PR #1767, opened 2021-05-25
[docs] Update `GridApi` descriptions

### LINKED ISSUE #1705: [docs] The description of apiRef is grammatically incorrect

## Current Behavior 😯

https://github.com/mui-org/material-ui-x/blob/e96be53e62d6a7f469a57f72cf3af3225972f012/packages/grid/_modules_/grid/models/api/gridParamsApi.ts#L9

https://github.com/mui-org/material-ui-x/blob/e96be53e62d6a7f469a57f72cf3af3225972f012/packages/grid/_modules_/grid/models/api/gridCsvExportApi.ts#L11

## Expected Behavior 🤔

```diff
-* Get the cell value of a row and field. 
+* Returns the cell value of a row and field. 
```

```diff
-* Export the grid data as CSV.
+* Exports the grid data as CSV.
```

## Context 🔦

This issue was spotted in #1620. @m4theushw has done it correctly which surfaces the inconsistency and consequently how the previous approach was not correct.

m4theushw:
### Breaking changes

- [XGrid] The `getEditCellValueParams` method was removed from the `apiRef`. Use `getEditCellPropsParams` instead.

  ```diff
  -const { value } = apiRef.current.getEditCellValueParams(id, field);
  +const { props: { value } } = apiRef.current.getEditCellPropsParams(id, field);
  ```

---

Fixes #1705

Also adds missing prefixes, like https://github.com/mui-org/material-ui-x/issues/1705#issuecomment-844252287.


packages/grid/_modules_/grid/models/api/gridEditRowApi.ts
```diff
@@ -4,68 +4,75 @@ import { GridRowId } from '../gridRows';
...
+   * @returns {GridEditCellPropsParams} The params.
    */
   getEditCellPropsParams: (rowId: GridRowId, field: string) => GridEditCellPropsParams;
   /**
    * Get the edit cell value params.
    * @param rowId
    * @param field
+   * @ignore - do not document
    */
   getEditCellValueParams: (rowId: GridRowId, field: string) => GridEditCellValueParams;
```
m4theushw, line 66: Not used. Let's remove it?

packages/grid/_modules_/grid/models/api/gridVirtualizationApi.ts
```diff
@@ -9,36 +9,41 @@ import { Optional } from '../../utils/utils';
...
-   * Get the current scroll position.
+   * Returns the current scroll position.
+   * @returns {GridScrollParams} The scroll positions.
    */
   getScrollPosition: () => GridScrollParams;
   /**
-   * Refresh the viewport cells according to the scroll positions
-   * @param forceRender
+   * Refreshes the viewport cells according to the scroll positions
+   * @param {boolean} forceRender If `true` forces a rerender. By default, it is `false`.
```
oliviertassinari, line 46: We add a comma in the core, we might do the same:

```suggestion
   * @param {boolean} forceRender If `true`, forces a rerender. By default, it is `false`.
```

packages/grid/_modules_/grid/models/api/gridEditRowApi.ts
```diff
@@ -4,68 +4,75 @@ import { GridRowId } from '../gridRows';
...
   /**
    * Commit the cell value changes to update the cell value.
    * @param update
+   * @ignore - do not document
    */
   commitCellChange: (params: GridEditCellPropsParams) => void;
```
m4theushw, line 72: I don't see a rationale to expose this method in the API. Although used in [this](https://material-ui.com/components/data-grid/editing/#edit-using-external-button) demo, the same thing can be done with `apiRef.current.setCellValue`.

packages/grid/_modules_/grid/models/api/gridPreferencesPanelApi.ts
```diff
@@ -0,0 +1,16 @@
...
+ * The preferences panel API interface that is available in the grid [[apiRef]].
+ */
+export interface GridPreferencesPanelApi {
+  /**
+   * Displays the preferences panel. The `newValue` argument controls the content of the panel.
+   * @param {GridPreferencePanelsValue} newValue The panel to open. Use `"filters"` or `"columns"`.
+   */
+  showPreferences: (newValue: GridPreferencePanelsValue) => void;
+  /**
+   * Hide the preferences panel.
```
oliviertassinari, line 13:
```suggestion
   * Hides the preferences panel.
```

packages/grid/_modules_/grid/models/api/gridEditRowApi.ts
oliviertassinari: @m4theushw "commit" means that the value is stored and will be taken part of the ctrl + z, ctrl + alt + z history feature. set value means that the value is only locally updated without updating the underlying rows. I would personally do the opposite, kill `setCellValue`.

m4theushw: If there would be a history feature then makes sense to have the "commit". I added the ignore to `setCellValue` for now.

Issue #1403, opened 2021-04-13
[RFC] Use event.defaultMuiPrevented to prevent default events


oliviertassinari:
## Summary 💡

This RFC highlights issues with the **current solution** and explores an **alternative** with fewer drawbacks. The conversation started with https://github.com/mui-org/material-ui-x/issues/1381.

## Motivation 🔦

The cell edit feature is soon to be released with the introduction of a new concept: `event.stopPropagation()` as a means to prevent the default behavior of the component.

https://github.com/mui-org/material-ui-x/blob/ca286ad1aa388a7ee8b6cb3609124f583679a45d/packages/grid/x-grid/src/tests/events.XGrid.test.tsx#L221-L230

I would like to highlight 5 important problems that the approach creates (before we release it).

1. The semantic is wrong (IMHO), but to be unbaised: confusing. `event.stopPropagation()` in the DOM and React stops the bubbling of the event to the parents but has **no impacts** on the element it listens to. It keeps calling the other listeners on the currentTarget element. In order to stop the other listeners to trigger on the same element, developers have to use `event.stopImmediatePropagation` in the DOM, which doesn't exist in the React realm.
2. `event.stopPropagation()` only works for events that are React synthetic events. The approach doesn't work on DOM events because there are no equivalents to `event.defaultPrevented` like React has (event.isPropagationStopped). So we can't scale the solution to all our use cases.
3. `event.stopPropagation()` stops the bubbling of the event. Meaning, if you want to handle the event on the parent, but still want to prevent the default behavior. You can't, you are screwed. Proof: https://codesandbox.io/s/material-demo-forked-1sstp?file=/demo.js. This is why it's commonly considered a practice to [void](https://css-tricks.com/dangers-stopping-event-propagation/). I agree that calling stopPropagation on the events the data grid is going to use for the cell editing makes sense, it's only when we want to disable the feature that it doesn't make any sense.
4. This is inconsistent. There is a prior-art in the main repository for using a different solution for the very same problem. It's currently deployed on the SwipeableDrawer and the Autocomplete. You can find it documented at https://next.material-ui.com/components/autocomplete/#events. When we introduced it, we thought about these problems, event.stopPropagation was an option we ignored because of the downsides I have listed before.
5. The parent callbacks are called before the component's callback. It's commonly the opposite of how callbacks run so that once the component yields to the parent, all the necessary actions have already been performed and the parent can continue.

## Proposed Solution 🧾

[...]

PR #1772, opened 2021-05-26
[DataGrid] Add support for edit components that use portal

### LINKED ISSUE #1439: [DataGrid] Cell editing with custom <Select> not working

## Current Behavior 😯
Custom select field is not showing meun item lists. The cell editing is changed to view mode on clicking Select field.

## Expected Behavior 🤔
Expected to show the list of menu items when clicking the Select field in cell edit mode

When cell is in edit mode, the Select field should show the dropdown select menu item options for user to select.

## Steps to Reproduce 🕹
I have created a sandbox here. https://codesandbox.io/s/material-demo-forked-eolox?file=/demo.js

m4theushw:
### Breaking changes

- [DataGrid] The `onEditCellChangeCommitted` prop won't called with an event when committing changes by clicking outside the cell.

---

Fixes #1439 

Due to the fact that we relay on the `blur` event to detect if the user clicked outside the cell, we need to ignore when the focus changes from the root cell element to one of its children. With components that use a portal, this logic doesn't work because there're in a different part of the DOM. This PR changes this behavior to listen for all `click` events in the document. To filter out events that were inside the cell, we check if before the click a `mouseup` event was fired. If it does, then the click was inside the cell or inside a portaled element and we shouldn't commit the changes yet. This is the same logic that ClickAwayListener uses. The reason to use `mouseup` instead of `click` is to be able to use the Select component. 

https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/
https://codesandbox.io/s/material-demo-forked-tprys?file=/demo.js


test/e2e/fixtures/DataGrid/PortaledEditCell.tsx
```diff
@@ -0,0 +1,50 @@
...
+      brand: 'Puma',
+    },
+  ],
+  columns: [
+    {
+      field: 'brand',
+      width: 100,
+      editable: true,
+      renderEditCell: ({ id, field, api }) =>
+        ReactDOM.createPortal(
```
oliviertassinari, line 26: This test case looks strange. 

- Why not use the `<Portal>` component as in the other test cases?
- Why not use a regular test case, I mean why does it need a more costly e2e test? Is it about the click event sequence? (Sebastian added a helper in the core for it)
- Why not use the Select component directly?

oliviertassinari (commented):
It doesn't seem to be working, the edit mode is not excited when clicking another cell: https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#cell-editing

test/e2e/fixtures/DataGrid/PortaledEditCell.tsx
m4theushw: > Why not use the <Portal> component as in the other test cases?

Updated. I didn't remember of it.

> Why not use a regular test case, I mean why does it need a more costly e2e test? Is it about the click event sequence? (Sebastian added a helper in the core for it)

Yes, I chose to do an e2e test because of the click sequence. The helper is not merged yet: https://github.com/mui-org/material-ui/pull/25741/files#diff-0468d4d0b40c7b10080399740c336c59638e96c9e10196e9537ea4fda5a80646. 

> Why not use the Select component directly (for a second test case of integration)?

I updated the PR to do an e2e test with Select and a regular test case with Portal.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -45,6 +47,21 @@ export function useGridEditRows(apiRef: GridApiRef) {
...
+  const handleDocumentClick = useEventCallback(() => {
+    const params = lastEditedCell.current;
+    if (insideCell.current || !params) {
+      insideCell.current = false;
+      return;
+    }
+    const cellCommitParams = apiRef.current.getEditCellPropsParams(params.id, params.field);
+    if (!cellCommitParams.props.error) {
+      // We commit the change when there are no error
+      apiRef.current.publishEvent(GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED, cellCommitParams);
```
m4theushw, line 61: I can't pass an event because `publishEvent` requires a `SyntheticEvent` (because of the `isPropagationStopped`) and I only have a native event.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -84,13 +101,18 @@ export function useGridEditRows(apiRef: GridApiRef) {
         api: apiRef.current,
       });
 
+      const doc = ownerDocument(apiRef.current.rootElementRef!.current as HTMLElement);
+      if (mode === 'view') {
+        doc.removeEventListener('click', handleDocumentClick);
```
m4theushw, line 106: I would also attach the listener here when changing to the edit mode but by doing that it [broke](https://www.argos-ci.com/mui-org/material-ui-x/builds/1424) the regression tests. The click done to switch between tests is incorrectly caught as a click to exit from the edit mode.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -251,20 +276,11 @@ export function useGridEditRows(apiRef: GridApiRef) {
...
-        // We commit the change when there are no error
-        apiRef.current.publishEvent(GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED, cellCommitParams, event);
-      }
-      apiRef.current.publishEvent(GRID_CELL_EDIT_EXIT, params, event);
-    },
-    [apiRef],
-  );
+  const handleCellMouseUp = React.useCallback((params: GridCellParams, event: React.MouseEvent) => {
+    if (event.isPropagationStopped() || params.cellMode === 'edit') {
+      insideCell.current = true;
```
m4theushw, line 281: Here is the magic. If a `mouseup` came from the cell, then the next click should be treated as it was inside the cell too.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
dtassone: Could we make this generic? Otherwise, there would be portal fixes in a lot of features, and each hook will attach and detach events...
If we add an event listener on the doc, and route it to publish some kind of `CLICK_OUTSIDE_ROOT`
And isolate that into a usePortal hook, it would be really nice.
Then each feature, can hook themselves to some kind of click on portal event...
What do you think?

m4theushw: The click we're interested may occur inside the root as well. What we need to check is if it happened outside the cell that is currently being edited. Since it's a very specific situation, I don't think we need to abstract it. This new event would have to be named like `GRID_EDIT_CELL_CLICK_OUTSIDE` and it would be attached only to the edit cell, which is the same thing I doing here but without exposing an event.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
```diff
@@ -147,31 +147,36 @@ export const useGridKeyboard = (
 
   const handleCellKeyDown = React.useCallback(
     (params: GridCellParams, event: React.KeyboardEvent) => {
-      if (!isGridCellRoot(event.target as HTMLElement)) {
+      // The target is not an element when triggered by a Select inside the cell
+      // See https://github.com/mui-org/material-ui/issues/10534
+      if (event.target instanceof HTMLElement && !isGridCellRoot(event.target)) {
```
oliviertassinari, line 152: doesn't work between different documents. 
```suggestion
      if (event.target.nodeType === 1 && !isGridCellRoot(event.target)) {
```

See https://github.com/mui-org/material-ui/blob/33850e52ca32787fbc02b64d3225bac2d451612c/packages/material-ui-utils/src/HTMLElementType.ts#L19

test/e2e/fixtures/DataGrid/SelectEditCell.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { id, value, api, field } = props;
+
+  const handleChange = (event) => {
+    const editProps = { value: event.target.value };
+    api.commitCellChange({ id, field, props: editProps });
+    api.setCellMode(id, field, 'view');
+  };
+
+  return (
+    <Select value={value} onChange={handleChange}>
```
oliviertassinari, line 16: How about we open the select as soon as it enters in edit mode? You can find this UX on AG Grid or Airtable. It feels like it currently requires two extra clicks that could be save in the majority of the cases.

I first saw it in test/e2e/index.test.ts.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
dtassone: what if we add a renderCell that use portals? Ie a select component in the grid view mode

test/e2e/fixtures/DataGrid/SelectEditCell.tsx
m4theushw: Updated the e2e test and also the other PR that uses a Select.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: > what if we add a renderCell that use portals? Ie a select component in the grid view mode

We wouldn't have problems because clicking outside the cell in view mode doesn't do anything besides moving the focus.

m4theushw:
I changed the way I detect if the click is outside the cell. For that, I created a new `GRID_CELL_FOCUS_OUT` event which is published by the `useGridFocus` hook. Different from a blur, it checks if the click came from a portal. @dtassone Could you take a look? I linked in the description a CodeSandbox with a Select.

Troy96:
@m4theushw Awesome! I have been waiting for this for some time. Can we expect this to go live and be available in coming few days?

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -33,7 +33,12 @@ export const GridRow = (props: GridRowProps) => {
   const publish = React.useCallback(
     (eventName: string) => (event: React.MouseEvent) => {
       // Ignore portal
-      if (!event.currentTarget.contains(event.target as HTMLElement)) {
+      // The target is not an element when triggered by a Select inside the cell
+      // See https://github.com/mui-org/material-ui/issues/10534
+      if (
+        (event.target as any).nodeType === 1 &&
+        !event.currentTarget.contains(event.target as Element)
```
dtassone, line 40: Could we wrap that into a function such as `isTargetWithinElement` or `isTargetFromSubTree` or ?
Make the code more readable

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -97,10 +98,23 @@ export const GridCell = React.memo((props: GridCellProps) => {
...
+
   const publish = React.useCallback(
     (eventName: string) => (event: React.SyntheticEvent) => {
       // Ignore portal
-      if (!event.currentTarget.contains(event.target as HTMLElement)) {
+      // The target is not an element when triggered by a Select inside the cell
+      // See https://github.com/mui-org/material-ui/issues/10534
+      if (
+        (event.target as any).nodeType === 1 &&
+        !event.currentTarget.contains(event.target as Element)
```
dtassone, line 116: reuse it here, remove comments and put it in the helper function

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
```diff
@@ -81,31 +81,36 @@ export const useGridKeyboard = (apiRef: GridApiRef): void => {
 
   const handleCellKeyDown = React.useCallback(
     (params: GridCellParams, event: React.KeyboardEvent) => {
-      if (!isGridCellRoot(event.target as HTMLElement)) {
+      // The target is not an element when triggered by a Select inside the cell
+      // See https://github.com/mui-org/material-ui/issues/10534
+      if ((event.target as any).nodeType === 1 && !isGridCellRoot(event.target as Element)) {
```
dtassone, line 86: same here

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
dtassone: unless there is a select dropdown within it

m4theushw: Here's a CodeSandbox with a Select in `renderCell`: https://codesandbox.io/s/material-demo-forked-kc7id?file=/demo.js

We only have a bug if a portaled component is used in `renderEditCell`.

packages/grid/_modules_/grid/components/GridRow.tsx
m4theushw: This will trigger the discussion about when to create helpers or not. 😂 I would prefer to keep it as it is because it's only used in two places now. But if we start using more I will definitely create a helper for it.

dtassone: That logic is pretty complex so I would definitely abstract it. 
Normally you can use a rule of 3, here it's repeated twice already, + complexity and readability benefits => clean code

m4theushw: If @oliviertassinari and @DanailH agree we can move this to a helper. I'm against it for the following reasons:
1. This conditional is only used in two places
2. We had past discussions about helpers and we didn't reach an agreement
3. 100% times I see a helper I go to its definition to see what it does, because I don't trust the name
4. It makes the code cleaner but the developer has to navigate to more files to understand 
5. By the name of the helper I don't know if `currentTarget` or `target` is used.

DanailH: In that particular case, I'm more in favour of keeping it like it is. If you name the helper `isTargetWithinElement` I would still need to check its implementation to understand what it does. I'm ok with helpers like `isSpaceKey` for example as it is clear what the helper does.

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -28,6 +32,7 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
           focus: { cell: { id, field }, columnHeader: null },
         };
       });
+      // TODO replace with constant
       apiRef.current.publishEvent('cellFocusChange');
```
DanailH, line 36: Don't we have that event already defined?

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -80,6 +90,48 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
...
   }, [logger, setGridState]);
 
+  const handleCellMouseUp = React.useCallback(
+    (params: GridCellParams) => {
+      const { cell } = apiRef.current.getState().focus;
+      if (!cell) {
+        return;
+      }
+
+      if (params.id === cell.id && params.field === cell.field) {
```
DanailH, line 100: Can we have a mismatch with the cell if we don't check both the `id` and the `field`?

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -80,6 +90,48 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
...
+  const handleDocumentClick = React.useCallback(
+    (event) => {
+      const isInsideFocusedCell = insideFocusedCell.current;
+      insideFocusedCell.current = false;
+
+      const { cell } = apiRef.current.getState().focus;
+      if (!cell || isInsideFocusedCell) {
+        return;
+      }
+
```
DanailH, line 116: Will this work?

```suggestion
      const { cell } = apiRef.current.getState().focus;
      if (!cell || insideFocusedCell.current) {
        return;
      }
      
      insideFocusedCell.current = false;
```

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
m4theushw: Yes, this handler is called when any cell is clicked. I just want `insideFocusedCell.current` to be true when the currently focused cell is clicked.

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
m4theushw: There's a problem with setting `insideFocusedCell.current` to false after checking if there's a cell focused. If there's no cell focused it will stay true, which can prevent the next click from working properly. I borrowed this logic from ClickAwayListener:

https://github.com/mui-org/material-ui/blob/aafb54acb7de52299f4c8cb14f0d4adcaac5cb90/packages/material-ui/src/ClickAwayListener/ClickAwayListener.tsx#L104-L105

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
m4theushw: There's only a `GRID_CELL_FOCUS` but it's mapped to the `focus` event.

packages/grid/_modules_/grid/components/GridRow.tsx
oliviertassinari: I'm all 💯 for https://github.com/mui-org/material-ui-x/pull/1772/files#r649207549. Some of the arguments are biased, some are not, but the arguments resonate with me. I also **never** trust a module I don't know by heart (nor I think anyone should ever), having to check the content of small helpers is mentally draining. Two repetitions are not a pattern.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -302,11 +310,12 @@ export function useGridEditRows(apiRef: GridApiRef) {
   );
 
   useGridApiEventHandler(apiRef, GRID_CELL_KEYDOWN, handleCellKeyDown);
-  useGridApiEventHandler(apiRef, GRID_CELL_BLUR, handleCellBlur);
   useGridApiEventHandler(apiRef, GRID_CELL_MOUSE_DOWN, preventTextSelection);
   useGridApiEventHandler(apiRef, GRID_CELL_DOUBLE_CLICK, handleDoubleClick);
   useGridApiEventHandler(apiRef, GRID_CELL_EDIT_ENTER, handleEnterEdit);
   useGridApiEventHandler(apiRef, GRID_CELL_EDIT_EXIT, handleExitEdit);
+  useGridApiEventHandler(apiRef, GRID_CELL_FOCUS_OUT, handleClickOutside);
```
oliviertassinari, line 317: This is suspicious. `GRID_CELL_FOCUS_OUT` if fire for click away and focus events. I don't get the naming nor the event. Why not?
```suggestion
  useGridApiEventHandler(apiRef, GRID_CELL_CLICK_OUTSIDE, handleClickOutside);
```

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -80,6 +90,48 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
...
 
+  const handleCellMouseUp = React.useCallback(
+    (params: GridCellParams) => {
+      const { cell } = apiRef.current.getState().focus;
+      if (!cell) {
+        return;
+      }
+
+      if (params.id === cell.id && params.field === cell.field) {
+        insideFocusedCell.current = true;
```
oliviertassinari, line 101: Did you copy the solution I was proposing for fixing ClickAwayListener with the Select? 😁

oliviertassinari (commented):
This demo behaves differently https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button. Clicking twice on a cell exit the edit mode

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: I updated the handler name. `handleClickOutside` was a inheritance from when I was adding a listener to the `document`. 

`GRID_CELL_CLICK_OUTSIDE` is not good name because the focus out can also be triggered pressing <kbd>Esc</kbd> or programmatically calling `setCellFocus`.

oliviertassinari:
@m4theushw On https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button. We were previously able to turn an unlimited amount of cells in edit mode. It's no longer the case. The first cell exits as soon as we click another.

packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
```diff
@@ -294,4 +371,50 @@ describe('<XGrid /> - Edit Rows', () => {
...
+            renderEditCell: () => <input type="checkbox" />,
+          },
+        ]}
+      />,
+    );
+    const cell = getCell(0, 0);
+    expect(cell).not.to.have.class('MuiDataGrid-cellEditing');
+    fireEvent.doubleClick(cell);
+    expect(cell).to.have.class('MuiDataGrid-cellEditing');
+    const checkbox = screen.getByRole('checkbox', { key: ' ' });
```
oliviertassinari, line 416: ?

```suggestion
    const checkbox = screen.getByRole('checkbox');
```

oliviertassinari:
https://github.com/mui-org/material-ui-x/pull/1772#issuecomment-862743018 Ok, here we go: #1403.

m4theushw:
> On https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button. We were previously able to turn an unlimited amount of cells in edit mode. It's no longer the case. The first cell exits as soon as we click another.

@oliviertassinari We can't edit multiple cells at the same time anymore because only one cell can have focus.

packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
m4theushw: I forgot to remove the `key`. My initial idea was to test pressing <kbd>Space</kbd> on a checkbox, because doing that triggers a `click` but not a `mouseup`. See https://bugzilla.mozilla.org/show_bug.cgi?id=1120106

oliviertassinari (approved):

oliviertassinari:
I have fixed the regression on https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button and added a regression test case. The fix is relatively simple but there is quite a lot of boilerplate that makes reading the diff a bit more challenging. I guess it's the cost of scaling with an event system, so fair enough.

m4theushw:
@oliviertassinari Looks good to me. 

Some points we could bring to #1824:

- Now we use `event.stopPropagation()` and `event.defaultMuiPrevented`. Maybe it's better to migrate everything to the second option to have only one way to stop propagation.
- There's a long list of events to stop propagation to make https://deploy-preview-1772--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button work. We could add a prop to allow users to completely opt-out from the mouse/keyboard interaction.
- By allowing to edit multiple cells at the same time, we kinda broke the focus logic. Only one cell could have focus and only the focused cell could be in edit mode.

oliviertassinari:
> only the focused cell could be in edit mode.

Why? I thoughts it was important for row edit to have all the cells of the row in edit mode

DanailH (approved):

docs/src/pages/components/data-grid/editing/StartEditButtonGrid.tsx
```diff
@@ -100,11 +100,11 @@ export default function StartEditButtonGrid() {
...
-  // Prevent from committing on blur
-  const handleCellBlur = React.useCallback(
-    (params, event?: React.SyntheticEvent) => {
-      if (params.cellMode === 'edit') {
-        event?.stopPropagation();
+  // Prevent from committing on focus out
+  const handleCellFocusOut = React.useCallback(
+    (params, event?: MouseEvent | React.SyntheticEvent) => {
+      if (params.cellMode === 'edit' && event) {
+        (event as any).defaultMuiPrevented = true;
```
dtassone, line 107: you could do an interface to make this a bit better
```tsx
interface MuiEvent extends React.SyntheticEvent {
  defaultMuiPrevented?: boolean
}
``` 
Also when you publish events you could use `as MuiEvent`

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -31,6 +31,10 @@ import { GridColumnResizeParams } from './params/gridColumnResizeParams';
 import { GridColumnVisibilityChangeParams } from './params/gridColumnVisibilityChangeParams';
 import { GridClasses } from './gridClasses';
 
+export type MuiEvent<E> = (E | React.SyntheticEvent) & {
```
m4theushw, line 34: @dtassone I added the `E` type variable so we can reuse it with different native events if necessary.

dtassone: Nice, but why not do an interface?

m4theushw: Interfaces can't extend union types.

https://stackoverflow.com/questions/56085306/error-message-an-interface-can-only-extend-an-object-type-or-intersection-of-o

m4theushw:
> Why? My assumption was that for row edit, we need to have all the cells of the row in edit mode. It was designed with this constraint in mind so far (I believe).

@oliviertassinari Probably we'll have problems implementing the row editing, because clicking outside a cell triggers `GRID_CELL_FOCUS_OUT` but this other cell could be in the same row. The cell editing is tied to the focus logic, which only allows one cell to receive focus. Maybe we need to introduce the row focus.

packages/grid/_modules_/grid/models/gridOptions.tsx
dtassone: because you shouldn't do a union type...
I would have done
```
interface MuiEvent<E> extends React.SyntheticEvent<E> {
...
}
```
But while typing, I noticed your `E | SyntheticEvent`. E is any, but yeah it could work anywhere 🤔

m4theushw: Yeah, but this event when triggered by the `document` is a native event. The union needs to be there.

oliviertassinari:
> clicking outside a cell triggers GRID_CELL_FOCUS_OUT but this other cell could be in the same row. The cell editing is tied to the focus logic, which only allows one cell to receive focus. Maybe we need to introduce the row focus.

@m4theushw I wouldn't worry too much about this. It was also the case when relying on the native blur event. In the core, there are a couple of places where we solve this by waiting with a setTimeout() to see if there isn't another focus event, before handling the blur.

m4theushw:
> @m4theushw Awesome! I have been waiting for this for some time. Can we expect this to go live and be available in coming few days?

@Troy96 Yes, it will be included in the next release which might be this week or next.

PR #1752, opened 2021-05-23
[docs] Fix 404 link

oliviertassinari:
Fix the broken link in https://material-ui.com/components/data-grid/pagination/#rendering.

PR #1774, opened 2021-05-27
[docs] Remove redundant customizable pagination section

ZeeshanTamboli:
There is already a customization section below for which the link was fixed in #1752 and this section seemed documenting the same thing.
Also fixed few small typos. 

Preview: https://deploy-preview-1774--material-ui-x.netlify.app/components/data-grid/pagination/


dtassone (approved):

packages/grid/_modules_/grid/hooks/utils/useResizeContainer.ts
```diff
@@ -34,7 +34,7 @@ export function useResizeContainer(apiRef): (size: ElementSize) => void {
             'The grid displays with a height of 0px.',
             '',
             'You can find a solution in the docs:',
-            'https://material-ui.com/components/data-grid/rendering/#layout',
+            'https://material-ui.com/components/data-grid/layout',
```
oliviertassinari, line 37:
```suggestion
            'https://material-ui.com/components/data-grid/layout/',
```

packages/grid/_modules_/grid/hooks/utils/useResizeContainer.ts
```diff
@@ -46,7 +46,7 @@ export function useResizeContainer(apiRef): (size: ElementSize) => void {
             'The grid displays with a width of 0px.',
             '',
             'You can find a solution in the docs:',
-            'https://material-ui.com/components/data-grid/rendering/#layout',
+            'https://material-ui.com/components/data-grid/layout',
```
oliviertassinari, line 49: URLs should have a trailing slash (`<link rel="canonical" href="https://material-ui.com/components/data-grid/layout/"/>`)

```suggestion
            'https://material-ui.com/components/data-grid/layout/',
```

oliviertassinari: But in the core, we never put hard-coded URLs like this in the source. We use redirections: https://github.com/mui-org/material-ui/blob/8cc7dd0568adde2b8c23804fabdadbd7cfff21d6/docs/public/_redirects#L15

ZeeshanTamboli: > Why, because this is a 404 for developers that have a previous version of the library installed (high version fragmentation)

I didn't get this. So suppose if we put in `_redirects`:
```
/r/data-grid/layout https://material-ui.com/components/data-grid/layout/
```
And in code:
```js
'You can find a solution in the docs:'
'https://material-ui.com/components/r/data-grid/layout',
```
Does this mean that if developers are in the old version, they will get warning in console with this new link `https://material-ui.com/components/r/data-grid/layout'`? How come? If they are using old version won't it will still show the old 404 link `data-grid/rendering/#layout`?

PR #1778, opened 2021-05-28
[DataGrid] Improve `headerClassName` type

### LINKED ISSUE #274: [RFC] Only allow a string for headerClassName
For simplicity, I think that `headerClassName` could only accept a string. Developers are used to this constraint to React. I would also argue they we have thought them to use the `clsx` helper.

DanailH:
### Breaking changes

- [DataGrid] Improve `headerClassName` type

  ```diff
  -cellClassName?: string | string[] | (params: GridCellParams) => string;
  +cellClassName?: string | (params: GridCellParams) => string;
  ```

   ```diff
   -headerClassName?: string | string[];
   +headerClassName?: string | (params: GridColumnHeaderParams) => string;
   ```

---

Fixes #274


packages/grid/data-grid/src/tests/columnHeaders.DataGrid.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <DataGrid
+            {...baselineProps}
+            columns={[{ field: 'brand', headerClassName: () => 'foobar' }]}
+          />
+        </div>,
+      );
+      expect(getColumnHeaderCell(0)).to.have.class('foobar');
+    });
+  });
+});
```
oliviertassinari, line 53: I'm confused. It's a TypeScript test but framed as a browser test. The browser test suite will never run it. I assumed you meant to write a browser test. How did you manage to assert that this test is running correctly?

packages/grid/_modules_/grid/models/gridColumnHeaderClass.ts
```diff
@@ -0,0 +1,11 @@
+import { GridColumnHeaderParams } from './params/gridColumnHeaderParams';
+
+/**
+ * A function used to process headerClassName params.
+ */
+export type GridColumnHeaderClassFn = (params: GridColumnHeaderParams) => string;
```
oliviertassinari, line 6: Why is this public?
```suggestion
type GridColumnHeaderClassFn = (params: GridColumnHeaderParams) => string;
```

packages/grid/_modules_/grid/models/gridColumnHeaderClass.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * The union type representing the [[GridColDef]] column header class type.
+ */
+export type GridColumnHeaderClassNamePropType = string | GridColumnHeaderClassFn;
```
oliviertassinari, line 11: Why is inlining the type not enough? It seems that we never use it more than once.
```suggestion
```

packages/grid/_modules_/grid/models/gridCellClass.ts
```diff
@@ -1,11 +1,11 @@
 import { GridCellParams } from './params/gridCellParams';
 
 /**
- * A function used to process cellClassParams.
+ * A function used to process cellClassName params.
  */
-export type GridCellClassFn = (params: GridCellParams) => string | string[];
+export type GridCellClassFn = (params: GridCellParams) => string;
```
oliviertassinari, line 6: Why does it need to be public?

```suggestion
type GridCellClassFn = (params: GridCellParams) => string;
```

packages/grid/_modules_/grid/models/gridCellClass.ts
```diff
@@ -1,11 +1,11 @@
...
 
 /**
  * The union type representing the [[GridColDef]] cell class type.
  */
-export type GridCellClassNamePropType = string | string[] | GridCellClassFn;
+export type GridCellClassNamePropType = string | GridCellClassFn;
```
oliviertassinari, line 11: It's never used more than once. Developers can use `GridColDef['cellClassName']` if they really need to access it. Shouldn't we inline it?

```suggestion
```

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -123,8 +124,19 @@ export const GridColumnHeaderItem = React.memo(
       [publish],
     );
 
+    const classNames = [classes?.columnHeader];
+
+    if (column.headerClassName) {
+      classNames.push(
+        clsx(
+          isFunction(column.headerClassName)
```
oliviertassinari, line 132: Shouldn't we drop this `isFunction` helper? What problem does it solve? I fail to understand it, there is already a built-in primitive in JavaScript it does this.

It also seems to

1. creates friction when writing new code, we have to figure out the write import path, base on the location of the file we want to use it into
2. creates an indirection that prevents clearly understand what it sets as a constraint on the public API.

```suggestion
          typeof column.headerClassName === 'function'
```

packages/grid/_modules_/grid/models/gridCellClass.ts
dtassone: why does it need to be private? If it is public, ppl can import it and type their function to make sure and avoid runtime errors

oliviertassinari: They can already do it `GridColDef['cellClassName']`

packages/grid/data-grid/src/tests/columnHeaders.DataGrid.spec.tsx
DanailH: Ou, that is weird. I took it from a similar test for the `cellClassName`. I'll check what is going on.

packages/grid/_modules_/grid/models/gridCellClass.ts
DanailH: I guess that applies to all the other comments - I followed the pattern done for the `cellClassName`. If we go for the not exporting it option then I'll change it for both cases.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
DanailH: `isFunction` seems to be used in quite a lot of places already. We had a similar discussion I think a couple of months ago regarding potentially unneeded helper functions. We never reached an agreement regarding this. We can always reopen the discussion and have a decision this time since it is inevitable that we will pick it up again at some point.

dtassone: I'm in favour of keeping it, it's an easy helper. And it does more than the `typeof...`. I do not understand the frictions.
image

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -123,8 +124,19 @@ export const GridColumnHeaderItem = React.memo(
       [publish],
     );
 
+    const classNames = [classes?.columnHeader];
+
+    if (column.headerClassName) {
+      classNames.push(
+        clsx(
```
m4theushw, line 131: You can drop the `clsx` if we're going to remove support for arrays.

packages/grid/data-grid/src/tests/columnHeaders.DataGrid.spec.tsx
oliviertassinari: I wasn't clear. The extension of the file determines if it's a TypeScript test or a regular test case. spec.tsx vs. test.tsx

packages/grid/_modules_/grid/models/gridCellClass.ts
oliviertassinari: It's not really important. I have raised the point as it seems to be a quick win. What draw my attention is that it's 1. inconsistent (we don't export a type for the each property of `GridColDef`, e.g. getValue ) 2. doesn't seem easy for a developer to discover (felt that the only way us to read the source, which developer's should have to). 3. the use case seems marginal with already an alternative.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
oliviertassinari: > It does more than the `typeof`

@dtassone Your example is demonstrating another value of removing the `isFunction` helper. it's telling you that the block is a dead code and should be removed.

I could find documentation about it in the documentation of TypeScript: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

> having to define a function to figure out if a type is a primitive is kind of **a pain**.

packages/grid/data-grid/src/tests/columnHeaders.DataGrid.spec.tsx
DanailH: Ok yes now I saw it, I added the wrong extension. I ran the tests and all passed and thought that it was ok since the test was fairly trivial.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
DanailH: Removed

dtassone (approved):

m4theushw (approved):

PR #1785, opened 2021-05-29
[DataGrid] Add `valueParser` to parse values entered by the user

### LINKED ISSUE #1118: [DataGrid] Filter does not work in conjunction with valueFormatter

## Current Behavior 😯

The column filter does not work if the value of the column was changed by the valueFormatter.
The search is still based on the field value.

## Expected Behavior 🤔

Values should be searched after valueFormatter.
A possibility to change the search would also be good if cells are specially rendered.

## Steps to Reproduce 🕹

https://codesandbox.io/s/vibrant-surf-2sq09?file=/src/demo.js
Try to search for the values A,B or C.

m4theushw:
Closes #1118

I'm not sure about the signature of `valueParser`. It's called with the user-entered value and the cell params. However, when applying it to the filter value I don't have a cell, so no params.

### DIFF
--- a/docs/src/pages/components/data-grid/columns/ValueParserGrid.js
+++ b/docs/src/pages/components/data-grid/columns/ValueParserGrid.js
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { DataGrid, GridValueFormatterParams } from '@material-ui/data-grid';
+import { DataGrid } from '@material-ui/data-grid';
 
 const rows = [
   {
@@ -16,7 +16,7 @@ const rows = [
   },
 ];
 
-export default function ValueFormatterGrid() {
+export default function ValueParserGrid() {
   return (
     <div style={{ height: 300, width: '100%' }}>
       <DataGrid
@@ -27,8 +27,8 @@ export default function ValueFormatterGrid() {
             field: 'date',
             headerName: 'Year',
             width: 150,
-            valueFormatter: (params: GridValueFormatterParams) =>
-              (params.value as Date).getFullYear(),
+            valueFormatter: (params) => params.value.getFullYear(),
+            valueParser: (value) => new Date(Number(value), 0, 1),
           },
         ]}
       />
--- a/docs/src/pages/components/data-grid/columns/columns.md
+++ b/docs/src/pages/components/data-grid/columns/columns.md
@@ -122,23 +122,26 @@ The value generated is used for filtering, sorting, rendering, etc unless overri
 
 ### Value formatter
 
-The value formatters allow you to format values for display as a string.
-For instance, you can use it to format a JavaScript `Date` object.
+The value formatter allows you to convert the value before displaying it.
+Common use cases include converting a JavaScript `Date` object to a date string or a `Number` into a formatted number (e.g. "1.000,50").
 
-```tsx
-const columns: GridColDef[] = [
-  {
-    field: 'date',
-    headerName: 'Year',
-    valueFormatter: (params: ValueFormatterParams) =>
-      (params.value as Date).getFullYear(),
-  },
-];
-```
+In the following demo, a formatter is used to display only the year of a JavaScript `Date` object.
 
 {{"demo": "pages/components/data-grid/columns/ValueFormatterGrid.js", "bg": "inline"}}
 
-The value generated is used for filtering, sorting, rendering in the cell and outside, etc unless overridden by a more specific configuration.
+The value generated is only used for rendering purposes.
+Filtering and sorting will not relay on the formatted value.
+Use the [`valueParser`](/components/data-grid/columns#value-parser) to support filtering.
+
+### Value parser
+
+The value parser allows you to convert the user-entered value to another one used for filtering or editing.
+Common use cases include parsing date strings to JavaScript `Date` objects or formatted numbers (e.g. "1.000,50") into `Number`.
+It can be understood as the inverse of [`valueFormatter`](/components/data-grid/columns#value-formatter).
+
+In the following demo, the user can filter for a year while a `Date` is used internally.
+
+{{"demo": "pages/components/data-grid/columns/ValueParserGrid.js", "bg": "inline"}}
 
 ### Render cell
 
--- a/packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
@@ -78,7 +78,7 @@ export const GridRowCells = React.memo((props: RowCellsProps) => {
     }
 
     const editCellState = editRowState && editRowState[column.field];
-    let cellComponent: React.ReactElement | null = null;
+    let cellComponent: React.ReactNode = null;
 
     if (editCellState == null && column.renderCell) {
       cellComponent = column.renderCell(cellParams);
--- a/packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
@@ -171,13 +171,15 @@ export function useGridEditRows(apiRef: GridApiRef) {
 
   const setCellValue = React.useCallback(
     (params: GridEditCellValueParams) => {
+      const column = apiRef.current.getColumn(params.field);
+      const parsedValue = column.valueParser
+        ? column.valueParser(params.value, apiRef.current.getCellParams(params.id, params.field))
+        : params.value;
       logger.debug(
-        `Setting cell id: ${params.id} field: ${
-          params.field
-        } to value: ${params.value?.toString()}`,
+        `Setting cell id: ${params.id} field: ${params.field} to value: ${parsedValue?.toString()}`,
       );
       const rowUpdate = { id: params.id };
-      rowUpdate[params.field] = params.value;
+      rowUpdate[params.field] = parsedValue;
       apiRef.current.updateRows([rowUpdate]);
       apiRef.current.publishEvent(GRID_CELL_VALUE_CHANGE, params);
     },
--- a/packages/grid/_modules_/grid/models/gridFilterItem.ts
+++ b/packages/grid/_modules_/grid/models/gridFilterItem.ts
@@ -1,7 +1,7 @@
 export interface GridFilterItem {
   id?: number;
   columnField?: string;
-  value?: string;
+  value?: any;
   operatorValue?: string;
 }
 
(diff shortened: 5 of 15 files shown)


packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
```diff
@@ -78,7 +78,7 @@ export const GridRowCells = React.memo((props: RowCellsProps) => {
     }
 
     const editCellState = editRowState && editRowState[column.field];
-    let cellComponent: React.ReactElement | null = null;
+    let cellComponent: React.ReactNode = null;
```
m4theushw, line 81: https://github.com/mui-org/material-ui-x/issues/1118#issuecomment-844484211

packages/grid/_modules_/grid/models/gridFilterItem.ts
```diff
@@ -1,7 +1,7 @@
 export interface GridFilterItem {
   id?: number;
   columnField?: string;
-  value?: string;
+  value?: any;
```
m4theushw, line 4: https://github.com/mui-org/material-ui-x/pull/1722#discussion_r640636206

packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
```diff
@@ -254,4 +255,33 @@ describe('<XGrid /> - Edit Rows', () => {
...
+
+  it('should apply the valueParser before saving the value', () => {
+    const valueParser = stub().withArgs('62').returns(1962);
+    render(
+      <TestCase
+        columns={[
+          { field: 'brand', editable: true },
+          { field: 'year', editable: true, valueParser },
+        ]}
+      />,
```
oliviertassinari, line 267: We don't need the apiRef, we might want to avoid one attraction for clarity:

```suggestion
      <XGrid
        {...baselineProps}
        columns={[
          { field: 'brand', editable: true },
          { field: 'year', editable: true, valueParser },
        ]}
      />,
```

packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
```diff
@@ -254,4 +255,33 @@ describe('<XGrid /> - Edit Rows', () => {
...
+    expect(cell.querySelector('input')!.value).to.equal('62');
+
+    fireEvent.keyDown(input, { key: 'Enter' });
+    expect(cell).not.to.have.class('MuiDataGrid-cellEditing');
+    expect(cell).to.have.text('1962');
+    expect(valueParser.callCount).to.equal(1);
+    expect(valueParser.args[0][0]).to.equal('62');
+    expect(valueParser.args[0][1].id).to.equal(1);
+    expect(valueParser.args[0][1].field).to.equal('year');
+    expect(valueParser.args[0][1].value).to.equal(1961);
```
oliviertassinari, line 285: If it's possible using a `.deep.equal` or `.include`, we might want to use it to make the test case easier to read (self declaration of the JS structure).

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -574,4 +574,32 @@ describe('<DataGrid /> - Filter', () => {
...
+        columns={[
+          {
+            field: 'amount',
+            type: 'number',
+            valueParser: (value) => (value as number) / 100,
+          },
+        ]}
+        operatorValue={'='}
+        value={50}
+        field={'amount'}
```
oliviertassinari, line 600: if we want to follow the existing convention:

```suggestion
        operatorValue="="
        value={50}
        field="amount"
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -122,23 +122,26 @@ The value generated is used for filtering, sorting, rendering, etc unless overri
...
-The value generated is used for filtering, sorting, rendering in the cell and outside, etc unless overridden by a more specific configuration.
+The value generated is only used for rendering purposes.
+Filtering and sorting will not relay on the formatted value.
+Use the [`valueParser`](/components/data-grid/columns#value-parser) to support filtering.
+
+### Value parser
+
+The value parser allows you to convert the user-entered value to another one used for filtering or editing.
+Common use cases include parsing date strings to JavaScript `Date` objects or formatted numbers (e.g. "1.000,50") into `Number`.
+It can be understood as the inverse of [`valueFormatter`](/components/data-grid/columns#value-formatter).
```
oliviertassinari, line 140:
```suggestion
It can be understood as the inverse of [`valueFormatter`](/components/data-grid/columns/#value-formatter).
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -122,23 +122,26 @@ The value generated is used for filtering, sorting, rendering, etc unless overri
...
-];
-```
+In the following demo, a formatter is used to display only the year of a JavaScript `Date` object.
 
 {{"demo": "pages/components/data-grid/columns/ValueFormatterGrid.js", "bg": "inline"}}
 
-The value generated is used for filtering, sorting, rendering in the cell and outside, etc unless overridden by a more specific configuration.
+The value generated is only used for rendering purposes.
+Filtering and sorting will not relay on the formatted value.
+Use the [`valueParser`](/components/data-grid/columns#value-parser) to support filtering.
```
oliviertassinari, line 134:
```suggestion
Use the [`valueParser`](/components/data-grid/columns/#value-parser) to support filtering.
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -122,23 +122,26 @@ The value generated is used for filtering, sorting, rendering, etc unless overri
...
+
+### Value parser
+
+The value parser allows you to convert the user-entered value to another one used for filtering or editing.
+Common use cases include parsing date strings to JavaScript `Date` objects or formatted numbers (e.g. "1.000,50") into `Number`.
```
oliviertassinari, line 139: en-US format:

```suggestion
Common use cases include parsing date strings to JavaScript `Date` objects or formatted numbers (e.g. "1,000.50") into `Number`.
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -122,23 +122,26 @@ The value generated is used for filtering, sorting, rendering, etc unless overri
 
 ### Value formatter
 
-The value formatters allow you to format values for display as a string.
-For instance, you can use it to format a JavaScript `Date` object.
+The value formatter allows you to convert the value before displaying it.
+Common use cases include converting a JavaScript `Date` object to a date string or a `Number` into a formatted number (e.g. "1.000,50").
```
oliviertassinari, line 126: en-US

```suggestion
Common use cases include converting a JavaScript `Date` object to a date string or a `Number` into a formatted number (e.g. "1,000.50").
```

try 

```js
console.log(new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' }).format(1000.5));
```

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -171,13 +171,15 @@ export function useGridEditRows(apiRef: GridApiRef) {
 
   const setCellValue = React.useCallback(
```
oliviertassinari, line 172: Shouldn't `setEditCellProps` be handled too. For instance, https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#controlled-editing

---

`Age` is supposed to be a number but is stored as a string. Shouldn't we update all the default columns to leverage this new `valueParser` property?

Capture d’écran 2021-05-30 à 21 41 13

m4theushw: > Shouldn't setEditCellProps be handled too. For instance, https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#controlled-editing

I think we need to discuss this. From what I understood, `setEditCellProps` should store in the `editRows` state the new value for the edit cell, which feeds the `value` prop. The value parser would only be applied when updating the row with the value stored in the state (a.k.a commiting). By the way, we use different approaches to create edit components:
- `GridEditInputCell` and `GridEditBooleanCell` publish the `GRID_CELL_EDIT_PROPS_CHANGE` event
- The `ValidateServerNameGrid` [demo](https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation) calls `setEditCellProps`
- The [demo](https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#custom-edit-component) with the rating component calls `commitCellChange`.

> Age is supposed to be a number but is stored as a string. Shouldn't we update all the default columns to leverage this new valueParser property?

We could, but what about update `GridEditInputCell` to parse the value and convert it to a number? We already do something similar with dates.

docs/src/pages/components/data-grid/columns/ValueParserGrid.js
```diff
@@ -0,0 +1,37 @@
...
+      <DataGrid
+        rows={rows}
+        columns={[
+          {
+            type: 'number',
+            field: 'date',
+            headerName: 'Year',
+            width: 150,
+            valueFormatter: (params) => params.value.getFullYear(),
+            valueParser: (value) => new Date(Number(value), 0, 1),
```
dtassone, line 31: Should we make it editable, so we can also test the editability with this demo?

m4theushw: If I make it editable, I'll have to use a custom component for the edit cell because the default one is converting the value to a date string (when it's a `Date`) and that doesn't work with `input[type=number]`. I would prefer to leave as it is to keep the code below the demo's toolbar simple.

dtassone: it feels like the valueFormatter should be a valueGetter as you are changing the type of value.

dtassone (approved):

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
oliviertassinari: @m4theushw I don't share the same perspective. Take this example:

https://github.com/mui-org/material-ui-x/blob/a24eba2a67657afe2a608bd29adc52fd629a7ca2/packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx#L41

We are doing the formatting manually inside the field for each value change.
For the filtering, we are also applying the formatted manually: https://github.com/mui-org/material-ui-x/blob/a24eba2a67657afe2a608bd29adc52fd629a7ca2/packages/grid/_modules_/grid/models/colDef/gridDateOperators.ts#L23

IMHO, we could abstract both with the new `valueParser` feature (we currently have two different implementations of `parseDate` and `parseDateTime`).

> We might ended up parsing an already parsed value if we choose to call valueParser in all places.

Then this sounds like we would uncover root issues with the data flow, that makes wrong assumptions.

Issue #1204, opened 2021-03-11
[XGrid] Multi-sort not working


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When the <kbd>Shift</kbd> or <kbd>Ctrl</kbd> key is pressed without DOM focus in the data grid, the feature is no working.

## Expected Behavior 🤔

When the <kbd>Shift</kbd> or <kbd>Ctrl</kbd> key is pressed without DOM focus in the data grid, the feature is still working.

## Steps to Reproduce 🕹

Steps:

1. Open https://material-ui.com/components/data-grid/sorting/#multi-column-sorting
2. Hold <kbd>Ctrl</kbd>
3. Click on two columns
4. 💥

![multi-sort](https://user-images.githubusercontent.com/3165635/110801298-601d7c00-827d-11eb-8c53-baf8fa394892.gif)

## Context 🔦

We can remove code by fixing this issue. The current logic listens to the *keydown* event on the data grid's root, which is not triggered when the <kbd>Ctrl</kbd> key is held with the DOM focus somewhere else. 
Instead, we could ask the event if other keys are pressed. It's, for instance, what we do here:

https://github.com/mui-org/material-ui/blob/ebeb1492b5e645eccdea920fbbf561fe2a0c3031/packages/material-ui-lab/src/TreeView/TreeView.js#L666

The other alternative is to listen on the document owning the data grid's root (but it seems more complexity than needed).

## Your Environment 🌎

v4.0.0-alpha.21

Issue #208, opened 2020-08-22
[data grid] Implement cell selection with range


oliviertassinari:
Turn the 1D selection (rows) into 2D (cells).

<img width="318" alt="Capture d’écran 2021-06-07 à 22 48 25" src="https://user-images.githubusercontent.com/3165635/121085721-843c1b00-c7e2-11eb-9b1f-6f2969b5b7fe.png">

### Benchmark

- https://glideapps.github.io/glide-data-grid/?path=/story/glide-data-grid-docs--search
- https://ag-grid.com/javascript-grid-range-selection/
- https://handsontable.com/docs/8.0.0/demo-selecting-ranges.html
- https://adazzle.github.io/react-data-grid/docs/examples/cell-range-selection
- https://www.telerik.com/kendo-react-ui/components/grid/selection/

### 🔥 Release (Edit: Sep142022)

We expect to release it in Q4 2022.
This will be a v6 feature, so if you're okay using alpha and beta packages, you'll probably be able to use it at least a month sooner. 

If anyone is willing to help steer the direction of the components, please consider giving us a [user interview](https://forms.gle/vsBv6CLPz9h57xg8A). We'd love to know more about your use cases, pain points, and expectations for this and other features.

m4theushw:
A couple of thoughts about the implementation of this feature:

- Rename `useGridSelection` -> `useGridRowSelection`
- Keep `selectionModel` and `onSelectionModelChange` only for row selection. We could rename them as well
- Create the `useGridRangeSelection` hook with its own state
- Add a `rangeSelectionModel` prop to control the range selection
- Clicking a cell firstly focuses it
- Clicking and moving the mouse selects other cells (focus remains in the first clicked)

**Note**: When I say "focus" I refer to update the `focus` state.

![image](https://user-images.githubusercontent.com/42154031/124192979-80777d80-da9c-11eb-81a6-19baf1e75441.png)

sergzz12:
Hello. Any idea on the timeline of the release of this feature?

Sajjad1432:
I wanna implement a Date Range picker in MUI DataGridPro is there any solution?

fabioel:
Hello. Would really love to see this feature implemented.

DmitriXBS:
## Some clarifications are needed about this feature request

### How are multiple selections made? (any or all of these)
1. click and drag to select a "block"
2. ctrl/cmd-click to select individual cells that may not be near each other
3. click to select, then shift-click to select a block of cells 
[...]

PR #1813, opened 2021-06-01
[DataGrid] Deselect row in non checkboxSelection grid

### LINKED ISSUE #1437: [DataGrid] Remove gridKeyboardMultipleKeySelector in useGridSelection

## Summary 💡

`useGridSelection` relays in the state to detect if a multiple key is pressed (Shift or CTRL). This could be read from the event. There's an opportunity for an enhancement here that was raised in https://github.com/mui-org/material-ui-x/pull/1422#discussion_r614878834. 

## Motivation 🔦

Use the event as single source of truth.

### LINKED ISSUE #1770: [DataGrid] Deselect row in non checkboxSelection grid

## Summary 💡

DataGrid with `checkboxSelection={false}` by default allows you to select one row by clicking on it. Once you selected a row, clicking on it again does nothing except for calling `onRowSelected` if provided (with same arguments as before).

I would expect that if I click on selected row it will be deselected (just like it behaves with checkboxSelection). With multiple selection of Xgrid you can deselect single row by holding `Ctrl` which does not work in DataGrid. Now after you select a row there is no way of going back to having no rows selected.

## Examples 🌈

Current behavior: https://codesandbox.io/s/gifted-wind-xg84s?file=/src/Demo.tsx
Deselecting with XGrid: https://material-ui.com/components/data-grid/selection/#multiple-row-selection

## Motivation 🔦

Without providing my own controlled `selectionModel` I see no way of implementing this really basic feature. It would be nice to provide a flag that allows deselection when clicking selected row that would trigger a `onRowSelected` with params with `isSelected: false`.

ZeeshanTamboli:
Fixes #1770 
Fixes #1437 

End-users can now use the <kbd>CTRL</kbd> key to unselect a row, even if multi selection is disabled. 

Preview: https://deploy-preview-1813--material-ui-x.netlify.app/components/data-grid/selection/


docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -27,14 +27,13 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
 
 ### Single row selection
 
-Single row selection is enable by default with the `DataGrid` component.
-For the `XGrid`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
+Single row selection is enable by default with the `DataGrid` component. While deselecting a row, hold the <kbd class="key">CTRL</kbd> key. For the `XGrid`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
```
oliviertassinari, line 30: One line per sentence is easier to review, we aim for writing docs like this.

```suggestion
Single row selection is enable by default with the `DataGrid` component.
To unselect a row, hold the <kbd class="key">CTRL</kbd> key and click on it.
For the `XGrid`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -30,6 +30,7 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
   const rowsLookup = useGridSelector(apiRef, gridRowsLookupSelector);
   const isMultipleKeyPressed = useGridSelector(apiRef, gridKeyboardMultipleKeySelector);
   const allowMultipleSelectionKeyPressed = React.useRef<boolean>(false);
+  const isMultipleKeyPressedRef = React.useRef<boolean>(false);
```
oliviertassinari, line 33: We shouldn't need for this. Isn't it `selectRowModel` that is called wight the wrong argument? 

```suggestion
```

oliviertassinari (dismissed):
We miss a description of the unselect behavior: https://material-ui.com/components/data-grid/accessibility/#selection

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
ZeeshanTamboli: I think it is called with the right argument. 
https://github.com/mui-org/material-ui-x/blob/c992eb3de48ed3f465cb8d3daacd45f13758224a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L166-L173

For deselection, key press should be detected. Just clicking on a selected row should not be deselected, but with a combination of supported keypress + click.
Not sure about the consideration of ref though. I considered it so that the key press boolean value can be persisted to check its value below.

ZeeshanTamboli:
> We miss a description of the unselect behavior: https://material-ui.com/components/data-grid/accessibility/#selection

Nice that you noted. I added it now.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
DanailH: Can't we just use `isMultipleKeyPressed` directly?

ZeeshanTamboli: > Can't we just use `isMultipleKeyPressed` directly?

No. I think when we click on the row the DataGrid re-renders and `isMultipleKeyPressed ` is initialised back to false. Storing it in a ref means the value is persisted by React across re-renders. Also, it's value change to `false` (if we leave the pressing of key) will not cause a re-render.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -70,27 +72,24 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
       const allowMultiSelect =
         allowMultipleOverride || allowMultipleSelectionKeyPressed.current || checkboxSelection;
 
-      if (allowMultiSelect) {
-        setGridState((state) => {
-          // eslint-disable-next-line prefer-object-spread
-          const selectionState: GridSelectionState = Object.assign({}, state.selection);
+      setGridState((state) => {
+        let selection: GridSelectionState = { ...state.selection };
```
dtassone, line 76: can we rename that to `selectionState`

ZeeshanTamboli: Renamed

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
oliviertassinari: This sounds like a great opportunity to refactor the logic. I would propose we remove `isMultipleKeyPressed` altogether. It's the last location we haven't yet refactored to remove the concept. multi-selection has the same bugs we had with multi-sorting. e.g. if you press <kbd>CTRL</kbd> outside of the data grid, it fails: #1204

docs/src/pages/components/data-grid/accessibility/accessibility.md
```diff
@@ -71,6 +71,7 @@ Use the arrow keys to move the focus.
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">A</kbd> | Select all rows                                   |
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">C</kbd> | Copy the currently selected row                   |
 |                                                        <kbd class="key">CTRL</kbd> + Click on cell | Enable multi-selection                            |
+|                                              <kbd class="key">CTRL</kbd> + Click on a selected row | Deselects a row                                   |
```
oliviertassinari, line 74:
```suggestion
|                                              <kbd class="key">CTRL</kbd> + Click on a selected row | Deselects the row                                   |
```

docs/src/pages/components/data-grid/accessibility/accessibility.md
```diff
@@ -71,6 +71,7 @@ Use the arrow keys to move the focus.
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">A</kbd> | Select all rows                                   |
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">C</kbd> | Copy the currently selected row                   |
 |                                                        <kbd class="key">CTRL</kbd> + Click on cell | Enable multi-selection                            |
```
oliviertassinari, line 73: **Off-topic**

It looks like we can do a follow-up to fix the grammar.

docs/src/pages/components/data-grid/accessibility/accessibility.md
```diff
@@ -71,6 +71,7 @@ Use the arrow keys to move the focus.
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">A</kbd> | Select all rows                                   |
 |                                             <kbd class="key">CTRL</kbd> + <kbd class="key">C</kbd> | Copy the currently selected row                   |
```
mbrookes, line 72:
```suggestion
|                                             <kbd class="key">CTRL</kbd> + <kbd class="key">C</kbd> | Copy the currently selected row(s)                |
```

docs/src/pages/components/data-grid/accessibility/accessibility.md
mbrookes: I don't see an issue with the grammar, but is the description correct?

docs/src/pages/components/data-grid/accessibility/accessibility.md
mbrookes: This seems to be describing Windows modifier keys.

docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -28,13 +28,14 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
...
 Single row selection is enable by default with the `DataGrid` component.
+To unselect a row, hold the <kbd class="key">CTRL</kbd> key and click on it.
 For the `XGrid`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
 
 {{"demo": "pages/components/data-grid/selection/SingleRowSelectionGrid.js", "bg": "inline"}}
 
 ### Multiple row selection [<span class="pro"></span>](https://material-ui.com/store/items/material-ui-pro/)
 
-To activate multiple selection, put focus the `XGrid` component and hold the <kbd class="key">CTRL</kbd> key while selecting rows.
+To activate multiple selection, put focus on the `XGrid` component and hold the <kbd class="key">CTRL</kbd> key while selecting rows.
```
mbrookes, line 38:
```suggestion
To select multiple rows, focus the `XGrid` component and hold the <kbd class="key">CTRL</kbd> key while selecting rows.
```
(Cmd on Mac)

mbrookes: The shift key also works for selecting multiple rows one at a time, but should instead allow range selection, for example selecting row 2, holding shift, and selecting row 5 should also select rows 2, 3 & 4.

docs/src/pages/components/data-grid/accessibility/accessibility.md
```diff
@@ -71,6 +71,7 @@ Use the arrow keys to move the focus.
...
 |                                                        <kbd class="key">CTRL</kbd> + Click on cell | Enable multi-selection                            |
+|                                              <kbd class="key">CTRL</kbd> + Click on a selected row | Deselects the row                                 |
```
oliviertassinari, line 74:
```suggestion
|                                              <kbd class="key">CTRL</kbd> + Click on a selected row | Deselect the row                                 |
```

docs/src/pages/components/data-grid/accessibility/accessibility.md
oliviertassinari: Ok, I was wondering about the plural, then: https://github.com/mui-org/material-ui-x/pull/1813/files#r644343155.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
m4theushw: To drop `isMultipleKeyPressed `, you just have to check if a multiple key is pressed in `handleRowClick` and set the third argument of `selectRowModel` to true. Don't forget to also check if the `disableMultipleSelection` prop is not true. I did something similar to fix the multi-sorting:

https://github.com/mui-org/material-ui-x/blob/c992eb3de48ed3f465cb8d3daacd45f13758224a/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts#L241

m4theushw (commented):
If you could add a test case would be awesome.

docs/src/pages/components/data-grid/accessibility/accessibility.md
ZeeshanTamboli: I can log a separate issue and possibly a PR for docs later to consider Mac modifier keys.

docs/src/pages/components/data-grid/selection/selection.md
ZeeshanTamboli: > The shift key also works for selecting multiple rows one at a time, but should instead allow range selection, for example selecting row 2, holding shift, and selecting row 5 should also select rows 2, 3 & 4.

Yup, we can consider this too in a separate PR.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
ZeeshanTamboli: Thanks, I'll refactor to drop `isMultipleKeyPressed`. 

> Don't forget to also check if the disableMultipleSelection prop is not true

I think you mean check for `disableMultipleSelection` to be true for DataGrid's single selection?

ZeeshanTamboli:
> If you could add a test case would be awesome.

Yes. Test case would really help. I'll add one.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
ZeeshanTamboli: I've refactored in https://github.com/mui-org/material-ui-x/pull/1813/commits/4021a0f2aeb8d8246cb633a87f44b2b668409c16

ZeeshanTamboli:
> > If you could add a test case would be awesome.
> 
> Yes. A test case would really help. I'll add one.

@m4theushw I added a test case.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -67,30 +71,29 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
+          (!disableMultipleSelection && isMultipleKey) ||
+          checkboxSelection;
 
-      if (allowMultiSelect) {
-        setGridState((state) => {
-          // eslint-disable-next-line prefer-object-spread
-          const selectionState: GridSelectionState = Object.assign({}, state.selection);
+        if (allowMultiSelect) {
           const isRowSelected =
             !allowMultiSelect || isSelected == null ? selectionState[id] === undefined : isSelected;
```
m4theushw, line 83: `allowMultiSelect` is always true in this branch. 😛

```suggestion
            isSelected == null ? selectionState[id] === undefined : isSelected;
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -67,30 +71,29 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
-          return { ...state, selection: selectionState };
-        });
-      } else {
-        setGridState((state) => {
-          const selectionState: GridSelectionState = {};
+        } else if (disableMultipleSelection && isMultipleKey && selectionState[id] !== undefined) {
+          selectionState = {};
+        } else {
+          selectionState = {};
           selectionState[id] = id;
```
m4theushw, line 93: You were not honoring `isSelected`. I think we can simplify this block with:

```suggestion
        } else {
          const isRowSelected =
            isSelected == null ? !isMultipleKey || selectionState[id] === undefined : isSelected;
          selectionState = {};
          if (isRowSelected) {
            selectionState[id] = id;
          }
```

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -12,6 +12,49 @@ describe('<DataGrid /> - Selection', () => {
...
+    fireEvent.click(screen.getByRole('cell', { name: 'Adidas' }));
+    expect(firstRow).not.to.have.class('Mui-selected');
+    expect(secondRow).to.have.class('Mui-selected');
+
+    // selecting another row with key press
+    fireEvent.click(screen.getByRole('cell', { name: 'Nike' }), { metaKey: true });
+    expect(secondRow).not.to.have.class('Mui-selected');
+    expect(firstRow).to.have.class('Mui-selected');
+
+    // deselecting a selected row without key press
```
m4theushw, line 48: You're not deselecting here. The first row is already selected.

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -12,6 +12,49 @@ describe('<DataGrid /> - Selection', () => {
...
+          ]}
+          columns={[{ field: 'brand', width: 100 }]}
+        />
+      </div>,
+    );
+    const firstRow = getRow(0);
+    const secondRow = getRow(1);
+
+    // selecting row on click without key press
+    fireEvent.click(screen.getByRole('cell', { name: 'Nike' }));
```
m4theushw, line 37: We have a helper to get the cell at X row and Y column.

```suggestion
    fireEvent.click(getCell(0, 0));
```

docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -28,13 +28,14 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
...
 Single row selection is enable by default with the `DataGrid` component.
+To unselect a row, hold the <kbd class="key">CTRL</kbd> key and click on it.
 For the `XGrid`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
 
 {{"demo": "pages/components/data-grid/selection/SingleRowSelectionGrid.js", "bg": "inline"}}
 
 ### Multiple row selection [<span class="pro"></span>](https://material-ui.com/store/items/material-ui-pro/)
 
-To activate multiple selection, put focus the `XGrid` component and hold the <kbd class="key">CTRL</kbd> key while selecting rows.
+To select multiple rows, focus on the `XGrid` component and hold the <kbd class="key">CTRL</kbd> key while selecting rows.
```
dtassone, line 38: Do we need the focus on the grid? I thought we fixed that 🤔

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
ZeeshanTamboli: I improved the comment. I wanted to test that `clicking on a selected row without key press should not deselect`.

docs/src/pages/components/data-grid/selection/selection.md
ZeeshanTamboli: Yes we fixed that 👍 . I rephrased it.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
ZeeshanTamboli: Updated

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -12,6 +12,49 @@ describe('<DataGrid /> - Selection', () => {
...
+            },
+          ]}
+          columns={[{ field: 'brand', width: 100 }]}
+        />
+      </div>,
+    );
+    const firstRow = getRow(0);
+    const secondRow = getRow(1);
+
+    // selecting row on click without key press
```
dtassone, line 36: it seems that each of these comments can be a seperate small test

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -12,6 +12,49 @@ describe('<DataGrid /> - Selection', () => {
   // TODO v5: replace with createClientRender
   const render = createClientRenderStrictMode();
 
+  it('should be selected and deselected when clicking on a row', () => {
```
dtassone, line 15: If we split, it will be easier to understand what should happen in this test and in the selection behavior, like holding CTRL to unselect, or selecting multiple rows...

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -12,6 +12,66 @@ describe('<DataGrid /> - Selection', () => {
   // TODO v5: replace with createClientRender
   const render = createClientRenderStrictMode();
 
+  describe('no checkboxSelection prop - selection/deselection', () => {
```
dtassone, line 15: that looks much better 👍

dtassone (approved):

m4theushw (commented):
I have pushed two commits to include <kbd>Shift</kbd> and <kbd>⌘ Command</kbd> as multiple keys both in the docs as in the tests too. In the multi-sorting we also support them: https://material-ui.com/components/data-grid/sorting/#multi-column-sorting

m4theushw (approved):

docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -28,13 +28,14 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
 ### Single row selection
 
 Single row selection is enable by default with the `DataGrid` component.
+To unselect a row, hold the <kbd class="key">CTRL</kbd> or <kbd class="key">Shift</kbd> (use <kbd class="key">⌘ Command</kbd> on macOS) key and click on it.
```
oliviertassinari, line 31: I would propose we only mention CTRL per https://material-ui.com/components/data-grid/accessibility/#key-assignment-conventions
```suggestion
To unselect a row, hold the <kbd class="key">CTRL</kbd> or <kbd class="key">Shift</kbd> key and click on it.
```

It's more or less based on https://www.w3.org/TR/wai-aria-practices/#kbd_common_conventions.

oliviertassinari: I would vote against using <kbd>Shift</kbd>. You would end up unselecting the current row if you are in multi-select mode and slightly miss your click target. At least, macOS doesn't support it not 

```suggestion
To unselect a row, hold the <kbd class="key">CTRL</kbd> (use <kbd class="key">⌘ Command</kbd> on macOS) key and click on it.
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -106,12 +110,25 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
+      checkboxSelection,
+      forceUpdate,
+      setGridState,
+    ],
   );
 
   const selectRow = React.useCallback(
     (id: GridRowId, isSelected = true, allowMultiple = false) => {
-      selectRowModel(id, apiRef.current.getRow(id), allowMultiple, isSelected);
+      selectRowModel({
```
oliviertassinari, line 126: Yes, thanks for this. 4 arguments start to be too hard to reason with.

docs/src/pages/components/data-grid/selection/selection.md
m4theushw: Yes, <kbd>Shift</kbd> is generally used for extending the selection. It doesn't makes sense but `isMultipleKeyPressed` supports it. I'll open a follow-up PR to fix the helper.

oliviertassinari: `isMultipleKeyPressed` is only used for the multi-header sorting. Which is an interesting problem. on AG Grid only <kbd>Shift</kbd> is handled. But on handsontable, a click on the header selects the whole column, so they only implement the multi-sorting wth <kbd>Ctrl</kbd>. IHMO the isMultipleKeyPressed abstraction is wrong, it should be handled on case by case.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -186,7 +185,7 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
         selectRowModel({
           id: params.id,
           row: params.row,
-          isMultipleKey: isMultipleKeyPressed(event),
+          isMultipleKey: event.metaKey || event.ctrlKey,
```
dtassone, line 188: not shift?

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
```diff
@@ -238,7 +238,8 @@ export const useGridSorting = (apiRef: GridApiRef, rowsProp: GridRowsProp) => {
 
   const handleColumnHeaderClick = React.useCallback(
     ({ colDef }: GridColumnHeaderParams, event: React.MouseEvent) => {
-      sortColumn(colDef, undefined, isMultipleKeyPressed(event));
+      const allowMultipleSorting = event.shiftKey || event.metaKey || event.ctrlKey;
```
dtassone, line 241: we have shift here

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -42,7 +42,7 @@ describe('<DataGrid /> - Selection', () => {
       expect(secondRow).to.have.class('Mui-selected');
     });
 
-    ['metaKey', 'ctrlKey', 'shiftKey'].forEach((key) => {
+    ['metaKey', 'ctrlKey'].forEach((key) => {
```
dtassone, line 45: and we don't have shift here.

docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -27,7 +27,7 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
 ### Single row selection
 
 Single row selection is enable by default with the `DataGrid` component.
-To unselect a row, hold the <kbd class="key">CTRL</kbd> or <kbd class="key">Shift</kbd> (use <kbd class="key">⌘ Command</kbd> on macOS) key and click on it.
+To unselect a row, hold the <kbd class="key">CTRL</kbd> key and click on it.
```
dtassone, line 30: We just have CTRL here. It's a bit confusing 😆

oliviertassinari: In https://github.com/mui-org/material-ui-x/pull/1813#discussion_r645833808, I have tried to argue why Shift is harmful to this feature.

m4theushw: Pressing <kbd>Shift</kbd> while clicking the rows selects the text too. It's better to not support this key.

hNBz68IIRK

oliviertassinari: Actually, we could open a new GitHub issue for https://material-ui.com/components/data-grid/selection/#multiple-row-selection. Shouldn't <kbd>Shift</kbd> should allow range selection, as it does on Excel or on a Tree View?

m4theushw: There's https://github.com/mui-org/material-ui-x/issues/208 but given the benchmark examples it seems more of a cell range selection. But I totally agree to allow the selection of a range of rows with <kbd>Shift</kbd>.

oliviertassinari: Ok, I'm adding a screenshot to #208 to avoid confusion :). It's different

PR #1816, opened 2021-06-02
[DataGrid] Fix fluid columns width when available `viewportWidth` < 0

### LINKED ISSUE #1582: [DataGrid] Fix width calc for fluid column on resize
Just need to hook to resize event to trigger a new calculation of column width when the grid resize

Discussed in #1574 

issue explained in #1554

DanailH:
Fixes #1582 

Reproduction: https://codesandbox.io/s/material-demo-forked-y2xj0 (when the grid loads start resizing the window to make it small to a point where the last column dissapears)

The problem was that when resizing and space was not enough the fluid columns did not default to the base width of `100px`.

https://user-images.githubusercontent.com/5858539/120652536-87ed3c00-c488-11eb-810b-0de3d1a07101.mov

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -32,6 +32,7 @@ import {
   visibleGridColumnsSelector,
 } from './gridColumnsSelector';
 import { useGridApiOptionHandler } from '../../root/useGridApiEventHandler';
+import { GRID_STRING_COL_DEF } from '../../../models/colDef/gridStringColDef';
 
 function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
   const numberOfFluidColumns = columns.filter((column) => !!column.flex && !column.hide).length;
@@ -50,14 +51,19 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
   }
 
   let newColumns = columns;
-  if (viewportWidth > 0 && numberOfFluidColumns) {
+  if (numberOfFluidColumns) {
     const flexMultiplier = viewportWidth / flexDivider;
-    newColumns = columns.map((column) => {
-      return {
-        ...column,
-        width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
-      };
-    });
+    newColumns = columns.map((column) =>
+      viewportWidth > 0
+        ? {
+            ...column,
+            width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
+          }
+        : {
+            ...column,
+            width: column.flex! ? GRID_STRING_COL_DEF.width : column.width,
+          },
+    );
   }
   return newColumns;
 }
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -53,17 +53,16 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
   let newColumns = columns;
   if (numberOfFluidColumns) {
     const flexMultiplier = viewportWidth / flexDivider;
-    newColumns = columns.map((column) =>
-      viewportWidth > 0
-        ? {
-            ...column,
-            width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
-          }
-        : {
-            ...column,
-            width: column.flex! ? GRID_STRING_COL_DEF.width : column.width,
-          },
-    );
+    newColumns = columns.map((column) => {
+      if (!column.flex) {
+        return column;
+      }
+      return {
+        ...column,
+        width:
+          viewportWidth > 0 ? Math.floor(flexMultiplier * column.flex!) : GRID_STRING_COL_DEF.width,
+      };
+    });
   }
   return newColumns;
 }


packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -50,14 +51,19 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
...
-      return {
-        ...column,
-        width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
-      };
-    });
+    newColumns = columns.map((column) =>
+      viewportWidth > 0
+        ? {
+            ...column,
+            width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
```
dtassone, line 60: I think you can do it in 1 line as flexMultiplier would be 0 if viewport is 0 and we merge col with the default type and string is default.
`width: column.flex! && flexMultiplier ? Math.floor(flexMultiplier * column.flex!) : column.width,`

m4theushw: > I think you can do it in 1 line as flexMultiplier would be 0 if viewport is 0 and we merge col with the default type and string is default.
> `width: column.flex! && flexMultiplier ? Math.floor(flexMultiplier * column.flex!) : column.width,`

The `column.width` might not be the default one. The updated widths are saved to the state and passed as argument.

https://github.com/mui-org/material-ui-x/blob/c992eb3de48ed3f465cb8d3daacd45f13758224a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L280

---

What about if we do the cloning only when the column is fluid?

```diff
diff --git a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
index 82559053..5525a787 100644
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -53,17 +53,18 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
   let newColumns = columns;
   if (numberOfFluidColumns) {
     const flexMultiplier = viewportWidth / flexDivider;
-    newColumns = columns.map((column) =>
-      viewportWidth > 0
-        ? {
-            ...column,
-            width: column.flex! ? Math.floor(flexMultiplier * column.flex!) : column.width,
-          }
-        : {
-            ...column,
-            width: column.flex! ? GRID_STRING_COL_DEF.width : column.width,
-          },
-    );
+    newColumns = columns.map((column) => {
+      if (!column.flex) {
+        return column;
+      }
+      return {
+        ...column,
+        width:
+          viewportWidth > 0
+            ? Math.floor(flexMultiplier * column.flex!)
+            : GRID_STRING_COL_DEF.width,
+      };
+    });
   }
   return newColumns;
 }
```

oliviertassinari (commented):
Do we have a reproduction, e.g. there is almost no information on the linked issue. What's the issue?

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
DanailH: > What about if we do the cloning only when the column is fluid?

That will work, I'll update it.

DanailH:
> Do we have a reproduction, e.g. there is almost no information on the linked issue. What's the issue?

I'll add a reproduction but the issue is that if you have only one fluid column (as a base case) and you start resing the grid, there is a point where there is no space for the fluid column to be displayed and the grid doesn't allow scrolling so you cant see the last X columns.

DanailH:
> > Do we have a reproduction, e.g. there is almost no information on the linked issue. What's the issue?
> 
> I'll add a reproduction but the issue is that if you have only one fluid column (as a base case) and you start resing the grid, there is a point where there is no space for the fluid column to be displayed and the grid doesn't allow scrolling so you cant see the last X columns.

PS: @oliviertassinari I added reproduction

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
DanailH: @m4theushw done

oliviertassinari:
@DanailH Great thanks for the reproduction! It seems to be exactly this https://github.com/mui-org/material-ui-x/issues/1561#issuecomment-832753596. In the realm of AG Grid, this would be expected behavior, and they would tell us: add [a `minWidth` property](https://ag-grid.com/react-grid/column-sizing/#column-flex). I personally feel that if we reproduce the flexbox model of CSS, then a very small column width is normal.

m4theushw (approved):

DanailH:
@oliviertassinari Visual test can be tricky in this situation as this only happens when you load the grid and then resize the window so that there isn't enough space for all the columns that already have fixed width. If you load the grid in a small space/window initially the regression doesn't happen.

dtassone (approved):

oliviertassinari:
> Visual test can be tricky in this situation 

@DanailH I'm completely lost. 

HEAD: I load the page like this, the last column is not visible: ✅

Capture d’écran 2021-06-04 à 13 40 11

https://codesandbox.io/s/material-demo-forked-y2xj0?file=/demo.js:0-587

PR: Same behavior

Capture d’écran 2021-06-04 à 13 41 10

https://codesandbox.io/s/material-demo-forked-y8m1q?file=/package.json

oliviertassinari:
I mean: flex: 1, means:

Capture d’écran 2021-06-04 à 13 46 39

https://codepen.io/oliviertassinari/pen/XWMExoz

Capture d’écran 2021-06-04 à 13 48 21

Expecting width: 100px here seems wrong.

m4theushw:
@oliviertassinari This PR is fixing the negative widths when there's no available space. A visual test case would not check that but an unit test can.

> I think that it only makes sense if we add minWidth: 100px.

So default it to zero?

oliviertassinari:
@m4theushw I think that it should default to the default `minWidth` value.

DanailH:
> @m4theushw I think that it should use the default `minWidth` value when there is not enough space, not the `width` value because `flex` overrides `width`.

@oliviertassinari the thing is that there is no `minWidth` on the `ColDef`, I can add one but that is a different PR. Regarding the `flex` it's just `sugar` on the `ColDef` that gets converted to pixels.

oliviertassinari:
@DanailH Ok, so we can close this PR, close #1582, and open a new standalone issue for https://github.com/mui-org/material-ui-x/issues/1561#issuecomment-832753596? 

The reasoning being:

1. the current logic is correct, it's how CSS work, what we aim to reproduce to avoid developers to learn yet another layout behavior
2. the changes here make the behavior inconsistent:
not working

DanailH:
@oliviertassinari Done, here is the ticket https://github.com/mui-org/material-ui-x/issues/1850. We can discuss it today during the sprint planning and bring it in.

oliviertassinari:
By close, I meant close, not merge, the behavior is now inconsistent. But it's fine, we can revert in #1850. Thanks for the new issue.

DanailH:
> By close, I meant close, not merge, the behavior is now inconsistent. But it's fine, we can revert in #1850. Thanks for the new issue.

Sorry I misunderstood. I'll change that once the other ticket is picked up.

Issue #1618, opened 2021-05-10
[DataGrid] Clicking on the filter icon in header title is not opening the filter panel


dtassone:
- open https://material-ui.com/components/data-grid/#mit-version
- add a fitler on a column 
- click on the filter icon of the filtered column
- The filter panel should open

oliviertassinari:
@dtassone If you could add the DataGrid component label, it would be great, it helps differentiate between MIT and commercial impacts, and once we work on more than a DataGrid, to differentiate the efforts.

m4theushw:
That's a regression from #1421. Because of the added `React.memo`, the `GridColumnHeaderItem` is only rendering when there's a prop changes. Since a click in the filter icon doesn't cause a prop change but a change in the state, it doesn't know when to show the preferences panel.  #1415 is also broke after pulling changes from master. 

Can we remove `React.memo`? It was added in https://github.com/mui-org/material-ui-x/pull/1421/commits/09c285bd3e2f0cd2b10f17c319aa9653174f9d5f. Does it fix the tab roving bug or is it for performance reasons?

https://github.com/mui-org/material-ui-x/blob/af5c05f572f787ed01c32241da78b805d94badfd/packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx#L45-L48

m4theushw:
Calling `getState` only gets the updated state in the handlers. All `useGridSelector` (like below) will use the old value. I think that passing state values as props is a solution. We'll have to move a bunch of them to `GridColumnHeadersItemCollection`.

https://github.com/mui-org/material-ui-x/blob/af5c05f572f787ed01c32241da78b805d94badfd/packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx#L58-L60

oliviertassinari:
As a general note, we have faced the same problem in the past: https://github.com/mui-org/material-ui-x/pull/666.

dtassone:
> > when the component is memoized which is normal I believe
> 
> @dtassone Could you expand on what you mean by normal? I can interpret it in two different ways:
> 
> 1. it's how React and the context works
> 2. it's an implementation tradeoff
> 
> Did you mean one of these two?

If the component is memoized then, React will do a shallow comparison of the props to know if it will update.
So if the props don't change then it won't rerun the component function.
This is the reason why, I was memoizing only the last component of the tree, ie the cells, but now we improved that a bit.

Check this example
[...]

PR #1819, opened 2021-06-03
[DataGrid] Fix cell height after changing grid's density

### LINKED ISSUE #1776: [DataGrid] Density selector not working

## Current Behavior 😯

The density selector is not rendering correctly.

## Expected Behavior 🤔

The density selector is rendering correctly.

## Steps to Reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/9fzgz
2. 

https://user-images.githubusercontent.com/3165635/119833881-c5d5e780-beff-11eb-9772-905f71346aee.mp4

## Context 🔦

Noticed on https://material-ui.com/components/data-grid/demo/

## Your Environment 🌎

v4.0.0-alpha.29

DanailH:
Fixes #1776 

The problem was related to grabbing the height from the state inside `GridRowCells` component which is memoized. Passing it as a prop solves the issue. I extended the existing density tests to cover that case.

 Preview: https://deploy-preview-1819--material-ui-x.netlify.app/components/data-grid/accessibility/#density-selector

### DIFF
--- a/packages/grid/data-grid/src/tests/toolbar.DataGrid.test.tsx
+++ b/packages/grid/data-grid/src/tests/toolbar.DataGrid.test.tsx
@@ -68,6 +68,11 @@ describe('<DataGrid /> - Toolbar', () => {
       expect(screen.getAllByRole('row')[1]).toHaveInlineStyle({
         maxHeight: `${Math.floor(rowHeight * COMPACT_DENSITY_FACTOR)}px`,
       });
+
+      // @ts-expect-error need to migrate helpers to TypeScript
+      expect(screen.getAllByRole('cell')[1]).toHaveInlineStyle({
+        maxHeight: `${Math.floor(rowHeight * COMPACT_DENSITY_FACTOR)}px`,
+      });
     });
 
     it('should decrease grid density when selecting comfortable density', () => {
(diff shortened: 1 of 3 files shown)


packages/grid/data-grid/src/tests/toolbar.DataGrid.test.tsx
```diff
@@ -105,6 +115,11 @@ describe('<DataGrid /> - Toolbar', () => {
       expect(screen.getAllByRole('row')[1]).toHaveInlineStyle({
         maxHeight: `${Math.floor(rowHeight * COMPACT_DENSITY_FACTOR)}px`,
       });
+
+      // @ts-expect-error need to migrate helpers to TypeScript
+      expect(screen.getAllByRole('cell')[1]).toHaveInlineStyle({
```
dtassone, line 120: we could just load the maxHeight and do an equal assert to avoid the ts-error

DanailH: Yes but that tests the internal, not the visual part. For me this way the test is more secure. If we change the way the height of the cell is set then if we check the `maxHeight` the test will stop working, no?

oliviertassinari (commented):
It looks like the bug resonates with https://github.com/mui-org/material-ui-x/issues/1618#issuecomment-838711349. 

Regarding the solution, did we consider to 1. only have a height property on the row, not on the cells? 2. No min-height/max-height, only height?

DanailH:
> It looks like the bug resonates with [#1618 (comment)](https://github.com/mui-org/material-ui-x/issues/1618#issuecomment-838711349).
> 
> Regarding the solution, did we consider to 1. only have a height property on the row, not on the cells? 2. No `min-height`/`max-height`, only `height`?

That was the easiest way to solve the issue. Regarding 2. it will still be the same problem I think as even if we only have height it still needs to come from the state and be based on the row height. Option 1. is more favorable. I will see if the effort of depending only on the row height is too big.

dtassone (approved):

oliviertassinari:
I have updated https://github.com/mui-org/material-ui-x/issues/1618#issuecomment-838711349 to keep track of the bugs created by the `useGridSelector` that doesn't work cross memo functions.

Issue #1804, opened 2021-05-31
[DataGrid] Uncatchable Errors thrown if row validation fails


xadie:
- [x ] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When row validation fails the exception thrown is uncatchable. (I guess function checkGridRowIdIsValid is responsible for throwing the error).

## Expected Behavior 🤔

Exception thrown should be able to be handled from the surrounding context. I would expect that if onError callback is provided, the raised error is catched and forwarded to the callback for instance.

## Steps to Reproduce 🕹

https://codesandbox.io/s/peaceful-mclean-nqcnyn?file=/src/App.tsx (v5)

- run example ~https://codesandbox.io/s/priceless-dewdney-9y9ko?file=/src/App.tsx~ (v4)
- check console output

## Context 🔦

Implementing error handling for an application that uses Datagrid for data display

dtassone:
The grid already comes with an error boundary so it should show an error overlay on errors.
There seems to be an issue with that

oliviertassinari:
Having a look at the problem, it seems that there are multiple opportunities:

1. I believe that the main argument we have used in the past to support the idea that the data grid needs a built-in error boundary is that showing a broken UI is worth that showing nothing. We all agreed to this. React as well, so much so that they made it the default behavior:
> errors that were not caught by any error boundary will result in unmounting of the **whole** React component tree. We debated this decision, but in our experience, it is worse to leave corrupted UI in place than to completely remove it.

https://reactjs.org/docs/error-boundaries.html

So it sounds to me that this argument has no foundations after all. We can remove the error boundary and still get the same outcome. @dtassone What's left to defend that we need a built-in ErrorBoundary?

2. The data grid stops the normal propagation of the error: https://codesandbox.io/s/elated-stallman-8y1cy?file=/src/App.tsx. In order to restore the default behavior, users have to rethrow it. Why should any component interfere with how developers want to implement error handling?
@xadie Would it help if there were no `onError` prop and you could use your own `ErrorBoundary` everywhere in the application? as suggested in:

> In practice, most of the time you’ll want to declare an error boundary component once and **use it** throughout your application.

https://reactjs.org/docs/error-boundaries.html

dtassone:
[...]

PR #1823, opened 2021-06-03
[DataGrid] Implement useControlState hook, and add control state on selectionModel

dtassone:
https://deploy-preview-1823--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-filter--control-selection

This PR allows integrating the concept of control state, within the grid state management by registering the controlled prop, the on-change handler and the state selector which allows retrieving where the prop model is stored in state.

Below is how you can register the prop `selectionModel` and its `onSelectionModelChange` as a controlled state prop.

```tsx
// You need to register a ControlStateItem to bind a model prop and onChange with its controlled state

export interface ControlStateItem<TModel> {
  stateId: string;
  propModel?: any;
  stateSelector: (state: GridState) => TModel;
  propOnChange?: (model: TModel) => void;
  onChangeCallback?: (model: TModel) => void;
}

  React.useEffect(() => {
    apiRef.current.registerControlState<GridSelectionModel>({
      stateId: 'selectionModel',
      propModel: props.selectionModel,
      propOnChange: props.onSelectionModelChange,
      stateSelector: gridSelectionStateSelector,
      onChangeCallback: (model: GridSelectionModel) => {
        apiRef.current.publishEvent(GRID_SELECTION_CHANGE, model);
      },
    });
  }, [apiRef, props.onSelectionModelChange, props.selectionModel]);
```

At this stage, the architecture seems to be working, and performing well. 

- [x] implement control selection
- [x] add tests and stories
- [x] remove legacy code
- [x] Remove onRowSelected & GridRowSelectedParams  

### Breaking Change

```diff
- props.onRowSelected
```
Use `onSelectionChange`.

### Next

- [ ] RegisterControlState for filter
- [ ] RegisterControlState for SortModel
- [ ] RegisterControlState for pagination substates so one for page, one for pagesize
- [ ] RegisterControlState for editRowsModel, not sure how it is going to work yet, is `mapStateToModel` enough 🤔 
- [ ] Cleanup, Remove redundant publishevent code, cleanup hook control mode, Fix publish event params (inconsistency btw callback and hook body)


packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+} from 'test/utils';
+import { expect } from 'chai';
+import { XGrid, GridComponentProps } from '@material-ui/x-grid';
+import { getCell, getRow } from 'test/utils/helperFn';
+import { spy } from 'sinon';
+import { GridSelectionModel } from '../../../_modules_/grid/models/gridSelectionModel';
+
+const isJSDOM = /jsdom/.test(window.navigator.userAgent);
+
+describe('<DataGrid /> - Control', () => {
```
oliviertassinari, line 15: This seems out of place. If you are testing the selection, shouldn't be moved to the selection test file?

packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    it('should not update the selection model when the selectionModelProp is set', () => {
+      const selectionModel: GridSelectionModel = [1];
+      render(<TestCase selectionModel={selectionModel} />);
+
+      expect(getRow(0)).to.not.have.class('Mui-selected');
+      expect(getRow(1)).to.have.class('Mui-selected');
+      fireEvent.click(getCell(0, 0));
+      expect(getRow(0)).to.not.have.class('Mui-selected');
```
oliviertassinari, line 69: again, split infinitive
```suggestion
      expect(getRow(0)).not.to.have.class('Mui-selected');
```

packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {
+  createClientRenderStrictMode,
+  // @ts-ignore
+  fireEvent,
+} from 'test/utils';
+import { expect } from 'chai';
+import { XGrid, GridComponentProps } from '@material-ui/x-grid';
+import { getCell, getRow } from 'test/utils/helperFn';
+import { spy } from 'sinon';
+import { GridSelectionModel } from '../../../_modules_/grid/models/gridSelectionModel';
```
oliviertassinari, line 11: Wrong import path

packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+      const ControlCase = (props: Partial<GridComponentProps>) => {
+        const { rows, columns, ...others } = props;
+        const [selectionModel, setSelectionModel] = React.useState<any>([0]);
+        const handleSelectionChange = React.useCallback((newModel) => {
+          if (newModel.length) {
+            setSelectionModel([...newModel, 2]);
+            return;
+          }
+          setSelectionModel(newModel);
+        }, []);
```
oliviertassinari, line 92: Could we remove the useCallback here? To make sure the API is flexible in the test

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -239,4 +211,23 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
+      propModel: props.selectionModel,
+      propOnChange: props.onSelectionModelChange,
+      stateSelector: (state) => state.selection,
+      mapStateToModel: (selectionState) => {
+        if (!selectionState) {
+          return undefined;
+        }
+        const model = Object.values(selectionState);
+        return model;
+      },
```
oliviertassinari, line 227: Something seems off. Why don't we store the model as is, in the state, and have another place for transformed value?

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -2,30 +2,89 @@ import * as React from 'react';
...
+      let shouldUpdate = true;
+      const updatedStateIds: string[] = [];
+      const controlStateMap = apiRef.current.controlStateRef.current!;
+
+      if (hasChanged && apiRef.current.controlStateRef) {
+        Object.keys(controlStateMap).forEach((stateId) => {
+          const controlState = controlStateMap[stateId];
+          const oldState = controlState.stateSelector(apiRef.current.state);
+          const newSubState = controlState.stateSelector(newState);
+          const hasSubStateChanged = !isDeepEqual(oldState, newSubState);
```
oliviertassinari, line 32: A performance bottleneck
```suggestion
          const hasSubStateChanged = oldState === newSubState;
```

docs/src/pages/components/data-grid/selection/ControlledSelectionGrid.tsx
```diff
@@ -15,8 +15,8 @@ export default function ControlledSelectionGrid() {
     <div style={{ height: 400, width: '100%' }}>
       <DataGrid
         checkboxSelection
-        onSelectionModelChange={(newSelection) => {
-          setSelectionModel(newSelection.selectionModel);
+        onSelectionModelChange={(newSelectionModel) => {
+          setSelectionModel(newSelectionModel);
```
oliviertassinari, line 19: I like this 👍

oliviertassinari: I would have personally done
```suggestion
        onSelectionModelChange={(event, newSelectionModel, details) => {
          setSelectionModel(newSelectionModel);
```

but it's already one step in the right direction, IMHO

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -2,30 +2,89 @@ import * as React from 'react';
...
+            );
+          }
+
+          if (hasSubStateChanged) {
+            if (controlState.propOnChange) {
+              const newModel = controlState.mapStateToModel
+                ? controlState.mapStateToModel(newSubState)
+                : newSubState;
 
+              if (!isDeepEqual(controlState.propModel, newModel)) {
```
oliviertassinari, line 49: Please, don't make deep equal a norm

```suggestion
              if (controlState.propModel === newModel) {
```

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -2,30 +2,89 @@ import * as React from 'react';
...
+                controlState.propOnChange(newModel);
+                shouldUpdate = controlState.propModel === undefined;
+              }
+            } else if (!controlState.propOnChange && controlState.propModel !== undefined) {
+              const oldModel = controlState.mapStateToModel
+                ? controlState.mapStateToModel(oldState)
+                : oldState;
+              shouldUpdate = !isDeepEqual(oldModel, controlState.propModel);
```
oliviertassinari, line 57:
```suggestion
              shouldUpdate = oldModel !== controlState.propModel;
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -239,4 +211,23 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
     });
     forceUpdate();
   }, [apiRef, setGridState, forceUpdate, isRowSelectable]);
+
+  React.useEffect(() => {
+    apiRef.current.registerControlState<GridSelectionModel, GridSelectionState>({
```
oliviertassinari, line 216: "register" would suggest it's run one and should be cleanup, It's not what happens.

```suggestion
    apiRef.current.updateControlState<GridSelectionModel, GridSelectionState>({
```

oliviertassinari (changes requested):
I don't think that the approach is flying. Changing a controlled prop should **directly** trigger a render, it doesn't.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
oliviertassinari: Oh boy, it's does indeed register. This can't work.

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -2,30 +2,89 @@ import * as React from 'react';
...
+            if (controlState.propOnChange) {
+              const newModel = controlState.mapStateToModel
+                ? controlState.mapStateToModel(newSubState)
+                : newSubState;
 
+              if (!isDeepEqual(controlState.propModel, newModel)) {
+                controlState.propOnChange(newModel);
+                shouldUpdate = controlState.propModel === undefined;
+              }
+            } else if (!controlState.propOnChange && controlState.propModel !== undefined) {
```
m4theushw, line 53: This branch is already only executed when there's no `controlState.propModel`.

```suggestion
            } else if (controlState.propModel !== undefined) {
```

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -2,30 +2,89 @@ import * as React from 'react';
...
   const forceUpdate = React.useCallback(
     () => apiRef.current.forceUpdate(() => apiRef.current.state),
     [apiRef],
   );
+
   const setGridState = React.useCallback(
     (stateUpdaterFn: (oldState: GridState) => GridState) => {
       const newState = stateUpdaterFn(apiRef.current.state);
       const hasChanged = apiRef.current.state !== newState;
+      let shouldUpdate = true;
```
m4theushw, line 23: Shouldn't it begin as `false` to update the state as little as possible? If I have an `controlState.propOnChange` set and the new model is equal to `controlState.propModel` it publishes the state change event.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -184,8 +158,6 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
   );
 
   useGridApiEventHandler(apiRef, GRID_ROW_CLICK, handleRowClick);
-  useGridApiOptionHandler(apiRef, GRID_ROW_SELECTED, onRowSelected);
```
m4theushw, line 187: Don't forget to remove this event from the constants file and to list it as a breaking change here.

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -392,21 +391,16 @@ export interface GridOptions {
...
-  onRowSelected?: (param: GridRowSelectedParams) => void;
   /**
    * Callback fired when the grid is resized.
    * @param param With all properties from [[GridResizeParams]].
    */
   onResize?: (param: GridResizeParams) => void;
   /**
    * Callback fired when the selection state of one or multiple rows changes.
-   * @param param With all properties from [[SelectionChangeParams]].
+   * @selectionModel with all the row ids [[GridRowId[]]].
```
m4theushw, line 401:
```suggestion
   * @param selectionModel with all the row ids [[GridRowId]][].
```

m4theushw (commented):
We have to manually call `setGridState` to update the state when the prop changes. Should we update the hook to also handle this update: prop -> state? I think we're gonna need a `mapModelToState` for that to sanitize the values that go the state (e.g. to unselect invisible rows).

https://github.com/mui-org/material-ui-x/blob/a2a1f8eca0f8b188a1ed40abbcc616dc3a837d76/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L217-L222

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
dtassone: because sometimes we need to change the state to be more optimized for retrieving data

packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
dtassone: yeah it's the auto import

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
dtassone: It runs once until you change the props. 
I think register is great as you tell the state that this prop and onChange are controlled. You don't update the state itself... Up to you!

packages/grid/data-grid/src/tests/control.DataGrid.test.tsx
dtassone: depends if you consider that you test the control feature or the selection

oliviertassinari: I would vote for testing the selection feature in-depth, as a end-to-end stress test, but not strong point of view.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
oliviertassinari: I have just made the connection, it's related to https://github.com/mui-org/material-ui-x/pull/1864#discussion_r652278480. When implementing the control pattern. If we transform the controlled prop into a different shape and use it as a the source of truth, then we need a method to do the opposite *OR* to store the cache, somewhere else.

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -235,12 +235,6 @@ export const GRID_ROW_LEAVE = 'rowLeave';
  */
 export const GRID_ROW_EDIT_MODEL_CHANGE = 'editRowModelChange';
 
-/**
- * Fired when a row is selected or unselected. Called with a [[GridRowSelectedParams]] object.
- * @event
- */
-export const GRID_ROW_SELECTED = 'rowSelected';
```
DanailH, line 242: 👍 I remember we discussed removing this event. Just need to remember to add it in the breaking changes list.

packages/grid/_modules_/grid/hooks/features/core/gridState.ts
```diff
@@ -33,28 +33,28 @@ import {
 } from '../virtualization/renderingState';
 
 export interface GridState {
-  rows: InternalGridRowsState;
-  editRows: GridEditRowsModel;
-  pagination: GridPaginationState;
-  options: GridOptions;
-  isScrolling: boolean;
   columns: GridInternalColumns;
+  columnMenu: GridColumnMenuState;
```
DanailH, line 37: I understand why we need to keep the options in the state but I'm a but concerned that we are introducing an extra step here. My rationale is that if I am providing options to the grid I'm not really interested in the state, i just want to force it. What would be the difference between having options and just exposing the entire state (which is already) the case and making developers use that to "control" it. It seems like 2 different ways to do the same thing in the end. Maybe we can consider dropping one of the 2 options (probably hide the state).

dtassone: Just to be clear, the state has not changed, I just reordered it alphabetically.

oliviertassinari: Then maybe what we should learn from this thread is that it would have been clearer to do the sorting in a dedicated PR.

dtassone: ok I revert

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
dtassone: by default it should let the state update unless it is controlled. This is not to control the number of state updates

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
dtassone: we can't === here at this stage due to the mapStateToModel that does an object.values on the state so === will never be equal. So atm if you have no mapStateToModel then it will use the === otherwise deep equal.

dtassone: looking to refactor

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
dtassone: Considering https://github.com/mui-org/material-ui-x/pull/1864#discussion_r654336662
I think we should stop transforming models in state and just transform them in selectors. So the state, would/should be the same as the props. And we can fix the === and remove mapModelToState as well

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
oliviertassinari: My issue with "register" is that it implicitly suggests we should also unregister to cleanup, which we don't. It seems to be a setter.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
dtassone: All done as mentioned above so `mapStateToModel` is gone

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -197,43 +166,64 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
-    const currentModel = Object.values(apiRef.current.getState().selection);
-    if (!isDeepEqual(currentModel, selectionModel)) {
-      apiRef.current.setSelectionModel(selectionModel || []);
+    // prop selectionModel changed
+    if (props.selectionModel === undefined) {
+      return;
+    }
+    const currentModel = apiRef.current.getState().selection;
+    if (currentModel !== props.selectionModel) {
+      setGridState((state) => ({ ...state, selection: props.selectionModel || [] }));
```
m4theushw, line 208: By not using `setSelectionModel`, the user can hijack the grid and create a DataGrid with multiple selection. Look https://codesandbox.io/s/material-demo-forked-essgk?file=/demo.js

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -197,43 +166,64 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
...
   React.useEffect(() => {
-    const currentModel = Object.values(apiRef.current.getState().selection);
-    if (!isDeepEqual(currentModel, selectionModel)) {
-      apiRef.current.setSelectionModel(selectionModel || []);
+    // prop selectionModel changed
+    if (props.selectionModel === undefined) {
+      return;
+    }
+    const currentModel = apiRef.current.getState().selection;
+    if (currentModel !== props.selectionModel) {
```
m4theushw, line 207: I think this is redundant because `applyControlStateConstraint` only updates if the model is different from the older one.

```suggestion
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -197,43 +166,64 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
   useGridApiMethod(apiRef, selectionApi, 'GridSelectionApi');
 
   React.useEffect(() => {
+    apiRef.current.registerControlState<GridSelectionModel>({
+      stateId: 'selectionModel',
```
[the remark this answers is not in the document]
DanailH: I guess because that is the name of the prop that is being controlled?

packages/grid/_modules_/grid/models/api/gridControlStateApi.ts
```diff
@@ -0,0 +1,22 @@
+import { GridState } from '../../hooks/features/core/gridState';
+import { ControlStateItem } from '../controlStateItem';
+
+/**
+ * The control state API interface that is available in the grid `apiRef`.
+ */
+export interface GridControlStateApi {
+  /**
+   * Register a control state that binds the model, the onChange prop, and the grid state together.
+   * @param [ControlStateItem<TModel, TState>]
```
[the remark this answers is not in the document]
DanailH: `to be registered`? 😄

packages/grid/_modules_/grid/hooks/root/useGridApiMethod.ts
```diff
@@ -12,16 +12,25 @@ export function useGridApiMethod<T extends Partial<GridApi>>(
...
+
+  React.useEffect(() => {
+    apiMethodsRef.current = apiMethods;
+  }, [apiMethods]);
+
+  React.useEffect(() => {
+    installMethods();
+  }, [installMethods]);
+
+  installMethods();
```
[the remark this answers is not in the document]
dtassone: to install api methods earlier in the grid lifecycle

packages/grid/_modules_/grid/useGridComponent.tsx
```diff
@@ -37,6 +38,7 @@ export const useGridComponent = (apiRef: GridApiRef, props: GridComponentProps)
   useLoggerFactory(apiRef, props);
   useApi(apiRef);
   useErrorHandler(apiRef, props);
+  useGridControlState(apiRef);
```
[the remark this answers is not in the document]
dtassone: the goal is to have error handling as early as possible as if there is an error in a hook before the error handling it won't be handled. Also as mentioned in #1804, it needs to be moved to a parent component

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
dtassone: what if you don't use control state

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -36,6 +37,7 @@ export type DataGridProps = Omit<
   pagination?: true;
   apiRef?: undefined;
   onRowsScrollEnd?: undefined;
+  selectionModel?: GridRowId | GridRowId[];
```
m4theushw, line 40: Should we update https://deploy-preview-1823--material-ui-x.netlify.app/api/data-grid/data-grid/?

packages/grid/_modules_/grid/models/api/gridControlStateApi.ts
```diff
@@ -0,0 +1,22 @@
+import { GridState } from '../../hooks/features/core/gridState';
+import { ControlStateItem } from '../controlStateItem';
+
+/**
+ * The control state API interface that is available in the grid `apiRef`.
+ */
+export interface GridControlStateApi {
+  /**
+   * Updates a control state that binds the model, the onChange prop, and the grid state together.
+   * @param {ControlStateItem<TModel>} controlState The [[ControlStateItem]] to be registered.
```
m4theushw, line 10:
```suggestion
   * @param {ControlStateItem<TModel>} controlState The [[ControlStateItem]] to be registered.
   * @ignore - do not document.
```

packages/grid/_modules_/grid/models/api/gridControlStateApi.ts
```diff
@@ -0,0 +1,22 @@
...
+   */
+  updateControlState: <TModel>(controlState: ControlStateItem<TModel>) => void;
+  /**
+   * Allows the internal grid state to apply the registered control state constraint.
+   * @param {GridState} state The new modified state that would be the next if the state is not controlled.
+   * @returns {shouldUpdate: boolean, postUpdate: () => void}, shouldUpdate let the state know if it should update, and postUpdate is a callback function triggered if the state has updated.
```
m4theushw, line 16:
```suggestion
   * @returns {shouldUpdate: boolean, postUpdate: () => void}, shouldUpdate let the state know if it should update, and postUpdate is a callback function triggered if the state has updated.
   * @ignore - do not document.
```

packages/grid/data-grid/src/DataGrid.tsx
dtassone: yes done

dtassone:
create single issue for the next todos

m4theushw (approved):

oliviertassinari:
@dtassone Could you update the PR to document the breaking changes? We are going to need it for the next release's changelog and for developers landing here. BTW, it has been your turn for the next release for the last 16 days. Are you up for making one Monday? :)

Bixtrix:
@dtassone @oliviertassinari It would have been nice to wrote that the multiple selection (onSelectionModelChange as array input) [now requires XGrid component](https://github.com/mui-org/material-ui-x/pull/1823/files#diff-36939e58755c91e3693a4f18aa7796b137ba1e32f65357a940fb358286048088R249) in the release notes

Tai-Iro:
Why is the error for multiple selections a thing? It's not mentioned anywhere in the docs or release notes, and it doesn't even stop multiple selections from working.
This change: https://github.com/mui-org/material-ui-x/pull/1823/files#diff-36939e58755c91e3693a4f18aa7796b137ba1e32f65357a940fb358286048088R247

oliviertassinari:
> now requires XGrid component in the release notes
> Why is the error for multiple selections a thing?

Note that it has always been documented under the Pro plan in https://material-ui.com/components/data-grid/selection/#multiple-row-selection. If it was possible to make work in `DataGrid` before, it was never intended nor documented.

Tai-Iro:
> > @Bixtrix multiple selection now requires XGrid component in the release notes
> > @Tai-Iro Why is the error for multiple selections a thing?
> 
> Note that multiple row selections has always been documented under the Pro plan in https://material-ui.com/components/data-grid/selection/#multiple-row-selection.
> 
> If it was possible to make work in `DataGrid` before with the `selectionModel` prop, it was a mistake from our side, sorry for the surprise. It was never intended nor documented.
> 

It was directly documented in the Controlled Selection example here: https://material-ui.com/components/data-grid/selection/#controlled-selection
And every example that uses checkboxes allows selecting multiple items, of which there are several.

m4theushw:
Yeah, with `checkboxSelection` it should allow multiple selection even with DataGrid. We missed an additional check here:

https://github.com/mui-org/material-ui-x/blob/3d1e5d8157bb12c2f27ef1223bc169890d5404db/packages/grid/data-grid/src/DataGrid.tsx#L254

Previously, there was no warning and the validation was done inside the hook.

PR #1828, opened 2021-06-04
[DataGrid] Fix preact try

dtassone:
https://github.com/preactjs/preact/issues/3055

### DIFF
--- a/packages/grid/_modules_/grid/hooks/root/useGridApiMethod.ts
+++ b/packages/grid/_modules_/grid/hooks/root/useGridApiMethod.ts
@@ -12,16 +12,25 @@ export function useGridApiMethod<T extends Partial<GridApi>>(
   const apiMethodsRef = React.useRef(apiMethods);
   const [apiMethodsNames] = React.useState(Object.keys(apiMethods));
 
-  React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


oliviertassinari:
@dtassone Do you have a report on how the exploration with Preact went? I'm asking because the community might want to know. For instance, they might resume the work based on what we have found.

packages/grid/_modules_/grid/hooks/root/useGridApiMethod.ts
```diff
@@ -12,16 +12,25 @@ export function useGridApiMethod<T extends Partial<GridApi>>(
...
+
+  React.useEffect(() => {
+    apiMethodsRef.current = apiMethods;
+  }, [apiMethods]);
+
+  React.useEffect(() => {
+    installMethods();
+  }, [installMethods]);
+
+  installMethods();
```
dtassone, line 35: The issue with Preact comes from a missing method `(getLocaleText)` on `apiRef`, in this PR I have tried to install the methods at an earlier stage of the React component lifecycle. 
While that attempt didn't work, I believe that if we manage to build `apiRef` at an earlier stage of the init process then it would fix this first Preact issue.

PR #1829, opened 2021-06-04
[DataGrid] Fix force reflow on scroll start and end

dtassone:
Following conversation from https://github.com/mui-org/material-ui-x/issues/1812#issuecomment-853267525

Preview: https://deploy-preview-1829--material-ui-x.netlify.app/components/data-grid/#commercial-version

**Before**

Capture d’écran 2021-06-04 à 13 25 37

https://codesandbox.io/s/material-demo-forked-7s1vw

**After**

Capture d’écran 2021-06-04 à 13 25 53

https://codesandbox.io/s/material-demo-forked-90ccn


oliviertassinari:
I have updated the PR's description with likes to the reproduction before and after. The main difference is about the initial delay before we see the page scrolling with CPU throttling at x4. The difference is marginal without the CPU throttling. So it doesn't seem to be a major win, it doesn't seem to harm either so 🤷‍♂️

oliviertassinari (approved):

PR #1831, opened 2021-06-05
[DataGrid] Fix React concurrent updates

### LINKED ISSUE #1784: [DataGrid] Rows are not rendered

## Current Behavior 😯
Sometimes the rows are not rendered.

## Steps to Reproduce 🕹
https://codesandbox.io/s/determined-haibt-ikjl1?file=/src/demo.js
The problem does not always occur. I have tried to reproduce it via codesandbox.

Open the example and refresh the browser.

image

## Context 🔦
In my application, I use Redux in conjunction with a web service. The first two renderings pass an empty array to the DataGrid, the third rendering contains the data.
When this happens very fast it happens again and again that I get empty rows.

I have tried to break it down as far as possible in the example.

## Order id 💳

#23119

oliviertassinari:
fix #1784

Preview: https://deploy-preview-1831--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-rows--defer-rendering

This method is triggered **twice** in the reproduction given in the GitHub Issue:

https://github.com/mui-org/material-ui-x/blob/4a57468f36acb3be4910693df733c0c1f5a20f85/packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts#L235

Inside this method, we schedule 3 updates, but unfortunately, React doesn't guarantee that the 3 updates (`updateStateIfChanged`) are done in the same order. What happens is that the last update that runs is not using the latest values of the state, leading to the bug. In order to solve it, we can use `useEventCallback` that is precisely meant for this problem: https://github.com/facebook/react/issues/14099#issuecomment-440013892. An alternative solution would be to read from the state directly in `updateStateIfChanged` instead of from the parent scope but it seems a lot of work to refactor.

The solution in #1807 focuses on avoiding the concurrent call to `refreshContainerSizes` to happen in the first place, wish sounds like being more efficient and that can be pursued as an independent effort to improve performance or even simplify the logic.

I have opened a new PR to emphasize what I think that we should expect from PR fixing bugs. A [failing](https://app.circleci.com/pipelines/github/mui-org/material-ui-x/5118/workflows/3dbe811f-8e9b-4958-8f55-79e5fbc8cdcd/jobs/25623) test case (cb7d37b) that is fast to iterate on, a description of the origin of the issue. And a fix, the closer to the root cause, the better. I'm spending time on this because I think this rigor is important in order to take over the incumbent solution, in a mature market.

Preview:

### DIFF
--- a/test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
+++ b/test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
@@ -7,8 +7,7 @@ export default function ConcurrentReactUpdate() {
   const [rows, setRows] = React.useState<GridRowModel[]>([]);
 
   React.useEffect(() => {
-    setTimeout(() => setRows([]), 0);
-    setTimeout(() => setRows([{ id: 1 }, { id: 2 }]), 1);
+    setTimeout(() => setRows([{ id: 1 }, { id: 2 }]), 0);
   }, []);
 
   return <DataGrid autoHeight columns={columns} rows={rows} />;
--- a/test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
+++ b/test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
@@ -1,0 +1,15 @@
+import * as React from 'react';
+import { DataGrid, GridRowModel } from '@material-ui/data-grid';
+
+const columns = [{ field: 'id', headerName: 'Id', width: 100 }];
+
+export default function ConcurrentReactUpdate() {
+  const [rows, setRows] = React.useState<GridRowModel[]>([]);
+
+  React.useEffect(() => {
+    setTimeout(() => setRows([]), 0);
+    setTimeout(() => setRows([{ id: 1 }, { id: 2 }]), 1);
+  }, []);
+
+  return <DataGrid autoHeight columns={columns} rows={rows} />;
+}
--- a/packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
+++ b/packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
@@ -11,6 +11,7 @@ import { isDeepEqual } from '../../utils/utils';
 import { gridColumnsTotalWidthSelector } from '../features/columns/gridColumnsSelector';
 import { GridState } from '../features/core/gridState';
 import { useGridSelector } from '../features/core/useGridSelector';
+import { useEventCallback } from '../../utils/material-ui-utils';
 import { useGridState } from '../features/core/useGridState';
 import { gridDensityRowHeightSelector } from '../features/density/densitySelector';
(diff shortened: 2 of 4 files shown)


test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { DataGrid, GridRowModel } from '@material-ui/data-grid';
+
+const columns = [{ field: 'id', headerName: 'Id', width: 100 }];
+
+export default function ConcurrentReactUpdate() {
```
dtassone, line 6: is this test not working on my PR?

oliviertassinari: When I tried to run https://github.com/mui-org/material-ui-x/pull/1807/files#diff-4b13b5b8ec20f6d20174b3716e9e03501d374fe0ca6b92c5faeb5ea20f3282b2R396 on master, it was passing.

oliviertassinari: Note that this test only runs on ChromeHeadless, not on Safari like in #1807, so definitely a tradeoff.

dtassone: then why don't we put it in my PR?

oliviertassinari: We can, for sure. The idea was to demonstrate a better test, and a fix closer to the root cause.

oliviertassinari:
I would be curious to try React 18 out. Per, https://github.com/reactwg/react-18/discussions/21, the order conflict might be solved.

test/e2e/fixtures/DataGrid/ConcurrentReactUpdate.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const [rows, setRows] = React.useState<GridRowModel[]>([]);
+
+  React.useEffect(() => {
+    setTimeout(() => setRows([]), 0);
+    setTimeout(() => setRows([{ id: 1 }, { id: 2 }]), 1);
```
oliviertassinari, line 11: Simpler test case

```suggestion
    setTimeout(() => setRows([{ id: 1 }, { id: 2 }]), 0);
```

dtassone:
Closing as we have #1807. It has all tests...

There is a regression on scrollbars. 

Screenshot 2021-06-17 at 12 17 47

oliviertassinari:
https://deploy-preview-1831--material-ui-x.netlify.app/components/data-grid/columns/#column-menu

oliviertassinari:
Regarding why it doesn't work, it seems to be because `useEventCallback` is plain wrong for the use case. We need to depend on the dependencies to rerun to update the state.

PR #1856, opened 2021-06-08
[docs] Add docs for `disableDensitySelector` option

### LINKED ISSUE #1851: [DataGrid] api missing disableDensitySelector
On the XGrid api page  - https://material-ui.com/api/x-grid/ is missing the disableDensitySelector, which is available in the code as I see it here: https://github.com/mui-org/material-ui-x/blob/f5642ae7fe19ed8b8998182c1a6f49c2ff7bb149/packages/grid/_modules_/grid/components/toolbar/GridDensitySelector.tsx
    // Disable the button if the corresponding is disabled    if (options.disableDensitySelector) {      return null;    }

This is also for the disableColumnFilter.

DanailH:
fix #1851

### DIFF
--- a/docs/pages/api-docs/data-grid/data-grid.md
+++ b/docs/pages/api-docs/data-grid/data-grid.md
@@ -29,7 +29,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">density</span> | <span class="prop-type">Density</span> | standard | Sets the density of the grid. |
 | <span class="prop-name">disableColumnMenu</span> | <span class="prop-type">boolean</span> | false | If `true`, the column menu is disabled. |
 | <span class="prop-name">disableColumnSelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the column selector is disabled. |
-| <span class="prop-name">disableDensitySelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the density selection is disabled. |
+| <span class="prop-name">disableDensitySelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the density selector is disabled. |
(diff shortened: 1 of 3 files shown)


docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -29,6 +29,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">density</span> | <span class="prop-type">Density</span> | standard | Sets the density of the grid. |
 | <span class="prop-name">disableColumnMenu</span> | <span class="prop-type">boolean</span> | false | If `true`, the column menu is disabled. |
 | <span class="prop-name">disableColumnSelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the column selector is disabled. |
+| <span class="prop-name">disableDensitySelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the density selection is disabled. |
```
m4theushw, line 32: Could we replace "selection" with "selector" in `GridOptions` to make it consistent with the other props?

```suggestion
| <span class="prop-name">disableDensitySelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the density selector is disabled. |
```

dtassone (approved):

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -99,7 +99,7 @@ export interface GridOptions {
    */
   disableColumnSelector?: boolean;
   /**
-   * If `true`, density selection is disabled.
+   * If `true`, density selector is disabled.
```
oliviertassinari, line 102: ?
It doesn't match the label in the documentation.

```suggestion
   * If `true`, the density selector is disabled.
```

Issue #1440, opened 2021-04-19
[DataGrid] Customize separator in CSV export


friedensstifter:
Hi,

i've worked a little bit with Data Grid 

1) ~Can you make the separator for csv export selectable? In Germany its a semicolon, not a comma. It would be great to export it with an ; instead of ,~ Fixed in #1859 
2) ~Can you add the choice, if a column should not be exported to csv? I'm creating links in some columns, they should not be exported. And i haven't found a way to prevent it~ Duplicate of #1435
3) ~The Coding of the file show cryptic letters instead of german umlaut. So, if i export the word "FakultÃ¤t" (Fakultät), Excel dont uses utf-8 for file coding. If i open the csv over Data import from file i can change the coding of the file. Is there a way to export in itf-8?~ Fixed in #1695 

Thanks,
friedensstifter

oliviertassinari:
For point 2. It's already reported in #1435

michallukowski:
The solution for point 3 would be to add sequence of bytes (0xEF, 0xBB, 0xBF) at the beginning of a csv file. This would allow Excel to automatically detect file encoding as UTF-8, but according to the Unicode standard, the BOM for UTF-8 files is not recommended. The implementation of this solution is very simple and requires replacement in useGridCsvExport.tsx: 

```const blob = new Blob([csv], { type: 'text/csv' });```

by

```const blob = new Blob([new Uint8Array([0xEF, 0xBB, 0xBF]),  [csv]], { type: 'text/csv' });```

It should be optional and by default it should work as it currently is. 

It would be great to add options in the GridToolbarExport that would allow you to configure: 

- csv separator: ",", ";", " " or any passed by developer
- decimal separator: "." or ","
- flag to generate UTF-8 or UTF-8 with BOM

**Comment by dtassone** [comment]
@michallukowski maybe you could do a PR for the UTF 8 fix 🤔

**Comment by oliviertassinari** [comment]
Regarding point 3. it was fixed in #1695 with:

```jsx
<GridToolbarExport
  csvOptions={{
    utf8WithBom: true,
  }}
/>
```

We could probably add documentation about this `utf8WithBom` new property. It's empty:

https://github.com/mui-org/material-ui-x/blob/fdeb8680fc1add53dc86b7b8d832b180abe18a8e/packages/grid/_modules_/grid/models/gridExport.ts#L6

Maybe:

```diff
[...]

PR #1859, opened 2021-06-09
[DataGrid] Allow to set in GridExportCsvOptions delimiter

michallukowski:
Add option in GridExportCsvOptions to set delimiter character

Fix 1. of #1440


dtassone:
FYI the comma as decimal separator is according to the culture. So in the uk they use 1, 000 and in Fr, they use 1 000 I think.
So not sure we should add it as an option. Maybe we could think of an excel formatter as @m4theushw mentioned somewhere (trying to recover his comment)

michallukowski:
> 
> 
> FYI the comma as decimal separator is according to the culture. So in the uk they use 1, 000 and in Fr, they use 1 000 I think.
> So not sure we should add it as an option. Maybe we could think of an export formatter as @m4theushw mentioned somewhere (trying to recover his comment)

You're talking about the thousand separator. Using Number.prototype.toString() we don't get the thousands separator which I think we shouldn't use in CSV. For example, in many European countries, the number 3.14 in the csv file is automatically treated by Excel as a string. If the decimal separator is ',' (3.14), Excel detects as a number. 
[https://en.wikipedia.org/wiki/Decimal_separator](https://en.wikipedia.org/wiki/Decimal_separator)

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
```diff
@@ -6,24 +6,43 @@ import {
   GridRowModel,
 } from '../../../../models';
```
oliviertassinari, line 8: No blank lines in the import

```suggestion
```

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -1,8 +1,34 @@
...
+ */
+export type CsvDelimiterCharacter = ',' | ';';
+
 /**
  * The options to apply on the CSV export.
  */
 export interface GridExportCsvOptions {
+  /**
+   * If `true`, the comma character used to separate the integer part from the fractional part of a number.
+   * This flag also changes the delimiter from `,` to `;`.
```
oliviertassinari, line 13: typo
```suggestion
   * This flag also changes the delimiter from `,` to `.`.
```

m4theushw:
@michallukowski I proposed in https://github.com/mui-org/material-ui-x/issues/1814#issuecomment-852678630 that we apply the value formatter to the CSV export too. With it, you would be able to change how numbers are formatted when exported. This is what Damien refers to. Regarding this PR, I would avoid adding specific options to control number formatting. We could leverage [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) that already provides everything. We could ask the user to pass a `Intl.NumberFormat` instance and use it to format values.

> Using Number.prototype.toString() we don't get the thousands separator which I think we shouldn't use in CSV.

`Number.prototype.toLocaleString()` gives the formatted value according to the current locale.

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -1,8 +1,34 @@
...
+   * @default false
+   */
+  commaAsDecimalSeparator?: boolean;
```
oliviertassinari, line 16: Could we remove this option and move it to a standalone PR? It seems to create friction for another change this we could merge as-is.

oliviertassinari (commented):
Can we add a test case for the new `delimiter` param? 

> Using Number.prototype.toString() we don't get the thousands separator which I think we shouldn't use in CSV.

Agree, I don't see why we would need to format a number in CSV export, it seems better to be able to rely on the CSV reader to automatically detect a number and handle it accordingly (e.g. for formatting based on the locale of the CSV reader or further math operations).

dtassone:
Yes so basically don't do anything special for numbers, we should consider using formatters on export

packages/grid/_modules_/grid/models/gridExport.ts
michallukowski: It's not a typo. When decimal separator is ',' then delimiter shouldn't also be ',' so it is automatically changed to ';'.

packages/grid/_modules_/grid/models/gridExport.ts
michallukowski: Ok. I'll get rid off this feature in this PR.

DanailH:
> Can we add a test case for the new `delimiter` param?
> 
> > Using Number.prototype.toString() we don't get the thousands separator which I think we shouldn't use in CSV.
> 
> Agree, I don't see **why** we would want to format a number in CSV exports. It seems better to be able to rely on the CSV reader to automatically detect a number and handle it accordingly (e.g. for formatting based on the locale of the CSV reader or further math operations).

This reminds me that I don't think that the CSV export currently uses the `valueFormatter` altho it does use `valueGetter`.

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
```diff
@@ -31,13 +31,16 @@ export function serialiseRow(
...
 }
 
-export function buildCSV(
-  columns: GridColumns,
-  rows: Map<GridRowId, GridRowModel>,
-  selectedRows: Record<string, GridRowId>,
-  getCellValue: (id: GridRowId, field: string) => GridCellValue,
-  delimiterCharacter: CsvDelimiterCharacter = ',',
-): string {
+type BuildCSVOptions = {
```
dtassone, line 34: It should be an interface

oliviertassinari: @dtassone Ah, right, my bad. Could you push this as a new commit directly on the PR so we can move forward with it?

oliviertassinari (approved):

PR #1030, opened 2021-02-11
[DataGrid] Implement CSV export

DanailH:
Fixes #197 

This is a draft, for now, to iron out the implementation but this initial version is stable and can be tested.

Let's use it to discuss the UI, the functionality, and the API.

ToDo:
- [x] Add tests
- [x] Add docs
- [x] Add translations

PR #1861, opened 2021-06-09
[DataGrid] Improve ptBR locale

aline-matos:
Add missing export translations.

### DIFF
--- a/packages/grid/_modules_/grid/locales/ptBR.ts
+++ b/packages/grid/_modules_/grid/locales/ptBR.ts
@@ -26,7 +26,7 @@ const ptBRGrid: Partial<GridLocaleText> = {
   toolbarFiltersTooltipShow: 'Exibir filtros',
   toolbarFiltersTooltipActive: (count) =>
     `${count} ${count !== 1 ? 'filtros' : 'filtro'} ${count !== 1 ? 'ativos' : 'ativo'}`,
-  
+
   // Export selector toolbar button text
   toolbarExport: 'Exportar',
   toolbarExportLabel: 'Exportar',
--- a/packages/grid/_modules_/grid/locales/ptBR.ts
+++ b/packages/grid/_modules_/grid/locales/ptBR.ts
@@ -30,7 +30,7 @@ const ptBRGrid: Partial<GridLocaleText> = {
   // Export selector toolbar button text
   toolbarExport: 'Exportar',
   toolbarExportLabel: 'Exportar',
-  toolbarExportCSV: 'Download como CSV',
+  toolbarExportCSV: 'Baixar como CSV',
 
   // Columns panel text
   columnsPanelTextFieldLabel: 'Localizar coluna',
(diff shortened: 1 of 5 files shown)


oliviertassinari:
How about we use this opportunity to add the same keys commented in the other locales?

https://github.com/mui-org/material-ui-x/blob/f5221ef2a1b4345c244cfc3704de67b39472e54a/packages/grid/_modules_/grid/constants/localeTextConstants.ts#L28-L31

It will make it easier for the other translators to handle.

packages/grid/_modules_/grid/locales/ptBR.ts
```diff
@@ -26,6 +26,11 @@ const ptBRGrid: Partial<GridLocaleText> = {
   toolbarFiltersTooltipShow: 'Exibir filtros',
   toolbarFiltersTooltipActive: (count) =>
     `${count} ${count !== 1 ? 'filtros' : 'filtro'} ${count !== 1 ? 'ativos' : 'ativo'}`,
+  
+  // Export selector toolbar button text
+  toolbarExport: 'Exportar',
+  toolbarExportLabel: 'Exportar',
+  toolbarExportCSV: 'Download como CSV',
```
m4theushw, line 33: It's better to translate the entire phrase.

```suggestion
  toolbarExportCSV: 'Baixar como CSV',
```

m4theushw:
@oliviertassinari Side note: the `toolbarExportLabel` key is not used. We could remove it from the other locales in another PR.

m4theushw (approved):

oliviertassinari:
> Side note: the toolbarExportLabel key is not used. We could remove it from the other locales in another PR.

@m4theushw Great 👁️ for the details! It was added in #1030, maybe it was refactored and forgot along the way. In this case, no need to translate it.

oliviertassinari:
> maybe it was refactored and forgot along the way. In this case, no need to translate it.

Actually, it depends, It looks like @DanailH added it so that a button can have a label if only the icon is visible. 

https://github.com/mui-org/material-ui-x/blob/45093913e390b9c71b89a3debf76a58a46e68a2f/packages/grid/_modules_/grid/components/toolbar/GridToolbarDensitySelector.tsx#L99

👍 for a new issue/a follow-up to make this consistent. I would personally argue that for the only icon use case, https://next.material-ui.com/system/screen-readers/#visually-hidden-elements has the potential to be more elegant. Less code, fewer translations.

oliviertassinari:
> maybe it was refactored and forgot along the way. In this case, no need to translate it.

@m4theushw Actually, it depends, It looks like @DanailH added it so that a button can have a label if **only the icon is visible**. 

https://github.com/mui-org/material-ui-x/blob/45093913e390b9c71b89a3debf76a58a46e68a2f/packages/grid/_modules_/grid/components/toolbar/GridToolbarDensitySelector.tsx#L99

👍 for a new issue/a follow-up to make this consistent between the different toolbar buttons.

I would personally argue that for the only icon use case, https://next.material-ui.com/system/screen-readers/#visually-hidden-elements has the potential to be more elegant. Less code, fewer translations. But If we want to make it work in a responsive way (hide the label on small screens), it might not fly.

dtassone (approved):

PR #1862, opened 2021-06-09
[DataGrid] Fix events naming

### LINKED ISSUE #1383: [RFC] Event naming conventions
## Summary 💡

This RFC proposes a set of rules on how events emitted by components and hooks should be named.

## Motivation 🔦

- The current [constants file](https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/constants/eventsConstants.ts) lacks consistency as different approaches are taken to name similar events, e.g. `GRID_CELL_DOUBLE_CLICK` and `GRID_CELL_KEYDOWN`.
- New events got to be introduced in the last refactorings, but depending on who created them, different conventions were adopted. https://github.com/mui-org/material-ui-x/pull/1289#discussion_r606541465 and https://github.com/mui-org/material-ui-x/pull/1380#discussion_r610614068.
- Facilitates the usage of the API as it's easier to guess the purpose of an event when it's named consistently.

## Proposed Solution 🧾

Based on my understanding and the discussions we had I summarised the following rules:

### 1. Prefix who publishes it

- When published by a component, use the name of the component as prefix, not the feature it was created for:

```diff
- export const GRID_COLUMN_REORDER_DRAG_OVER = 'columnReorderDragOver';
+ export const GRID_COLUMN_HEADER_DRAG_OVER = 'columnHeaderDragOver';
``` 

- The feature name should be used when emitted by a hook, e.g. `GRID_COLUMN_ORDER_CHANGE` and `GRID_COLUMN_RESIZE_STOP`. `GRID_PAGE_CHANGE` and `GRID_PAGE_SIZE_CHANGE` are exceptions as it's very clear what they do.

### 2. Use the event handler as suffix

When the event is published by a React [event handler](https://reactjs.org/docs/events.html) (`onMouseUp`, `onKeyDown`, `onDoubleClick`), the prop name should be used as suffix by converting it to uppercase and removing the "on". 

Prop name | Event constant
------------ | ----------------
`onDragOver` | `GRID_COLUMN_HEADER_DRAG_OVER`
`onClick` | `GRID_COLUMN_HEADER_CLICK`
`onDoubleClick` | `GRID_ROW_DOUBLE_CLICK`

### 3. Avoid abbreviations

```diff
- export const GRID_COL_RESIZE_START = 'colResizeStart';
+ export const GRID_COLUMN_RESIZE_START = 'columnResizeStart';
``` 

### 4. Present tense always

```diff
- export const GRID_PAGE_CHANGED = 'pageChanged';
+ export const GRID_PAGE_CHANGE = 'pageChange';

- export const GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED = 'cellEditPropsChangeCommitted';
+ export const GRID_CELL_EDIT_PROPS_CHANGE_COMMIT = 'cellEditPropsChangeCommit';
```

### 5. Constant name and value should match

The constant name is reserved for internal use only. In all demos and docs, the value of the constant is what users will be encouraged to use. In this context, the constant value should be the camelCase version of the constant name without "GRID". 

```js
export const GRID_COLUMN_HEADER_CLICK = 'columnHeaderClick';
export const GRID_SORT_MODEL_CHANGE = 'sortModelChange';
export const GRID_MULTIPLE_KEY_PRESS_CHANGE = 'multipleKeyPressChange';
```

## Unresolved Questions ❓

In rule n. 2 should the prop name (`onMouseUp`) or the DOM event (`mouseup`) be used as suffix?

m4theushw:
### Breaking changes

- [XGrid] The following events were renamed:

  1. `columnHeaderNavigationKeydown` to `columnHeaderNavigationKeyDown`
  2. `columnResizeCommitted` to `columnWidthChange`
  3. `rowsUpdated` to `rowsUpdate`
  4. `rowsCleared` to `rowsClear`
  5. `columnsUpdated` to `columnsChange`

- [XGrid] The following DOM events were removed because they were not necessary for the grid to work:

  1. `focusout`
  2. `keydown`
  3. `keyup`
  
  If you need them, attach a listener using the `apiRef`, as following:
  
  ```jsx
  apiRef.current.rootElementRef?.current?.addEventListener('focusout', event => {
    // ...
  });
  ```

---

Preview: https://deploy-preview-1862--material-ui-x.netlify.app/components/data-grid/events/#catalog-of-events
Closes #1383


packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
```diff
@@ -332,5 +332,5 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
   useGridApiEventHandler(apiRef, GRID_COLUMN_RESIZE_STOP, handleResizeStop);
 
   useGridApiOptionHandler(apiRef, GRID_COLUMN_RESIZE, options.onColumnResize);
-  useGridApiOptionHandler(apiRef, GRID_COLUMN_RESIZE_COMMITTED, options.onColumnResizeCommitted);
+  useGridApiOptionHandler(apiRef, GRID_COLUMN_WIDTH_CHANGE, options.onColumnResizeCommitted);
```
oliviertassinari, line 335: This looks inconsistent in 3 different ways:

1. `GRID_COLUMN_RESIZE` first, then `GRID_COLUMN_WIDTH_CHANGE` last.
2. `GRID_COLUMN_WIDTH_CHANGE` that fires `props.onColumnResizeCommitted()`
3. There are no events that are fired when the width change because of `flex: 1` and the container width change

oliviertassinari (commented):
I'm confused. Why do we have `GRID_COLUMN_HEADER_KEYDOWN` but `GRID_COLUMN_HEADER_NAVIGATION_KEY_DOWN`?

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
m4theushw: > I'm confused. Why do we have GRID_COLUMN_HEADER_KEYDOWN but GRID_COLUMN_HEADER_NAVIGATION_KEY_DOWN (KEY_DOWN vs. KEYDOWN)?

Fixed.

> 1. GRID_COLUMN_RESIZE fire first, then GRID_COLUMN_WIDTH_CHANGE last.

I think it would be better to remove `GRID_COLUMN_WIDTH_CHANGE` because it's only used to call `onColumnResizeCommitted`.

> 2. GRID_COLUMN_WIDTH_CHANGE that fires props.onColumnResizeCommitted()

I would open another PR to rename this prop.

> 3. There are no events that are fired when the width change because of flex: 1 and the container width change

The new column width after changing the size of the container can be retrieved by subscribing to `columnsChange`.

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -417,7 +390,7 @@ export const GRID_COLUMN_ORDER_CHANGE = 'columnOrderChange';
  * @ignore - do not document.
  * @event
  */
-export const GRID_ROWS_UPDATED = 'rowsUpdated';
+export const GRID_ROWS_UPDATE = 'rowsUpdate';
```
dtassone, line 393: GRID_ROWS_CHANGE then? 
as you remove 'update'  in columns

oliviertassinari:
> KEY_DOWN vs. KEYDOWN Fixed.

@m4theushw Are you sure? I can still see different casing used in https://deploy-preview-1862--material-ui-x.netlify.app/components/data-grid/events/#catalog-of-events. For instance `columnHeaderKeydown`, there might be mote

packages/grid/_modules_/grid/constants/eventsConstants.ts
oliviertassinari: @dtassone and how would you name the two other events? `GRID_ROWS_SET`, `GRID_ROWS_CLEAR`.

If we update the row update events to only fire one, then `GRID_ROWS_CHANGE` seem definitely better.

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
oliviertassinari: > I think it would be better to remove GRID_COLUMN_WIDTH_CHANGE because it's only used to call onColumnResizeCommitted

I don't follow. Do you mean removing the `onColumnResizeCommitted` prop?

> I would open another PR to rename this prop.

Won't this only move the problem somewhere else, making the new name of `onColumnResizeCommitted` inconsistent with `onColumnResize`?

> The new column width after changing the size of the container can be retrieved by subscribing to columnsChange.

I assume `columnsChange` also fire when we are finished to resize the columns. Which suggests that we don't need a `GRID_COLUMN_WIDTH_CHANGE` event (I mean, the part on the width changed).

m4theushw: `GRID_COLUMN_WIDTH_CHANGE` could be removed as well as the `onColumnResizeCommitted` prop. We fix `GRID_COLUMN_RESIZE` to be called in the same conditions of the removed event . We do these things in another PR.

> I assume columnsChange also fire when we are finished to resize the columns. Which suggests that we don't need a GRID_COLUMN_WIDTH_CHANGE event (I mean, the part on the "width changed").

Yes, the most reliable is `columnsChange`. We also don't need `GRID_COLUMN_RESIZE_START` or `GRID_COLUMN_RESIZE_STOP`. These events are only used inside this hook, where their handler could be called directly.

packages/grid/_modules_/grid/constants/eventsConstants.ts
m4theushw: What's the point to have `GRID_ROWS_CLEAR`? It's only used by the sorting to clear the sorting once all the records are updated. `GRID_ROWS_SET` could be used instead.

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -225,7 +225,7 @@ export function useGridColumns(apiRef: GridApiRef, { columns }: { columns: GridC
       const column = apiRef.current.getColumn(field);
       apiRef.current.updateColumn({ ...column, width });
 
-      apiRef.current.publishEvent(GRID_COLUMN_RESIZE_COMMITTED, {
+      apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE, {
```
oliviertassinari, line 228: I don't think that we have advanced the discussion enough to make this change. For instance, in #1905, we talked about adding `GRID_COLUMN_ORDER_CHANGE_COMMITTED`, on top of `GRID_COLUMN_ORDER_CHANGE`.

```suggestion
      apiRef.current.publishEvent(GRID_COLUMN_RESIZE_COMMITTED, {
```

m4theushw: In this case, `GRID_COLUMN_WIDTH_CHANGE` is the most appropriate name because it's fired every time that the width is changed. `GRID_COLUMN_RESIZE_COMMITTED` would only make sense to be fired when the user stops the resizing.

oliviertassinari: > it's fired every time that the width is changed. 

@m4theushw But not in this case: https://codesandbox.io/s/uwcz8?file=/demo.tsx, see the "id" column width

resize

The "width change" terminology would not describe when the event fire, not its purpose: to help control the column resizing feature. It wasn't introduced to subscribe to the width change, to for instance display a sticky row at the bottom (cc @dtassone for thoughts).

dtassone: How about we adopt the committed suffix as in core. I agree that we should always match propName event and hooks together as it would be easier to maintain.

For the use of `CHANGE` in the prop name, should we keep for controllable prop models?

oliviertassinari: > For the use of CHANGE in the prop name, should we keep it for controllable prop models?

In theory, one can live without the other (feedback vs. order), so no preferences.

m4theushw: @oliviertassinari In your example, the event is not fired during window resizing because we're manually updating the state in `useGridColumns`, instead of calling `setColumnIndex`.

> The "width change" terminology would not describe when the event fire, not its purpose: the event is here to help control the column resizing feature. It wasn't introduced to subscribe to the width change.

It's published by a method called `setColumnWidth`. Makes sense an event named `GRID_COLUMN_WIDTH_CHANGE` to be published when the width is changed.

> Changing this name would also break the event name <-> prop name mapping convention we have. For instance GRID_COLUMN_RESIZE_COMMITTED <-> onColumnResizeCommitted. I don't think that we should allow to change one without the other.

I would open another PR to fix this. https://github.com/mui-org/material-ui-x/pull/1862#discussion_r648767712

oliviertassinari: I see, so we need two events

m4theushw: > Ahhhhh. So is the following correct?

Yes, and `GRID_COLUMN_WIDTH_CHANGE` would call `onColumnWidthChange` (former `onColumnResizeCommitted`).

For this second event, maybe we should use DEBOUNCE instead of COMMITED, like `GRID_DEBOUNCED_RESIZE`. Because when the width is changed the value is already commited to the state. I think most users just want the last width of a column.

oliviertassinari: Ok, but don't we still need to update this PR to introduce `GRID_COLUMN_RESIZE_COMMITTED` back, alongside `GRID_COLUMN_WIDTH_CHANGE`?

m4theushw: I would not introduce `GRID_COLUMN_RESIZE_COMMITTED` in this PR. We first need to fix https://github.com/mui-org/material-ui-x/pull/1862#discussion_r652804358 to call the API instead of updating directly the state. Once `GRID_COLUMN_WIDTH_CHANGE` is called in all occasions we can add a debounced version of it.

oliviertassinari: OK, I don't have the bandwidth to truly understand the problem. I'm resolving this discussion and trusting you will get the best out of the situation.

packages/grid/_modules_/grid/constants/eventsConstants.ts
dtassone: Well I like granular events and those are for internal use but ultimately they are change events.
However the clear and set occurs in 1 rendering so you can't publish change as it's for internal reset of rows to put new ones. Users should only know about the new updated set.

dtassone: If we have doubts, how about we keep those for the next iteration?

oliviertassinari: Looking at the publish:

- When GRID_ROWS_CLEARED is triggered, GRID_ROWS_SET is always triggered.

At the listeners:

- GRID_ROWS_SET is listed by `applySorting`, `applySorting`
- GRID_ROWS_CLEARED is listened by "empty sorting"
- GRID_ROWS_UPDATED is listened by `applySorting`, `applyFilters`

So it seems that we only need one event.

dtassone: GRID_ROWS_CLEARED allows to clear the sortedRows state.

dtassone: the split between GRID_ROWS_UPDATED and GRID_ROWS_SET was more for a user perspective where one would want to subscribe to row updates

m4theushw: `GRID_ROWS_SET` is published by `setRows`. But this method is mostly used in demos that programmatically call the API, so the user that calls the method doesn't need an event and we can remove it. The only use of `setRows` outside demos is in `updateRows`, which also publishes `GRID_ROWS_UPDATED `.

Side note: `setRows` is actually appending rows, renaming to `addRows` might be better.

dtassone: Originally the idea was to do a reset with a clear and a set, to avoid sortedRows error... If we remove the clear, we can remove the set as well and just rename the updated to change.

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -33,9 +33,11 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
           focus: { cell: { id, field }, columnHeader: null },
         };
       });
+      // TODO replace with constant
+      apiRef.current.publishEvent('cellFocusChange');
```
dtassone: reminder, fix constant

oliviertassinari (approved):

PR #1864, opened 2021-06-10
[DataGrid] Persist selection from `selectionModel` when page changes

### LINKED ISSUE #1782: [DataGrid] Prevent the grid from cleaning up the `selectionModel` when `selectionModel` prop is provided

## Current Behavior 😯

## Expected Behavior 🤔

## Steps to Reproduce 🕹

Steps:

## Context 🔦

## Environement

v4.0.0-alpha.30

DanailH:
Fix #1782

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/selection/gridSelectionSelector.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/gridSelectionSelector.ts
@@ -3,15 +3,19 @@ import { gridRowsLookupSelector, GridRowsLookup } from '../rows/gridRowsSelector
 import { GridState } from '../core/gridState';
 import { GridSelectionState } from './gridSelectionState';
 import { GridRowId, GridRowModel } from '../../../models/gridRows';
+import { optionsSelector } from '../../utils/optionsSelector';
+import { GridOptions } from '../../../models/gridOptions';
 
 export const gridSelectionStateSelector = (state: GridState) => state.selection;
 export const selectedGridRowsCountSelector: OutputSelector<
   GridState,
   number,
-  (res: GridSelectionState) => number
-> = createSelector<GridState, GridSelectionState, number>(
+  (res: GridSelectionState, options: GridOptions) => number
+> = createSelector<GridState, GridSelectionState, GridOptions, number>(
   gridSelectionStateSelector,
-  (selection) => Object.keys(selection).length,
+  optionsSelector,
+  (selection, options) =>
+    options.selectionModel ? options.selectionModel.length : Object.keys(selection).length,
 );
 
 export const selectedGridRowsSelector = createSelector<
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -209,7 +209,7 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
       const newSelectionState = { ...state.selection };
       let hasChanged = false;
       Object.keys(newSelectionState).forEach((id: GridRowId) => {
-        if (!rowsLookup[id] && !selectionModel) {
+        if (!rowsLookup[id] || !selectionModel) {
           delete newSelectionState[id];
           hasChanged = true;
         }
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {
+  GRID_PAGE_CHANGE,
   GRID_ROW_CLICK,
   GRID_ROW_SELECTED,
   GRID_SELECTION_CHANGED,
@@ -183,9 +184,16 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
     [disableSelectionOnClick, selectRowModel],
   );
 
+  const handlePageChange = React.useCallback(() => {
+    if (selectionModel) {
+      apiRef.current.setSelectionModel(selectionModel);
+    }
+  }, [apiRef, selectionModel]);
+
   useGridApiEventHandler(apiRef, GRID_ROW_CLICK, handleRowClick);
   useGridApiOptionHandler(apiRef, GRID_ROW_SELECTED, onRowSelected);
   useGridApiOptionHandler(apiRef, GRID_SELECTION_CHANGED, onSelectionModelChange);
+  useGridApiOptionHandler(apiRef, GRID_PAGE_CHANGE, handlePageChange);
 
   // TODO handle Cell Click/range selection?
   const selectionApi: GridSelectionApi = {
@@ -201,7 +209,7 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
       const newSelectionState = { ...state.selection };
       let hasChanged = false;
       Object.keys(newSelectionState).forEach((id: GridRowId) => {
-        if (!rowsLookup[id]) {
+        if (!rowsLookup[id] && !selectionModel) {
           delete newSelectionState[id];
           hasChanged = true;
         }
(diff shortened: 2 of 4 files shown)


packages/grid/_modules_/grid/hooks/features/selection/gridSelectionSelector.ts
```diff
@@ -3,15 +3,19 @@ import { gridRowsLookupSelector, GridRowsLookup } from '../rows/gridRowsSelector
...
   number,
-  (res: GridSelectionState) => number
-> = createSelector<GridState, GridSelectionState, number>(
+  (res: GridSelectionState, options: GridOptions) => number
+> = createSelector<GridState, GridSelectionState, GridOptions, number>(
   gridSelectionStateSelector,
-  (selection) => Object.keys(selection).length,
+  optionsSelector,
+  (selection, options) =>
+    options.selectionModel ? options.selectionModel.length : Object.keys(selection).length,
```
dtassone, line 18: you should not use the `options.selectionModel` here as the grid read the state. So the state is the source of truth and should reflect what is in options.

dtassone: Yes so if you set the model it updates the `state.selection`, then if you select a row, the state will be updated. Therefore, the source of truth is `state.selection`

m4theushw (commented):
Could we have a test case where it begins with a `selectionModel`, goes to another page, selects another row and checks that `onSelectionModelChange` was called with the initial selected ids and also the new selected id after changing page? It would be awesome to have it to prevent a regression in the future.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -201,7 +209,7 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
       const newSelectionState = { ...state.selection };
```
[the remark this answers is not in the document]
DanailH: Updated

DanailH:
I've updated the PR and made the solution work. We discussed it yesterday with @dtassone and for now, I'll keep this PR open as a reference for this one #1823 .

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -198,10 +211,14 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
 
   React.useEffect(() => {
     setGridState((state) => {
+      if (selectionModel) {
+        return state;
+      }
```
oliviertassinari, line 216: The two changes in this useEffect looks KO. The behavior should be controlled **invariant**. Meaning, if the uncontrolled selection unselects when the `rows` changes, then the controlled version of the selection should behave identically (unselect).

oliviertassinari: Here, we stop unselecting if controlled 💥

packages/grid/_modules_/grid/hooks/features/selection/gridSelectionSelector.ts
oliviertassinari: I can understand why we would want to keep the selector state only, we have never used the options before, it can feel scary, a new pattern. :+1: for trying a solution without.

For the prop updating the state or not. In https://github.com/mui-org/material-ui/blob/2c4dc994da164a53b253f03bcd2425a52655f175/packages/material-ui-utils/src/useControlled.js#L8 we skip the state. However, here, we have a double notion of prop controlled and state controlled, with the state with more priority. So we probably want to update the state.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -94,8 +98,16 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
         data: row,
         isSelected: selectionState[id] !== undefined,
       };
+
+      let newSelectionModel = Object.values(selectionState);
+      if (selectionModel) {
+        newSelectionModel = selectionModel.includes(rowModelParams.id)
+          ? selectionModel.filter((selection) => selection !== rowModelParams.id)
+          : [...selectionModel, rowModelParams.id];
+      }
```
oliviertassinari, line 107: Wait, are we implementing the logic that is inside the above `setGridState` twice?

oliviertassinari (changes requested):

DanailH:
https://github.com/mui-org/material-ui-x/pull/1864#discussion_r652277004

The thing is that this selector is only used to display the `Row count` in the footer and the `Select All` checkbox.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
DanailH: No, it looks like it but it is much simpler. That one only appends the newly selected row id to the already existing selection model so that the value of the params that the developer will get from the `onSelectionModelChange` is correct. There is no state or options update.

oliviertassinari: I don't understand, want what about multi-selection? It seems that it should be exactly the same, so enforced with a shared abstraction

dtassone: We use an object for the selection state to avoid looping through an array to know if a row is selected, otherwise, if you have 1000k rows we would have a perf hit. 
https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/components/GridViewport.tsx#L80

However, for the prop, using an object is an option but it seemed to make less sense, and it is more convenient to use an array.

oliviertassinari: @dtassone The caching makes sense, it definitely seems valid. What do you think about we stored the controlled prop as is, in the store, and store a private cache, a dedicated state, in another part of the state (one for instance that we wouldn't expose for developers than want to store the state in their data base). But more importantly, have updates on the `state.selectionModal` being the source of truth, and triggering changes automatically, to its cache, say `state.cache.selectionModal`?

dtassone: Well the controlled prop to me is already stored and is the value of the prop itself. Why would we need to duplicate the values? 
And the private cache is actually the grid state and it is the source of truth. 
We could do a selector that transforms the state to make it more storable.

oliviertassinari: > Why would we need to duplicate the values?

To fix the first comment of this thread.

DanailH: Actually, I was looking in that use case yesterday and what I struggle to understand is how would you differentiate multiselection using `CTRL` from the normal one? The only way I see it is that if it is not multi-selection then basically every time the `onSelectionModelChange` is called we need to always pass just one ID (the lasted one)?

Then for the muti-selection and checkbox selection, you would append/remove IDs based on what is being selected/deselected.

dtassone: Considering #1823 and the equality issue and the discussion here, how about we transform state to be optimised in a selector. As they are memoised it will only do the transformation when the underlying model change. That way it would allow us to store the model in state, fix the equality issue #1823 and avoid the model duplication.
The cache mentioned by @oliviertassinari would actually be provided by the selector memoization. What do you think?

oliviertassinari: @dtassone This could work if we also expose the raw state. We need the raw state to perform the control feedback loop.

DanailH:
Just to clarify because I'm a bit lost from all the discussions - should I stop the effort on this PR or fix only the problem where what is emitted from the `onSortModel` change is not correct? Basically not fixing the underlying problem with controlled/uncontrolled state but only the event?

oliviertassinari:
@DanailH I think that we should fix the `onSelectionModelChange` event. This event should fire when the selectionModel is emptied. No matter what happens next, we know that this behavior should be in, and is very unlikely to go away. So it's safe, unrelated to another other effort, and small. I think perfect for a first step.

DanailH:
I'm closing this PR in favour of this one -> https://github.com/mui-org/material-ui-x/compare/master...DanailH:bug/DataGrid-1782-call-onSelectionModelChange-when-selection-state-changes?expand=1

The listed PR has the fix mentioned here https://github.com/mui-org/material-ui-x/pull/1864#issuecomment-867001003

PR #1871, opened 2021-06-11
[XGrid] Fix calculation of the scroll position in `scrollToIndexes`

### LINKED ISSUE #1818: [X-GRID] ScrollToIndexes is inconsistently scrolling rows into view

## Current Behavior 😯

When selecting a row at random then scrolling to that row anytime the `selectedRow` state changes, the `scrollToIndexes` method inconsistently scrolls (more like jumps) the row into view. If you look at my example and press the 'select a random row and scroll to it' button, you'll see it occasionally misses the row entirely and I need to scroll either up or down. 
What's interesting is it seems to produce different results with a smaller data set. I have tried replacing my ID keys in the assumption that I had shared IDs which gave me a the wrong `index`, but this didn't make a difference.

I'm using `useGridAPIRef()` to access the following methods to achieve my goal:
- `getRowIndex()` to get the row index of the row I wish to scroll to
- `getColumnindex()` as above but for columns. I am not interested in horiztonal scrolling but it seems to be a required param for the following
- `scrollToIndexes()` which I understand takes a object parameter in this shape: `{ rowIndex: 1, colIndex: 1 }` src: https://github.com/mui-org/material-ui-x/blob/726235cfc1510e62a2bce84c6a2b7b9ea8433589/packages/grid/_modules_/grid/models/gridCell.ts#L9
- `scroll()` to manually scroll (since `scrollToIndexes` doesn't seem to work).

I'm maybe misunderstanding how to properly scroll to any given row in the table but I was unable to find any documentation demonstrating how to do it so I had to do a bit of a deep dive into the source for `scrollToIndexes`.

## Expected Behavior 🤔

`scrollToIndexes()` should scroll to the specified row in the table everytime, regardless of previous scroll position. The row should then be visible in the table without the user needing to scroll the table at all.

## Steps to Reproduce 🕹

Link: https://codesandbox.io/s/material-uix-gridscroll-to-indexes-demo-simple-bzphc

Steps:

1. select the 'enable scrollToIndexes()` button. this will use the out of the box `scrollToIndexes()` functionality
2. Select a row at random by clicking the 'Select a random row' button in the example 
3. the row selected will either be visible in the table without the user scrolling, or it won't.
4. the row selected will never be at the same position (not always in the middle/top/bottom) - seemingly random enough

I have attempted to fix the problem myself by creating a similar scroll function (called `custom`). It works by taking `n rows` and mulitpliying it by `rowHieght`. Example: 
- I want to select row #200
- assume I'm at the top of the table (no scroll)
- 20 * pixel hieght of each row (30) = 20 * 30 = 600 px to scroll
- remove some abitrary value (because the calculation above doesn't seem to be 100%)
- `scrollPositionFromTop -= 50;`
- use `scroll({ top: scrollPositionFromTop, left: 0 });`

## Context 🔦

I'm trying to programatically select a row (from a separate component) and scroll the row into view to save the user from scrolling hundreds of rows to find the one they're making changes to. I have a `useEffect` which listens for the selected row ID changing (which is essentially a pointer to the row in the table for this example). When the ID changes,`scrollToIndexes` fires off and should scroll the row into view for the user to see. I can select a row fine, but it's the scrolling part that is proving difficult to consistently achieve.

## Your Environment 🌎

<details>
  
```
Using Chrome both locally and in my example code.

```
 System:
    OS: Linux 5.11 Fedora 33 (Workstation Edition) 33 (Workstation Edition)
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 1.22 GB / 15.26 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.16.0 - /usr/bin/node
    npm: 7.13.0 - /usr/local/bin/npm
  Managers:
    Composer: 2.0.12 - /usr/bin/composer
    pip3: 20.2.2 - /usr/bin/pip3
    RubyGems: 3.1.4 - /usr/bin/gem
    Yum: 4.6.1 - /usr/bin/yum
  Utilities:
    Make: 4.3 - /usr/bin/make
    GCC: 10.2.1 - /usr/lib64/ccache/gcc
    Git: 2.30.2 - /usr/bin/git
    Clang: 11.0.0 (Fedora - /usr/lib64/ccache/clang
    Mercurial: 5.4.2 - /usr/bin/hg
    Subversion: 1.14.1 - /usr/bin/svn
  Servers:
    Apache: 2.4.46 - /usr/sbin/apachectl
  Virtualization:
    Docker: 20.10.6 - /usr/bin/docker
  IDEs:
    Nano: 5.3 - /usr/bin/nano
    VSCode: 1.55.1 - /usr/bin/code
    Vim: 8.2 - /usr/bin/vim
  Languages:
    Bash: 5.0.17 - /usr/bin/bash
    Go: 1.15.8 - /usr/bin/go
    Perl: 5.32.1 - /usr/bin/perl
    PHP: 7.4.16 - /usr/bin/php
    Python: 3.9.2 - /usr/bin/python
    Python3: 3.9.2 - /usr/bin/python3
    Ruby: 2.7.2 - /usr/bin/ruby
  Databases:
    MySQL: undefined - /usr/bin/mysql
    SQLite: 3.34.1 - /usr/bin/sqlite3
  Browsers:
    Chrome: 89.0.4389.114
    Firefox: 87.0
</details>

## Order id 💳

I don't have an Order ID.

m4theushw:
Fixes #1818
Preview: https://codesandbox.io/s/material-uix-gridscroll-to-indexes-demo-simple-forked-0oi0o?file=/src/App.js

This is an attempt to fix the related bug. It took me some good time to understand how the virtualization works. :)

From my investigation, I found that the amount of scroll needed to make a row visible is less than what I would normally expect (rowIndex * rowHeight). Take as example the top position of the row at the index 5 below. In theory, its position is 250px but I had to scroll just 200px to see it. The same occurs with `scrollToIndexes`. It was using the viewport height, which might include space for partially visible rows. Now it's using the same height that triggers the page change in the virtualization.

Mfl74Wvk3z


DanailH:
Not sure if it ready for review but I can see 3 main problems:
 1. When navigating with the keyboard the grids start to scroll immediately, not when the last visible is focused and you want to move to the next one that is not visible,
 2. When you scroll the grid the normal way and then select a row there is a "jump" that positions the cell you selected to the top edge of the viewport.
 3. The scroll to bottom using the keyboard shortcut seems to not working.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -222,32 +220,23 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
...
-        if (isRowIndexAbove) {
-          scrollCoordinates.top = scrollPosition; // We put it at the top of the page
-          logger.debug(`Row is above, setting top to ${scrollCoordinates.top}`);
-        } else if (isRowIndexBelow) {
-          // We make sure the row is not half visible
-          scrollCoordinates.top = scrollPosition - viewportHeight + rowHeight;
-          logger.debug(`Row is below, setting top to ${scrollCoordinates.top}`);
+        const scrollPosition = currentRowPage * gridState.containerSizes!.renderingZoneScrollHeight;
+
+        if (windowRef.current!.scrollTop !== scrollPosition) {
```
dtassone, line 233: you just want to scroll when you can see the cell as @DanailH mentioned

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -233,10 +233,14 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
...
-        const pageHeight = gridState.containerSizes!.viewportPageSize * rowHeight;
+
+        // The amount of space taken by the first partially visible row
+        const partialRowHeight = windowRef.current!.scrollTop % rowHeight;
 
         isRowIndexAbove = windowRef.current!.scrollTop > scrollPosition;
-        isRowIndexBelow = scrollPosition + rowHeight > windowRef.current!.scrollTop + pageHeight;
+        isRowIndexBelow =
+          scrollPosition + rowHeight >
+          windowRef.current!.scrollTop + viewportHeight + partialRowHeight;
```
m4theushw, line 243: Fixes a bug where the grid wasn't scrolling if the row was partially visible.

BPYpligb7o

m4theushw:
> When you scroll the grid the normal way and then select a row there is a "jump" that positions the cell you selected to the top edge of the viewport.

@DanailH Yeah, this was in the original CodeSandbox provided by the user. I removed it now.

> The scroll to the bottom using the keyboard shortcut seems to not working.

Fixed. I think it's ready for review now.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -220,23 +222,33 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
...
           : params.rowIndex - paginationState.page * paginationState.pageSize;
 
         const currentRowPage = elementIndex / gridState.containerSizes!.viewportPageSize;
-        const scrollPosition = currentRowPage * gridState.containerSizes!.renderingZoneScrollHeight;
-
-        if (windowRef.current!.scrollTop !== scrollPosition) {
-          logger.debug(`Setting top to ${scrollCoordinates.top}`);
-          scrollCoordinates.top = scrollPosition;
-          needScroll = true;
+        const viewportHeight = gridState.containerSizes!.renderingZoneScrollHeight;
```
dtassone, line 234: it's a misleading var name I think

m4theushw: True, I changed to `height`. Probably I'll revisit this hook to fix the jump, then we can discuss that again.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -225,24 +225,28 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
...
           : params.rowIndex - paginationState.page * paginationState.pageSize;
 
         const currentRowPage = elementIndex / gridState.containerSizes!.viewportPageSize;
-        const scrollPosition = currentRowPage * gridState!.viewportSizes.height;
-        const viewportHeight = gridState.viewportSizes.height;
+        const height = gridState.containerSizes!.renderingZoneScrollHeight;
+        const scrollPosition = currentRowPage * height;
+
+        // The amount of space taken by the first partially visible row
+        const partialRowHeight = windowRef.current!.scrollTop % rowHeight;
```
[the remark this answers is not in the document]
oliviertassinari: See the logic that we copied from WAI, https://github.com/mui-org/material-ui/blob/2c4dc994da164a53b253f03bcd2425a52655f175/packages/material-ui/src/useAutocomplete/useAutocomplete.js#L328-L348. It's working smoothly https://next.material-ui.com/components/autocomplete/ and is exactly the same case.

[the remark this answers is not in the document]
oliviertassinari: Wouldn't we get this problem solved if we get the root cause of #1911?

oliviertassinari (commented):
I don't think that we should fix scrollToIndexes, it seems to behave correctly. What I think we should double down on is: https://github.com/mui-org/material-ui-x/issues/1911#issuecomment-861914182

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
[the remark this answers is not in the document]
m4theushw: > Wouldn't we get this problem solved if we get the root cause of #1911?

Yes, but #1911 will require a lot of work. We could first push this one to solve the pain of the user that opened #1818.

Does this diagram help to understand the logic?

scroll position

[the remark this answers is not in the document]
oliviertassinari: > Does this diagram help to understand the logic?

I meant that the current logic on HAD already take care of this.

> We could first push this one to solve the pain of the user that opened #1818.

From what I understand, this solution would only work if we reverse the wrong behavior of #1911.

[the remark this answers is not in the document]
oliviertassinari: > I meant that the current logic on HEAD already takes care of this.

Ah, no, right, my bad. We forgot to account for the height of the row 😆 

```diff
diff --git a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
index 10878a5a..a19b4626 100644
--- a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
@@ -235,7 +235,7 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
         const viewportHeight = gridState.viewportSizes.height;

         isRowIndexAbove = windowRef.current!.scrollTop > scrollPosition;
-        isRowIndexBelow = windowRef.current!.scrollTop + viewportHeight < scrollPosition;
+        isRowIndexBelow = windowRef.current!.scrollTop + viewportHeight < scrollPosition + rowHeight;

         if (isRowIndexAbove) {
           scrollCoordinates.top = scrollPosition; // We put it at the top of the page
```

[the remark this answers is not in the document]
oliviertassinari: > But something is still off 😆

Found it, `viewportHeight` value is wrong too.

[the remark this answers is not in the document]
m4theushw: But we don't use `viewportHeight` here. We use `renderingZoneScrollHeight` for virtualization, which I think it's not right value.

oliviertassinari:
@m4theushw Is there something left to do on this PR?

m4theushw:
@oliviertassinari I've simplified the logic. We can push this one or wait for the virtualization logic to be redone.

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -449,6 +449,35 @@ describe('<XGrid /> - Rows', () => {
...
+          />,
+        );
+        const gridWindow = document.querySelector('.MuiDataGrid-window')!;
+        // Scroll to the rowIndex=5, which is the 2nd row in page 1
+        gridWindow.scrollTop = renderingZoneHeight + rowHeight;
+        gridWindow.dispatchEvent(new Event('scroll'));
+        // Scroll to rowIndex=4, which is the 1st row in page 1
+        apiRef.current.scrollToIndexes({ rowIndex: 4, colIndex: 0 });
+        gridWindow.dispatchEvent(new Event('scroll'));
+        expect(gridWindow.scrollTop).to.equal(renderingZoneHeight);
```
oliviertassinari, line 478: This doesn't look right. AFAIK, it should be
```suggestion
        expect(gridWindow.scrollTop).to.equal(4 * rowHeight);
```

We display the 5th row, so we need to scroll 4-row heights. Anything else would push us toward changing the scroll logic when it's the rendering logic that is wrong.

oliviertassinari (changes requested):

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
m4theushw: #1969 doesn't solve the bug when the given `rowIndex` is way before the first visible row. This test is lying.

oliviertassinari: Which bug are we talking about? If we are talking about #1818, I think that it should be put on hold until #1911 is fixed.

m4theushw: You mentioned that this PR passes with the logic from #1969. From what I understood that means that we should only merge the test case here. I'm saying that the bug of #1818 remains and the test case is not right. I'll close this one and investigate how to address #1911.

oliviertassinari: - :+1: for an extra test case. 
- #1911 I would expect the solution not to be too complex. I think that we should first look into why the scrollTop is too small compared to what it should be, then once fixed, look into why the rendered rows are not continuous. It's likely a small refactoring.

oliviertassinari: @m4theushw Coming back to a discussion we had. Do you agree that this assertion would be correct? https://github.com/mui-org/material-ui-x/pull/1871/files#r660182225. If you do, I think that we could merge a new test case. If you don't, we can chat more about it.

Either way, what's the intended outcome of keeping the PR open for a couple of weeks? We could restart the effort from a blank branch.

m4theushw: I do agree that https://github.com/mui-org/material-ui-x/pull/1871/files#r660182225 is the correct assertion, but it doesn't work this way. The virtualization changes from page 0 to 1 before that (at the "jump"). If we merge this one it will be too tied to the virtualization implementation. Better to close it and to fix the root cause.

oliviertassinari: I was going to say that the value of this test case is to increase our test coverage, but I feel that it duplicate with https://github.com/mui-org/material-ui-x/blob/4066078512087318ec22e2b80763564d67b0a22e/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx#L480

So, after all, I don't see any value in this PR 😁, outside of the discussion it has created.

m4theushw: I'm testing it after rebasing with master. It seems to pass with the proposed assertion.

m4theushw: Yeah, it passes but visually it's wrong.

https://user-images.githubusercontent.com/42154031/124522376-6eae1700-ddc9-11eb-8014-16c907d76211.mp4

LorenGr:
I am seeing that this topic was closed. But the bug is still present. Is there a plan to fix this please?
I made a demo here : https://codesandbox.io/s/datagridprodemo-material-demo-forked-xhl3m?file=/demo.js:660-675
the scrolling looses synch with the selected row when you scroll down a lot

m4theushw:
@LorenGr The calculation is right, but the virtualization is rendering the wrong rows. You can subscribe to #2673 to get notified when the new virtualization engine gets released.

PR #990, opened 2021-02-06
[DataGrid] Resolve the api ref at the same time as any other ref

oliviertassinari:
Effectively, it means that we can stop documenting `useApiRef`, and replace it completely in the public API with:

```jsx
function MyApp() {
  const apiRef = React.useRef();

  return <DataGrid apiRef={apiRef} />
}
```

It's a logical follow-up with the direction taken in #933. More details in https://github.com/mui-org/material-ui-x/pull/939#discussion_r567235523. I like the approach, at least, it's sounds and consistent with the rest of the codebase. You can find the same API on, the Tabs for instance (`action` prop).

PR #1877, opened 2021-06-12
[DataGrid] Add useGridApiContext to access the GridApiContext

### LINKED ISSUE #1401: [DataGrid] Should fail with an explicit error when context is missing

## Current Behavior 😯

When the context provider of the data grid is missing, it fails with an obscure message.

## Expected Behavior 🤔

It should fail with a clear error message.

## Steps to Reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/data-grid-bug-loading-empty-row-forked-x20om?file=/src/App.js

Screenshot 2021-04-13 at 00 23 26

Note that it won't reproduce in our local env as we share the same react context instance. Use https://codesandbox.io/s/data-grid-bug-loading-empty-row-forked-2i56s?file=/src/App.js instead.

```jsx
import { GridOverlay } from "@material-ui/data-grid";

export default function App() {
  return (
    <div className="App" style={{ height: "500px" }}>
      <GridOverlay />
    </div>
  );
}
```

## Context 🔦

This is a follow-up on #1326 where we didn't leverage the feedback as much as we could. The first time I saw this problem was in https://github.com/mui-org/material-ui/issues/686.

## Your Environment 🌎

v4.0.0-alpha.24

m4theushw:
Closes #1401


packages/grid/_modules_/grid/hooks/root/useGridApiContext.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+import { GridApiContext } from '../../components/GridApiContext';
+
+export function useGridApiContext() {
+  const apiRef = React.useContext(GridApiContext);
+
+  if (apiRef === undefined) {
```
oliviertassinari, line 7: What do you think about raising a warning instead of throwing and only in dev mode to not have any bundle size impact? Is it important enough to throw?

dtassone: I think it is important to throw as users would have issues at a later stage.

We could also consider combining it with useGridApiRef so you have a single point to get or create the apiRef 🤔

packages/grid/data-grid/src/tests/components.DataGrid.test.tsx
```diff
@@ -45,4 +48,25 @@ describe('<DataGrid /> - Components', () => {
       expect(document.querySelectorAll('.customFooter').length).to.equal(0);
     });
   });
+
+  it('should throw if a component is used without proving the context', function test() {
```
dtassone, line 52: proving?

DanailH (approved):

packages/grid/_modules_/grid/hooks/root/useGridApiContext.ts
m4theushw: > What do you think about raising a warning instead of throwing and only in dev mode to not have any bundle size impact?

@oliviertassinari  I tested with the docs and the warning wasn't sent to the console, but to the terminal that was running the dev server. A throw in this situation facilitates for the user to know where's the problem.

> We could also consider combining it with useGridApiRef so you have a single point to get or create the apiRef 🤔

@dtassone Let's imagine that there's another grid inside the grid, like a child row. How will we know if a apiRef has to be created or the one in the context should used? Maybe we need to add an argument to force to create the apiRef even if there's a context in the tree.

oliviertassinari: > but to the terminal that was running the dev server.

@m4theushw True, I was actually just talking about this problem in https://github.com/mui-org/material-ui/issues/18119#issuecomment-860877215. It's a Next.js thing only, that they could solve by forwarding the console log too (not only the error).

> How will we know if a apiRef has to be created or the one in the context should used?

💯. I would even push it one step further is unbundle useGridApiRef. I think that we need 3 hooks not 1.

- `useGridApiContext` to consume the API ref
- `const apiRef = React.useRef<GridApi>(createGridApi());` to create the apiRef and store it.
- `useGridApiRef`to read the API ref from the child data grid **ONLY**

I did a first step in this direction with #990

Issue #605, opened 2020-11-19
[DataGrid] Allow "select all" to only select the active page


NickLont:
When Pagination is enabled in XGrid when the "Select all" checkbox is enabled, all possible options from all pages are selected. I would like the option to select only the current paginated page options (exactly like [that](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/selection/#with-paging))
 

- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

When the pagination number per page is set to 5, when the "select all" checkbox is clicked only the current 5 options should be selected and not every possible option.
## Examples 🌈

Example from the DevExtremeREACTIVE page [here](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/selection/#with-paging)

## Motivation 🔦

Building an application with a table of contacts, the use cases where a bulk action would need to affect every single contact is rare, compared to selecting only the ones currently paginated.

oliviertassinari:
@NickLont Thanks for the issue. It seems that both approaches have their use cases. From my perspective, the default should be to select only the visible rows. Most end-users will be confused otherwise.

I like how Crowdin display a message to suggest the selection of all the rows as soon as the whole page is selected:

<img width="1189" alt="Capture d’écran 2020-11-19 à 15 56 30" src="https://user-images.githubusercontent.com/3165635/99682761-e79d7000-2a7f-11eb-9d63-0fe5d051e40f.png">

Outside of DevExpress, we can also use react-table to benchmark the difference: https://react-table.tanstack.com/docs/api/useRowSelect#example. `getToggleAllPageRowsSelectedProps()` is for the current page while `getToggleAllRowsSelectedProps()` is for everything.

oliviertassinari:
Gmail

<img width="1013" alt="Capture d’écran 2020-11-23 à 15 03 27" src="https://user-images.githubusercontent.com/3165635/99971165-1b88d600-2d9d-11eb-9671-5dbca048c54b.png">

oliviertassinari:
For implementing the "select all" behavior when only the current page is selected, we might need #246.

dtassone:
Benchmark for Ag grid 
it has 2 modes
https://www.ag-grid.com/react-grid/row-selection/#select-everything-or-just-filtered

PR #1879, opened 2021-06-12
[DataGrid] `Select all` checkbox click should select only filtered rows after filter

### LINKED ISSUE #1141: [DataGrid] Select All feature does not work on column filtering
- [X] The issue is present in the latest release.
- [X] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Select All feature accurately selects active rows when only 1 filter is active. If a subsequent filter is added, Select All feature breaks.

## Expected Behavior 🤔

Select All feature should update total number of active rows when additional filters are added. 

## Steps to Reproduce 🕹

Steps:

1. Add a single column filter to X Grid table https://material-ui.com/components/data-grid/#commercial-version
2. Select All checkbox on top left
3. Add a second column filter.
4. Uncheck and check Select All checkbox again.

## Your Environment 🌎

System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 12.18.4 - ~\PMT-App\node_modules\.bin\node.CMD
    Yarn: Not Found
    npm: 6.14.10 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 88.0.4324.190
    Edge: Spartan (44.18362.449.0)
  npmPackages:
    @material-ui/core: 4.11.0 => 4.11.0
    @material-ui/icons: 4.11.2 => 4.11.2
    @material-ui/lab: 4.0.0-alpha.56 => 4.0.0-alpha.56
    @material-ui/pickers:  3.2.10
    @material-ui/styles:  4.11.3
    @material-ui/system:  4.11.3
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @material-ui/x-grid: 4.0.0-alpha.20 => 4.0.0-alpha.20
    @material-ui/x-license:  4.0.0-alpha.20
    @types/react:  17.0.0
    react: 16.13.1 => 16.13.1
    react-dom: 16.13.1 => 16.13.1

## Order id 💳

n/a

### LINKED ISSUE #1863: [DataGrid] Filter and check all Checkbox selects all items

## Current Behavior 😯
When I filter the items in the grid and use the select all checkbox it selects the items thats filtered out too. It seems to work correctly if the filter only return a couple of items. 

## Expected Behavior 🤔

selectionModel should only have items from the filtered items when I use the select all checkbox.

## Steps to Reproduce 🕹

```javascript
import { DataGrid } from "@material-ui/data-grid";
import { useDemoData } from "@material-ui/x-grid-data-generator";

export default function DisableClickSelectionGrid() {
  const { data } = useDemoData({
    dataSet: "Commodity",
    rowLength: 300,
    maxColumns: 6
  });

  return (
    <div style={{ height: 400, width: "100%" }}>
      <DataGrid
        checkboxSelection
        disableSelectionOnClick
        {...data}
        onSelectionModelChange={(selection) => {
          console.log("selection", selection);
        }}
      />
    </div>
  );
}
```
[https://codesandbox.io/s/material-demo-forked-3d6n8?file=/demo.tsx](https://codesandbox.io/s/material-demo-forked-3d6n8?file=/demo.tsx)

Steps:

1. Click the Desk colum filter, filter on "D-1"
2. Check the select all checkbox
3. The console will log 300, expected 31

## Your Environment 🌎

<details>
  
```
```
</details>

## Order id 💳
22396

ZeeshanTamboli:
Fix point 3. of https://github.com/mui-org/material-ui-x/issues/1141#issuecomment-850977514
Closes #1863
Preview: https://deploy-preview-1879--material-ui-x.netlify.app/components/data-grid/


packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -239,4 +242,10 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
     });
     forceUpdate();
   }, [apiRef, setGridState, forceUpdate, isRowSelectable]);
+
+  React.useEffect(() => {
+    if (totalSelectedRows > Object.keys(visibleRowIds).length) {
```
dtassone, line 247: you can use `visibleGridRowCountSelector`
Also not sure about the logic here. What if we change rows to another set with the same length?

ZeeshanTamboli: > Also not sure about the logic here. What if we change rows to another set with the same length?

So when filtering, is it possible that the rows filtered (visible) are equal to the total rows? Also this logic only runs when the rows are _selected_ and then the filtering is applied, so the visible rows are less after filtering. Would there be a case when the rows are of same length after filtering? I don't think so.

ZeeshanTamboli: > you can use visibleGridRowCountSelector

I need to pass `visibleRowIds` below to `selectRows`. Should I still select `visibleRowCount` for count to replace `Object.keys(visibleRowIds).length`? I think we should as the selector will be memoized.

m4theushw: > So when filtering, is it possible that the rows filtered (visible) are equal to the total rows?

Yes, if all rows satisfy the filter than all rows will be visible.

---

You can't just check if there're more selected rows than visible rows. The behavior becomes inconsistent if the size of the selection is smaller than the number of visible rows.

I think you need to do the same thing we do when the `rows` prop is changed, but with `visibleRowIds`.

https://github.com/mui-org/material-ui-x/blob/a2a1f8eca0f8b188a1ed40abbcc616dc3a837d76/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L203-L208

@DanailH was working in https://github.com/mui-org/material-ui-x/pull/1864 to fix a bug around this part of the code too.

ZeeshanTamboli: > So when filtering, is it possible that the rows filtered (visible) are equal to the total rows?
> > Yes, if all rows satisfy the filter then all rows will be visible.

Okay, so then the point is that this condition won't satisfy, right? And it would not select rows by deselecting others and selecting visible ones. It would work as it works now in master.

> The behavior becomes inconsistent if the size of the selection is smaller than the number of visible rows.

Again, then this effect's condition won't satisfy and selecting rows won't run.

------

**What this effect does?**

I will like to give an example: 
For grid with 200 rows.
The total selected rows are all 200, after filter the visible rows become 40. The dependencies change, this effect's condition satisfies and now the select rows will deselect others and select only 40. If you add another filter, the rows are now 37, it will select only 37. This is main issue of #1141.

-----

> I think you need to do the same thing we do when the `rows` prop is changed, but with `visibleRowIds`.
>
> https://github.com/mui-org/material-ui-x/blob/a2a1f8eca0f8b188a1ed40abbcc616dc3a837d76/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L203-L208
>
> @DanailH was working in https://github.com/mui-org/material-ui-x/pull/1864 to fix a bug around this part of the code too. 

Does this effect #1141? Can we focus on the main issue and the related comments issues I linked in description?

Also I would suggest once to look at whats happening currently and after the fix for the issues attached in the description.

ZeeshanTamboli: > you can use visibleGridRowCountSelector

@dtassone I used this selector now.

m4theushw: > Again, then this effect's condition won't satisfy and selecting rows won't run.

Yeah, but you might have one selected row, adds another filter where this selected row doesn't satisfy it and the previous row will still selected. If you select all, that means having a larger selection than visible rows, it unselects those rows that are not more visible. That's the inconsistency I noticed.

> Does this effect #1141? 

No

> Is this fix you are suggesting for the related issue of #1141?

Yes, I would keep it simple. The selection should always contain only visible rows. In the future, we can add an option to allow to accumulate selection even when rows become invisible.

dtassone (approved):

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -239,4 +246,10 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
     });
     forceUpdate();
   }, [apiRef, setGridState, forceUpdate, isRowSelectable]);
+
+  React.useEffect(() => {
+    if (totalSelectedRows > visibleRowCount) {
+      selectRows(visibleRowIds, true, true);
+    }
+  }, [totalSelectedRows, visibleRowIds, selectRows, visibleRowCount]);
```
dtassone, line 254: Really not sure about this useEffect. 
If any of the dependency changes then it will select the rows.

dtassone: better to use apiRef for the api call as it avoids adding the selectRows as a dependency

ZeeshanTamboli: > better to use apiRef for the api call as it avoids adding the selectRows as a dependency

Then apiRef will be a dependency.

dtassone: yes but that's fine as it's a ref it doesn't change like the function itself

ZeeshanTamboli: Okay. Added `apiRef`. I was wondering if refs don't change why we need it as a dependency and why does hooks eslint complains. Usually we don't add refs as dependency.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
ZeeshanTamboli: > Yeah, but you might have one selected row, adds another filter where this selected row doesn't satisfy it and the previous row will still be selected.

Yeah, even I see this behaviour. 

> The selection should always contain only visible rows

Are you referring with pagination, that it should select only the current page visible rows? #605?

oliviertassinari (changes requested):
Are we should about the test coverage? I mean, we add one new test that is passing on HEAD. And a second test is weird. Why does it apply the filter twice? What about the second bug #1141 shouldn't we test it too?

oliviertassinari:
I have opened #1919 to prove my point. The test case is not testing the issue, and we can break this down into smaller chunks: #1919 is one of them.

ZeeshanTamboli:
@oliviertassinari 
> Are we sure about the test coverage? I mean, we add one new test that is passing on HEAD. 

My bad. The test was wrong. The test with the below change if updated will be correct. Pretty much what you did in #1919. This fails in master HEAD. I was setting the filter in initial mounting itself rather than setting it after the render. This leans me to now follow more TDD approach. Thanks.

```diff
diff --git a/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx b/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
index ac18ad6b..3bc07b97 100644
--- a/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
@@ -129,20 +129,16 @@ describe('<XGrid /> - Selection', () => {
   });

   it('should select only filtered rows after filter is applied', () => {
-    render(
-      <Test
-        checkboxSelection
-        filterModel={{
-          items: [
-            {
-              columnField: 'brand',
-              operatorValue: 'contains',
-              value: 'a',
-            },
-          ],
-        }}
-      />,
-    );
+    render(<Test checkboxSelection />);
+    apiRef.current.setFilterModel({
+      items: [
+        {
+          columnField: 'brand',
+          operatorValue: 'contains',
+          value: 'a',
+        },
+      ],
+    });
     const selectAll = screen.getByRole('checkbox', {
       name: /select all rows checkbox/i,
     });
```


> And a second test is weird. Why does it apply the filter twice?

It is testing the main issue of #1141 
```
Current Behavior 😯
Select All feature accurately selects active rows when only 1 filter is active. If a subsequent filter is added, Select All feature breaks.

Expected Behavior 🤔
Select All feature should update total number of active rows when additional filters are added.
```
`If a subsequent filter is added, Select All feature breaks.`
This the reason I added the second filter but not same, it is a different filter constraint. But here again I will need to use ` apiRef.current.setFilterModel`. 
Does it make sense? 🤔 
Actually this statement by user: `Select All feature accurately selects active rows when only 1 filter is active` is invalid now after April release. Select All breaks even after one filter. The issue was reported in March. So this is DataGrid related as well.

> What about the second bug #1141 shouldn't we test it too (checkbox)?

Yes we should which I missed. I  I think it is `Select All` initially  after filter (intermediate state) and then again `Select All` to check that the intermediate state is gone and none of the rows are selected?

The test adequacy is not good. I can rewrite them and add the missed one (checkbox).

> we can break this down into smaller chunks: #1919 is one of them. 

Yes. All seemed related to me with `Select All` so I included them in one PR.

oliviertassinari:
@ZeeshanTamboli Ahhhhh. Now I get why you added the second filter in the test! Be very careful with what developers write as expected vs. actually. In more than 50% of the time, we overrule them. What we care about hearing is the pain point.

So basically, if all the rows are selected without filters, and we add a filter, then, only the new visible rows are selected. Personally, this behavior has never crossed my mind. I didn't cover it in https://github.com/mui-org/material-ui-x/issues/1141#issuecomment-850977514. I would personally vote against it because it prevents the end-users to fine-tune their selection. But no strong point of view. cc @mui-org/x

At the very list, we should break down this PR in 3:

- memo issue, this is a standalone bug. I have it in #1919. I recommend we merge this one first
- checkbox issue, this is a standalone bug. We have a test case for it in #1919 already, and the fix in this PR (likely but I didn't test it).
- your new behavior this is a standalone discussion. We continue the discussion, see if we want it or not.

dtassone:
> if all the rows are selected without filters, and we add a filter, then, only the new visible rows are selected.

I would vote against it in the native behavior, as ppl select a row then apply a filter to quickly find their next pick and select it. So I expect that they would want the selection to be aggregated

ZeeshanTamboli:
@oliviertassinari 
> It seems that we should break down this PR in smaller chunks. It's overloaded.
> 
> 1. checkbox issue, this is a standalone bug. I have it in #1919. I recommend we merge this one first

Sounds good.

> 2. memo issue, this is a standalone bug. We have a test case for it in #1919 already, and the fix in this PR (likely but I didn't test it).

Should I rename this PR and have only memo fix in this? Change the description? Or should we close this PR and create a new one for just memo?

> 3. your new behavior is something isolated. We continue the discussion, see if we want it or not.

Yes. That's what the useEffect was doing in this PR. We can close this or only do memo fix here. We can have discussion in issue #1141. Not sure.

oliviertassinari:
@ZeeshanTamboli If you could rebase the PR on HEAD an focus on this problem in: 

https://github.com/mui-org/material-ui-x/blob/a75eb4f70a30a998324b03fa9c60c98df7b0c8f3/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx#L157-L158

and only in this PR, it would be 👌. One bug at a time.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -118,6 +118,7 @@ export const useGridFilter = (
         return {
           ...state,
           visibleRows: {
+            ...state.visibleRows,
             visibleRowsLookup,
             visibleRows: Object.entries(visibleRowsLookup)
               .filter(([, isVisible]) => isVisible)
```
oliviertassinari, line 124: The logic of the data grid is not consistent regarding this behavior. Sometimes, we spread extra props even if we override them all, sometimes, we don't. The change seems fair as a systematic approach. It will reduce the cost of future updates.

oliviertassinari (approved):

PR #1880, opened 2021-06-12
[docs] Fix 404 links

oliviertassinari:
@m4theushw I think that we should flatten the URLs. It's inconsistent and will likely create multiple issues later down the road as we fully integrate with the main repo API generation scripts. For instance, it makes it harder to automatically link API pages from the JSDocs as we do in https://twitter.com/olivtassinari/status/1401611864539283457. How about?

- `/api/data-grid/x-grid/` -> `/api/x-grid/`
- `/api/data-grid/data-grid/` -> `/api/data-grid/`
- `/api/data-grid/` -> `/components/data-grid/api/`

### DIFF
--- a/docs/pages/api-docs/data-grid/data-grid.md
+++ b/docs/pages/api-docs/data-grid/data-grid.md
@@ -20,7 +20,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">autoHeight</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid height is dynamic and follow the number of rows in the grid. |
 | <span class="prop-name">autoPageSize</span> | <span class="prop-type">boolean</span> | false | If `true`, the pageSize is calculated according to the container size and the max number of rows to avoid rendering a vertical scroll bar. |
 | <span class="prop-name">checkboxSelection</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid get a first column with a checkbox that allows to select rows. |
-| <span class="prop-name">classes</span> | <span class="prop-type">GridClasses</span> |   | Override or extend the styles applied to the component. See [CSS API](/api/data-grid/#css) below for more details. |
+| <span class="prop-name">classes</span> | <span class="prop-type">GridClasses</span> |   | Override or extend the styles applied to the component. See [CSS API](/api/data-grid/data-grid/#css) below for more details. |
 | <span class="prop-name">className</span> | <span class="prop-type">string</span> |   | CSS classname to add on the outer container. |
 | <span class="prop-name">columnBuffer</span> | <span class="prop-type">number</span> | 2 | Number of columns rendered outside the grid viewport. |
 | <span class="prop-name">columnTypes</span> | <span class="prop-type">GridColumnTypesRecord</span> |   | Extend native column types with your new column types. |
--- a/docs/pages/api-docs/data-grid/x-grid.md
+++ b/docs/pages/api-docs/data-grid/x-grid.md
@@ -21,7 +21,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">autoHeight</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid height is dynamic and follow the number of rows in the grid. |
 | <span class="prop-name">autoPageSize</span> | <span class="prop-type">boolean</span> | false | If `true`, the pageSize is calculated according to the container size and the max number of rows to avoid rendering a vertical scroll bar. |
 | <span class="prop-name">checkboxSelection</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid get a first column with a checkbox that allows to select rows. |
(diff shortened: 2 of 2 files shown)


m4theushw (approved):
The problem with `/api/data-grid/` -> `/components/data-grid/api/` is that it breaks the hierarchy.  The [API Reference](https://material-ui.com/api/data-grid/) is the entrance point for everything that the `data-grid` package provides. Its children should be in an URL that reflects that.

- DataGrid's API Reference - `/something/data-grid`
  - &lt;DataGrid/&gt; - `/something/data-grid/data-grid`
  - GridColDef - `/something/data-grid/grid-col-def`
  - useGridSelector - `/something/data-grid/use-grid-selector`

> For instance, it makes it harder to automatically link API pages from the JSDocs as we do in https://twitter.com/olivtassinari/status/1401611864539283457.

The link to the demo in your tweet was manually added. We just have to do the same in our components. I didn't understand why it's harder to do.

---

Should we use the markdown link syntax with absolute URLs instead of `[[FooBar]]` with relative URLs? VSCode can't render a proper link. There's also [@link](https://jsdoc.app/tags-inline-link.html) which is supported now.

image

oliviertassinari:
> The API Reference is the entrance point for everything that the data-grid package provides. Its children should be in an URL that reflects that.

Then I guess it depends on the respective. One could argue that `/components/data-grid/api/` is only the page that references all the other relevant APIs for the data grid, like `/api/popper`. 

> The link to the demo in your tweet was manually added

It's done automatically by the API generation scripts, by [this function](https://github.com/mui-org/material-ui/blob/fe22aa70a611e42073ea03f64f26b849392009ed/docs/scripts/buildApi.ts#L410) precisely.

oliviertassinari:
I didn't realize we did this as well:

```diff
diff --git a/docs/src/pages/components/data-grid/sorting/sorting.md b/docs/src/pages/components/data-grid/sorting/sorting.md
index 75b4dd05..2d751f54 100644
--- a/docs/src/pages/components/data-grid/sorting/sorting.md
+++ b/docs/src/pages/components/data-grid/sorting/sorting.md
@@ -1,5 +1,6 @@
 ---
 title: Data Grid - Sorting
-components: DataGrid, XGrid
 ---

 # Data Grid - Sorting
@@ -82,8 +83,3 @@ The grid exposes a set of methods that enables all of these features using the i
 - `getSortModel`: Get the sort model currently applied to the grid.
 - `setSortModel`: Set the sort model and trigger the sorting of rows.
 - `onSortModelChange`: Callback fired when the column sorting changed before the grid has sorted its rows.
+
+## API
+
+- [DataGrid](/api/data-grid/data-grid/)
+- [XGrid](/api/data-grid/x-grid/)
```

m4theushw:
> We have been using components instead of hard coding the markdown so that API pages of the components could link the relevant pages of the docs where they are demoed. We break this feature. For instance, we won't get a link back to the demos once we introduce the API page for GridToolbarExport.

I had to remove `components` and hard code the URLs to the demos because the docs is based in the docs from the core and I can't change the way that markdown pages are rendered.

> It's done automatically by the API generation scripts, by this function precisely. The problem here is with:

I didn't know that we were automatically annotating TypeScript definitions. We could support links, but it has to be done manually.

We should discuss how to scope APIs that belong to a different product. The URLs and tools used by the core work well only with the core. We're trying to workaround these limitations to make DataGrid work with them too, but I think we should accept that a different approach has to be taken.

dtassone (approved):

oliviertassinari:
> I had to remove components and hard code the URLs to the demos because the docs is based in the docs from the core and I can't change the way that markdown pages are rendered.

@m4theushw If we kept the URL structure under https://material-ui.com/api/data-grid/, this wouldn't have been an issue right?

> I didn't know that we were automatically annotating TypeScript definitions. We could support links, but it has to be done manually.

For the non-component API pages, I agree because we generate these pages manually, there are no predictibale patterns.
For the component, we can because we generate the pages automatically.

> We should discuss how to scope APIs that belong to a different product. The URLs and tools used by the core work well only with the core. We're trying to work around these limitations to make DataGrid work with them too, but I think we should accept that a different approach (with completely different URLs) has to be taken.

I think that depends on the UX we want to create when browsing the docs. It really depends on how we frame the problem. So it's a continuation of this discussion: https://github.com/mui-org/material-ui/pull/26477#issuecomment-851059745.

If we create a feeling of isolation of realm in the documentation, shouldn't we get these URLs?

- `/components/data-grid/overview/` -> `/x/data-grid/overview/`
- `/api/data-grid/grid-toolbar-export/` -> `/x/api/grid-toolbar-export/` 

Maybe a structuring question, let's add more constraints to shake the tree and see what would work. How would MUI-X support two themes?

--- 

Regarding the tools. It's likely that we will need to update them, agree, especially to support other realms: unstyled/system/design-kit. Still, I think that it's really important we share the same logic. We don't have the bandwidth to own two different implementations of the same documentation. If we were >100 like Google, AWS then maybe, to experiment. But even at this scale, they share a LOT of the same structure between products.

PR #1899, opened 2021-06-13
[DataGrid] Escape regular expression characters in filters

### LINKED ISSUE #1857: [DataGrid] Wildcard char in filters crashes
Filtering a column using contains `*` FireFox throws a `SyntaxError: nothing to repeat` error. 

## Current Behavior 😯

When doing a column filter for Contains * on a string type column the browser throws an error.

image

Firefox Error:
image

Chrome Error:
image

## Expected Behavior 🤔

String Column contains `*` should either be treated as a wild card or return no results.

## Steps to Reproduce 🕹

Using the data grid demo on sandbox: https://codesandbox.io/s/7vhnz

Steps:

1. Select First Name column's "Filter" option
2. In the Filter Pane select Contains option with a Value of *
- image

## Context 🔦

I was attempting to filter for non-empty strings in a column and attempted `*` as a sort of wild card character. 

## Your Environment 🌎

<details>
  
```
  Unable to provide
```
</details>

Repeatable on the Material-UI provided demo:  https://codesandbox.io/s/7vhnz?file=/index.tsx

Browsers: Firefox and Chrome

ZeeshanTamboli:
Fixes #1857 

Solution from https://github.com/mui-org/material-ui-x/issues/1857#issuecomment-857054361

https://deploy-preview-1899--material-ui-x.netlify.app/components/data-grid/

### DIFF
--- a/packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts
+++ b/packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts
@@ -2,8 +2,8 @@ import { GridFilterInputValue } from '../../components/panel/filterPanel/GridFil
 import { GridFilterItem } from '../gridFilterItem';
 import { GridFilterOperator } from '../gridFilterOperator';
 
-function escapeRegExp(string: string): string {
-  return string.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+function escapeRegExp(value: string): string {
+  return value.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
 }
 
 export const getGridStringOperators: () => GridFilterOperator[] = () => [
--- a/packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts
+++ b/packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts
@@ -2,6 +2,10 @@ import { GridFilterInputValue } from '../../components/panel/filterPanel/GridFil
 import { GridFilterItem } from '../gridFilterItem';
 import { GridFilterOperator } from '../gridFilterOperator';
 
+function escapeRegExp(string: string): string {
+  return string.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+}
+
 export const getGridStringOperators: () => GridFilterOperator[] = () => [
   {
     value: 'contains',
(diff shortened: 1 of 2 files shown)


oliviertassinari (changes requested):
Please add a test case

packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts
```diff
@@ -2,6 +2,10 @@ import { GridFilterInputValue } from '../../components/panel/filterPanel/GridFil
 import { GridFilterItem } from '../gridFilterItem';
 import { GridFilterOperator } from '../gridFilterOperator';
 
+function escapeRegExp(string: string): string {
```
dtassone, line 5: can we avoid `string: string` please? You can call it `value` or `regex` or...?

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -184,6 +184,14 @@ describe('<DataGrid /> - Filter', () => {
         expect(getColumnValues()).to.deep.equal(expected);
       });
     });
+
+    ['contains', 'startsWith', 'endsWith'].forEach((operatorValue) => {
+      it('should escape RegExp characters if applied as filter values', () => {
+        const regExpToEscape = '[-[]{}()*+?.,\\^$|#s]';
+        render(<TestCase value={regExpToEscape} operatorValue={operatorValue} />);
+        expect(getColumnValues()).to.deep.equal([]);
```
oliviertassinari, line 192: I don't think that it's enough, we should still be able to find the result.

ZeeshanTamboli: If the assertion is checking `[]`, it is ignoring the input value and returning zero rows if it is a regex. So what do we want to support? Should we escape regex and show empty or able to filter without any crash even if it is a RegExp? Like support * as all rows?
Also I checked with Commodity `contains` filter one case wherein `soy*` worked before as `starts with soy` and now it simply returns zero rows for `soy*` in Xgrid demo.  I will have to check that.

dtassone: IMO soy* should just search for soy*? So we don't support regex search
Otherwise if a value is *, it is not filterable

ZeeshanTamboli: > IMO soy* should just search for soy*

This works with the current fix. 
Example: If a row contains `soy*bean` value string, filter operation `contains` `soy*` will return this row. What happens is the filter regex is now `soy\*bean` making * as literal instead of regex special. 
 I could add a test case for this.

dtassone: Yes it's the correct behaviour. And yes please add a test so it's a bit more robust.

ZeeshanTamboli: I added one.

dtassone (approved):

PR #1903, opened 2021-06-14
[DataGrid] Use binary search to find column indexes in virtualization

Janpot:
Noticed when playing around with the [virtualization demo](https://material-ui.com/components/data-grid/virtualization/#column-virtualization) in codesandbox that when I tried 100,000 columns, it would hit codesandbox infinite loop detection. Having worked on my own [toy implementation](https://mui-plus.vercel.app/components/DataGrid/virtualization) of a virtualized Material UI datagrid, I realized it was looping through all possible columns to find viewport intersection. Since it has to search an ordered array, a binary search is possible here and drastically reduces iterations on high column counts. I reviewed the Material UI Datagrid implementation and found that [my implementation](https://github.com/Janpot/mui-plus/blob/03b78402418b9772ccb3bfea52a1e986fc2be520/packages/mui-plus/src/utils/virtualization.ts#L11) was easy to port. Quick measurement of performance showed about a 200x performance improvement of the `getColumnFromScroll` function execution time on 100,000 columns (~10ms => ~0.05ms). 
I realize ofcourse that wanting to render 100k columns is very much an edge-case, so feel free to discard this PR if the added complexity isn't worth it. On 1000 columns, like in the current example, the performance improvement is much less dramatic.

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualColumns.ts
@@ -26,7 +26,7 @@ type UseVirtualColumnsReturnType = [
   UpdateRenderedColsFnType,
 ];
 
-// Uses binary search to avoid looping through all positions
+// Uses binary search to avoid looping through all possible positions
 function getIdxFromScroll(
   offset: number,
   positions: number[],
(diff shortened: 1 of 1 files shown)


oliviertassinari:
> I realize ofcourse that wanting to render 100k columns is very much an edge-case

Can the same optimization be applied to the row virtualization? This would make it a more appealing. Regarding the complexity of a dichotomy, with an abstraction that we could use in more places, I think that it could potentially do it.

Janpot:
> Can the same optimization be applied to the row virtualization?

Absolutely, the algorithm is agnostic in terms of direction. I've extracted a `getIdxFromScroll` that can be reused in other virtualization scenarios of variable sized items.

oliviertassinari:
@Janpot It looks great 👌.

It's impressive to see what you have built on https://mui-plus.vercel.app/components/DataGrid/pinned in 3 months.

- [useResizeObserver](https://github.com/Janpot/mui-plus/blob/master/packages/mui-plus/src/DataGrid/useResizeObserver.ts) is one I have advocated adding in https://github.com/mui-org/material-ui/issues/22519. (there are other utils that you could have imported from `@material-ui/utils`).
- The virtualization is interesting. For curiosity, I have tried to create a comparator (we always looking for ways to make it fast, it's really hard).
  - MUI+ https://codesandbox.io/s/recursing-jackson-4xgsl, deployed to production: https://csb-4xgsl.netlify.app/. The average frame rate when I scroll
  Capture d’écran 2021-06-15 à 13 57 12

  - XGrid https://codesandbox.io/s/awesome-shtern-6pyxl, deployed to production: https://csb-6pyxl.netlify.app/. The average frame rate when I scroll
  Capture d’écran 2021-06-15 à 13 57 07

MUI+ performs about +44% better than XGrid (36 FPS vs. 25 FPS) and doesn't suffer from blocking screens every time the window jumps (the **red** bars). @dtassone Could you have a closer look at the implementation? Maybe we can replace our implementation with this one?

Janpot:
Oh that's interesting, I hadn't done any real benchmarking yet. I'm glad it's useful. Happy to answer any questions. I converted the styles as per your suggestion, thanks!

oliviertassinari (approved):

dtassone (approved):

PR #1910, opened 2021-06-15
[DataGrid] Remove disableClickEventBubbling

### LINKED ISSUE #1296: [DataGrid] Remove disableClickEventBubbling,  as you can use event.stopPropagation
We can now remove this prop as it is only useful for clicks and it does not really scale for all the other events. 
If you want to stop the click propagation, you can call `event.stopPropagation`

```diff
-<DataGrid disableClickEventBubbling />
+<DataGrid onCellClick={(event)=> event.stopPropagation}/>
```

relate #947

m4theushw:
### Breaking changes

- [DataGrid] The `disableClickEventBubbling` prop was removed because the same outcome can be obtained calling `event.stopPropagation()`.

  ```diff
  -<DataGrid disableClickEventBubbling />
  +<DataGrid onCellClick={(event) => event.stopPropagation()} />
  ```

---

Closes #1296


dtassone (approved):

missbruni:
I can see that this is now not the case because https://github.com/mui-org/material-ui-x/releases/tag/v4.0.0-alpha.32 has also removed the ability to stop the event propagation. So how are we meant to be stopping the cell to go into view mode when commited please ? @m4theushw

m4theushw:
> I can see that this is now not the case because https://github.com/mui-org/material-ui-x/releases/tag/v4.0.0-alpha.32 has also removed the ability to stop the event propagation. So how are we meant to be stopping the cell to go into view mode when commited please ? @m4theushw

@missbruni Could you open a new issue with a reproduction? I'm not sure if it was ever possible to stop the cell from going to the view mode once it's commited, but that's something we need to allow.

missbruni:
@m4theushw sure will do. This is how we were handling our custom row editing implementation. We were doing a stopPropagation inside the callback to stop the cell from going back to view mode, but now with the new versions since we are not getting this event, this is no longer possible. I will raise one now.

missbruni:
> > I can see that this is now not the case because https://github.com/mui-org/material-ui-x/releases/tag/v4.0.0-alpha.32 has also removed the ability to stop the event propagation. So how are we meant to be stopping the cell to go into view mode when commited please ? @m4theushw
> 
> @missbruni Could you open a new issue with a reproduction? I'm not sure if it was ever possible to stop the cell from going to the view mode once it's commited, but that's something we need to allow.

@m4theushw sure will do. This is how we were handling our custom row editing implementation. We were doing a stopPropagation inside the callback to stop the cell from going back to view mode, but now with the new versions since we are not getting this event, this is no longer possible. I will raise one now.

missbruni:
@m4theushw here, https://github.com/mui-org/material-ui-x/issues/2146
Any comments / recommendation to work around this issue for us would be much appreciated as we rely on it for our own row editing implementation and until we have a workaround we are unable to take advantage of the bug fixes for versions 32 and above!

PR #1913, opened 2021-06-16
[DataGrid] Include MUI core component localizations in `localeText`

### LINKED ISSUE #1107: [DataGrid] Pagination localization

## Current Behavior 😯

Currently of the footer of the table look like this while using pagination:
108724352-83c68f80-7525-11eb-84ff-af8a01f825b6

The word "of" is somehow not changeable and it affects the localization of the DataGrid when implementing it on another language.

## Context 🔦

I am trying to implement a DataGrid with Pagination on German. Everything seems fine until this small issue.

### LINKED ISSUE #1854: [DataGrid] Changing ```Rows per page``` localization missing keys
Hello,
I was able to change the Data grid localization thanks to [this](https://material-ui.com/components/data-grid/localization/).  I also used [this](https://github.com/mui-org/material-ui-x/blob/HEAD/packages/grid/_modules_/grid/constants/localeTextConstants.ts) to know which keys are supported.

I was not able to change the `Rows per page` locale. Am I missing something ?
It looks like we are missing a key for this ?

## Summary 💡

Add a keys in the [DataGrid localization](https://material-ui.com/components/data-grid/localization/#translation-keys). 

Another solution could be: having a props `labelRowsPerPage` like in the [TablePagination](https://material-ui.com/api/table-pagination/) component.

DanailH:
Fix #1107 
FIx #1854 

I had a thought about how to solve the problem mentioned in the 2 tickets above and I have a possible solution without introducing any duplication and without adding regressions to the way the grid localization currently works. 

I added a **Breaking Change** label because I needed to change some of the types but even with this we still support **v5**, people can use the localization with the theme and in addition, they can just provide all the localizations through the `localeText` prop.

The only duplication is that I needed to list the components for which there are localizations in the core (but without any actual values, only their names).

ToDo:

- [x] Update the Localization docs


packages/grid/_modules_/grid/constants/localeTextConstants.ts
```diff
@@ -98,4 +98,12 @@ export const GRID_DEFAULT_LOCALE_TEXT: GridLocaleText = {
...
   booleanCellTrueLabel: 'true',
   booleanCellFalseLabel: 'false',
+
+  // Core components keys
+  MuiTablePagination: {},
+  MuiPagination: {},
+  MuiAlert: {},
+  MuiBreadcrumbs: {},
+  MuiRating: {},
+  MuiAutocomplete: {},
```
dtassone, line 108: We don't natively use all these components. Why should we add them all?

oliviertassinari (commented):
It also seems that a large part of the issue was with the lack of documentation on his aspect (the core components). So adding something about it might help. I mean, we likely want to explain that developers have to translate the core components used. And how it works.

packages/grid/_modules_/grid/constants/localeTextConstants.ts
oliviertassinari: One question I would have is, let's say we want to implement Ant Design, as a design specification. How does these change fit it?

DanailH:
> It also seems that a large part of the issue was with the lack of documentation on his aspect (the core components). So adding something about it might help. I mean, we likely want to explain that developers have to translate the core components used. And how it works.

I know what you mean. The thing is that the solution that @m4theushw described here https://github.com/mui-org/material-ui-x/issues/1854#issuecomment-856867956 is missing from the docs (or at least from the Localization docs). Probably it makes sense adding a section for it.

packages/grid/_modules_/grid/constants/localeTextConstants.ts
DanailH: We don't, correct but for some of them, it does make sense to have them.  Technically we only use the `MuiTablePagination` so we can only keep that one and add the rest on demand if we happen to use them in the future. The benefit is that that way all the localizations are kept together.

@oliviertassinari regarding the Ant Design - I didn't understand the problem - is it if we use a different pagination component?

Actually, these are only here because the way that the `getLocaleText` API works is that it throws an error if the key doesn't exist. If that was not the case we woudn't need to keep these here at all.

oliviertassinari:
> the solution that @m4theushw described here #1854 (comment)

Not necessarily, what I had in mind is to make sure we have the documentation counterpart of the solution we deploy. Solutions don't have to be source changes only, it could also heavily depends on the documentation.

We have a couple of different options, ranging from:

1. Hide the notion of core components to the developers, have each translation key flat in the locale, no need to update the docs
2. A documentation only fix, explain how the problem can be solved with the componentsProps prop.
3. The hybrid, what we have started to do here. We are partially coupled with the core components, but still get a single place where developers can find all the props that can be translated, updating the docs still seems important.

DanailH:
@oliviertassinari Understood. I tried to go with 1. but listing the props flat is a problem since we have to respect the typings and more importantly because we support both v4 and v5.

Option 2. is the easiest and would require the least effort.
Option 3. - if we can agree on the implementation of course I will update the docs with what has changed and add examples.

packages/grid/_modules_/grid/utils/getGridLocalization.ts
```diff
@@ -27,20 +27,18 @@ export const getGridLocalization = (
       components: {
         MuiDataGrid: {
           defaultProps: {
-            localeText: gridTranslations,
+            localeText: { ...gridTranslations, ...coreTranslations?.components },
           },
         },
-        ...coreTranslations?.components,
```
m4theushw, line 33: We need to revert this line and its v4 counterpart below. If another component from the core is used, it will not get translations.

packages/grid/_modules_/grid/constants/localeTextConstants.ts
m4theushw: I would keep only `MuiTablePagination`. If we add all of them it might create doubts for who is doing translations.

oliviertassinari: > is it if we use a different pagination component?

Different design specs can be supported in different ways. The most likely path we will take is that we will expose the same MuiPagination structure, but styled differently. The only thing that we would need to do is to reinject the new components. So we might be able to use the same translation keys, as we would display the same content.

packages/grid/_modules_/grid/utils/getGridLocalization.ts
oliviertassinari: @m4theushw It seems that you envision a different solution. This change seems to match with the intent to have all the translations from happening inside MuiDataGrid.localeText, even for the core components.

m4theushw: I have no problem with appending the core translations into `MuiDataGrid.localeText`. My point is that if another component from the core (e.g. Alert) is used outside the DataGrid it will not get translations. That's because in the docs we recommend to use the locale from the data-grid package to create the theme. Open [this](https://codesandbox.io/s/create-react-app-forked-4545g?file=/src/App.js) CodeSandbox and note that the standalone TablePagination is not translated.

DanailH: I see your worry - I will update the docs to match this implementation. That's why it's marked as a breaking change because you will no longer import all translations from the `X` package, you would need a mixed solution depending on your case.

docs/src/pages/components/data-grid/localization/localization.md
```diff
@@ -44,6 +44,14 @@ Note that `createMuiTheme` accepts any number of arguments.
 If you are already using the [translations of the core components](/guides/localization/#locale-text), you can add `bgBG` as a new argument.
 The same import works with `XGrid` as it's an extension of `DataGrid`.
 
+**Note**: if you want to pass in language translations directly to the grid and not use the `createMuiTheme` in combination with the `ThemeProvider` you can directly load the language translations you need from the `DataGrid` or `XGrid` package.
```
dtassone, line 47: I would remove `you need`

dtassone (approved):

docs/src/pages/components/data-grid/localization/localization.md
```diff
@@ -44,6 +44,14 @@ Note that `createMuiTheme` accepts any number of arguments.
 If you are already using the [translations of the core components](/guides/localization/#locale-text), you can add `bgBG` as a new argument.
 The same import works with `XGrid` as it's an extension of `DataGrid`.
 
+**Note**: if you want to pass in language translations directly to the grid and not use the `createMuiTheme` in combination with the `ThemeProvider` you can directly load the language translations from the `DataGrid` or `XGrid` package.
```
m4theushw, line 47:
```suggestion
If you want to pass language translations directly to the grid without using `createMuiTheme` and `ThemeProvider`, you can directly load the language translations from the `@material-ui/data-grid` or `@material-ui/x-grid` package.
```

docs/src/pages/components/data-grid/localization/localization.md
```diff
@@ -44,6 +44,14 @@ Note that `createMuiTheme` accepts any number of arguments.
 If you are already using the [translations of the core components](/guides/localization/#locale-text), you can add `bgBG` as a new argument.
 The same import works with `XGrid` as it's an extension of `DataGrid`.
```
m4theushw, line 45: Just to avoid breaking the line in the the period sign.

image


```suggestion
The same import works for `XGrid` as it's an extension of `DataGrid`.
```

docs/src/pages/components/data-grid/localization/localization.md
```diff
@@ -44,6 +44,14 @@ Note that `createMuiTheme` accepts any number of arguments.
 If you are already using the [translations of the core components](/guides/localization/#locale-text), you can add `bgBG` as a new argument.
```
m4theushw, line 44: If the user is already using the translations from core, using this approach will break the core components, because we don't append anymore the core translation in our translations. We need to explain that `localeText` has to be used. Also, could you update the PR's description to include a migration path?

DanailH: Am no, why you can add both the translations from the `DataGrid` along with the translations from the core. You can do :
```
const theme = createMuiTheme(
  {
    palette: {
      primary: { main: '#1976d2' },
    },
  },
  CoreBgBG,
  DataGridBgBG,
);
```
Regarding the migration, one would just need to add the translations from the core as an additional argument to the `createMuiTheme`.

packages/grid/_modules_/grid/utils/getGridLocalization.ts
oliviertassinari: > Open this CodeSandbox and note that the standalone TablePagination is not translated.

The table component not translated sounds like the expected behavior, it does no longer leak.

packages/grid/_modules_/grid/models/api/gridLocaleTextApi.ts
```diff
@@ -1,7 +1,12 @@
+import { Localization as CoreLocalization } from '@material-ui/core/locale';
+
+// @ts-ignore
```
oliviertassinari, line 3: It might be better to use 
```suggestion
// @ts-expected-error THE REASON
```

docs/src/pages/components/data-grid/localization/localization.md
oliviertassinari: @DanailH 👍 for adding this as an example.

DanailH: Added

packages/grid/_modules_/grid/models/api/gridLocaleTextApi.ts
DanailH: Done

docs/src/pages/components/data-grid/localization/localization.md
```diff
@@ -42,7 +42,35 @@ const theme = createMuiTheme(
...
+
+const theme = createMuiTheme(
+  {
+    palette: {
+      primary: { main: '#1976d2' },
+    },
+  },
+  bgBG,
+  coreBgBG,
+);
```
m4theushw, line 60: 👍

m4theushw (approved):

packages/grid/_modules_/grid/models/api/gridLocaleTextApi.ts
```diff
@@ -1,7 +1,12 @@
+import { Localization as CoreLocalization } from '@material-ui/core/locale';
+
+// @ts-expect-error We have different structure in v5
+type MuiComponentsLocalization = CoreLocalization['props'] | CoreLocalization['components'];
+
 /**
  * Set the types of the texts in the grid.
  */
-export interface GridLocaleText {
+export interface GridLocaleText extends Partial<MuiComponentsLocalization> {
```
oliviertassinari, line 9: This type is not correct. 

1. This should fail: https://codesandbox.io/s/material-demo-forked-d8djr?file=/demo.tsx, it doesn't. Fixed in #2117.
2. The tests of https://github.com/mui-org/material-ui/pull/27196 shouldn't, they do. Fixed in #2117

PR #1919, opened 2021-06-17
[DataGrid] Fix double-click issue

oliviertassinari:
Fix the double click issue: https://github.com/mui-org/material-ui-x/issues/1141#issuecomment-798547109. In the recording, you can see that I click twice, and the state stays the same. *checked indeterminate* is the **same state** as *unchecked indeterminate*. In core v5 we removed the visual distinction to match screen readers.

Note that the fix is mostly removing code we don't need* (*assuming it's better to not react to clicks on the checkbox when there are no rows).

I worked on this because #1879 didn't seem to go in a rigorous direction. I'm mostly interested in the test case and isolation. I'm a believer that the quality of the tests ultimately drives the quality of the code.

### DIFF
--- a/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
@@ -13,6 +13,10 @@ import { GridApiRef, GridComponentProps, useGridApiRef, XGrid } from '@material-
 
 const isJSDOM = /jsdom/.test(window.navigator.userAgent);
 
+function getSelectedRows(apiRef) {
+  return Array.from(apiRef.current.getSelectedRows().keys());
+}
+
 describe('<XGrid /> - Selection', () => {
   // TODO v5: replace with createClientRender
   const render = createClientRenderStrictMode();
@@ -122,7 +126,7 @@ describe('<XGrid /> - Selection', () => {
 
   it('should clean the selected ids when the rows prop changes', () => {
     const { setProps } = render(<Test selectionModel={[0, 1, 2]} checkboxSelection />);
-    expect(apiRef.current.getSelectedRows()).to.have.keys([0, 1, 2]);
+    expect(getSelectedRows(apiRef)).to.deep.equal([0, 1, 2]);
     setProps({
       rows: [
         {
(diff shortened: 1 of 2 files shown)


packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
```diff
@@ -126,4 +133,30 @@ describe('<XGrid /> - Selection', () => {
...
+          columnField: 'brand',
+          operatorValue: 'contains',
+          value: 'Puma',
+        },
+      ],
+    });
+    expect(getColumnValues(1)).to.deep.equal(['Puma']);
+    fireEvent.click(selectAll);
+    // TODO fix, should be only 2
+    expect(Array.from(apiRef.current.getSelectedRows().keys())).to.deep.equal([0, 1, 2]);
```
dtassone, line 154: fine to use apiRef to assert the state but your test coverage would be higher if you check the `Mui-selected` css class on the row

packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
```diff
@@ -126,4 +133,30 @@ describe('<XGrid /> - Selection', () => {
...
+        {
+          columnField: 'brand',
+          operatorValue: 'contains',
+          value: 'Puma',
+        },
+      ],
+    });
+    expect(getColumnValues(1)).to.deep.equal(['Puma']);
+    fireEvent.click(selectAll);
+    // TODO fix, should be only 2
```
ZeeshanTamboli, line 153: Yup, this gets fixed by removing `React.memo` after which we get the updated visible rows in `GridHeaderCheckbox`.

ZeeshanTamboli (approved):
Nice removal of unnecessary state usage from `GridHeaderCheckbox`. Looks good to me!

packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
oliviertassinari: My bad, no, I got trapped. We can't, the selected rows 0 and 1 are not visible.

PR #1922, opened 2021-06-17
[DataGrid] Make the CSV export respect the `valueFormatter`

### LINKED ISSUE #1814: [DataGrid] valueGetter does not apply to CSV export

## Current Behavior 😯

The CSV export does not get the column value that is displayed in the table through valueGetter. So in the case of objects, it shows [Object object] rather than the value derived from valueGetter of the column.

## Expected Behavior 🤔

The exported CSV column that use valueGetter should show the value with valueGetter applied.

## Steps to Reproduce 🕹

https://codesandbox.io/s/material-demo-forked-h262z

Steps:

1. Export to CSV
2. Compare values between the table and the CSV file.

## Context 🔦

Related to https://github.com/mui-org/material-ui-x/issues/1145. Not sure if this is a duplicate or tracked separately for X-Grid.

Our customers would like to be able to export the table content to do additional work on the data. To display what's on the table, we need:

1. valueGetter to apply to column values in the CSV to reflect what is being displayed in the table
2. additional way to change the value just for the export. For example, something like `exportValueGetter` that is run just for export (I included an example in the codesandbox above).

## Your Environment 🌎

<details>
  
```
  System:
    OS: macOS 11.0.1
  Binaries:
    Node: 16.1.0 - ~/.n/bin/node
    npm: 7.11.2 - ~/.n/bin/npm
  Browsers:
    Chrome: 90.0.4430.212
    Edge: Not Found
    Firefox: Not Found
    Safari: 14.0.1
  npmPackages:
    @material-ui/core: ^4.11.0 => 4.11.2
    @material-ui/icons: ^4.5.1 => 4.11.2
    @material-ui/lab: ^4.0.0-alpha.57 => 4.0.0-alpha.57
    @material-ui/pickers: ^3.2.7 => 3.2.10
    @material-ui/styles:  4.11.2
    @material-ui/system:  4.11.2
    @material-ui/types:  5.1.0
    @material-ui/utils: ^4.10.2 => 4.11.2
    @material-ui/x-grid: ^4.0.0-alpha.27 => 4.0.0-alpha.30
    @types/react:  17.0.8
    react: 16.14.0 => 16.14.0
    react-dom: ~16.9.0 => 16.9.0
    typescript:  3.9.7
```
</details>

## Order id 💳

[Order #25281] (May 25, 2021)

DanailH:
Fix #1814


packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
```diff
@@ -35,12 +37,12 @@ interface BuildCSVOptions {
...
   rows: Map<GridRowId, GridRowModel>;
   selectedRows: Record<string, GridRowId>;
-  getCellValue: (id: GridRowId, field: string) => GridCellValue;
+  getCellParams: (id: GridRowId, field: string) => GridCellParams;
   delimiterCharacter: GridExportCsvDelimiter;
 }
 
 export function buildCSV(options: BuildCSVOptions): string {
-  const { columns, rows, selectedRows, getCellValue, delimiterCharacter } = options;
+  const { columns, rows, selectedRows, getCellParams, delimiterCharacter } = options;
```
dtassone, line 45: Should we make it an opt out option?

DanailH: Applying the same logic then we should also make `delimiterCharacter` optional. Although you need either 'getCellParams' or `getCellValue` here to get the actual cell value because we only have access to the cell ID and using the `apiRef` in here is not ideal as that file is with helper/utility functions that are grid agnostic.
IMO we can keep it for now as it is.

dtassone: That's ok, I meant using the value formatter, there might be cases where one would prefer the raw value, maybe it should be column based as well 🤔 
What do you think? @mui-org/x

m4theushw: The first step was to apply the same `valueFormatter` to the CSV export, that might be sufficient for most of the use cases. If someone requests for a per-column formatting, we can add an exclusive value formatter for the export: https://github.com/mui-org/material-ui-x/issues/1814#issuecomment-852678630

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
```diff
@@ -1,5 +1,5 @@
 import {
```
m4theushw, line 1: There's a typo in the name of this folder / file: `seralizers` -> `serializers` and `csvSeraliser` -> `csvSerializer`

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
dtassone: Yes that's what we said originally. I'm just being extra cautious here about numbers, ie `1,000`, should be exported as `"1,000"`, would that be interpreted as numbers by excel? 🤔

dtassone: Just tested and it seems ok. Let's not do the option for now

dtassone (approved):

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
m4theushw: Yeah, we had a PR about that: https://github.com/mui-org/material-ui-x/pull/1656#issuecomment-840122724 If the OS is correctly configured, we just need to use the same formatting in the CSV export.

packages/grid/_modules_/grid/hooks/features/export/seralizers/csvSeraliser.ts
DanailH: Nice, done

m4theushw (approved):

PR #1923, opened 2021-06-17
[DataGrid] Improve updateRows performance

N2D4:
Given `n` total rows and `m` row updates, reduces asymptotic complexity of `updateRows` from `O(n*m)` to `O(n+m)`.

We do this by performing all updates together, instead of each one individually.

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
@@ -187,9 +187,8 @@ export const useGridRows = (
       });
       if (updatedLookup) {
         internalRowsState.current.idRowsLookup = updatedLookup;
+        setGridState((state) => ({ ...state, rows: { ...internalRowsState.current } }));
       }
-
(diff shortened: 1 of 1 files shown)


packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
```diff
@@ -174,16 +175,20 @@ export const useGridRows = (
           addedRows.push(partialRow);
           return;
         }
-        const lookup = { ...internalRowsState.current.idRowsLookup };
 
-        lookup[id] = {
+        if (!updatedLookup) {
```
dtassone, line 179: when can that be true?

N2D4: Only on the very first iteration (we could initialize `updatedLookup` too, but that would waste resource if we only add/remove rows)

dtassone (approved):

m4theushw (approved):

PR #1925, opened 2021-06-18
[DataGrid] Make GRID_ROWS_CLEAR private

oliviertassinari:
### Breaking changes

- [DataGrid] Make GRID_ROWS_CLEAR private (#1925) @oliviertassinari 
 The `GRID_ROWS_CLEAR` event was always triggered alongside `GRID_ROWS_SET`. You can listen to the latter event only.

---

~See the rationale in https://github.com/mui-org/material-ui-x/pull/1862#discussion_r654370166. When GRID_ROWS_CLEARED triggers, `GRID_ROWS_SET` is also triggered, no need to update the sorting state **twice**.~

~Note that this makes `GRID_ROWS_CLEARED` an orphan, nobody is listening to the event.~

~If I'm missing something and is really required, then it surfaces a hole in the test suite, the CI is green.~

### DIFF
--- a/docs/src/pages/components/data-grid/events/events.json
+++ b/docs/src/pages/components/data-grid/events/events.json
@@ -162,7 +162,6 @@
     "description": "Fired when the user stops resizing a column. Called with an object <code>{ field: string }</code>."
   },
   { "name": "columnOrderChange", "description": "Fired when the user ends resizing a column." },
-  { "name": "rowsClear", "description": "Fired when the grid is emptied." },
   {
     "name": "columnsChange",
     "description": "Fired when the columns state is changed.\nCalled with an array of strings correspoding to the field names."
--- a/packages/grid/_modules_/grid/constants/eventsConstants.ts
+++ b/packages/grid/_modules_/grid/constants/eventsConstants.ts
@@ -412,7 +412,10 @@ export const GRID_ROWS_UPDATE = 'rowsUpdate';
 export const GRID_ROWS_SET = 'rowsSet';
 
 /**
- * Fired when the grid is emptied.
+ * Implementation detail.
+ * Fired to reset the sortedRow when the set of rows changes.
+ * It's important as the rendered rows are coming from the sortedRow
+ * @ignore - do not document.
  * @event
  */
 export const GRID_ROWS_CLEAR = 'rowsClear';
--- a/packages/grid/_modules_/grid/hooks/features/filter/visibleGridRowsState.ts
+++ b/packages/grid/_modules_/grid/hooks/features/filter/visibleGridRowsState.ts
@@ -2,7 +2,6 @@ import { GridRowId } from '../../../models/gridRows';
 
 export interface VisibleGridRowsState {
   visibleRowsLookup: Record<GridRowId, boolean>;
-
   visibleRows?: GridRowId[];
 }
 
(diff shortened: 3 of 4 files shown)


dtassone:
I don't think it is dead code at least when I wrote it, it was not.
If I remember well it is to clear the sortedRows used sortedGridRowIdsSelector, to avoid getting an error when a row was added or removed and not in the lookup.

dtassone:
it seems ok now https://deploy-preview-1925--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-streaming--slow-update-grid

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
```diff
@@ -3,7 +3,6 @@ import {
   GRID_COLUMN_HEADER_CLICK,
   GRID_COLUMN_HEADER_KEYDOWN,
   GRID_COLUMNS_UPDATED,
-  GRID_ROWS_CLEARED,
```
dtassone, line 6: you can remove the constant as well and update the api

dtassone:
> I don't think it is dead code at least when I wrote it, it was not.
> If I remember well it is to clear the sortedRows used sortedGridRowIdsSelector, to avoid getting an error when a row was added or removed and not in the lookup.

I think the rendering optimisation fixed it

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
oliviertassinari: This would be a breaking change. I think that we can decorate the two. My objective here is to simplify the logic.

dtassone: we are in alpha. Let's do it I would say

m4theushw (commented):
This PR broke https://deploy-preview-1925--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-filter--commodity-with-new-rows-via-api

The following diff passes on HEAD but here doesn't:

```diff
diff --git a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
index c800f9cb..591610b2 100644
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -162,6 +162,7 @@ describe('<XGrid /> - Rows', () => {

     it('should allow to reset rows with setRows and render after 100ms', () => {
       render(<TestCase />);
+      expect(getColumnValues()).to.deep.equal(['Nike', 'Adidas', 'Puma']);
       const newRows = [
         {
           id: 3,
@@ -169,10 +170,9 @@ describe('<XGrid /> - Rows', () => {
         },
       ];
       apiRef.current.setRows(newRows);
-
-      clock.tick(50);
-      expect(getColumnValues()).to.deep.equal(['Nike', 'Adidas', 'Puma']);
-      clock.tick(50);
+      // Force an update before the 100ms
+      apiRef.current.forceUpdate(() => apiRef.current.state);
+      clock.tick(100);
       expect(getColumnValues()).to.deep.equal(['Asics']);
     });
```

My suggestion is to try to remove the timeout from the custom `forceUpdate` in `useGridRows` or to reset the sorted rows after updating the rows in this hook (might break the separation of concern).

oliviertassinari:
> This PR broke

@m4theushw Oh wow, great catch. I have added a failing test case to avoid regression in the future. I have also updated the assertion to illustrate the behavior. I have reduced the ambition of the changes.

dtassone:
Should we use this PR to rename the event to match the convention CLEARED => CLEAR?

oliviertassinari:
I'm waiting for #1862 to be merged to rebase.

m4theushw (approved):

dtassone (approved):

PR #1872, opened 2021-06-11
[DataGrid] Rename classes according to new convention

DanailH:
Fixes points 2. and 4. of #408 

### Breaking Changes

The main data grid component:

- `.data-container` is removed
- `.MuiDataGrid-columnHeaderSortable` is renamed to `.MuiDataGrid-columnHeader--sortable`
- `.MuiDataGrid-columnHeaderCenter` is renamed to `.MuiDataGrid-columnHeader--alignCenter`
- `.MuiDataGrid-columnHeaderRight` is renamed to `.MuiDataGrid-columnHeader--alignRight`
- `.MuiDataGrid-columnHeader-draggable` is renamed to `.MuiDataGrid-columnHeaderDraggableContainer`
- `.MuiDataGrid-columnHeaderSortable` is renamed to `.MuiDataGrid-columnHeader--sortable`
- `.MuiDataGrid-columnHeaderMoving` is renamed to `.MuiDataGrid-columnHeader--moving`
- `.MuiDataGrid-columnHeaderSorted` is renamed to `.MuiDataGrid-columnHeader--sorted`
- `.MuiDataGrid-columnHeaderNumeric` is renamed to `.MuiDataGrid-columnHeader--numeric`
- `.MuiDataGrid-columnHeader-dropZone` is renamed to `.MuiDataGrid-columnHeaderDropZone`
- `.MuiDataGrid-columnSeparatorResizable` is renamed to `.MuiDataGrid-columnSeparator--resizable`
- `.MuiDataGrid-cellWithRenderer` is renamed to `.MuiDataGrid-cell--withRenderer`
- `.MuiDataGrid-cellLeft` is renamed to `.MuiDataGrid-cell--textLeft`
- `.MuiDataGrid-cellRight` is renamed to `.MuiDataGrid-cell--textRight`
- `.MuiDataGrid-cellCenter` is renamed to `.MuiDataGrid-cell--textCenter`
- `.MuiDataGrid-cellEditing` is renamed to `.MuiDataGrid-cell--editing`
- `.MuiDataGrid-cellEditable` is renamed to `.MuiDataGrid-cell--editable`
- `.MuiDataGrid-editCellBoolean` is renamed to `.MuiDataGrid-editBooleanCell`
- `.MuiDataGrid-editCellInputBase` is renamed to `.MuiDataGrid-editInputCell`
- `.MuiDataGrid-scrollArea-left` is renamed to `.MuiDataGrid-scrollArea--left`
[...]

PR #1928, opened 2021-06-18
[test] Fix tests

m4theushw:
The pipeline broke after the merge of #1872.


oliviertassinari:
@m4theushw Thanks for the quick patch, It was fixed even before I had the time to raise it 😁

oliviertassinari:
Back to business as usual

Issue #1197, opened 2021-03-10
[DataGrid] Copying a cell (cmd +c) selects and copies the whole row


Tollwood:
As a user I want to copy cell content so that I can use it in other places. 
Select + right click copy works fine. Using the keyboard shortcut (cmd/ctrl + c) does select and copy the whole row.

- [x ] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

The whole row is selected and copied  instead of just the selection

## Expected Behavior 🤔

only the selection should be copied

## Steps to Reproduce 🕹
Steps:

1. open https://material-ui-x.netlify.app/storybook/?path=/story/x-grid-demos-playground--commodity
2. select / double click a cell
3. use keyboard shortcut (cmd +c / ctrl + c)
4. The whole row gets selected and copied

## Context 🔦

I want to copy cell content from my table and search for this value, use it in another part of the application as an input

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  tested in chrome
  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.15.1 - ~/.nvm/versions/node/v14.15.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.1/bin/npm
  Browsers:
    Chrome: 88.0.4324.192
    Edge: Not Found
    Firefox: 85.0.2
    Safari: 14.0.2
  npmPackages:
    @material-ui/core: ^4.11.3 => 4.11.3 
    @material-ui/icons: ^4.11.2 => 4.11.2 
    @material-ui/lab: ^4.0.0-alpha.57 => 4.0.0-alpha.57 
    @material-ui/pickers: ^3.2.10 => 3.2.10 
    @material-ui/styles:  4.11.3 
    @material-ui/system:  4.11.3 
    @material-ui/types:  5.1.0 
    @material-ui/utils:  4.11.2 
    @material-ui/x-grid: 4.0.0-alpha.22 => 4.0.0-alpha.22 
    @material-ui/x-license:  4.0.0-alpha.21 
    @types/react: ^16.9.0 => 16.14.4 
    react: 17.0.1 => 17.0.1 
    react-dom: 17.0.1 => 17.0.1 
[...]

PR #1929, opened 2021-06-18
[DataGrid] Allow to copy the selected rows to the clipboard

m4theushw:
Related to #199. I didn't use "closes" because this issue is much bigger and we can't really implement it now.

Preview: https://deploy-preview-1929--material-ui-x.netlify.app/components/data-grid/demo/


oliviertassinari:
Is it related to #1197?

m4theushw:
@oliviertassinari Interesting, this issue was never mentioned in #199. This PR is actually an enhancement / bugfix of the current copy feature. If you <kbd>CTRL</kbd> + <kbd>c</kbd> in a row it selects the text then uses the copy command to write to the clipboard, this is not good. I'm leveraging the Clipboard API now. The old copy doesn't support copying the selected rows, now it does and the fields are exported separated by TAB so it can directly be pasted into Excel.  I added support for pressing <kbd>ALT</kbd> + <kbd>c</kbd> to copy including the headers.

From the issues you mentioned in #1197:

> The virtualization is not supported, the whole row isn't copied:

Fixed.

> When we use CTRL+c in an editable cell, the edit mode enter, this is very wrong.

Fixed.

> In the future, we might see developers asking for: a. customizing the conversion from cell value to text, b. customizing the delimiter, asking for imperative copy APIs. We can wait

a. The CSV export is used internatelly, so value formatters are supported.
b. No, we use TAB
c. We have an imperative API: `copySelectedRowsToClipboard(includeHeaders)`

packages/grid/_modules_/grid/hooks/root/useEvents.ts
```diff
@@ -88,6 +89,13 @@ export function useEvents(apiRef: GridApiRef): void {
...
 
+  useNativeEventListener(
+    apiRef,
+    apiRef.current.rootElementRef!,
+    GRID_KEYDOWN,
+    getHandler(GRID_KEYDOWN),
+  );
+
   React.useEffect(() => {
     if (apiRef.current.rootElementRef?.current) {
```
m4theushw, line 100: This effect doesn't work. When it runs, `apiRef.current.rootElementRef?.current` is null. It's already to be removed by #1862, so a native event listener is better.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,80 @@
...
+        .map((column) => serialiseCellValue(column.headerName || column.field, '\t'))
+        .join('\t')}\r\n`;
+
+      const data = [...selectedRows.keys()].reduce<string>(
+        (acc, id) =>
+          `${acc}${serialiseRow(id, filteredColumns, apiRef.current.getCellParams, '\t').join(
+            '\t',
+          )}\r\n`,
+        includeHeaders ? headers : '',
+      );
```
oliviertassinari, line 49: Did you consider depending even more on the CSV export feature? It could be safer.

oliviertassinari (commented):
Great start, much better than what we had better 👍 

1. If the row selection is empty, should <kbd>CTRL</kbd> + <kbd>c</kbd> copy the focused cell?
2. If text is selected natively, inside a read cell, or inside an edit cell, the native <kbd>CTRL</kbd> + <kbd>c</kbd> behavior is overridden, not great.
3. I froze my laptop doing a ctrl c on the [3m cells demo](https://deploy-preview-1929--material-ui-x.netlify.app/components/data-grid/#commercial-version) :laugh:

packages/grid/_modules_/grid/hooks/features/rows/useGridParamsApi.ts
```diff
@@ -78,9 +78,17 @@ export function useGridParamsApi(apiRef: GridApiRef) {
     (id: GridRowId, field: string) => {
       const colDef = apiRef.current.getColumn(field);
       const value = apiRef.current.getCellValue(id, field);
-      const baseParams = getBaseCellParams(id, field);
+      const row = apiRef.current.getRow(id);
       const params: GridCellParams = {
-        ...baseParams,
```
m4theushw, line 83: I got an improvement from 20ms to 8ms by removing this spread operator.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,98 @@
...
+  const handleKeydown = React.useCallback(
+    (event: KeyboardEvent) => {
+      const isModifierKeyPressed = event.ctrlKey || event.metaKey || event.altKey;
+      if (event.key.toLowerCase() !== 'c' || !isModifierKeyPressed) {
+        return;
+      }
+
+      const selection = window.getSelection();
+      const anchorNode = selection?.anchorNode;
+      if (anchorNode && apiRef.current.windowRef?.current?.contains(anchorNode)) {
```
oliviertassinari, line 81: I don't think that it's working. As soon as I select and unselect something, ctrl + c do no longer work.

m4theushw: If the focus is not inside the grid it won't work. Can you check if that's the case? We can't listen to the events in the window.

oliviertassinari: I don't know, for me, I can almost, not use the feature

https://user-images.githubusercontent.com/3165635/123002156-e172c780-d3b1-11eb-81b3-3f54837042bf.mp4

m4theushw: Fixed. There's always a `Selection` object, but I was not checking if the selected string is not empty.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,101 @@
...
+      const isModifierKeyPressed = event.ctrlKey || event.metaKey || event.altKey;
+      if (event.key.toLowerCase() !== 'c' || !isModifierKeyPressed) {
+        return;
+      }
+
+      const selection = window.getSelection();
+      if (
+        selection?.anchorNode &&
+        apiRef.current.windowRef?.current?.contains(selection.anchorNode) &&
+        selection.toString().length
```
oliviertassinari, line 83: Could we avoid a false-ish comparator?

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,101 @@
...
+    (event: KeyboardEvent) => {
+      const isModifierKeyPressed = event.ctrlKey || event.metaKey || event.altKey;
+      if (event.key.toLowerCase() !== 'c' || !isModifierKeyPressed) {
+        return;
+      }
+
+      const selection = window.getSelection();
+      if (
+        selection?.anchorNode &&
+        apiRef.current.windowRef?.current?.contains(selection.anchorNode) &&
```
oliviertassinari, line 82: What's the use case for these two comparisons?

m4theushw: The currently selected element may be outside the `windowRef`. If it's inside the grid, then I let the browser to handle the CTRL + C.

We can't use `apiRef.current.windowRef?.current?.contains(selection?.anchorNode)` because ? returns undefined and this method requires `Node` or null.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,80 @@
...
+      textarea.select();
+      document.execCommand('copy');
+      apiRef.current.rootElementRef!.current!.removeChild(textarea);
+    },
+    [apiRef],
+  );
+
+  const copySelectedRowsToClipboard = React.useCallback(
+    (includeHeaders = false) => {
+      const selectedRows = apiRef.current.getSelectedRows();
```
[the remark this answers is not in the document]
dtassone: We could make it more generic 🤔 
`copyToClipboard(rowsOrCells, {includeHeaders, useFormattedValue}) `

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
oliviertassinari: OK, so we should be able to remove it.
```suggestion
```

The handler only triggers if the focus is in the grid, and the next condition don't copy if any text is already selected, not overriding the user againt copy and paste feature.

m4theushw: Yeah, it makes sense. I just removed these two lines.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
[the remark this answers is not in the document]
m4theushw: @dtassone Yes, but at this moment we don't have cell selection yet.  About the `useFormattedValue` I think that users will want to control columns individually. We could create a separate value formatter just for the clipboard. AG Grid has one.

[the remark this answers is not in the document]
oliviertassinari: +1 for making `copySelectedRowsToClipboard` undocumented (private) for now.

DanailH:
Is there a change in the actual functionality, I mean does the docs needs to be updated?

m4theushw:
> Is there a change in the actual functionality, I mean does the docs needs to be updated?

@DanailH The actual functionality was not working very well. The docs in general doesn't need updates. Only the accessibility page was updated to mention the new shortcut to copy including headers: <kbd>Alt</kbd> + <kbd>c</kbd>.

packages/grid/_modules_/grid/hooks/features/rows/useGridParamsApi.ts
oliviertassinari: Wow, it's pretty wild 👏!

On the full 100,000 rows data set, filtering goes from

Capture d’écran 2021-06-24 à 18 57 06

https://material-ui.com/components/data-grid/#commercial-version

to

Capture d’écran 2021-06-24 à 18 57 12

https://deploy-preview-1929--material-ui-x.netlify.app/components/data-grid/#commercial-version

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -0,0 +1,96 @@
...
+  const writeToClipboardPolyfill = React.useCallback(
+    (data: string) => {
+      const doc = ownerDocument(apiRef.current.rootElementRef!.current!);
+      const span = doc.createElement('span');
+      span.style.whiteSpace = 'pre';
+      span.style.userSelect = 'all';
+      span.style.opacity = '0px';
+      span.textContent = data;
+
+      apiRef.current.rootElementRef!.current!.appendChild(span);
```
oliviertassinari, line 25: is there a specific for appending in the data grid and not in the document directly? I'm asking so we could have this method be outside of the scope of the hook.

m4theushw: I never considered appending to the document, but we might have a reason to not do it. If the document doesn't have a `body` tag we can't append. I know it's a pretty absurd case, but it may happen.

oliviertassinari: It's safe, see https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/utils/getScrollbarSize.js, we never had any issue with it

m4theushw: Updated.

oliviertassinari (approved):

PR #1932, opened 2021-06-19
[test] Test vertical scrollbar

oliviertassinari:
Add a test case for https://github.com/mui-org/material-ui-x/pull/1831#issuecomment-863122707. We could have easily missed it. I have added a visual and function test base the functional test is flaky to catch the issue. The order of resolution between the two effects varies.

### DIFF
--- a/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
+++ b/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
@@ -444,9 +444,9 @@ describe('<DataGrid /> - Layout & Warnings', () => {
     // A function test counterpart of ScrollbarOverflowVerticalSnap.
     it('should not have a horizontal scrollbar if not needed', () => {
       const TestCase = () => {
-        const data = useData(100, 2);
+        const data = useData(100, 1);
         return (
-          <div style={{ width: 300, height: 300 }}>
+          <div style={{ width: 500, height: 300 }}>
             <DataGrid {...data} />
           </div>
         );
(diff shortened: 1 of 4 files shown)


oliviertassinari:
Ok, I got the screenshot test to fail as expected:

Capture d’écran 2021-06-19 à 22 20 57

The function test is not failing however, the events fire twice. It seems that triggering this problem is actually hard. AFAIK we don't have any test case for the scrollbar, so these new ones won't harm.

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -439,6 +440,22 @@ describe('<DataGrid /> - Layout & Warnings', () => {
         40 + 30 * baselineProps.rows.length,
       );
     });
+
+    // A function test counterpart of ScrollbarOverflowVerticalSnap.
+    it('should not have a horizontal scrollbar if not needed', () => {
+      const TestCase = () => {
+        const data = useData(100, 1);
```
dtassone, line 447: this is a simple case. The test should cover when we have the total column width that is within the container width - scroll width ie 
container width of 500, total column width of 495 no scroll bar if no vertical scrollbar, also we could check with a total width of 485 so no scroll bar even if there is a vertical one
Does that make sense?

oliviertassinari: It does make sense, for another PR. For now, it fixes this condition which is a step forward.

dtassone (approved):

oliviertassinari:
Ok, so now, if somebody try to apply the changes I did in #1831, the CI will fail.

PR #1949, opened 2021-06-20
[DataGrid] Fix scrollToIndexes offscreen row

oliviertassinari:
One chunk of #1818

I have updated the logic to match [the one we use](https://github.com/mui-org/material-ui/blob/2c4dc994da164a53b253f03bcd2425a52655f175/packages/material-ui/src/useAutocomplete/useAutocomplete.js#L328-L348) in the Autocomplete component. It wasn't computed correctly.

@m4theushw I have extracted it from #1871 because closing #1818 will be a larger effort. I have also simplified the test case to be virtualization implementation **invariant**.

A live demo of the bug fixed in this PR: https://material-ui.com/components/data-grid/#commercial-version

scroll

### DIFF
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -438,5 +438,25 @@ describe('<XGrid /> - Rows', () => {
         expect(virtualRowsCount).to.equal(7);
       });
     });
+
+    describe('scrollToIndexes', () => {
+      it('should scroll correctly when the given index is partially visible at the bottom', () => {
+        const headerHeight = 40;
+        const rowHeight = 50;
+        const border = 1;
+        render(
+          <TestCaseVirtualization
+            hideFooter
+            headerHeight={headerHeight}
+            height={headerHeight + 4 * rowHeight + 10 + border * 2}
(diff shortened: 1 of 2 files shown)


packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -438,5 +438,25 @@ describe('<XGrid /> - Rows', () => {
...
+            hideFooter
+            headerHeight={headerHeight}
+            height={headerHeight + 4 * rowHeight + 10 + border * 2}
+            nbCols={2}
+            rowHeight={rowHeight}
+          />,
+        );
+        const gridWindow = document.querySelector('.MuiDataGrid-window')!;
+        apiRef.current.scrollToIndexes({ rowIndex: 4, colIndex: 0 });
+        expect(gridWindow.scrollTop).to.equal(rowHeight);
```
oliviertassinari, line 458: It equals 0 without the fix, even if the selected row is only visible at 20%.

m4theushw (approved):

oliviertassinari:
I have merged but something still feels wrong about the assertion. I would expect the scroll top to be 40 pixels, the minimum required to see the focuses row, not 50 pixels.

Issue #1920, opened 2021-06-16
[DataGrid] Entering value manually in a date field is not working


davidalmas:
- [x ] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Hi, 

When I try to enter a date (eg 15/06/2021) I get an error typing the year field, it's like 1901. However, if I choose the date from the native date picker, the problem doesn't reproduce itself.

## Expected Behavior 🤔

When i type 2021 the year must be correct.

## Steps to Reproduce 🕹

You can see these materials-ui url that the error persists.

https://material-ui.com/components/data-grid/editing/

m4theushw:
It's not working because when, for example, "2" is typed into the year it creates a `Date` in 1902. I think we should only publish `GRID_CELL_EDIT_PROPS_CHANGE` when the date is "complete" (has at least 3-digit for year) and valid. Also, we could leverage `valueParser` to do the parsing: https://github.com/mui-org/material-ui-x/pull/1785#discussion_r642119534.

oliviertassinari:
What about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#two_digit_years_map_to_1900_%E2%80%93_1999?

hetanthakkar1:
@oliviertassinari I am on it!

dtassone:
ATM, we have a simple component for editing every type of column `GridEditInputCell`. I think dates can have their own component now to manage how to handle change and other behavior specific to the `date` column type. 
Then, to render the new `GridEditDateCell`, use the `renderEditCell` prop of the native date column definition `GRID_DATE_COL_DEF`.

Freundschaft:
has this actually been fixed? I still see the issue on the newest version

PR #1955, opened 2021-06-24
[DataGrid] Improve the editing API

### LINKED ISSUE #1503: [DataGrid] Focus isn't set on the input type radio
Focus isn't set on the input type radio

bug 3

### LINKED ISSUE #1824: [DataGrid] Improve API to control cell editing through custom components

## Summary 💡

While I was creating the components for #1817 I had some problems with the cell editing API. I'm listing all of them here because I think it demands work in this entire API and not just in a particular method.

1. **Inconsistencies on how to save values** - I first raised this in https://github.com/mui-org/material-ui-x/pull/1785#discussion_r642617499 but I faced the same problem in #1817. We have different components using different APIs to do the same thing. [Officially](https://material-ui.com/components/data-grid/editing/#custom-edit-component), users should use `commitCellChange` but this same method has its problems. Which API both us and developers should use to create custom edit components?

    - `GridEditInputCell` and `GridEditBooleanCell` publish the `GRID_CELL_EDIT_PROPS_CHANGE` event
    - The `ValidateServerNameGrid` [demo](https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation) calls `setEditCellProps`
    - The [demo](https://deploy-preview-1785--material-ui-x.netlify.app/components/data-grid/editing/#custom-edit-component) with the rating component calls `commitCellChange`.

2. **Calling `commitCellChange` without `setCellMode` reverts the value when clicking outside the cell** - Open [this](https://codesandbox.io/s/material-demo-forked-cpvrk?file=/demo.js) CodeSandbox, change a value in the favorite color column and click outside the cell to save. It reverts to the original value, but it shouldn't. The reason to not call `setCellMode` would be if I want to stay in the edit mode after selecting a value (e.g. a Slider).

3. **Keyboard support?** - Open [this](https://codesandbox.io/s/material-demo-forked-btjyf?file=/demo.js) CodeSandbox and select a new value for a cell of the favorite color column with the keyboard. When <kbd>Enter</kbd> is pressed it doesn't change to the view mode, even using `setCellMode`. In #1817 I had to call `setEditCellProps` when the value is selected through the keyboard.

4. **Components that use a backdrop** - Since I changed in #1772 to relay in the click, instead of blur, we now have a new problem. If a component has a backdrop, clicking "outside" it doesn't exit from the edit mode because for us the click was inside it. That being said, we should provide a new API so users can handle the outside click themselves. That means publishing `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED` and `GRID_CELL_EDIT_EXIT`.

    https://github.com/mui-org/material-ui-x/blob/b1aac5cd14d3e5bba44cdc612dcf35a78c0a50d8/packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts#L65-L71

### LINKED ISSUE #2035: DataGrid not executing `onEditCellChange`-handlers when using `singleSelect`-Cells

## Current Behavior 😯

When editing a `singleSelect` cell and commiting the change, the `onEditCellChangeComitted`- and `onEditCellChange`-handler of the DataGrid component aren't being executed.

## Expected Behavior 🤔

The `onEditCellChangeComitted`- and `onEditCellChange`-handler should get executed when editing/commiting a change in the DataGrid component

## Steps to Reproduce 🕹
Steps:

1. Go to the [Sandbox](https://codesandbox.io/s/relaxed-star-nl4qf?file=/src/App.tsx)
2. Open the Console
3. Change Name and see that the change is being logged in the console
4. Change the Birthyear and see that nothing is being logged in the console

## Context 🔦

I'm trying to do a Project Table where I can change Items from a Database.

## Your Environment 🌎

<details>
  Browser: Chrome Version 91.0.4472.124 (Official Build) (64-bit)
```
npx: installed 2 in 10.847s

  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.17.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - C:\Program Files\nodejs\yarn.CMD
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.906.0), Chromium (91.0.864.64)    
  npmPackages:
    @material-ui/data-grid: ^4.0.0-alpha.33 => 4.0.0-alpha.33 
    @material-ui/styles:  4.11.4
    @material-ui/system:  4.11.3
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @types/react: ^17.0.0 => 17.0.10
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.1.2 => 4.3.2

```
</details>

m4theushw:
### Breaking changes

- [DataGrid] The `props` key in the first argument of `commitCellChange` was removed to promote the use of the value already stored in the state. To update the value in the state, call `setEditCellProps` before.

  ```diff
  -apiRef.current.commitCellChange({ id: 1, field: 'name', props: { value: 'Ana' } });
  +apiRef.current.setEditCellProps({ id: 1, field: 'name', props: { value: 'Ana' } });
  +apiRef.current.commitCellChange({ id: 1, field: 'name' });
  ```

- [DataGrid] Calling `commitCellChange` in a cell in view mode will throw an error. Make sure to first enter the edit mode.

  ```diff
  +apiRef.current.setCellMode(1, 'name', 'edit');
  apiRef.current.commitCellChange({ id: 1, field: 'name' });
  ```

- [DataGrid] The `setCellValue` was removed from the API. Use `commitCellChange` or `updateRows` in place.

  ```diff
  -apiRef.current.setCellValue({ id: 1, field: 'name', value: 'Ana' });
  +apiRef.current.updateRows([{ id: 1, name: 'Ana' }]);
  ```

  or

  ```diff
  -apiRef.current.setCellValue({ id: 1, field: 'name', value: 'Ana' });
  +apiRef.current.setCellMode(1, 'name', 'edit');
  +apiRef.current.setEditCellProps({ id: 1, field: 'name', props: { value: 'Ana' } });
  +apiRef.current.commitCellChange({ id: 1, field: 'name' });
  ```

- [DataGrid] The `getEditCellProps` was removed because `getEditCellPropsParams` offers the same functionality.

  ```diff
  -const props = apiRef.current.getEditCellProps(1, 'name');
  +const { props } = apiRef.current.getEditCellPropsParams(1, 'name');
  ```

  **Note**: This method will now throw an error if the cell is in view mode.

---

Closes #1824 
Closes #1503
Closes #2035 

This PR contains the following improvements to the editing API:

1. All edit components now use the same pattern to save the value: `setEditCellProps` -> `commitCellChange` -> `setCellMode`.
2. `commitCellChange` will use the value in the `editRows` state, not from the params. Previously, it was possible to commit directly the value to the row, without saving it to the state. This method will throw an error if the cell is not in the edit mode, also to force the point 1.
3. The value parser is now applied in `setEditCellProps` instead of `commitCellChange`.
4. `setCellValue` was removed from the API but its behavior was merged into `commitCellChange`
5. `getEditCellProps` was removed because it was only called in `getEditCellPropsParams`.
6. Demos that use the Rating or Slider are now keyboard-accessible.

bty9sKAIPM

Preview: https://deploy-preview-1955--material-ui-x.netlify.app/components/data-grid/demo/


packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -82,32 +83,25 @@ export function useGridEditRows(apiRef: GridApiRef) {
 
   const setCellMode = React.useCallback(
     (id, field, mode: GridCellMode) => {
-      let hasChanged = false;
+      const isInEditMode = apiRef.current.getCellMode(id, field) === 'edit';
```
DanailH, line 86: Maybe it makes sense to add an `enum` for the `edit` and `view` modes?

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -258,12 +240,12 @@ export function useGridEditRows(apiRef: GridApiRef) {
       if (!params.isEditable || event.isPropagationStopped()) {
         return;
       }
+      setCellMode(params.id, params.field, 'edit');
```
DanailH, line 243:
```suggestion

      setCellMode(params.id, params.field, 'edit');
      
```

Just to make it a bit more readable 😉

oliviertassinari (commented):
Looks great, I have updated the issue description as it closes #1503 too.

packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
```diff
@@ -372,7 +372,7 @@ describe('<XGrid /> - Edit Rows', () => {
     expect(input.value).to.equal('1961');
 
     fireEvent.change(input, { target: { value: '62' } });
-    expect(cell.querySelector('input')!.value).to.equal('62');
+    expect(cell.querySelector('input')!.value).to.equal('1962');
```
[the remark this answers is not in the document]
m4theushw: How #1920 is related to this change? Why should it be "62"? In real life, If I wanted to keep the year as "62" I would also need to format the `value` prop before updating the internal state. This test case doesn't do this to allow asserting that the value parser ran.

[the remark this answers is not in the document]
oliviertassinari: I didn't notice the custom `const valueParser = stub().withArgs('62').returns(1962);`. It looks alright then.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
oliviertassinari: What do you mean?

DanailH: Not to use the string directly => 
```
enum CellEditMode {
  View = 'view',
  Edit = 'edit',
}
```
?

m4theushw: Following this line of thinking we also need to create an enum for `asc` | `desc` and other constant values we might have. I think we can discuss if it's time to do this abstraction. We have a enum for feature mode but it's used in a lot of places, so it makes sense, but `CellEditMode` would be used in only one hook. IDK, for now TypeScript already suggests the values for us.

DanailH: Not sure, I used the same approach for the density and it is used in fewer places. 

> but CellEditMode would be used in only one hook.

And when you are calling the API and passing the mode.

m4theushw: I started to add this enum but I had to change a bunch of new files that were not modified by this PR. Better to do in another PR. I still think it's a bit redundant to have the TS type and the enum, but I'm open for opinions.

DanailH (approved):

dtassone (approved):

Issue #18, opened 2020-06-10
[DataGrid] Bundle size is way too large


oliviertassinari:
The current bundle size and bundling strategy is not flying. It's not living up to the quality level our audience expects from the core components. https://bundlephobia.com/result?p=@material-ui/data-grid

## Tree-shaking

First, tree-shaking is not working. (tested with CRA). This means that, for instance, developers bundle all the locales. I think that the solution is to drop the current bundling strategy and copy what the core is doing. I think that this bundling problem is hard enough, we don't have the resources to solve the problems twice because we use two different solutions between different packages.

**in**

```jsx
import * as React from 'react';
import { GridToolbar } from '@material-ui/data-grid';

export default function DataGridDemo() {
  console.log('GridToolbar', GridToolbar)
  return null;
}
```

**out**, in the production bundle

<img width="466" alt="Screenshot 2021-04-24 at 20 16 03" src="https://user-images.githubusercontent.com/3165635/115968824-ec3be800-a539-11eb-8695-bea8c7dba7ad.png">

<img width="686" alt="Capture d’écran 2021-06-12 à 12 44 35" src="https://user-images.githubusercontent.com/3165635/121773390-f793b300-cb7b-11eb-9b3c-28b5e4adf955.png">

## Inefficiencies

Second, here is a couple of bundle size-reduction opportunities:

- [x] **clsx**. (win ~0.3 kB gzipped) Replace classnames with clsx.
- [ ] **license**. (win ~0.7 kB) I believe we can remove the need from including `packages/license/src/encoding/base64.ts` in the bundle of the users. The possible solution would work as follow:
  - client-side, use the built-in `window.atob` and `window.btoa`, only run the license check if the window is defined (ignore server-side, and CI)
  - server-side, us the built-in `Buffer.from('Hello World!').toString('base64');` and `Buffer.from(b64Encoded, 'base64').toString()` utils.
- [x] **lodash**. The dependency was inlined in the repository (I guess to hide, [on the surface](https://npm.anvaka.com/#/view/2d/%2540material-ui-x%252Fgrid), the dependency on this library), I don't think that it removes the fact that we depend on it. I would be in favor of completely removing its modules from the codebase.
  Recommended migration plan:
[...]

PR #1956, opened 2021-06-24
[DataGrid] Add single select column type

### LINKED ISSUE #1320: [DataGrid] Create select column type
String extension type that allows displaying a select options component in filters, and edit mode.

The first version can be only client-side. And should work with a memoized selector that get all distinct value of a column for better performance.

### Benchmark

-

DanailH:
Fixes #1320 

Had to add a new prop in the `ColDef` - `valueOptions` that is used to determine the select options for that column.

Preview: https://deploy-preview-1956--material-ui-x.netlify.app/components/data-grid/#commercial-version


packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+      const newValue = event.target.value;
+      const editProps = { value: newValue };
+
+      setValueState(newValue);
+      api.publishEvent(
+        GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED,
+        { id, field, props: editProps },
+        event,
+      );
+      api.setCellMode(id, field, 'view');
```
m4theushw, line 37: Selecting an item with the keyboard is not closing the select. I had to use a different approach to save the value in [`renderEditProgress`](https://github.com/mui-org/material-ui-x/blob/master/packages/grid/x-grid-data-generator/src/renderer/renderEditProgress.tsx). In #1955 I improved this API to make very clear what each methods does.

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -71,6 +71,10 @@ export interface GridColDef {
    * @default 'string'
    */
   type?: GridColType;
+  /**
+   * Used in combination with `type='select'`. Array with possible cell values.
+   */
+  valueOptions?: Array<string>;
```
m4theushw, line 77: We use more the foo[] syntax.

```suggestion
  valueOptions?: string[];
```

packages/grid/_modules_/grid/models/colDef/gridSelectOperators.ts
```diff
@@ -0,0 +1,42 @@
...
+  {
+    value: 'is',
+    getApplyFilterFn: (filterItem: GridFilterItem) => {
+      if (!filterItem.columnField || !filterItem.value || !filterItem.operatorValue) {
+        return null;
+      }
+
+      return ({ value }): boolean => {
+        return (
+          filterItem.value?.localeCompare((value && value.toString()) || '', undefined, {
```
m4theushw, line 15: Why not a simple string comparison, e.g. `filterItem.value === value`?

m4theushw:
We could add another column to the employee dataset to use this new column type. In commodity, all columns that use a select need a custom one because the options are rendered differently.

image

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
DanailH: It was a generic type initially

packages/grid/_modules_/grid/models/colDef/gridSelectOperators.ts
DanailH: This seems to be the approach for the string operators and they are the same as the select ones.

DanailH:
> We could add another column to the employee dataset to use this new column type. In commodity, all columns that use a select need a custom one because the options are rendered differently.
> 
> image

Yes I was thinking of updating the `x-data-grid-generator` but I want to push it for another PR to keep that one smaller.

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
DanailH: Yes, I saw the PR. maybe it makes sense to wait with this one until the other one is merged because I would have to change the way the cell edit mode is handled.

m4theushw: Ops, I linked the wrong component. [`renderEditStatus`](https://github.com/mui-org/material-ui-x/blob/master/packages/grid/x-grid-data-generator/src/renderer/renderEditStatus.tsx) is based on Select.

packages/grid/_modules_/grid/models/colDef/gridColType.ts
```diff
@@ -4,5 +4,5 @@ export const GRID_DATE_COLUMN_TYPE = 'date';
 export const GRID_DATETIME_COLUMN_TYPE = 'dateTime';
 export const GRID_BOOLEAN_COLUMN_TYPE = 'boolean';
 
-export type GridNativeColTypes = 'string' | 'number' | 'date' | 'dateTime' | 'boolean';
+export type GridNativeColTypes = 'string' | 'number' | 'date' | 'dateTime' | 'boolean' | 'select';
```
oliviertassinari, line 7: Are we sure about the name of the type? I'm asking because of Airtable double type:

Capture d’écran 2021-06-25 à 01 23 48

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
```diff
@@ -0,0 +1,70 @@
...
+    (event, reason) => {
+      if (reason === 'backdropClick') {
+        api.setCellMode(id, field, 'view');
+      }
+    },
+    [api, field, id],
+  );
+
+  return (
+    <Select
```
oliviertassinari, line 49: Do we want to implement this type with a dropdown list? Airtable uses an combobox even for the single select.

Another concern, the demo used in the issue is a country selector, this use case is not supported by the Select, or barely (too slow).

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -71,6 +71,10 @@ export interface GridColDef {
    * @default 'string'
    */
   type?: GridColType;
+  /**
+   * Used in combination with `type='select'`. Array with possible cell values.
```
oliviertassinari, line 75:
```suggestion
   * To be used in combination with `type='select'`. Array with possible cell values.
```

oliviertassinari (commented):
> No problem, I suggested it because I had to download the PR to test it.

Same pain here. I think that it signals that the developers can't easily discover the full scope of this feature. How about we replace `EditContractType` and `EditRateType` as to simple direct application of the new column type? It might be too basic use cases to stress test the feature, but it would already be a start.

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -71,6 +71,10 @@ export interface GridColDef {
...
+   */
+  valueOptions?: string[];
```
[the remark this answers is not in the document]
DanailH: Likewise - I wasn't sure what would have been the best approach so I made it work so we can discuss it. Having a label does make sense.

packages/grid/_modules_/grid/models/colDef/gridColType.ts
DanailH: Regarding the multiple select - we can set it up by renaming the `select` type to `singleSelect` or keeping it as it is but then the have the multiple one be called `multipleSelect`

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
DanailH: No opinion here - whatever we choose for the default these column types can be extended.

oliviertassinari: The main issue for me is the upfront bundle size we impose on the developers. It seems KO to force the bundling of a column type that the developers will never use: a Select, à DatePicker, a TimePicker, an Autocomplete, etc. Before you know it, your bundle is 100 kB gzipped. Tree shaking is not helping.

In #18, I talked about how we over bundle from the core components.

How about we make this an optional colum type developers have to manually import? And then, use the Autocomplete, not the Select?

m4theushw: Select is already included in the bundle because of the pagination. But manually importing this column type may be a start.

+1 for using Autocomplete. We would be able to also use it in the filter panel.

packages/grid/_modules_/grid/models/colDef/gridColType.ts
m4theushw: `singleSelect` would be better. `multipleSelector` is completely different (e.g. needs a "contains" operator).

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
DanailH: Idk, I think it looks a bit weird if you have autocompleted with 2 options. 

Also to be clear - @oliviertassinari by not bundling you mean that if devs want to use the `select` they will load it from the grid package and spread it in the `ColDef` for that column right?

packages/grid/_modules_/grid/models/colDef/gridColType.ts
oliviertassinari: +1 for removing potential confusion

```suggestion
export type GridNativeColTypes = 'string' | 'number' | 'date' | 'dateTime' | 'boolean' | 'singleSelect';
```

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
oliviertassinari: Honestly, it's hard to see what we should prioritize here. What seems clearer is that:

- A rich date column type should be tree-shakeable. It's not OK to add 20kB? for our date picker.
- We can use `<Select>` built-in as it will always be here for the pagination
- `<Select>` doesn't work for a country select, but we aim to fix this in the core
- If we add Autocomplete in the filter, we add 5kb gzipped for everybody. 

How about we start simple, keeping the current tradeoff of the PR?

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -71,6 +71,10 @@ export interface GridColDef {
    * @default 'string'
    */
   type?: GridColType;
+  /**
+   * To be used in combination with `type='select'`. Array with possible cell values.
```
oliviertassinari, line 75: This is not a React prop.

```suggestion
   * To be used in combination with `type: 'select'`. This is an array of the possible cell values and labels.
```

oliviertassinari (commented):
Things that seems left:

- Handle `valueOptions` when is a rich object
- Simplify Commodity data set with the new built-in type (e.g. EditContractType and EditRateType)

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -4,11 +4,25 @@ import TextField, { TextFieldProps } from '@material-ui/core/TextField';
 import { unstable_useId as useId } from '@material-ui/core/utils';
 import { GridLoadIcon } from '../../icons/index';
 import { GridFilterInputValueProps } from './GridFilterInputValueProps';
+import { GridColDef } from '../../../models/colDef/gridColDef';
+
+const renderSelectOptions = ({ valueOptions }: GridColDef) =>
+  valueOptions!.map((option) =>
```
m4theushw, line 10: We need to add an empty option or the user will open the select and see the first option "selected".  See #1961.

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
m4theushw: > How about we start simple, keeping the current tradeoff of the PR?

I would go on this direction too. We use more the native select because of the ClickAwayListener issue, once the new SelectField is released we can start using it on this column type, so for developers will be clearer and acceptable (I hope so) the increase in the bundle size.

packages/grid/_modules_/grid/models/colDef/gridSelectOperators.ts
oliviertassinari: Moved to #1968

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -17,19 +17,30 @@ describe('<DataGrid /> - Filter', () => {
...
+        country: 'Germany',
+      },
+    ],
+    columns: [
+      { field: 'brand' },
+      { field: 'isPublished', type: 'boolean' },
+      {
+        field: 'country',
+        type: 'select',
+        valueOptions: ['United States', 'Germany', 'France'],
```
dtassone, line 41: what if you want to render a template in the select? Ie flag and country?
Also I would have expected the grid to just generate the options by getting all the different values of a column without needing to provide those options.

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
DanailH: So we can have the initial version as it is.

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
DanailH: I thought about that but what if there is a value that no column has, then it is a problem of not having a way go get it.

Regarding the template - based on that discussion https://github.com/mui-org/material-ui-x/pull/1956#discussion_r658350955 we can start simple and extend it. For custom templates, the column type can be extended. I can have the same argument for the date picker and for any of the other column types.

packages/grid/_modules_/grid/components/cell/GridEditSelectCell.tsx
DanailH: Fixed

docs/pages/api-docs/data-grid/grid-col-def.md
```diff
@@ -12,30 +12,31 @@ import { GridColDef } from '@material-ui/data-grid';
...
+| <span class="prop-name optional">valueFormatter<sup><abbr title="optional">?</abbr></sup></span>    | <span class="prop-type">(params: GridValueFormatterParams) =&gt; GridCellValue</span>              |                                                  | Function that allows to apply a formatter before rendering its value.                                           |
+| <span class="prop-name optional">valueGetter<sup><abbr title="optional">?</abbr></sup></span>       | <span class="prop-type">(params: GridValueGetterParams) =&gt; GridCellValue</span>                 |                                                  | Function that allows to get a specific data instead of field to render in the cell.                             |
+| <span class="prop-name optional">valueOptions<sup><abbr title="optional">?</abbr></sup></span>      | <span class="prop-type">string \| []</span>                                                        |                                                  | To be used in combination with `type: 'singleSelect'`. This is an array of the possible cell values and labels. |
```
oliviertassinari, line 40: This is outdated.

packages/grid/x-grid-data-generator/src/commodities.columns.tsx
```diff
@@ -222,6 +234,8 @@ export const getCommodityColumns: (editable?: boolean) => any[] = (editable = fa
       headerName: 'Counterparty Country',
       generateData: randomCountry,
       renderCell: renderCountry,
+      type: 'singleSelect',
+      valueOptions: COUNTRY_ISO_OPTIONS,
```
oliviertassinari, line 238: Would it make sense?

```suggestion
      valueOptions: COUNTRY_ISO_OPTIONS,
      editable: true,
```

oliviertassinari (commented):
In the commodity data set, I can't find one column that uses the render cell of the `singleSelect` type. How about we kill `renderEditCell` in `field: 'rateType'`?

Otherwise, it looks good

packages/grid/x-grid-data-generator/src/commodities.columns.tsx
DanailH: Am that one is editable (I tested it before)

oliviertassinari: I don't understand. Are we looking at the same code?

docs/pages/api-docs/data-grid/grid-col-def.md
DanailH: It is the same as in the `gridColDef`. Maybe we can say it is `This is an array of the possible cell string values and/or labels.` ?

DanailH:
> In the commodity data set, I can't find one column that uses the render cell of the `singleSelect` type. How about we kill `renderEditCell` in `field: 'rateType'`?
> 
> Otherwise, it looks good

Done. I trashed a couple of more `renderEdit*` components. The ones that were not using the `Autocomplete` are now using the default render cell component.

docs/pages/api-docs/data-grid/grid-col-def.md
oliviertassinari: The description is already up to date. I was making a reference to the type, that is outdated, compared to the implementation:

Capture d’écran 2021-06-29 à 15 54 53

DanailH: Hmm seems to be a bug with the docs generation script. `Array<string | { value: any; label: string }>` is converted to `string \| []`

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+} from '../../constants/eventsConstants';
+import { isEscapeKey } from '../../utils/keyboardUtils';
+
+const renderSingleSelectOptions = (option) =>
+  typeof option === 'string' ? (
+    <MenuItem key={option} value={option}>
+      {option}
+    </MenuItem>
+  ) : (
+    <MenuItem key={option.value} value={option}>
```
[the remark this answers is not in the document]
DanailH: Added

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+        { id, field, props: editProps },
+        event,
+      );
+      api.publishEvent(GRID_CELL_EDIT_EXIT, { id, field }, event);
+    }
+  };
+
+  const handleClose = (event, reason) => {
+    if (reason === 'backdropClick' || isEscapeKey(event.key)) {
+      api.publishEvent(GRID_CELL_EDIT_EXIT, { id, field });
```
m4theushw, line 57: Better to use our own APIs.

```suggestion
      api.setCellMode(id, field, 'view');
```

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+    getValue,
+    hasFocus,
+    ...other
+  } = props;
+
+  const handleChange = (event) => {
+    const editProps = { value: event.target.value };
+
+    if (event.key) {
+      api.publishEvent(GRID_CELL_EDIT_PROPS_CHANGE, { id, field, props: editProps }, event);
```
m4theushw, line 44:
```suggestion
      api.setEditCellProps({ id, field, props: editProps }, event);
```

packages/storybook/src/stories/grid-columns.stories.tsx
```diff
@@ -316,3 +316,50 @@ export const ValueGetterAndFormatter = () => {
...
+      value: 'fr',
+      label: 'France',
+    },
+    {
+      value: 'it',
+      label: 'Italy',
+    },
+  ];
+
+  const [data] = React.useState({
```
m4theushw, line 340: We could remove `React.useState`.

m4theushw (commented):
There's an e2e test case added to test the support of portaled components, e.g. Select. We could update it to use this new column type. You have to render all columns in the full featured demo to see the columns that use the `singleSelect`. I saw that you updated the `maxColumns` in https://github.com/mui-org/material-ui-x/pull/1956/commits/0f4d1277ae9ba0d097c224a7a730115c48aca984 but it was reverted.

packages/grid/x-grid-data-generator/src/services/static-data.ts
```diff
@@ -77,252 +77,252 @@ export const COUNTRY_OPTIONS = [
...
-  { code: 'VU', label: 'Vanuatu', phone: '678' },
-  { code: 'WF', label: 'Wallis and Futuna', phone: '681' },
-  { code: 'WS', label: 'Samoa', phone: '685' },
-  { code: 'XK', label: 'Kosovo', phone: '383' },
-  { code: 'YE', label: 'Yemen', phone: '967' },
-  { code: 'YT', label: 'Mayotte', phone: '262' },
-  { code: 'ZA', label: 'South Africa', phone: '27' },
-  { code: 'ZM', label: 'Zambia', phone: '260' },
-  { code: 'ZW', label: 'Zimbabwe', phone: '263' },
+  { value: 'AD', code: 'AD', label: 'Andorra', phone: '376' },
```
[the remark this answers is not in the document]
DanailH: Ok. We can remove the `code` but `value` needs to remain as it is in the type.

DanailH:
> There's an e2e test case added to test the support of portaled components, e.g. Select. We could update it to use this new column type. You have to render all columns in the full featured demo to see the columns that use the `singleSelect`. I saw that you updated the `maxColumns` in [0f4d127](https://github.com/mui-org/material-ui-x/commit/0f4d1277ae9ba0d097c224a7a730115c48aca984) but it was reverted.

I can enable it back to the full column list, not a problem. Actually even with the current setup of 20 columns, you can see the new col type (for filters for example) but the render cell component is custom. I'll bump it.

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
DanailH: This is different than what we've used for now. I don't remember discussing that pattern. The API is only called directly in the data generator package.

packages/grid/x-grid-data-generator/src/services/static-data.ts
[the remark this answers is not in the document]
m4theushw: Yes, mean remove `code` and use only `value`. I updated the comment.

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
m4theushw: In #1955 I changed the other components to use the same approach. The imperative API is better because we can enforce the type of the params.

The demo about custom components is already using the API methods: https://material-ui.com/components/data-grid/editing/#custom-edit-component

docs/pages/api-docs/data-grid/grid-col-def.md
m4theushw: This will be fixed by #2003.

m4theushw (approved):
Looks good. I think you'll need to merge with master first because I merged #2003.

m4theushw:
Sorry to close it. 😁

oliviertassinari:
A nice addition, I was looking at the new GitHub project view WIP: https://twitter.com/github/status/1407731478096756739 and of course, they have a single select. Looks close to Airtable and Notion.

PR #1959, opened 2021-06-24
[DataGrid] Fix number edit cell output

oliviertassinari:
A reproduction:

1. Open https://codesandbox.io/s/material-demo-forked-3vkgp?file=/demo.js
2. Edit one cell with the age
3. Log the rows

Capture d’écran 2021-06-24 à 15 31 24

#1955 spark me the idea to fix it.


m4theushw (approved):
Looks good! For the future, we need to update `GridEditInputCell` to also leverage the value parser when the column type .is a date.

PR #1964, opened 2021-06-25
[DataGrid] Fix scrollToIndexes offscreen column

m4theushw:
While #1949 fixed the bug in the y-axis, there's a similar problem in the x-axis too. It happens because we consider a column as visible if its left position is inside the viewport, without accounting the width.

In https://material-ui.com/components/data-grid/demo/

zRZL7b9OLy

Preview: https://deploy-preview-1964--material-ui-x.netlify.app/components/data-grid/demo/

### DIFF
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -487,7 +487,7 @@ describe('<XGrid /> - Rows', () => {
         const gridWindow = document.querySelector('.MuiDataGrid-window')!;
         expect(gridWindow.scrollLeft).to.equal(0);
         apiRef.current.scrollToIndexes({ rowIndex: 0, colIndex: 2 });
-        expect(gridWindow.scrollLeft).to.equal(60);
+        expect(gridWindow.scrollLeft).to.equal(columnWidth * 3 - width);
       });
     });
   });
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -277,13 +277,18 @@ describe('<XGrid /> - Rows', () => {
 
     let apiRef: GridApiRef;
     const TestCaseVirtualization = (
-      props: Partial<XGridProps> & { nbRows?: number; nbCols?: number; height?: number },
+      props: Partial<XGridProps> & {
+        nbRows?: number;
+        nbCols?: number;
+        width?: number;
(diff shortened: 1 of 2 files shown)


packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -466,6 +471,24 @@ describe('<XGrid /> - Rows', () => {
...
+          { field: 'id', width: columnWidth },
+          { field: 'firstName', width: columnWidth },
+          { field: 'lastName', width: columnWidth },
+          { field: 'age', width: columnWidth },
+        ];
+        render(<TestCaseVirtualization width={width + border * 2} rows={rows} columns={columns} />);
+        const gridWindow = document.querySelector('.MuiDataGrid-window')!;
+        expect(gridWindow.scrollLeft).to.equal(0);
+        apiRef.current.scrollToIndexes({ rowIndex: 0, colIndex: 2 });
+        expect(gridWindow.scrollLeft).to.equal(60);
```
oliviertassinari, line 490: Correct?

```suggestion
        expect(gridWindow.scrollLeft).to.equal(columnWidth * 3 - width);
```

oliviertassinari (commented):
Here is what happens when I press <kbd>Arrow Left</kbd>, from the last column. It's 🎄. 

https://user-images.githubusercontent.com/3165635/123522289-b7c0e580-d6bc-11eb-897e-0d4604249730.mp4

I would recommend we kill the `isColumnVisibleInWindow` method and use the same strategy of #1969. We can implement a correct behavior with /2 the code. The columns have an extra later of complexity that we don't have yet on the rows: variable width, but `columnsMeta` being a cache don't make it more complex.

However, the new test case is a SOLID step forward, so I'm also happy to take it slower and move forward with what we already have.

oliviertassinari:
> While #1949 fixed the bug in the y-axis

I have pushed the solution one step forward in for the y-axis in #1969, but it's still not that great. The `.focus()` we call is messing up with the scroll, it's why we get big jumps of scroll when we <kbd>Arrow Up</kbd>.

m4theushw:
> Here is what happens when I press Arrow Left, from the last column. It's 🎄.

@oliviertassinari I'm aware of this bug, it's not good true. But to fix it we need to fix the logic of `scrollToIndexes`, which is for another PR. This PR is a quick-win without side effects.

oliviertassinari (approved):
I'm not approving for the logic (I think that we will be able to better and simpler), but for the test case that looks correct, we move forward 👍.

PR #1966, opened 2021-06-25
[DataGrid] Emit `onSelectionModelChange` when selection state changes

### LINKED ISSUE #1782: [DataGrid] Prevent the grid from cleaning up the `selectionModel` when `selectionModel` prop is provided

DanailH:
Fixes #1782


packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
```diff
@@ -138,6 +138,22 @@ describe('<XGrid /> - Selection', () => {
...
+    );
+    setProps({
+      rows: [
+        {
+          id: 1,
+          brand: 'Nike',
+        },
+      ],
+    });
+    expect(onSelectionModelChange.callCount).to.equal(2);
```
oliviertassinari, line 154: Could we also assert that the callback value is correct?

packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
```diff
@@ -138,6 +138,22 @@ describe('<XGrid /> - Selection', () => {
     expect(getSelectedRows(apiRef)).to.deep.equal([0]);
   });
 
+  it('should call onSelectionModelChange when selection state changes', () => {
+    const onSelectionModelChange = spy();
```
oliviertassinari, line 142: A small detail, I believe that we have used this convention in the codebase.

```suggestion
    const handleSelectionModelChange = spy();
```

oliviertassinari (commented):
👍 for this behavior, it also fixes the uncontrolled behavior.

packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
oliviertassinari: To leave as a comment or to assert right after the first render?

```suggestion
    // TODO should only fire once
    expect(onSelectionModelChange.callCount).to.equal(2);
```

DanailH: Just to be sure - the correct behavior in the case of server side pagination will be an empty array for the `selectionModel` because the selection state will be empty (there won't be any matching ids).
I'm asking because there is already a test that is asserting that new rows have the same ids they will remain selected - https://github.com/mui-org/material-ui-x/blob/master/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx#L127

DanailH: Updated

packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
```diff
@@ -138,6 +138,27 @@ describe('<XGrid /> - Selection', () => {
     expect(getSelectedRows(apiRef)).to.deep.equal([0]);
   });
 
+  it('should call onSelectionModelChange when selection state changes', () => {
+    const handleSelectionModelChange = spy();
+    const { setProps } = render(
+      <Test selectionModel={[0]} onSelectionModelChange={handleSelectionModelChange} />,
+    );
+    expect(getSelectedRows(apiRef)).to.deep.equal([0]);
+    expect(handleSelectionModelChange.getCall(0).args[0]).to.deep.equal({ selectionModel: [0] });
```
oliviertassinari, line 147:
```suggestion
    // TODO should not fire
    expect(handleSelectionModelChange.getCall(0).args[0]).to.deep.equal({ selectionModel: [0] });
```

oliviertassinari (commented):
Looks good

dtassone:
not sure how this fix is going to fit with my control state PR, it seems redundant, and we will have to clean these changes

DanailH:
@dtassone  It's not a fix per se but rather a missed interaction that should emit the `onSelectionModelChange`.

dtassone:
I think it will go see https://github.com/mui-org/material-ui-x/pull/1823/files#diff-87fa96c5d959b8acbc3c76f30e1750e8293f462996dd1a1e8cba8e72553d2f5a

DanailH:
Well if that is the case then why is the linked ticket assigned to me, we are essentially doing double work here...

oliviertassinari:
The PR looks solid, I would personally 👍 for a merge. One step at a time.

oliviertassinari (approved):

DanailH:
@oliviertassinari @dtassone  Ok, I applied the test to the control state PR https://github.com/mui-org/material-ui-x/pull/1823. This is what the test should look like taking into account the changes in the linked PR. The only part that is failing is the part of clearing the selection after loading rows with different IDs.

```tsx
it('should call onSelectionModelChange when selection state changes', () => {
    const handleSelectionModelChange = spy();
    const { setProps } = render(
      <div style={{ width: 300, height: 300 }}>
        <XGrid 
          columns={[{ field: 'brand' }]}
          rows={[
            {
              id: 0,
              brand: 'Nike',
            },
          ]}
          selectionModel={[0]} 
          onSelectionModelChange={handleSelectionModelChange} />
      </div>
    );
    expect(getSelectedRows(apiRef)).to.deep.equal([0]);
    setProps({
      rows: [
        {
          id: 1,
          brand: 'Nike',
        },
      ],
    });
    expect(handleSelectionModelChange.callCount).to.equal(1);
    expect(handleSelectionModelChange.getCall(0).args[0]).to.deep.equal([]);
    // TODO: Fix - it sould clear the selection because the rows have different IDs
    expect(getSelectedRows(apiRef)).to.deep.equal([]);
  });
```

I can either push this directly to the control state PR or @dtassone you can add it manually.

dtassone:
IMO we could merge #1823 and then reapply/fix this issue on top of it

oliviertassinari:
Great 👍

Issue #1036, opened 2021-02-12
[DataGrid] Unify publishing events params


dtassone:
## Summary 💡

We currently have descrepancies between all the events params. Therefore this issue is about unifying all those events params.

link in #885 

> onPageChange should have two arguments (event, page). It shouldn't include unrelated state, like pageCount, pageSize,

## Examples 🌈
`onPageChange`
`onCellClick`

## Motivation 🔦

Unify the usage of the grid.

m4theushw:
I'm closing this one because the example of `onPageChange` will be fixed once #2099 is merged. All other situations that I can imagine we kinda already solved when `useGridControlState` was implemented. We can keep #1301 to discuss the params as it has more information.

PR #1967, opened 2021-06-25
[DataGrid] Rename onColumnResizeCommitted to onColumnWidthChange

m4theushw:
### Breaking changes

- [DataGrid] Rename `onColumnResizeCommitted` to `onColumnWidthChange`.

  ```diff
  -<DataGrid onColumnResizeCommitted={...} />
  +<DataGrid onColumnWidthChange={...} />
  ```

---

Renamed to match the event name. Related to https://github.com/mui-org/material-ui-x/pull/1862#discussion_r652865464

I don't know what to do with the type in the argument, it's shared with `onColumnResize`. #1036 would touch this part.

### DIFF
--- a/docs/pages/api-docs/data-grid/data-grid.md
+++ b/docs/pages/api-docs/data-grid/data-grid.md
@@ -65,8 +65,8 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">onColumnHeaderEnter</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse enter event comes from a column header element. |
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
-| <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
-| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
+| <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired while a column is being resized. |
+| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when the width of a column is changed. |
 | <span class="prop-name">onColumnVisibilityChange</span> | <span class="prop-type">(param: GridColumnVisibilityChangeParams) => void</span> |   | Callback fired when a column visibility changes. |
 | <span class="prop-name">onError</span> | <span class="prop-type">(args: any) => void</span> |   | Callback fired when an exception is thrown in the grid, or when the `showError` API method is called. |
 | <span class="prop-name">onEditCellChange</span> | <span class="prop-type">(params: GridEditCellParams) => void</span> |   |  Callback fired when the edit cell value changed. |
--- a/docs/pages/api-docs/data-grid/data-grid.md
+++ b/docs/pages/api-docs/data-grid/data-grid.md
@@ -66,7 +66,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
-| <span class="prop-name">onColumnResizeCommitted</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
+| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
 | <span class="prop-name">onColumnVisibilityChange</span> | <span class="prop-type">(param: GridColumnVisibilityChangeParams) => void</span> |   | Callback fired when a column visibility changes. |
 | <span class="prop-name">onError</span> | <span class="prop-type">(args: any) => void</span> |   | Callback fired when an exception is thrown in the grid, or when the `showError` API method is called. |
 | <span class="prop-name">onEditCellChange</span> | <span class="prop-type">(params: GridEditCellParams) => void</span> |   |  Callback fired when the edit cell value changed. |
--- a/docs/pages/api-docs/data-grid/x-grid.md
+++ b/docs/pages/api-docs/data-grid/x-grid.md
@@ -70,7 +70,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
-| <span class="prop-name">onColumnResizeCommitted</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
+| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
 | <span class="prop-name">onColumnVisibilityChange</span> | <span class="prop-type">(param: GridColumnVisibilityChangeParams) => void</span> |   | Callback fired when a column visibility changes. |
 | <span class="prop-name">onError</span> | <span class="prop-type">(args: any) => void</span> |   | Callback fired when an exception is thrown in the grid, or when the `showError` API method is called. |
 | <span class="prop-name">onEditCellChange</span> | <span class="prop-type">(params: GridEditCellParams) => void</span> |   |  Callback fired when the edit cell value changed. |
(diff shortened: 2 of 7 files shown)


oliviertassinari (commented):
👍 For renaming, but we miss the `onColumnResizeCommitted` prop to be fired and the `GRID_COLUMN_RESIZE_COMMITTED` event.

m4theushw:
@oliviertassinari I didn't understand you. We had `GRID_COLUMN_RESIZE_COMMITTED`, then it was renamed to `GRID_COLUMN_WIDTH_CHANGE` to match the convention. Now I'm updating the prop to reflect the new name. The only event we might introduce is a debounced version: https://github.com/mui-org/material-ui-x/pull/1862#discussion_r652865464

oliviertassinari:
To be more precise on what I was thinking of:

```diff
diff --git a/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx b/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
index af8783bc..cf6e472b 100644
--- a/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
+++ b/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
@@ -105,12 +105,8 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
     clearTimeout(stopResizeEventTimeout.current);
     stopResizeEventTimeout.current = setTimeout(() => {
       apiRef.current.publishEvent(GRID_COLUMN_RESIZE_STOP);
-      apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE, {
-        element: colElementRef.current,
-        colDef: colDefRef.current,
-        api: apiRef,
-        width: colDefRef.current?.width,
-      });
+      apiRef.current.publishEvent(GRID_COLUMN_RESIZE_COMMITTED);
+      apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE);
     });

     logger.debug(
@@ -132,12 +128,8 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
     newWidth = Math.max(MIN_COL_WIDTH, newWidth);

     updateWidth(newWidth);
-    apiRef.current.publishEvent(GRID_COLUMN_RESIZE, {
-      element: colElementRef.current,
-      colDef: colDefRef.current,
-      api: apiRef,
-      width: newWidth,
-    });
+    apiRef.current.publishEvent(GRID_COLUMN_RESIZE);
+    apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE);
   });

   const handleColumnResizeMouseDown = useEventCallback(
@@ -201,6 +193,8 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
     clearTimeout(stopResizeEventTimeout.current);
     stopResizeEventTimeout.current = setTimeout(() => {
       apiRef.current.publishEvent(GRID_COLUMN_RESIZE_STOP);
+      apiRef.current.publishEvent(GRID_COLUMN_RESIZE_COMMITTED);
+      apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE);
     });

     logger.debug(
@@ -227,12 +221,8 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
     newWidth = Math.max(MIN_COL_WIDTH, newWidth);

     updateWidth(newWidth);
-    apiRef.current.publishEvent(GRID_COLUMN_RESIZE, {
-      element: colElementRef.current,
-      colDef: colDefRef.current,
-      api: apiRef,
-      width: newWidth,
-    });
+    apiRef.current.publishEvent(GRID_COLUMN_RESIZE);
+    apiRef.current.publishEvent(GRID_COLUMN_WIDTH_CHANGE);
   });

   const handleTouchStart = useEventCallback((event) => {
@@ -332,5 +322,6 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
   useGridApiEventHandler(apiRef, GRID_COLUMN_RESIZE_STOP, handleResizeStop);

   useGridApiOptionHandler(apiRef, GRID_COLUMN_RESIZE, options.onColumnResize);
+  useGridApiOptionHandler(apiRef, GRID_COLUMN_RESIZE_COMMITTED, options.onColumnResizeCommitted);
   useGridApiOptionHandler(apiRef, GRID_COLUMN_WIDTH_CHANGE, options.onColumnWidthChange);
 };
```

m4theushw:
GRID_COLUMN_RESIZE_COMMIT[TED] would only be published if the user resizes the column, not when the column is resized because the container was resized. I think that if developers want to persist the width of the columns they also want to do it when the container changes. To do that we need to publish this event when the container resize stops, which is the same of creating a debounced GRID_COLUMN_WIDTH_CHANGE. Shouldn't we take this direction (debounce)?

oliviertassinari:
> GRID_COLUMN_RESIZE_COMMIT[TED] would only be published if the user resizes the column, not when the column is resized because the container was resized.

@m4theushw Agree, the sketch diffs tried to implement this behavior. 

> I think that if developers want to persist the width of the columns they also want to do it when the container changes.

This one is tough. My thought process was: developers resize the column in order to be able to better see them this seems closer to what is relevant between two sessions. Persisting the final size of the column seems too much depending on the browser's window, one that is less guaranteed to stay stable.

m4theushw:
> GRID_COLUMN_RESIZE: when the user resizes the column. Use case: controlling the resizing to implement a max-width

@oliviertassinari If we update GRID_COLUMN_WIDTH_CHANGE to be fired when the width of a column is changed for any reason then there's no need need for this event, nether the onColumnResize prop.

> GRID_COLUMN_RESIZE_COMMITTED: when the user persists the column he is resizing. Use case: storing the resized columns in the DB

The developer could listen to GRID_COLUMN_RESIZE_STOP for this use case.

---

What do you think about pushing forward this PR? It only renames a prop to match its event. I'll open another issue so we can discuss the different resize events. We could use `apiRef.current.setColumnWidth` to update the width during resize. Also, I just noticed that the resizing in touch devices is not working (already know why). 🙊

docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -66,7 +66,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
-| <span class="prop-name">onColumnResizeCommitted</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
+| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
```
flaviendelangle, line 69: I feel like the description of `onColumnResize` and `onColumnWidthChange` could be clearer.
To better indicate that `onColumnResize` is triggered multiple times during the resize process and `onColumnWidthChange` only at the end.

m4theushw: Updated.

oliviertassinari: @m4theushw We probably want to update the markdown too?

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -70,7 +70,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
-| <span class="prop-name">onColumnResizeCommitted</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
+| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resized. |
```
oliviertassinari, line 73:
```suggestion
| <span class="prop-name">onColumnWidthChange</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when the width of a column is changed. |
```

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -70,7 +70,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
```
oliviertassinari, line 72:
```suggestion
| <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired while a column is being resized. |
```

docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -66,7 +66,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">onColumnHeaderLeave</span> | <span class="prop-type">(param: GridColumnHeaderParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a column header element. |
 | <span class="prop-name">onColumnOrderChange</span> | <span class="prop-type">(param: GridColumnOrderChangeParams, event: React.MouseEvent) => void</span> |   | Callback fired when a column is reordered. |
 | <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired when a column is resizing. |
```
oliviertassinari, line 68:
```suggestion
| <span class="prop-name">onColumnResize</span> | <span class="prop-type">(param: GridColumnResizeParams) => void</span> |   | Callback fired while a column is being resized. |
```

docs/pages/api-docs/data-grid/data-grid.md
m4theushw: @oliviertassinari Yeah, I missed the markdown. Now it's fixed. I also updated https://deploy-preview-1967--material-ui-x.netlify.app/components/data-grid/columns/#resizing to explain better these two callbacks.

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -84,7 +84,10 @@ Alternatively, to disable all columns resize, set the prop `disableColumnResize=
 
 {{"demo": "pages/components/data-grid/columns/ColumnSizingGrid.js", "disableAd": true, "bg": "inline"}}
 
-To capture changes of a column width during resizing use `onColumnResize` or `onColumnResizeCommitted`.
+To capture changes in the width of a column there are two callbacks that are called:
+
+- `onColumnResize`: Called while a column is being resized.
+- `onColumnWidthChange`: Called after the width of a column is changed, but not during resizing.
```
oliviertassinari, line 90: The "but not during resizing" feels off. As a developer, I think that I would expect it to be always a one-on-one binding. Great to have it documented nevertheless.

oliviertassinari (approved):

flaviendelangle (approved):

Issue #1911, opened 2021-06-15
[DataGrid] Virtualization not respecting scrollTop


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

End-users can feel any difference between virtualization enable and disabled

## Expected Behavior 🤔

End-users shouldn't feel any difference between virtualization enable and disabled

## Steps to Reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/material-uix-gridscroll-to-indexes-demo-simple-forked-on0dc
2. See how the scroll-top is **discontinue** disorienting end-users.
![scrollTop](https://user-images.githubusercontent.com/3165635/122136528-f9db6300-ce42-11eb-8f4c-37e43005ff17.gif)

You can also see there is a blank space at the bottom, but it's #414.

## Context 🔦

Raised by @m4theushw in #1871.

>  I found that the amount of scroll needed to make a row visible is less than what I would normally expect (rowIndex * rowHeight)

It's KO. It would only make sense for this use case: https://ag-grid.com/react-grid/massive-row-count/.

I suspect the best course of action is to rewrite the virtualization logic. Taking the following constraints into account in the design.

- Built-in support for no virtualization #1781
- Variable row height #417
- Dynamic row height #438
- Performance #1903
- Research into overscan #270

oliviertassinari:
Looking at the reproduction, we are completely off. We have 34 rows of 50px height with a viewport of 290px height. So the max scroll top should be = 34 * 50 - 290 = 1410px. Instead, it's 1250px.

m4theushw:
> Looking at the reproduction, we are completely off. We have 34 rows of 50px height with a viewport of 290px height. So the max scroll top should be = 34 * 50 - 290 = 1410px. Instead, it's 1250px 🙃.

Yeah, when I was playing with the virtualization to understand it I noticed that we use a different calculation for the height of the container. If we do the right thing the user will have to scroll more, but there will be no discontinuation when reaching the top.

[...]

PR #1969, opened 2021-06-26
[DataGrid] Better fix the scrollToIndexes logic

oliviertassinari:
This is a follow-up https://github.com/mui-org/material-ui-x/pull/1949#issuecomment-866995274, I'm pushing the quality of the solution one step further, by doing a plain copy & paste of the `Autocomplete`'s logic.

The previous behavior: https://codesandbox.io/s/material-demo-forked-j7i1x?file=/package.json.

https://user-images.githubusercontent.com/3165635/123521590-e8525080-d6b7-11eb-94db-3e259e36a549.mp4

The new behavior: https://codesandbox.io/s/material-demo-forked-nzkwd?file=/package.json

https://user-images.githubusercontent.com/3165635/123521674-63b40200-d6b8-11eb-8e7d-8e945eacefd8.mp4

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
@@ -217,33 +217,21 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
         }
       }
 
-      let isRowIndexAbove = false;
-      let isRowIndexBelow = false;
-
       if (params.rowIndex != null) {
         const elementIndex = !options.pagination
           ? params.rowIndex
           : params.rowIndex - paginationState.page * paginationState.pageSize;
 
-        const currentRowPage = elementIndex / gridState.containerSizes!.viewportPageSize;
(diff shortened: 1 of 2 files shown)


dtassone:
it doesn't really work on https://deploy-preview-1969--material-ui-x.netlify.app/components/data-grid/
Try keeping the arrow down
then up, 
then go down with the scroll thumb quite a bit 
and repeat the keyboard arrows test

oliviertassinari:
@dtassone True, now, what's the root cause? Is it the issue from the changes or from the rendering logic. AFAIK, it's surfacing that the rendering logic is not correct, which is a step forward (enforced by the test case).

m4theushw (approved):
It's not a perfect solution but it's better than what we have. The new logic still doesn't fix #1818 but I can improve it in #1871. In my tests, when I call `scrollToindexes` with a rowIndex much lower than the one that is currently displayed it doesn't scroll.

oliviertassinari:
@m4theushw Yes, one correct test case at a time :). Now, I think that we get the fix of #1818 for free by solving #1911.

PR #1720, opened 2021-05-20
[DataGrid] Exploring Plugin api

dtassone:
Plugin api exploration

Fix filters in a separate PR.
- extract filter changes (fix nb calls to onChange function)
- fix control api on filters

PR #1989, opened 2021-06-28
[XGrid] Close column header menu when resizing column

### LINKED ISSUE #1931: [XGrid] Resizing a column does not close the menu

## Current Behavior 😯

r2f5ywyz62

## Expected Behavior 🤔

It should close any menu as soon as the user starts resizing a column.

## Steps to Reproduce 🕹

Steps:

1. Go to https://material-ui.com/components/data-grid/demo/
2. Open the menu of a column
3. Resize another column

flaviendelangle:
Closes #1931 

Preview: https://deploy-preview-1989--material-ui-x.netlify.app/components/data-grid/columns/#resizing


flaviendelangle:
I did not find tests about this menu so I created a new file
But we may want to test the basic features of this menu and not just the behavior modified in this PR

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      const menuIcon = getColumnHeaderCell(0).querySelector('.MuiDataGrid-menuIcon');
+
+      fireEvent.click(menuIcon!.querySelector('.MuiDataGrid-menuIconButton'));
+
+      expect(menuIcon).to.have.class('MuiDataGrid-menuOpen');
+
+      const separator = getColumnHeaderCell(0).querySelector('.MuiDataGrid-iconSeparator');
+      fireEvent.mouseDown(separator);
+      await raf();
```
oliviertassinari, line 63: Do we need the raf and mouse up events to reproduce the bug in the test case?

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      fireEvent.mouseUp(separator);
+
+      expect(menuIcon).not.to.have.class('MuiDataGrid-menuOpen');
```
oliviertassinari, line 66: Is there an aria attribute we can use instead of the classes API? Like aria-expanded.

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+            ]}
+          />
+        </div>,
+      );
+
+      expect(document.querySelectorAll('.MuiGridMenu-root')).to.have.length(0);
+
+      const menuIcon = getColumnHeaderCell(0).querySelector('.MuiDataGrid-menuIcon');
+
+      fireEvent.click(menuIcon!.querySelector('.MuiDataGrid-menuIconButton'));
```
oliviertassinari, line 57: If we could query with role + label, it would be better.

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
flaviendelangle: If I do not do the `mouseUp` event, then the next bug is failing
But maybe there is just a cleanup that needs to be done between the two tests.

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
flaviendelangle: There is a `aria-expanded` indeed, I switch :+1:

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
flaviendelangle: https://app.circleci.com/pipelines/github/mui-org/material-ui-x/5588/workflows/9535d3d3-6d3b-45e7-9aff-0581fab8113a/jobs/28784?invite=true#step-111-229

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
flaviendelangle:
```suggestion
      fireEvent.click(menuIcon!.querySelector('.MuiDataGrid-menuIconButton'));
```

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
flaviendelangle: To give a more recent example :

Here is the CI with the RAF / mouseUp : https://app.circleci.com/pipelines/github/mui-org/material-ui-x/5630/workflows/73b90c24-27b9-4a06-949c-00477018a3e8

Here is the CI without the RAF / mouseUp : https://app.circleci.com/pipelines/github/mui-org/material-ui-x/5632/workflows/65bf1f80-87bf-4341-a45b-4515eeec074d

With only this commit between the two : https://github.com/mui-org/material-ui-x/pull/1989/commits/034d088c4835cddf11e13c1cb9ce936c2f5d7113

I tried with just the mouseUp and not the Raf and the tests are passing if you prefer.

flaviendelangle:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

There is probably a missing cleanup somewhere when a parent of this icon is unmounted while being mousedown

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/1989/commits/351dc207be5931e8aa2e774943c87c3540669d38

This should fix this error

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
```diff
@@ -295,6 +295,12 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
       setGridState((oldState) => ({
         ...oldState,
         columnResize: { ...oldState.columnResize, resizingColumnField: field },
+        columnMenu: oldState.columnMenu.open
+          ? {
+              ...oldState,
```
dtassone, line 300: `...oldState.columnMenu`

dtassone (changes requested):

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
```diff
@@ -36,8 +36,8 @@ describe('<XGrid /> - Column Headers', () => {
     ],
   };
 
-  describe('GridColumnHeaderMenu', () => {
-    it('should close menu when resizing a column', async () => {
+  describe.only('GridColumnHeaderMenu', () => {
```
dtassone: don't forget to remove the `.only`

flaviendelangle: I just removed it
I did not know this feature, it's great but I'm forgetting to remove it all the time :grimacing:

oliviertassinari: eslint will fail the build if it's still around

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
oliviertassinari: Ahhh ok, it's a regression from #1720. It also explains why column resizing is broken since 4.0.0-alpha.31.

packages/grid/x-grid/src/tests/columnHeaders.DataGrid.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      );
+
+      const columnCell = getColumnHeaderCell(0);
+
+      const menuIconButton = columnCell.querySelector('button[aria-label="Menu"]');
+
+      fireEvent.click(menuIconButton);
+      expect(menuIconButton!.getAttribute('aria-expanded')).to.equal('true');
+
+      const separator = columnCell.querySelector('.MuiDataGrid-iconSeparator');
```
m4theushw, line 60: We could use `GRID_COLUMN_HEADER_SEPARATOR_RESIZABLE_CSS_CLASS` here.

dtassone (approved):

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
```diff
@@ -327,6 +327,13 @@ export const useGridColumnResize = (apiRef: GridApiRef) => {
     };
   }, [apiRef, handleTouchStart, stopListening]);
 
+  React.useEffect(() => {
+    // TODO properly handle listener cleanup on grid unmount
```
dtassone, line 331: Run tests without this effect

flaviendelangle: If I remove the effect, I have the same error as described here :
https://github.com/mui-org/material-ui-x/pull/1989#discussion_r661220534

flaviendelangle: @dtassone do you think we better merge this PR as it is or improve the behavior of `useGridColumnResize` to avoid this non-conditional cleanup on unmount ?

dtassone: IMO we can do it in a separate PR, and we could capture the refactoring in a separate issue.

packages/grid/x-grid/src/tests/columnHeaders.XGrid.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+      const columnCell = getColumnHeaderCell(0);
+
+      const menuIconButton = columnCell.querySelector('button[aria-label="Menu"]');
+
+      fireEvent.click(menuIconButton);
+      await waitFor(() => expect(screen.queryByRole('menu')).not.to.equal(null));
+
+      const separator = columnCell.querySelector('.MuiDataGrid-iconSeparator');
+      fireEvent.mouseDown(separator);
+      await waitFor(() => expect(screen.queryByRole('menu')).to.equal(null));
```
m4theushw, line 66: Trigger a `mouseup` to remove the added listeners. Then, you'll be able remove that `useEffect`. This comment applies to the other test case too.

```suggestion
      await waitFor(() => expect(screen.queryByRole('menu')).to.equal(null));
      fireEvent.mouseUp(separator);
```

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/1989#discussion_r661042972

It does work, but @oliviertassinari you not a huge fan

m4theushw: We can leave a TODO note to remove the `mouseup` later. I think that changing the code to only make this test to pass is worse. In a real world scenario, we'll have the mouse up after the mouse down so the listeners will be removed.

flaviendelangle: OK fine for me, I'll add the mouseup + TODO :+1:

oliviertassinari: The mouse up do look better. It better isolate the origin of the issue.

packages/grid/x-grid/src/tests/columnHeaders.XGrid.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+      const columnCell = getColumnHeaderCell(0);
+
+      const menuIconButton = columnCell.querySelector('button[aria-label="Menu"]');
+
+      fireEvent.click(menuIconButton);
+      await waitFor(() => expect(screen.queryByRole('menu')).not.to.equal(null));
+
+      const separator = columnCell.querySelector('.MuiDataGrid-iconSeparator');
+      fireEvent.mouseDown(separator);
+      // TODO remove mouseUp once useGridColumnReorder will handle cleanup properly
```
oliviertassinari, line 66: We will likely get this for free with a fix of #2007.

oliviertassinari (approved):

packages/grid/x-grid/src/tests/columnHeaders.XGrid.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </div>,
+      );
+
+      const columnWithMenuCell = getColumnHeaderCell(0);
+      const columnToResizeCell = getColumnHeaderCell(1);
+
+      const menuIconButton = columnWithMenuCell.querySelector('button[aria-label="Menu"]');
+
+      fireEvent.click(menuIconButton);
+      await waitFor(() => expect(screen.queryByRole('menu')).not.to.equal(null));
```
oliviertassinari, line 90: Ideally, we wouldn't need for an away, but it's not obvious how we could make it happen. Maybe we would need to disable the transition in the render factory helper, or to mock the time. I'm raising the point because await often lead to slower and less reliable tests. Anyway, not a blocker.

flaviendelangle: I'm also in favor of fixing this behavior in a later PR to avoid blocking this one

PR #1995, opened 2021-06-29
[DataGrid] Remove focus on cell when its row is removed from the data

### LINKED ISSUE #1958: [XGrid] After updating row props with an array where item was deleted, data grid stops working

## Current Behavior 😯

After deleting row from the data source and providing it to XGrid via prop `rows` XGrid stops working correctly.

(The workaround with unmounting the grid after the data source changes and re-create it with the updated rows is unfortunately not acceptable solution since it causes unwanted flickering.)

## Expected Behavior 🤔

XGrid works correctly after changing the rows array without the need to unmount and mount the whole grid again.

## Steps to Reproduce 🕹

Example usage is on https://codesandbox.io/s/delicate-snowflake-1llbd.

Steps:

1. click the menu button in the last column of the table on any row
2. click 'Remove'
3. the row is deleted from the data source and it also disappears from the grid
4. click the menu button in the last column of the table on any row (or perform any other action on the grid like selecting a row)
5. exception is thrown

image

## Context 🔦

I would like to use the grid in a way that it dynamically renders the rows passed in props even after changes to the array (deleting, adding...). If there is any way that doesn't include complete removal of the grid component from the DOM and its re-creation, I would appreciate it.

## Your Environment 🌎

<details>
  
```
Chrome browser was used.

 System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 7.3.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 91.0.4472.114
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.48)
  npmPackages:
    @emotion/styled:  10.0.27
    @material-ui/styles:  4.11.4
    @material-ui/system:  4.11.3
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @material-ui/x-grid: ^4.0.0-alpha.32 => 4.0.0-alpha.32
    @material-ui/x-license:  4.0.0-alpha.32
    @types/react: 17.0.3 => 17.0.3
    react: 17.0.2 => 17.0.2
    react-dom: 17.0.2 => 17.0.2
    styled-components: 5.2.1 => 5.2.1
    typescript: ~4.2.4 => 4.2.4
```
</details>

## Order id 💳

Will be provided once we will buy the license which we don't have yet.

flaviendelangle:
Fixes #1958

### DIFF
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -1,8 +1,12 @@
 import * as React from 'react';
-import { createClientRenderStrictMode, fireEvent } from 'test/utils';
+import {
+  createClientRenderStrictMode,
+  // @ts-expect-error need to migrate helpers to TypeScript
+  fireEvent,
+} from 'test/utils';
 import { useFakeTimers } from 'sinon';
 import { expect } from 'chai';
-import { getCell, getColumnValues, raf } from 'test/utils/helperFn';
+import { getCell, getColumnValues } from 'test/utils/helperFn';
 import {
   GridApiRef,
   GridComponentProps,
--- a/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import { createClientRenderStrictMode } from 'test/utils';
+import { createClientRenderStrictMode, fireEvent } from 'test/utils';
 import { useFakeTimers } from 'sinon';
 import { expect } from 'chai';
-import { getCell, getColumnValues } from 'test/utils/helperFn';
+import { getCell, getColumnValues, raf } from 'test/utils/helperFn';
 import {
   GridApiRef,
   GridComponentProps,
@@ -30,16 +30,19 @@ describe('<XGrid /> - Rows', () => {
         autoHeight: isJSDOM,
         rows: [
           {
+            id: 0,
             clientId: 'c1',
             first: 'Mike',
             age: 11,
           },
           {
+            id: 1,
             clientId: 'c2',
             first: 'Jack',
             age: 11,
           },
           {
+            id: 2,
             clientId: 'c3',
             first: 'Mike',
             age: 20,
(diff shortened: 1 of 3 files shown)


flaviendelangle:
The issue occurred when removing a row that had a cell focused when removing it.
The next time we clicked on the document, it published the event `GRID_CELL_FOCUS_OUT` which tried to pass the value of the cell but unsucessfully.

I'm not sure what the best solution here is.
I went for a simple `useEffect` that checks if the focused cell has been removed whenever the `rows` prop changes. It solves the issue but may not be the best solution.


Another issue here is that the typing of `getCellValue` assure that is does return a value. But if we pass an invalid `id`, it returns `undefined`. I think that in a development environment it should throw an explicit error instead of letting the rest of the code crash with errors like `cannot access X of undefined`.

```
getCellValue: (id: GridRowId, field: string) => GridCellValue;
```

Of course this would require a new test once a solution is decided.

dtassone:
> The issue occurred when removing a row that had a cell focused when removing it.
> The next time we clicked on the document, it published the event `GRID_CELL_FOCUS_OUT` which tried to pass the value of the cell but unsucessfully.
> 
> I'm not sure what the best solution here is.
> I went for a simple `useEffect` that checks if the focused cell has been removed whenever the `rows` prop changes, which solves the issue.
> 
> Of course this would require a new test once a solution is decided.
> 
> Another issue here is that the typing of `getCellValue` assures that is does return a value. But if we pass an invalid `id`, it returns `undefined`. I think that in a development environment it should throw an explicit error instead of letting the rest of the code crash with errors like `cannot access X of undefined`.
> 
> ```ts
> getRow: (id: GridRowId) => GridRowModel;
> ```

The solution looks fine.
I don't think we should throw an error if the id is invalid, I think it should just return null and we should update the type.
One could check if a row is in in the grid using apiRef and getRow

flaviendelangle:
OK, I'll add some test for the current behavior change
And for the type change I'll do another PR after

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -468,4 +472,74 @@ describe('<XGrid /> - Rows', () => {
...
+      render(<TestCase rows={baselineProps.rows} />);
+
+      fireEvent.click(getCell(0, 0));
+      expect(apiRef.current.getState().focus.cell).to.deep.equal({
+        id: baselineProps.rows[0].id,
+        field: baselineProps.columns[0].field,
+      });
+    });
+
+    it('should reset focus when removing the row containing the focus cell', async () => {
```
oliviertassinari, line 526:
```suggestion
    it('should reset focus when removing the row containing the focus cell', () => {
```

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -468,4 +472,74 @@ describe('<XGrid /> - Rows', () => {
...
+      const { setProps } = render(<TestCase rows={baselineProps.rows} />);
+
+      fireEvent.click(getCell(0, 0));
+      setProps({ rows: baselineProps.rows.slice(1) });
+      expect(apiRef.current.getState().focus.cell).to.equal(null);
+    });
+
+    it('should not reset focus when removing a row not containing the focus cell', async () => {
```
oliviertassinari, line 534:
```suggestion
    it('should not reset focus when removing a row not containing the focus cell', () => {
```

oliviertassinari (commented):
The bug is fixed: https://codesandbox.io/s/zen-darkness-fyjex 👍. I wonder about:

1. Would we have the same issue with a removed column? Maybe we shouldn't are yet.
2. Is the cleanup done soon enough? It seems to, it runs in the same effect timing than `convertGridRowsPropToState`.

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
flaviendelangle: 1. I think that there is a broader subject than just row removal here indeed. Could there be the same behavior for other things than just focus ?

2. I also think so.

dtassone (approved):

magrinj:
@flaviendelangle Hey ! Thanks for the fix, when this one will be published in an npm release ?

oliviertassinari:
@magrinj Next up on our release round-robin is Damien. The last release was 2 weeks ago, so I assume it will happen soon or later.

Issue #1815, opened 2021-06-02
[DataGrid] Setting page with ApiRef in useEffect does not work anymore


dtassone:
Broken story
https://material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-pagination--page-2-api

Expected to render rows 51-100 of 2000

Instead it renders rows 1-50 of 2000

TODO add as snap test

oliviertassinari:
@dtassone When you open an issue, please always follow the issue template. For instance, we miss a codesandbox and the version.

oliviertassinari:
Regarding the change of behavior, it originates from #1583 (not from the refactoring part, from the fix of #1517). As a side note. I should have moved the refactoring to a different PR, this is a great example of what I did [sucks](https://github.com/mui-org/material-ui-x/pull/1583#issuecomment-834750710).

flaviendelangle:
The `useEffect` in the story component occurs just after the `useEffect` in `useData` is triggered.
So the story component has not yet rerendered with the data.

If I add a check on the rows length in the `useEffect` in the story component, the `setPage` occurs after the story component has rerenderer with the data

```ts
  React.useEffect(() => {
    if (data.rows.length) {
      apiRef.current.setPage(1);
    }
  }, [data, apiRef]);
```

But, it's still to soon because it occurs before the `useEffect` in `useGridPagination` has updated the `pagination` state with `rowCount` > 0.

For me the real issue here is that we have async cascade state update, so the `useEffect` in the story component can be called with an intermediate state.

We should trigger the `setGridState` currently in the [useEffect in useGridPagination](https://github.com/mui-org/material-ui-x/blob/1210f1a7cbefff1ad30e6f326db6925975c33c80/packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts#L94) synchronously whenever the `visibleRowCount` changes, not in a `useEffect`.

flaviendelangle:
Relates to #2044 (check after refactoring the pagination with the new controlled state pattern)

flaviendelangle:
I confirm that #2099 fixes this one

PR #1996, opened 2021-06-29
[core] Remove unused event 'cellFocusChange'

flaviendelangle:
What is the use case of this event ? It is never mentioned in an issue and never used internally.
We already have a `GRID_CELL_FOCUS` which gives access to the cell being focused whereas this one just says "the focused cell has changed".

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
+++ b/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
@@ -1,13 +1,14 @@
 import * as React from 'react';
 import { ownerDocument } from '@material-ui/core/utils';
 import {
-    GRID_CELL_CLICK,
-    GRID_CELL_DOUBLE_CLICK,
-    GRID_CELL_MOUSE_UP,
-    GRID_CELL_FOCUS_OUT,
-    GRID_COLUMN_HEADER_BLUR,
-    GRID_COLUMN_HEADER_FOCUS,
-    GRID_CELL_MODE_CHANGE, GRID_CELL_FOCUS_CHANGE,
+  GRID_CELL_CLICK,
+  GRID_CELL_DOUBLE_CLICK,
+  GRID_CELL_MOUSE_UP,
+  GRID_CELL_FOCUS_OUT,
+  GRID_COLUMN_HEADER_BLUR,
+  GRID_COLUMN_HEADER_FOCUS,
+  GRID_CELL_MODE_CHANGE,
+  GRID_CELL_FOCUS_CHANGE,
 } from '../../../constants/eventsConstants';
 import { GridApiRef } from '../../../models/api/gridApiRef';
 import { GridFocusApi } from '../../../models/api/gridFocusApi';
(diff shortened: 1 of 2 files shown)


m4theushw:
I would go further and remove it. Nobody subscribes to this event. We only use the `GRID_CELL_FOCUS_OUT` for the row editing feature.

flaviendelangle:
PR updated to remove it :+1:

m4theushw:
Looking at the blame, I think it was added to match the convention of when the state is updated, publish an event. The use case for this event is not clear. It's also fired for when a column header receives focus, which is odd for an event containing "cell".

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -33,11 +33,9 @@ export const useGridFocus = (apiRef: GridApiRef): void => {
           focus: { cell: { id, field }, columnHeader: null },
         };
       });
-      // TODO replace with constant
-      apiRef.current.publishEvent('cellFocusChange');
```
dtassone, line 37: It should have been a constant, probably something that was missed in a review. 
The idea was to publish an event when the grid focus state changes so ppl can subscribe to state change. Now you can use 'cellFocus' to know which cell will be next to get the focus but the grid state won't have changed at that stage. So it's slightly different

oliviertassinari (approved):

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
oliviertassinari: > The idea was to publish an event when the grid focus state changes so ppl can subscribe to state change.

This seems to resonate with the potential confusion about what these events are. This is not clear to me. Are the events coming from user events, regardless of if the data grid commits them or not. For instance, a GRID_PAGE_CHANGE fires as long as the user clicks to change the current state, and regardless of if the outcome changing the state or not. Or does it trigger once a state change is persisted? 

Here, seems `cellFocusChange` only fire 50% of the time, it seems to be more of a legacy logic we broke along the way, but for which nobody complained about. So I would 👍 for removing it and see if developers miss it.

m4theushw: For these events that are fired after a state update, we could add a reason to `setState` and developers can listen to GRID_STATE_CHANGE. We could go further and extend the new useControlState hook to allow to subscribe to updates in just a part of the state.

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
flaviendelangle: Being able to listen to state changes would help making sure that code like #1815 only runs when the Grid is ready enough.

But on the other hand, I think this pattern could easily be overused. A lot of events would have their equivalent with `GRID_STATE_CHANGE`.

oliviertassinari: > We could go further and extend the new useControlState hook to allow to subscribe to updates in just a part of the state.

@m4theushw It sounds close to the point 2. of https://github.com/mui-org/material-ui-x/issues/1507#issuecomment-828485605.

@flaviendelangle Agree, I think that the event should fire to signal that the action was initiated, not to signal that it has been persisted, usually, developers don't need to know. For instance, we could fire the event on a page change, event if the state is controlled and the change, was therefore not committed. So it's really to signal the end-user intent, not to signal the outcome.

PR #1998, opened 2021-06-30
[XGrid] Add option to select only visible rows on the current page

### LINKED ISSUE #605: [DataGrid] Allow "select all" to only select the active page

DanailH:
Fixes #605 

Added `checkboxSelectionVisibleOnly` that when `true`, Select All will only select rows visible on the current paginated page.

ToDo:

- [x] Docs
- [x] Tests

### DIFF
--- a/docs/pages/api-docs/data-grid/x-grid.md
+++ b/docs/pages/api-docs/data-grid/x-grid.md
@@ -21,6 +21,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">autoHeight</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid height is dynamic and follow the number of rows in the grid. |
 | <span class="prop-name">autoPageSize</span> | <span class="prop-type">boolean</span> | false | If `true`, the pageSize is calculated according to the container size and the max number of rows to avoid rendering a vertical scroll bar. |
 | <span class="prop-name">checkboxSelection</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid get a first column with a checkbox that allows to select rows. |
+| <span class="prop-name">checkboxSelectionVisibleOnly</span> | <span class="prop-type">boolean</span> | false | To be used in combination with `checkboxSelection`. If `true`, `Select All` selects only the rows on the current page. |
 | <span class="prop-name">classes</span> | <span class="prop-type">GridClasses</span> |   | Override or extend the styles applied to the component. See [CSS API](/api/data-grid/x-grid/#css) below for more details. |
 | <span class="prop-name">className</span> | <span class="prop-type">string</span> |   | CSS classname to add on the outer container. |
 | <span class="prop-name">columnBuffer</span> | <span class="prop-type">number</span> | 2 | Number of columns rendered outside the grid viewport. |
--- a/docs/pages/api-docs/data-grid/x-grid.md
+++ b/docs/pages/api-docs/data-grid/x-grid.md
@@ -21,7 +21,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">autoHeight</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid height is dynamic and follow the number of rows in the grid. |
 | <span class="prop-name">autoPageSize</span> | <span class="prop-type">boolean</span> | false | If `true`, the pageSize is calculated according to the container size and the max number of rows to avoid rendering a vertical scroll bar. |
 | <span class="prop-name">checkboxSelection</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid get a first column with a checkbox that allows to select rows. |
-| <span class="prop-name">checkboxSelectionVisibleOnly</span> | <span class="prop-type">boolean</span> | false | To be used in combination with `checkboxSelection`. If `true`, `Select All` selects only the rows on the current page. |
+| <span class="prop-name">checkboxSelectionVisibleOnly</span> | <span class="prop-type">boolean</span> | false | If `true`, the "Select All" header checkbox selects only the rows on the current page. To be used in combination with `checkboxSelection`. |
 | <span class="prop-name">classes</span> | <span class="prop-type">GridClasses</span> |   | Override or extend the styles applied to the component. See [CSS API](/api/data-grid/x-grid/#css) below for more details. |
 | <span class="prop-name">className</span> | <span class="prop-type">string</span> |   | CSS classname to add on the outer container. |
 | <span class="prop-name">columnBuffer</span> | <span class="prop-type">number</span> | 2 | Number of columns rendered outside the grid viewport. |
--- a/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
+++ b/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
@@ -4,6 +4,7 @@ import {
   GRID_SELECTION_CHANGE,
 } from '../../constants/eventsConstants';
 import { useGridSelector } from '../../hooks/features/core/useGridSelector';
+import { gridPaginatedVisibleSortedGridRowIdsSelector } from '../../hooks/features/pagination/gridPaginationSelector';
 import { visibleSortedGridRowIdsSelector } from '../../hooks/features/filter/gridFilterSelector';
 import { gridTabIndexColumnHeaderSelector } from '../../hooks/features/focus/gridFocusStateSelector';
 import { gridRowCountSelector } from '../../hooks/features/rows/gridRowsSelector';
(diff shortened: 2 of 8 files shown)


packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
```diff
@@ -28,7 +35,10 @@ export const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnH
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
       const checked = event.target.checked;
-      apiRef!.current.selectRows(visibleRowIds, checked, !event.target.indeterminate);
+      const rowsToBeSelected = checkboxSelectionVisibleOnly
+        ? paginatedVisibleRowIds
```
flaviendelangle, line 39: We are running both selectors even when the callback is not called
Is it totally negligible performance-wise or should we add a callback version of the selectors ?
Something like :

```ts
const getVisibleRowIds = useGridSelectorCallback(apiRef, visibleSortedGridRowIdsSelector)
const getPaginatedVisibleRowIds = useGridSelectorCallback(apiRef, gridPaginatedVisibleSortedGridRowIdsSelector)

...

const handleChange = () => {
  const rowsToBeSelector = checkboxSelectionVisibleOnly
    ? getPaginatedVisibleRowIds()
    : getVisibleRowIds()
}
```

And even without it, some selectors are performing actions like filter on large table, but from what I understand, a selector is never memoized, it run at each render of the component using `useGridSelector`.
But this is not exactly related to this PR.

```ts
export const useGridSelector = <State>(
  apiRef: GridApiRef | undefined,
  selector: (state: any) => State,
) => {
  const [state] = useGridState(apiRef!);
  return selector(state);
};
```

dtassone: we memoize at the selector level using `createSelector` from 'reselect'
https://github.com/reduxjs/reselect#creating-a-memoized-selector

flaviendelangle: Oh indeed, my bad.
So forget my second point.

And for the 1st one, I suppose you consider that selectors like `visibleSortedGridRowIdsSelector` will always be memoized somewhere by another component that does need it on every render ?

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -21,6 +21,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">autoHeight</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid height is dynamic and follow the number of rows in the grid. |
 | <span class="prop-name">autoPageSize</span> | <span class="prop-type">boolean</span> | false | If `true`, the pageSize is calculated according to the container size and the max number of rows to avoid rendering a vertical scroll bar. |
 | <span class="prop-name">checkboxSelection</span> | <span class="prop-type">boolean</span> | false | If `true`, the grid get a first column with a checkbox that allows to select rows. |
+| <span class="prop-name">checkboxSelectionVisibleOnly</span> | <span class="prop-type">boolean</span> | false | To be used in combination with `checkboxSelection`. If `true`, `Select All` selects only the rows on the current page. |
```
dtassone, line 24: Should this option be only xgrid 🤔

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
dtassone: If it's a selector it will be memoized and recomposed at the first call when the state change. It doesn't matter if it is called 1 or 100 times.

docs/pages/api-docs/data-grid/x-grid.md
DanailH: Can be. It doesn't seem to be that critical to not give it for free. Also since the free version pagination is auto-enabled after a certain amount of rows it makes sense to allow that option as well.

dtassone: Yes, but we could argue that this is for the advanced use cases where users would want to define specific grid behavior.

DanailH: done

dtassone: @oliviertassinari what do you think? Up for putting it in XGrid as it's an advanced configurable grid or prefers to put it in DataGrid as well

oliviertassinari: Regarding DataGrid vs. XGrid, no objections for XGrid only.

---

I can't find this feature in AG Grid. In react-table, I can find it exposed as getToggleAllPageRowsSelectedProps vs. getToggleAllRowsSelectedProps: https://github.com/tannerlinsley/react-table/blob/8ec25cabccf915991ec00cf67bfb26a83aa0c0b2/docs/src/pages/docs/api/useRowSelect.md. react-table. In both cases, the default behavior is to select all the pages. So I think that we are good on this front.

My main question would be: should it be a boolean or an enum? We could imagine something in this order in the long run. 

```jsx
checkboxSelectionMode?: 'all' | 'page' | 'ask'
```

or leave it to the developers, and improve the apiRef to have more than `selectRows()` but something with pre-existing options.

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -55,6 +55,11 @@ export interface GridOptions {
    * @default false
    */
   checkboxSelection?: boolean;
+  /**
+   * To be used in combination with `checkboxSelection`. If `true`, `Select All` selects only the rows on the current page.
```
oliviertassinari, line 59: I think that we should revert the sentence, the to be used seems secondary information.

```suggestion
   * If `true`, the "Select All" header checkbox selects only the rows on the current page. To be used in combination with `checkboxSelection`. 
```

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -12,6 +12,7 @@ const FORCED_PROPS: Partial<GridComponentProps> = {
   pagination: true,
   apiRef: undefined,
   onRowsScrollEnd: undefined,
+  checkboxSelectionVisibleOnly: false,
```
oliviertassinari, line 15: I would sort ASC, to make it easier to find and reduce likeness of conflicts (forced, prop-type, types, etc.)

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
oliviertassinari: @flaviendelangle 👍 for calling the selector directly in `handleChange()` with the state from `useGridState`. We don't need to run the slice each time the visible rows change. In this very instance, it probably doesn't really matter, but I think that it could set an interesting precedent for this more systematized "lazy" pattern. 

The rule could be:

- `useGridSelector` if the value is used in the render method.
- call the selector directly if used inside a handler.

Redux **nor** Recoil supports this pattern, why? what's the downside?

flaviendelangle: I can see a down side when we are using the callback in an effect and want it to run whenever the dependencies of the callback are changing.

docs/pages/api-docs/data-grid/x-grid.md
DanailH: I was wondering if there may be a need for more values that's why I called it `checkboxSelectionVisibleOnly`. If in the future we decide to also change the filtering selection that can still fall under the same option because the name is not related to the pagination (the same applies for the infinite loading) 🤔

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
DanailH: This discussion seems to go off-topic from that is in the PR. We can always move it to a dedicated discussion.

I've updated the PR to call the selectors directly inside the `handleChange`.

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
```diff
@@ -4,21 +4,22 @@ import {
...
 import { isNavigationKey, isSpaceKey } from '../../utils/keyboardUtils';
 import { useGridApiContext } from '../../hooks/root/useGridApiContext';
+import { optionsSelector } from '../../hooks/utils/optionsSelector';
 
 export const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnHeaderParams>(
   function GridHeaderCheckbox(props, ref) {
     const [, forceUpdate] = React.useState(false);
     const apiRef = useGridApiContext();
-    const visibleRowIds = useGridSelector(apiRef, visibleSortedGridRowIdsSelector);
+    const { checkboxSelectionVisibleOnly } = useGridSelector(apiRef, optionsSelector);
```
flaviendelangle, line 21: https://github.com/mui-org/material-ui-x/pull/1952#discussion_r656968843

Like @oliviertassinari , I'd prefer to avoid destructuring for widely used objects like `options`. 
It make it easier to search for usage across the codebase.

```suggestion
    const options = useGridSelector(apiRef, optionsSelector);
```

flaviendelangle (approved):

PR #2003, opened 2021-06-30
[docs] Fix description of union types

m4theushw:
Fixes https://github.com/mui-org/material-ui-x/pull/1956#discussion_r660642444


docs/scripts/buildApi.ts
```diff
@@ -45,7 +45,19 @@ function generateType(type) {
...
+      let text = '{ ';
+      text += type.declaration.children
+        .map((child) => {
+          let memberText = child.name;
+          if (child.flags.isOptional) memberText += '?';
+          return `${memberText}: ${child.type ? generateType(child.type) : 'any'}`;
+        })
+        .join('; ');
+      text += ' }';
+      return text;
```
oliviertassinari, line 60: Are there more cases like https://github.com/TypeStrong/typedoc-default-themes/blob/a87b8eab47a2ad9e7faa267494d7504f6756614f/src/default/partials/type.hbs#L169 that we might not have spotted yet? 😁

m4theushw: Yes, certainly. But I would go on demand here. It's hard to translate this template to JavaScript without knowing when each case is used.

oliviertassinari: Ok, good to know, @mui-org/x keep an 👀 open for them.

DanailH (approved):

PR #2008, opened 2021-07-01
[DataGrid] Allow to customize the columns exported as CSV

### LINKED ISSUE #1435: [DataGrid] Allow to remove column from CSV export

## Summary 💡
I would like to be able to customize whether I need to export some columns. 

## Examples 🌈
https://codesandbox.io/s/github/gregnb/mui-datatables?file=/examples/csv-export/index.js:678-686

## Motivation 🔦
For example with action columns.

## Order id 💳
19652

flaviendelangle:
Closes #1435 

Preview: https://deploy-preview-2008--material-ui-x.netlify.app/components/data-grid/export/#customize-which-columns-are-exported


flaviendelangle:
@mui-org/x are we OK with this behavior ? 
The two options should be working just like in AG Grid.

This PR needs some demo, and I need to implement it as props of `GridToolbarExport`, but I would like to fix the format first.

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -23,6 +23,16 @@ export interface GridExportCsvOptions {
    * @default false
    */
   utf8WithBom?: boolean;
+  /**
+   * The columns exported in the CSV
+   * This should only be used if you want to restrict the columns exports
+   */
+  columnKeys?: string[];
```
DanailH, line 30: I think it will be better to just have a value like `notExportable` or `disableExport` inside the `ColDef` and use it to know what column is exportable and what not.

flaviendelangle: https://github.com/mui-org/material-ui-x/issues/1435#issuecomment-825492594

For me that was the discussion here.
If we follow the AG Grid solution, the "not exportable" columns are flagged with a param in the export method (which will also be available as in the prop `csvOptions` in `GridToolbarExport`)
If we follow the material-table solution, the "not exportable" columns are flagged with a param in the `ColDef`

I tend to prefer the AG Grid approach here because it easily allow to to several export buttons with different configurations. Which is a use case that I had quite often on my previous job.

---

Of course we could also implement both (right now or later).

DanailH: That comment is regarding the delimiter as far as I can see. That to me is a different use case. 

I'm not saying that we should remove what you have done here but rather add the option in the `ColDef` as well - logic is that there are certain columns that we can be 100% sure that devs wouldn't want to export such as the checkbox or ID or if they have a column with buttons that do different actions.
Since the CSV export is a free feature it makes sense to provide the option to mark those columns as not exportable somehow in the definition.

flaviendelangle: I mis-read the comment then.

> Since the CSV export is a free feature it makes sense to provide the option to mark those columns as not exportable somehow in the definition.

Do we agree that the "AG Grid" solution (= the current state of the PR) does allow the free user to use it (by passing a prop to `GridToolbarExport` ? Or am I missing something ?

---

And no problem to implement both.
I suppose that if both were used at the same time, the `columnKeys` would have the priority over the `ColDef`

flaviendelangle: Done :+1:

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -143,6 +143,11 @@ export interface GridColDef {
    * Allows setting the filter operators for this column.
    */
   filterOperators?: GridFilterOperator[];
+  /**
+   * If `true`, the column will not be included in exports
+   * @default true
```
DanailH, line 148: As it is now the default is `false` 😉

DanailH:
Can we also update the docs here https://material-ui.com/components/data-grid/export/#data-grid-export. Maybe add a line saying how can columns be excluded from the export.

DanailH (approved):

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
 
+### Only export some columns
+
+By default, the CSV will only contain the visible columns of the table.
+
+To export all columns, including the hidden one, you can use the `allColumns` option in the `csvOptions` property of `GridToolbarExport`
+
+To exclude additional columns, you can use the `disableExport` property available in `GridColDef`
+
+To define the exact columns you want to export, you can use the `columnKeys` option in the `csvOptions` property of `GridToolbarExport`
```
dtassone, line 25: columnKeys => keys are called fields in `GridColDef`, I think `keys` is breaking the ubiquitous language.

flaviendelangle: I took the naming from AG Grid but field would indeed match better our own naming.

flaviendelangle: Perhaps we should also rename `allColumns` into `allFields`

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -143,6 +143,11 @@ export interface GridColDef {
    * Allows setting the filter operators for this column.
    */
   filterOperators?: GridFilterOperator[];
+  /**
+   * If `true`, the column will not be included in exports
```
m4theushw, line 147:
```suggestion
   * If `true`, this column will not be included in exports.
```

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -23,6 +23,16 @@ export interface GridExportCsvOptions {
    * @default false
    */
   utf8WithBom?: boolean;
+  /**
+   * The columns exported in the CSV
+   * This should only be used if you want to restrict the columns exports
+   */
+  fields?: string[];
+  /**
+   * If `true`, the hidden columns will also be exported
```
m4theushw, line 32:
```suggestion
   * If `true`, the hidden columns will also be exported.
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
 
 {{"demo": "pages/components/data-grid/export/ExportSelectorGrid.js", "bg": "inline"}}
 
+### Only export some columns
+
+By default, the CSV will only contain the visible columns of the table.
```
m4theushw, line 19:
```suggestion
By default, the CSV will only contain the visible columns of the grid.
```

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -23,6 +23,16 @@ export interface GridExportCsvOptions {
    * @default false
    */
   utf8WithBom?: boolean;
+  /**
+   * The columns exported in the CSV
+   * This should only be used if you want to restrict the columns exports
```
m4theushw, line 28:
```suggestion
   * The columns exported in the CSV.
   * This should only be used if you want to restrict the columns exports.
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
 
 {{"demo": "pages/components/data-grid/export/ExportSelectorGrid.js", "bg": "inline"}}
 
+### Only export some columns
```
m4theushw, line 17: Developers can opt to export more columns than only those that are visible.

```suggestion
### Customize which columns are exported
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
+
+By default, the CSV will only contain the visible columns of the table.
+
+To export all columns, including the hidden one, you can use the `allColumns` option in the `csvOptions` property of `GridToolbarExport`
+
+To exclude additional columns, you can use the `disableExport` property available in `GridColDef`
+
+To define the exact columns you want to export, you can use the `fields` option in the `csvOptions` property of `GridToolbarExport`
```
m4theushw, line 25: I think is missing which approach takes precedence. We have two main ways: `fields` and `disableExport`. If `fields` is false, then developers can also use `allColumns`.

```suggestion
To include or hide other columns, there are two ways:

1. Define the exact columns to be exported with the `fields` attribute in the `csvOptions` prop of [`GridToolbarExport`](/components/data-grid/components/#toolbar). 
Set `allColumns` in `csvOptions` to true to include hidden columns, instead of only the visible ones.

2. Set the `disableExport` attribute to true in each `GridColDef`.

<!-- demo -->
```

dtassone:
@flaviendelangle don't hesitate to resolve the code review comment once you did the change...

packages/grid/_modules_/grid/hooks/features/export/useGridCsvExport.tsx
```diff
@@ -2,34 +2,48 @@ import * as React from 'react';
...
   const getDataAsCsv = React.useCallback(
     (options?: GridExportCsvOptions): string => {
       logger.debug(`Get data as CSV`);
 
+      let exportedColumns: GridColDef[];
+
+      if (options?.fields) {
+        exportedColumns = options.fields
+          .map((field) => columns.find((column) => column.field === field))
+          .filter((column): column is GridColDef => !!column);
```
oliviertassinari, line 31: Oh nice, first time I see this TS feature used: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates, I have learned something 👍

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -143,6 +143,11 @@ export interface GridColDef {
    * Allows setting the filter operators for this column.
    */
   filterOperators?: GridFilterOperator[];
+  /**
+   * If `true`, this column will not be included in exports
```
oliviertassinari, line 147:
```suggestion
   * If `true`, this column will not be included in exports.
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
 
 {{"demo": "pages/components/data-grid/export/ExportSelectorGrid.js", "bg": "inline"}}
 
+### Customize which columns are exported
```
oliviertassinari, line 17: To keep the headers short?

```suggestion
### Customize exported columns
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
+
+By default, the CSV will only contain the visible columns of the grid.
+
```
oliviertassinari, line 20: No real need for two `<p>` here:

```suggestion
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
+To include or hide other columns, there are two ways:
+
+1. Define the exact columns to be exported with the `fields` attribute in the `csvOptions` prop of [`GridToolbarExport`](/components/data-grid/components/#toolbar).
+   Set `allColumns` in `csvOptions` to true to include hidden columns, instead of only the visible ones.
```
oliviertassinari, line 24: I think that we could make it even more obvious for the developers:

````suggestion
    ```jsx
   <GridToolbarExport csvOptions={{ fields: ['id', 'name'] }} />
   ```

   Set `allColumns` in `csvOptions` to true to include hidden columns, instead of only the visible ones.
````

File: docs/src/pages/components/data-grid/export/export.md [hunk 15]
```diff
@@ -14,6 +14,16 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
+2. Set the `disableExport` attribute to true in each `GridColDef`.
```
oliviertassinari, line 25:
````suggestion
2. Set the `disableExport` attribute to true in each `GridColDef`.
  
  ```jsx
  <DataGrid
    columns={[
      { field: 'id', disableExport: true },
      { field: 'brand' },
    ]}
  ```
````

packages/grid/_modules_/grid/hooks/features/export/useGridCsvExport.tsx
flaviendelangle: I love this feature
It's not usefull very often, but for non-nullish filters it makes the result a lot more readable.

dtassone (approved):

PR #2010, opened 2021-07-01
[DataGrid] apiRef.current.getRow can return null

flaviendelangle:
See https://github.com/mui-org/material-ui-x/pull/1995#issuecomment-871248683

For `getRowParams` / `getCellParams` / `getBaseCellParams` I could return `null` instead of throwing, but it would mean checking if the return value is null in a lot more places.

For `getCellValue` if I return `null` we could not differentiate a cell with the value `null` and a cell with a non existing ID

For `selectRow`, I had to update the test because we were asking for a row that do not exist but we never checked the return value of the function in the tests. I don't know what is the best behavior here, does it make sense to return `null` instead of crashing ?


packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -113,9 +113,15 @@ export const useGridSelection = (apiRef: GridApiRef): void => {
 
   const selectRow = React.useCallback(
     (id: GridRowId, isSelected = true, allowMultiple = false) => {
+      const row = apiRef.current.getRow(id);
+
+      if (!row) {
+        throw new Error(`No row with id #${id} found`);
```
dtassone, line 119: Remove the throw here pls just return;

flaviendelangle: Done :+1:

dtassone (approved):

Issue #1993, opened 2021-06-29
[DataGrid] React does not recognize the `defaultProps` prop on a DOM element


jbty:
- [ x ] The issue is present in the latest release.
- [ x ] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Reproduction: https://codesandbox.io/s/material-demo-forked-xeimg?file=/demo.js

On dev server when i use the Datagrid component I get this red warning :

```
Warning: React does not recognize the `defaultProps` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `defaultprops` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

    at div
    at http://localhost:3000/static/js/vendors~main.chunk.js:1343:73
    at TablePagination (http://localhost:3000/static/js/vendors~main.chunk.js:41213:84)
    at http://localhost:3000/static/js/vendors~main.chunk.js:61000:11
    at div
    at http://localhost:3000/static/js/vendors~main.chunk.js:59109:13
    at http://localhost:3000/static/js/vendors~main.chunk.js:60935:11
    at div
    at Jc (http://localhost:3000/static/js/vendors~main.chunk.js:64737:11)
    at Zc (http://localhost:3000/static/js/vendors~main.chunk.js:64635:193)
    at Xc (http://localhost:3000/static/js/vendors~main.chunk.js:64673:13)
    at div
    at NoSsr (http://localhost:3000/static/js/vendors~main.chunk.js:27107:24)
    at http://localhost:3000/static/js/vendors~main.chunk.js:59020:11
    at Uc (http://localhost:3000/static/js/vendors~main.chunk.js:64613:18)
    at http://localhost:3000/static/js/vendors~main.chunk.js:65077:16
    at http://localhost:3000/static/js/vendors~main.chunk.js:65098:11
    at StyledComponent (http://localhost:3000/static/js/vendors~main.chunk.js:289635:28)
    at div
    at http://localhost:3000/static/js/vendors~main.chunk.js:1343:73
    at Paper (http://localhost:3000/static/js/vendors~main.chunk.js:28609:84)
    at FaDataTab (http://localhost:3000/static/js/3.chunk.js:264:22)
    at div
    at http://localhost:3000/static/js/vendors~main.chunk.js:1343:73
[...]

PR #2012, opened 2021-07-02
[core] Add @material-ui/lab as peer dependency

### LINKED ISSUE #1878: @material-ui/x-grid-data-generator error: 'withStyles' is not exported from '@material-ui/core/styles'
Material-UI 5.0.0-alpha.35 or higher is not compatible with x-grid-data-generator, since withStyles is not longer in @material-ui/core.

This is the specific error:
```
./node_modules/@material-ui/x-grid-data-generator/node_modules/@material-ui/lab/esm/Rating/Rating.js
Attempted import error: 'withStyles' is not exported from '@material-ui/core/styles'.

```

m4theushw:
Fixes #1878 

Noticed while debugging #1993 that demos depending in `@material-ui/x-grid-data-generator` won't work with MUI v5 because the version of the lab is fixed in v4. This PR makes the lab a peer dependency.

image

Before: https://codesandbox.io/s/material-demo-forked-1vzw6?file=/package.json

After: https://codesandbox.io/s/material-demo-forked-nzqjp?file=/package.json

---

I'm aware that clicking in any cell to edit crashes the demo. I'll fix this in another PR. The first step was to make them at least run with v5.

image

### DIFF
--- a/packages/grid/x-grid-data-generator/package.json
+++ b/packages/grid/x-grid-data-generator/package.json
@@ -23,7 +23,6 @@
   },
   "dependencies": {
     "@material-ui/icons": "^4.9.1",
-    "@material-ui/lab": "^4.0.0-alpha.54",
     "@types/chance": "^1.1.0",
     "chance": "^1.1.6",
     "clsx": "^1.0.4",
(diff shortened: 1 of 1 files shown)


flaviendelangle:
https://github.com/mui-org/material-ui-x/issues/1878#issuecomment-860913529

I think we want to do the same for `@material-ui/icons`

flaviendelangle (approved):

Issue #1907, opened 2021-06-15
[DataGrid] rowsPerPage selection dropdown is not shown


tobiasbueschel:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When rendering a `DataGrid` component, the rows per page dropdown selection is not shown.

```jsx
      <DataGrid
        pageSize={pageSize}
        onPageSizeChange={handlePageSizeChange}
        rowsPerPageOptions={[5, 10, 20]}
        pagination
        {...data}
      />
```

This also happens when I look at the Code Sandbox of the official MUI example shown here:

- https://material-ui.com/components/data-grid/pagination/#page-size
- https://codesandbox.io/s/o1iep?file=/demo.tsx

![image](https://user-images.githubusercontent.com/13087421/121983511-040a4c80-cdc4-11eb-80cd-11794265d909.png)

## Expected Behavior 🤔

![image](https://user-images.githubusercontent.com/13087421/121983561-1b493a00-cdc4-11eb-9719-ef8b6337ac9e.png)

## Steps to Reproduce 🕹

https://codesandbox.io/s/o1iep?file=/demo.tsx

## Context 🔦

It doesn't allow us to dynamically change the page size.

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
  System:
    OS: macOS 11.4
  Binaries:
    Node: 16.2.0 - ~/.nvm/versions/node/v16.2.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.14.0 - ~/Development/project/node_modules/.bin/npm
  Browsers:
    Chrome: 91.0.4472.101
    Edge: Not Found
    Firefox: Not Found
    Safari: 14.1.1
  npmPackages:
    @material-ui/core: ~4.11.4 => 4.11.4
    @material-ui/data-grid: ~4.0.0-alpha.30 => 4.0.0-alpha.31
    @material-ui/icons: ~4.11.2 => 4.11.2
    @material-ui/lab: ~4.0.0-alpha.58 => 4.0.0-alpha.58
    @material-ui/pickers: ~3.3.10 => 3.3.10
    @material-ui/styles:  4.11.4
    @material-ui/system:  4.11.3
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
[...]

Issue #1325, opened 2021-03-31
[DataGrid] Use explicit definition instead of React.FC


m4theushw:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

We're typing many components with `React.FC`. One of the [problems](https://github.com/facebook/create-react-app/pull/8177) with this approach is that it provides an implicit definition of the `children` prop, even for those components that don't use it. We should favor an explicit definition instead.

```diff
diff --git a/packages/grid/_modules_/grid/components/GridRowCount.tsx b/packages/grid/_modules_/grid/components/GridRowCount.tsx
index 4d888ce..45ca4a3 100644
--- a/packages/grid/_modules_/grid/components/GridRowCount.tsx
+++ b/packages/grid/_modules_/grid/components/GridRowCount.tsx
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import { GridApiContext } from './GridApiContext';

-export const GridRowCount: React.FC<{ rowCount: number }> = ({ rowCount }) => {
+export interface GridRowCountProps {
+  rowCount: number
+}
+
+export const GridRowCount = (props: GridRowCountProps) => {
+  const { rowCount } = props;
   const apiRef = React.useContext(GridApiContext);

   if (rowCount === 0) {
```

PR #2014, opened 2021-07-02
[DataGrid] Warn when pageSize is not present in rowsPerPageOptions

flaviendelangle:
One chunk of #1907


flaviendelangle:
The snapshots changes are expected, we don't pass `rawPerPageOptions` as props so the default one are used (`DEFAULT_GRID_OPTIONS.rowsPerPageOptions`)

flaviendelangle:
@oliviertassinari , I updated the behavior as discussed yesterday :+1:

flaviendelangle:
Their is an error on a test because the effect runs with the default `pageSize` and then with the value passed by the user.
It should be fixed by the clean controlled state. It put this PR on hold until it's done :+1:

oliviertassinari (commented):
👍 for the current direction

packages/grid/_modules_/grid/components/GridPagination.tsx
```diff
@@ -79,6 +79,25 @@ export const GridPagination = React.forwardRef<
...
+  const hasPageSizeInRowsPerPageOptions = options.rowsPerPageOptions?.includes(
+    paginationState.pageSize,
+  );
+
+  React.useEffect(() => {
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      options.rowsPerPageOptions &&
+      !hasPageSizeInRowsPerPageOptions
+    ) {
```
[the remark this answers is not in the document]
flaviendelangle: You're right, and running it in the render would also avoid this issue https://github.com/mui-org/material-ui-x/pull/2014#issuecomment-874717496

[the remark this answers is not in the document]
flaviendelangle: My bad, it's not enough
If I put it in the render, on the 1st render I have the default pagination.
So I'll have to wait to the controlled mode rework.

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -320,6 +320,34 @@ describe('<DataGrid /> - Pagination', () => {
         );
         expect(getColumnValues(0)).to.deep.equal(['7', '8']);
       });
+
+      it('should display a warning if the pageSize is not in the rowsPerPageOptions', () => {
+        const pageSize = 12;
+
+        const TestCase: React.FC = () => {
```
m4theushw, line 327: Just a small thing. We stopped using `React.FC` in #1325.

```suggestion
        const TestCase = () => {
```

flaviendelangle: We can use `React.VoidFunctionComponent` to exclude the children

oliviertassinari:
@flaviendelangle I have rebased and updated the tests to pass. A codesandbox with the new DX: https://codesandbox.io/s/material-demo-forked-pjlyc?file=/demo.tsx. However, having to include 100 in the list, that's not flying 😁. The simplest might be to add a hack for 100 and to leave a TODO.

Updating the options in an effect? It's borderline. I think that it would only work under two conditions: 1. We initiate the state with the right initial options (KO currently) 2. We make the child of the state provider a memo, to always stop the React rendering so that children only render once per option updates (KO currently).

Otherwise, we can use React, create a dedicated context for the options (top-level props).

m4theushw:
In #1720 a new `GridRootPropsContext` context was added with the root props. You can use it instead of querying the state. We get for free the advantage of not having to add 100 to the list. You just have to solve the dependency cycle, maybe add the prop you need in the parent element and pass it to the pagination component.

flaviendelangle:
I'll finish this one once the controlled pagination is merged
And I agree, looking at the input props should help

flaviendelangle:
https://app.circleci.com/pipelines/github/mui-org/material-ui-x/6357/workflows/3eb1a423-62eb-40b7-b3ff-5cf21a9beef9/jobs/33830

@oliviertassinari if I set a global variable to only warn once, it's not reset between two tests so 2 of the 3 tests don't have the warning.
I'm using a ref to bypass this issue.

packages/grid/_modules_/grid/hooks/features/useGridComponents.tsx
```diff
@@ -1,8 +1,73 @@
...
+  ColumnSortedAscendingIcon: GridArrowUpwardIcon,
+  ColumnSortedDescendingIcon: GridArrowDownwardIcon,
+  ColumnResizeIcon: GridSeparatorIcon,
+  DensityCompactIcon: GridViewHeadlineIcon,
+  DensityStandardIcon: GridTableRowsIcon,
+  DensityComfortableIcon: GridViewStreamIcon,
+  ExportIcon: GridSaveAltIcon,
+};
+
+export const DEFAULT_GRID_SLOTS_COMPONENTS: GridApiRefComponentsProperty = {
```
flaviendelangle, line 55: Same reason as https://github.com/mui-org/material-ui-x/pull/2213/files#r676538144
It's to avoid circle import

oliviertassinari:
@flaviendelangle 

Capture d’écran 2021-07-27 à 12 15 29

You have discarded one of my commit. Is it on purpose?

packages/grid/_modules_/grid/hooks/features/useGridComponents.tsx
```diff
@@ -1,8 +1,73 @@
 import * as React from 'react';
+import MUICheckbox from '@material-ui/core/Checkbox';
+
```
oliviertassinari, line 3:
```suggestion
```

oliviertassinari (commented):
Looks good

flaviendelangle:
@oliviertassinari what was the goal of adding the `100` in `rowsPerPageOptions` ?

oliviertassinari:
@flaviendelangle To dodge the initial value issue. You fixed this using the root props context.

packages/grid/_modules_/grid/components/GridPagination.tsx
```diff
@@ -80,6 +83,25 @@ export const GridPagination = React.forwardRef<
     };
   };
 
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const warnedOnceMissingPageSizeInRowsPerPageOptions = React.useRef(false);
+    if (
+      !warnedOnceMissingPageSizeInRowsPerPageOptions.current &&
+      !rootProps.rowsPerPageOptions!.includes(rootProps.pageSize ?? DEFAULT_PAGE_SIZE)
```
m4theushw, line 91: We can give a default value for the prop in:

https://github.com/mui-org/material-ui-x/blob/75ae2f73204939c73ae3b8f5597b6541fefe34dc/packages/grid/data-grid/src/DataGrid.tsx#L53

The same for the XGrid.

```suggestion
      !rootProps.rowsPerPageOptions!.includes(rootProps.pageSize)
```

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -292,6 +334,37 @@ describe('<DataGrid /> - Pagination', () => {
...
+        `Material-UI: The page size \`${pageSize}\` is not preset in the \`rowsPerPageOptions\``,
+      ]);
+    });
+
+    it('should display a warning if the default pageSize given as props is not in the prop rowsPerPageOptions', () => {
+      expect(() => {
+        render(<BaselineTestCase rowsPerPageOptions={[25, 50]} />);
+        // @ts-expect-error need to migrate helpers to TypeScript
+      }).toWarnDev([
+        `Material-UI: The page size \`${DEFAULT_PAGE_SIZE}\` is not preset in the \`rowsPerPageOptions\``,
```
m4theushw, line 365: I wouldn't mind hardcoding the value if we want to avoid exporting constants.

```suggestion
        `Material-UI: The page size \`100\` is not preset in the \`rowsPerPageOptions\``,
```

packages/grid/_modules_/grid/components/GridPagination.tsx
flaviendelangle: If we do so, then `useGridPageSize` will always be in controlled mode no ?
We could give the default props only to the context and not to the feature hooks but I don't think it's a good idea.

m4theushw: We can use the state value, If think. One test related to auto size will fail, maybe we don't display the warning in this condition because there's no way to change the rows per page.

```suggestion
      !rootProps.rowsPerPageOptions!.includes(rootProps.pageSize ?? paginationState.pageSize)
```

oliviertassinari (commented):
It looks clean

m4theushw (approved):

PR #2034, opened 2021-07-05
[DataGrid] Remove 'hide: true' from a column should correctly resize the others column

### LINKED ISSUE #2004: [DataGrid] Hidden columns widths still used in calculation

## Current Behavior 😯

https://user-images.githubusercontent.com/5405326/123953124-2c3ea180-d9a7-11eb-917e-154e26012cd8.mp4

## Expected Behavior 🤔

Should update the column as soon as the column changes.

## Steps to Reproduce 🕹

Steps:

2. Hide a column
3. 💥

## Context 🔦

Reported by @si7o in https://github.com/mui-org/material-ui-x/issues/708#issuecomment-871336411.

## Your Environment 🌎

v4.0.0-alpha.32

flaviendelangle:
Fixes #2004 

Based on 2nd diff in this comment https://github.com/mui-org/material-ui-x/issues/2004#issuecomment-871781521
The issue with the code in the comment was that when we changed the `columns` prop to remove a column, `upsertColumnsState` was neither removing them from the columns list nor publishing the event.


flaviendelangle:
I'll add a test :+1:

packages/grid/_modules_/grid/models/api/gridColumnApi.ts
```diff
@@ -47,8 +47,9 @@ export interface GridColumnApi {
   /**
    * Updates the definition of multiple columns at the same time.
    * @param {GridColDef[]} cols The new column [[GridColDef]] objects.
+   * @param {boolean} allColumnsProvided Determines if the `cols` argument contains all the columns or just the updated ones
    */
-  updateColumns: (cols: GridColDef[]) => void;
+  updateColumns: (cols: GridColDef[], allColumnsProvided?: boolean) => void;
```
dtassone, line 52: `allColumnsProvided` is this one to flag if it is a reset or an update? 🤔

flaviendelangle: The issue was that @oliviertassinari was trying to use `updateColumns` when the column prop change.
But when we removed column from the prop, they were not removed from the state because `upsertColumnsState` keeps the previous columns in the internal state.
So I added a flag on `updateColumns` to ask the code not to keep the previous ones.

It's indeed a flag to differentiate update and reset

If you don't want to pollute `updateColumn`, we could add a method `setColumns` and the effect that runs when the `columns` prop changes would call `setColumns`. 
`updateColumns` could also call `setColumns` to keep as much logic in common as possible.

dtassone: > If you don't want to pollute updateColumn, we could add a method setColumns and the effect that runs when the columns prop changes would call setColumns.
updateColumns could also call setColumns to keep as much logic in common as possible.

Yes that sounds much cleaner.

flaviendelangle: Updated

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -160,13 +165,32 @@ export function useGridColumns(apiRef: GridApiRef, { columns }: { columns: GridC
     [columnsMeta.positions, getColumnIndex],
   );
 
+  const setColumnsWithNewState = React.useCallback(
```
dtassone, line 168: rename to `setColumnsState`.

dtassone (approved):

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -375,6 +380,61 @@ describe('<DataGrid /> - Layout & Warnings', () => {
         });
       });
 
+      it('should resize flex: 1 column when setting hide: false on a column to avoid exceeding grid width', async () => {
```
oliviertassinari, line 383:
```suggestion
      it('should resize flex: 1 column when setting hide: false on a column to avoid exceeding grid width', () => {
```

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -375,6 +380,61 @@ describe('<DataGrid /> - Layout & Warnings', () => {
...
+        const TestCase: React.FC<DataGridProps> = (props) => (
```
oliviertassinari, line 384: We don't allow FC in the codebase (x an core). See https://github.com/mui-org/material-ui-x/pull/1436 why 

```suggestion
        const TestCase = (props: DataGridProps) => (
```

oliviertassinari (commented):
The added test case is not failing on HEAD. I don't know what's testing, but not the bug. The logic looks great otherwise, one abstraction where it's meant to be one.

flaviendelangle:
@oliviertassinari Indeed I was a bit quick on this one
If we update the columns through the props, it worked, it only failed with updates through `apiRef`

I'll make a new PR to move the text

PR #2036, opened 2021-07-05
[docs] Better explain what happens in the future

oliviertassinari:
Preview: https://deploy-preview-2036--material-ui-x.netlify.app/components/data-grid/getting-started/#license-key-installation

Try to solve this pain point:

```
Hi,

my company, XXX, just bought PRO licences of MaterialUI.
We received an email with the key and an expiry date in one year.

As developers we need to know how to integrate this key in our apps but we do not know what will occur in one year if we strictly follow the instructions in the "licence key installation" documentation.
Can you please elaborate on what will occur so that we can elaborate a strategy to integrate the key in pour apps.

Thanks.
```

They had already found https://material-ui.com/components/data-grid/getting-started/#license-key-installation

### DIFF
--- a/docs/src/pages/components/data-grid/getting-started/getting-started.md
+++ b/docs/src/pages/components/data-grid/getting-started/getting-started.md
@@ -190,8 +190,12 @@ LicenseInfo.setLicenseKey(
 
 The grid checks the key without making any network requests.
 
+#### What's the expiration policy?
+
+The licenses are perpetual, the license key will work forever.
+However, access to updates is not.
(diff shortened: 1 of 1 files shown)


McFoggy:
Thank you Olivier that clarifies the whole thing and answer our questions.

flaviendelangle (approved):

oliviertassinari:
It might duplicate a bit with the content on the pricing page and [its FAQ](https://next.material-ui.com/branding/pricing/) but since the question was technical and not sales-related, I think that it's a safe improvement.

Issue #1933, opened 2021-06-19
[DataGrid] Consider a different virtualization implementation


oliviertassinari:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Based on this benchmark https://github.com/mui-org/material-ui-x/pull/1903#issuecomment-861440717 there might be a different implementation of the virtualization that could yield better performance. We are currently underperforming compared to popular existing alternative s

## Motivation 🔦

A data grid is meant to make it easy to navigate information, the quality of the UX is heavily impacted by its rendering performance. 

Before we commit to a different direction, I think that it's critical we investigate the state of the art with a rigorous benchmark of the existing solution. We can measure the average frame rate when scrolling vertically relatively quickly (relative to the speed of reading/scanning with the yet). My viewport shows 9 columns and 20 rows. So far (*to update & complement*), we have, sorting by performance: 

- [React Virtualized](https://github.com/bvaughn/react-virtualized)
  - codesandbox: https://codesandbox.io/s/gallant-brook-v5d0c
  - deployed to production: https://csb-v5d0c.netlify.app/
  - Average frame rate: 45 FPS
  - Strategy: Use the native scroll feature. On scroll, each time a new row needs to be displayed, add it to the dom, and apply an absolute positioning, matching its position. Overscale 10 rows in each direction. ⚠️ doesn't render the correct DOM structure for a data grid.

  <img width="191" alt="Capture d’écran 2021-06-20 à 01 06 36" src="https://user-images.githubusercontent.com/3165635/122657448-f0fcd100-d163-11eb-98c6-97751cc009cb.png">

- [MUI+](https://github.com/Janpot/mui-plus)
  - codesandbox: https://codesandbox.io/s/recursing-jackson-4xgsl
  - deployed to production: https://csb-4xgsl.netlify.app/
  - Average frame rate: 36 FPS
  - Strategy: Use the native scroll feature. On scroll, each time a new row needs to be displayed, shift the translate value of the whole row container to account for rows that are unmounted. Overscan of 3 rows in each direction.

  <img width="192" alt="Capture d’écran 2021-06-15 à 13 57 12" src="https://user-images.githubusercontent.com/3165635/122048566-ba305f00-cde1-11eb-9cb0-eaf89b90198e.png">
  
- AG Grid
  - codesandbox: https://codesandbox.io/s/hopeful-ride-4wb7n
[...]

PR #2038, opened 2021-07-05
[XGrid] Add `onVirtualPageChange` prop

DanailH:
Part of https://github.com/mui-org/material-ui-x/issues/1247

Based on this comment https://github.com/mui-org/material-ui-x/pull/1715#issuecomment-869044841

### DIFF
--- a/docs/pages/api-docs/data-grid/x-grid.md
+++ b/docs/pages/api-docs/data-grid/x-grid.md
@@ -91,6 +91,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onRowsScrollEnd</span> | <span class="prop-type">(param: GridRowScrollEndParams) => void</span> |   | Callback fired when scrolling to the bottom of the grid viewport. |
 | <span class="prop-name">onSelectionModelChange</span> | <span class="prop-type">(param: GridSelectionModelChangeParams) => void</span> |   | Callback fired when the selection state of one or multiple rows changes. |
 | <span class="prop-name">onSortModelChange</span> | <span class="prop-type">(param: GridSortModelParams) => void</span> |   | Callback fired when the sort model changes before a column is sorted. |
+| <span class="prop-name">onVirtualPageChange</span> | <span class="prop-type">(param: GridVirtualPageChangeParams) => void</span> |   | Callback fired when the virtual page changes. |
 | <span class="prop-name">page</span> | <span class="prop-type">number</span> | 1   |  Set the current page. |
 | <span class="prop-name">pageSize</span> | <span class="prop-type">number</span> | 100 | Set the number of rows in one page. |
 | <span class="prop-name">pagination</span> | <span class="prop-type">boolean</span> | false | If `true`, pagination is enabled. |
--- a/packages/grid/_modules_/grid/constants/eventsConstants.ts
+++ b/packages/grid/_modules_/grid/constants/eventsConstants.ts
@@ -457,3 +457,9 @@ export const GRID_STATE_CHANGE = 'stateChange';
  * @event
  */
 export const GRID_COLUMN_VISIBILITY_CHANGE = 'columnVisibilityChange';
+
+/**
+ * Fired when a the virtual page changes. Called with a [[GridVirtualPageChangeParams]] object.
+ * @event
+ */
+export const GRID_VIRTUAL_PAGE_CHANGE = 'virtualPageChange';
--- a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
@@ -1,5 +1,9 @@
 import * as React from 'react';
-import { GRID_SCROLL, GRID_ROWS_SCROLL } from '../../../constants/eventsConstants';
+import {
+  GRID_SCROLL,
+  GRID_ROWS_SCROLL,
+  GRID_VIRTUAL_PAGE_CHANGE,
+} from '../../../constants/eventsConstants';
 import { GridApiRef } from '../../../models/api/gridApiRef';
 import { GridVirtualizationApi } from '../../../models/api/gridVirtualizationApi';
 import { GridCellIndexCoordinates } from '../../../models/gridCell';
@@ -25,6 +29,7 @@ import { InternalRenderingState } from './renderingState';
 import { useGridVirtualColumns } from './useGridVirtualColumns';
 import { gridDensityRowHeightSelector } from '../density/densitySelector';
 import { scrollStateSelector } from './renderingStateSelector';
+import { useGridApiOptionHandler } from '../../root/useGridApiEventHandler';
 
 export const useGridVirtualRows = (apiRef: GridApiRef): void => {
   const logger = useLogger('useGridVirtualRows');
@@ -148,6 +153,14 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
         setRenderingState({ virtualPage: nextPage });
         logger.debug(`Changing page from ${page} to ${nextPage}`);
         requireRerender = true;
+
+        apiRef.current.publishEvent(GRID_VIRTUAL_PAGE_CHANGE, {
+          currentPage: nextPage,
+          nextPage: nextPage + 1 < containerProps.lastPage ? nextPage + 1 : containerProps.lastPage,
+          firstRowIndex: nextPage * containerProps.viewportPageSize,
+          pageSize: containerProps.viewportPageSize,
+          api: apiRef,
+        });
       } else {
         if (!containerProps.isVirtualized && page > 0) {
           logger.debug(`Virtualization disabled, setting virtualPage to 0`);
@@ -405,4 +418,5 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
     GRID_SCROLL,
     preventViewportScroll,
   );
+  useGridApiOptionHandler(apiRef, GRID_VIRTUAL_PAGE_CHANGE, options.onVirtualPageChange);
 };
--- a/packages/grid/_modules_/grid/models/gridOptions.tsx
+++ b/packages/grid/_modules_/grid/models/gridOptions.tsx
@@ -30,6 +30,7 @@ import { GridResizeParams } from './params/gridResizeParams';
 import { GridColumnResizeParams } from './params/gridColumnResizeParams';
 import { GridColumnVisibilityChangeParams } from './params/gridColumnVisibilityChangeParams';
 import { GridClasses } from './gridClasses';
+import { GridVirtualPageChangeParams } from './params/gridVirtualPageChangeParams';
 
 export type MuiEvent<E> = (E | React.SyntheticEvent) & {
   defaultMuiPrevented?: boolean;
@@ -431,6 +432,10 @@ export interface GridOptions {
    * Callback fired when the state of the grid is updated.
    */
   onStateChange?: (params: any) => void;
+  /**
+   * Callback fired when the virtual page changes.
+   */
+  onVirtualPageChange?: (params: GridVirtualPageChangeParams) => void;
   /**
    * Set the current page.
    * @default 1
--- a/packages/grid/_modules_/grid/models/params/gridVirtualPageChangeParams.ts
+++ b/packages/grid/_modules_/grid/models/params/gridVirtualPageChangeParams.ts
@@ -1,0 +1,25 @@
+/**
+ * Object passed as parameter of the virtual page change event.
+ */
+export interface GridVirtualPageChangeParams {
+  /**
+   * The index of the current page.
+   */
+  currentPage: number;
+  /**
+   * The index of the next page.
+   */
+  nextPage: number;
+  /**
+   * The index of the first row in the current page.
+   */
+  firstRowIndex: number;
+  /**
+   * The size of the virtual page.
+   */
+  pageSize: number;
+  /**
+   * Api that let you manipulate the grid.
+   */
+  api: any;
+}
--- a/packages/grid/data-grid/src/DataGrid.tsx
+++ b/packages/grid/data-grid/src/DataGrid.tsx
@@ -27,6 +27,7 @@ export type DataGridProps = Omit<
   | 'licenseStatus'
   | 'options'
   | 'onRowsScrollEnd'
+  | 'onVirtualPageChange'
   | 'pagination'
   | 'scrollEndThreshold'
 > & {
@@ -38,6 +39,7 @@ export type DataGridProps = Omit<
   disableMultipleColumnsSorting?: true;
   disableMultipleSelection?: true;
   onRowsScrollEnd?: undefined;
+  onVirtualPageChange?: undefined;
   pagination?: true;
 };
 
@@ -180,6 +182,19 @@ DataGrid.propTypes = {
         [
           `Material-UI: \`onRowsScrollEnd\` is not a valid prop.`,
           'onRowsScrollEnd is not available in the MIT version.',
+          '',
+          'You need to upgrade to the XGrid component to unlock this feature.',
+        ].join('\n'),
+      );
+    }
+    return null;
+  }),
+  onVirtualPageChange: chainPropTypes(PropTypes.any, (props: any) => {
+    if (props.onVirtualPageChange != null) {
+      return new Error(
+        [
+          `Material-UI: \`onVirtualPageChange\` is not a valid prop.`,
+          'onVirtualPageChange is not available in the MIT version.',
           '',
           'You need to upgrade to the XGrid component to unlock this feature.',
         ].join('\n'),
--- a/packages/grid/x-grid/src/tests/events.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/events.XGrid.test.tsx
@@ -20,6 +20,7 @@ import {
 } from '@material-ui/x-grid';
 import { getCell, getColumnHeaderCell, getRow } from 'test/utils/helperFn';
 import { spy } from 'sinon';
+import { useData } from 'packages/storybook/src/hooks/useData';
 
 describe('<XGrid /> - Events Params', () => {
   // TODO v5: replace with createClientRender
@@ -68,6 +69,16 @@ describe('<XGrid /> - Events Params', () => {
     return (
       <div style={{ width: 300, height: 300 }}>
         <XGrid apiRef={apiRef} {...baselineProps} {...props} />
+      </div>
+    );
+  };
+
+  const TestVirtualization = (props: Partial<XGridProps>) => {
+    apiRef = useGridApiRef();
+    const data = useData(50, 10);
+    return (
+      <div style={{ width: 300, height: 300 }}>
+        <XGrid rows={data.rows} columns={data.columns} apiRef={apiRef} {...props} />
       </div>
     );
   };
@@ -283,5 +294,24 @@ describe('<XGrid /> - Events Params', () => {
     gridWindow.scrollTop = 12345;
     gridWindow.dispatchEvent(new Event('scroll'));
     expect(handleOnRowsScrollEnd.callCount).to.equal(1);
+  });
+
+  it('call onVirtualPageChange when the virtual page changes', () => {
+    const handleOnVirtualPageChange = spy();
+    const { container } = render(
+      <div style={{ width: 300, height: 300 }}>
+        <TestVirtualization onVirtualPageChange={handleOnVirtualPageChange} />
+      </div>,
+    );
+    const gridWindow = container.querySelector('.MuiDataGrid-window');
+    // arbitrary number to make sure that the bottom of the grid window is reached.
+    gridWindow.scrollTop = 12345;
+    gridWindow.dispatchEvent(new Event('scroll'));
+
+    expect(handleOnVirtualPageChange.callCount).to.equal(1);
+    expect(handleOnVirtualPageChange.lastCall.args[0].currentPage).to.equal(15);
+    expect(handleOnVirtualPageChange.lastCall.args[0].nextPage).to.equal(16);
+    expect(handleOnVirtualPageChange.lastCall.args[0].firstRowIndex).to.equal(45);
+    expect(handleOnVirtualPageChange.lastCall.args[0].pageSize).to.equal(3);
   });
 });
--- a/docs/src/pages/components/data-grid/events/events.json
+++ b/docs/src/pages/components/data-grid/events/events.json
@@ -181,5 +181,9 @@
   {
     "name": "columnVisibilityChange",
     "description": "Fired when a column visibility changes. Called with a GridColumnVisibilityChangeParams object."
+  },
+  {
+    "name": "virtualPageChange",
+    "description": "Fired when a the virtual page changes. Called with a GridVirtualPageChangeParams object."
   }
 ]


packages/grid/_modules_/grid/models/params/gridVirtualPageChangeParams.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export interface GridVirtualPageChangeParams {
+  /**
+   * The index of the current page.
+   */
+  currentPage: number;
+  /**
+   * The index of the next page.
+   */
+  nextPage: number;
```
oliviertassinari, line 12: I don't think that we should expose these params. They are specific to the current virtualization implementation.

```suggestion
```

packages/grid/_modules_/grid/models/params/gridVirtualPageChangeParams.ts
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * The index of the first row in the current page.
+   */
+  firstRowIndex: number;
+  /**
+   * The size of the virtual page.
+   */
+  pageSize: number;
```
oliviertassinari, line 20: This seems more adequate

```suggestion
  /**
   * The index of the last row in the viewport.
   */
  lastRowIndex: number;
```

packages/grid/_modules_/grid/models/params/gridVirtualPageChangeParams.ts
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * The index of the current page.
+   */
+  currentPage: number;
+  /**
+   * The index of the next page.
+   */
+  nextPage: number;
+  /**
+   * The index of the first row in the current page.
```
oliviertassinari, line 14:
```suggestion
   * The index of the first row in the viewport.
```

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -27,6 +27,7 @@ export type DataGridProps = Omit<
   | 'licenseStatus'
   | 'options'
   | 'onRowsScrollEnd'
+  | 'onVirtualPageChange'
```
oliviertassinari, line 30: What about naming it?

```suggestion
  | 'onViewportRowChange'
```

Per the `GRID_VIEWPORT_ROW_CHANGE` event name proposal in https://github.com/mui-org/material-ui-x/pull/1715#pullrequestreview-689902319.

oliviertassinari (changes requested):

packages/grid/data-grid/src/DataGrid.tsx
DanailH: Updated

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -287,4 +298,21 @@ describe('<XGrid /> - Events Params', () => {
...
+      </div>,
+    );
+    const gridWindow = container.querySelector('.MuiDataGrid-window');
+    // scroll 10 rows.
+    gridWindow.scrollTop = 12345;
+    gridWindow.dispatchEvent(new Event('scroll'));
+
+    expect(handleOnVirtualPageChange.callCount).to.equal(1);
+    expect(handleOnVirtualPageChange.lastCall.args[0].firstRowIndex).to.equal(48);
+    expect(handleOnVirtualPageChange.lastCall.args[0].lastRowIndex).to.equal(50);
```
oliviertassinari, line 316: This looks suspicious. The default row height is 52 pixels. We have 300 pixels in height, so space for 5.7 rows. Removing the header, I would expect that we get 
```suggestion
    expect(handleOnVirtualPageChange.lastCall.args[0].firstRowIndex).to.equal(46 = 50 - 1 - Math.ceil((300 - 52 - 2)/52));
    expect(handleOnVirtualPageChange.lastCall.args[0].lastRowIndex).to.equal(50);
```

oliviertassinari (changes requested):
For the tests, we have more related to the virtualization inside `rows.XGrid.test.tsx`. Would it be a better place for them? Also, I think that we should better control the rendering parameters. The same feedback as @m4theushw did on the PR of @flaviendelangle https://github.com/mui-org/material-ui-x/pull/1986#discussion_r662678044. IMHO, it's easier to resonate based on the size of the viewport, rather than the size of the whole data grid container.

Otherwise, for the tests, I think that we should add more. For instance, we could assert that the initial event is correct, that if a row is visible by 1 pixel, it's included in the viewport. That if we scrollTop = 10 pixels, it triggers (not need to go to the bottom).

oliviertassinari:
Feedback I did in the last O3: we can't add a demo for this feature until #1911 is fixed. The data grid is displaying rows that have nothing to do with what the given scroll top should display. To make progress here, we would need to work in TDD.

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
DanailH: But that event is still being emitted when the virtualization changes, not when you scroll. Basically, it won't emit values such as `0-4`, `1-5`, `2-6` but rather `6-10`, `10-14`, `14-18` etc.

I think we should first agree on the behaviour we would want.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -144,6 +149,17 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
         top: containerProps.isVirtualized ? rzScrollTop : scrollTop,
       };
 
+      if (page !== nextPage) {
+        const newPage = page < nextPage ? nextPage + 1 : nextPage;
+        const nextPageFirstRowIndex = newPage * containerProps.viewportPageSize;
```
dtassone, line 154: instead of recalculating the first row index, you can get it from the rendering state `state.rendering`

dtassone: which you can just pass as argument

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -405,4 +422,5 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
     GRID_SCROLL,
     preventViewportScroll,
   );
+  useGridApiOptionHandler(apiRef, GRID_VIEWPORT_ROW_CHANGE, options.onViewportRowChange);
```
dtassone, line 425: `onViewportRowChange` => `onViewportRowsChange`

oliviertassinari:
```suggestion
  useGridApiOptionHandler(apiRef, GRID_VIEWPORT_ROWS_CHANGE, options.onViewportRowsChange);
```

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
DanailH: The thing is that I only have access to the current rendering state and I need to send the next rendering state values, not the current `page` but the `nextPage`. Even if I used the values from the `state.rendering` I would still need to calculate them. Also, the scroll direction (up or down) makes a difference.

dtassone: you should  trigger the event after the state is updated

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
oliviertassinari: > I think we should first agree on the behaviour we would want.

We have started this discussion in https://github.com/mui-org/material-ui-x/pull/1715#issuecomment-866723993. I didn't realize this wasn't clear yet. I mean, I thought that I answered the question and that you upvoted to confirm.

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -287,4 +298,21 @@ describe('<XGrid /> - Events Params', () => {
     gridWindow.dispatchEvent(new Event('scroll'));
     expect(handleOnRowsScrollEnd.callCount).to.equal(1);
   });
+
+  it('call onViewportRowsChange when the viewport rows change', () => {
+    const handleOnViewportRowsChange = spy();
```
oliviertassinari, line 303:
```suggestion
    const handleViewportRowsChange = spy();
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
DanailH: Ok, not a problem, I was just still thinking in the context of the virtualization.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
DanailH: This will be removed altogether as it will be decoupled from the virtualization.

oliviertassinari: Using the state that determines what to render is an interesting idea (proposed by @dtassone), with a high potential. I mean, conceptually speaking, the virtualization needs to know exactly the same information as we want to fire in the event (plus the overscan).

Sure, the virtualization currently doesn't compute this state right: #1911, but it will need to. In react-virtualized, the component uses this information under [onRowsRendered](https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md) with the viewport start and end, as well as the overscan. It's then [used by](https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/InfiniteLoader/InfiniteLoader.js#L119) the `InfiniteLoader` to have a rich loading logic.

We get the same API in react-window, under [onItemsRendered](https://react-window.vercel.app/#/api/FixedSizeList). 

Looking at the source of mui-plus, it seems to be done right: https://github.com/Janpot/mui-plus/blob/11a4cb121e7e3d3ad0b5c1190e23927313a9abc3/packages/mui-plus/src/utils/virtualization.ts#L69. It also relevant to mention that for the more advanced use cases, variable height and dynamic height, it's doesn't seem realistic to duplicate the logic.

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -451,3 +451,9 @@ export const GRID_STATE_CHANGE = 'stateChange';
  * @event
  */
 export const GRID_COLUMN_VISIBILITY_CHANGE = 'columnVisibilityChange';
+
+/**
+ * Fired when a the viewport rows change. Called with a [[GridViewportRowsChange]] object.
```
m4theushw, line 456:
```suggestion
 * Fired when the viewport rows change. Called with a [[GridViewportRowsChange]] object.
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -287,4 +298,21 @@ describe('<XGrid /> - Events Params', () => {
...
+    const { container } = render(
+      <div style={{ width: 300, height: 300 }}>
+        <TestVirtualization onViewportRowsChange={handleOnViewportRowsChange} />
+      </div>,
+    );
+    const gridWindow = container.querySelector('.MuiDataGrid-window');
+    // scroll 10 rows.
+    gridWindow.scrollTop = 12345;
```
m4theushw, line 311: We could call `scrollToIndexes` instead.

oliviertassinari: The risk would be that `scrollToIndexes` doesn't behave correctly, or we don't understand how it behaves at a deep enough level to write/read the test cae.

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
DanailH: Updated

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
DanailH: Updated

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
DanailH: I updated that based on the changed functionality

oliviertassinari (commented):
@DanailH Could you use this case:

```jsx
import { XGrid } from '@material-ui/x-grid';
import { useDemoData } from '@material-ui/x-grid-data-generator';

export default function XGridDemo() {
  const { data } = useDemoData({
    dataSet: 'Commodity',
    rowLength: 100000,
    editable: true,
  });

  return (
    <div style={{ height: 520, width: '100%' }}>
      <XGrid
        {...data}
        loading={data.rows.length === 0}
        rowHeight={38}
        checkboxSelection
        onViewportRowsChange={(params) => console.log('onViewportRowsChange', params)}
        disableSelectionOnClick
      />
    </div>
  );
}
```

and test that it supports all the required cases? Here is one example where it doesn't meet the expected outcome:

https://user-images.githubusercontent.com/3165635/127188314-93856985-b906-452a-9c1b-59e4e0ca46ee.mp4

Regarding what should be the expected outcome, 1. I would be in favor of saying whatever is the scroll-top value which would mean not respecting what's displayed, but at least, it would eventually be correct once we fix the rendering. 2. Now, if it's simpler to have it based on the rendering state, then why not. 3. We could also give up on the feature and wait for #1911 or #1933 to complete the effort (they can be considered blockers). Merging this PR as a first step in the right direction, with the right API and a beginning of a support (but noting that can be used by developers yet, so keeping it undocumented).

packages/grid/_modules_/grid/models/params/gridViewportRowsChange.ts
```diff
@@ -0,0 +1,17 @@
...
+  /**
+   * The index of the first row in the viewport.
+   */
+  firstRowIndex: number;
+  /**
+   * The index of the last row in the viewport.
+   */
+  lastRowIndex: number;
+  /**
+   * Api that let you manipulate the grid.
```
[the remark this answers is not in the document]
DanailH: I'll take care of that as part of https://github.com/mui-org/material-ui-x/issues/2189

DanailH:
> Regarding what should be the expected outcome, 1. I would be in favor of saying whatever is the scroll-top value which would mean not respecting what's displayed, but at least, it would eventually be correct once we fix the rendering. 2. Now, if it's simpler to have it based on the rendering state, then why not. 3. We could also give up on the feature and wait for #1911 or #1933 to complete the effort (they can be considered blockers). Merging this PR as a first step in the right direction, with the right API and a beginning of support (but noting that can be used by developers yet, so keeping it undocumented).

@oliviertassinari I'm aware of the limitation. The thing is that if we use the rendering state the part from it we need is the `renderingZoneScroll` but that one is cleared every time the virtual page changes. I would suggest just rely on the real scroll position and once the "jump" is fixed this feature will also start showing the correct values.

Otherwise, we again come to the fact that we need to rely on the virtualization implementation and not show the rows from `1-5` then `2-6` but `1-5` then `6-10`. Anything else would be just a way to make the current implementation work with the false values from the rendering state.

Regarding not documenting it - the only mentioned is the one in the `x-grid.md`. We can keep it as is for now and add a demo later once we know it works as expected.

oliviertassinari:
@DanailH Ok, I have spent 1/2 hour looking at this, to better understand what the problem is in #2228. The approach there has one downside, it notifies about the visible rows **and** the overscan, which we would probably want to ignore to minimize data loading, but it's functional, developers can rely on it. My guess is that we can improve this later on, once we refactor the virtualization.

DanailH:
> My guess is that we can fix the x2 loading later on, once we refactor the virtualization. What do you think about it?

Sorry I didn't get what x2 loading means. If it is about prefetching rows then I agree we can tackle that problem once we sort out the virtualization.

Btw you probably saw but the next row index is the one that is about to be displayed, so if rows from `0-5` are shown when the callback from `onViewportRowsChange` is called the result `1-6` will be that row number 6 is the one about to be shown.

oliviertassinari:
> Sorry I didn't get what x2 loading means.

Yeah right, I wasn't clear. Open https://deploy-preview-2228--material-ui-x.netlify.app/, there are 11 rows visible. Actual:

Capture d’écran 2021-07-28 à 15 48 22

Expected: `{ firstRowIndex: 0, lastRowIndex: 10 }`. 

> I agree we can tackle that problem once we sort out the virtualization.

For me, the most interesting part is that no matter what we use for the virtualization, by definition of the problem, we will always have to compute the first row index and last row index state, **and** use in a React render method to only render the current components in the viewport. It's why I tried on using `state.rendering.renderContext.firstRowIdx` in #2228. You can check the source of the benchmark source in https://github.com/mui-org/material-ui-x/issues/1933, AFAIK they all have this notion.

> Btw you probably saw but the next row index is the one that is about to be displayed, so if rows from 0-5 are shown when the callback from onViewportRowsChange is called the result 1-6 will be that row number 6 is the one about to be shown.

I didn't notice it, where do you see this?

---

Regarding going forward, we can 1. merge the current changes, but it's not usable by developers. we would need to come back to it. 2. explore further on using scrollTop, basically use the same logic as in scrollToIndex, we might be able to salvage it once we fix the virtualization, but developers can't rely on it yet, it's unsynchronized from what's visible. 3. we continue forward in the direction explore in #2228, we clean it up, it can be used by developers. The downside is that developers would have no control over what's fetched

DanailH:
Yes, I saw the exploration PR. The problem is `state.rendering.renderContext.firstRowIdx` doesn't show what is currently visible, it shows what is on the current virtual page. So these values only update when the virtual page changes (what I used in the initial implementation). It's just how currently the virtualization works. Changing those values, as far as I know, will break how virtualization works.

That is why I opted for directly calculating those myself by checking the `scrollTop`.

The deployment fails but if it builds you can check here https://deploy-preview-2038--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-rows--test-on-viewport-rows-change&globals=measureEnabled:false to see the result (open the console).

oliviertassinari:
@DanailH Yeap, I agree, these are fine concerns. I think that it really depends on the strategy we want to take on this problem.

> The deployment fails but if it builds you can check here https://deploy-preview-2038--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-rows--test-on-viewport-rows-change&globals=measureEnabled:false to see the result (open the console).

The link is no longer working for me, I didn't get a notification to access time. Netlify has permalinks. I could find it back:  https://6102ab9375027e1bd3e92749--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-rows--test-on-viewport-rows-change with the commit. The callback is not usable for a developer but the test case looks correct, so it's a step forward.

I would personally go with 3. The results wouldn't really be great because the callback won't have exactly the right values, it would be off by a couple of rows. However, the primary use case for this callback seems: lazy load data, loading 20 instead of 10 rows seems to be an acceptable downside, some might even argue that it's needed anyway for over scanning. 

I'm happy with any of these 3 options.

DanailH:
Alright I've updated the implementation so now it matches what we discussed on Monday and what has been explored there  #2228

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -97,6 +97,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onRowsScrollEnd</span> | <span class="prop-type">(param: GridRowScrollEndParams) => void</span> |   | Callback fired when scrolling to the bottom of the grid viewport. |
 | <span class="prop-name">onSelectionModelChange</span> | <span class="prop-type">(model: GridSelectionModel) => void</span> |   | Callback fired when the selection state of one or multiple rows changes. |
 | <span class="prop-name">onSortModelChange</span> | <span class="prop-type">(model: GridSortModel) => void</span> |   | Callback fired when the sort model changes before a column is sorted. |
+| <span class="prop-name">onViewportRowsChange</span> | <span class="prop-type">(param: GridViewportRowsChangeParams) => void</span> |   | Callback fired when the virtual page changes. |
```
oliviertassinari, line 100:
```suggestion
| <span class="prop-name">onViewportRowsChange</span> | <span class="prop-type">(params: GridViewportRowsChangeParams) => void</span> |   | Callback fired when the rows in the viewport change. |
```

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -440,3 +440,9 @@ export const GRID_STATE_CHANGE = 'stateChange';
  * @event
  */
 export const GRID_COLUMN_VISIBILITY_CHANGE = 'columnVisibilityChange';
+
+/**
+ * Fired when the viewport rows change. Called with a [[GridViewportRowsChange]] object.
```
oliviertassinari, line 445:
```suggestion
 * Fired when the rows in the viewport change. Called with a [[GridViewportRowsChange]] object.
```

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -413,6 +414,10 @@ export interface GridOptions {
    * Callback fired when the state of the grid is updated.
    */
   onStateChange?: (params: any) => void;
+  /**
+   * Callback fired when the virtual page changes.
```
oliviertassinari, line 418:
```suggestion
   * Callback fired when the rows in the viewport change.
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -278,13 +291,47 @@ describe('<XGrid /> - Events Params', () => {
...
+      >
+        <TestVirtualization rowHeight={rowHeight} onViewportRowsChange={handleViewportRowsChange} />
+      </div>,
+    );
+    const gridWindow = container.querySelector('.MuiDataGrid-window');
+    // scroll 8 rows so that the renderContext is updated. To be changed to a scroll of 1 row.
+    gridWindow.scrollTop = pageSize * rowHeight * 2;
+    gridWindow.dispatchEvent(new Event('scroll'));
+    await waitFor(() => {
+      expect(handleViewportRowsChange.callCount).to.equal(2); // should be 1
```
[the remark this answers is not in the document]
DanailH: When I was thinking about it my assumption was that it should fire when you start scrolling, not on the initial render. We can have this as expected behaviour. My assumption was that you will always start with part of the data and would need to load more later when scrolling occurs?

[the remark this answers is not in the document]
DanailH: Updated. Added another expect to validate the first call.

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -278,13 +291,47 @@ describe('<XGrid /> - Events Params', () => {
...
+    const border = 1;
+    const pageSize = 4;
+    const handleViewportRowsChange = spy();
+    const { container } = render(
+      <div
+        style={{ width: 300, height: headerHeight + pageSize * rowHeight + offset + border * 2 }}
+      >
+        <TestVirtualization rowHeight={rowHeight} onViewportRowsChange={handleViewportRowsChange} />
+      </div>,
+    );
```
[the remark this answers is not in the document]
DanailH: Had to update the test because of this mistake.

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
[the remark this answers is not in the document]
oliviertassinari: >  My assumption was that you will always start with part of the data and would need to load more later when scrolling occurs?

How would a developer know how much data to load initially?

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -278,13 +287,48 @@ describe('<XGrid /> - Events Params', () => {
...
+    const rowHeight = 40;
+    const pageSize = 4;
+    const handleViewportRowsChange = spy();
+    const { container } = render(
+      <div style={{ width: 300, height: headerHeight + rowHeight * pageSize }}>
+        <TestVirtualization rowHeight={rowHeight} onViewportRowsChange={handleViewportRowsChange} />
+      </div>,
+    );
+
+    expect(handleViewportRowsChange.callCount).to.equal(1); // should be 1
```
oliviertassinari, line 311:
```suggestion
    expect(handleViewportRowsChange.callCount).to.equal(1);
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -278,13 +287,48 @@ describe('<XGrid /> - Events Params', () => {
...
+    expect(handleViewportRowsChange.lastCall.args[0].firstRowIndex).to.equal(0); // should be 1
```
oliviertassinari, line 312: Why 1?

```suggestion
    expect(handleViewportRowsChange.lastCall.args[0].firstRowIndex).to.equal(0);
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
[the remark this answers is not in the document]
DanailH: Yes, now its correct to expect 2.

packages/grid/_modules_/grid/hooks/features/infiniteLoader/useGridInfiniteLoader.ts
```diff
@@ -2,50 +2,103 @@ import * as React from 'react';
...
+      }
+    },
+    [apiRef, options, visibleColumns, containerSizes],
+  );
 
   const handleGridScroll = React.useCallback(() => {
-    if (!containerSizes) {
+    if (!props.onRowsScrollEnd && !props.onViewportRowsChange) {
       return;
     }
```
m4theushw, line 66: Why to check if the props are being used? The user could be listening to the event.

```suggestion
```

packages/grid/x-grid/src/tests/events.XGrid.test.tsx
```diff
@@ -303,13 +312,49 @@ describe('<XGrid /> - Events Params', () => {
...
+  });
+
+  it('call onViewportRowsChange when the viewport rows change', async () => {
+    const headerHeight = 40;
+    const rowHeight = 40;
+    const pageSize = 4;
+    const handleViewportRowsChange = spy();
+    const { container } = render(
+      <div style={{ width: 300, height: headerHeight + rowHeight * pageSize }}>
+        <TestVirtualization rowHeight={rowHeight} onViewportRowsChange={handleViewportRowsChange} />
```
m4theushw, line 332: Shouldn't we set the header height?

```suggestion
        <TestVirtualization headerHeight={headerHeight} rowHeight={rowHeight} onViewportRowsChange={handleViewportRowsChange} />
```

packages/grid/_modules_/grid/hooks/features/infiniteLoader/useGridInfiniteLoader.ts
```diff
@@ -2,50 +2,103 @@ import * as React from 'react';
...
+    if (props.onRowsScrollEnd) {
+      handleRowsScrollEnd(scrollPosition);
+    }
+  }, [props, apiRef, handleRowsScrollEnd]);
+
+  // TODO: Check if onViewportRowsChange works as expected once virtualization is reworked
+  React.useEffect(() => {
+    const renderContext = renderState.renderContext!;
+
+    if (!props.onViewportRowsChange || !renderContext) {
```
m4theushw, line 79:
```suggestion
    if (!renderContext) {
```

packages/grid/_modules_/grid/hooks/features/infiniteLoader/useGridInfiniteLoader.ts
```diff
@@ -2,50 +2,103 @@ import * as React from 'react';
...
 
     const scrollPosition = apiRef.current.getScrollPosition();
-    const scrollPositionBottom =
-      scrollPosition.top + containerSizes.windowSizes.height + options.scrollEndThreshold;
 
-    if (scrollPositionBottom < containerSizes.dataContainerSizes.height) {
-      isInScrollBottomArea.current = false;
+    if (props.onRowsScrollEnd) {
+      handleRowsScrollEnd(scrollPosition);
+    }
```
oliviertassinari, line 72: Same feedback as in https://github.com/mui-org/material-ui-x/pull/2038/files#r683870957

```suggestion
    handleRowsScrollEnd(scrollPosition);
```

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -305,4 +305,8 @@ export enum GridEvents {
    * Fired when a column visibility changes. Called with a [[GridColumnVisibilityChangeParams]] object.
    */
   columnVisibilityChange = 'columnVisibilityChange',
+  /**
+   * Fired when the rows in the viewport change. Called with a [[GridViewportRowsChange]] object.
+   */
+  gridViewportRowsChange = 'viewportRowsChange',
```
m4theushw, line 311: We don't use the `grid` prefix anymore.

```suggestion
  viewportRowsChange = 'viewportRowsChange',
```

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -551,6 +552,10 @@ export interface GridOptions {
    * @internal
    */
   onStateChange?: (params: any, event: MuiEvent<{}>, details?: any) => void;
+  /**
+   * Callback fired when the rows in the viewport change.
+   */
+  onViewportRowsChange?: (params: GridViewportRowsChangeParams) => void;
```
m4theushw, line 558:
```suggestion
  onViewportRowsChange?: (params: GridViewportRowsChangeParams, event: MuiEvent<{}>, details?: any) => void;
```

PR #2051, opened 2021-07-05
[DataGrid] Fix manual entry in date fields

### LINKED ISSUE #1920: [DataGrid] Entering value manually in a date field is not working

m4theushw:
Fixes #1920
Preview: https://deploy-preview-2051--material-ui-x.netlify.app/components/data-grid/editing/

The TODO comments can be addressed once #1955 is merged.

### DIFF
--- a/packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import InputBase, { InputBaseProps } from '@material-ui/core/InputBase';
-import { GRID_CELL_EDIT_PROPS_CHANGE } from '../../constants/eventsConstants';
 import { GridCellParams } from '../../models/params/gridCellParams';
 
 export function GridEditDateCell(props: GridCellParams & InputBaseProps) {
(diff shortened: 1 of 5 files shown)


packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
```diff
@@ -0,0 +1,78 @@
...
+      setValueState(newValue);
+
+      if (newValue === '') {
+        const editProps = { value: null };
+        // TODO use api.setEditCellProps
+        api.publishEvent(GRID_CELL_EDIT_PROPS_CHANGE, { id, field, props: editProps }, event);
+        return;
+      }
+
+      // TODO move to value parser
```
oliviertassinari, line 36: I'm not sure it would make a difference anymore since we have custom logic in the filtering for dates.

dtassone (approved):

PR #2075, opened 2021-07-06
[core] Configure Renovate and remove Dependabot

flaviendelangle:
Same config as https://github.com/mui-org/material-ui/pull/27003/files

Some lines are not usefull right now but I doesn't hurt so I think it's best to keep the two repo synchronized to limit the mental overhead.

I also added these changes : https://github.com/mui-org/material-ui/pull/27115, @oliviertassinari if the `allowedPostUpgradeCommands` is configured repo by repo you will also have to do it for this one :+1:


dtassone (approved):

oliviertassinari:
@flaviendelangle Thanks for the follow-up.

> Some lines are not usefull right now but I doesn't hurt so I think it's best to keep the two repo synchronized to limit the mental overhead.

This seems fair, however, the two repositories aren't completely identical. We need to handle `@storybook` here, we don't use it in the main repo (to incentive creating more docs)

flaviendelangle:
They may be improvements to do indeed.
Do you see other dependencies than the Storybook ? Now that we have a correct basic behavior, we can take the time to improve it.

oliviertassinari:
@flaviendelangle I can't think of anything else. I believe Storybook is the only difference between the x and the core repo. I mean, in the core, we have more packages only to make sure the integration with third-party works well e.g. react-spring, this is it.

PR #2080, opened 2021-07-06
[DataGrid] Do not emit GRID_FILTER_MODEL_CHANGE event in applyFilterLinkOperator when called from another filter update method

### LINKED ISSUE #2013: [core] Story Server Filter Via Props is broken

## Current Behavior 😯

The story currently crashes because `onFilterModelChange` is called with 0 items and the callback `onFilterChange` assumes that their is one item.

image

## Expected Behavior 🤔

The Story should not crash

## Steps to Reproduce 🕹

Go to [This Story](https://material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-filter--server-filter-via-props&globals=measureEnabled:false)

flaviendelangle:
Fixes #2013


dtassone:
that issue is fixed in #1909

PR #2084, opened 2021-07-06
[DataGrid] Make GridToolbarXXX props overridable

### LINKED ISSUE #2083: [DataGrid] GridToolbarFilterButton props spread doesn't follow other Toolbar components
<GridToolbarColumnsButton color="inherit" /> render a "inherit" button, <GridToolbarFilterButton color="inherit" /> render a "primary" button.

- [x ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯
GridToolbarColumnsButton props spreading comes before override, not after.

## Expected Behavior 🤔
Props spreading should come after.

## Your Environment 🌎

<details>
  <summary>System info</summary>

System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 14.17.2 - C:\Program Files\nodejs\node.EXE
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.64)
  npmPackages:
    @material-ui/core: ^5.0.0-beta.0 => 5.0.0-beta.0
    @material-ui/data-grid: ^4.0.0-alpha.33 => 4.0.0-alpha.33
    @material-ui/icons: ^5.0.0-beta.0 => 5.0.0-beta.0
    @material-ui/lab: ^5.0.0-alpha.38 => 5.0.0-alpha.38
    @material-ui/private-theming:  5.0.0-alpha.36
    @material-ui/styled-engine:  5.0.0-alpha.34
    @material-ui/styles: ^5.0.0-beta.0 => 5.0.0-beta.0
    @material-ui/system:  5.0.0-alpha.38
    @material-ui/types:  6.0.1
    @material-ui/unstyled:  5.0.0-alpha.39
    @material-ui/utils:  5.0.0-alpha.36
    @types/react: ^17.0.13 => 17.0.13
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.3.5 => 4.3.5
</details>

tifosiblack:
Fix #2083

### DIFF
--- a/packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
+++ b/packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
@@ -31,6 +31,9 @@ const useStyles = makeStyles(
 );
 
 export interface GridToolbarFilterButtonProps extends Omit<TooltipProps, 'title' | 'children'> {
+  /**
+   * The props used for each slot inside.
+   */
   componentProps?: {
     button?: ButtonProps;
   };
@@ -40,7 +43,7 @@ export const GridToolbarFilterButton = React.forwardRef<
   HTMLButtonElement,
   GridToolbarFilterButtonProps
 >(function GridToolbarFilterButton(props, ref) {
-  const { componentProps, ...rest } = props;
+  const { componentProps = {}, ...other } = props;
   const classes = useStyles();
   const apiRef = useGridApiContext();
   const options = useGridSelector(apiRef, optionsSelector);
@@ -101,7 +104,7 @@ export const GridToolbarFilterButton = React.forwardRef<
 
   const OpenFilterButtonIcon = apiRef!.current.components!.OpenFilterButtonIcon!;
   return (
-    <Tooltip title={tooltipContentNode} enterDelay={1000} {...rest}>
+    <Tooltip title={tooltipContentNode} enterDelay={1000} {...other}>
       <Button
         ref={ref}
         size="small"
@@ -113,7 +116,7 @@ export const GridToolbarFilterButton = React.forwardRef<
           </Badge>
         }
         onClick={toggleFilter}
-        {...componentProps?.button}
+        {...componentProps.button}
       >
         {apiRef!.current.getLocaleText('toolbarFilters')}
       </Button>
--- a/packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
+++ b/packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
@@ -3,7 +3,7 @@ import { Theme } from '@material-ui/core/styles';
 import { makeStyles } from '@material-ui/styles';
 import Badge from '@material-ui/core/Badge';
 import Button, { ButtonProps } from '@material-ui/core/Button';
-import Tooltip from '@material-ui/core/Tooltip';
+import Tooltip, { TooltipProps } from '@material-ui/core/Tooltip';
 import { capitalize } from '@material-ui/core/utils';
 import { gridColumnLookupSelector } from '../../hooks/features/columns/gridColumnsSelector';
 import { useGridSelector } from '../../hooks/features/core/useGridSelector';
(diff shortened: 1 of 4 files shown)


packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
```diff
@@ -95,7 +95,6 @@ export const GridToolbarFilterButton = React.forwardRef<HTMLButtonElement, Butto
       <Tooltip title={tooltipContentNode} enterDelay={1000}>
         <Button
           ref={ref}
-          {...props}
           onClick={toggleFilter}
```
flaviendelangle, line 98: I don't think we want to be able to totally skip `toggleFilter`
It should be after the props spread.
And not directly related to this PR, the `onClick` should either be removed from the typing of the component or called in `toggleFilter`. Right now the typing allows to pass a `onClick` but does nothing with it.

tifosiblack: I guess I can exclude it from ButtonProps

tifosiblack: I can do the same for other toolbar component

oliviertassinari (dismissed):
The convention is for the props to be spread on the root element and last. So for instance on the tooltip, not the button. Note that the extra props of the tooltip goes to the button but we probably don't want to talk about it a componentsProps.button would do it.

tifosiblack:
> The convention is for the props to be spread on the root element and last. So for instance on the tooltip, not the button. Note that the extra props of the tooltip goes to the button but we probably don't want to talk about it a componentsProps.button would do it.

title is required for tooltip, should I use `Omit<TooltipProps, 'title'>` or use `Partial<TooltipProps>` ?

oliviertassinari:
Omit sounds great

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
oliviertassinari: > I don't think we want to be able to totally skip toggleFilter

@flaviendelangle Agree, only with  #1403.

oliviertassinari:
Best to review with:

Capture d’écran 2021-07-07 à 01 24 00

tifosiblack:
> Best to review with:
> 
> Capture d’écran 2021-07-07 à 01 24 00

Indeed, I used prettier formatter, but it changed a lot the file whitespaces...

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
```diff
@@ -30,85 +30,96 @@ const useStyles = makeStyles(
...
+        ref={ref}
+        size="small"
+        color="primary"
+        aria-label={apiRef!.current.getLocaleText('toolbarFiltersLabel')}
+        startIcon={
+          <Badge badgeContent={counter} color="primary">
+            <OpenFilterButtonIcon />
+          </Badge>
+        }
+        onClick={toggleFilter}
```
flaviendelangle, line 118: Either call `componentProps.button?.onClick?.()` in `toggleFilter` or omit `toggleFilter` from the `componentProps.button` interface.

@oliviertassinari do we want to allow the override of any prop ?
For instance here, the user could override `startIcon` which is already customizable with the `OpenFilterButtonIcon` (but he currenctly can't remove the badge).

flaviendelangle (changes requested):

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
tifosiblack: Not an "easy" change after all ;)
IMO we should be able to change completely the appearance of the button as long as the behavior is kept.

flaviendelangle: I'm also in favor of letting the user change the  `startIcon`, and it's a modification the user could do by mistake.

oliviertassinari: > do we want to allow the override of any prop ?

@flaviendelangle Yes, I think so. Developers should be able to override `startIcon`.

The case of the event handlers is interesting. We don't have a policy around it yet. In the core, If I look at `SliderUnstyled.js`, `componentProps` overrides the event handlers. If I look at `SwitchUnstyled`, the event handlers of `componentProps` are merged.

cc @michaldudak and @mnajdova. Maybe we should merge?

michaldudak: We should definitively have consistency in this matter across the codebase. My preference for event handlers and classNames is to merge. I suppose it'll create less WTF moments among developers 😉

flaviendelangle (approved):

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
oliviertassinari: @flaviendelangle I have noticed that you merged without merging the onClick handlers. Was it on purpose? Is this something we should handle in a follow-up?

tifosiblack: I didn't see you already merged, I may add another PR to deal with event merging.

oliviertassinari: > I may add another PR to deal with event merging.

@tifosiblack Flavien is off until Monday, another PR he can review by the time he comes back would be awesome 👌

flaviendelangle: Definitely an error on my side

PR #2085, opened 2021-07-06
[XGrid] Add ability to disable reorder on some columns

### LINKED ISSUE #1161: [XGrid] Disable reordering on a specific column
First off: great work with this product. 

I have a custom checkbox column - I am not using the checkbox built ins. This column should not be reorderable, however there is no ColDef option to specify that. The other areas of column functionality are optional (resizable, sortable, etc). It makes sense to also have a "reorderable"  col def option.

### LINKED ISSUE #1768: [XGrid] Disable checkbox column draggable
The checkbox column is draggable and it shouldn't be. Is there any config on the table that may prevent this? I see it happening in the demo as well so I'm guessing its an issue with XGrid and not my setup. You can see the column being dragged here, using the same code for the table configs.

https://codesandbox.io/s/gttt1?file=/demo.tsx

Thanks!

flaviendelangle:
Closes #1161 
Closes #1768 

Implemented behavior :

- A column with `disableReorder: true` can't be drag&dropped
- If the first visible column of the grid has `disableReorder: true`, we can't drag&drop a column left of it
- If the last visible column of the grid has `disableReorder: true`, we can't drag&drop a column right of it

Right now the behavior is at the drag&drop level, I do not block the updates made with `apiRef`.
Do we want to do it?

Do you prefer? 

```jsx
  /**
   * If `true`, this column can not be reordered
   * @default false
   */
  disableReorder?: boolean;
```

~OR~

```
  /**
   * If `true`, this column can be reordered
   * @default true
   */
  orderable?: boolean;
```

We have both (`disableExport` but `resizable`), which is not great from a user perspective I think.

I did not implement the behavior on the column panel yet. Preview: https://deploy-preview-2085--material-ui-x.netlify.app/components/data-grid/columns/#column-reorder.


oliviertassinari:
For the name of the property, I would refer to https://material-ui.com/guides/api/#property-naming

packages/storybook/src/stories/grid-reorder.stories.tsx
```diff
@@ -23,3 +25,27 @@ export const ReorderSmallDataset = () => {
     </div>
   );
 };
+
+export const DisableReorderOnSomeColumn = () => {
```
oliviertassinari, line 29: Shouldn't we have a demo for this feature, as https://www.ag-grid.com/angular-grid/column-moving/#lock-position?

---

**Off topic**. It's IMHO, why we shouldn't work storybook first, but docs first. In the past, I was collecting a list of issues that we're creating because we were working Storybook first.

- https://github.com/mui-org/material-ui-x/issues/98 source: missing SSR support
- https://github.com/mui-org/material-ui-x/issues/209 source: isolated environment
- keyboard even conflicts when mounting two data grid on the same page (I can't find the GitHub issue related to it). source: isolated environment
- https://github.com/mui-org/material-ui-x/issues/895 source: isolated environment
- https://github.com/mui-org/material-ui-x/issues/974 source: isolated environment.

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -218,10 +227,151 @@ describe('<XGrid /> - Reorder', () => {
...
+        { field: 'desc', disableReorder: true },
+        { field: 'type' },
+      ];
+
+      const Test = () => {
+        apiRef = useGridApiRef();
+
+        return (
+          <div style={{ width: 300, height: 300 }}>
+            <XGrid apiRef={apiRef} rows={rows} columns={columns} onPageChange={() => {}} />
```
oliviertassinari, line 356:
```suggestion
            <XGrid apiRef={apiRef} rows={rows} columns={columns} />
```

x the new test case. I'm taking care of the old one in #2088

packages/storybook/src/stories/grid-reorder.stories.tsx
flaviendelangle: You're right, I'll add a demo

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
flaviendelangle: > For the name of the property, I would refer to https://material-ui.com/guides/api/#property-naming

@oliviertassinari 

This page is for props, I don't think the shorthand argument really applies here.
But I am 100% in favor of consistency. 

Both approaches can make sense

- Always favor `default: false` (meaning that we should have `disableResize` instead of `resizable`)
or
- Always favor the affirmative form (meaning that we should have `exportable` instead of `disableExport`)

Most options are in the affirmative form so I think we should rename `disableExport` into `exportable` (my PR has not yet been merged) and on this PR `orderable` ?

flaviendelangle:
> For the name of the property, I would refer to https://material-ui.com/guides/api/#property-naming

@oliviertassinari 

This page is for props, I don't think the shorthand argument really applies here.
But I am 100% in favor of consistency. 

Both approaches can make sense

- Always favor `default: false` (meaning that we should have `disableResize` instead of `resizable`)
or
- Always favor the affirmative form (meaning that we should have `exportable` instead of `disableExport`)

Most options are in the affirmative form so I think we should rename `disableExport` into `exportable` (my PR has not yet been merged) and on this PR `orderable` ?

flaviendelangle:
@mui-org/x  

I have two behavior points to discuss

AG Grid differentiate
1) the columns that can't be moved and that are automatically placed at the beginning of the grid (`lockPosition: true`)
2) the column that can't be moved but for which we can move columns around it (`suppressMovable: true`)

```
The Age column is locked as the first column in the scrollable area of the grid. It is not possible to move this column, or have other columns moved over it to impact its position. As a result the Age column marks the beginning of the scrollable area regardless of its position.
The Athlete column has moving suppressed. It is not possible to move this column, but it is possible to move other columns around it.
```

Do you think it's something we want to implement ?
I did a single solution with a bit more of magic since I'm checking the position of the target column to decide if the user is allowed to move the drag column on it.

---

Do we want to also block the re-ordering when it is done with `apiRef.current.setColumnIndex` ? 
Right now I am only blocking the drag&drop.

oliviertassinari:
> Both approaches can make sense

Yeah, right, the shorthand advantage is specific to the props. I will update this section to mention the second advantage: as a developer, you always know what the default value is by looking at the name of the prop.

> Do you think it's something we want to implement ?

I think that the difference is not negligeable. If we wanted to be lazy and maximize flexibility, we could say to the developers:

1. `suppressMovable: true`: add event handlers on the column headers and use #1403. 
2. `lockPosition: true` it could be implemented with a controlled API.

I think that we should start small, only do the minimum feature to solve the pain reported.

flaviendelangle:
https://github.com/mui-org/material-ui-x/pull/2085#issuecomment-875456845

That sounds good for me :+1:

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -206,10 +215,151 @@ describe('<XGrid /> - Reorder', () => {
...
+        );
+      };
+
+      render(<Test />);
+      expect(getColumnHeadersTextContent()).to.deep.equal(['brand', 'desc', 'type']);
+      const dragCol = getColumnHeaderCell(1).firstChild!;
+      const targetCol = getColumnHeaderCell(0).firstChild!;
+
+      fireEvent.dragStart(dragCol);
+      fireEvent.dragEnter(targetCol);
```
m4theushw, line 250: If the column can't be reordered we don't even allow the user to start dragging it. We could check if the `draggable` attribute is false or if `GRID_COLUMN_HEADER_DRAGGING_CSS_CLASS` was not added.

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -206,10 +215,151 @@ describe('<XGrid /> - Reorder', () => {
...
+          <div style={{ width: 300, height: 300 }}>
+            <XGrid apiRef={apiRef} rows={rows} columns={columns} />
+          </div>
+        );
+      };
+
+      render(<Test />);
+      expect(getColumnHeadersTextContent()).to.deep.equal(['brand', 'desc', 'type']);
+      const dragCol = getColumnHeaderCell(1).firstChild!;
+      const targetCol = getColumnHeaderCell(0).firstChild!;
```
m4theushw, line 317: The index of the last visible column is 2, no?

```suggestion
      const targetCol = getColumnHeaderCell(2).firstChild!;
```

packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx
```diff
@@ -103,27 +103,33 @@ export const useGridColumnReorder = (apiRef: GridApiRef): void => {
...
+        const targetColVisibleIndex = apiRef.current.getColumnIndex(params.field, true);
+        const targetCol = apiRef.current.getColumn(params.field);
+        const dragColIndex = apiRef.current.getColumnIndex(dragColField, false);
+        const visibleColumnAmount = apiRef.current.getVisibleColumns().length;
+
+        const canBeReordered =
+          !targetCol.disableReorder ||
+          (targetColVisibleIndex > 0 && targetColVisibleIndex < visibleColumnAmount - 1);
+
+        const cursorMoveDirectionX = getCursorMoveDirectionX(cursorPosition.current, coordinates);
```
m4theushw, line 119: For another PR:

This `cursorMoveDirectionX` is wrong. It's always "right" the first time. We need to save the cursor position when the user starts dragging

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -24,6 +24,25 @@ import {
 
 const isJSDOM = /jsdom/.test(window.navigator.userAgent);
 
+function createDragOverEvent(target: ChildNode) {
+  const dragOverEvent = createEvent.dragOver(target);
+  // Safari 13 doesn't have DragEvent.
+  // RTL fallbacks to Event which doesn't allow to set these fields during initialization.
+  Object.defineProperty(dragOverEvent, 'clientX', { value: 1 });
```
m4theushw, line 31: For another PR:

Add an argument to allow to change the `clientX`. We use this value to check if the user is moving to the left or right.

packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx
```diff
@@ -103,27 +103,33 @@ export const useGridColumnReorder = (apiRef: GridApiRef): void => {
...
         }
 
         cursorPosition.current = coordinates;
       }
     },
-    [apiRef, dragCol, logger],
+    [apiRef, dragColField, logger],
   );
 
   const handleDragEnd = React.useCallback(
```
m4theushw, line 135: Small bug here. We need to check if the user started to drag the cell.

https://github.com/mui-org/material-ui-x/blob/cbd2179dc605d8f07bca15e60ef89ffe5fd600d4/packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx#L96-L98

jLW0lPu6QF

oliviertassinari (commented):
I have tested the behavior with a `disableReorder` on a column in the middle: https://codesandbox.io/s/material-demo-forked-2c26p?file=/demo.js. It's weird. But 👍  for ignoring this for now.

I mean, the new prop allows preventing end-users to start the drag and drop from the column, it's already great. It comes with a bonus behavior (that it likely outside of the scope of  `disableReorder`) of preventing other columns to go past it.

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -206,10 +215,151 @@ describe('<XGrid /> - Reorder', () => {
...
+      let apiRef: GridApiRef;
+      const rows = [{ id: 0, brand: 'Nike' }];
+      const columns = [
+        { field: 'brand' },
+        { field: 'desc', disableReorder: true },
+        { field: 'type' },
+      ];
+
+      const Test = () => {
+        apiRef = useGridApiRef();
```
oliviertassinari, line 340: We don't the apiRef in this test and in the other, should we kill it?

packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx
flaviendelangle: What is the issue on the video ?
I can check if `originColumnIndex.current` is not null.

m4theushw: The age column is not reorderable but I select its text I can reorder. We just have to copy the same 3 lines from `handleDragOver`.

dtassone (approved):

packages/grid/x-grid/src/tests/reorder.XGrid.test.tsx
```diff
@@ -206,10 +215,151 @@ describe('<XGrid /> - Reorder', () => {
...
+      };
+
+      render(<Test />);
+      expect(getColumnHeadersTextContent()).to.deep.equal(['brand', 'desc', 'type']);
+      const dragCol = getColumnHeaderCell(1).firstChild! as HTMLElement;
+      const targetCol = getColumnHeaderCell(0).firstChild!;
+
+      fireEvent.dragStart(dragCol);
+
+      expect(dragCol.classList.contains(GRID_COLUMN_HEADER_DRAGGING_CSS_CLASS)).to.equal(false);
```
m4theushw, line 251:
```suggestion
      expect(dragCol).to.have.attribute('draggable', 'false');
      expect(dragCol).not.to.have.class(GRID_COLUMN_HEADER_DRAGGING_CSS_CLASS);
```

m4theushw (approved):
Good job!

PR #2086, opened 2021-07-06
[core] Renovate : Group storybook updates

flaviendelangle:


oliviertassinari (commented):
Initially, @dtassone wanted to handle Storybook updates less frequently, so we asked dependabot to ignore it (I believe it's no like in renovate where we can pick the frequency per entry). But did we ever update it? If we want to update Storybook once a week, happy with the changes 👍

oliviertassinari:
Also, the approach changed in https://github.com/mui-org/material-ui/pull/27115#issuecomment-874696040

flaviendelangle:
@oliviertassinari behavior updated to keep in sync with the core

PR #2087, opened 2021-07-06
[test] Test the validation before saving a value

m4theushw:
I just noticed that when calling `commitCellChange` we're not checking if there's an error. This could cause the invalid value being commited when it shouldn't. This situation may occur with custom edit components.


packages/grid/x-grid/src/tests/editRows.XGrid.test.tsx
```diff
@@ -577,4 +577,32 @@ describe('<XGrid /> - Edit Rows', () => {
...
+    const { setProps } = render(<TestCase editRowsModel={{}} />);
+    const cell = getCell(1, 0);
+    cell.focus();
+    fireEvent.doubleClick(cell);
+    const input = cell.querySelector('input')!;
+    expect(input.getAttribute('aria-invalid')).to.equal(null);
+    fireEvent.change(input, { target: { value: 'n' } });
+    setProps({ editRowsModel: { 1: { brand: { error: true, value: 'n' } } } });
+    fireEvent.keyDown(input, { key: 'Enter' });
+    expect(input.getAttribute('aria-invalid')).to.equal('true');
```
oliviertassinari, line 591: To improve the debug experience?

```suggestion
    expect(input).to.have.attribute('aria-invalid', 'true');
```

oliviertassinari (commented):
Should `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED` fire if there is an error or not? The name of the event is weird. It says commit**ted** even if it wasn't yet or might not be.

m4theushw:
> Should `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED` fire if there is an error or not? The name of the event is weird. It says commit**ted** even if it wasn't yet or might not be.

@oliviertassinari True, `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED` shouldn't be fired. I pushed one more commit fixing this and also allowing developers to stop the propagation of this event and `GRID_CELL_EDIT_PROPS_CHANGE` . 

In #1955 I changed all edit components to call the API directly, but by doing that the events would not be published and developers couldn't stop the propagation. To fix this, I updated `commitCellChange` and `setEditCellProps` here to be mere event publishers. The purpose of each one is as following:
- `setEditCellProps` publishes `GRID_CELL_EDIT_PROPS_CHANGE` only
- `commitCellChange` checks if there's an error and, if it doesn't, then publishes `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED`. The commit logic was moved to a handler that subscribes to this event, so developers can stop it.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -137,14 +130,15 @@ export function useGridEditRows(apiRef: GridApiRef) {
     [options.isCellEditable],
   );
 
-  const setEditCellProps = React.useCallback(
+  const changeCellEditProps = React.useCallback(
```
dtassone, line 133: `setEditCellProps`, `changeCellEditProps` => pretty confusing

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -168,6 +162,16 @@ export function useGridEditRows(apiRef: GridApiRef) {
     [apiRef, forceUpdate, logger, setGridState],
   );
 
+  const handleCellEditCellProps = React.useCallback(
```
dtassone, line 165: handleCellEditCellProps => handleEditCellProps

packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx
```diff
@@ -25,7 +25,7 @@ export function GridEditInputCell(props: GridCellParams & InputBaseProps) {
       const newValue = event.target.value;
       const editProps = { value: newValue };
       setValueState(newValue);
-      api.setEditCellProps({ id, field, props: editProps });
+      api.changeCellEditProps({ id, field, props: editProps }, event);
```
dtassone, line 28: I don't get the purpose of this new method

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: Yeah, the names are a bit weird, we can discuss. Before #1824, some components were directly publishing `GRID_CELL_EDIT_PROPS_CHANGE`, but that is not good because TypeScript can't suggested the arguments. This new method is just a helper to publish the event. The actual logic of updating the state is in `setEditCellProps`. I can't just call `setEditCellProps` because the prop won't be called.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -52,13 +52,7 @@ export function useGridEditRows(apiRef: GridApiRef) {
     if (params.cellMode === 'view') {
       return;
     }
-
-    const cellCommitParams = apiRef.current.getEditCellPropsParams(params.id, params.field);
-    if (!cellCommitParams.props.error) {
-      // We commit the change when there is no error
```
dtassone, line 58: how are you handling this now? error was a sort of predefined prop

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -168,6 +163,16 @@ export function useGridEditRows(apiRef: GridApiRef) {
     [apiRef, forceUpdate, logger, setGridState],
   );
 
+  const handleCellEditPropsChange = React.useCallback(
+    (params: GridEditCellPropsParams, event?: React.SyntheticEvent) => {
+      if (event?.isPropagationStopped()) {
```
dtassone, line 168: Is that not handled in the publishEvent?

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -333,10 +355,11 @@ export function useGridEditRows(apiRef: GridApiRef) {
       getCellMode,
       isCellEditable,
       commitCellChange,
-      setEditCellProps,
+      setEditCellProps, // TODO don't expose, update the editRowsModel prop directly
```
dtassone, line 358: why this comment?

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: `commitCellChange` now checks if there's an error. If there's no error, it publishes `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED`. The handler of this event is who updates the row.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: No, `publishEvent` prevents from publishing an event if the given event object has the propagation stopped. Once the event is published, it will call all the listeners. If the user wants to stop the propagation of an event in `onEditCellChange`, we need to check by hand if the propagation was stopped.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: To reduce confusion between `setEditCellValue`. In edit components, the current value is only thing that needs to be updated in the majority of the cases. Outside an edit components, developers can control the state with the `editRowsModel` prop to set the error or other prop.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
dtassone: Yes so one way to do it, is to override the emit behavior of the event emitter so we avoid this check everywhere 🤔

dtassone (approved):

oliviertassinari:
> True, GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED shouldn't be fired.

@m4theushw To me it wasn't obvious. In the documentation we use the event to listen to the event that signals that the end-users are done (enter, click away, etc.), we even suggest that validation should be done after. From what I understand we have no documentation for using `error: true` in the model. Do we even need it? How about we remove this feature, waiting for a strong use case to come up?

m4theushw:
> In the documentation we use the event to listen to the action that signals that the end-user is done (enter, click away, etc.), we even suggest that validation should be done after. 

@oliviertassinari That's wrong. Although we say to do the validation after, the demo is doing the validation before, while the user is changing the value.

> From what I understand we have no documentation for using error: true in the editRowsModel. Do we even need it?

Now we have documentation about `error: true` in #2143. I updated the editing docs to better reflect what we do in the demos: https://deploy-preview-2143--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation

> How about we remove this feature, waiting for a strong use case to come up?

Based on the new docs what do you think we can remove? The `GRID_CELL_EDIT_COMMIT` (previously named `GRID_CELL_EDIT_PROPS_CHANGE_COMMITTED`) can be used if the developer wants to send to the server BEFORE updating the row, like to do additional validation in the server.

oliviertassinari:
@m4theushw Ok, I didn't saw #2143 when I commented. We are good

PR #2092, opened 2021-07-07
[core] Fix missing git source in packages

msftenhanceprovenance:
With the rise in supply chain attacks and OSS dependencies being used as a attack vector, Microsoft is working with our ecosystem partners, such as the Linux Foundation's OpenSSF, to enable OSS consumers to track packages back to their public sources.
We've identified that the following packages published to NPM do not report where sources can be found, typically accomplished by including a link to your GitHub repository in your `package.json` REPOSITORY field. This PR was created to add this value, ensuring future releases will include this provenance information.
Published NPM packages with repository information:
	*@material-ui/data-grid

### DIFF
--- a/packages/grid/data-grid/package.json
+++ b/packages/grid/data-grid/package.json
@@ -29,6 +29,11 @@
     "typescript": "tsc -p tsconfig.json",
     "build": "cd ../ && rollup --config rollup.data-grid.config.js"
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mui-org/material-ui-x.git",
+    "directory": "packages/grid/data-grid"
+  },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


oliviertassinari (commented):
Nice bot. I'm not sure what happened when we created the packages. We have already these in the core since the beginning, so it should have been in the x repo too, proof: https://github.com/mui-org/material-ui/blob/0230df94ca877fdd63a42c3e276f75c3a4316eea/packages/material-ui-lab/package.json#L15. We could also double-check for other differences. For instance we also forgot about the homepage.

To push an extra commit to fix all the other packages we have in this repository.

flaviendelangle (approved):

dtassone (approved):

PR #2098, opened 2021-07-07
[DataGrid] Row editing

### LINKED ISSUE #204: [DataGrid] Implement Row editing
### Benchmark

- https://ant.design/components/table/#components-table-demo-edit-row
- https://material-table.com/#/docs/features/editable
- https://demos.telerik.com/kendo-ui/grid/editing
- https://www.telerik.com/kendo-react-ui/components/grid/editing/
- https://ej2.syncfusion.com/react/demos/#/material/grid/normal-edit
- https://editor.datatables.net/

m4theushw:
Closes #204
Based on #2087
Previews: https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#row-editing

Based on the benchmarks, I think that the easiest path to implement this feature is to leverage the current cell editing, but with the difference that the value is committed once for the entire row, instead of on each cell change. There's already an `editMode` prop which is not being used, so I'm repurposing it to be used to choose between `cell`  and `row`.

TODO

- [x] Tests
- [x] Update docs
- [x] Add `onEditRowChangeCommit` prop
- [x] Update edit components to support the row editing (select shouldn't open straightaway)


oliviertassinari (commented):
It looks like an awesome starting point!

oliviertassinari:
1. We can remove "Row & Cell editing", it's here
https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/#upcoming-features
2. It's done? https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/getting-started/#feature-comparison

Capture d’écran 2021-08-10 à 01 40 13

3. The colors feel like a play garden 😁 cc @danilo-leal 

Capture d’écran 2021-08-10 à 01 51 33

https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#control-with-external-buttons

4. On this demo https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#conditional-validation, the state is lost, sometimes

https://user-images.githubusercontent.com/3165635/128788832-5b9f9604-8967-4d8b-b9a8-975f6cee72d3.mp4

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -206,10 +267,33 @@ export function useGridEditRows(
...
+        throw new Error(`You can't commit changes when the edit mode is 'cell'`);
+      }
+
+      const model = apiRef.current.getEditRowsModel();
+      const row = model[id];
+      if (!row) {
+        throw new Error(`Row at id: ${id} is not being editted`);
+      }
+
+      const hasFieldWithError = Object.keys(row).some((field) => !!row[field].error);
```
flaviendelangle, line 285: Could we use directly `Object.values` instead of extracting the keys and then getting the value of each key ?

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -104,6 +129,38 @@ export function useGridEditRows(
     [apiRef, forceUpdate, logger, setGridState],
   );
 
+  const setRowMode = React.useCallback(
```
flaviendelangle, line 132: To type the arguments and the return type, I propose to always do the following :
I avoids inconsistencies between the public API signatures and the method signature.
For instance in another hook the public API was setting an argument as mandatory and the method was setting it as optional. TypeScript was not complaining since the public API was stricter.

x other methods

```suggestion
  const setRowMode = React.useCallback<GridEditRowApi['setRowMode']>(
```

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -104,6 +129,38 @@ export function useGridEditRows(
...
+          delete newEditRowsState[id];
+        }
+        return { ...state, editRows: newEditRowsState };
+      });
+      forceUpdate();
+    },
+    [apiRef, columns, forceUpdate, setGridState],
+  );
+
+  const getRowMode = React.useCallback(
```
flaviendelangle, line 154: Same 
```suggestion
  const getRowMode = React.useCallback<GridEditRowApi['getRowMode']>(
```

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -291,37 +427,59 @@ export function useGridEditRows(
...
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_STOP, handleCellEditStop);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_COMMIT, handleCellEditCommit);
+
+  // Events exclusive to row editing
+  useGridApiEventHandler(apiRef, GRID_ROW_EDIT_START, handleRowEditStart);
+  useGridApiEventHandler(apiRef, GRID_ROW_EDIT_STOP, handleRowEditStop);
+  useGridApiEventHandler(apiRef, GRID_ROW_EDIT_COMMIT, handleRowEditCommit);
+
   useGridApiOptionHandler(apiRef, GRID_EDIT_CELL_PROPS_CHANGE, options.onEditCellPropsChange);
+  useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_COMMIT, options.onCellEditCommit);
```
flaviendelangle, line 466: Use the `props` instead of the `options`

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -125,6 +126,11 @@ export interface GridOptions {
    * @default false
    */
   disableSelectionOnClick?: boolean;
+  /**
+   * Controls whether to use the cell or row editing.
+   * @default "cell"
+   */
+  editMode?: 'row' | 'cell';
```
flaviendelangle, line 133: We could introduce a variable `GridEditMode` since it's used 5 times

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -605,10 +605,6 @@ describe('<XGrid /> - Rows', () => {
       };
     });
 
-    afterEach(() => {
-      clock.restore();
-    });
```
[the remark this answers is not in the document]
m4theushw: We actually don't fake the time in this group. Initially, I was faking the time in #2239, then I reverted but forgot to remove the `clock.restore();`.

packages/grid/_modules_/grid/models/gridOptions.tsx
m4theushw: I added a `GridEditMode` type alias.

m4theushw:
> On this demo https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#conditional-validation, the state is lost, sometimes

@oliviertassinari I couldn't reproduce it consistently, it happened almost randomly here. I suppose it loses the state because the time between clicks is too small. There's the first click, the internal state is updated with the new value, then it calls `onEditRowsModelChange` which also causes a rerender, but when it's called the second click already happened. I propose to debounce in a second PR the value update of each edit component:

https://github.com/mui-org/material-ui-x/blob/aa637c3eabcfad097b2c59af5f4efcf4041080d3/packages/grid/_modules_/grid/components/cell/GridEditBooleanCell.tsx#L38

If you open this other demo in local host https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation, the UX of erasing the content of the text input is not good. It's lagging because there's a rerender each time a character is pressed.

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -68,6 +74,8 @@ export function GridRow(props: GridRowProps) {
     isFunction(getRowClassName) && getRowClassName(apiRef!.current.getRowParams(id));
   const cssClasses = clsx(className, rowClassName, classes?.row, {
     'Mui-selected': selected,
+    [`${GRID_CSS_CLASS_PREFIX}-row--editing`]: apiRef.current.getRowMode(id) === 'edit',
+    [`${GRID_CSS_CLASS_PREFIX}-row--editable`]: editMode === 'row',
```
DanailH, line 78: Should we move those in the classes constant? Or better yet with GridClasses from here #2320

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -175,7 +180,7 @@ export const GridCell = React.memo(function GridCell(props: GridCellProps) {
       data-mode={cellMode}
       aria-colindex={colIndex + 1}
       style={style}
-      tabIndex={tabIndex}
+      tabIndex={cellMode === 'view' || !isEditable ? tabIndex : -1}
```
DanailH, line 183: Didn't we have an `enum` for the different cell modes? If not maybe we should have one.

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -109,6 +109,12 @@ export const GRID_CELL_FOCUS = 'cellFocus';
  */
 export const GRID_CELL_FOCUS_OUT = 'cellFocusOut';
 
+/**
+ * Fired when a cell gains focus. Called with a [[GridCellParams]] object.
+ * @event
+ */
+export const GRID_CELL_FOCUS_IN = 'cellFocusIn';
```
DanailH, line 116: I guess these would need rebasing and removing the `GRID` prefix.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -110,7 +140,39 @@ export function useGridEditRows(
...
+      });
+      forceUpdate();
+    },
+    [apiRef, columns, forceUpdate, setGridState],
+  );
+
+  const getRowMode = React.useCallback<GridEditRowApi['getRowMode']>(
+    (id) => {
+      if (props.editMode === 'cell') {
+        return 'view';
```
DanailH, line 168: I think we should move those into an `enum`. Typing it every time seems error-prone.

DanailH:
I tested it and it is an awesome feature 🚀.

The only thing I was able to find is when you add new rows and the rows exceed the overflow of the window you are not automatically scrolled to the newly created row. It makes sense for the user to get scrolled to that row since when you create it it is in `edit` mode. The same happens when you use pagination and you have a `pageSize=5` and you have 5 rows but then you add another one, which goes to the next page - it makes sense for the pagination to open the second page.

packages/grid/_modules_/grid/components/cell/GridCell.tsx
m4theushw: Yeah, I remember the discussion. The problem with adding an enum is that users won't be able to use plain strings. I tried to add an enum for the edit mode in https://github.com/mui-org/material-ui-x/pull/2098#discussion_r685855801 but we would force everyone to use the enum value, losing the simplicity.

m4theushw:
> The only thing I was able to find is when you add new rows and the rows exceed the overflow of the window you are not automatically scrolled to the newly created row. 

Now it's scrolling to the new row.

> The same happens when you use pagination and you have a pageSize=5 and you have 5 rows but then you add another one, which goes to the next page - it makes sense for the pagination to open the second page.

Interesting, we could investigate how to make `scrollToIndexes` support pagination.

oliviertassinari (commented):
I have personally seen enough. Happy to go to production with the current version. I have left more comments to polish. Great work, well done!

> On this demo https://deploy-preview-2098--material-ui-x.netlify.app/components/data-grid/editing/#conditional-validation, the state is lost, sometimes

I have found it using console.warn to get the stack track. It wasn't easy, took me half an hour! The reset happens when the double click triggers. I have pushed a commit 35df73f that replicates the same logic of `setCellMode` in `setRowMode`. We can skip the whole section if we are already in edit mode. Fixed

> It's lagging because there's a rerender each time a character is pressed.

I remember making the same point to Damien in a PR"s review. One possible solution could be to keep the state of the input local, and only sync with the parent data grid on commits.

> We could investigate how to make scrollToIndexes support pagination.

We could, maybe we should wait for a developer to complain?

packages/grid/_modules_/grid/components/cell/GridCell.tsx
DanailH: Hmm I'm not sure I understand. Having the `view` coming from a `enum` won't limit users to use simple strings to check: https://www.typescriptlang.org/play?#code/KYOwrgtgBAsg9gE2FA3gKClAagS2AdygF4oByANz31IBo0BfNNAYzhAGcAXKZ4AGz7wkxMpQKkA3E1Yc4fYADo+cAOYAKXgKHIiu2IkW4CASjRA

Or maybe I'm missing something?

m4theushw: If we use the enum to type the argument, then passing a string won't work anymore.

https://www.typescriptlang.org/play?#code/KYOwrgtgBAsg9gE2FA3gKClAagS2AdygF4oByANz31IBo0BfNNAMzBAGMAXHOEKT4AGdOACgiJgALlgSAlKgxR2vQXAA2wAHRq4AcxGlmcOACMAhgCdSsgNwMmA4QcoFrdoA

DanailH: Maybe we can use a combination of a `type` and an `enum` then. I've done it like this for the `density` feature https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/models/gridDensity.ts

m4theushw: I've created the enums `GridCellModes`, `GridRowModes` and `GridEditModes`. I'm not typing the argument with them, just using their value.

DanailH (approved):

oliviertassinari:
🎊

Issue #2044, opened 2021-07-05
[DataGrid] Implement control state/prop for all props using useGridControlState hook


dtassone:
link to #190 

## Summary 💡

Following core work of #1823, we now have to bind all props with their onChange handler.
Todo includes:
-  [x] RegisterControlState for selection #1823
-  [x] RegisterControlState for filter #1909 
-  [x] RegisterControlState for SortModel #2095 
-  [x] RegisterControlState for pagination substates so one for page, one for pagesize #2099 
-  [x] RegisterControlState for editRowsModel #2143
- [ ] Cleanup, Remove redundant publishevent code, cleanup hook control mode, Fix publish event params (inconsistency btw callback and hook body)

dtassone:
link to #190

dtassone:
FYI I'm picking up control sort model

flaviendelangle:
The `filterModel` is a single prop in `GridOptions`, which correspond to the `filter` key of the state.
But for the pagination, we don't have this equivalent.

The `pagination` state contains two different type of data.

### Descriptions of the rows 

- `paginationMode` which is coming from a prop (with a default fallback)
- `rowCount` which is coming from a prop if `paginationMode="server"` and from the `rows` (in which case it's a derived state)


### Descriptions of the current pagination

- `pageSize`
- `page`

For this part, we can have a controlled / uncontrolled behavior.
But unless the `filterModel`, these are two props, not a single `paginationModel` 

```ts
<Grid page={0} pageSize={20} />
```

Instead of 

```ts
<Grid paginationModel={{ page: 0, pageSize: 20 }} /> 
```

---

I think we want to split the state :

- the `paginationModel` part which can be controlled or uncontrolled
- the `rowCount` should not be derived stored in the state taken fril computed with the option or a selector

flaviendelangle:
The `filterModel` is a single prop in `GridOptions`, which correspond to the `filter` key of the state.
But for the pagination, we don't have this equivalent, we have several options that either can or can't be controlled.

### Descriptions of the rows 

- `paginationMode` which is coming from a prop (with a default fallback)
- `rowCount` which is coming from a prop if `paginationMode="server"` and from the `rows` (in which case it's a derived state)


### Descriptions of the current pagination

- `pageSize`
- `page`

For this part, we can have a controlled / uncontrolled behavior.
But unless the `filterModel`, these are two props, not a single `paginationModel` 

```ts
<Grid page={0} pageSize={20} />
```

[...]

PR #2095, opened 2021-07-07
[DataGrid] Improve controllable sorting

dtassone:
Part of #2044 (sorting)

```diff
// Props
- onSortModelChange?: (params: GridSortModelParams) => void;
+ onSortModelChange?: (model: GridSortModel) => void;

// Usage
- <DataGrid onSortModelChange = {(params: GridSortModelParams)=>  setSortModel(params.model)} />
+ <DataGrid onSortModelChange = {(model: GridSortModel)=> setSortModel(model) } />

```

PR #2099, opened 2021-07-08
[DataGrid] Improve controllable pagination

### LINKED ISSUE #1815: [DataGrid] Setting page with ApiRef in useEffect does not work anymore

flaviendelangle:
Part of #2044 
Fixes #1815 

- [x] Implement new controllable pagination with `updateControlState`
- [x] Stop using options for pagination
- [x] Split `useGridPagination` into `useGridPage` and `useGridPageSize` to have 1 hook = 1 sub state = 1 controlled item
- [x] Add new tests for pagination to cover the control use case and some edge cases discovered curing the rework 

```diff
// Props
- onPageChange?: (params: GridPageChangeParams) => void;
+ onPageChange?: (page: number) => void;

// Usage
- <DataGrid onPageChange = {(params: GridPageChangeParams)=>  setPage(params.page)} />
+ <DataGrid onPageChange = {(page: number)=> setPage(page) } />
```

```diff
// Props
- onPageSizeChange?: (params: GridPageChangeParams) => void;
+ onPageSizeChange?: (pageSize: number) => void;

// Usage
- <DataGrid onPageSizeChange = {(params: GridPageChangeParams)=>  setPageSize(params.pageSize)} />
+ <DataGrid onPageSizeChange = {(pageSize: number)=> setPageSize(pageSize) } />
```


packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
```diff
@@ -66,60 +56,89 @@ export const useGridPagination = (apiRef: GridApiRef): void => {
...
+    }
+
+    setGridState((oldState) => ({
+      ...oldState,
+      pagination: {
+        ...oldState.pagination,
+        page: options.page!,
+      },
+    }));
+  }, [setGridState, options.page]);
```
dtassone, line 105: you can use the props directly

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
```diff
@@ -66,60 +56,89 @@ export const useGridPagination = (apiRef: GridApiRef): void => {
...
     setGridState,
     forceUpdate,
-    options.paginationMode,
     visibleRowCount,
     options.rowCount,
     options.autoPageSize,
-    containerSizes?.viewportPageSize,
     options.pageSize,
     options.page,
+    containerSizes?.viewportPageSize,
```
dtassone, line 141: why do we need this one now?

flaviendelangle: Because if we updated the state [here](https://github.com/mui-org/material-ui-x/blob/0c733adb212748d20c2c661e99eff1b43fcc566a/packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts#L115) due to a viewport height change.

Then we must check if the pagination state is still valid (= rerun `applyContraints`)
But if we run it directly on the `setGridState` link above. Then we have cases were we update both `pageSize` and `page` in the same `setGridState`. And your hook does not like it.

So the workaround I found is to update `page` and `pageSize` separately.
And then in a third `setGridState` that is run whenever the rwo first are run, I update de constraints.

But it's quite ugly.
If you have a better approach I'm interested.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: That's a point I have some trouble understanding.
In the whole codebase, when should we use the props and when should we use the options ?

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: I could remove this specific variable from the 3rd useEffect by checking if we are in `autoPageSize` in the 1st useEffect and apply the constraint directly their if so.
It would be a bit more optimized, but even more hard to read.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
dtassone: we refactored hooks to take props as 2nd arguments so now you could use the props directly. Props are also now merged with the default props. So it's better to use props directly as it avoids a cycle to update options

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
```diff
@@ -66,60 +56,89 @@ export const useGridPagination = (apiRef: GridApiRef): void => {
...
+      (!options.autoPageSize || !containerSizes?.viewportPageSize)
+    ) {
+      return;
+    }
+
+    setGridState((oldState) => ({
+      ...oldState,
+      pagination: {
+        ...oldState.pagination,
+        pageSize: options.autoPageSize ? containerSizes?.viewportPageSize! : options.pageSize!,
```
dtassone, line 119: if you set autoPageSize and pageSize then you should be in some kind of invalid state. As if pageSize is controlled or set then, the state won't be updated to the viewportPageSize.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: OK, I'm changing it for all the `options` in `useGridPagination` then

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: I kept the old priority 

```tsx
          pageSize:
            (options.autoPageSize ? containerSizes?.viewportPageSize : options.pageSize) ||
            state.pagination.pageSize,
```

But you're right, the control state should have the priority over the `autoPageSize` to avoid inconsistency.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: https://app.circleci.com/pipelines/github/mui-org/material-ui-x/5958/workflows/d0ac1224-6900-4924-8cc6-be7dcaa96629/jobs/31174

This test fails if I use the props.
Before, the `useEfffect` in `useGridPagination` used `options.page` and `useGridSelector(apiRef, visibleGridRowCountSelector)` and both were always synchronized.
On the 1st run I have the default options page (0 page) and the default visible rows state (0 row).
O, the 2nd run I have the user options page (1 page) and the real visible rows state (1 row).

But now I'm using directly the prop.
So on the 1st run, I have the user options page (1 page) but I don't have the real visible rows state. I would have it if I ran the selector in the useEffect. But since it's in the render, it's still the value computed before `useGridFilter` updated the state.

```
1st render: Object{render: 0, effect: 3}
2nd render (caused by the initial `setGridState`: Object{render: 3, effect: 3}
3rd render (caused by the `setProps({ page: 2 })`) : Object{render: 3, effect: 3}
```

dtassone:
Could you add some tests for the control mode, similar to sorting #2095  or filter #1909

flaviendelangle:
@dtassone I added those 

https://github.com/mui-org/material-ui-x/pull/2099/files#diff-f6c6ed07bb3e3a333dedc80ef2cbb693915e79bf20fd4eb4f45dcdd9b50bdbefR193

https://github.com/mui-org/material-ui-x/pull/2099/files#diff-f6c6ed07bb3e3a333dedc80ef2cbb693915e79bf20fd4eb4f45dcdd9b50bdbefR112

https://github.com/mui-org/material-ui-x/pull/2099/files#diff-f6c6ed07bb3e3a333dedc80ef2cbb693915e79bf20fd4eb4f45dcdd9b50bdbefR84

For the `pageSize` I have some trouble clicking on the menus item but I should indeed add some tests.

flaviendelangle:
I did a very basic `rows` generator to avoid having to check all the time which brand is 1st, 2nd or 3rd in the list.
For features like pagination, I think boring data like `{ id: n, name: 'Element ${n}' }` is easier to read for the tests.
But maybe it does already exist in `useData`. I don't find a real documentation for this package, it could be interesting to have one.

```ts
const getData = (rowCount: number) =>
  Array.from({ length: rowCount }).map((_, i) => ({ id: i, name: `Element ${i}` }));
```

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -102,7 +102,7 @@ describe('<DataGrid /> - Pagination', () => {
       expect(onPageChange.callCount).to.equal(2);
```
dtassone, line 102: can you add the assertion on the call argument please

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -157,14 +157,62 @@ describe('<DataGrid /> - Pagination', () => {
...
+          page={0}
+          rowsPerPageOptions={[1, 2, 3]}
+        />,
       );
+
+      fireEvent.mouseDown(document.querySelector('.MuiSelect-selectMenu'));
+      expect(screen.queryAllByRole('option').length).to.equal(3);
+
+      fireEvent.click(screen.queryAllByRole('option')[1]);
+      expect(onPageSizeChange.callCount).to.equal(1);
```
dtassone, line 182: same here, please assert the call arguments

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -157,14 +157,62 @@ describe('<DataGrid /> - Pagination', () => {
...
+      fireEvent.mouseDown(document.querySelector('.MuiSelect-selectMenu'));
+      expect(screen.queryAllByRole('option').length).to.equal(3);
+
+      fireEvent.click(screen.queryAllByRole('option')[1]);
       expect(getColumnValues()).to.deep.equal(['Element 0']);
-      setProps({ page: 1, pageSize: 2 });
-      expect(getColumnValues()).to.deep.equal(['Element 2', 'Element 3']);
+    });
+
+    it('should control pageSize state when the prop and the onChange are set', () => {
```
dtassone, line 195: You are missing the tests when there is only the onchange event without model
```
<BaselineTestCase
  
            onPageSizeChange={(newPageSize) => setPageSize(newPageSize)}
```

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
flaviendelangle: :facepalm: I copied another test without updating the assertions

dtassone:
> I did a very basic `rows` generator to avoid having to check all the time which brand is 1st, 2nd or 3rd in the list.
> For features like pagination, I think boring data like `{ id: n, name: 'Element ${n}' }` is easier to read for the tests.
> But maybe it does already exist in `useData`. I don't find a real documentation for this package, it could be interesting to have one.
> 
> ```ts
> const getData = (rowCount: number) =>
>   Array.from({ length: rowCount }).map((_, i) => ({ id: i, name: `Element ${i}` }));
> ```

I wrote useData for that a while ago

flaviendelangle:
I think I mixed `useData` and `useDemoData`
`useData` is fine here, I'm making the switch

packages/grid/_modules_/grid/hooks/features/pagination/useGridPagination.ts
flaviendelangle: This is no longer an issue after the split of `usePagination` into two hooks

packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
```diff
@@ -14,46 +15,90 @@ import { useGridSelector } from '../features/core/useGridSelector';
...
+  gridPageSelector,
+} from '../features/pagination/gridPaginationSelector';
 import { optionsSelector } from '../utils/optionsSelector';
 import { useLogger } from '../utils/useLogger';
 import { useGridApiEventHandler } from './useGridApiEventHandler';
-
-export const useGridContainerProps = (apiRef: GridApiRef) => {
+import { GridComponentProps } from '../../GridComponentProps';
+
+function getBrowserScrollbarSize(doc: Document, element: HTMLElement): number {
```
DanailH, line 27: Should we move this to the `utils.ts` file?

flaviendelangle: I was in a dedicated hook before but I had to move it closer to where it's used.
But I agree, it would be better in `utils.ts`

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
```diff
@@ -18,8 +18,8 @@ function CustomPagination() {
     <Pagination
       className={classes.root}
       color="primary"
-      count={state.pagination.pageCount}
-      page={state.pagination.page + 1}
+      count={state.page.pageCount}
```
dtassone, line 21: why renaming pagination to page? 
I don't think page suits very well here

flaviendelangle: I split `state.pagination` (handled by `useGridPagination`) into `state.page` (handled by `useGridPage`) and `state.pageSize` (handled by `useGridPageSize`)

But I can rename the `useGridPage` into `useGridPagination` and `state.page` into `state.pagination`

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const getPageCount = (rowCount: number, pageSize: GridPageSizeState): number => {
+  if (pageSize > 0 && rowCount > 0) {
+    return Math.ceil(rowCount / pageSize);
+  }
+
+  return 0;
+};
+
+const clampPage = (pageState: GridPageState): GridPageState => {
```
dtassone, line 20: Can we rename it to something more meaningful and more common?

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+    [setGridState, forceUpdate, logger],
+  );
+
+  React.useEffect(() => {
+    apiRef.current.updateControlState({
+      stateId: 'pageSize',
+      propModel: props.pageSize,
+      propOnChange: props.onPageSizeChange,
+      stateSelector: (state) => state.pageSize,
```
dtassone, line 39: this can be in the pagination state as well.
We can have 2 hooks working on the same substate in the case of pagination I think

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
flaviendelangle: What do you have in mind ?
I took the name of the lodash function that does the same thing : https://lodash.com/docs/#clamp

packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
```diff
@@ -14,46 +15,77 @@ import { useGridSelector } from '../features/core/useGridSelector';
...
     options.pagination,
-    paginationState.page,
-    paginationState.pageSize,
+    options.pageSize,
+    pageState.currentPage,
+    pageSizeState,
     visibleRowsCount,
   ]);
 
+  const getScrollbarSize = React.useCallback(() => {
```
dtassone, line 60: we should already have one

flaviendelangle: Their was one in `useGridScrollbarSizeDetector.tsx` that was updating the `options.scrollbarWidth`
But now that I don't use the option, I need to store it elsewhere.

It is only used in `useGridContainerProps.ts` so call it directly here.
I would store it in a `state.scrollbarWidth` but I don't see a lot of benefits to do it.

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
dtassone: I would keep `useGridPage` but rename the state to pagination. These 2 hooks should be in the `pagination` directory
Then those 2 pagination hooks share the same state which is fine.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
flaviendelangle: I was trying to be as close as possible from 1 hook = 1 substate
I suppose we have 2 approaches :

1) We have a single sub state `state.pagination` used by both hooks
2) We have two sub state `state.page` and `state.pageSize`.

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2099#discussion_r673157407

Fine for me to use the same sub state called `pagination` :+1:

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
dtassone: I would have put something like `getValidPage` 🤔 
I was not familiar with `clamp`

flaviendelangle: It is maybe a wording only used by `lodash`
Anyway, if it not clear to someone in the team, we should probably rename it :+1:

packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
dtassone: well why don't you use options.scrollbarWidth, we don't need to stop using options completely, we can just use it when it is useful.

flaviendelangle: For me the option pattern makes the code quite hard to follow and to debug.

You have optional props
Then a hook `useOptionsProp` that put those props in the state and sometime applies a default value, sometime merge the value with a default value (for `classes` and `localeText`) and sometime does not apply any default value.
Then a hook `useGridScrollbarSizeDetector` that updates the `scrollbarSize` option but only if the prop is null.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
dtassone: Reason for 1. would be, as these 2 hooks serve the same feature family then they can share the same substate I think. 

It could also be flat  as 2 . but then other prop would also need to be in the state such as pageCount.  Therefore, if you keep it flat you might pollute the first level of the state by adding your props directly, there which would impact the state  shape visibility... Therefore I lean towards 1.

packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
dtassone: I understand your confusion and options should be refactored a bit now. However I think in the case of scrollbarSize it is still useful, you could refactor useGridScrollbarSizeDetector to its own state but I think it's a bit outside this PR scope.

flaviendelangle: I can go back using the option.
When an effect uses both options and props, they may be inconsistencies when the user updates 2 props at once (their is a render where the prop used directly is updated but the prop used through the option process is not).

But I suppose for the scrollbarWidth, it should not occurs.

flaviendelangle: For me, all the "options" that can be initialized synchronously before the Grid DOM is mounted should be preprocessed in some `useMemo`s

Something like 

```ts
const props = useGridPreprocessedProps(rawProps)
```

And most of them are. Their is only the `scrollbarSize` which is a bit different since you pass the grid root (I suppose they may be situations where the scrollwidth is not the same in various places of the DOM).
So if in a different PR I put the `scrollbarbarWidth` out of the options (in a dedicated substate or in `state.scrollbar`), then I think we can have every use case currently covered by the options without the prop to state pattern.

dtassone: Props are already merged with their default value. 
https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/GridComponent.tsx#L38

In the case of the scrollbar, this can't work as the default value is dynamic hence why we need to use options at this stage.

flaviendelangle: Yes, my proposal allow to stop using the option (or any state-based solution) for `classes` and `localeText`.
But the the scrollbarWidth we need the grid to be mounted to determine the default value.

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -25,145 +26,371 @@ describe('<DataGrid /> - Pagination', () => {
...
+
+      rows = document.querySelectorAll('.MuiDataGrid-viewport [role="row"]');
+      expect(rows.length).to.equal(expectedViewportRowsLengthAfter);
+
+      expect(onPageSizeChange.lastCall.args[0]).to.equal(expectedViewportRowsLengthAfter);
+    });
+  });
+
+  describe('other props', () => {
+    it('should react to an update of rowCount', () => {
```
m4theushw, line 380: We don't need to group these tests. They can be one level up.

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -25,145 +26,371 @@ describe('<DataGrid /> - Pagination', () => {
...
-            <DataGrid {...baselineProps} rowCount={5} {...props} />
-          </div>
-        );
-      };
+    it('should not call onPageChange on initialisation or on page prop change', () => {
+      const onPageChange = spy();
 
-      const { setProps } = render(<Test />);
-      expect(document.querySelector('.MuiTablePagination-root')).to.have.text(
-        'Rows per page:1001-5 of 5',
```
m4theushw, line 67: Can we make this test to pass without removing this assertion?

m4theushw (commented):
Everything seems to be working well. 👌 My only concern are the tests. It's a bit hard to review all of them. Can we not change the tests we had and add another group only with test cases for the controllable pagination? It will be easier to review and to spot possible regressions. In the filtering, there's a group dedicated to them: https://github.com/mui-org/material-ui-x/blob/a1d05f2a2c041df6234f56b6107b1df65458e158/packages/grid/x-grid/src/tests/filtering.XGrid.test.tsx#L426

There's a pattern I noticed we're doing:

1. Assert that the state is updated when neither the prop nor the onChange is present
2. Assert that the state is not updated when the prop is present
3. Assert that the state is updated when the onChange is present but not the prop
4. Assert that the state is totally controlled when the prop and the onChange is present

flaviendelangle:
@m4theushw I should have waited for another PR before reorganizing the tests.
The ones impacted by the control mode are

For `page`

1. Assert that the state is updated when neither the prop nor the onChange is present
`should apply new page when clicking on next / previous button and onPageChange is not defined and page is not controlled`

2. Assert that the state is not updated when the prop is present
`should not change the page state when clicking on next button and a page prop is provided`

3. Assert that the state is updated when the onChange is present but not the prop
`should call onPageChange and apply new page when clicking on next / previous button and page is not controlled`

4. Assert that the state is totally controlled when the prop and the onChange is present
`should call onPageChange with the correct page when clicking on next / previous button when page is controlled

For `pageSize`

1. Assert that the state is updated when neither the prop nor the onChange is present

2. Assert that the state is not updated when the prop is present
`should not change the pageSize state when clicking on a page size option when pageSize prop is provided`

3. Assert that the state is updated when the onChange is present but not the prop
`should call onPageChange and apply the new pageSize when clicking on a page size option and pageSize is not controlled`

4. Assert that the state is totally controlled when the prop and the onChange is present
`should call onPageSizeChange with the correct page when clicking on a page size option when pageSize is controlled`

flaviendelangle:
https://github.com/mui-org/material-ui-x/pull/2099#discussion_r675248613

@m4theushw sorry I could not answer on your thread 
It's to give priority to the `pageSize` when both `pageSize` and `autoPageSize` are defined.

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
```diff
@@ -19,7 +19,7 @@ function CustomPagination() {
       className={classes.root}
       color="primary"
       count={state.pagination.pageCount}
-      page={state.pagination.page + 1}
+      page={state.pagination.currentPage + 1}
```
dtassone, line 22: why not just page? We know already it's the current one 😄

docs/src/pages/components/data-grid/events/events.json
```diff
@@ -125,14 +125,8 @@
...
   },
-  {
-    "name": "pageChange",
-    "description": "Fired when the current page change. Called with a GridPageChangeParams object."
-  },
-  {
-    "name": "pageSizeChange",
-    "description": "Fired when the page size change. Called with a GridPageChangeParams object."
-  },
+  { "name": "pageChange", "description": "Fired when the current page changes." },
```
dtassone, line 128:
```diff
-current
```

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
flaviendelangle: It was to avoid `state.page.page`
But now that we have a single state for both hooks I can rename it :+1:

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -0,0 +1,114 @@
...
+    [setGridState, forceUpdate, logger],
+  );
+
+  React.useEffect(() => {
+    apiRef.current.updateControlState({
+      stateId: 'page',
+      propModel: props.page,
+      propOnChange: props.onPageChange,
+      stateSelector: (state) => state.pagination.currentPage,
+      onChangeCallback: (model) => {
```
dtassone, line 61: we could rename 'model' to page here

packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts
```diff
@@ -32,16 +32,17 @@ export const useGridContainerProps = (apiRef: GridApiRef) => {
 
   const getVirtualRowCount = React.useCallback(() => {
     logger.debug('Calculating virtual row count.');
-    if (options.pagination && !options.autoPageSize) {
-      const rowsLeft = visibleRowsCount - paginationState.page * paginationState.pageSize;
+    if (options.pagination && (!options.autoPageSize || options.pageSize)) {
```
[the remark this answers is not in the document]
flaviendelangle: @m4theushw sorry I could not answer on your thread
It's to give priority to the pageSize when both pageSize and autoPageSize are defined.

See Pagination prop: autoPageSize should give priority to the controlled pageSize

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -44,7 +44,7 @@ export const useGridPage = (
         ...oldState,
         pagination: applyValidPage({
           ...oldState.pagination,
-          currentPage: page,
+          page: page,
```
dtassone, line 47: lint, short syntax here

dtassone (approved):

dtassone:
Nice one 👍

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
m4theushw: Would ne nice to keep `page`. The diff will be smaller.

packages/grid/x-grid/src/tests/pagination.XGrid.test.tsx
```diff
@@ -7,51 +7,92 @@ import { getColumnValues } from 'test/utils/helperFn';
...
+            <XGrid {...basicData} apiRef={apiRef} pagination pageSize={1} />
+          </div>
+        );
+      };
+
+      render(<GridTest />);
+
+      expect(getColumnValues()).to.deep.equal(['0']);
+      act(() => {
+        apiRef.current.setPage(1);
```
m4theushw, line 36: Interesting, tests with `setPage` don't need `act` but `setPageSize` won't pass without it.

m4theushw (approved):
Perfect.

oliviertassinari:
I have updated the PR's description so I can copy & paste its BC content during the next release.

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -501,7 +500,6 @@ export const DEFAULT_GRID_OPTIONS: GridOptions = {
   filterMode: GridFeatureModeConstant.client,
   headerHeight: 56,
   localeText: GRID_DEFAULT_LOCALE_TEXT,
-  pageSize: 100,
```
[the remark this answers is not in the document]
oliviertassinari: Ah, it's a controlled prop! So there is more to it than the breaking changes we document.

HEAD: you can change the page size, https://codesandbox.io/s/material-demo-forked-vofk8
PR: you can't https://codesandbox.io/s/material-demo-forked-l54jb

I'm updating the description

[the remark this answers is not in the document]
flaviendelangle: Indeed, only provided the `pageSize` and keeping the internal state update is not possible anymore.
If it's something we want, we may provide a `defaultPageSize` property but I'm not sure it's worth it.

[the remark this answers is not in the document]
oliviertassinari: We could introduce default props, we started without to see if it flies without 🤷‍♂️

PR #2101, opened 2021-07-08
[DataGrid] Add `minWidth` to `GridColDef`

### LINKED ISSUE #1850: [DataGrid] Introduces `minWidth` in `ColDef` for columns with fluid width
Based on these discussions:
https://github.com/mui-org/material-ui-x/issues/1561#issuecomment-832753596
https://github.com/mui-org/material-ui-x/pull/1816#issuecomment-855771150

## Summary 💡

Have the ability to set `minWidth` for columns.

## Examples 🌈

A possible solution can be the following (adding optional `minWidth` for fluid columns):

```tsx
<DataGrid
  columns={[
    {
      field: 'id',
      flex: 1,
      minWidth: 50,
    },
    {
      field: 'username',
      width: 200,
    },
    {
      field: 'age',
      flex: 0.3,
      minWidth: 80,
    },
  ]}
  rows={{
    id: 1,
    username: '@MaterialUI',
    age: 20,
  }}
 />
```

## Motivation 🔦

In order for the fluid column width feature to be consistent, we need to introduce a way to set the `minWidth` of a fluid column.

## Benchmark

- https://ag-grid.com/javascript-grid/column-sizing/#column-flex

DanailH:
Fixes #1850 

A follow up on this PR -> https://github.com/mui-org/material-ui-x/pull/1954

The `minWidth` also works in combination with the column resize feature as in if `minWidth` is provided the resizing will be restricted and the width of the column cannot be smaller than the set `minWidth`.

I used the same example, with just a small tweak.

Preview: https://deploy-preview-2101--material-ui-x.netlify.app/components/data-grid/columns/#minimum-width


docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -60,6 +60,8 @@ It works by dividing the remaining space in the grid among all flex columns in p
 For example, consider a grid with a total width of 500px that has three columns: the first with `width: 200`; the second with `flex: 1`; and third with `flex: 0.5`.
 The first column will be 200px wide, leaving 300px remaining. The column with `flex: 1` is twice the size of `flex: 0.5`, which means that final sizes will be: 200px, 200px, 100px.
 
+To set a minimum width for a `flex` column set the `minWidth` property in `GridColDef`.
```
dtassone, line 63: is it only for flex?

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -82,6 +84,8 @@ By default, `XGrid` allows all columns to be resized by dragging the right porti
 To prevent the resizing of a column, set `resizable: false` in the `GridColDef`.
 Alternatively, to disable all columns resize, set the prop `disableColumnResize={true}`.
 
+To restrict resizing a column under a certain width set the `minWidth` property in `GridColDef`.
```
dtassone, line 87: Shouldn't we do a maxWidth too?

DanailH: Hmm, that is interesting. It makes sense to add `maxWidth` as well to have all the relevant props supported, but maybe in a separate PR.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
```diff
@@ -96,7 +96,7 @@ export const GridRowCells = React.memo(function GridRowCells(props: RowCellsProp
     const cellProps: GridCellProps = {
       value: cellParams.value,
       field: column.field,
-      width: column.width!,
+      width: column.minWidth ? Math.max(column.minWidth, column.width!) : column.width!,
```
flaviendelangle, line 99: When is this line usefull (same thing for its equivalent in `GridColumnHeaderItem`)
Shouldn't the logic in `updateColumnsWidth` be enough ?
When I remove it, the `Resizing` doc still works.

flaviendelangle:
We should probably update the `Resizing` documentation with a line describing the `minWidth` (and in the future the `maxWidth`) impact on the resize.

And add a `minWidth` to one of the two resizable columns of the doc example.

oliviertassinari (dismissed):
A test case where the new logic doesn't work https://codesandbox.io/s/material-demo-forked-hkgdo?file=/demo.js. There is enough space not to have a scrollbar, and yet we display one.

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -42,6 +42,10 @@ export interface GridColDef {
    * If set, it indicates that a column has fluid width. Range [0, ∞).
    */
   flex?: number;
+  /**
+   * If set, it sets the minimum width a column can have. Commonly used in fluid columns.
+   */
```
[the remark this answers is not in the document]
DanailH: The default is only for resizing, its a local logic.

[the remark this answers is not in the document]
oliviertassinari: Oh ok. Should we consider having a default too for the minWidth? It doesn't have to be the same as for resizing. I wonder.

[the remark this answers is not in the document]
DanailH: I was thinking about it and if we do and a developer sets a lower `width` in the `GridColDef` then we will force them to also provide `minWidth`. I don't think it will bring much value to set a default `minWidth`.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
DanailH: Actually, this is needed for the initial load - the reason is that `updateColumnsWidth` is only called when there is a change. If this is removed then setting `minWidth` won't be respected on the first load.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -147,7 +147,7 @@ export function GridColumnHeaderItem(props: GridColumnHeaderItemProps) {
     ...classNames,
   );
 
-  const width = column.width!;
+  const width = column.minWidth ? Math.max(column.minWidth, column.width!) : column.width!;
```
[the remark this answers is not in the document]
DanailH: https://github.com/mui-org/material-ui-x/pull/2101#discussion_r666900814

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
flaviendelangle: When do we set the flex column width for the 1st load then ?

DanailH: Actually my bad, it is called it's just only respecting `fluid` columns

flaviendelangle: Oh indeed
So I suppose we want to improve the `updateColumnsWidth` (and maybe rename it `preprocessColumnsWidth` or something like that).

oliviertassinari: > Example: minWidth > width, you update the width accordingly in updateColumnsWidth, so width = minWidth and then the original value of the width is lost.

Why is it an issue to lose the original value?

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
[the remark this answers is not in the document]
oliviertassinari: >  I don't think it will bring much value to set a default minWidth.

To me, the use case is identical to having a min-width on the column resizing logic. I believe the min width we have on resizing is meant to not have any columns that completely shrink up to a point where the end-users can't interact with it.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
flaviendelangle: I can agree that modifying the column like we do feels wrong.
For instance, we have to put `column.width!` everywhere because we know that the typing is wrong and that after this point the width is always defined. 
I would prefer having an explicit type change.

For instance
```
interface PreProcessedGridColDef extends GridColDef {
 processedValues: {
   width: number
 }
}

const preprocessColumns: (rawColumns: GridColDef[]): PreProcessedGridColDef[]
```

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -58,10 +58,24 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number) {
...
+
+      if (viewportWidth > 0) {
+        if (column.minWidth && column.minWidth > flexColumnWidth) {
+          newWidth = column.minWidth;
+        } else {
+          newWidth = flexColumnWidth;
+        }
+      } else {
+        newWidth = minColumnWidth;
+      }
```
m4theushw, line 74: A small simplification:

```suggestion
      let newWidth = column.minWidth! ?? GRID_STRING_COL_DEF.width;
      if (viewportWidth > 0) {
        const flexColumnWidth = Math.floor(flexMultiplier * column.flex!);
        newWidth =
          column.minWidth && column.minWidth > flexColumnWidth ? column.minWidth : flexColumnWidth;
      }
```

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
m4theushw: The `useGridColumns` already updates the widths in the initial load, but it doesn't anything when there's no fluid column. If we add an else case to the conditional below, we'll be able to compute the widths for non-fluid columns and remove this leaking logic. 

https://github.com/mui-org/material-ui-x/blob/3c0082e6ae06790fde25e4f2772b892162ffaa89/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L55

m4theushw (commented):
Open https://codesandbox.io/s/material-demo-forked-8soge?file=/demo.js and note that the age column is hidden. I think we forgot to consider `minWidth` when computing the available viewport width:

https://github.com/mui-org/material-ui-x/blob/3c0082e6ae06790fde25e4f2772b892162ffaa89/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L46

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
DanailH: Updated and removed

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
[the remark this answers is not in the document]
DanailH: Updated

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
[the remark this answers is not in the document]
DanailH: Updated - added a default `minWidth` of 50 and removed the constant from the `useColumnResize`. Added also a new `Minimum width` section to the docs.

packages/grid/_modules_/grid/models/colDef/gridStringColDef.ts
```diff
@@ -5,6 +5,7 @@ import { getGridStringOperators } from './gridStringOperators';
 
 export const GRID_STRING_COL_DEF: GridColTypeDef = {
   width: 100,
+  minWidth: 10,
```
m4theushw, line 8: Is not 10px too little? Now if I resize a column, the separator invades the 3 dot icon.

image

```suggestion
  minWidth: 48,
```

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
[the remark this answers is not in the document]
m4theushw: Don't forget to update the default value in the JSDoc comment.

DanailH:
Ok, I approved argos - it was complaining because of the `50px` `minWidth`. It would be great if I can merge it 😅

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -50,6 +50,14 @@ To change the width of a column, use the `width` property available in `GridColD
 
 {{"demo": "pages/components/data-grid/columns/ColumnWidthGrid.js", "bg": "inline"}}
 
+### Minimum width
+
+By default, the columns have a minimum width of 50 pixels.
```
oliviertassinari, line 55: We have moved away from using "pixels" in the core because it could be confused with the screen pixels, here we cover CSS pixels, so px seems more appropriate. Would it make more sense? It's almost the same so not something major.

```suggestion
By default, the columns have a minimum width of 50px.
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -50,6 +50,14 @@ To change the width of a column, use the `width` property available in `GridColD
...
+This is an arbitrary, easy to remember value.
```
oliviertassinari, line 56: https://english.stackexchange.com/questions/293323/easy-to-use-versus-easy-to-use/293336

```suggestion
This is an arbitrary, easy-to-remember value.
```

oliviertassinari (commented):
Great to not have the logic leak outside of the two relevant hooks 👍

docs/pages/api-docs/data-grid/grid-col-def.md
```diff
@@ -29,6 +29,7 @@ import { GridColDef } from '@material-ui/data-grid';
...
 | <span class="prop-name optional">hide<sup><abbr title="optional">?</abbr></sup></span>              | <span class="prop-type">boolean</span>                                                             | <span class="prop-default">false<br /></span>    | If `true`, hide the column.                                                                                     |
 | <span class="prop-name optional">hideSortIcons<sup><abbr title="optional">?</abbr></sup></span>     | <span class="prop-type">boolean</span>                                                             | <span class="prop-default">false<br /></span>    | Toggle the visibility of the sort icons.                                                                        |
+| <span class="prop-name optional">minWidth<sup><abbr title="optional">?</abbr></sup></span>          | <span class="prop-type">number</span>                                                              | <span class="prop-default">50<br /></span>       | Sets the minimum width a column can have. It's commonly used in fluid columns.                                  |
```
dtassone, line 32:
```
- Sets the minimum width a column can have. It's commonly used in fluid columns.
+Sets the minimum width of a column.
```

dtassone (approved):

docs/src/pages/components/data-grid/columns/columns.md
DanailH: Created a follow up ticket to not forget about this discussion -> https://github.com/mui-org/material-ui-x/issues/2174

DanailH:
Created a follow up ticket for `maxWidth` -> https://github.com/mui-org/material-ui-x/issues/2174

Issue #1826, opened 2021-06-03
[data grid][docs] Follow-up on docs generation


oliviertassinari:
#1529 is a major step forward toward exhaustive documentation of the API. These are opportunities I have seen that we could tackle next:

- [ ] 1. The new [Events](https://master--material-ui-x.netlify.app/components/data-grid/events/) documentation page is limited. Frustration: https://github.com/mui-org/material-ui-x/issues/2001#issuecomment-872292432
  - Events are not specific to the Pro plan. Developers can listen to them with regular props: https://codesandbox.io/s/material-demo-forked-2lp71?file=/demo.js. We can add a new second about it.
  - Events can be imported as constants, there is no mention of it
  - "Fired when the grid is unmounted." Is this event really relevant?
- [ ] 2. The details of `GridColDef` seems hard to find. In the h2 [Column definitions](https://master--material-ui-x.netlify.app/components/data-grid/columns/#column-definitions), we could have a h3 about the [GridColDef API](https://master--material-ui-x.netlify.app/api/data-grid/grid-col-def/).
- [x] 3. The apiRef blocks miss the href in https://github.com/mui-org/material-ui-x/blob/0fd56fd31047707700317f14bebec7fb2bf4c37a/docs/src/pages/components/data-grid/filtering/filtering.md#L161
- [ ] 4. The apiRef blocks can be improved on how the new component pattern displays the information. It's a follow-up on https://github.com/mui-org/material-ui-x/pull/1529#discussion_r639248789. It seems that we have the following opportunities. We can improve the density of information, there is a lot of wasted space. We can reduce the number of clicks to access the information.
- [ ] 5. The apiRef blocks break the copy & paste the method name in https://material-ui.com/components/data-grid/filtering/#apiref

https://user-images.githubusercontent.com/3165635/126902292-76eef161-cfac-454b-933b-4480dbaec30f.mp4

- [ ] 6. The apiRef blocks pattern can be used in many other parts: density, edit row, csv export, and in the core too.
- [x] 7. We have a JSON docs generator in the core repository for React components (not interfaces). We should be able to use it for `DataGrid`, `XGrid`, `GridToolbarContainer`, `GridToolbarExport`, etc.
- [ ] 8. Once we have this JSON generated. We can move to the next step, read it to render the API page. We have a page component to read it in the core repository, we should be able to use it too.
[...]

PR #1529, opened 2021-04-30
[docs] Automatically generate API docs

m4theushw:
This PR covers the automatic generation of the API pages using `typedoc`. 

Preview:
https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/

Closes #446
To be merged after #1572

### Summary of changes

1. All API pages of the DataGrid are not in a new submenu under Components API.

image

2. Added a [page](https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/) to group the exports into categories. This is what users looking for the DataGrid component will see now.

3. Added a [page](https://deploy-preview-1529--material-ui-x.netlify.app/components/data-grid/events/) describing how to subscribe to events. This page also contains a table automatically populated with all the events. I had to do some workarounds to get the same styling from the component API page here. The worst part I think is using the `MarkdownElement` (a component from the docs) inside a demo, but users can't fork the demo so maybe we can live with that. We could separate the events into groups if necessary. 

4. Generated API pages for the [GridApi](https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/grid-api/), [GridColDef](https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/grid-col-def/), [GridCellParams](https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/grid-cell-params/) and [GridRowParams](https://deploy-preview-1529--material-ui-x.netlify.app/api/data-grid/grid-row-params/). Which other APIs should we have documentation?

---

TODO:

- [x] Docs for events (https://ag-grid.com/javascript-grid/grid-events/)
- [x] Index page (https://www.telerik.com/kendo-react-ui/components/grid/api/)
- [x] Add missing renders in `generateType` (https://github.com/TypeStrong/typedoc-default-themes/blob/master/src/default/partials/type.hbs)
- [x] Add links to other API pages: "Allows to update a column [[GridColDef]] model."
- [x] Filter out properties with "@ignore - do not document"
- [x] Prettify generated pages
- [x] Assert in CircleCI that json files are up to date

PR #2102, opened 2021-07-08
[docs] Generate api doc for the GridExportCSVOptions interface

### LINKED ISSUE #1924: [DataGrid] Toolbar Export delimiter is not documented
Hi! I 'am using Material UI DataGrid Grid Toolbar Export to export data to csv. Can i change delimiter and encoding?

Thank you!

flaviendelangle:
Closes #1924 

@m4theushw I needed to edit the `pages.js` on the core (directly in my local `node_modules` for now) for the doc to build.
Could we do like for the `components` part of the doc and add a fallback in the `pages.js` file of the X repo ?

Preview: https://deploy-preview-2102--material-ui-x.netlify.app/api/data-grid/grid-api/


flaviendelangle:
I think it would be better to re-use at least some parts of the logic of the `buildApi.ts` of the core.

Because for the components, we don't want to re-implement everything
For instance the inheritance management : https://material-ui.com/api/button/#inheritance

oliviertassinari:
@flaviendelangle You could use the existing tooling of the main mono-repository to generate the .json files per component. Even if we can't use them in the docs, it would get us one step closer to when we swap the docs to use v5 over v4.

flaviendelangle:
@oliviertassinari would this JSON be used in a documentation page or just generated to prepare for further works ?

oliviertassinari:
> Do you think it's still a good idea to show the interfaces I added to the autogenerated doc?

@flaviendelangle I assume you are talking about `GridCsvExportOption`.

- It makes me think about https://github.com/mui-org/material-ui-x/pull/1529#discussion_r639248789. 
- If we add a new page, is there a way we could link it from the apiRef page
- AG Grid embed the API page inside the export page https://ag-grid.com/react-grid/csv-export/#api. If we don't embed, I think that we should link the page.

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -19,13 +19,13 @@ To enable the CSV export you need to compose a toolbar containing the `GridToolb
...
 
-1. Define the exact columns to be exported with the `fields` attribute in the `csvOptions` prop of [`GridToolbarExport`](/components/data-grid/components/#toolbar).
+1. Define the exact columns to be exported with the `fields` attribute in the [`csvOptions`](/api/data-grid/grid-export-csv-options) prop of [`GridToolbarExport`](/components/data-grid/components/#toolbar).
 
 ```jsx
 <GridToolbarExport csvOptions={{ fields: ['id', 'name'] }} />
 ```
 
-Set `allColumns` in `csvOptions` to true to include hidden columns, instead of only the visible ones.
+Set `allColumns` in [`csvOptions`](/api/data-grid/grid-export-csv-options) to true to include hidden columns, instead of only the visible ones.
```
oliviertassinari, line 28: The URL is a 301
```suggestion
Set `allColumns` in [`csvOptions`](/api/data-grid/grid-export-csv-options/) to true to include hidden columns, instead of only the visible ones.
```

docs/src/pages.js
```diff
@@ -68,4 +73,27 @@ datagrid.children =
         { pathname: '/components/data-grid/group-pivot', title: '🚧 Group & Pivot' },
       ];
 
+dataGridComponentAPI.children =
+  process.env.PULL_REQUEST === 'false'
```
oliviertassinari, line 77: I doubt that we should approach unstable docs page like this. We duplicate the content. Instead, we could make it additive. For now, maybe remove this branch for now?

docs/pages/api-docs/data-grid/test-doc-gen.md
```diff
@@ -0,0 +1,21 @@
+# GridRowParams Interface
+
+<p class="description">Object passed as parameter in the column <a href="/api/data-grid/grid-col-def/">GridColDef</a> cell renderer.</p>
+
+## Import
```
oliviertassinari, line 5: To remove?

docs/src/pages.js
flaviendelangle: I did what was done for the `components` part (above in the same file)
But I do not understand the goal of this ternary to be honest.

oliviertassinari: The goal was to be able to work on new pages in dev mode without releasing them in production.

flaviendelangle: Is the process of releasing a new doc page documented ?
I suppose that at some point we need to add it to the core repo for it to be released.

oliviertassinari: > Is the process of releasing a new doc page documented ?

It's not documented when it comes to X. For the core, it's inside the CONTIRBUTION.md guide. Understanding how it works requires understanding how the production is structured. To make it short, we have two different websites (core, x). In production, we use a CDN, we proxies this two webistes to share the same URLs. 

So the sync is about:

1. making sure we share the same code infra for consistency (e.g. same CSS). This is done with x importing the core repository (monorepo dependency pointing to a git commit) and using the code of the docs of the core.
2. make sure the links in X are up to date. this is currently done automatically by importing the pages from the core repo. In the future, we will solve that by not showing the links in the first place. At least, it's what we have discussed doing with @danilo-leal.
3. make sure the links in core are up to date, pointing to X correctly. This is done manually, by the X team making PRs on the core. In the future, we will solve that by not showing the links.

flaviendelangle: OK it's a lot clearer now.
Indeed the 3. is a limitation that would be nice to remove but I suppose it's far from being simple.

m4theushw:
You need to open another PR in the core repository to add this new page to its pages file: https://github.com/mui-org/material-ui/blob/4fe0df5b4dfbe161cf1ad8bb5183c61506b556b6/docs/src/pages.js#L174

Without it, this page will not appear in the sidebar nav when the user is not navigating in a non-X page (e.g. https://material-ui.com/api/chip/). It will only appear while inside a X page.

> Could we do like for the components part of the doc and add a fallback in the pages.js file of the X repo ?

I'm OK with it. When I did #1529 I forked the core repo, added the new page to it, then updated the #1529's package.json to use my fork as the monorepo version. Before merging, I reverted the package.json.

> How about we use the existing tooling of the main mono-repository to generate one .json file per component we have here? Even if we can't use them in the docs, it would get us one step closer to when we swap the docs to use v5 over v4.

The tool here is already capable of generating a JSON for interfaces that are part of the `GridApi` or events. We use this JSON file to embed API docs into the markdown, see [1](https://material-ui.com/components/data-grid/filtering/#apiref) and [2](https://material-ui.com/components/data-grid/events/#catalog-of-events). If you like, you can experiment with embedding non-GridApi interfaces. We have a bunch of small interfaces missing docs. However, the process of creating a PR here, another one in the core repo and then sync the merge of both doesn't scale, so embedding might be better.

image

m4theushw (approved):
Looks good. We can explore the embedding-approach in another PR, if you like.

docs/pages/api-docs/data-grid/index.md
```diff
@@ -13,6 +13,7 @@
 - [GridCellParams](/api/data-grid/grid-cell-params/)
 - [GridColDef](/api/data-grid/grid-col-def/)
 - [GridRowParams](/api/data-grid/grid-row-params/)
+- [GridToolbarExportProps](/api/data-grid/grid-toolbar-export-props/)
```
m4theushw, line 16: Broken link.

```suggestion
- [GridExportCsvOptions](/api/data-grid/grid-export-csv-options/)
```

m4theushw (changes requested):
I approved too early. 🤦‍♂️

m4theushw (approved):
Now it's really good.

oliviertassinari:
> The tool here is already capable of generating a JSON for interfaces that are part of the GridApi or events. We use this JSON file to embed API docs into the markdown, see 1 and 2. If you like, you can experiment with embedding non-GridApi interfaces. We have a bunch of small interfaces missing docs. However, the process of creating a PR here, another one in the core repo, and then sync the merge of both doesn't scale, so embedding might be better.

@m4theushw Ok, I should clarify my previous point because I wasn't clear. There are a couple of different topics as to how we can improve the current situation. I have updated #1826 with every idea I could get. The proposals I have there are unrelated to the pain of having a **double** sync loop (it's not only x that needs to be updated when core changes, but core when x changes). This double sync pain, I believe the [docs redesign](https://www.notion.so/material-ui/Rearranging-the-Documentation-ed351b1778294d9b9eca90d6df08287d) will take are of it. But again I believe that it's a different problem.

+1 for investing more on the embedded API demos, not only for apiRef but for other interfaces, e.g. hooks, flat simple objects.

cc @flaviendelangle

Issue #2103, opened 2021-07-08
[DataGrid] Column header is moving when keyboard arrow keys are pressed


Deligar:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When all columns have property flex the header is moving when keyboard arrow keys are pressed (header doesn't match columns after moving)

Before:
![image](https://user-images.githubusercontent.com/65401960/124922789-86241480-e002-11eb-8f6c-914c4071113d.png)

After:
![image](https://user-images.githubusercontent.com/65401960/124922936-a653d380-e002-11eb-8422-cddc395ba028.png)

## Expected Behavior 🤔

Header should remain still

## Steps to Reproduce 🕹

Steps:

1. Open https://material-ui.com/components/data-grid/columns/#fluid-width
2. Click on the first cell
3. Press right arrow key on the keyboard
4. Header sould move to the left
5. Press left arrow key on the keyboard
6. Header should go back in place

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
  Don't forget to mention which browser you used.
  Output from `npx @material-ui/envinfo` goes here.
```
</details>

Browser: Google Chrome v91.0.4472.124

flaviendelangle:
I reproduct your behavior on the published documentation.

And on the master branch I have a different behavior (but also not great).
https://user-images.githubusercontent.com/3309670/124929373-bc609480-e000-11eb-9f75-f05fa392a71e.mp4

Same thing on this PR : https://deploy-preview-2084--material-ui-x.netlify.app/components/data-grid/columns/#fluid-width

flaviendelangle:
Relates to #2011

m4theushw:
This bug also happens when all columns are not using fluid widths . The problem is that `isColumnVisibleInWindow` is returning false when the column to scroll is the last one and it ends at the very end of the viewport, leaving no space after it. In #1964 we improved this logic to consider the column width, but it seems that we missed the column-to-scroll-is-at-the-very-end scenario. If we look at the latest release, the bug is worse because it moves just by navigating to the second column. With #1964 it only moves when navigating to the last column.

[...]

Issue #2011, opened 2021-07-01
[DataGrid] Gap after the last column when all the column widths are set to fluid


ps-nchang:
A slight gap is shown between the last column and DataGrid's border when all the column's widths are set to fluid with `flex` attribute and with the following three options turned on for the DataGrid
`showCellRightBorder`, `showColumnRightBorder`, and `disableExtendRowFullWidth`

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯
When all the column's widths are set to fluid with `flex` attribute, each column width must have been calculated but there is a slight gap between the last column and the border. This behavior might have something to do with `disableExtendRowFullWidth` option is being turned on.  The is not an issue if the column widths are set with explicit pixel sizes. 
The same behavior occurs in X-Grid widget, too.

## Expected Behavior 🤔
There should be no gap between the last column and the border even when all the column's widths are set to fluid with `flex` attribute and even if `showCellRightBorder`, `showColumnRightBorder`, and `disableExtendRowFullWidth` options are set.

## Steps to Reproduce 🕹

Please see this https://codesandbox.io/s/material-demo-forked-qrcgo?file=/demo.js to show the issue. Pay attention to the last column's gap and the border.

<img width="417" alt="Capture d’écran 2021-07-24 à 23 43 12" src="https://user-images.githubusercontent.com/3165635/126881646-46e8a8d9-7295-4a3c-a2dd-fa2b08e5e1f3.png">

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
npx: installed 2 in 3.408s

  System:
    OS: macOS 10.15.7
  Binaries:
    Node: 14.16.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.12 - /usr/local/bin/npm
  Browsers:
    Chrome: 91.0.4472.114
    Edge: Not Found
    Firefox: 89.0
    Safari: 14.1.1
  npmPackages:
    @emotion/react: ^11.4.0 => 11.4.0 
    @emotion/styled: ^11.3.0 => 11.3.0 
    @material-ui/core: ^5.0.0-alpha.33 => 5.0.0-alpha.33 
    @material-ui/data-grid: ^4.0.0-alpha.30 => 4.0.0-alpha.30 
[...]

PR #2104, opened 2021-07-08
[DataGrid] Avoid trying to scroll right when focusing a cell in the last column if 100% in viewport 

flaviendelangle:
Part of #2103 (followed by #2110)

### DIFF
--- a/.github/workflows/issue-mark-duplicate.yml
+++ b/.github/workflows/issue-mark-duplicate.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: mark-duplicate
-        uses: actions-cool/issues-helper@2.x
+        uses: actions-cool/issues-helper@v2
         with:
           actions: 'mark-duplicate'
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/docs/scripts/buildApi.ts
+++ b/docs/scripts/buildApi.ts
@@ -4,7 +4,7 @@ import { writeFileSync, mkdirSync } from 'fs';
 import path from 'path';
 import kebabCase from 'lodash/kebabCase';
 import * as prettier from 'prettier';
-import { renderInline as renderMarkdownInline } from '../../node_modules/@material-ui/monorepo/docs/src/modules/utils/parseMarkdown';
+import { renderInline as renderMarkdownInline } from '../../node_modules/@material-ui/monorepo/docs/packages/markdown';
 
 type Api = {
   name: string;
--- a/package.json
+++ b/package.json
@@ -113,7 +113,6 @@
     "nyc": "^15.1.0",
     "playwright": "^1.6.1",
     "prettier": "^2.0.5",
-    "pretty-format-v24": "npm:pretty-format@24",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "rollup": "^2.28.2",
--- a/packages/grid/_modules_/grid/GridComponentProps.ts
+++ b/packages/grid/_modules_/grid/GridComponentProps.ts
@@ -76,4 +76,8 @@ export interface GridComponentProps extends GridOptionsProp {
    * Set the whole state of the grid.
    */
   state?: Partial<GridState>;
+  /**
+   * @ignore
+   */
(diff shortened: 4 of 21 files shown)


dtassone:
image
Not completely fixed.

That said, I'm not sure about the way you fix this issue. I think it is due to a scrollbar width being mistakenly added on the column header.

flaviendelangle:
@dtassone for me the issue was that `renderedColState.rightEmptyWidth === 1` in the example given in #2013 and `3` in the example given in #2011 because of the `Math.floor` in the flex column width computation.

So we add a `GridEmptyCell` at the right of the table in `GridColumnHeader`, which has a `10px padding`.
Meaning that we add `20px - 1px => 19px` to the header.

My fix reduces the  `renderedColState.rightEmptyWidth` to zero in both examples so their is no `GridEmptyCell` on the right anymore.

But I have probably broken something else on the table you screen, I'll check

dtassone:
The fix seems a bit of a hack and doesn't solve the calculation issue. We have an extra 11px on the scrollWidth of the column headers that should not be there.

m4theushw:
I did a small investigation about this bug in https://github.com/mui-org/material-ui-x/issues/2103#issuecomment-876692676. I think we can take a different direction to fix it, because the problem is not the empty cell. Not that we shouldn't remove the empty cell, but the problems is not it. 😛

flaviendelangle:
OK so from what I understand there as several issues here.

I think that Grid with at least 1 flex column should always have a total column width a least equal to the viewport width. Unless there is a better solution to avoid this gap problem : https://github.com/mui-org/material-ui-x/issues/2011#issuecomment-873189677
To do so, we can remove the `Math.floor` but I think it's not great to have non-integer column width.

---

I did reproduce your video @dtassone. 
And indeed my conclusion on the empty cell is wrong, I removed its padding and I still have the same behavior.
Do we agree that removing it's padding is still something we want to do though ?

--- 

I'll implement @m4theushw 's fix from https://github.com/mui-org/material-ui-x/issues/2103#issuecomment-876692676

flaviendelangle:
@m4theushw do you see a difference here https://github.com/mui-org/material-ui-x/pull/2104/files#diff-1c5bf15ff6f5b30287003a45d20a597797db104b7edab8d778f00e96f2dc06c1R97

Between :

```ts
if (
    lastColIndex < positions.length - 1 ||
    (positions[positions.length - 1] && totalWidth > availableWidth)
) {
    lastColIndex -= 1; // Last fully visible column
}
```

and 

```ts
const lastColumnRightPosition = positions[lastColIndex] + visibleColumns[lastColIndex].width
if (
    lastColIndex < positions.length - 1 ||
    (positions[positions.length - 1] && lastColumnRightPosition > availableWidth)
) {
    lastColIndex -= 1; // Last fully visible column
}
```

For me their are equivalent (and the 1st is lighter) but I'm not sure.

m4theushw:
> @m4theushw do you see a difference here https://github.com/mui-org/material-ui-x/pull/2104/files#diff-1c5bf15ff6f5b30287003a45d20a597797db104b7edab8d778f00e96f2dc06c1R97

Yeah, they seem to be equivalent. I think the solution is around improving this while-loop. This last decrement we do doesn't seem right. But I'm OK if we need an extra logic to check if we're at the last column.

https://github.com/mui-org/material-ui-x/blob/2a1aeef4f3d5d5da473e4c3120c49df19657e867/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualColumns.ts#L90-L94

flaviendelangle:
@m4theushw I replaced the `while` loop with a `findIndex` to avoid the `-= 1` at the end.
But the behavior is similar.

flaviendelangle:
@m4theushw I reworked the `while` loop to make it a bit more explicit.

oliviertassinari:
I have done a quick exploration in #2162 for another issue. It seems to solve this one even if it wasn't the intent. Two birds 🕊 with one stone 🗿.

oliviertassinari:
> It seems to solve this one even if it wasn't the intent. Two birds 🕊 with one stone 🗿.

I was wrong; I have posted a new comment on why in #2103

PR #2108, opened 2021-07-09
[DataGrid] Fix support for `@material-ui/core@4.12`

### LINKED ISSUE #2106: [DataGrid] Broken support of core v4.12.1
in data-grid version 4.0.0-alpha.33, i cannot build my typescript project because the data-grid.d.ts:519 references Localization$1['components'] in @material-ui/core/localization, which does not exist.
I have tried with the listed version @material-ui/core@4.9.12 and the latest version @material-ui/core@4.12.1.
The problem is also in version 4.0.0-alpha.32 but not present in 4.0.0-alpha.31 as it does not reference the localization of material ui.

The error message:
Property 'components' does not exist on type 'Localization'.

Maybe it is meant to be only for material-ui@5.0.0? i havent tried with that, as it is not tagged as the latest version yet.

DanailH:
Fixes #2106

A full explanation of the problem can be found here -> https://github.com/mui-org/material-ui-x/issues/2106#issuecomment-876996807

- v4.11.4: https://codesandbox.io/s/material-demo-forked-dtst2?file=/demo.js
- v4.12.1: https://codesandbox.io/s/material-demo-forked-skokv?file=/demo.js
- v5.0.0.beta-0: https://codesandbox.io/s/material-demo-forked-n0qqf?file=/demo.js

### DIFF
--- a/packages/grid/_modules_/grid/utils/utils.ts
+++ b/packages/grid/_modules_/grid/utils/utils.ts
@@ -26,15 +26,14 @@ export function isMuiV5(): boolean {
 }
 
 export function muiStyleAlpha(color: string, value: number): string {
-  if (isMuiV5() || !(styles as any)?.fade) {
+  if ((styles as any)?.alpha) {
     return (styles as any)?.alpha(color, value);
   }
-
   return (styles as any)?.fade(color, value);
 }
 
 export function createTheme(): styles.Theme {
-  if (isMuiV5() || !(styles as any)?.createMuiTheme) {
+  if ((styles as any)?.createTheme) {
     return (styles as any)?.createTheme();
   }
   return (styles as any)?.createMuiTheme();
--- a/packages/grid/_modules_/grid/utils/utils.ts
+++ b/packages/grid/_modules_/grid/utils/utils.ts
@@ -22,7 +22,6 @@ export function getThemePaletteMode(palette: any): string {
 }
 
 export function isMuiV5(): boolean {
-  // Changed from `alpha` to 'fade' because mui/core@4.12.0 adds 'alpha' in v4
   return !('fade' in styles);
 }
 
--- a/packages/grid/_modules_/grid/utils/utils.ts
+++ b/packages/grid/_modules_/grid/utils/utils.ts
@@ -21,14 +21,23 @@ export function getThemePaletteMode(palette: any): string {
   return palette.type || palette.mode;
 }
 
-export function isMuiV5(): boolean {
-  return !('fade' in styles);
+export function getMuiVersion(): string {
+  if (!('fade' in styles)) {
+    return 'v5';
+  }
+
+  if ('fade' in styles && 'alpha' in styles) {
+    return 'v4.12';
+  }
+
+  return 'v4';
 }
 
 export function muiStyleAlpha(color: string, value: number): string {
   if ((styles as any)?.alpha) {
     return (styles as any)?.alpha(color, value);
   }
+  // deprecated name in v4.12 and v5
   return (styles as any)?.fade(color, value);
 }
 
@@ -36,6 +45,7 @@ export function createTheme(): styles.Theme {
   if ((styles as any)?.createTheme) {
     return (styles as any)?.createTheme();
   }
+  // deprecated name in v4.12 and v5
   return (styles as any)?.createMuiTheme();
 }
 
--- a/packages/grid/_modules_/grid/utils/utils.ts
+++ b/packages/grid/_modules_/grid/utils/utils.ts
@@ -26,14 +26,15 @@ export function isMuiV5(): boolean {
 }
 
 export function muiStyleAlpha(color: string, value: number): string {
-  if (isMuiV5()) {
+  if (isMuiV5() || !(styles as any)?.fade) {
     return (styles as any)?.alpha(color, value);
   }
+
   return (styles as any)?.fade(color, value);
 }
 
 export function createTheme(): styles.Theme {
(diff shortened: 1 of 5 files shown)


packages/grid/_modules_/grid/utils/utils.ts
```diff
@@ -22,7 +22,8 @@ export function getThemePaletteMode(palette: any): string {
 }
 
 export function isMuiV5(): boolean {
-  return 'alpha' in styles;
+  // Changed from `alpha` to 'fade' because mui/core@4.12.0 adds 'alpha' in v4
```
flaviendelangle, line 25: Could we check several properties to reduce the risk of a future break ?

oliviertassinari: Nice trick but this looks like a comment for a the PR's description, not the source

```suggestion
```

oliviertassinari (dismissed):
"Update isMuiV5 helper" assuming what we go after is "Fix support for @material-ui/core@4.12", we also need to update `muiStyleAlpha` and `createTheme` to remove the deprecation warnings. We can no longer test the version, we need to feature detect.

DanailH:
@oliviertassinari Yes, I had a quick chat with @mnajdova this morning about it that was the easiest solution to the problem.

oliviertassinari:
> The drawback with doing that is that we need to also update the core version we support. Basically to use the grid one will need to use core@4.12.0 > , no ?

No need, we can feature detect.

packages/grid/_modules_/grid/utils/utils.ts
```diff
@@ -22,18 +22,19 @@ export function getThemePaletteMode(palette: any): string {
 }
 
 export function isMuiV5(): boolean {
-  return 'alpha' in styles;
+  return !('fade' in styles);
 }
 
 export function muiStyleAlpha(color: string, value: number): string {
-  if (isMuiV5()) {
+  if (isMuiV5() || !(styles as any)?.fade) {
```
oliviertassinari, line 29: Why not?

```suggestion
  if ((styles as any)?.alpha) {
```

x the other case

DanailH: I wanted to still utilize the `isMuiV5` helper but in those two cases, it doesn't make much sense.

We would still need to keep it tho, `isMuiV5` as it is used on other places as well.

DanailH:
The tests are failing because of depreciation warnings. Basically now we need a way to know if it is v4.12 because I don't see how we can detect things like `<TablePagination onChangePage />` or `<TablePagination onPageChange />`

oliviertassinari:
@DanailH Arf, maybe we should do what you have proposed initially then, drop <v4.12 versions support.

packages/grid/_modules_/grid/utils/utils.ts
```diff
@@ -21,21 +21,31 @@ export function getThemePaletteMode(palette: any): string {
...
+  if ('fade' in styles && 'alpha' in styles) {
+    return 'v4.12';
+  }
+
+  return 'v4';
 }
 
 export function muiStyleAlpha(color: string, value: number): string {
-  if (isMuiV5()) {
+  if ((styles as any)?.alpha) {
```
oliviertassinari, line 37: A good lesson for us, feature detection >> version matching.

bartvde:
any chance there will be a release soon that has this fix incorporated? We are running into: 

```Warning: Failed prop type: Invalid prop `modifiers` of type `array` supplied to `ForwardRef(Popper)`, expected `object`.``` 

because of this one. TIA.

**Comment by striky1** [comment]
> any chance there will be a release soon that has this fix incorporated? We are running into:
> 
> `` Warning: Failed prop type: Invalid prop `modifiers` of type `array` supplied to `ForwardRef(Popper)`, expected `object`. ``
> 
> because of this one. TIA.

Same here + I still have errors:

```
index.js:1 Material-UI: The key `selectLabel` provided to the classes prop is not implemented in ForwardRef(TablePagination).
You can only override one of the following: root,toolbar,spacer,caption,selectRoot,select,selectIcon,input,menuItem,actions.
```

I tried downgrade back of `/core` to `4.11.4` but it didn't help.

Current package.json:

```
    "@material-ui/core": "^4.12.2",
    "@material-ui/data-grid": "^4.0.0-alpha.33",
    "@material-ui/icons": "^4.11.2",
    "@material-ui/lab": "^4.0.0-alpha.58",
    "@material-ui/pickers": "^3.3.10",
    "@material-ui/styles": "^4.11.4",
```

Thank you

**Comment by Primajin** [comment]
```
    "@material-ui/core": "^4.12.2",
    "@material-ui/icons": "^4.11.2",
    "@material-ui/lab": "4.0.0-alpha.60",
    "@material-ui/pickers": "^3.3.10",
    "@material-ui/x-grid": "^4.0.0-alpha.34",
```
seems to have resolved the issue

striky1:
> ```
>     "@material-ui/core": "^4.12.2",
>     "@material-ui/icons": "^4.11.2",
>     "@material-ui/lab": "4.0.0-alpha.60",
>     "@material-ui/pickers": "^3.3.10",
>     "@material-ui/x-grid": "^4.0.0-alpha.34",
> ```
> 
> 
>      
>   
> 
> seems to have resolved the issue

but you are using `x-grid` instead of `data-grid` what is not MIT version

oliviertassinari:
@striky1 He's a Pro plan user. This fix was released in MUI X version: 4.0.0-alpha.34, regardless of the plan (MIT or commercial).

PR #2110, opened 2021-07-09
[DataGrid] Fix flex columns and resize

### LINKED ISSUE #2011: [DataGrid] Gap after the last column when all the column widths are set to fluid

### LINKED ISSUE #2103: [DataGrid] Column header is moving when keyboard arrow keys are pressed

dtassone:
follow up from #2104 
fix #2011 
fix #2103 

Preview: https://deploy-preview-2110--material-ui-x.netlify.app/components/data-grid/columns/#fluid-width

### DIFF
--- a/packages/grid/_modules_/grid/components/cell/GridEmptyCell.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridEmptyCell.tsx
@@ -22,6 +22,7 @@ export const GridEmptyCell = React.memo(function GridEmptyCell({
         lineHeight: `${height - 1}px`,
         minHeight: height,
         maxHeight: height,
+        padding: 0,
       }}
       className={GRID_CELL_CSS_CLASS}
     />
--- a/packages/storybook/src/stories/grid-columns.stories.tsx
+++ b/packages/storybook/src/stories/grid-columns.stories.tsx
@@ -1,3 +1,4 @@
+import { useDemoData } from '@material-ui/x-grid-data-generator';
 import * as React from 'react';
 import { GridColDef, XGrid, GridColTypeDef, GridValueGetterParams } from '@material-ui/x-grid';
 import CreateIcon from '@material-ui/icons/Create';
@@ -363,3 +364,31 @@ export const SingleSelectColumnType = () => {
     </div>
   );
 };
+export function FlexLayoutGrid() {
+  const { data } = useDemoData({
+    dataSet: 'Commodity',
+    rowLength: 15,
+    maxColumns: 6,
+  });
+
+  const tmpData = JSON.parse(JSON.stringify(data));
+  tmpData.columns.forEach((item) => {
+    item.flex = 1;
+  });
+  return (
+    <div style={{ height: '400px', width: '400px' }}>
+      <div style={{ display: 'flex', height: '100%' }}>
+        <div style={{ flexGrow: 1 }}>
+          <XGrid
+            autoHeight
+            showCellRightBorder
+            showColumnRightBorder
+            disableExtendRowFullWidth
+            rows={tmpData.rows}
+            columns={tmpData.columns}
+          />
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSeparator.tsx
+++ b/packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSeparator.tsx
@@ -11,7 +11,7 @@ export interface GridColumnHeaderSeparatorProps extends React.HTMLAttributes<HTM
   height: number;
 }
 
-export const GridColumnHeaderSeparator = React.memo(function GridColumnHeaderSeparator(
+export const GridColumnHeaderSeparator = function GridColumnHeaderSeparator(
   props: GridColumnHeaderSeparatorProps,
 ) {
   const { resizable, resizing, height, ...other } = props;
@@ -38,4 +38,4 @@ export const GridColumnHeaderSeparator = React.memo(function GridColumnHeaderSep
       <ColumnResizeIcon className="MuiDataGrid-iconSeparator" />
     </div>
   );
-});
+};
--- a/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
+++ b/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
@@ -157,7 +157,6 @@ export const useStyles = makeStyles(
         '& .MuiDataGrid-columnSeparator': {
           position: 'absolute',
           right: 0,
-          zIndex: 100,
           display: 'flex',
           flexDirection: 'column',
           justifyContent: 'center',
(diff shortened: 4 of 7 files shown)


packages/storybook/src/stories/grid-columns.stories.tsx
```diff
@@ -363,3 +364,31 @@ export const SingleSelectColumnType = () => {
...
+            showColumnRightBorder
+            disableExtendRowFullWidth
+            rows={tmpData.rows}
+            columns={tmpData.columns}
+          />
+        </div>
+      </div>
+    </div>
+  );
+}
```
oliviertassinari, line 394: We can already reproduce the issue in the docs. What's the use case for the story?

```suggestion
```

packages/grid/_modules_/grid/components/cell/GridEmptyCell.tsx
```diff
@@ -22,6 +22,7 @@ export const GridEmptyCell = React.memo(function GridEmptyCell({
         lineHeight: `${height - 1}px`,
         minHeight: height,
         maxHeight: height,
+        padding: 0,
```
oliviertassinari, line 25: This looks off topic

```suggestion
```

oliviertassinari (changes requested):
- We miss a test case
- There is a CSS regression, see argos once it has finish running

packages/grid/_modules_/grid/components/cell/GridEmptyCell.tsx
flaviendelangle: The issue is that the cell always have a padding of 10px
But we do not want it on the GridEmptyCell.

Maybe we should add a class to this cell and set the padding to 0 in the rootStyle instead of here.

dtassone:
> * We miss a test case
> * There is a CSS regression, see argos once it has finish running

Yes I moved the separator to avoid negative margin that creates issue on resize

packages/grid/_modules_/grid/components/cell/GridEmptyCell.tsx
dtassone: yes

packages/storybook/src/stories/grid-columns.stories.tsx
dtassone: working on flex columns, and see if it works. Where is #2011 in the doc?

m4theushw:
About #2011, the double border is still there.

image

The separator is being displayed when `showColumnRightBorder` is used.

image

dtassone:
> About #2011, the double border is still there.

A separate issue that can be fixed in another PR. We need to keep PR small

m4theushw:
I'm confused, we have this PR and #2104 both trying to fix #2103.

About the solution proposed to fix it here, it's still trying to scroll, but since we removed the padding from the empty cell it can't. We're not fixing the root cause https://github.com/mui-org/material-ui-x/issues/2103#issuecomment-876692676. To reproduce, use fixed size columns.

https://user-images.githubusercontent.com/42154031/125085562-1a44b900-e0a1-11eb-9730-9755e87405e6.mp4

flaviendelangle:
@m4theushw for me this PR is a follow up on #2004 because there are several issues to be fixed around the column width.

The scroll event should be fixed here https://github.com/mui-org/material-ui-x/pull/2104#issuecomment-876965867

dtassone:
the separator creates an extra 10px on the columns header due to its positioning. The problem is the icon shape that is a square 24x24 instead of just a bar 24x2 which would be easier to position

dtassone:
I managed to fix the icon and put the bar on the right side of the icon so we can position it properly without using negative margin. let me know what you think

packages/storybook/src/stories/grid-columns.stories.tsx
oliviertassinari: Ok, could we make it a Snap story in this case? This could complement the function test to make sure the rendering with sub-pixels is still correct

oliviertassinari (changes requested):
Please isolate the fix of #2011 and #2103. Looking at the changes, here are no reason to bundle them.

oliviertassinari:
We can resume the effort later on

PR #2112, opened 2021-07-09
[DataGrid] Fix support for singleSelect with numeric values

### LINKED ISSUE #2111: [DataGrid] Single select column type not filtering with number values
- [x ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

The 'singleSelect' filters everything when the column values are numbers. The valueOptions are specified as objects: [{label:"One", value:1}, {label:"Two", value: 2}]

## Expected Behavior 🤔

Should work like an equal filter.

m4theushw:
Fixes #2111 

Preview: https://deploy-preview-2112--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-columns--single-select-column-type&globals=measureEnabled:false

Column "country" has as options `{ value: string; label: string }[]`
Column "fruit" has as options `{ value: number; label: string }[]`
Column "rating" has as options `number[]`


packages/grid/_modules_/grid/models/colDef/gridSingleSelectOperators.ts
```diff
@@ -6,13 +6,14 @@ export const getGridSingleSelectOperators: () => GridFilterOperator[] = () => [
   {
     value: 'is',
     getApplyFilterFn: (filterItem: GridFilterItem) => {
-      if (!filterItem.columnField || !filterItem.value || !filterItem.operatorValue) {
```
m4theushw, line 9: We're already validating if there's a `columnField` and `operatorValue` before calling the filter function. Related to https://github.com/mui-org/material-ui-x/issues/1961#issuecomment-869177167

docs/pages/api-docs/data-grid/grid-col-def.md
```diff
@@ -38,6 +38,6 @@ import { GridColDef } from '@material-ui/data-grid';
...
 | <span class="prop-name optional">valueGetter<sup><abbr title="optional">?</abbr></sup></span>       | <span class="prop-type">(params: GridValueGetterParams) =&gt; GridCellValue</span>                 |                                                  | Function that allows to get a specific data instead of field to render in the cell.                             |
-| <span class="prop-name optional">valueOptions<sup><abbr title="optional">?</abbr></sup></span>      | <span class="prop-type">(string \| { label: string; value: any })[]</span>                         |                                                  | To be used in combination with `type: 'singleSelect'`. This is an array of the possible cell values and labels. |
+| <span class="prop-name optional">valueOptions<sup><abbr title="optional">?</abbr></sup></span>      | <span class="prop-type">any[]</span>                                                               |                                                  | To be used in combination with `type: 'singleSelect'`. This is an array of the possible cell values and labels. |
```
flaviendelangle, line 41: Why not using the same type as https://github.com/mui-org/material-ui-x/pull/2112/files#diff-011e3c058a45f4b4092aa522938afa28991cc4f218800f436d9f1b24c1c48195R77 ?

m4theushw: It seems to be related to the docs generator. I'll open another PR to fix it.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -44,16 +44,23 @@ export function GridFilterInputValue(props: GridTypeFilterInputValueProps & Text
 
   const onFilterChange = React.useCallback(
     (event) => {
+      let value = event.target.value;
+      if (type === 'singleSelect') {
```
oliviertassinari, line 48: It's not obvious why we do this. Maybe would make it clearer?

```suggestion
      // NativeSelect only cast the value to a string. 
      if (type === 'singleSelect') {
```

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -626,49 +648,158 @@ describe('<DataGrid /> - Filter', () => {
...
+              filter: {
+                items: [{ id: 123, columnField: 'voltage', value: '', operatorValue: 'is' }],
+                linkOperator: GridLinkOperator.And,
+              },
+            }}
+          />,
+        );
+        expect(getColumnValues()).to.deep.equal(['Hair Dryer', 'Dishwasher', 'Microwave']);
+        fireEvent.change(screen.getByLabelText('Value'), { target: { value: '220' } });
+        clock.tick(600);
```
flaviendelangle, line 733: Could we document why we need to wait 600ms ?

docs/pages/api-docs/data-grid/grid-col-def.md
m4theushw: This is not a docs bug. Typing with `any` means that the object definition is useless. I updated to only allow string and numbers.

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
m4theushw, line 733:
```suggestion
        clock.tick(600); // Wait for the debounce
```

flaviendelangle (approved):

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -74,7 +74,7 @@ export interface GridColDef {
   /**
    * To be used in combination with `type: 'singleSelect'`. This is an array of the possible cell values and labels.
    */
-  valueOptions?: Array<any | { value: any; label: string }>;
+  valueOptions?: Array<string | number | { value: any; label: string }>;
```
dtassone, line 77: Can we use GridCellValue?

flaviendelangle: I don't think we currently support `Date`

It will be considered as an object here https://github.com/mui-org/material-ui-x/pull/2112/files#diff-e03ef6ac8071c6dbcd6ceddc294a6ceb9692826571ae45332126f34b953d35efR8

dtassone: How would I do a single select for a list of country for example, with the flag and country name? 
I think the select options should consider renderCell, valueFormatter or getter...

flaviendelangle: I'm not following you

dtassone: Well it's off topic: but I think this feature is a bit too basic at this stage and should consider renderCell, valueGetter, valueFormatter...

flaviendelangle: This PR add the support of `number`, it's a step forward but I do agree that we still don't support everything we should

m4theushw: This `singleSelect` is only a column type, like the others. The user can provide a custom cell or a different data structures.

PR #2116, opened 2021-07-10
[DataGrid] Support style prop

oliviertassinari:
If we support the `className` prop, then it makes sense to support the `style` prop too. I got trapped in https://github.com/mui-org/material-ui/pull/27196.

### DIFF
--- a/packages/grid/x-grid/src/tests/layout.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/layout.XGrid.test.tsx
@@ -58,7 +58,26 @@ describe('<XGrid /> - Layout', () => {
         </div>,
       );
 
-      expect(document.querySelector(`.${className}`)).to.equal(container.firstChild.firstChild);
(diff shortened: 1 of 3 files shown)


packages/grid/x-grid/src/tests/layout.XGrid.test.tsx
```diff
@@ -58,7 +58,26 @@ describe('<XGrid /> - Layout', () => {
...
+            {...baselineProps}
+            style={{
+              border: 0,
+            }}
+          />
+        </div>,
+      );
+
+      // @ts-expect-error need to migrate helpers to TypeScript
+      expect(document.querySelector('.MuiDataGrid-root')).toHaveInlineStyle({
```
oliviertassinari, line 78: I didn't use the role="grid" because I trust more that we set the root class name on the root than the aria.

dtassone (approved):

PR #2130, opened 2021-07-12
[core] Correctly test column visibility switch impact on column width

flaviendelangle:
@oliviertassinari this do fail when using a commit older than https://github.com/mui-org/material-ui-x/pull/2034


oliviertassinari (commented):
Thanks for the cleanup, looks great. I guess we could have kept the other test if we have no other like this.

flaviendelangle:
As you want, it does indeed test something that could break

oliviertassinari (approved):

PR #2131, opened 2021-07-12
[test] Remove Firefox and Edge from BrowserStack

m4theushw:


oliviertassinari (approved):
More context: https://github.com/mui-org/material-ui/pull/27179#issuecomment-877678754,  I'm not sure that this change will really have an impact (most of the concurrency is used by the core), but we have the stats in CircleCI so we can always try and compare one week later.

oliviertassinari:
From the last 7 days:

Capture d’écran 2021-07-17 à 16 55 16

oliviertassinari:
I'm closing, we are back to a more normal DX, last 7 days

Capture d’écran 2021-07-24 à 23 40 02

PR #2139, opened 2021-07-13
[core] Use props instead of options for event handler

flaviendelangle:
The `options` pattern should be removed in the future now that we pass the props as the 2nd argument of the hooks.


flaviendelangle:
The CI error comes from https://github.com/mui-org/material-ui-x/pulls
So nothing related to this PR :+1:

oliviertassinari (commented):
For the issues with the CI, you can rebase on HEAD. We should have fixed it all now.
For the removal of the options. For me, the real problem is that we update them in an effect, this is not great. As long as we fix this, no preferences on either we keep this notion of options or merge everything under props. In any cases, we will need to find a way to differentiate top level props (currently "options") with sub component props.

flaviendelangle:
For variables coming from the props, without any default value, I don't see a benefit using a prop to state pattern.
I do agree that having it in a useEffect causes additional problems. And that for some variables with more complex behavior we may want to keep something similar to the option but without the useEffect (not sure though).

But for the event handlers I think using directly the props is enough and removes a layer of complexity.

m4theushw (approved):
We only have to use the options selector for `classes` and `localeText`. Besides these two, using the props is better.

PR #1994, opened 2021-06-29
[DataGrid] Allow passing styles and Popper props to GridPanel

sebastianfrey:
Closes #1808.

PR #2117, opened 2021-07-10
[DataGrid] Fix localeText type

oliviertassinari:
Fix the build of https://github.com/mui-org/material-ui/pull/27196. See the added test cases for a simpler reproduction of the issue.

It's a follow-up on https://github.com/mui-org/material-ui-x/pull/1913/files#r667394689, required to fix #2039, with https://github.com/mui-org/material-ui/pull/27196.

Issue #1902, opened 2021-06-13
[DataGrid] First class support of v5


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 🔦

A list of tasks to do in order to drop @material-ui/core v4 and only support v5 going forward.

## Details

- [x] Wait for v5 to have enough [adoption](https://docs.google.com/spreadsheets/d/1KP94DulRRmUk2eBCWwhYQYIbL1wvoLFD0tKeLqeRH8c/edit#gid=609237852) -> 1/3 relative to v4 could be a great threshold to prioritize this effort.
- [x] Update docs, createMuiTheme->createTheme #2281
- [x] Migrate the docs and dev tooling to v5 two weeks before the stable release of v5. #2441 
- [ ] Remove `getDoc` prop. See https://github.com/mui-org/material-ui/pull/26696 for why. It solves our cross-document issue at the same time.  This is wrong ⬇️

https://github.com/mui-org/material-ui-x/blob/9dd5cf94061d1733566b7b66982fae554ebcf242/packages/grid/_modules_/grid/components/panel/GridPanelWrapper.tsx#L28

First step in #2224

- [ ] Remove duplicated modules, for instance: https://github.com/mui-org/material-ui-x/blob/a4df486a95b72847224b71291fcdc3211feea3d1/packages/grid/_modules_/grid/utils/material-ui-utils.ts#L7
- [ ] Remove conditional code. #2575 
https://github.com/mui-org/material-ui-x/blob/1a0ad36f7380ba6bff04fa8f2b6366e001988734/packages/grid/_modules_/grid/components/GridPagination.tsx#L107
- [ ] Revert 36124c6?
- [x] Rename imports from @material-ui to @mui. #2571

bestickley:
I have "@material-ui/core": "^5.0.0-beta.0" installed (along with peer deps) and "@material-ui/data-grid": "^4.0.0-alpha.33" installed. I'm getting the error:
```
error when starting dev server:
Error: Build failed with 1 error:
node_modules/@material-ui/data-grid/dist/index-esm.js:1:433: error: Could not resolve "@material-ui/styles" (mark it as external to exclude it from the bundle)
    at failureErrorWithLog (/Users/stickb/Code/green-boost/starter/ui/node_modules/esbuild/lib/main.js:1449:15)
[...]

PR #2140, opened 2021-07-13
[core] Copy getClasses from the core (removed in v5)

flaviendelangle:
#1994 starts using `getClasses` that was recently removed from the core (due to the migration away from `enzyme`)
See https://github.com/mui-org/material-ui/commit/2569eb9d40c648a73e62860fd993ad535ca9af5d#diff-f075d81b14df1f31fb0699c4516fe3d78490351867feeb4fb869d5dbe8c6128e

@oliviertassinari


flaviendelangle:
https://app.circleci.com/pipelines/github/mui-org/material-ui-x/6063/workflows/274e60b1-d36b-4d0d-a6be-dbb0b3222d88/jobs/31907

Copy paste doen't work, I'll try to fix

oliviertassinari:
I meant:

```diff
diff --git a/packages/grid/_modules_/grid/components/panel/GridPanel.test.tsx b/packages/grid/_modules_/grid/components/panel/GridPanel.test.tsx
index bf5027df..b8d6e468 100644
--- a/packages/grid/_modules_/grid/components/panel/GridPanel.test.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridPanel.test.tsx
@@ -4,11 +4,9 @@ import {
   // @ts-expect-error JS
   createMount,
   // @ts-expect-error JS
-  getClasses,
-  // @ts-expect-error JS
   describeConformance,
 } from 'test/utils';
-import { GridPanel, useGridApiRef, GridApiContext } from '@material-ui/data-grid';
+import { GridPanel, gridPanelClasses as classes, useGridApiRef, GridApiContext } from '@material-ui/data-grid';
 import { Popper } from '@material-ui/core';

 describe('<GridPanel />', () => {
@@ -16,8 +14,6 @@ describe('<GridPanel />', () => {
   // TODO v5: replace with createClientRender
   const render = createClientRenderStrictMode();

-  let classes;
-
   function Wrapper(props) {
     const apiRef = useGridApiRef();
     apiRef.current.columnHeadersContainerElementRef = {
@@ -28,10 +24,6 @@ describe('<GridPanel />', () => {
     return <GridApiContext.Provider value={apiRef} {...props} />;
   }

-  before(() => {
-    classes = getClasses(<GridPanel open={false} />);
-  });
-
   describeConformance(<GridPanel disablePortal open />, () => ({
     classes,
     inheritComponent: Popper,
diff --git a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
index 7f5416ed..ac458b4a 100644
--- a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
@@ -41,6 +41,11 @@ const useStyles = makeStyles(
   { name: 'MuiGridPanel', defaultTheme },
 );

+export const gridPanelClasses = {
+  root: 'MuiGridPanel-root',
+  paper: 'MuiGridPanel-paper',
+};
+
 export const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>(function GridPanel(
   props,
   ref,
@@ -103,6 +108,3 @@ export const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>(functi
     </Popper>
   );
 }) as (props: GridPanelProps) => JSX.Element;
-
-// @ts-ignore TODO migrate to v5 gridPanelClasses pattern, this is only for tests
-GridPanel.useStyles = useStyles;
```

oliviertassinari:
@mui-org/x think that we need to take over this effort as Flavien is off until next Monday.

m4theushw:
@oliviertassinari I can take it as it's breaking master and blocking all new PRs.

m4theushw:
@oliviertassinari Do you have any suggestion to fix the Netlify deploy? I think it fails because it's using a node_modules cache.

oliviertassinari:
@m4theushw It's flaky, I don't kno why

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "noImplicitAny": false,
     "types": ["node", "requestidlecallback"],
     "baseUrl": "./",
+    "skipLibCheck": true,
```
oliviertassinari, line 21: I had a look locally for why this line. I get:

Capture d’écran 2021-07-14 à 21 50 04

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import clsx from 'clsx';
 import { Theme } from '@material-ui/core/styles';
 import { makeStyles } from '@material-ui/styles';
+import { generateUtilityClasses } from '@material-ui/unstyled';
```
oliviertassinari, line 5: unstyled is not a dependency of the packages, it would break people's build. Trying to open a codesandbox from the docs preview. 

Would this approach work instead? https://github.com/mui-org/material-ui-x/blob/a4df486a95b72847224b71291fcdc3211feea3d1/packages/grid/_modules_/grid/utils/material-ui-utils.ts#L61

m4theushw: Hmm I noticed this utils file but I found easier to import from @material-ui/unstyled. I didn't realize that it would break the build in v4. I'll add the `generateUtilityClasses` to it.

oliviertassinari: Actually, I'm updating #1902 so we don't forget about this

tsconfig.json
m4theushw: We had a similar problem in #2106. The docs in the core already enables this flag: https://github.com/mui-org/material-ui/blob/8a4b8952077b64e90ac9d23c116a45aed672db9c/docs/tsconfig.json#L11 TypeScript also encourages to set it to true: https://www.typescriptlang.org/tsconfig#skipLibCheck

If we don't want to encourage enabling this flag, we should address this in the core.

oliviertassinari: > We had a similar problem in #2106

I assume it's a reference to `Property 'components' does not exist on type 'Localization'.`. It seems to be the problem in reverse (thrid-party that tries to use data-grid but fail because its internal types are wrong). I have faced it when working on https://github.com/mui-org/material-ui/pull/27196, and fixed it in #2117.

oliviertassinari: Regarding `skipLibCheck`, TypeScript has the value to false by default. My concern is with developers that would complain because we break their build.

oliviertassinari (commented):
It's green 🎉

tsconfig.json
m4theushw: We don't need the `skipLibCheck` anymore, I already removed it. Since we created our local classes utility there're no more errors from @material-ui/unstyled.

oliviertassinari:
@m4theushw Thanks. I'm rebasing my other PR now :)

PR #2141, opened 2021-07-13
[DataGrid] Remove stateId argument from GridApi getState method

flaviendelangle:
@dtassone, we only have 2 calls of `getState` with a `stateId`, both in the same hook (`useGridFilter`)
Is this something we want to keep ?
95% of the time, the code written is `const page = apiRef.current.getState().page` instead of `const page = apiRef.current.getState('page')` so I'm not sure the overhead is worth it here.

But with this PR you can use the `stateId` with the correct typing without specifying it as a generic.

@m4theushw the doc generation does not like the generic, I'm loosing the `StateId` definition here. Same thing for the type returned.
It could be interesting to handle it. But I would more be in favor of droping the `stateId` param on `getState` (see above)


oliviertassinari (commented):
> 95% of the time, the code written is

Then +1 for removing these 5% percent. For instance, it's not supported in https://react-redux.js.org/api/hooks#usestore.

```diff
-const filterState = apiRef.current.getState('filter');
+const filterState = apiRef.current.getState().filter;
```

flaviendelangle:
In general, I'm in favor of avoiding methods with separate behavior like this one.
If the `stateId` was something usefull, I would have prefered splitting the two :

```ts
getState: () => GridState
getSubState: <StateId extends keyof GridState>(stateId: StateId) => GridState[StateId]
```

packages/grid/_modules_/grid/models/api/gridStateApi.ts
```diff
@@ -11,7 +11,7 @@ export interface GridStateApi {
    * @param {string} stateId The part of the state to be returned.
    * @returns {any} The whole state or part of it.
```
m4theushw, line 12: We also need to remove the "If `stateId` is present, only the referred part is returned." from the main description. 

```suggestion
   * @returns {any} The whole state.
```

--- 

I'll fix the issue with the docs generator in another PR. We might have other cases where we miss the generics.

oliviertassinari:
The API is public, I have added a new label. If we could update the description of the PR to document it, it would be perfect. Actually, are we confortable with making the whole state public? If we are not, we can make the method private.

m4theushw:
> Actually, are we confortable with making the whole state public? If we are not, we can make the method private.

@oliviertassinari Yes, I would keep the state and this method public. For some parts of the state (e.g. `focus`) there's no alternative way to access them besides using `getState`.

flaviendelangle:
For the stable version, we may want to make it private and only expose methods for the wanted substates.
That would allow us to consider the other sub states as internal and change their structure without any breaking change.

But for now I do agree that exposing it is simpler.

oliviertassinari:
OK, I guess that once the component is stable and we need to change one part of the state, then we can change the public `getState` method to keep it unchanged for developers.

m4theushw (approved):

PR #2143, opened 2021-07-14
[DataGrid] Improve controllable cell edit

m4theushw:
### Breaking changes

- [XGrid] Improve cell edit controllable behavior

  - The `cellEditPropsChange` event was renamed to `editCellPropsChange`.
  - The `cellEditPropsChangeCommitted` event was renamed to `cellEditCommit`.
  - The `cellValueChange` event was removed. Listen to `cellEditCommit` to detect when the value is committed.
  - The `editRowModelChange` event was renamed to `editRowsModelChange`.

- [DataGrid] Improve cell edit controllable behavior

  - The `onEditCellChange` prop was renamed to `onEditCellPropsChange`.
  - The `onEditCellChangeCommitted` prop was renamed to `onCellEditCommit`.
  - The `onEditRowModelChange` prop was removed. Use the new `onEditRowsModelChange` prop.
    ```diff
    -onEditRowModelChange?: (params: GridEditRowModelParams)
    +onEditRowsModelChange?: (editRowsModel: GridEditRowsModel)
    ```

---

Part of #2044

Preview: https://deploy-preview-2143--material-ui-x.netlify.app/components/data-grid/editing/


packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -315,48 +274,58 @@ export function useGridEditRows(apiRef: GridApiRef) {
...
   useGridApiOptionHandler(apiRef, GRID_CELL_MODE_CHANGE, options.onCellModeChange);
-  useGridApiOptionHandler(apiRef, GRID_ROW_EDIT_MODEL_CHANGE, options.onEditRowModelChange);
 
   useGridApiMethod<GridEditRowApi>(
     apiRef,
     {
       setCellMode,
       getCellMode,
-      isCellEditable,
+      isCellEditable, // TODO don't expose, user already controls which cells are editable
```
m4theushw, line 326: This method only serves to set `isEditable` in `GridCellParams` whose single usage is inside this very hook. We could call it directly.

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -146,22 +146,16 @@ export const GRID_CELL_DRAG_OVER = 'cellDragOver';
 export const GRID_CELL_DRAG_END = 'cellDragEnd';
 
 /**
- * Fired when the props of the edit input changes. Called with a [[GridEditCellPropsParams]] object.
+ * Fired when the props of the edit cell changes. Called with a [[GridEditCellPropsParams]] object.
  * @event
  */
-export const GRID_CELL_EDIT_PROPS_CHANGE = 'cellEditPropsChange';
+export const GRID_EDIT_CELL_PROPS_CHANGE = 'editCellPropsChange';
```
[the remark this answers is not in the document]
oliviertassinari: In the naming of the public API, do we want to consider the "cell edit" feature the same as the "row edit" feature?

Did you consider this prefix convention?

- `GRID_EDIT`: cell + row edit
- `GRID_CELL_EDIT`: cell edit
- `GRID_ROW_EDIT`: row edit

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -315,48 +274,58 @@ export function useGridEditRows(apiRef: GridApiRef) {
     [apiRef],
   );
 
-  const handleDoubleClick = React.useCallback(
+  const handleCellEditExit = React.useCallback(
+    (params: GridCellParams, event?: React.SyntheticEvent) => {
+      // TODO check if its propagation was stopped
```
oliviertassinari, line 279: To be clearer, the event would continue to propagate

```suggestion
      // TODO check if event internal handling was prevented
```

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -315,48 +274,58 @@ export function useGridEditRows(apiRef: GridApiRef) {
...
+      // When dispatched by the document, the event is not passed
+      if (!event || !isKeyboardEvent(event)) {
+        return;
+      }
+
+      if (isCellEditCommitKeys(event.key)) {
+        apiRef.current.publishEvent(GRID_CELL_NAVIGATION_KEY_DOWN, params, event);
+        return;
+      }
+      if (isEscapeKey(event.key) || isDeleteKeys(event.key)) {
```
oliviertassinari, line 291: Time to stop using abstractions that don't solve problems? (we need isDeleteKeys)

```suggestion
      if (event.key === 'Enter' || isDeleteKeys(event.key)) {
```

docs/src/pages/components/data-grid/editing/ValidateRowModelControlGrid.js
```diff
@@ -65,7 +65,7 @@ export default function ValidateRowModelControlGrid() {
         rows={rows}
         columns={columns}
         editRowsModel={editRowsModel}
-        onEditCellChange={handleEditCellChange}
+        onEditCellPropsChange={handleCellEditPropsChange}
```
oliviertassinari, line 68: The demo is no longer working, test https://deploy-preview-2143--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation vs. https://material-ui.com/components/data-grid/editing/#client-side-validation.

Shouldn't we use? 

```suggestion
        onEditRowsModelChange={handleCellEditPropsChange}
```

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -315,48 +274,58 @@ export function useGridEditRows(apiRef: GridApiRef) {
...
-  useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_PROPS_CHANGE, options.onEditCellChange);
-  // TODO remove because onEditCellChange does the same
-  useGridApiOptionHandler(apiRef, GRID_CELL_VALUE_CHANGE, options.onCellValueChange);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_ENTER, handleCellEditEnter);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_EXIT, handleCellEditExit);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_COMMIT, handleCellEditCommit);
+  useGridApiEventHandler(apiRef, GRID_EDIT_CELL_PROPS_CHANGE, handleEditCellPropsChange);
+
+  useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_COMMIT, options.onCellEditCommit);
+  useGridApiOptionHandler(apiRef, GRID_EDIT_CELL_PROPS_CHANGE, options.onEditCellPropsChange);
```
oliviertassinari, line 317: I don't get the value proposition of `onEditCellPropsChange`.

```suggestion
  // TODO remove, use onEditRowsModelChange directly
  useGridApiOptionHandler(apiRef, GRID_EDIT_CELL_PROPS_CHANGE, options.onEditCellPropsChange);
```

oliviertassinari (commented):
It's a great step for the cleanup 👍

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
DanailH: Am it should be `Escape` instead of `Enter`.

dtassone:
> ### Breaking changes
> * [XGrid] Improve cell edit controllable behavior
>   
>   * The `cellEditPropsChange` event was renamed to `editCellPropsChange`.
>   * The `cellEditPropsChangeCommitted` event was renamed to `cellEditCommit`.

I think it should be renamed to `editCellPropsCommit`

>   * The `cellValueChange` event was removed. Listen to `cellEditCommit` to detect when the value is committed.
>   * The `editRowModelChange` event was renamed to `editRowsModelChange`.
> * [DataGrid] Improve cell edit controllable behavior
>   
>   * The `onEditCellChange` prop was renamed to `onEditCellPropsChange`.
>   * The `onEditCellChangeCommitted` prop was renamed to `onCellEditCommit`.

And here it should be `onEditCellPropsCommit`

>   * The `onEditRowModelChange` prop was removed. Use the new `onEditRowsModelChange` prop.
>     ```diff
>     -onEditRowModelChange?: (params: GridEditRowModelParams)
>     +onEditRowsModelChange?: (editRowsModel: GridEditRowsModel)
>     ```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -58,23 +58,22 @@ The editable cells have a green background for better visibility.
...
 {{"demo": "pages/components/data-grid/editing/ValueGetterGrid.js", "bg": "inline"}}
 
 ### Client-side validation
 
-Follow the following steps to validate the value in the cells:
-
-- Set the event handler `onEditCellChange`. It's invoked when a change is triggered by the edit input component.
-- Set the event handler `onEditCellChangeCommitted` to validate or persist the new value. This handler is invoked when an end-user requests a change to be committed and before the cell reverts to view mode.
+To validate the value in the cells, use `onEditCellPropsChange` to set the `error` attribute of the respective field when the value is invalid.
+Once this attribute is true, the value will never be commited.
```
dtassone, line 75: `If the error attribute is true... `

docs/src/pages/components/data-grid/editing/ValidateRowModelControlGrid.js
m4theushw: Fixed. All demos now use `onEditRowsModelChange`.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: This `isDeleteKeys`  and other similar abstractions we could bring to inside the hook. The probability of using it again in another feature is very small.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: No, I was meaning to add the following code:

```ts
if (!params.isEditable || event.isPropagationStopped()) {
  return;
}
```

`handleCellEditEnter` already has it but this method doesn't. However, with #1403 all of this manual checks will go away.

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -207,25 +211,18 @@ export interface GridOptions {
...
+  onEditCellPropsChange?: (params: GridEditCellPropsParams, event?: React.SyntheticEvent) => void;
   /**
-   * Callback fired when the EditRowModel changed.
-   * @param handler
+   * Callback fired when the cell changes are going to be committed.
+   * @param {GridEditCellPropsParams} params With all properties from [[GridEditCellPropsParams]].
+   * @param {React.SyntheticEvent} event The event that caused this prop to be called.
    */
-  onEditRowModelChange?: (params: GridEditRowModelParams) => void;
+  onCellEditCommit?: (params: GridEditCellPropsParams, event?: React.SyntheticEvent) => void;
```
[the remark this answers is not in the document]
m4theushw: According to our event naming convention, we must use the present tense. Here the prop was renamed to follow the event it listens to.  We don't need to rename other props because this was the only one that was wrong.

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -207,25 +211,18 @@ export interface GridOptions {
...
    */
-  onEditCellChangeCommitted?: (
-    params: GridEditCellPropsParams,
-    event?: React.SyntheticEvent,
-  ) => void;
+  onEditCellPropsChange?: (params: GridEditCellPropsParams, event?: React.SyntheticEvent) => void;
   /**
-   * Callback fired when the EditRowModel changed.
-   * @param handler
+   * Callback fired when the cell changes are going to be committed.
```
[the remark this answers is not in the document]
m4theushw: Renamed. My argument to use "going to" was to emphasize that if the user wants to make a server request **before** updating the row this is the callback that he needs to use.

packages/grid/_modules_/grid/constants/eventsConstants.ts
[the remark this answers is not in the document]
m4theushw: > Did you consider this prefix convention?

Yes, GRID_EDIT_CELL already follows this convention.

m4theushw:
> I think it should be renamed to editCellPropsCommit

@dtassone My rationale is that we don't commit all props, but only the value prop. Also, `cellEditCommit` is simpler and would resonate well with its row counterpart: `rowEditCommit`. I'm open for suggestions.

> And here it should be onEditCellPropsCommit

The prop matches the event, so whatever name we decide it will be renamed as well.

packages/grid/_modules_/grid/constants/eventsConstants.ts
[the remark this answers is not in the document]
oliviertassinari: Ok, I got confused when I saw an event with `GRID_EDIT` as a prefix but again CELL later on. It makes sense.

packages/grid/_modules_/grid/models/gridOptions.tsx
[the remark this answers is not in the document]
oliviertassinari: I meant that we will have https://next.material-ui.com/api/slider/ `onChangeCommitted` to rename.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -326,48 +273,59 @@ export function useGridEditRows(
...
-    props.onEditCellChangeCommitted,
-  );
-  // TODO rename to onEditCellPropsChange
-  useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_PROPS_CHANGE, props.onEditCellChange);
-  // TODO remove because onEditCellChange does the same
-  useGridApiOptionHandler(apiRef, GRID_CELL_VALUE_CHANGE, props.onCellValueChange);
-  useGridApiOptionHandler(apiRef, GRID_CELL_MODE_CHANGE, props.onCellModeChange);
-  useGridApiOptionHandler(apiRef, GRID_ROW_EDIT_MODEL_CHANGE, props.onEditRowModelChange);
+  useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_COMMIT, options.onCellEditCommit);
+  // TODO remove, use onEditRowsModelChange directly
```
oliviertassinari, line 316: If we are going in this direction. Should we update the docs now to not use `onEditRowsModelChange` or do it later?

oliviertassinari (approved):

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: > Should we update the docs now to not use onEditRowsModelChange or do it later?

If you meant `onEditCellPropsChange` I already updated all markdown pages and demos to use `onEditRowsModelChange`.

oliviertassinari: I'm referring to `onEditCellPropsChange` yes. We can see in https://master--material-ui-x.netlify.app/components/data-grid/editing/#server-side-validation that we document the prop. Is it expected?

m4theushw: Hmmm, maybe we can't remove `onEditCellPropsChange` yet. To do server-side validation, the user will want to make the server request only when a particular field is changed. With `onEditRowsModelChange` we loose the field that caused the change.

oliviertassinari: Then it's not clear to me. Why is the value not the only prop we care about for the edit cell?

m4theushw: To be clear. When I say "the props of a field changes" I'm meaning the props in the `editRows` state.

image

Going back to https://github.com/mui-org/material-ui-x/pull/2143#discussion_r675085418. If we use `onEditRowsModelChange` for validation, a server request might be made even if the field was not changed. This method will be shared with the row editing, so multiple fields could be being modified at the same time. Imagine an user registration form. I only want to validate the username when the user changes the username field, not when the name or password is changed. I can only have this granularity with `onEditCellPropsChange`.

oliviertassinari: Ok, I see, the alternative would be to add a new argument to `onEditRowsModelChange`. We have often solve this type of problem like this in the past

PR #2149, opened 2021-07-14
[DataGrid] Add quick filter demo

### LINKED ISSUE #2037: [DataGrid] Example with search bar
I think there is no example of a search bar using in any of the examples under the table and DataGrid filtering tab ,in the documentation. I think there should be one or more examples where the use of the search bar should be shown in a table created with the DataGrid component.

dtassone:
closes #2037 

Preview: https://deploy-preview-2149--material-ui-x.netlify.app/components/data-grid/filtering/#quick-filter

### DIFF
--- a/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
+++ b/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { IconButton, TextField } from '@material-ui/core';
+import IconButton from '@material-ui/core/IconButton';
+import TextField from '@material-ui/core/TextField';
 import {
   DataGrid,
   GridToolbarColumnsButton,
@@ -11,7 +12,7 @@ import { useDemoData } from '@material-ui/x-grid-data-generator';
 import ClearIcon from '@material-ui/icons/Clear';
 import SearchIcon from '@material-ui/icons/Search';
 
-const QuickSearchToolbar = (props) => {
+function QuickSearchToolbar(props) {
   return (
     <GridToolbarContainer
       style={{
@@ -24,13 +25,16 @@ const QuickSearchToolbar = (props) => {
         <GridToolbarDensitySelector />
       </div>
       <TextField
+        variant="standard"
         value={props.value}
         onChange={props.onChange}
-        placeholder="Search..."
+        placeholder="Search…"
         InputProps={{
           startAdornment: <SearchIcon fontSize="small" />,
           endAdornment: (
             <IconButton
+              title="Clear"
+              aria-label="Clear"
               size="small"
               style={{ visibility: props.value ? 'visible' : 'hidden' }}
               onClick={props.clearSearch}
@@ -42,7 +46,7 @@ const QuickSearchToolbar = (props) => {
       />
     </GridToolbarContainer>
   );
-};
+}
 
 export default function QuickFilteringGrid() {
   const { data } = useDemoData({
@@ -57,9 +61,9 @@ export default function QuickFilteringGrid() {
   const requestSearch = (searchValue) => {
     setSearchText(searchValue);
     const searchRegex = new RegExp(`.*${searchValue}.*`, 'ig');
-    const filteredRows = data.rows.filter((o) => {
-      return Object.keys(o).some((k) => {
-        return searchRegex.test(o[k].toString());
+    const filteredRows = data.rows.filter((row) => {
+      return Object.keys(row).some((k) => {
+        return searchRegex.test(row[k].toString());
       });
     });
     setRows(filteredRows);
--- a/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
+++ b/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import PropTypes from 'prop-types';
 import IconButton from '@material-ui/core/IconButton';
 import TextField from '@material-ui/core/TextField';
 import {
@@ -65,6 +66,12 @@ function QuickSearchToolbar(props) {
     </GridToolbarContainer>
   );
 }
+
+QuickSearchToolbar.propTypes = {
+  clearSearch: PropTypes.func.isRequired,
+  onChange: PropTypes.func.isRequired,
+  value: PropTypes.string.isRequired,
+};
 
 export default function QuickFilteringGrid() {
   const { data } = useDemoData({
--- a/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
+++ b/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
@@ -5,8 +5,6 @@ import TextField from '@material-ui/core/TextField';
 import {
   DataGrid,
   escapeRegExp,
-  GridToolbarColumnsButton,
-  GridToolbarContainer,
   GridToolbarDensitySelector,
   GridToolbarFilterButton,
 } from '@material-ui/data-grid';
@@ -20,11 +18,21 @@ const defaultTheme = createMuiTheme();
 const useStyles = makeStyles(
   (theme) => ({
     root: {
-      padding: theme.spacing(1, 1, 0),
+      padding: theme.spacing(0.5, 0.5, 0),
       justifyContent: 'space-between',
+      display: 'flex',
+      alignItems: 'flex-start',
+      flexWrap: 'wrap',
     },
-    searchInputUnderline: {
-      '&:before': {
+    textField: {
+      [theme.breakpoints.down('xs')]: {
+        width: '100%',
+      },
+      margin: theme.spacing(1, 0.5, 1.5),
+      '& .MuiSvgIcon-root': {
+        marginRight: theme.spacing(0.5),
+      },
+      '& .MuiInput-underline:before': {
         borderBottom: `1px solid ${theme.palette.divider}`,
       },
     },
@@ -36,9 +44,8 @@ function QuickSearchToolbar(props) {
   const classes = useStyles();
 
   return (
-    <GridToolbarContainer className={classes.root}>
+    <div className={classes.root}>
       <div>
-        <GridToolbarColumnsButton />
         <GridToolbarFilterButton />
         <GridToolbarDensitySelector />
       </div>
@@ -47,9 +54,9 @@ function QuickSearchToolbar(props) {
         value={props.value}
         onChange={props.onChange}
         placeholder="Search…"
+        className={classes.textField}
         InputProps={{
           startAdornment: <SearchIcon fontSize="small" />,
-          classes: { underline: classes.searchInputUnderline },
           endAdornment: (
             <IconButton
               title="Clear"
@@ -63,7 +70,7 @@ function QuickSearchToolbar(props) {
           ),
         }}
       />
-    </GridToolbarContainer>
+    </div>
   );
 }
 
--- a/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
+++ b/docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
@@ -3,6 +3,7 @@ import IconButton from '@material-ui/core/IconButton';
 import TextField from '@material-ui/core/TextField';
 import {
   DataGrid,
+  escapeRegExp,
   GridToolbarColumnsButton,
   GridToolbarContainer,
   GridToolbarDensitySelector,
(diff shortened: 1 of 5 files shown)


natem122787 (approved):

docs/src/pages/components/data-grid/filtering/filtering.md
```diff
@@ -162,15 +162,17 @@ const filterModel: GridFilterModel = {
 
 {{"demo": "pages/components/data-grid/filtering/FilterApiNoSnap.js", "bg": "inline", "hideToolbar": true}}
 
-## 🚧 Quick filter
+## Quick filter
 
-> ⚠️ This feature isn't implemented yet. It's coming.
+The grid does not natively include quick filtering.
+However, it could be easily implemented as in the demo below.
```
oliviertassinari, line 168: Subjective, what is easy for one can be hard for another

```suggestion
However, it could be implemented as in the demo below.
```

oliviertassinari: It seems that we don't need to use the hypothetical form

```suggestion
However, it can be implemented as in the demo below.
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+  GridToolbarColumnsButton,
+  GridToolbarContainer,
+  GridToolbarDensitySelector,
+  GridToolbarFilterButton,
+} from '@material-ui/data-grid';
+import { useDemoData } from '@material-ui/x-grid-data-generator';
+import ClearIcon from '@material-ui/icons/Clear';
+import SearchIcon from '@material-ui/icons/Search';
+
+const QuickSearchToolbar = (props) => {
```
oliviertassinari, line 14:
```suggestion
function QuickSearchToolbar(props) {
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import { IconButton, TextField } from '@material-ui/core';
```
oliviertassinari, line 2:
```suggestion
import IconButton from '@material-ui/core/IconButton';
import TextField from '@material-ui/core/TextField';
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+    rowLength: 100,
+    maxColumns: 6,
+  });
+  const [searchText, setSearchText] = React.useState('');
+  const [rows, setRows] = React.useState<any[]>(data.rows);
+
+  const requestSearch = (searchValue: string) => {
+    setSearchText(searchValue);
+    const searchRegex = new RegExp(`.*${searchValue}.*`, 'ig');
+    const filteredRows = data.rows.filter((o: any) => {
```
oliviertassinari, line 59: no shorthands

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+    dataSet: 'Commodity',
+    rowLength: 100,
+    maxColumns: 6,
+  });
+  const [searchText, setSearchText] = React.useState('');
+  const [rows, setRows] = React.useState<any[]>(data.rows);
+
+  const requestSearch = (searchValue: string) => {
+    setSearchText(searchValue);
+    const searchRegex = new RegExp(`.*${searchValue}.*`, 'ig');
```
oliviertassinari, line 58: regexp value to escape?

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <GridToolbarDensitySelector />
+      </div>
+      <TextField
+        value={props.value}
+        onChange={props.onChange}
+        placeholder="Search..."
+        InputProps={{
+          startAdornment: <SearchIcon fontSize="small" />,
+          endAdornment: (
+            <IconButton
```
oliviertassinari, line 33: AFAIK we also need to label the button to be accessible

```suggestion
            <IconButton
              title="Clear"
              aria-label="Clear"
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
oliviertassinari: Should we ignore diatrics too in the match?

oliviertassinari: For context, in the main repo, we have used two different search libraries so far: 

- [match-sorter](https://github.com/kentcdodds/match-sorter) in https://next.material-ui.com/components/autocomplete/#advanced
- [flexsearch](https://github.com/nextapps-de/flexsearch) in https://next.material-ui.com/components/material-icons/

I wonder how far we should go. If we don't go very far, how about we encourage a library to go beyond the demo, so our users don't have to think?

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+    >
+      <div>
+        <GridToolbarColumnsButton />
+        <GridToolbarFilterButton />
+        <GridToolbarDensitySelector />
+      </div>
+      <TextField
+        value={props.value}
+        onChange={props.onChange}
+        placeholder="Search..."
```
oliviertassinari, line 29:
```suggestion
        placeholder="Search…"
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,87 @@
...
+        justifyContent: 'space-between',
+      }}
+    >
+      <div>
+        <GridToolbarColumnsButton />
+        <GridToolbarFilterButton />
+        <GridToolbarDensitySelector />
+      </div>
+      <TextField
+        value={props.value}
```
oliviertassinari, line 27: We need to force the variant as it's standard in v4 and outlined in v5 by default.

```suggestion
        variant="standar"
        value={props.value}
```

oliviertassinari: We can involve @danilo-leal to improve the look and feel for the new search input:

Capture d’écran 2021-07-15 à 16 55 06

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
oliviertassinari: I'm actually wondering when we should use ellipses. We have often used them in the docs (main search, icons search) but Google doesn't seem to use them Google Search, YouTube, Google Maps.

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
danilo-leal: Screen Shot 2021-07-16 at 00 22 39

Quick iteration on it. Just added a little bit of x-axis/top padding on the div wrapping the buttons and the search input + also made the input underline a little bit lighter. Those types of interventions already boost the look a bit :)

Could definitely go further if appropriated now.

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
m4theushw: We don't need to ignore diacritics because the strings are all ASCII. I would keep it very simple, using even `row[k].toString().indexOf(...)`. If we keep the Regex approach, then at least let's escape special characters. We could add a note pointing to these search libraries.

```suggestion
    const searchRegex = new RegExp(escapeRegExp(searchValue), 'i');
```

m4theushw (commented):
Looks good. I pushed dc9aa69d with small changes to match @danilo-leal's proposal. The input height was not right because of the `box-sizing: border-box` we apply to the root. What do you think?

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
```diff
@@ -0,0 +1,119 @@
...
+
+function QuickSearchToolbar(props) {
+  const classes = useStyles();
+
+  return (
+    <GridToolbarContainer
+      className={classes.toolbar}
+      style={{
+        justifyContent: 'space-between',
+      }}
```
oliviertassinari, line 45: root convention + no inline style

```suggestion
      className={classes.root}
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
```diff
@@ -0,0 +1,119 @@
...
+import ClearIcon from '@material-ui/icons/Clear';
+import SearchIcon from '@material-ui/icons/Search';
+import { createMuiTheme } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/styles';
+
+const defaultTheme = createMuiTheme();
+const useStyles = makeStyles(
+  (theme) => ({
+    toolbar: {
+      '&&': {
```
oliviertassinari, line 21: This is hiding something
```suggestion
```

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
```diff
@@ -0,0 +1,119 @@
...
+        padding: theme.spacing(1, 1, 0),
+      },
+    },
+    search: {
+      boxSizing: 'unset',
```
oliviertassinari, line 26: better set the box-sizing we need, but even then, it's hiding something

```suggestion
```

dtassone:
@oliviertassinari not sure what the issue is with proptypes again 🤔

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from 'react';
+import IconButton from '@material-ui/core/IconButton';
+import TextField from '@material-ui/core/TextField';
+import {
+  DataGrid,
+  escapeRegExp,
```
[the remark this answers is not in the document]
dtassone: Are we sure about escaping the regexp? We could keep it simple...
If we do we could import an escapeRegExp func or write it in the demo, it could be simpler 🤔

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
oliviertassinari: I have found what it was hiding. The styles of `MuiDataGrid-toolbarContainer` have an extra level of specificity. We don't need to have it. Since it can be composed, I propose we flatten the styles of `GridToolbarContainer` and similar components.

dtassone: I removed it

m4theushw: > The styles of MuiDataGrid-toolbarContainer have an extra level of specificity. We don't need to have it.

The same for `box-sizing: border-box`. We apply it on everything but `CssBaseline` already sets it.

oliviertassinari: @m4theushw CssBaseline is not guaranteed to be present. It's not straightforward to introduce in an existing codebase. For new project, we do encourage its usage.

m4theushw: @oliviertassinari We can keep `box-sizing: border-box` but applying only on those elements that really need it. The ".MuiDataGrid-root *" selector is so broad that it overwrites the new value set by the TextField.

m4theushw (approved):

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.js
oliviertassinari: I have opened #2177, a good one for the community.

oliviertassinari: > We can keep box-sizing: border-box but applying only on those elements that really need it.

It's what I did in the core, I won't object 😁

> The ".MuiDataGrid-root *" selector is so broad that it overwrites the new value set by the TextField.

Right, we can even argue that it can create bugs for developers with custom render cells.

Issue #1234, opened 2021-03-15
Replace @mui/styles with @mui/system


ro7584:
Hi ~

I was wondering if it could add `@material-ui/styles` as dependency for the `@material-ui/grid`?
The file below was included by `@material-ui/data-grid`, but missing in [package.json](https://github.com/mui-org/material-ui-x/blob/v4.0.0-alpha.22/packages/grid/data-grid/package.json).
https://github.com/mui-org/material-ui-x/blob/v4.0.0-alpha.22/packages/grid/_modules_/grid/utils/material-ui-utils.ts#L3

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

oliviertassinari:
@ro7584 Interesting. This import is strictly speaking wrong. 

https://github.com/mui-org/material-ui-x/blob/32fbaf095a0cc4391c248253b3a02e354f8f8ff2/packages/grid/_modules_/grid/utils/material-ui-utils.ts#L3

We are not allowed to import from this package. Only from the core.

As you can see, it's getting resolved with 

https://github.com/mui-org/material-ui-x/blob/32fbaf095a0cc4391c248253b3a02e354f8f8ff2/packages/grid/_modules_/grid/utils/material-ui-utils.ts#L15-L16

Now, the problem is that `@material-ui/styles` is a singleton. If we add it as a dependency to the data grid, it could lead to hard to debug CSS rendering issue. I would recommend we simply snooze the issue until we push #441 further (drop v4 support).

Are you facing a specific issue?

ro7584:
You're right.   
Recently, I'm trying to use [Plug'n'Play](https://yarnpkg.com/features/pnp) and get `@material-ui/style` resolved error at webpack dev server.

Glad to know that it's getting resolved in v5.
Feel free to snooze this issue. 👍

bclehmann:
It might be worth noting that Material 5 is not actually supported. On [NPM ](https://www.npmjs.com/package/@material-ui/data-grid) it says that you can use DataGrid with Material 5 when in fact you cannot.

oliviertassinari:
@bclehmann It's supported, but there are a few limitations, like this one that can be workedaround.

bclehmann:
I'm not sure that it makes sense to call it a limitation if merely importing the module will prevent compilation.

[...]

PR #2151, opened 2021-07-14
Correct readme and package.json Peer Dependencies

bclehmann:
#1234 Notes that x-grid and data-grid don't actually work with Material 5.

This corrects `package.json` and readme files to take this into account.


oliviertassinari:
This is correct

PR #2155, opened 2021-07-16
[DataGrid] Use Intl.Collator for string comparison

### LINKED ISSUE #1968: [DataGrid] Custom localeCompare is not fit for usage in filtering of large arrays

## Summary 💡

We are using a **custom** `String.prototype.localeCompare()` for filtering:

https://github.com/mui-org/material-ui-x/blob/d9e6ad15e0f66aa5728376ad2666716b82c0015f/packages/grid/_modules_/grid/models/colDef/gridStringOperators.ts#L32-L34

However, this API wasn't designed to handle "large" arrays. It's [slow](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#performance). (`.localeCompare(a)` is OK, `.localeCompare(a, b)` is KO).

### Filtering

Open https://material-ui.com/components/data-grid/#commercial-version and filter with equals (localeCompare):

Capture d’écran 2021-06-25 à 17 11 07

Capture d’écran 2021-06-25 à 17 10 40

Do the same contains (without localeCompare):

Capture d’écran 2021-06-25 à 17 11 43

Capture d’écran 2021-06-25 à 17 11 40

See the https://codesandbox.io/s/patient-dew-vg0hw?file=/src/index.tsx for a simpler benchmark.

Capture d’écran 2021-06-26 à 15 45 22

⚠️ I would also challenge that we don't solve the diacritics issue for the other operators. A bug to fix.

## Context 🔦

This was raised in https://github.com/mui-org/material-ui-x/pull/1956#discussion_r658338663.

m4theushw:
Closes #1968 
Preview: https://deploy-preview-2155--material-ui-x.netlify.app/components/data-grid/#commercial-version
Benchmark: https://material-ui.com/components/data-grid/#commercial-version

#### Sorting

localeCompare / Intl.Collator

image

#### Filtering (equals)

localeCompare / Intl.Collator

image

### DIFF
--- a/packages/grid/_modules_/grid/utils/sortingUtils.ts
+++ b/packages/grid/_modules_/grid/utils/sortingUtils.ts
@@ -23,6 +23,8 @@ export const gridNillComparer = (v1: GridCellValue, v2: GridCellValue): number |
   return null;
 };
 
+const collator = new Intl.Collator();
+
 export const gridStringNumberComparer: GridComparatorFn = (
   value1: GridCellValue,
   value2: GridCellValue,
@@ -33,7 +35,7 @@ export const gridStringNumberComparer: GridComparatorFn = (
   }
 
   if (typeof value1 === 'string') {
-    return value1.localeCompare(value2!.toString());
+    return collator.compare(value1!.toString(), value2!.toString());
   }
   return (value1 as any) - (value2 as any);
 };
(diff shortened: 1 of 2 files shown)


packages/grid/_modules_/grid/utils/sortingUtils.ts
```diff
@@ -33,7 +35,7 @@ export const gridStringNumberComparer: GridComparatorFn = (
   }
 
   if (typeof value1 === 'string') {
-    return value1.localeCompare(value2!.toString());
+    return collator.compare(value1!.toString(), value2!.toString());
```
oliviertassinari, line 38: Did you measure a difference in performance? I wouldn't expect one here.

oliviertassinari: Should we update all the other instances in the codebase?

oliviertassinari (approved):

PR #2159, opened 2021-07-16
[DataGrid] Add minHeight to the rendering zone

### LINKED ISSUE #2133: [DataGrid] Header is not scrollable when all items are filtered out

## Current Behavior 😯

It's not possible to scroll a table when all items are filtered out and there are a lot of columns so that horizontal scroll is shown.

## Expected Behavior 🤔

I'd expect the header is scrollable nevertheless. At least you can see what column has filtering applied on.

## Steps to Reproduce 🕹

Steps:

1. Go to this table https://material-ui.com/components/data-grid/filtering/#customize-the-filters
2. Open filtering and select any column which is not visible without scrolling to the right
3. Apply a filter to filter all items out e.g. Country is Andorra
4. Now it's not possible to scroll the table

See the GIF example: 
Jul-12-2021 16-43-34

## Your Environment 🌎

<details>
  
```
 System:
    OS: macOS 11.4
  Binaries:
    Node: 14.15.3 - ~/.asdf/installs/nodejs/14.15.3/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.14.9 - ~/.asdf/installs/nodejs/14.15.3/bin/npm
  Browsers:
    Chrome: 91.0.4472.114
    Edge: Not Found
    Firefox: 87.0
    Safari: 14.1.1
  npmPackages:
    @material-ui/core: 4.11.1 => 4.11.1 
    @material-ui/icons: 4.5.1 => 4.5.1 
    @material-ui/lab: 4.0.0-alpha.56 => 4.0.0-alpha.56 
    @material-ui/pickers: 3.1.2 => 3.1.2 
    @material-ui/styles: 4.11.1 => 4.11.1 
    @material-ui/system:  4.11.3 
    @material-ui/types:  5.1.0 
    @material-ui/utils:  4.11.2 
    @types/react:  17.0.0 
    react: 16.12.0 => 16.12.0 
    react-dom: 16.12.0 => 16.12.0 
    typescript: 4.2.3 => 4.2.3 

  Browser:
    Google Chrome Version 91.0.4472.114
```
</details>

## Order id 💳

#27213

m4theushw:
Fixes #2133 

Preview: https://deploy-preview-2159--material-ui-x.netlify.app/components/data-grid/components/

### DIFF
--- a/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
+++ b/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
@@ -1,10 +1,7 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import { useGridSelector } from '../../hooks/features/core/useGridSelector';
-import {
-  gridDataContainerSizesSelector,
-  gridScrollBarSizeSelector,
-} from '../../hooks/root/gridContainerSizesSelector';
+import { gridDataContainerSizesSelector } from '../../hooks/root/gridContainerSizesSelector';
 import { useGridApiContext } from '../../hooks/root/useGridApiContext';
 
 type GridDataContainerProps = React.HTMLAttributes<HTMLDivElement>;
@@ -13,15 +10,11 @@ export function GridDataContainer(props: GridDataContainerProps) {
   const { className, ...other } = props;
   const apiRef = useGridApiContext();
   const dataContainerSizes = useGridSelector(apiRef!, gridDataContainerSizesSelector);
-  const scrollBar = useGridSelector(apiRef!, gridScrollBarSizeSelector);
 
   const style: any = {
     minWidth: dataContainerSizes?.width,
+    minHeight: dataContainerSizes?.height,
   };
-
-  if (scrollBar.hasScrollY) {
-    style.minHeight = dataContainerSizes?.height!;
-  }
 
   return <div className={clsx('MuiDataGrid-dataContainer', className)} style={style} {...other} />;
 }
--- a/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
+++ b/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
@@ -12,6 +12,7 @@ export function GridDataContainer(props: GridDataContainerProps) {
   const dataContainerSizes = useGridSelector(apiRef!, gridDataContainerSizesSelector);
 
   const style: any = {
+    // TODO remove min
     minWidth: dataContainerSizes?.width,
     minHeight: dataContainerSizes?.height,
   };
--- a/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
+++ b/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
@@ -521,13 +521,20 @@ describe('<DataGrid /> - Layout & Warnings', () => {
 
     describe('autoHeight', () => {
       it('should have the correct intrinsic height', () => {
+        const headerHeight = 40;
+        const rowHeight = 30;
         render(
           <div style={{ width: 300 }}>
-            <DataGrid {...baselineProps} headerHeight={40} rowHeight={30} autoHeight />
+            <DataGrid
+              {...baselineProps}
+              headerHeight={headerHeight}
+              rowHeight={rowHeight}
+              autoHeight
+            />
           </div>,
         );
         expect(document.querySelector('.MuiDataGrid-main')!.clientHeight).to.equal(
-          40 + 30 * baselineProps.rows.length,
+          headerHeight + rowHeight * baselineProps.rows.length,
         );
       });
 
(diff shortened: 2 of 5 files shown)


oliviertassinari (commented):
Did you consider moving the fix outside of the virtualization? I mean, AFAIK, this current fix is coupled to it (`renderingZone`). We would face the same issue if we disable the virtualization.

```diff
diff --git a/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx b/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
index a2a6732e..19718d93 100644
--- a/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
+++ b/packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
@@ -17,6 +17,7 @@ export function GridDataContainer(props: GridDataContainerProps) {

   const style: any = {
     minWidth: dataContainerSizes?.width,
+    minHeight: 1,
   };

   if (scrollBar.hasScrollY) {
```

dtassone (approved):

flaviendelangle:
https://www.argos-ci.com/mui-org/material-ui-x/builds/2415

Their are scrollbar appearing on several snapshots.

oliviertassinari:
@m4theushw The changes surface how the dataContainerSizes.height state is wrong in https://deploy-preview-2159--material-ui-x.netlify.app/components/data-grid/events/#subscribing-to-events:

Capture d’écran 2021-07-21 à 23 08 33

But I don't think that we should take care of it now, it was already wrong before (even if we couldn't see the symptoms).

m4theushw:
> @m4theushw The changes surface how the dataContainerSizes.height state is wrong in https://deploy-preview-2159--material-ui-x.netlify.app/components/data-grid/events/#subscribing-to-events:

@oliviertassinari I fixed this in da6786c. As consequence, it also removed the empty space after the last row when not using virtualization. When virtualization is enabled, the scrollbar size is not added to the height:

https://github.com/mui-org/material-ui-x/blob/9613bdd0611fa43a53561421f7c783f32769c5dd/packages/grid/_modules_/grid/hooks/root/useGridContainerProps.ts#L179-L183

Before: 

image

After:

image

oliviertassinari:
We get a regression: 

Capture d’écran 2021-07-22 à 00 17 46

It looks like we miss a functional test case for it.

m4theushw:
> There are scrollbar appearing on several snapshots.

@flaviendelangle Yes, the bug was the scrollbar not appearing. 😁

oliviertassinari:
https://github.com/mui-org/material-ui-x/pull/2159#issuecomment-884535893

Exactly why functional test with visual test doesn't work 😆

m4theushw:
> We get a regression:

@oliviertassinari Fixed in 033ec03. Now we have a good reason to add the scrollbar size: if `autoHeight` is true. 

From what I can see all other visual test failures are related to https://github.com/mui-org/material-ui-x/pull/2159#issuecomment-884533906. If there's another regression we will discover later.

> Loking at the code, it looks like the abstraction in the state is wrong. The height values aren't meant to be the same. From what I understand, the scroll container should use totalSizes (account for the horizontal scrollbar height if needed), while the scrollable element dataContainerSizes (only account for the space needed to render the children, it's not a minHeight, it's a height ⚠️⚠️⚠️).

Yeah, doesn't make sense to replicate the values. That's something to be solved in another PR.

packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
```diff
@@ -13,15 +10,11 @@ export function GridDataContainer(props: GridDataContainerProps) {
   const { className, ...other } = props;
   const apiRef = useGridApiContext();
   const dataContainerSizes = useGridSelector(apiRef!, gridDataContainerSizesSelector);
-  const scrollBar = useGridSelector(apiRef!, gridScrollBarSizeSelector);
 
   const style: any = {
     minWidth: dataContainerSizes?.width,
+    minHeight: dataContainerSizes?.height,
```
oliviertassinari, line 16: We should be able to force the dimension. From what I understand, using min allows for the value to go beyond what should never happen (hiding root issues).

```suggestion
    // TODO remove min
    minWidth: dataContainerSizes?.width,
    minHeight: dataContainerSizes?.height,
```

oliviertassinari (commented):
Great to have more test cases, they move us one step in the right direction.

packages/grid/_modules_/grid/components/containers/GridDataContainer.tsx
m4theushw: Yes. it was always "min". We can try to set the exact dimension in another PR. If it doesn't work it's because `dataContainerSizes.height` is wrong at some point.

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -519,15 +519,41 @@ describe('<DataGrid /> - Layout & Warnings', () => {
...
-        </div>,
-      );
-      expect(document.querySelector('.MuiDataGrid-main')!.clientHeight).to.equal(
-        40 + 30 * baselineProps.rows.length,
-      );
+    describe('autoHeight', () => {
+      it('should have the correct intrinsic height', () => {
+        render(
+          <div style={{ width: 300 }}>
+            <DataGrid {...baselineProps} headerHeight={40} rowHeight={30} autoHeight />
```
flaviendelangle, line 526: We could extract `headerHeight` into its own variable for readability in the total height formula

flaviendelangle (approved):

Issue #610, opened 2020-11-20
add support for table row onClick or onContextMenu or other keyboard mouse event


vcx100g:
Support mouse keyboard actions for table row!

Render each row will also include onClick or onContextMenu or keyboard context menu or most of the popular mouse keyboard events. Row action are needed in most table (data grid) based app. Without mouse and keyboard action, it is difficult for user on desktop PC.

- [ ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

oliviertassinari:
@vcx100g Do you have an example? Is it related to #517?

vcx100g:
right click on mouse button to open context menu for table row!

![Screenshot_20201122_034100](https://user-images.githubusercontent.com/47863776/99886090-b4700200-2c74-11eb-9f27-72a489a2f2e7.png)

oliviertassinari:
@vcx100g One option is to render a custom cell and combine this demo: https://material-ui.com/components/menus/#context-menu.

vcx100g:
I have try using the renderCell, the "margin/padding" and border between cell do not work with open context menu. Only the context (text or button or image) inside the cell work when right click. Also with selected a row, click context menu button on keyboard cannot trigger anything.

When I develop on tableRow @material-ui, I have to set onContextMenu to whole table row to make it work.

zj9495:
Any update? I think add `onRowContextMenu` and `onCellContextMenu` events is a good choice, developers can render the context menu by themselves.

Issue #2138, opened 2021-07-12
[DataGrid / X-Grid] Add possibility to pass in component for row and cell


Primajin:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

With many material UI components I can overwrite what element should be rendered in the end.
Passing in components allows me to pass in a special checkbox component via https://material-ui.com/api/data-grid/data-grid/#slots but no component for the rows and cells themselves.

## Examples 🌈

```js
<DataGrid
  components={{ RowWrapper: '<a/>', CellWrapper: '<span/>' }}
/>
```

## Motivation 🔦

I would like to have rows / cells rendered as something else than a div and thus provide an element via components
For example to attach to the middle mouse click the row should be an anchor that I can point to the detail page of the data. A mere onClick handler doesn't fire the event.

## Order id 💳
ORDER:25229

dtassone:
For cells, you can use the `renderCell` function on the column object. https://material-ui.com/components/data-grid/columns/#render-cell

Primajin:
@dtassone Hmm but that only renders stuff inside the cell - it does not change the cell element.

In the above example a `<strong>` tag is provided - if you check the code it will render the same old div with a `strong` inside. What I want to achieve is that in this example the div becomes a strong itself and gets all the classes applied.

![image](https://user-images.githubusercontent.com/1742115/125418160-e5e40c8f-cd48-4f1c-85b0-e2353a6ccf67.png)

So basically like you do with Typography - you can provide a `p` but make it look like a `h1`: https://material-ui.com/components/typography/#changing-the-semantic-element

component | elementType |   | The component used for the root node. Either a string to use a HTML element or a component. Overrides the behavior of the variantMapping prop.
-- | -- | -- | --

This would really help us in our app! 👍🏻

m4theushw:
We can add a new `Cell` and `Row` slots to allow to pass custom components. The same for the `componentsProps`.

To listen for click events made by the middle button, there's a workaround involving attaching an `auxclick` event listener to the grid's root. To have it only firing when clicking the row, you just have to check if the `event.target` comes from an element inside a row. Here's a very basic example to give an idea how to do it:

```jsx
export default function Basic() {
  const [gridRoot, setGridRoot] = React.useState(null);

  React.useEffect(() => {
    const handleAuxClick = (event) => {
      if (event.button === 1) {
        console.log("middle click");
      }
[...]

PR #2161, opened 2021-07-17
[DataGrid] Add right click events

zj9495:
ref #610

### DIFF
--- a/docs/pages/api-docs/data-grid/data-grid.md
+++ b/docs/pages/api-docs/data-grid/data-grid.md
@@ -53,7 +53,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">nonce</span> | <span class="prop-type">string</span> |   | Nonce of the inline styles for [Content Security Policy](https://www.w3.org/TR/2016/REC-CSP2-20161215/#script-src-the-nonce-attribute). |
 | <span class="prop-name">onCellClick</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a click event comes from a cell element. |
 | <span class="prop-name">onCellDoubleClick</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a double click event comes from a cell element. |
-| <span class="prop-name">onCellRightClick</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a right click event comes from a cell element. |
+| <span class="prop-name">onCellContextMenu</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a right click event comes from a cell element. |
 | <span class="prop-name">onCellOver</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse over event comes from a cell element. |
 | <span class="prop-name">onCellOut</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse out comes from a cell element. |
 | <span class="prop-name">onCellEnter</span> | <span class="prop-type">(params: GridCellParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse enter event comes from a cell element. |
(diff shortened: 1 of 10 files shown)


oliviertassinari:
This approach seems unscalable. How about we update GridSlotsComponentsProps to support new slots?

zj9495:
> This approach seems unscalable. How about we update GridSlotsComponentsProps to support new slots?

My option is add context menu event to allow users to combine Menu component to impelement a context menu, it is easly and quickly. How about it? 
image

image

flaviendelangle:
I would also be in favor of supporting custom row component (and exporting our component to allow basic wrappers).

oliviertassinari:
> I would also be in favor of supporting custom row component (and exporting our component to allow basic wrappers).

@flaviendelangle To be clean, are you encouraging the following diff?

```diff
diff --git a/packages/grid/_modules_/grid/components/cell/GridCell.tsx b/packages/grid/_modules_/grid/components/cell/GridCellOwner.tsx
similarity index 100%
rename from packages/grid/_modules_/grid/components/cell/GridCell.tsx
rename to packages/grid/_modules_/grid/components/cell/GridCellOwner.tsx
diff --git a/packages/grid/_modules_/grid/models/gridSlotsComponent.ts b/packages/grid/_modules_/grid/models/gridSlotsComponent.ts
index 16d66352..b6226a01 100644
--- a/packages/grid/_modules_/grid/models/gridSlotsComponent.ts
+++ b/packages/grid/_modules_/grid/models/gridSlotsComponent.ts
@@ -30,6 +30,7 @@ import { GridPanel } from '../components/panel/GridPanel';
 import { GridColumnUnsortedIcon } from '../components/columnHeaders/GridColumnUnsortedIcon';
 import { GridApiRefComponentsProperty } from './api/gridComponentsApi';
 import { GridIconSlotsComponent } from './gridIconSlotsComponent';
+import { GridCell } from '@material-ui/data-grid';

 /**
  * Grid components React prop interface containing all the overridable components.
@@ -122,6 +123,7 @@ export const DEFAULT_GRID_SLOTS_COMPONENTS: GridApiRefComponentsProperty = {
   Footer: GridFooter,
   Header: GridHeader,
   PreferencesPanel: GridPreferencesPanel,
+  Cell: 'div',
   LoadingOverlay: GridLoadingOverlay,
   NoResultsOverlay: GridNoResultsOverlay,
   NoRowsOverlay: GridNoRowsOverlay,
```

In practice, it would mean that:

1. the author of this PR can provide his own cell slot, add his own custom event handlers.
2. we would make #2138 a bit easier to solve (to confirm with the author because he might have changed his point of view)
3. we give up on the idea to have `renderCell`/`renderEditCell` to render the **whole** cell, not only the child. https://github.com/bvaughn/react-virtualized/blob/abe0530a512639c042e74009fbf647abdb52d661/source/List/List.example.js#L200-L212

No objections. I even doubt that 3. can scale with the need to have tree data?

Primajin:
I think that could work well, is there an easy way to play with this like in a code sandbox online? Otherwise I can  try to patch my locally installed MUI with the diff above and see if that helps

m4theushw:
@zj9495 Do you still plan to work in this PR towards the suggestion made in https://github.com/mui-org/material-ui-x/pull/2161#issuecomment-884496507? We might even get #2138 fixed for free. I also agree that adding one more callback prop won't scale. We already have a bunch of `onCellXXX` props and some of them are not used internally. The one proposed here would fall in the same situation.

---

@Primajin You can create a [codesandbox](https://ci.codesandbox.io/status/mui-org/material-ui-x/pr/2161/builds/150983) based on the code from this PR but it won't help to play with the proposed diff. It's better to download the repo and apply the diff.

zj9495:
> @zj9495 Do you still plan to work in this PR towards the suggestion made in [#2161 (comment)](https://github.com/mui-org/material-ui-x/pull/2161#issuecomment-884496507)? We might even get #2138 fixed for free. I also agree that adding one more callback prop won't scale. We already have a bunch of `onCellXXX` props and some of them are not used internally. The one proposed here would fall in the same situation.
> 
> @Primajin You can create a [codesandbox](https://ci.codesandbox.io/status/mui-org/material-ui-x/pr/2161/builds/150983) based on the code from this PR but it won't help to play with the proposed diff. It's better to download the repo and apply the diff.

It seems that this PR does not meet the community’s plan, let me close it

PR #2162, opened 2021-07-17
[DataGrid] Fix scrollToIndexes behavior

### LINKED ISSUE #2103: [DataGrid] Column header is moving when keyboard arrow keys are pressed

oliviertassinari:
### Breaking changes

- [DataGrid] Remove public `apiRef.current.isColumnVisibleInWindow()` as it servers private use cases.

---

Fix the bug reported in https://github.com/mui-org/material-ui-x/pull/1964#pullrequestreview-693344458 with the solution proposed. I have spent 1 hour on this, nothing crazy. 

It also appears to fix #2103 and closes #2104.

So two bug fixes, one extra test case, and 50% less code (`isColumnVisibleInWindow` seems not to have use cases for developers, so I have removed it).

Preview: https://deploy-preview-2162--material-ui-x.netlify.app/components/data-grid/#commercial-version

---

If you play with the preview, you will notice one last issue we have with the `scrollToIndexes` logic. See this big jump when I press <kbd>Arrow Left</kbd>?:

https://user-images.githubusercontent.com/3165635/126049490-d3dcc6de-5772-44fd-b431-adce1ba1e1d5.mp4

The `.focus()` call for the new active cell wrecks the correct scroll position that the new logic tries to set. You can also notice it in the other direction when using <kbd>Arrow Up</kbd>, it drives me nuts.

I didn't look into how to solve this. It might be a timing issue, the browser auto-scroll when focusing an element off-screen, so maybe we need to force a layout computation when we change the scroll, no idea. I will open a new GitHub issue, the current UX is not great.


packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -504,6 +504,7 @@ describe('<XGrid /> - Rows', () => {
         );
         const gridWindow = document.querySelector('.MuiDataGrid-window')!;
         gridWindow.scrollTop = offset;
+        gridWindow.dispatchEvent(new Event('scroll')); // Similate browser behavior
```
oliviertassinari, line 507:
```suggestion
        gridWindow.dispatchEvent(new Event('scroll')); // Simulate browser behavior
```

packages/grid/x-grid/src/tests/rows.XGrid.test.tsx
```diff
@@ -529,6 +530,27 @@ describe('<XGrid /> - Rows', () => {
...
+          { field: 'id', width: columnWidth },
+          { field: 'firstName', width: columnWidth },
+          { field: 'lastName', width: columnWidth },
+          { field: 'age', width: columnWidth },
+        ];
+        render(<TestCaseVirtualization width={width + border * 2} rows={rows} columns={columns} />);
+        const gridWindow = document.querySelector('.MuiDataGrid-window')!;
+        expect(gridWindow.scrollLeft).to.equal(0);
+        apiRef.current.scrollToIndexes({ rowIndex: 0, colIndex: 2 });
+        gridWindow.dispatchEvent(new Event('scroll')); // Similate browser behavior
```
oliviertassinari, line 549:
```suggestion
        gridWindow.dispatchEvent(new Event('scroll')); // Simulate browser behavior
```

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
```diff
@@ -193,51 +209,38 @@ export const useGridVirtualRows = (apiRef: GridApiRef): void => {
...
-          logger.debug(`Scrolling to the right, scrollLeft: ${scrollCoordinates.left}`);
-        }
-        if (gridState.rendering.renderingZoneScroll.left > scrollCoordinates.left) {
-          scrollCoordinates.left = columnsMeta.positions[params.colIndex];
-          logger.debug(`Scrolling to the left, scrollLeft: ${scrollCoordinates.left}`);
-        }
+
+      if (params.colIndex != null) {
+        scrollCoordinates.left = scrollIntoView({
+          clientHeight: windowRef.current!.clientWidth,
```
m4theushw, line 215: I noticed a difference of 1px between `clientWidth` and the [old values](https://github.com/mui-org/material-ui-x/pull/2162/files#diff-7904e8c986f79f8bbc1627613039ba9fc46cc3f000aef6b4883c56f4aff2311aL210-L211). Should we calculate the inner width by ourselves?

```suggestion
          clientHeight: gridState.containerSizes!.windowSizes.width - gridState.scrollBar.scrollBarSize.y,
```

In Chrome, it misses the right border. In Firefox it works fine. This doesn't happen in https://material-ui.com/components/data-grid/demo/

DVOMOyM6Lk

m4theushw (commented):
IMO, we shouldn't adopt this new approach for `rowIndex`. The virtualization doesn't respect the `scrollTop` which might create new problems. Once #1911 is done we could use it.

dtassone (approved):

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
oliviertassinari: I have tried to reproduce on BS with Windows 10 and Chrome 91 but can't:

Capture d’écran 2021-07-20 à 11 45 17

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -92,7 +92,7 @@ export const useStyles = makeStyles(
         '& .MuiDataGrid-columnHeader:focus-within, & .MuiDataGrid-cell:focus-within': {
           outline: `solid ${muiStyleAlpha(theme.palette.primary.main, 0.5)} 1px`,
           outlineWidth: 1,
-          outlineOffset: -2,
+          outlineOffset: -1,
```
oliviertassinari, line 95: It's off topic, but seems simple enough, it's meant so the outline fits perfectly.

before: 
Capture d’écran 2021-07-20 à 11 48 53

after:
Capture d’écran 2021-07-20 à 11 50 17

oliviertassinari:
> IMO, we shouldn't adopt this new approach for rowIndex. The virtualization doesn't respect the scrollTop which might create new problems. Once #1911 is done we could use it.

@m4theushw To be clear, I didn't change the approach used by `rowIndex`. This change is about updating `colIndex` to use the approach we use in HEAD for `rowIndex`. It's also about creating an abstraction so both axes use the same logic (`scrollIntoView`). Now that we give layout information to `scrollIntoView`, in the future, we can add more options, to it, similar to https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView what accepts.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
oliviertassinari: > Should we calculate the inner width by ourselves?

I think that it would be better to reduce the number of dependencies here. The more we can rely on direct access to the DOM API, the better. It would be great to get to the root cause of what you have found. Is it a sub-pixel alignment issue? Is it a CSS outline rendering glitch? Did you zoom?

m4theushw: > I have tried to reproduce on BS with Windows 10 and Chrome 91 but can't:

I noticed it only happens if the Windows's scaling is above 100%. Here's 125% because the resolution is 1920x1080. To simulate it, open Chrome in BrowserStack and zoom to 125%. Use https://deploy-preview-2162--material-ui-x.netlify.app/components/data-grid/demo/ to test. The `clientWidth` should be 830 but it's 829.

oliviertassinari: I can reproduce with a zoom of 125% on Windows and macOS. I can also reproduce on AG Grid.

Capture d’écran 2021-07-20 à 15 14 08

I guess Chrome is rounding numbers when we read the DOM but differently based on what we read.

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts
m4theushw: Yes, let's see if some user complains about. With the current logic we use today, this glitch doesn't occur. That was my only concern.

oliviertassinari: In any case, 💯 for having found this. Even if I don't that we should try hard to fix it at the state of the component (high opportunity cost). 

I think that later on, once we need to find new ways to further differentiate in this mature market, it could matter.

PR #2166, opened 2021-07-18
[core] Support `docs:api` script in Windows OS

ZeeshanTamboli:
I was running this script on Windows OS and I got `/' was unexpected at this time` error.
Apparently, Windows platforms do not support single quotes on the command line, so escaping it in double quotes.

This works in all OS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "start": "yarn docs:dev",
     "docs:dev": "yarn workspace docs dev",
     "docs:api": "yarn docs:api:build",
-    "docs:api:build": "cross-env BABEL_ENV=development babel-node --config-file ./docs/babel.config.js -i '/node_modules/(?!@material-ui)/' -x .ts,.tsx,.js ./docs/scripts/buildApi.ts ./docs/pages/api-docs/data-grid",
+    "docs:api:build": "cross-env BABEL_ENV=development babel-node --config-file ./docs/babel.config.js -i \"/node_modules/(?!@material-ui)/\" -x .ts,.tsx,.js ./docs/scripts/buildApi.ts ./docs/pages/api-docs/data-grid",
     "docs:build": "yarn workspace docs build",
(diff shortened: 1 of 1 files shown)


oliviertassinari (commented):
~What about `yarn prettier`? Does it work too on Windows?~

package.json
```diff
@@ -5,7 +5,7 @@
     "start": "yarn docs:dev",
     "docs:dev": "yarn workspace docs dev",
     "docs:api": "yarn docs:api:build",
-    "docs:api:build": "cross-env BABEL_ENV=development babel-node --config-file ./docs/babel.config.js -i '/node_modules/(?!@material-ui)/' -x .ts,.tsx,.js ./docs/scripts/buildApi.ts ./docs/pages/api-docs/data-grid",
+    "docs:api:build": "cross-env BABEL_ENV=development babel-node --config-file ./docs/babel.config.js -i \"/node_modules/(?!@material-ui)/\" -x .ts,.tsx,.js ./docs/scripts/buildApi.ts ./docs/pages/api-docs/data-grid",
```
oliviertassinari, line 8: This seems to work too.

```suggestion
    "docs:api:build": "cross-env BABEL_ENV=development babel-node -x .ts,.tsx,.js ./docs/scripts/buildApi.ts ./docs/pages/api-docs/data-grid",
```

ZeeshanTamboli: Ok this works for me as well, but better if @m4theushw can confirm with this suggested change.

m4theushw: Yes, it works. The reason to not ignore node_modules was because, at the time the script was made, one of the imports was using ES6. Now this file uses the CJS syntax.

https://github.com/mui-org/material-ui/blob/f60126f95d7b36fa12133887990615b9f47e994a/docs/src/modules/utils/parseMarkdown.js#L5

m4theushw (approved):

Issue #2146, opened 2021-07-14
[DataGrid] Fix docs, add missing `onCellFocusOut`


missbruni:
On release https://github.com/mui-org/material-ui-x/releases/tag/v4.0.0-alpha.32 it seems onEditCellChangeCommitted is now not returning the event, on the release itself there seems to be no explanation of how to go around in case you do want to stop the default component behaviour for this event. 

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Without being able to stop event propagation we are no longer able to stop the cell from going into view mode once commited with clicking out on a edit more custom row implementation. 

Using alpha 33
https://codesandbox.io/s/material-demo-forked-dqx3v

## Expected Behavior 🤔

Using alpha 31
https://codesandbox.io/s/material-demo-forked-fzthu

## Steps to Reproduce 🕹

- Go on demo links for expected or current scenarios
- Click on Edit for a row
- Edit a cell and click out (commit) to the next one

You will see on expected that the cell remains on edit mode and on current it doesnt. This is due to onEditCellChangeCommitted not having access to the event any longer. 

## Context 🔦

We are trying to remain on edit mode upon committing a cell change on clicking outside as we have a custom row implementation with a master save. If this is an intentional change, it would be appreciated if we could be given a workaround for this behaviour.

m4theushw:
Clicking outside a cell does two things: switch back to view mode and commits the value. Stopping the propagation in `onEditCellChangeCommitted` only prevents the commiting. Thus, the row will not be updated, but the cell will still go to the view mode. To prevent it from reverting to view mode, cancel the default behavior of the outside click with the `onCellFocusOut`:

```jsx
<DataGrid
  onCellFocusOut={(params, event) => {
    event.defaultMuiPrevented = true;
  }}
/>
```

There's an example using this handler to also achieve this master save implementation: https://material-ui.com/components/data-grid/editing/#edit-using-external-button

---

@oliviertassinari The example in the docs I mentioned is labeled as a Pro feature but it's just a custom implementation using the props. What is the original intention of having it labeled like it? Is the ability to stop the propagation of events?

dtassone:
@m4theushw the external button demo uses apiRef which is why it is marked as a pro demo. Stopping the propagation of events works in MIT as well

missbruni:
[...]

PR #2167, opened 2021-07-18
[docs] Add missing API docs

ZeeshanTamboli:
Part of #2146 to add doc of `onCellFocusOut` prop.

Also added some other props from `GridOptions` which were missing.

**Preview:**
XGrid: https://deploy-preview-2167--material-ui-x.netlify.app/api/data-grid/x-grid/
DataGrid: https://deploy-preview-2167--material-ui-x.netlify.app/api/data-grid/data-grid/


packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -143,8 +143,8 @@ export interface GridOptions {
   editRowsModel?: GridEditRowsModel;
   /**
    * Filtering can be processed on the server or client-side.
-   * Set it to 'client' if you would like to handle filtering on the client-side.
```
oliviertassinari, line 146: This still seem relevant, why remove it?

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -90,6 +97,7 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onRowsScrollEnd</span> | <span class="prop-type">(param: GridRowScrollEndParams) => void</span> |   | Callback fired when scrolling to the bottom of the grid viewport. |
 | <span class="prop-name">onSelectionModelChange</span> | <span class="prop-type">(param: GridSelectionModelChangeParams) => void</span> |   | Callback fired when the selection state of one or multiple rows changes. |
 | <span class="prop-name">onSortModelChange</span> | <span class="prop-type">(param: GridSortModelParams) => void</span> |   | Callback fired when the sort model changes before a column is sorted. |
+| <span class="prop-name">onStateChange</span> | <span class="prop-type">(params: any) => void</span> |   | Callback fired when the state of the grid is updated. |
```
oliviertassinari, line 100: This is private API, to not document.

docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -84,6 +89,7 @@ import { DataGrid } from '@material-ui/data-grid';
 | <span class="prop-name">onRowLeave</span> | <span class="prop-type">(param: GridRowParams, event: React.MouseEvent) => void</span> |   | Callback fired when a mouse leave event comes from a row container element. |
 | <span class="prop-name">onSelectionModelChange</span> | <span class="prop-type">(param: GridSelectionModelChangeParams) => void</span> |   | Callback fired when the selection state of one or multiple rows changes. |
 | <span class="prop-name">onSortModelChange</span> | <span class="prop-type">(param: GridSortModelParams) => void</span> |   | Callback fired when the sort model changes before a column is sorted. |
+| <span class="prop-name">onStateChange</span> | <span class="prop-type">(params: any) => void</span> |   | Callback fired when the state of the grid is updated. |
```
oliviertassinari, line 92: Private API

oliviertassinari (commented):
Thanks for the fixes. It's easy to have this docs outdated, the sooner we upgrade the docs infra to v5, so we can use the docs generator of the main repository, the better.

packages/grid/_modules_/grid/models/gridOptions.tsx
ZeeshanTamboli: The default is `client`, so I added `@default`. So I removed it  from here as I thought we need not explicitly set it to `client` because it is default.

docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -44,15 +46,18 @@ import { DataGrid } from '@material-ui/data-grid';
...
 | <span class="prop-name">isRowSelectable</span> | <span class="prop-type">(params: GridRowParams) => boolean</span> |   | Determines if a row can be selected. |
 | <span class="prop-name">loading</span> | <span class="prop-type">boolean</span> |  false | If `true`, a  loading overlay is displayed. |
 | <span class="prop-name">localeText</span> | <span class="prop-type">GridLocaleText</span> |   | Set of text labels used in the grid. You can find all the translation keys supported in [the source](https://github.com/mui-org/material-ui-x/blob/HEAD/packages/grid/_modules_/grid/constants/localeTextConstants.ts) in the GitHub repository. |
 | <span class="prop-name">logger</span> | <span class="prop-type">Logger</span> | null | Pass a custom logger in the components that implements the 'Logger' interface. |
-| <span class="prop-name">logLevel</span> | <span class="prop-type">string | false</span> | false | Allows to pass the logging level or false to turn off logging. |
+| <span class="prop-name">logLevel</span> | <span class="prop-type">string or boolean false</span> | false | Allows to pass the logging level or false to turn off logging. |
```
m4theushw, line 54: I would keep the older syntax. We keep the same TS definition in the docs.

image

```suggestion
| <span class="prop-name">logLevel</span> | <span class="prop-type">string | false</span> | false | Allows to pass the logging level or false to turn off logging. |
```

docs/pages/api-docs/data-grid/data-grid.md
m4theushw: Well, if it's private we should consider removing this prop. It's no used internally.

packages/grid/_modules_/grid/models/gridOptions.tsx
m4theushw: In the core, these comments are used in boolean props. There, the default value goes to `@default` and the comment only mentions what the non-default value does.

image

I don't have any preference here, as long as we keep consistency: remove from all other feature mode props or keep this line.

docs/pages/api-docs/data-grid/data-grid.md
```diff
@@ -27,13 +27,15 @@ import { DataGrid } from '@material-ui/data-grid';
...
 | <span class="prop-name">disableDensitySelector</span> | <span class="prop-type">boolean</span> | false | If `true`, the density selector is disabled. |
 | <span class="prop-name">disableExtendRowFullWidth</span> | <span class="prop-type">boolean</span> | false | If `true`, rows will not be extended to fill the full width of the grid container. |
 | <span class="prop-name">disableSelectionOnClick</span> | <span class="prop-type">boolean</span> | false | If `true`, the selection on click on a row or cell is disabled. |
 | <span class="prop-name">error</span> | <span class="prop-type">any</span> |   | An error that will turn the grid into its error state and display the error component. |
 | <span class="prop-name">editRowsModel</span> | <span class="prop-type">GridEditRowsModel</span> | undefined  | Set the edit rows model of the grid. |
+| <span class="prop-name">filterMode</span> | <span class="prop-type">GridFeatureMode</span> | client | Filtering can be processed on the server or client-side. Set it to `server` if you would like to handle filtering on the server-side. |
```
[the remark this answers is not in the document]
ZeeshanTamboli: Done

docs/pages/api-docs/data-grid/data-grid.md
ZeeshanTamboli: I did this because in https://material-ui.com/api/data-grid/data-grid/ it is getting displayed as:

screenshot 1

The pipe ( | ) was being considered as a separator. I escaped it in markdown now.

docs/pages/api-docs/data-grid/data-grid.md
ZeeshanTamboli: I see it being used here https://github.com/mui-org/material-ui-x/blob/e6bed6b36d362810e49af90e4c6271d3c41ba93c/packages/grid/_modules_/grid/hooks/features/core/useGridApi.ts#L38
Am I missing something?

m4theushw: I was talking about the prop. We don't use it internally. XGrid's users could still get state updates by subscribing to `GRID_STATE_CHANGE`.

oliviertassinari: @m4theushw We have mentioned it in some of the issues. I think that we should consider it experimental at this point. It's unclear what problem it solves.

packages/grid/_modules_/grid/models/gridOptions.tsx
oliviertassinari: In this case, we can move slower: only add the default mention, it's already an improvement.

ZeeshanTamboli: I already have added the `@default`  tag.

docs/pages/api-docs/data-grid/data-grid.md
ZeeshanTamboli: So based on this conversation there are 2 things:
1. Remove the doc and the prop both.
2. Only remove the doc.

What do we do?

ZeeshanTamboli: In any case, I removed this API from docs.

ZeeshanTamboli:
I have also updated the `onSelectionModelChange` API doc in commit https://github.com/mui-org/material-ui-x/pull/2167/commits/63432bd5473d84839a2ec1593eb2f4d4e9390613

ZeeshanTamboli:
I am a bit confused here that the https://material-ui.com/api/data-grid/data-grid/ is still showing `(param: GridSortModelParams) => void` for `onSortModelChange`.
Capture 2

While on master HEAD it has been updated: https://github.com/mui-org/material-ui-x/blob/master/docs/pages/api-docs/data-grid/data-grid.md.

Did we release the docs?

m4theushw (approved):
Good work. 👍

m4theushw:
> Did we release the docs?

We had a problem during the deploy. Now it's updated.

PR #2173, opened 2021-07-20
[DataGrid] Remove unused `editMode` prop

ZeeshanTamboli:
While adding missing API docs I noticed this prop was not in use.


m4theushw:
👍 for removing it. I'm repurposing this prop in #2098 to control between cell editing and row editing.

ZeeshanTamboli:
> 👍 for removing it. I'm repurposing this prop in #2098 to control between cell editing and row editing.

Oh I see. Should I close this? You just need to change the JSDoc description. Describe it as row or cell instead of client, sever.

m4theushw:
> Oh I see. Should I close this? You just need to change the JSDoc description. Describe it as row or cell instead of client, sever.

No, we can keep this one. The row editing feature will take some time to be ready.

flaviendelangle (approved):

ZeeshanTamboli:
Can somebody look into why the CI is failing? I re-tried it multiple times.

flaviendelangle:
@ZeeshanTamboli it's also the case on other PR's, we do have a problem.
But it's not linked to your PR :+1:

m4theushw:
@oliviertassinari Again Netlify is failing because it didn't find a dependency we don't use. What did you do in #2140 to make it rerun?

m4theushw (approved):

oliviertassinari:
> What did you do in #2140 to make it rerun?

I reverted changes done to the yarn.lock. Which is a crappy solution, hiding the problem. Since then, we changed the file in https://github.com/mui-org/material-ui-x/pull/2164. So I guess it's why it fail again. I could find https://answers.netlify.com/t/builds-always-fail-until-clearing-cache-and-re-trying/11612/13.

m4theushw:
@oliviertassinari It's weird because I opened #2171 a couple of hours before this one and Netlify didn't fail.

Does https://answers.netlify.com/t/build-cache-is-now-updated-in-deploy-preview-builds/26344 make any sense? If this production branch's cache is wrong all PRs based on it will fail.

image

oliviertassinari:
@m4theushw This PR is based on top of the commit that I have selected:

Capture d’écran 2021-07-20 à 23 46 42

It's pretty new.

oliviertassinari:
@m4theushw I see that you pushed this branch to production, we will see if it solves the issue.

m4theushw:
@oliviertassinari I clicked this "Re-run" button:

image

Only thing I noticed was this new deploy:

image

Issue #924, opened 2021-01-25
[RFC] Better isolate DataGridPro from DataGrid


oliviertassinari:
## Summary 💡

The two components should be better isolated. `XGrid` should be able to grow in complexity and package size without harming the bundle size of `DataGrid`. In the current configuration, both packages have the exact same size

## Motivation 🔦

- `DataGrid` is in competition in the space of a similar open-source data grid. The bundle size is one of the dimensions developers look at. I think that the biggest decision influence happens when one package looks disproportionally bigger than an equivalent solution. So this issue might be the one that has the most influence on this matter. I'm not sure that we will need to the same depth as we did in the main repository.

  If you look at the bundle size of ag-Grid, you can see that it can quickly get out of hand: [ag-grid-community](https://bundlephobia.com/result?p=ag-grid-community): 193 kB gzipped. They have tried to isolate features, but the core seems to be about the same size: [@ag-grid-community/core](https://bundlephobia.com/result?p=@ag-grid-community/core): 191 kB gzipped (it's only with the enterprise features that it seems to make a difference).

- The license should be better isolated. It's not OK to have all the content of `/packages/grid/_modules_` exported under an MIT licensed package.
- This effort should allow us to make the grid, and its feature more composable. It would be great to invest time in a plugin system. Some prior-arts: https://datatables.net/manual/plug-ins/.

oliviertassinari:
@dtassone Has made a quick-win proposal to initiate this direction. We could create two `GridComponent`

https://github.com/mui-org/material-ui-x/blob/a0b719281f82460d0e7f829af40add2538c0e46e/packages/grid/_modules_/grid/GridComponent.tsx#L48

This won't solve the problem but put us in a better place. It will force us to start handling the matter as we build new features, in order to identify ways to reduce the duplication.

oliviertassinari:
@mbrookes How important do you think this is? Would it make sense to do it post v5.0.0 stable?

I would personally advocate doing it before the Premium features, regardless of the stable nature. For the Pro features, IMHO, the main value is about forcing us to isolate features (tree shaking, code isolation, etc.)

mbrookes:
It's important, but it can wait if there are other priorities.

oliviertassinari:
Regarding sharing code, we might want to have `@material-ui/x-grid` depending on `@material-ui/data-grid`. It would have the following advantages:

1. Progressive migration, people adding XGrid in a few cases don't pay the bundle size overhead twice
[...]

PR #2176, opened 2021-07-20
[core] Isolate DataGrid and XGrid

dtassone:
TODO 
- [x] refactor how to inject 'useThemeProps' as it's breaking immunity
- [x] refactor selection model to accept 'GridRowId' as it's breaking immunity

Part of #924


oliviertassinari:
@mui-org/x Who wants to own and continue these changes?

flaviendelangle:
I can take a look :+1: 
But do we agree that this is a first draft that won't remove the XGrid code contained in the hooks from the DataGrid bundle ?

oliviertassinari:
> do we agree that this is a first draft that won't remove the XGrid code contained in the hooks from the DataGrid bundle?

Agree, it's meant to be a step that allows us to build new hooks for XGrid only.

flaviendelangle:
@mui-org/x I'm removing the properties from `DataGrid` that could only have one value (`pagination` for instance) to clarify the TypeScript interface.
Do you see a reason not to ?

oliviertassinari:
> In a lot of our examples we are putting <DataGrid pagination {....otherProps /> but the pagination prop is always true in DataGrid so it's misleading.

@flaviendelangle We do such so that developers can replace `<DataGrid>` with `<XGrid>` and get the same behavior. It's for progressive adoption.

flaviendelangle:
@oliviertassinari ok
I'm not a huge fan of this approach. 

What is the downside of enabling the pagination by default on `XGrid` ?

oliviertassinari:
> What is the downside of enabling the pagination by default on XGrid ?

@flaviendelangle I can think of one: most XGrid applications would be without the pagination

Does this downside have more weight than the downsides of` <DataGrid pagination>`? I don't know, I'm not emotionally invested in any of the two.

flaviendelangle:
@oliviertassinari I don't think it's a very important subject today indeed.
Can I remove all the other props useless in `DataGrid` and only keep `pagination` ?

packages/grid/_modules_/grid/hooks/features/export/useGridCsvExport.tsx
```diff
@@ -38,7 +38,7 @@ export const useGridCsvExport = (apiRef: GridApiRef): void => {
       return buildCSV({
         columns: exportedColumns,
         rows: visibleSortedRows,
-        selectedRowIds: selection,
+        selectedRowIds: Array.isArray(selection) ? selection : [selection],
```
m4theushw, line 41: We could change below to use `gridArraySelectionStateSelector` and remove this if-is-an-array check.

https://github.com/mui-org/material-ui-x/blob/f79a1fd0c32ddc9a7d2e1cdf87e81d369ce6827e/packages/grid/_modules_/grid/hooks/features/export/useGridCsvExport.tsx#L20

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -186,7 +189,7 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
   React.useEffect(() => {
     // Rows changed
     setGridState((state) => {
-      const newSelectionState = [...state.selection];
+      const newSelectionState = getArraySelectionModel(state.selection);
```
m4theushw, line 192: Reuse `gridArraySelectionStateSelector`

```suggestion
      const newSelectionState = gridArraySelectionStateSelector(state);
```

Same for the `isRowSelectable`'s effect.

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -1,273 +1,69 @@
...
-  disableMultipleColumnsFiltering?: true;
-  disableMultipleColumnsSorting?: true;
-  disableMultipleSelection?: true;
-  onRowsScrollEnd?: undefined;
-  selectionModel?: GridRowId | GridRowId[];
-  pagination?: true;
-};
-
-const MAX_PAGE_SIZE = 100;
+export const DATA_GRID_DEFAULT_PROPS = DEFAULT_GRID_OPTIONS;
```
m4theushw, line 29: Is there a reason to export `DATA_GRID_DEFAULT_PROPS` and `X_GRID_DEFAULT_PROPS`?

packages/grid/data-grid/src/tests/pagination.DataGrid.test.tsx
```diff
@@ -384,12 +383,12 @@ describe('<DataGrid /> - Pagination', () => {
 
       const footerHeight = document.querySelector('.MuiDataGrid-footerContainer')!.clientHeight;
       const expectedViewportRowsLengthBefore = Math.floor(
-        (heightBefore - DEFAULT_GRID_OPTIONS.headerHeight - footerHeight) /
-          DEFAULT_GRID_OPTIONS.rowHeight,
+        (heightBefore - DATA_GRID_DEFAULT_PROPS.headerHeight - footerHeight) /
```
m4theushw, line 386: We have been setting manually the header and row height for the test. This file is the only one where the default value is used. I would use the props to set the value.

https://github.com/mui-org/material-ui-x/blob/f79a1fd0c32ddc9a7d2e1cdf87e81d369ce6827e/packages/grid/x-grid/src/tests/rows.XGrid.test.tsx#L317-L322

packages/grid/data-grid/src/DataGrid.tsx
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2176#discussion_r677743672

To use them here
But if we stop using them in test I can stop exporting them

flaviendelangle:
@mui-org/x "refactor how to inject 'useThemeProps' as it's breaking immunity"
I'm struggling with this point.
Damien removed it from the `DataGrid` code but I don't see why.

m4theushw:
> @mui-org/x "refactor how to inject 'useThemeProps' as it's breaking immunity"
> I'm struggling with this point.
> Damien removed it from the `DataGrid` code but I don't see why.

I don't know why but by removing it we lost the ability to change the language [via the theme](https://material-ui.com/components/data-grid/localization/#locale-text). ⚠

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -1,273 +1,70 @@
...
+          <GridHeaderPlaceholder />
+          <GridBody />
+          <GridFooterPlaceholder />
+        </GridErrorHandler>
+      </GridRoot>
+    </GridContextProvider>
   );
 });
 
+DataGridRaw.defaultProps = DEFAULT_GRID_OPTIONS;
```
m4theushw, line 65: If we re-add the `useThemeProps`, by setting the default props we won't get the theme translations yet. This default options object has a `localeText` prop with the English translation and the [`getThemeProps`](https://github.com/mui-org/material-ui/blob/bb3b9adbd42d8d0aa5f50b5da60f7ee377118a48/packages/material-ui-styles/src/getThemeProps/getThemeProps.js#L15-L17)  used internally doesn't overwrite the value if there's a default one.

flaviendelangle: Good idea to test it correctly :+1:

flaviendelangle: Fixed

packages/grid/x-grid/src/XGrid.tsx
```diff
@@ -17,15 +28,29 @@ if (process.env.NODE_ENV !== 'production' && RELEASE_INFO === '__RELEASE' + '_IN
...
+          <GridHeaderPlaceholder />
+          <GridBody />
+          <GridFooterPlaceholder />
+        </GridErrorHandler>
+      </GridRoot>
+    </GridContextProvider>
+  );
 });
 
+XGridRaw.defaultProps = DEFAULT_GRID_PROPS_FROM_OPTIONS;
```
oliviertassinari, line 55: The `defaultProps` API is in its way of deprecation in React, please never use it.

```suggestion
```

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -235,7 +290,7 @@ describe('<DataGrid /> - Selection', () => {
       const row0 = getRow(0);
       expect(row0).to.have.class('Mui-selected');
 
-      setProps({ selectionModel: 1 });
+      setProps({ selectionModel: [1] });
```
oliviertassinari, line 293: For any refactoring, changes in tests should receive high scrutiny. Do we need these changes in this PR? Why? Could we isolate them?

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -1,274 +1,43 @@
...
-            '',
-            'You need to upgrade to the XGrid component to unlock multiple selection.',
-          ].join('\n'),
-        );
-      }
-      return null;
-    },
-  ),
-} as any;
+DataGrid.propTypes = DATA_GRID_PROPTYPES;
```
oliviertassinari, line 43: AFAIK, we can't isolate the prop types in a different file. It prevents https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types to remove them in production. Also, in v5 they will be automatically generated anyway.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -142,7 +146,7 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
 
   const setSelectionModel = React.useCallback(
     (model: GridSelectionModel) => {
-      apiRef.current.selectRows(model, true, true);
+      apiRef.current.selectRows(getArraySelectionModel(model), true, true);
```
oliviertassinari, line 149: Why do we need this change here?

packages/grid/data-grid/src/tests/selection.DataGrid.test.tsx
flaviendelangle: I think it was to have the test pass because Damien did not have the time to handle properly the non-array  version.
But it's now implemented so I can revert this change.

packages/grid/x-grid/src/XGrid.tsx
flaviendelangle: We are currently using the default props.
I do agree that it would be better not to.
Could we do it in a distinct PR to avoid making this one heavier ?

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: I can edit `selectRows` to handle `GridRowId | GridRowId[]` otherwise

m4theushw (approved):

packages/grid/x-grid/src/XGrid.tsx
oliviertassinari: Oh yeah, I didn't realize that we were already using defaultProps. My comment is off topic.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2176/files#diff-36939e58755c91e3693a4f18aa7796b137ba1e32f65357a940fb358286048088L61

On `master` we are doing this.
Which is wrong according to the typing because `selectionModel` is always an array in the interface.

https://github.com/mui-org/material-ui-x/blob/fa346f0fbe3d9b9eea9bb403fe4675f544d6abf9/packages/grid/_modules_/grid/models/gridSelectionModel.ts

I'm fixing this by supporting the non-array format.

Then I can either preprocess the prop like it was done in the DataGrid but in both components.
Or handle the non-array version across the codebase.
Which has the advantage of sticking to what the user is giving me. But that's debatable.

flaviendelangle: The test are not different
I just added one for the array version in DataGrid.

flaviendelangle: I cleaned it to only convert the non-array to array once from the props and once from the imperative update.

flaviendelangle: @oliviertassinari  I just saw that the `setSelectionModel` does not handle the non-array version
Do we want to add it for consistency with the props or do we want to only handle the array version here ?

If so, I can remove the line you pointed. 

The issue came from the fact that the hook method it not typed by importing the public API
So we have

```ts
  const setSelectionModel = React.useCallback(
    (model: GridSelectionModel) => {
```

But

```ts
  /**
   * Updates the selected rows to be those passed to the `rowIds` argument.
   * Any row already selected will be unselected.
   * @param {GridRowId[]} rowIds The row ids to select.
   */
  setSelectionModel: (rowIds: GridRowId[]) => void;
```

Which made me make that mistake.

____

I removed the non-array version from `setSelectionModel` since it was not documented and it's not the subject of the PR to improve it :+1:

oliviertassinari (commented):
looks good

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -1,276 +1,43 @@
...
+    </GridContextProvider>
   );
 });
 
+DataGridRaw.defaultProps = DEFAULT_GRID_PROPS_FROM_OPTIONS;
+
 export const DataGrid = React.memo(DataGridRaw);
 
 // @ts-ignore
-DataGrid.propTypes = {
```
oliviertassinari, line 83: I didn't notice this change. I think that we should revert the `propTypes` change. The `propTypes` needs to be in the same file for https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types to trigger. Actually, we also need to replace the TypeScrip bundler to use the Babel bundler of the core that leverages this optimization. Once we move the infra to v5. the prop-types will be automatically generated and in the same file.

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2176/files/5ca14f2e2feaa13e2ce70bc67130fdbb85d2eb79#r680403715

https://github.com/mui-org/material-ui-x/blob/master/packages/grid/data-grid/src/DataGrid.tsx#L44

I don't know why their is an outdated code visible here :thinking: 
However I forgot to delete the file containing the split prop types, I'll do a PR and merge it as it is trivial

flaviendelangle: #2432

PR #2179, opened 2021-07-21
[DataGrid] Use event.defaultMuiPrevented to prevent the default behavior

### LINKED ISSUE #1403: [RFC] Use event.defaultMuiPrevented to prevent default events

m4theushw:
### Breaking changes

- [DataGrid] Replace `event.stopPropagation()` with `event.defaultMuiPrevented = true` in all callback props. 

  Previously, calling `event.stopPropagation` would stop the propagation of an event and also prevent its the default behavior from running. Now, these two actions can be controlled independently:

  - Set `event.defaultMuiPrevented` to `true` to prevent the default action of an event

  - Call `event.stopPropagation` to stop an event from propagating to parent elements

  ```diff
  <DataGrid
    onCellDoubleClick={(params, event) => {
  -   event.stopPropagation();
  +   event.defaultMuiPrevented = true;
    }}
  />
  ```

- [XGrid] Replace `event.stopPropagation()` with `event.defaultMuiPrevented = true` in all event listeners. 

  ```diff
  apiRef.current.subscribeEvent('cellDoubleClick', (params, event) => {
  - event.stopPropagation();
  + event.defaultMuiPrevented = true;
  });
  ```

  For more information, check this [page](https://material-ui.com/components/data-grid/events/#disabling-the-default-behavior).

---

Closes #1403 

Preview: https://deploy-preview-2179--material-ui-x.netlify.app/components/data-grid/events/#disabling-the-default-behavior

TODO

- [x] Add section in https://material-ui.com/components/data-grid/events/
- [x] Remove all `event.isPropagationStopped()`
- [x] Improve the definition of `MuiEvent`
- [x] List the breaking changes


docs/src/pages/components/data-grid/events/events.md
```diff
@@ -20,14 +20,33 @@ You can subscribe to one of the [events emitted](/components/data-grid/events/#c
...
+
+```tsx
+<DataGrid
+  onCellClick={(params: GridCellParams, event: MuiEvent) => {
+    event.defaultMuiPrevented = true;
+  }}
+/>
+```
+
+Usually, double clicking a cell will put it into [edit mode](/components/data-grid/editing).
```
oliviertassinari, line 46:
```suggestion
Usually, double clicking a cell will put it into [edit mode](/components/data-grid/editing/).
```

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -206,139 +206,166 @@ export interface GridOptions {
   /**
    * Callback fired when the edit cell value changes.
    * @param {GridEditCellPropsParams} params With all properties from [[GridEditCellPropsParams]].
-   * @param {React.SyntheticEvent} event The event that caused this prop to be called.
+   * @param {MuiEvent} event The event that caused this prop to be called.
```
oliviertassinari, line 209:
```suggestion
   * @param {MuiEvent<React.SyntheticEvent>} event The event that caused this prop to be called.
```

? x the other

---

I would personally be fine if I had to do this in my codebase.:

```jsx
(event as any).defaultMuiPrevented = true;
```

```suggestion
   * @param {React.SyntheticEvent} event The event that caused this prop to be called.
```

but I can understand if we don't like this idea 😆

packages/storybook/src/stories/grid-rows.stories.tsx
```diff
@@ -482,11 +481,11 @@ export function EditRowsControl() {
   }, []);
 
   const onCellEditCommit = React.useCallback(
-    (params: GridEditCellPropsParams, event?: React.SyntheticEvent) => {
+    (params: GridEditCellPropsParams, event) => {
```
oliviertassinari, line 484: Why not doing? 
```suggestion
    (params: GridEditCellPropsParams, event?: MuiEvent<React.SyntheticEvent>) => {
```
instead of using as later down

docs/src/pages/components/data-grid/events/events.md
```diff
@@ -20,14 +20,33 @@ You can subscribe to one of the [events emitted](/components/data-grid/events/#c
...
+
+## Disabling the default behavior
+
+Depending on the use case, it might be necessary to disable the default action taken by an event.
+The `MuiEvent` passed to the event handler has a `defaultMuiPrevented` property to control when the default behavior can be executed or not.
+Set it to `true` to block the default handling of an event and implement your own.
+
+```tsx
+<DataGrid
+  onCellClick={(params: GridCellParams, event: MuiEvent) => {
```
oliviertassinari, line 40: ?
```suggestion
  onCellClick={(params: GridCellParams, event: MuiEvent<React.SyntheticEvent>) => {
```

docs/src/pages/components/data-grid/events/events.md
```diff
@@ -8,7 +8,7 @@ title: Data Grid - Events
 
 ## Subscribing to events
```
m4theushw, line 9: This page is not good. My idea is to remove the Pro badge from the page and move it to this section only. Then, add a new section only for the event props, e.g. `onCellClick`, explaining that they are like subscribing to events but with less granularity. For the catalog of events, I want to improve the docs generator to include a third column with the event prop that is mapped to each event name. I think that with our convention GRID_CELL_CLICK -> onCellClick I can discover the mappings. cc @oliviertassinari

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -87,7 +87,7 @@ Server-side validation works like client-side [validation](#validation).
 - Validate the value in the server.
 - Once the server responds, set the `error` attribute to false if it is valid. This allows to commit it.
 
-**Note:** To prevent the default client-side behavior, use `event.stopPropagation()`.
+**Note:** To prevent the default client-side behavior, set `event.defaultMuiPrevented` to `true`.
```
dtassone, line 90: if the user call `stopPropagation()`, should it also stop the propagation in MUI? 
If so we could override the function to something like
```ts
event.stopPropagation = ()=> {
  defaultMuiPrevented = true;
base.stopPropagation();
}
```
That would also keep the feature backward compatible.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -245,7 +235,7 @@ export function useGridEditRows(
   const handleCellKeyDown = React.useCallback(
     (params: GridCellParams, event) => {
       const { id, field, cellMode, isEditable } = params;
-      if (!isEditable || event.isPropagationStopped()) {
+      if (!isEditable) {
```
dtassone, line 238: why not `event.defaultMuiPrevented` here and further below?

packages/grid/_modules_/grid/hooks/root/useApi.ts
```diff
@@ -4,15 +4,26 @@ import { GridSubscribeEventOptions } from '../../utils/eventEmitter/GridEventEmi
...
   const publishEvent = React.useCallback(
-    (name: string, params: any, event?: React.SyntheticEvent) => {
-      if (!event || !event.isPropagationStopped || !event.isPropagationStopped()) {
-        apiRef.current.emit(name, params, event);
+    (
+      name: string,
+      params: any,
+      event: MuiEvent<React.SyntheticEvent | DocumentEventMap[keyof DocumentEventMap] | {}> = {},
+    ) => {
+      event.defaultMuiPrevented = false;
```
dtassone, line 22: if we set `event.defaultMuiPrevented = true;` and then publish then it will be reset here and be published.
Why not check `defaultMuiPrevented` everywhere there `isPropagationStopped`

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
```diff
@@ -12,7 +13,15 @@ export function useGridApiEventHandler(
 
   React.useEffect(() => {
     if (handler && eventName) {
-      return apiRef.current.subscribeEvent(eventName, handler, options);
+      const enhancedHandler = (
```
dtassone, line 16: that could go out of this scope in a separate function. Not sure about redefining the handler as it changes the ref of the handler func potentially breaking immutability `if (handler === otherHandler)`  🤔

docs/src/pages/components/data-grid/editing/editing.md
m4theushw: I think it's better to keep them independent, like calling `stopPropagation()` and `preventDefault()` in a JS event. This will be a real breaking change but we have to do it.

packages/grid/_modules_/grid/models/gridOptions.tsx
m4theushw: We started to use this `MuiEvent`. Would be nice to keep it, because of the special property. Sometimes using it can lead to terrible typings when the event can be fired both by a synthetic event as well as by a native event, e.g. `MuiEvent<React.MouseEvent | DocumentEventMap['click']>`.

packages/storybook/src/stories/grid-rows.stories.tsx
m4theushw: Fixed.

packages/grid/_modules_/grid/hooks/root/useApi.ts
m4theushw: We still want the event to be published. Only the listeners should not run if `event.defaultMuiPrevented` is `true`.

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
m4theushw: We don't have any place where we compare the handler that is actually passed to `subscribeEvent`. I mean, compare the handler with the one that is saved in the event emitter. The ref would not change.

Another approach would be to move the `event.defaultMuiPrevented` validation to the event emitter, which is a very low level way to do it.

PR #2181, opened 2021-07-21
[DataGrid] Won't be merged - Do not loose the resized columns width when running the flex allocation

### LINKED ISSUE #2129: [XGrid] Resized column state is lost when resizing column or viewport and at least on column has flex enabled

## Current Behavior 😯

https://user-images.githubusercontent.com/3165635/125062727-01430480-e0af-11eb-871f-131328f4e3f6.mp4

https://codesandbox.io/s/material-demo-forked-ch5hy?file=/demo.tsx

## Expected Behavior 🤔

https://user-images.githubusercontent.com/3165635/125063170-70b8f400-e0af-11eb-857f-7482d4f28a6f.mp4

https://plnkr.co/edit/VO2y32CCitTxbgar

## Steps to Reproduce 🕹

See above actual behavior

## Context 🔦

Found when working on #1850

## Your Environment 🌎

@material-ui/x-grid@v4.0.0-alpha.33

flaviendelangle:
Fixes #2129 
Preview: https://deploy-preview-2181--material-ui-x.netlify.app/components/data-grid/columns/#resizing
Bug reproduction: https://codesandbox.io/s/material-demo-forked-gubpt?file=/demo.js

@mui-org/x , when doing `apiRef.current.setColumnWidth(myField, myWidth)`, do you think it should rerun the flex process ?
For instance if I have :

```
columns = [{ field: 'id', flex: 1 }, { field: 'brand', width: 100 }]
width = 500

=>

columnsAfterFlex = [{ field: 'id', width: 400 }, { field: 'brand', width: 100 }]
```

If I do `apiRef.current.setColumnWidth('brand', 150)`, do you intend the `id` column to have a width of 350px or 400px ?

In the issue we have the case where we resize through the UI and we clearly don't want to rerun the flex process because it destroys the resize we just did (and others we may have done)
But for the API call I suppose both can make sense.

____

As for the implementation in this PR, I added an argument to `updateColumns` / `updateColumn` that is available in the public API. I'm not sure creating 2 new methods make sense here. The column update methods could take a `behavior?: { shouldApplyFlex?: boolean }` for scalability if we have other variants of the column updates.

____

Another point, currently when mounting a Grid, we run 3 times `updateColumnsWidth` 

1) In the useEffect that tracks the props
2) In the useEffect that tracks the viewport width on 1st render
3) In the useEffect that tracks the viewport width after the viewport is set

With this PR we are down to 2.
I could remove the first of the two. But it means that the columns are not initialized before the viewport is mounted in the DOM. Which makes sense I think. But we run quite a lot of code before which need the columns (for instance we run `applySorting`).
So it would be a deeper change.


docs/pages/api-docs/data-grid/grid-api.md
```diff
@@ -88,8 +88,8 @@ import { GridApi } from '@material-ui/x-grid';
...
-| <span class="prop-name">updateColumns</span>                                                      | <span class="prop-type">(cols: GridColDef[]) =&gt; void</span>                                                                                                                           | Updates the definition of multiple columns at the same time.                                                                                                   |
+| <span class="prop-name">updateColumn</span>                                                       | <span class="prop-type">(col: GridColDef, shouldApplyFlex?: boolean) =&gt; void</span>                                                                                                   | Updates the definition of a column.                                                                                                                            |
```
dtassone, line 91: It's not great to add this argument here

flaviendelangle: It's still a draft :+1: 
I was redacting the description of the PR.
I have cases with `shouldApplyFlex = true` and `shouldApplyFlex = false` in `apiRef.current.updateColumn(s)` in other hooks so I can't just rely on a default value.

I can create new methods in `GridAPI` but for me it make sense when the user update the columns through the API to let him decide if he want to reapply the flex.

oliviertassinari:
> If I do apiRef.current.setColumnWidth('brand', 150), do you intend the id column to have a width of 350px or 400px ?

@flaviendelangle I think that it depends on what's the current state of the column width. The state of the props is the initial starting point. 

So in your case:

```
columns = [{ field: 'id', flex: 1 }, { field: 'brand', width: 100 }]
width = 500

=>

columnsAfterFlex = [{ field: 'id', width: 400 }, { field: 'brand', width: 100 }]
```

I don't think that it makes sense to store the computed width as the input width, it should stay a derived state, not a source of truth. This, IMHO, would make a lot more sense:

```diff
-columnsAfterFlex = [{ field: 'id', flex: 1, width: 400 }, { field: 'brand', width: 100 }]
+columnsAfterFlex = [{ field: 'id', flex: 1, computedWidth: 400 }, { field: 'brand', width: 100, computedWidth: 100 }]
```

Then, if we call `apiRef.current.setColumnWidth('brand', 150)`

we get this state change:

```diff
-columnsAfterFlex = [{ field: 'id', flex: 1, computedWidth: 400 }, { field: 'brand', width: 100, computedWidth: 100 }]
+columnsAfterFlex = [{ field: 'id', flex: 1, computedWidth: 350 }, { field: 'brand', width: 150, computedWidth: 150 }]
```

flaviendelangle:
@oliviertassinari I'm totally in favor of differentiating clearly the input properties of the `column` object and the computed ones.
We have a lot of `column.width!` which shows that we often need to loose TypeScript typechecking because we now better than him what is in this property depending on where we call it (I was talking about it here https://github.com/mui-org/material-ui-x/pull/2101#discussion_r667764931)

docs/pages/api-docs/data-grid/grid-api.md
oliviertassinari: I would personally expect to have `shouldApplyFlex` always true

flaviendelangle: @oliviertassinari that means that for manual resize through the header, you could cause a shrink on the flex column down to 50px. This is a change of behavior that would encourage the users to put a width or minWidth on their flex column. Whereas today increasing the width of a column by resizing it will never cause the other columns to shrink.

But I suppose it make sense.

flaviendelangle: @oliviertassinari 

In the doc we have the following line :

> flex doesn't work together with width. If you set both flex and width in GridColDef, flex will override width.

And indeed our current flex implementation does not care about the `width` prop
How do we want to make it evolve ?

Currently my PRs behavior is to never allow a flex column to have less computedWidth than its with

For instance :

```js
columns = [{ field: 'id', flex: 1, width: 100 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 200 }, { field: 'brand', width: 100 }]
```

but 

```js
columns = [{ field: 'id', flex: 1, width: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 300 }, { field: 'brand', width: 100 }]
```

Whereas today we would have 

```js
columns = [{ field: 'id', flex: 1, width: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 200 }, { field: 'brand', width: 100 }]
```

But if we want to keep ignoring the `width` property when `flex` is defined. Then the user should do the following in order to avoid having its flex column squashed to much.

```js
columns = [{ field: 'id', flex: 1, minWidth: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 300 }, { field: 'brand', width: 100 }]
```

The two can make sens. I find the 1st one a bit more straightforward.

flaviendelangle:
@oliviertassinari 

In the doc we have the following line :

> flex doesn't work together with width. If you set both flex and width in GridColDef, flex will override width.

And indeed our current flex implementation does not care about the `width` prop
How do we want to make it evolve ?

Currently my PRs behavior is to never allow a flex column to have less computedWidth than its with

For instance :

```js
columns = [{ field: 'id', flex: 1, width: 100 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 200 }, { field: 'brand', width: 100 }]
```

but 

```js
columns = [{ field: 'id', flex: 1, width: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 300 }, { field: 'brand', width: 100 }]
```

Whereas today we would have 

```js
columns = [{ field: 'id', flex: 1, width: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 200 }, { field: 'brand', width: 100 }]
```

But if we want to keep ignoring the `width` property when `flex` is defined. Then the user should do the following in order to avoid having its flex column squashed to much.

```js
columns = [{ field: 'id', flex: 1, minWidth: 300 }, { field: 'brand', width: 100 }]
viewportWidth = 300

=>

columnsAfterFlex = [{ field: 'id', computedWidth: 300 }, { field: 'brand', width: 100 }]
```

The two can make sens. I find the 1st one a bit more straightforward.

oliviertassinari:
@flaviendelangle Great point. Flex should have more priority: https://codepen.io/oliviertassinari/pen/jOmGZag, this is how CSS works. In this case, I would vote for a new property: `forcedWidth` to allow end-users to override the flex logic when using the column resize feature.

packages/grid/x-grid/src/tests/columns.XGrid.test.tsx
```diff
@@ -164,5 +164,94 @@ describe('<XGrid /> - Columns', () => {
...
+
+      it('should resize the flex width after resizing a column with api', () => {
+        const twoColumns = [
+          { field: 'id', width: 100, flex: 1 },
+          { field: 'brand', width: 100 },
+        ];
+
+        render(<Test columns={twoColumns} />);
+
+        expect(getColumnHeaderCell(0).style.width).to.equal('198px');
```
oliviertassinari, line 184: We have utils for this, the main advantage is that it throw if the component can't be found and render the element details when the assertion fails. I propose we use it here, x the other cases
```suggestion
        expect(getColumnHeaderCell(0)).toHaveInlineStyle({ width: '198px' });
```

flaviendelangle:
Oh indeed my implementation goes against the CSS logic.

If we follow the `flex` CSS logic, then from what I understand, the `flex` column `computedWidth`:

- could be smaller than its `width` property (= the `width` property is not used when their is a `flex` as told in the docs)
- could not be smaller than its `minWidth` property
- could not be larger than its `maxWidth` property

And we could add a `forcedWidth` in the future but its the same than putting `minWidth = maxWidth` right ?

oliviertassinari:
> And we could add a forcedWidth in the future

I think that we should do it in this PR, because if we don't, we solve #2129 but regress on the resizing feature. If we have correct test cases, we would even break the CI.

> its the same than putting minWidth = maxWidth right ?

width, minWidth, maxWidth, and flex, are about reproducing the CSS layout algorithm. `forcedWidth` would be about allowing end-users to override all of this. Now, `minWidth = maxWidth` could work, we could start like this

flaviendelangle:
I'm not sure to understand when you would want to use the `forcedWidth`, could you give me an example ?

flaviendelangle:
I updated the behavior to match the CSS flex as described here https://github.com/mui-org/material-ui-x/pull/2181#issuecomment-884797776

packages/grid/_modules_/grid/hooks/features/columns/columnsUtils.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // Compute the width of flex columns
+  if (totalFlexUnits && widthToAllocateInFlex > 0) {
+    const widthPerFlexUnit = totalFlexUnits > 0 ? widthToAllocateInFlex / totalFlexUnits : 0;
+
+    for (let i = 0; i < stateColumns.length; i += 1) {
+      const column = stateColumns[i];
+
+      if (!column.hide && column.flex && column.flex > 0) {
+        const flexColumnWidth = Math.floor(widthPerFlexUnit * column.flex);
```
dtassone, line 61: as for #2110, it creates issues to use the `Math.floor` here

dtassone: https://github.com/mui-org/material-ui-x/pull/2110/files#diff-ab3374e32d059e6ec4d957bf3002146d73129b88296f556ae19061e1c15512c6R63

DanailH:
I'm not sure if it is appropriate at this stage to discuss the functionality but wouldn't it be better if when resizing the `flex` is just removed from the column (or even from all of the columns). My argument is that if you are going to resize you are looking to control the space in the grid your self so keeping the `flex` is probably not worth it?

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -7,149 +7,46 @@ import {
...
-      if (!column.flex) {
-        return {
-          ...column,
-          width: Math.max(column.minWidth!, column.width!),
-        };
-      }
-
-      let newWidth = column.minWidth! ?? GRID_STRING_COL_DEF.width;
-      if (viewportWidth > 0) {
-        const flexColumnWidth = Math.floor(flexMultiplier * column.flex!);
```
oliviertassinari, line 68: We forgot this logic, e.g. `min-width: 246.923px;`, if it was made on purpose, I think that we should break the changes down. One problem at a time.

oliviertassinari (commented):
The refactoring and extra test cases look good. However, I can't see any change in behavior. How about we aim to merge this as an internal refactor and then come back to our problem (#2129)? 

> wouldn't it be better if when resizing the flex is just removed from the column

Removing the flex prop from the column definition sounds better than updating min and max width to match.

flaviendelangle:
https://github.com/mui-org/material-ui-x/pull/2181#issuecomment-885628268

@DanailH you should remove it from all the columns
Which when resizing from the separator in the header would causes jumps. I don't think it would work well.

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
flaviendelangle: There is behavior changes
If you

1) go to this demo : http://localhost:3001/components/data-grid/columns/#resizing
2) set a column flex to 1
3) resize this column
4) resize it again

Between 3 and 4 the column you resize will take bake its original width (that's my video https://github.com/mui-org/material-ui-x/issues/2129#issuecomment-878280780)
But the tests are not testing this particular scenario.

___

OK to split this PR

@oliviertassinari for 4. you see another solution than mine if we want to keep round numbers ?

https://github.com/mui-org/material-ui-x/pull/2104/commits/7b26f5ee3086088ba568a047c13ed5e7ccd22b3c#diff-ab3374e32d059e6ec4d957bf3002146d73129b88296f556ae19061e1c15512c6R70

___

I don't understand what you mean by removing the flex when resizing.
Do you mean just ignore the property if a column has been resized since the last prop column update ?
It would cause jumps when resizing with the selector in the header no ?

oliviertassinari: @flaviendelangle Ok, all good.

1. No strong point of view
2. Yes, it seems fine
3. Yes, but we can always wait to see if it's enough. So it seems that we fix the problem in #2011 here. I guess we could add test cases in another PR and then close #2011.

oliviertassinari (commented):
I looks good

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
flaviendelangle: OK, I let this PR on hold and I create the one with the test

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2211

See this one for the tests

DanailH (approved):

flaviendelangle:
@DanailH I won't merge this one, I splitting the modifications into several smaller chunks

For the current one : https://github.com/mui-org/material-ui-x/pull/2235

oliviertassinari:
@flaviendelangle Come back to https://github.com/mui-org/material-ui-x/pull/2181#discussion_r676202491

> 1. Increase test coverage to feel safer during our refactoring
> 2. Fix the DB storing by introducing computedWidth
> 3. Fix state update to be more efficient
> 4. Fix flex width to use the whole space with a round number [DataGrid] Gap after the last column when all the column widths are set to fluid #2011
> 5. Fix [XGrid] Resized column state is lost when resizing column or viewport and at least on column has flex enabled #2129

It seems that we have done 1. 2. 4. What do you think about skipping 3. and doing 5. directly?

flaviendelangle:
@oliviertassinari the n°3 is in https://github.com/mui-org/material-ui-x/pull/2286
But it is independent from 5. so I agree that we can do 5 directly.

PR #2188, opened 2021-07-22
[typescript] Improve type coverage of colDef

### LINKED ISSUE #2187: [typescript] Avoid `api: any` or `colRef: any` due to circular import

## Summary 💡

Currently, when importing the `GridApi` or `GridColDef` object would cause a circular import, we replace it with an `any`
For instance

```ts
/**
 * The column of the current header component.
 */
colDef: any;
/**
 * API ref that let you manipulate the grid.
 */
api: any;
```

We can use dynamic import to avoid circular dependency and type correctly those properties

```ts
/**
 * The column of the current header component.
 */
colDef: import('../colDef/gridColDef').GridColDef;
/**
 * API ref that let you manipulate the grid.
 */
api: import('../api/gridApiRef').GridApiRef;
```

## Motivation 🔦

Typing correctly will help us catching bugs and inconsistencies.

For `GridColumnHeaderParams` we have calls that gives the `apiRef` and other that gives `apiRef.current`

```ts
const getColumnHeaderParams = React.useCallback(
  (field: string): GridColumnHeaderParams => ({
    field,
    colDef: apiRef.current.getColumn(field),
    api: apiRef!.current,
  }),
  [apiRef],
);
```

```ts
export type GridColumnHeaderClassFn = (params: GridColumnHeaderParams) => string;

const headerClassName = isFunction(column.headerClassName)
  ? column.headerClassName({ field: column.field, colDef: column, api: apiRef })
  : column.headerClassName;
```

It's the only one I found with inconsistencies inside a single type.
I'm opening a second issue to propose to unify the behaviors across the various calls so that the only breaking change on this one is `GridColumnHeaderParams`

Furthermore, having the autocomplete is a big help for large objects like `api` or `colRef`

flaviendelangle:
Fixes #2187 

Other public methods that are not typed

- `onStateChange` is not typed (but may be removed https://github.com/mui-org/material-ui-x/pull/2167)


flaviendelangle:
@m4theushw I removed the `event` and `props` on some of the custom cell renderers because the typing says that they should not be here.
I would like your confirmation on this point :+1:

flaviendelangle:
Follow up on https://github.com/mui-org/material-ui-x/issues/2189#issuecomment-884938508
If we want to stop exposing `apiRef`, I suppose that we don't want to type it on the public methods while it's still their.
The `columns` / `colDef` typing is still relevant I think.

docs/src/pages/components/data-grid/columns/RenderExpandCellGrid.js
```diff
@@ -120,7 +120,161 @@ renderCellExpand.propTypes = {
   /**
    * The column of the row that the current cell belongs to.
    */
-  colDef: PropTypes.any.isRequired,
+  colDef: PropTypes.shape({
```
oliviertassinari, line 123: this looks way too verbose for a demo

flaviendelangle: This line has been generated, but I guess it's an error in the TS => PropTypes generator.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -28,11 +30,11 @@ export interface GridCellParams {
   /**
    * The column of the row that the current cell belongs to.
    */
-  colDef: any;
+  colDef: GridColDef;
   /**
    * GridApi that let you manipulate the grid.
    */
-  api: any;
+  api: GridApi;
```
dtassone, line 37: we could add `undefined` as for datagrid it won't be there

packages/storybook/src/stories/grid-sorting.stories.tsx
```diff
@@ -256,8 +256,8 @@ export const CustomComparator = () => {
         params.getValue(params.id, 'age') || 'x'
       }`,
     sortComparator: (v1, v2, cellParams1, cellParams2) =>
-      cellParams1.api.getCellValue(cellParams1.id, 'age') -
-      cellParams2.api.getCellValue(cellParams2.id, 'age'),
+      (cellParams1.api.getCellValue(cellParams1.id, 'age') as number) -
```
dtassone, line 259: Why has this one change?

packages/grid/_modules_/grid/models/params/gridCellParams.ts
flaviendelangle: Do we have a plan to export different interfaces for `XGrid` and `DataGrid` ?
I can set `api: GridApi | undefined` and let the user of `XGrid` do `api!.getState()` for instance
But if we have a way to generate a different interface it would be nice

packages/storybook/src/stories/grid-sorting.stories.tsx
flaviendelangle: Before, `api` was of type `any` so `cellParams1.api.getCellValue(cellParams1.id, 'age')` and `cellParams1.api.getCellValue(cellParams2.id, 'age')` was of type `any`

Now they are of type `GridCellValue`
So TypeScript complains that I am trying to do a substraction on a type that potentially does not support it (`object` for instance`)

I would like to add a generic to `getCellValue` (https://github.com/mui-org/material-ui-x/issues/2190) to do `cellParams1.api.getCellValue<number>(cellParams1.id, 'age')` instead of casting but it requires changes in the doc gen.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
dtassone: I agree that it would be nice to generate a different interface but I think at this stage it would add more complexity so I suggest we keep the idea for the next iterations.
I'm working on the split bw the 2 components in #2176, and as mentioned in https://github.com/mui-org/material-ui-x/issues/2189#issuecomment-885546633, I think we should approach the setting of params in a generic way

flaviendelangle: OK to let `api: GridApi` and do another PR with https://github.com/mui-org/material-ui-x/issues/2189#issuecomment-885546633 and `api: GridApi | undefined`

Or do you want `api: GridApi | undefied` even though in reality it is currently never undefined ?

oliviertassinari:
How did we get away from the circular dependency in the types?

flaviendelangle:
@oliviertassinari the type-only import ( `import type { MyType }` ) handles circular imports perfectly from what I read.
I'm not sure if, when importing a type with `import { MyType }` there is a real circular import issue or if it's just a limitation of the ESLint rule.

https://github.com/microsoft/TypeScript/pull/35200
https://github.com/import-js/eslint-plugin-import/issues/1453

oliviertassinari:
@flaviendelangle Ok, I didn't know that TypeScript support circular type declaration. It sounds interesting to give it a try and see if it flies at scale.

flaviendelangle:
I tried to build `yarn release:build` and the `.d.ts` files have been generated correctly.

https://github.com/mui-org/material-ui-x/pull/2213/files#diff-b98630eb301ddfb1b291caade7edafda784e434b852f21efc232324774d4857eR37
Here (other PR) I just fixed the circular import rather than adding `type` to dozens of imports. 
But as long as our imports are only types (and if classes / enum, only used as types), this solution should work.

packages/grid/x-grid-data-generator/src/renderer/renderEditContractType.tsx
```diff
@@ -32,9 +32,9 @@ function EditContractType(props: GridCellParams) {
 
   const handleChange = (event) => {
     const editProps = { value: event.target.value };
-    api.setEditCellProps({ id, field, props: editProps }, event);
+    api.setEditCellProps({ id, field, props: editProps });
```
m4theushw, line 35: In #2220 I changed this component to call the new `setEditCellValue`, which is the correct method and this one does requires the event. We can remove the event now but it will come back later.

docs/pages/api-docs/data-grid/grid-cell-params.md
```diff
@@ -14,9 +14,9 @@ import { GridCellParams } from '@material-ui/data-grid';
 
 | Name                                                                                         | Type                                                                              | Description                                                      |
 | :------------------------------------------------------------------------------------------- | :-------------------------------------------------------------------------------- | :--------------------------------------------------------------- |
-| <span class="prop-name">api</span>                                                           | <span class="prop-type">any</span>                                                | GridApi that let you manipulate the grid.                        |
+| <span class="prop-name">api</span>                                                           | <span class="prop-type">GridApi&lt;&gt;</span>                                    | GridApi that let you manipulate the grid.                        |
```
m4theushw, line 17: `GridApi<>` doesn't seem right. We can open another PR to fix the line below to also check if the array is not empty.

https://github.com/mui-org/material-ui-x/blob/9f336333ef1e537c9ac56b81defa3634f294642a/docs/scripts/buildApi.ts#L81

packages/grid/_modules_/grid/models/params/gridColumnResizeParams.ts
```diff
@@ -9,11 +12,11 @@ export interface GridColumnResizeParams {
   /**
    * The column of the current header component.
    */
-  colDef: any;
+  colDef: GridStateColDef;
   /**
    * API ref that let you manipulate the grid.
    */
-  api: any;
+  api: GridApiRef;
```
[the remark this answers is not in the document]
flaviendelangle: What are you suggesting ?
We do not type the API argument in callback because once #2236 is merged we will be able to remove them ?

packages/grid/_modules_/grid/models/params/gridCellParams.ts
DanailH: I think that one is no longer relevant since it will be removed in https://github.com/mui-org/material-ui-x/issues/1821?

packages/grid/_modules_/grid/models/params/gridColumnResizeParams.ts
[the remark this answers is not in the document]
DanailH: Same as https://github.com/mui-org/material-ui-x/pull/2188#discussion_r682599852. I don't think it's worth changing it even.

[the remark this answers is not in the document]
m4theushw: My suggestion was only for consistency. Some params object include the value, others the ref.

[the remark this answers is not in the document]
flaviendelangle: Yes @m4theushw but I did not want to apply major breaking changes in this PR, just do the typing part without modifying the behavior.

PR #2192, opened 2021-07-22
run netlify

oliviertassinari:
Test that Netlify uses the cache of the production env.


m4theushw:
They seem to use the cache of the last **published** deploy in production.

We already had one deploy in production that was meant to fix the Netlify issue but other deploys after it kept failing: 🤷‍♂️

image

oliviertassinari:
Perfect, it's what I wanted to confirm. We are good.

oliviertassinari:
To be clear, my assumption is that Netlify only update its cache if the build is successful. it was only since 3 hours ago.

PR #2194, opened 2021-07-22
[docs] Describe how to export custom rendered cells

### LINKED ISSUE #2185: [docs] Explain that valueFormatter can be used in the CSV export
### How to Export to CSV if column value has custom format 

Hello! Probably this is wrong place for my issue. Please refer me right place for it.

I use XGrid. For some columns in my table I have custom value: object: e.g {a:200,b:400}.
I use renderCell for render these custom values - Drawing simple chart.

When I export data as csv I got for these columns next value, example:

`Name,Description,[object Object]`

This is reasonable because csvSerializer just not process this case:

```
const serialiseCellValue = (value: any, delimiterCharacter: GridExportCsvDelimiter) => {
  if (typeof value === 'string') {
    const formattedValue = value.replace(/"/g, '""');
    return formattedValue.includes(delimiterCharacter) ? `"${formattedValue}"` : formattedValue;
  }
  return value;
};
```

Is there way to setup CSV export for such case. E.g provide some exportFormatter function?

Thanks!

m4theushw:
Closes #2185 

Preview: https://deploy-preview-2194--material-ui-x.netlify.app/components/data-grid/export/#export-custom-rendered-cells


dtassone (approved):

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -194,6 +194,8 @@ const columns: GridColDef[] = [
 
 {{"demo": "pages/components/data-grid/columns/RenderCellGrid.js", "bg": "inline"}}
 
+**Note**: It is recommended to also set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/) the data.
```
DanailH, line 197: It would be better to point directly to the section.

```suggestion
**Note**: It is recommended to also set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/#export-custom-rendered-cells) the data.
```

mbrookes (approved):

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -8,12 +8,14 @@ title: Data Grid - Export
...
 ## CSV export
 
-You are able to export the displayed data to CSV with an API call, or using the grid UI.
-
-To enable the CSV export you need to compose a toolbar containing the `GridToolbarExport` component, and apply it using the `Toolbar` key in the grid `components` prop.
+The DataGrid allows the data to be exported to CSV by composing a [toolbar](/components/data-grid/components/#toolbar) with the `GridToolbarExport` component. Use the `components` prop to assign the custom toolbar.
 
 {{"demo": "pages/components/data-grid/export/ExportSelectorGrid.js", "bg": "inline"}}
 
+Alternatively, you can export data using the imperative API available in XGrid:
```
oliviertassinari, line 15: A detail. I wonder if we shouldn't prefix these sections with apiRef and the pro icon. I wonder because we might be able to:

- avoid frustration for developers using the DataGrid, reading this section only to later realize it's a paid feature.
- create some convention, for instance, we could later use it to identify where we should like the section about how to use the apiRef.

m4theushw: If we add the apiRef header in each export output (CSV, Excel, print) we can't have specific links to them, unless we name the sections as "CSV - apiRef". Another alternative would be to have only one "apiRef" at the bottom of the page with the consolidated methods from all outputs, but we don't have a interface that groups them.

oliviertassinari: > we can't have specific links to them

the docs uses a increment `#api-ref`, `#api-ref-2`, etc. I believe.

PR #2196, opened 2021-07-23
[DataGrid] Fix `disableMultipleColumnsFiltering` console warning

ZeeshanTamboli:
Warning text should be `filtering` instead of `sorting` for `disableMultipleColumnsFiltering` prop.

### DIFF
--- a/packages/grid/data-grid/src/DataGrid.tsx
+++ b/packages/grid/data-grid/src/DataGrid.tsx
@@ -148,7 +148,7 @@ DataGrid.propTypes = {
       return new Error(
         [
           `Material-UI: \`<DataGrid disableMultipleColumnsFiltering={false} />\` is not a valid prop.`,
-          'Only single column sorting is available in the MIT version.',
+          'Only single column filtering is available in the MIT version.',
           '',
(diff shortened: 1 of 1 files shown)


flaviendelangle (approved):

oliviertassinari:
@flaviendelangle Two quick notes regarding merging pull requests. The UI allows to modify the final commit:

Capture d’écran 2021-07-31 à 13 48 34

I think that we should leverage it:

1. **The title**. When there is a single commit, GitHub uses the title of the commit instead of the title of the PR for the final commit. This means in practice that we end-up with:

  ```sh
  - fix warning (#2196) @ZeeshanTamboli
  ```

  In the generated changelog, instead of:

  ```sh
  - [DataGrid] Fix disableMultipleColumnsFiltering console warning (#2196) @ZeeshanTamboli
  ```

2. **The message**. When there are multiple commits. GitHub use the titles of the commits to generating a description. This leads to noise when using `git log`. 

  Capture d’écran 2021-07-31 à 13 46 37

  or in the GitHub UI:

  Capture d’écran 2021-07-31 à 13 46 25

  This one is not that important, compared to 1. it matters a lot less.

PR #2208, opened 2021-07-25
[DataGrid] Canonical controlled state behavior

oliviertassinari:
I have used #2200 as an entry to find myself interest in the controlled state. Along the way, I have noticed a couple of anti-patterns that I went on fixing:

- The presence of a onXXX prop shouldn't change the behavior of the state
- Using deepEqual to compare props is no longer relevant since we compare the reference in the useGridControlState
- The user's onXXX prop is supposed to be called after the component has handled the logic. It's how we aim to make it work in the core. It's not always possible, for instance, when we need to do `event.isMuiDefaultPrevented`.
- The event should be triggered, regardless of if the change was persisted in the state or not.
- Change the logic to have fewer branches

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
+++ b/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
@@ -1,13 +1,13 @@
 import React from 'react';
 import { GridApiRef } from '../../../models/api/gridApiRef';
 import { GridControlStateApi } from '../../../models/api/gridControlStateApi';
-import { ControlStateItem } from '../../../models/controlStateItem';
+import { GridControlStateItem } from '../../../models/controlStateItem';
 import { useGridApiMethod } from '../../root/useGridApiMethod';
 
 export function useGridControlState(apiRef: GridApiRef) {
-  const controlStateMapRef = React.useRef<Record<string, ControlStateItem<any>>>({});
+  const controlStateMapRef = React.useRef<Record<string, GridControlStateItem<any>>>({});
 
-  const updateControlState = React.useCallback((controlStateItem: ControlStateItem<any>) => {
+  const updateControlState = React.useCallback((controlStateItem: GridControlStateItem<any>) => {
     const { stateId, stateSelector, ...others } = controlStateItem;
 
     controlStateMapRef.current[stateId] = {
--- a/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
+++ b/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
@@ -64,9 +64,7 @@ export function useGridControlState(apiRef: GridApiRef) {
               controlState.propOnChange(model);
             }
 
-            if (controlState.onChangeCallback) {
-              controlState.onChangeCallback!(model);
-            }
+            apiRef.current.publishEvent(controlState.changeEvent, model);
           });
         },
       };
--- a/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
+++ b/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
@@ -19,7 +19,7 @@ export function useGridControlState(apiRef: GridApiRef) {
 
   const applyControlStateConstraint = React.useCallback(
     (newState) => {
-      let shouldUpdate = true;
+      let ignoreSetState = false;
       const updatedStateIds: string[] = [];
       const controlStateMap = controlStateMapRef.current!;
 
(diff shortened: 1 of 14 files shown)


flaviendelangle:
0 test broken on #2099 with this PR

> The presence of a onXXX prop shouldn't change the behavior of the state

How we it changing the behavior before ?

> The user's onXXX prop is supposed to be called after the component has handled the logic. It's how we aim to make it work in the core. It's not always possible, for instance, when we need to do event.isMuiDefaultPrevented.

OK to favor consistency with the core

oliviertassinari:
> How we it changing the behavior before?

I'm not exactly sure. I was mostly referring to the skip update code was was discriminating the presence of a prop callback or not to assign `shouldUpdate`. I have renamed the variable to `ignoreSetState` to be biased toward: true => action. false => default.

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
```diff
@@ -18,29 +18,25 @@ export const useGridState = (
...
 
-        postUpdate();
+        if (apiRef.current.publishEvent) {
+          const params: GridStateChangeParams = { api: apiRef.current, state: newState };
+          apiRef.current.publishEvent(GRID_STATE_CHANGE, params);
+        }
       }
 
-      return hasChanged;
+      postUpdate();
```
m4theushw, line 37: It should only be called if `ignoreSetState` is false. The following test passes on HEAD but here it doesn't.

```diff
diff --git a/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx b/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
index f709fad2..4c453247 100644
--- a/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
+++ b/packages/grid/x-grid/src/tests/selection.XGrid.test.tsx
@@ -15,6 +15,7 @@ import {
   GridSelectionModel,
   useGridApiRef,
   XGrid,
+  GRID_SELECTION_CHANGE
 } from '@material-ui/x-grid';

 const isJSDOM = /jsdom/.test(window.navigator.userAgent);
@@ -322,5 +323,15 @@ describe('<XGrid /> - Selection', () => {
       expect(getRow(1)).to.have.class('Mui-selected');
       expect(getRow(2)).to.have.class('Mui-selected');
     });
+
+    it('should not publish GRID_SELECTION_CHANGE if the selection state did not change ', () => {
+      const onSelectionChange = spy();
+      const selectionModel = [];
+      render(<Test selectionModel={selectionModel} />);
+      apiRef.current.subscribeEvent(GRID_SELECTION_CHANGE, onSelectionChange);
+      expect(onSelectionChange.callCount).to.equal(0);
+      apiRef.current.setSelectionModel(selectionModel);
+      expect(onSelectionChange.callCount).to.equal(0);
+    });
   });
 });
```

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
```diff
@@ -19,49 +19,53 @@ export function useGridControlState(apiRef: GridApiRef) {
...
-              controlStateMap[stateId].onChangeCallback!(model);
+            const controlState = controlStateMap[stateId];
+            const model = controlStateMap[stateId].stateSelector(newState);
+
+            if (controlState.propOnChange) {
+              controlState.propOnChange(model);
+            }
+
+            if (controlState.onChangeCallback) {
+              controlState.onChangeCallback!(model);
```
m4theushw, line 68: I was thinking if shouldn't we add a `modelChangeEvent` to `ControlStateItem` and publish automatically the event. This callback in 100% of the cases is only used to publish the event.

packages/grid/_modules_/grid/hooks/features/core/useGridState.ts
oliviertassinari: Ah, well spotted, I'm adding a new commit to fix this.

packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
oliviertassinari: Ok done

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -301,24 +301,21 @@ export const useGridFilter = (
...
-      const hasItemsWithoutIds = filterModel.items.find((item) => item.id == null);
+    if (props.filterModel !== undefined && props.filterModel.items.length > 1) {
+      const hasItemsWithoutIds = props.filterModel.items.find((item) => item.id == null);
       if (hasItemsWithoutIds) {
         throw new Error(
           "The 'id' field is required on filterModel.items when you use multiple filters.",
         );
       }
     }
     const oldFilterModel = apiRef.current.state.filter;
```
flaviendelangle, line 317: We could use `getState()`

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
oliviertassinari: Agree, I'm merging to show a conflict with https://github.com/mui-org/material-ui-x/pull/2176/files#r680404558. We can fix this in a follow up

PR #2210, opened 2021-07-26
[docs] Handle generics in api doc generation

### LINKED ISSUE #2190: [docs] The doc api builder should support generics
The generics are currently not well supported by the doc generator

## Current Behavior 😯

The following line of a TypeScript interface generates a Markdown line without the Generic which makes the documentation unusable if we add generics to the public methods.

```ts
getCellValue: <Value extends GridCellValue = GridCellValue>(id: GridRowId, field: string) => Value;
```

```md
 <span class="prop-type">(id: GridRowId, field: string) =&gt; Value</span> 
```

## Expected Behavior 🤔

Should output something like

```md
<span class="prop-type">\<Value extends GridCellValue = GridCellValue\>(id: GridRowId, field: string) =&gt; Value</span>
```

## Steps to Reproduce 🕹

Edit the `getCellValue` in https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/models/api/gridParamsApi.ts to match the code example above and run `yarn docs:api:build`

flaviendelangle:
Closes #2190 

image

I worked with a generic-based `getCellValue` on my local env to also handle the default value, but I don't want to change the `getCellValue` signature here.

I'm currently just reproducting the TypeScript syntax in the column of the table. But with more complex generics, it could become heavy. If you have a better idea on how to represent the generics.


docs/scripts/buildApi.ts
```diff
@@ -30,11 +30,32 @@ function generateType(type, needsParenthesis = false) {
...
+
+          if (generic.type) {
+            genericLine += ` extends ${generateType(generic.type)}`;
+          }
+
+          if (generic.default) {
+            genericLine += ` = ${generateType(generic.default)}`;
+          }
+
+          return genericLine;
```
m4theushw, line 51: We can do the same with less lines:

```suggestion
          let genericLine = generic.name;
          genericLine += generic.default ? ' =' : ' extends';
          return `${genericLine} ${generateType(generic.default || generic.type)}`;
```

m4theushw: Yeah, I see. If you like we can use the one-line if syntax: `if (something) genericLine  += '...';` Just to keep all branches small.

flaviendelangle: Done :+1:

m4theushw (approved):

PR #2213, opened 2021-07-26
[core] Use rootProps instead of options in Grid components except for classes

flaviendelangle:
Next step for migrating away from `options`


packages/grid/_modules_/grid/hooks/features/useGridComponents.tsx
```diff
@@ -1,8 +1,72 @@
...
+  GridTableRowsIcon,
+  GridTripleDotsVerticalIcon,
+  GridViewHeadlineIcon,
+  GridViewStreamIcon,
+} from '../../components';
+import { GridColumnUnsortedIcon } from '../../components/columnHeaders/GridColumnUnsortedIcon';
+import { ErrorOverlay } from '../../components/ErrorOverlay';
+import { GridNoResultsOverlay } from '../../components/GridNoResultsOverlay';
+
+const DEFAULT_GRID_SLOTS_ICONS: GridIconSlotsComponent = {
```
flaviendelangle, line 37: I had to move the `DEFAULT_GRID_SLOTS_ICONS` outside of `gridSlotsComponent.ts`
Otherwise, any component listed here could not use the context due to circular import.

packages/grid/_modules_/grid/components/GridPagination.tsx
```diff
@@ -43,13 +43,12 @@ export const GridPagination = React.forwardRef<
 >(function GridPagination(props, ref) {
   const classes = useStyles();
   const apiRef = useGridApiContext();
+  const rootProps = React.useContext(GridRootPropsContext)!;
```
DanailH, line 46: Can we get rid of the `!` at the end?

flaviendelangle:
```ts
const GridRootPropsContext = React.createContext<GridComponentProps | undefined>(undefined);
```

We could do the following :

```
const useGridRootProps = () => {
  const contextValue = React.useContext(GridRootPropsContext)

  if (!contextValue) {
    throw new Error('useGridRootProps should only be used inside the DataGrid / XGrid component
  }
}
```

Or remove the `undefined` by passing the default props :

```ts
const GridRootPropsContext = React.createContext<GridComponentProps>({ ...DEFAULT_GRID_OPTIONS, rows: [], columns: [] });
```

packages/grid/_modules_/grid/hooks/utils/useGridRootProps.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
```
oliviertassinari, line 2:
```suggestion
```

packages/grid/_modules_/grid/hooks/utils/useGridRootProps.ts
```diff
@@ -0,0 +1,15 @@
...
+import { GridRootPropsContext } from '../../context/GridRootPropsContext';
+
+export const useGridRootProps = () => {
+  const contextValue = React.useContext(GridRootPropsContext);
+
+  if (!contextValue) {
+    throw new Error(
+      'Material-UI: useGridRootProps should only be used inside the DataGrid / XGrid component',
```
oliviertassinari, line 10: en_US spacing rule 

```suggestion
      'Material-UI: useGridRootProps should only be used inside the DataGrid/XGrid component',
```

packages/grid/_modules_/grid/hooks/utils/useGridRootProps.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { GridRootPropsContext } from '../../context/GridRootPropsContext';
+
+export const useGridRootProps = () => {
```
DanailH, line 4: I think that is better. I added @m2mathew but other than that I'm ok with the change.

DanailH (approved):

PR #2220, opened 2021-07-27
[DataGrid] Polish cell editing

m4theushw:
### Breaking changes

- [DataGrid] Replace the `onCellModeChange` prop with `onCellEditEnter` or `onCellEditExit`.

- [DataGrid] Rename `onCellEditEnter` to `onCellEditStart` and `onCellEditEnd` to `onCellEditStop`.

  ```diff
  <DataGrid 
  - onCellEditEnter={...}
  - onCellEditExit={...}
  + onCellEditStart={...}
  + onCellEditStop={...}
  />
  ```

- [XGrid] The `setEditCellProps` API call is not available anymore. Use the [controlled editing](https://material-ui.com/components/data-grid/editing/#controlled-editing) or `setEditRowsModel`.

  ```diff
  -apiRef.current.setEditCellProps({ id, field, props: { ...props, error: true } });
  +apiRef.current.setEditRowsModel({
  +  ...oldModel,
  +  [id]: {
  +    ...oldModel[id],
  +    [field]: { ...oldModel[id][field], error: true },
  +  },
  +});
  ```

---

- 8bdbf6f - removes the `onCellModeChange` prop because there're already specific events for entering/exiting the edit mode
- e2c76f3 - adds the `onCellEditEnter` and `onCellEditExit` prop to replace the above prop
- 41144de - I was hoping we would be able to make `isCellEditable` private but we can't, it's used to set a CSS class in the cell
- fe1eb32 - don't expose `setEditCellProps` because users can change the model controlling the state or with `setEditRowsModel`
- 22f72f0 - the GRID_CELL_EDIT_ENTER event was being published even if the cell was not editable

Preview: https://deploy-preview-2220--material-ui-x.netlify.app/components/data-grid/editing/


flaviendelangle:
> [DataGrid] Replace the onCellModeChange prop with onCellEditEnter or onCellEditExit.

What's the goal of this rename ?

m4theushw:
> > [DataGrid] Replace the onCellModeChange prop with onCellEditEnter or onCellEditExit.
> 
> What's the goal of this rename ?

@flaviendelangle It's not a rename, but a removal. Some reasons:

1. If the users wants to cancel the enter/exit default behavior he has to use events, now there're props for that too. https://github.com/mui-org/material-ui-x/issues/2146#issuecomment-879968581
2. `onCellModeChange` is not used in any demo and it's kinda redudant. We already have events like GRID_CELL_EDIT_ENTER where the user can attach a custom behavior and by listing to this event the user is also listening for the mode change.
3. Keep the exposed API small

docs/src/pages/components/data-grid/editing/ValidateServerNameGrid.js
```diff
@@ -47,20 +47,26 @@ export default function ValidateServerNameGrid() {
         clearTimeout(promiseTimeout);
         clearTimeout(keyStrokeTimeoutRef.current);
 
-        apiRef.current.setEditCellProps({
-          id,
-          field,
-          props: { ...props, error: true },
+        let oldModel = apiRef.current.getEditRowsModel();
+        apiRef.current.setEditRowsModel({
```
flaviendelangle, line 51: Would it make sense to introduce a callback version of `setEditRowsModel` ?

```suggestion
        apiRef.current.setEditRowsModel(oldModel => ({
          ...oldModel,
          [id]: {
            ...oldModel[id],
            [field]: { ...oldModel[id][field], error: true },
          },
        });
```

It would reduce the risk of a user passing an outdated model.

m4theushw: We could, but in the docs we encourage the use of the `editRowsModel` prop, which doesn't suffer from this outdated-model risk. The imperative API is kinda hidden. We can keep the callback version in the backlog because there're others `setSomethingModel` methods and, for consistency, they should also support it.

flaviendelangle:
Quite a lot of git conflict
But otherwise it's good for me

docs/pages/api-docs/data-grid/x-grid.md
```diff
@@ -83,6 +82,8 @@ import { XGrid } from '@material-ui/x-grid';
 | <span class="prop-name">onError</span> | <span class="prop-type">(args: any) => void</span> |   | Callback fired when an exception is thrown in the grid, or when the `showError` API method is called. |
 | <span class="prop-name">onEditCellPropsChange</span> | <span class="prop-type">(params: GridEditCellPropsParams, event: React.SyntheticEvent) => void</span> |   |  Callback fired when the edit cell value changes. |
 | <span class="prop-name">onCellEditCommit</span> | <span class="prop-type">(params: GridEditCellPropsParams, event: React.SyntheticEvent) => void</span> |   | Callback fired when the cell changes are going to be committed. |
+| <span class="prop-name">onCellEditEnter</span> | <span class="prop-type">(params: GridCellParams, event: React.SyntheticEvent) => void</span> |   | Callback fired when the cell turns to edit mode. |
```
DanailH, line 85: Does `onCellEditStart` and `onCellEditEnd` or `onCellEditStop` make more sense?

m4theushw: Yes, we already use this suffix in the resize events.

packages/storybook/src/stories/grid-rows.stories.tsx
```diff
@@ -688,10 +698,31 @@ export function ValidateEditValueServerSide() {
         clearTimeout(promiseTimeout);
         clearTimeout(keyStrokeTimeoutRef.current);
 
-        apiRef.current.setEditCellProps({ id, field, props: { ...props, error: true } });
+        let oldModel = apiRef.current.getEditRowsModel();
```
oliviertassinari, line 701: Using "old" feels strange. It's the current value of the model at the time it's called. Maybe we should reverse the tradeoff? Call "new" for the new value we aim to push?

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -302,25 +304,23 @@ export function useGridEditRows(
...
   useGridApiEventHandler(apiRef, GRID_COLUMN_HEADER_DRAG_START, handleColumnHeaderDragStart);
-  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_ENTER, handleCellEditEnter);
-  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_EXIT, handleCellEditExit);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_START, handleCellEditEnter);
+  useGridApiEventHandler(apiRef, GRID_CELL_EDIT_STOP, handleCellEditExit);
   useGridApiEventHandler(apiRef, GRID_CELL_EDIT_COMMIT, handleCellEditCommit);
   useGridApiEventHandler(apiRef, GRID_EDIT_CELL_PROPS_CHANGE, handleEditCellPropsChange);
 
   useGridApiOptionHandler(apiRef, GRID_CELL_EDIT_COMMIT, options.onCellEditCommit);
-  // TODO remove, use onEditRowsModelChange directly
```
oliviertassinari, line 311: Did we find cases where using `onEditCellPropsChange` is relevant over `onEditRowsModelChange` or can we still remove the `onEditCellPropsChange` prop?  I recall there was something about extra props that I didn't understand the use case.

oliviertassinari (commented):
This docs is outdated https://deploy-preview-2220--material-ui-x.netlify.app/components/data-grid/editing/#events (cellEditEnter/etc.)

packages/storybook/src/stories/grid-rows.stories.tsx
m4theushw: I updated everything to `newModel`.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
m4theushw: To do server-side validation `onEditCellPropsChange` is relevant. While we don't provide a reason argument we need to support this prop. In the mean time, I kept its usage low leveraging `onEditRowsModelChange` whenever I could.

PR #2232, opened 2021-07-29
[DataGrid] Fix CSV export when selected row id is number

### LINKED ISSUE #2156: [DataGrid] Export selected rows

## Summary 💡

We'd like to have the ability to export only selected rows instead of the whole set of rows.

## Motivation 🔦

Sometimes it's required to export part of the data, filtering might help, but can be a bit complex for that use case. So selecting and exporting desired entries is a better option here.

## Order id 💳

#27213

flaviendelangle:
Closes #2156 

CSV export does not work when row ID are numbers and some rows are selected


DanailH (approved):

oliviertassinari:
@flaviendelangle I have no idea how you found this, It was a long short reading #2156, well done!

Issue #2227, opened 2021-07-28
[RFC] Clean `x-data-grid` and `x-data-grid-pro`  exports


flaviendelangle:
Relates to #924 
Relates to #18 

## Proposals

1. Stop exporting everything from the grid

```
export * from './components';
export * from './constants';
export * from './hooks';
export * from './locales';
export * from './models';
export * from './utils';
export * from './GridComponentProps';
```

We currently export from the module root almost every variable / type ever exported from any file inside the grid codebase.
We should be able to keep some variables / types private 

And even if we have an `export *` cascade, we almost always import from the leafs to avoid circular imports.

Here are a few things we should stop exporting without reason:

- the features hooks (`DataGrid` exports `useGridColumnReorder` even if it does not use it). We may export them when we'll have a hook-only API but currently they can never be properly used outside of the Grid.

- the internal utilities, (`isMultipleKey`, `isEscapeKey`, `isDesc`, ...)

- the internal data transformation functions (`mergeGridColTypes`, ...)

2. Group the event constants into a single enum

We are exporting dozens of variables (`GRID_RESIZE`, `GRID_KEYDOWN`, `GRID_COLUMN_HEADER_FOCUS`) when we could only export a single enum `GridEvent`. 
This would be more explicit and would also allow to strongly type `publishEvent` first argument (and later on type the second argument).

3. Group the CSS class variables

Probably related to #1950 

## Motivations

- Limit the breaking changes. If we export everything, renaming an internal variable or type is theoretically a breaking change. In a PR currently in work, I stop exporting `DEFAULT_GRID_SLOTS_COMPONENTS` because I moved it inside the single file that is using it, it's breaking because the variable is exported even if never used by anyone.

- Make the import clearer. Having hundreds of imports is a bit overwhelming.

- Never export something related to XGrid in the DataGrid package

- Make it easier to reduce bundle size of DataGrid by knowing what is in it

m4theushw:
> 1. Stop exporting everything from the grid

TypeDoc only generates documentation for what is exported. By not exporting a particular interface we might lose its docs. We have to have this in mind.

> the internal utilities, (isMultipleKey, isEscapeKey, isDesc, ...)

These we can definitely stop exporting them ASAP.

> the internal data transformation functions (mergeGridColTypes, ...)

Same as above. +1 for not exporting.

> 2. Group the event constants into a single enum

[...]

PR #2233, opened 2021-07-29
[DataGrid] Don't export the internal utils

flaviendelangle:
Part of #2227 

The following variables and types are not exported anymore from `@material-ui/data-grid` and `@material-ui/x-grid`

```diff
// utils.domUtils.ts

- findParentElementFromClassName
- getRowEl
- isGridCellRoot
- isGridHeaderCellRoot
- getIdFromRowElem
- getFieldFromHeaderElem
- findHeaderElementFromField
- findGridCellElementsFromCol
- escapeOperandAttributeSelector
- getGridColumnHeaderElement
- getGridRowElement
- getGridCellElement
- isOverflown

// utils/exportAs.ts
- exportAs

// utils/getGridLocalization
- LocalizationV4
- LocalizationV5
- Localization
- getGridLocalization

// utils/keyboardUtils.ts
- isHomeOrEndKeys
- isPageKeys
- isDeleteKeys
- printableCharRegex
- isPrintableKey
- GRID_MULTIPLE_SELECTION_KEYS
- GRID_CELL_EXIT_EDIT_MODE_KEYS
- GRID_CELL_EDIT_COMMIT_KEYS
- isMultipleKey
- isCellEnterEditModeKeys
- isCellExitEditModeKeys
- isCellEditCommitKeys
- isNavigationKey
- isKeyboardEvent
- isHideMenuKey

// utils/material-ui-utils.ts
- DistributiveOmit
- InternalStandardProps
- useEventCallback
- useEnhancedEffect
- useThemeProps
- composeClasses
- generateUtilityClass
- generateUtilityClasses

// utils/mergeUtils.ts
- mergeGridColTypes
- removeUndefinedProps
- mergeGridOptions

// utils/sortingUtils.ts
- nextGridSortDirection
- isDesc
- gridNillComparer
- gridStringNumberComparer
- gridNumberComparer
- gridDateComparer

// utils/utils.ts
- isDeepEqual
- isNumber
- isFunction
- isObject
- getMuiVersion
- muiStyleAlpha
- createTheme
- localStorageAvailable
- Optional
- getDataGridUtilityClass
- escapeRegExp
- getThemePaletteMode
```

These three utils are used in the demo.
Do we want to export them or let the user re-implement them ?

```
- getThemePaletteMode
- escapeRegExp
- isOverflown
```

For `escapeRegExp` and `isOverflown`, I don't think it's the role of MUI to provide such generic methods.
For `getThemePaletteMode`, the core could export it but if it does not I suppose we shouldn't do it


test/regressions/TestViewer.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import { useFakeTimers } from 'sinon';
 import { withStyles } from '@material-ui/styles';
-import { createTheme } from '@material-ui/data-grid';
+import { createTheme } from '../../packages/grid/_modules_/grid/utils/utils';
```
oliviertassinari, line 6: Interesting, we can't use `@material-ui/data-grid/utils/utils` because of the `_modules_` folder structure. This folder structure prevents us to split MIT and Commercially licensed code. So we need to move most of the modules away. Once we do how do XGrid and DataGrid share the utils? I guess we have to either keep the utils in or export them as `unstable_`. Anyway, it's a problem for our future self. Making the utils private is the right thing to do.

oliviertassinari:
> Do we want to export them or let the user re-implement them ?

+1 for re-implementation

test/regressions/TestViewer.js
flaviendelangle: Yes, removing the unnecessary exports is a first step to help us reorganize the codebase between the MIT and the Commercially licensed code.

And there is some code that is not Grid related and should not be exported and not be in a `grid` folder but probably in a `_shared_` directory that will be accessed by the grids but also the new X components.
That's the case for a lot of the utils.

packages/grid/_modules_/grid/utils/getGridLocalization.ts
```diff
@@ -2,13 +2,13 @@ import { GridLocaleText } from '../models/api/gridLocaleTextApi';
...
 import { getMuiVersion } from './utils';
 
-export interface LocalizationV4 {
+interface LocalizationV4 {
   props: {
     MuiDataGrid: Pick<GridOptions, 'localeText'>;
   };
 }
 
-export interface LocalizationV5 {
```
DanailH, line 11: Nice catch. I bet there are even more places like this where functions or typings are exported even though they are not used anywhere else. We should keep an eye on that.

DanailH (approved):

test/regressions/TestViewer.js
oliviertassinari: >  but probably in a _shared_ directory that will be accessed by the grids but also the new X components.

If it needs to be access cross component, then we need a similar package to `@material-ui/utils`. Maybe we should name them:

- `@mui/core-utils` (prev `@material-ui/utils`)
- `@mui/x-utils`

Rupert-com:
OK, copied that there have been changes applied. 

**pre Update**
there was this way to access the object shapes.
```typescript
import { createTheme, DataGrid, GridCellParams, GridColDef, GridEditCellPropsParams, GridRowModel, GridRowSelectedParams, GridSelectionModelChangeParams, GRID_CELL_EDIT_PROPS_CHANGE } from '@material-ui/data-grid'
[...]

const renderEditCell = (params: GridCellParams) => {
    return <EditTypeOfProduct {...params} Companys={values.Companys} />
}
```
After the update
``` bash
Module '"@material-ui/data-grid"' has no exported member 'createTheme'.
Module '"@material-ui/data-grid"' has no exported member 'GridRowSelectedParams'.
Module '"@material-ui/data-grid"' has no exported member 'GridSelectionModelChangeParams'.
```

Is there a new alternative way?

oliviertassinari:
@Rupert-com

- createTheme import from the core package:
```jsx
import { createTheme } from '@material-ui/core/styles';
```
- GridRowSelectedParams see #1823
- GridSelectionModelChangeParams see #1823

PR #2235, opened 2021-07-29
[core] New column internal field: computedWidth

flaviendelangle:
WIP

Part of the split of #2181 

- Store the real width of the columns in `column.computedWidth`
- Create distinct types for the columns given in the props (`GridColDef`) and columns stored in the state (`GridStateColDef`), the latter has additional prepossessing (currently we only have `computedWidth`)


flaviendelangle:
The argos diff seems to be due to the removal of the `Math.floor` in the flex column width computation.

docs/src/pages/components/data-grid/columns/RenderExpandCellGrid.tsx
```diff
@@ -121,7 +121,7 @@ function renderCellExpand(params: GridCellParams) {
   return (
     <GridCellExpand
       value={params.value ? params.value.toString() : ''}
-      width={params.colDef.width}
+      width={params.colDef.computedWidth}
```
oliviertassinari, line 124: Should this variable be in the public API landscape? Yeah, maybe.

packages/grid/_modules_/grid/hooks/features/columns/gridColumnsSelector.ts
```diff
@@ -9,44 +13,47 @@ export const allGridColumnsSelector = createSelector<
...
+  const positions: number[] = [];
 
-    const totalWidth = visibleColumns.reduce((totalW, curCol) => {
-      positions.push(totalW);
-      return totalW + curCol.width!;
-    }, 0);
+  const totalWidth = visibleColumns.reduce((totalW, curCol) => {
+    positions.push(totalW);
+    return totalW + curCol.computedWidth;
+  }, 0);
```
oliviertassinari, line 36: Maybe use the default name we use for the reducer accumulator instead of a strangely named variable?

```suggestion
  const totalWidth = visibleColumns.reduce((acc, curCol) => {
    positions.push(acc);
    return acc + curCol.computedWidth;
  }, 0);
```

oliviertassinari (commented):
It looks good. Regarding the direct value of this PR for the developers, IMHO, it's mean that now, when they store the state to restore a previous session, the width that they are going to restore, is true to how the end-users have manually resized the column, regardless of what was the user-agent window width.

oliviertassinari:
> The argos diff seems to be due to the removal of the Math.floor in the flex column width computation.

Maybe we can restore the Math.floor and work on #2011 independently?

docs/src/pages/components/data-grid/columns/RenderExpandCellGrid.tsx
flaviendelangle: I think so
Otherwise, custom cell renderer that really need the width of the cell would need to capture it with observers of other complex approach which seems counterproductive.

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -160,13 +160,13 @@ export interface GridColDef {
...
 }
 
-export interface GridColumnProp extends Omit<GridColDef, 'filterOperators'> {
-  filterOperators?: GridFilterOperator[] | string;
-}
-
 export type GridColumns = GridColDef[];
 export type GridColTypeDef = Omit<GridColDef, 'field'> & { extendType?: GridNativeColTypes };
 
+export interface GridStateColDef extends GridColDef {
```
DanailH, line 166: Why do we need to extend here, can't we just add it directly to the `GridColDef`, if it is going to be publicly available?

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -36,57 +37,60 @@ import { useGridApiOptionHandler } from '../../root/useGridApiEventHandler';
...
-        if (!column.flex) {
-          viewportWidth -= Math.max(column.minWidth!, column.width!);
-        } else {
-          flexDivider += column.flex;
-        }
+type RawGridColumnsState = Omit<GridColumnsState, 'lookup'> & {
+  lookup: { [field: string]: GridColDef | GridStateColDef };
+};
+
+function getStateColumns(
```
DanailH, line 44: As far as I can see this handles only the `width` of the columns? Maybe we can have a more descriptive name, like `getComputedColumns` or `recomputeColumnsWidth` or something else?

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
flaviendelangle: It allows to differenciate the input columns and the state columns
If I add it to `GridColDef`, then the users could think they can provid a `computedWidth`.


`GridColDef` = the prop `columns` given to `DataGrid` and `XGrid`
`GridStateColDef` = the columns stored in the state and returned in callbacks.

The only difference is currenctly the `computedWidth` but I think their will be more preprocessings and additions in the future.

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
flaviendelangle: If you prefer the naming `computedColumns` over `stateColumns` I can rename `GridStateColDef` into `GridComputedColDef` and `getStateColumns` into `getComputedColumns`

I'm note a fan to limit the method name to the concept of `width`.

DanailH: My argument was because it only takes care of the `width` but as it is now is also ok.

DanailH (approved):

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
flaviendelangle: We can also argue that its best to keep a name related to the width for now and rename it to its generic naming when a 2nd preprocessing occurs.

docs/src/pages/components/data-grid/columns/RenderExpandCellGrid.js
```diff
@@ -118,7 +118,7 @@ function renderCellExpand(params) {
   return (
     <GridCellExpand
       value={params.value ? params.value.toString() : ''}
-      width={params.colDef.width}
+      width={params.colDef.computedWidth}
```
oliviertassinari, line 121: This PR could be considered breaking by some, e.g. say the user copied this demo in the past and uses flex: 1. https://codesandbox.io/s/material-demo-forked-zpk7c?file=/demo.js. I would have personally not used "core" as the primary label, but it's a detail. We will see if people complain.

flaviendelangle: You're right
I guess it depends if you want to be exhaustive in your breaking change or if you want to focus on the main ones.
We can update the label if you prefer ?

oliviertassinari: > I guess it depends if you want to be exhaustive in your breaking change or if you want to focus on the main ones.

It's a long time dilemma discussion @eps1lon and I have had. I personally think that there is a grey scale, bug fixes change behaviors, wrong behaviors developers might rely on.

When developers raise a change of behavior in an issue that wasn't documented in the changelog or versioning, it's usually a good sign that we haven't been strict enough

At the end of the day, I think that it's about proving the best update experience as possible for developers. Over communication might get us into https://overreacted.io/npm-audit-broken-by-design/, but user communication might even be worse.

eps1lon: npm-audit is problematic because we can't communicate as well as we can with SemVer. This is a breaking change. A breaking change is not an "opinion" or grey area. It is well defined.

Issue #1821, opened 2021-06-03
[DataGrid] Should not expose apiRef


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Developers can access apiRef in the community plan, while we document it a Pro plan feature.

## Expected Behavior 🤔

Developers shouldn't be able to access apiRef in `DataGrid`, **only** in `XGrid`.

## Steps to Reproduce 🕹

Steps:

1. https://codesandbox.io/s/material-demo-forked-59fsr?file=/demo.tsx

```jsx
export default function DataGridDemo() {
  return (
    <div style={{ height: 400, width: "100%" }}>
      <DataGrid
        onRowSelected={(parmas) => {
          console.log(Array.from(parmas.api.current.getSelectedRows().entries()));
        }}
        rows={rows}
        columns={columns}
        pageSize={5}
        checkboxSelection
      />
    </div>
  );
}
```

## Context 🔦

Saw it in https://github.com/mui-org/material-ui-x/issues/1820

As a side note, the current types and naming looks strange:

1. Why aren't we reducing the types to `GridApiRef` instead of any?

https://github.com/mui-org/material-ui-x/blob/fa346f0fbe3d9b9eea9bb403fe4675f544d6abf9/packages/grid/_modules_/grid/models/params/gridRowSelectedParams.ts#L15-L18

2. Why are naming the prop `api` if we don't provide `GridApi`? Shouldn't it be called `apiRef`?

3. Why expose apiRef when we could provide the api directly?

## Your Environment 🌎

4.0.0-alpha.30

dtassone:
> Why aren't we reducing the types to GridApiRef instead of any?

due to circle dependencies

> Why are naming the prop api if we don't provide GridApi? Shouldn't it be called apiRef?

Yes there is an inconsistency in that, originally I wanted to provide `apiRef.current` so it's a bit nicer for the user but I think it can create issues.

> Why expose apiRef when we could provide the api directly?

If we expose api then the state of the api that was attached to current at the time we pass might be outdated if we use it outside a function or a useEffect. Hence why we use a ref.

oliviertassinari:
@dtassone Ok, so we should only rename api -> apiRef?

dtassone:
yes for consistency everywhere.

oliviertassinari:
[...]

Issue #1301, opened 2021-03-26
[DataGrid] Standardize/Reconcile all event handlers and params


dtassone:
1. Events should always be fired with the `params` of the entity it comes from, or with the params relative to its action
ie `onCellClick` has `GridCellParams` 
And `onFilterModelChange` has `GridFilterModelParams` 

Getting params, can be done using apiRef
- `apiRef.current.getCellParams(id: GridRowId, field: string): GridCellParams`;
- We should add `apiRef.current.getColumnHeaderParams(field: string): GridColumnHeaderParams`;

Then it should have the event as a second parameter. 
- This is because event is optional in some cases
ie `options.onColumnReorder(params: GridColumnParams, event?: React.SyntheticEvent); `
```tsx
// in hook
setColumnIndex = (field, newIndex, event?) => {
   apiRef.publishEvent(GRID_COLUMN_REORDER, apiRef.current.getColumnHeaderParams(field), event)
}

// in Api
setColumnIndex: (field, newIndex)=> void
```

2. Events should be granular so it's easily catchable by our users.

Imagine I want to log the cell that users are trying to edit. I could subscribe to the double click event but there are many different ways of getting into edit mode, keyboard, calling the api directly, and one could do it on click...
With granular events, I can easily hook into the event and achieve the above.

```tsx
export function EditCellWithMessageGrid() {
  const apiRef = useGridApiRef();
  const [message, setMessage] = React.useState('');

  React.useEffect(()=> {
    return apiRef.current.subscribeEvent('cellEnterEdit', (param: GridCellParams, event?: React.SyntheticEvent)=> {
      LogUser(`user ${username} is editing `, params) 
      setMessage(`Editing cell with value: ${param.value} at row: ${param.rowIndex}, column: ${param.field},
                        triggered by ${event!.type}
      `);
    });
  }, [apiRef])

  React.useEffect(()=> {
    return apiRef.current.subscribeEvent('cellExitEdit', ()=> {
      LogUser(`user ${username} stopped editing `, params) 
      setMessage('');
    });
  }, [apiRef])

  return (
    <div style={{ height: 400, width: '100%' }}>
      <XGrid
        {...baselineEditProps}
        apiRef={apiRef}
        autoHeight
      />
      { message && <Alert severity="info">{message}</Alert> }
    </div>
  );
}
```

oliviertassinari:
Closing for now, we have fixed 80% of the pain.

PR #2236, opened 2021-07-29
[XGrid] Add options param each callback option in `XGrid`

DanailH:
Part 1 of 2 from #1821 

I'll split the effort into 2 PR. The first one (this one) adds an additional `options` param to each callback and only for `XGrid` it adds `api: apiRef.current`  to that object. The idea is that later on if we decide we can also add a `reason` to the same object.

Unfortunately due to circular dependency, the `api` can only be typed at the place where the assignment happens, in `useGridApiEventHandler`


flaviendelangle:
> Unfortunately due to circular dependency, the api can only be typed at the place where the assignment happens, in useGridApiEventHandler.

@DanailH even with the following ?

```ts
import type { GridApi } from '../path/to/api/gridApi'

**Comment by DanailH** [comment]
> > Unfortunately due to circular dependency, the api can only be typed at the place where the assignment happens, in useGridApiEventHandler.
> 
> @DanailH even with the following ?
> 
> ```ts
> import type { GridApi } from '../path/to/api/gridApi'
> ```

Yes

File: packages/grid/_modules_/grid/hooks/root/useApi.ts [hunk 1]
```diff
@@ -23,7 +23,7 @@ export function useApi(apiRef: GridApiRef): void {
       if (event && isSynthenticEvent(event) && event.isPropagationStopped()) {
         return;
       }
-      apiRef.current.emit(name, params, event);
+      apiRef.current.emit(name, params, event, {});
```
m4theushw, line 26: Why the empty object? Don't we want to pass the API to XGrid's users that are subscribing to the events? We could use the same `MuiCallbackOptions`.

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
```diff
@@ -2,6 +2,26 @@ import * as React from 'react';
...
+import { optionsSelector } from '../utils/optionsSelector';
+import { GridApi } from '../../models/api/gridApi';
+
+// TODO: Remove once [[GridApi]] cycle dependency is fixed
+/**
+ * Callback options.
+ */
+export interface MuiCallbackOptions {
+  api?: GridApi;
+  reason?: string;
```
m4theushw, line 15: The reason is not used yet. I would add it later.

```suggestion
```

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -136,7 +136,7 @@ export interface GridOptions {
    * Callback fired when the EditRowModel changes.
    * @param {GridEditRowsModel} editRowsModel With all properties from [[GridEditRowsModel]].
```
m4theushw, line 137:
```suggestion
   * @param {GridEditRowsModel} editRowsModel With all properties from [[GridEditRowsModel]].
   * @param {MuiCallbackOptions} options Additional options for this callback.
```

x other props.

m4theushw (commented):
We also need to update http://localhost:3001/components/data-grid/events/#subscribing-to-events.

I wonder if users will not get confused with this third argument. There's already "params", now "options" too. To get worse, `subscribeEvent` has an `options` argument to control if the event should fire first. The alternative would be to pass the API directly as the third argument and move the reason to the params.

oliviertassinari:
> Part 1 of 2 from #1821

I'm confused by the PR's description. #1821 is about the DataGrid only, and yet, we change the XGrid API here. Could it be because we have worked on #1301 here? 

> I wonder if users will not get confused with this third argument. There's already "params", now "options" too. To get worse, subscribeEvent has an options argument to control if the event should fire first. The alternative would be to pass the API directly as the third argument and move the reason to the params.

From what I understand, we have three proposals so far:

- Option 1, proposed by @DanailH 
  - `onCellBlur(params: GridCellParams, event: MuiEvent<React.SyntheticEvent>, options?: { api: any })`
  - `onPageChange?: (page: number, options?: { api: any, reason: string }) => void;`
- Option 2, proposed by @m4theushw 
  - `onCellBlur(params: GridCellParams, event: MuiEvent<React.SyntheticEvent>, api: any)`
  - `onPageChange?: (page: number, api: any) => void;`
- Option 3, proposed by @oliviertassinari
  - `onCellBlur(event: MuiEvent<React.SyntheticEvent>, params: GridCellParams)`
  - `onPageChange?: (event: MuiEvent<React.SyntheticEvent>, page: number, params?: { api: any, reason: string }) => void;`
  
  To be closer to the core's API, e.g. with the [Autocomplete](https://github.com/mui-org/material-ui/blob/a0cccf538eab4ea50b0601e7296ac583f16004f9/packages/material-ui/src/useAutocomplete/useAutocomplete.d.ts#L255-L268).
  
I agree with Matheus, IMHO 1 is KO because "params" and "options" are synonymous.

packages/grid/_modules_/grid/hooks/root/useApi.ts
DanailH: It can be added but after the latest comments, it should be `api` directly.

DanailH:
@oliviertassinari As far as I understood the problem described in #1821 is twofold:
1. Remove `api` from each callback
2. Make sure `api` is still available only for `XGrid`

What this PR does is handle 2. from the list.

Regarding the approach - I made it as an object so that later it will be easier to add `reason` if we choose to and because as far as I remember this was what we agreed on last Monday when we were discussing the ticket.

If we want to go with the second option from your list that is also ok. I just want to agree on this before doing the change.

packages/grid/_modules_/grid/hooks/root/useApi.ts
DanailH: Updated

DanailH:
Ok, I've updated the PR with what We've discussed yesterday. I also fixed some of the existing docs in the `gridOptions` and updated the `/data-grid/events/#subscribing-to-events` section in the docs

packages/grid/_modules_/grid/GridComponentProps.ts
```diff
@@ -67,11 +67,15 @@ export interface GridComponentProps extends GridOptionsProp {
   /**
    * Set a callback fired when the state of the grid is updated.
    */
-  onStateChange?: (params: GridStateChangeParams, event: MuiEvent<{}>) => void; // We are overriding the handler in GridOptions to fix the params type and avoid the cycle dependency
+  onStateChange?: (params: GridStateChangeParams, event: MuiEvent<{}>, options: any) => void; // We are overriding the handler in GridOptions to fix the params type and avoid the cycle dependency
```
m4theushw, line 70:
```suggestion
  onStateChange?: (params: GridStateChangeParams, event: MuiEvent<{}>, details: any) => void; // We are overriding the handler in GridOptions to fix the params type and avoid the cycle dependency
```

packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
```diff
@@ -67,7 +75,7 @@ export function useGridControlState(apiRef: GridApiRef) {
         },
       };
     },
-    [apiRef],
+    [apiRef, props],
```
m4theushw, line 78:
```suggestion
    [apiRef, props.signature],
```

packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts
```diff
@@ -41,8 +43,14 @@ export function useGridControlState(apiRef: GridApiRef) {
         if (hasSubStateChanged) {
           if (controlState.propOnChange) {
             const newModel = newSubState;
+            const details =
+              props.signature === Signature.XGrid
+                ? {
+                    api: apiRef.current,
+                  }
+                : {};
```
m4theushw, line 51: Just a small detail.

```suggestion
            const details = props.signature === Signature.XGrid ? { api: apiRef.current } : {};
```

The same in `useApi`.

packages/grid/_modules_/grid/hooks/root/useApi.ts
```diff
@@ -5,13 +5,17 @@ import { useLogger } from '../utils/useLogger';
...
 import { MuiEvent } from '../../models/gridOptions';
+import { Signature } from './useGridApiEventHandler';
+import { useGridSelector } from '../features/core/useGridSelector';
+import { optionsSelector } from '../utils/optionsSelector';
 
 const isSynthenticEvent = (event: any): event is React.SyntheticEvent => {
   return event.isPropagationStopped !== undefined;
 };
 
 export function useApi(apiRef: GridApiRef): void {
```
m4theushw, line 16:
```suggestion
export function useApi(apiRef: GridApiRef, { signature }: Pick<GridComponentProps, 'signature'>): void {
```

packages/grid/_modules_/grid/hooks/root/useApi.ts
```diff
@@ -5,13 +5,17 @@ import { useLogger } from '../utils/useLogger';
...
   const logger = useLogger('useApi');
+  const { signature } = useGridSelector(apiRef, optionsSelector);
```
m4theushw, line 18: We're moving away from the options in #2213.

```suggestion
```

packages/grid/_modules_/grid/hooks/root/useApi.ts
DanailH: I'll keep it as `props` based on this discussion https://github.com/mui-org/material-ui-x/pull/2275#discussion_r682581815

docs/src/pages/components/data-grid/events/events.md
```diff
@@ -8,7 +8,7 @@ title: Data Grid - Events
 
 ## Subscribing to events
 
-You can subscribe to one of the [events emitted](/components/data-grid/events/#catalog-of-events) by calling `apiRef.current.subscribeEvent()` with the name of the event and a handler. The handler will be called with two arguments: an object with information related to the event and a `MuiEvent` containing the DOM event or the React synthetic event, when available.
+You can subscribe to one of the [events emitted](/components/data-grid/events/#catalog-of-events) by calling `apiRef.current.subscribeEvent()` with the name of the event and a handler. The handler will be called with three arguments: an object with information related to the event, a `MuiEvent` containing the DOM event or the React synthetic event, when available and a `MuiCallbackDetails` containing the `GridApi` if `XGrid` is being used.
```
m4theushw, line 11: I think two items is the limit to separate with commas.

```suggestion
You can subscribe to one of the [events emitted](/components/data-grid/events/#catalog-of-events) by calling `apiRef.current.subscribeEvent()` with the name of the event and a handler. The handler will be called with three arguments:

1. an object with information related to the event
2. a `MuiEvent` containing the DOM event or the React synthetic event, when available
3. a `MuiCallbackDetails` containing the `GridApi` only if `XGrid` is being used.
```

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
```diff
@@ -10,22 +29,28 @@ export function useGridApiEventHandler(
   options?: { isFirst?: boolean },
 ) {
   const logger = useLogger('useGridApiEventHandler');
+  const { signature } = useGridSelector(apiRef, optionsSelector);
```
m4theushw, line 32: We could get it from the `GridRootPropsContext` context, but this context is only inserted after the hook.

m4theushw (approved):

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
DanailH: Updated.

flaviendelangle: Am I missing something or is the context provider instantiated in the `GridComponent` which is also the component which calls most of the `useGridApiEventHandler` ?

If so, it will always be undefined even after initialization since its not in a child of the provider.

DanailH: Am, not sure but I changed it and tested it and it seems to work.

flaviendelangle: It should be undefined in every call from a hook that is inside `GridComponent` which is problematic.

@m4theushw even with #2176 it will be the same.
We could quite easily move the the props preprocessing + context provider up if we want to use it in the feature.
It would make sense for me.

DanailH: Well it does work as is but sure.

DanailH: Updated

flaviendelangle: Where does the value come from then ?

m4theushw: We don't need this line:

```suggestion
```

The details actually is set in https://github.com/mui-org/material-ui-x/pull/2236/files#diff-940e97227abe24d94ec31abb3fe7f6af369e3fcc9711995fefeca5876e737ee1R28 and passed to the handler.

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
```diff
@@ -2,6 +2,25 @@ import * as React from 'react';
...
 import { useLogger } from '../utils/useLogger';
+import { useGridSelector } from '../features/core/useGridSelector';
+import { optionsSelector } from '../utils/optionsSelector';
+import { GridApi } from '../../models/api/gridApi';
+
+// TODO: Remove once [[GridApi]] cycle dependency is fixed
+/**
+ * Callback details.
+ */
+export interface MuiCallbackDetails {
```
oliviertassinari, line 13: This type is specific to the data grid, shouldn't it be?

```suggestion
export interface GridCallbackDetails {
```

packages/grid/_modules_/grid/hooks/root/useGridApiEventHandler.ts
oliviertassinari: I have unresolved the suggestion, the changes in this file are not needed. It's strange, why do we try to override the behavior of `subscribeEvent`? what happened for the feedback to be missed?

PR #1421, opened 2021-04-15
[DataGrid] refactoring on focus cell and cols

dtassone:
- fix #1416
- Help with #1430
- fix #1493
- fix wrong initial tab index. it should be on the first header cell, not the first body cell.
- fix #1420

Preview: https://deploy-preview-1421--material-ui-x.netlify.app/components/data-grid/editing/#edit-using-external-button

PR #2239, opened 2021-07-30
[DataGrid] Drop autoFocus

m4theushw:
We've been using `autoFocus` to set the focus when a cell enters the edit mode. The problem with this approach is that in #2098 multiple cells will be in edit mode at the same time, so using this prop on multiple elements is wrong. This PR modifies the current behavior to use the state to control which cell should receive focus.

Preview: https://deploy-preview-2239--material-ui-x.netlify.app/components/data-grid/demo/


oliviertassinari (commented):
Note that iOS is supposed to not support the `autoFocus` attribute: https://caniuse.com/?search=autofocus. 

We have a regression on the keyboard navigation with the row checkbox selection feature. The last version that seems to behave correctly is v4.0.0-alpha.30: https://codesandbox.io/s/material-demo-forked-v9cqw. it's the input that is supposed to be focused when on the checkbox selection column, the arrow keys are supposed to work too. The correct behavior was introduced in #1421, but without any tests. I think an example of what we shouldn't do: bundle 5 fixes, add no tests.

flaviendelangle:
Should we document the focus in this [custom edit component demo](https://deploy-preview-2239--material-ui-x.netlify.app/components/data-grid/editing/#custom-edit-component) ?

m4theushw:
> Should we document the focus in this [custom edit component demo](https://deploy-preview-2239--material-ui-x.netlify.app/components/data-grid/editing/#custom-edit-component) ?

@flaviendelangle I prefer to keep it as it is. But in the row editing yes will we have to explain better how to create custom edit components.

m4theushw:
> Note that iOS is supposed to not support the `autoFocus` attribute: https://caniuse.com/?search=autofocus.
> 
> We have a regression on the keyboard navigation with the row checkbox selection feature. The last version that seems to behave correctly is v4.0.0-alpha.30: https://codesandbox.io/s/material-demo-forked-v9cqw. it's the input that is supposed to be focused when on the checkbox selection column, the arrow keys are supposed to work too. The correct behavior was introduced in #1421, but without any tests. I think an example of what we shouldn't do: bundle 5 fixes, add no tests.

@oliviertassinari I fixed this in 20ef450. I had the same problem with the checkbox not checking, but in the row editing. I didn't think we would have this problem here, however we had. The solution is a bit hacky.

m4theushw:
@mui-org/x Could someone review this one? It would be nice to check if there're no regressions. It's the only thing blocking #2098, which is already done.

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -145,13 +145,17 @@ export const GridCell = React.memo(function GridCell(props: GridCellProps) {
...
+    }
+
+    const doc = ownerDocument(apiRef!.current.rootElementRef!.current as HTMLElement)!;
 
-    if (
-      hasFocus &&
-      cellRef.current &&
-      (!doc.activeElement || !cellRef.current!.contains(doc.activeElement))
-    ) {
+    if (cellRef.current && (doc.activeElement || !cellRef.current!.contains(doc.activeElement))) {
```
oliviertassinari, line 158: AFAIK we will never see doc.activeElement empty, unless on IE https://github.com/sveltejs/sapper/issues/332#issue-346813546

```suggestion
    if (cellRef.current && !cellRef.current!.contains(doc.activeElement!)) {
```

oliviertassinari (commented):
Nothing blocking, only me explore any possible opportunities.

packages/grid/_modules_/grid/components/cell/GridEditBooleanCell.tsx
```diff
@@ -41,11 +43,17 @@ export function GridEditBooleanCell(
     setValueState(value);
   }, [value]);
 
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
m4theushw: I updated all input components. The remaining uses of ` React.useLayoutEffect` are for another PR.

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -172,17 +190,24 @@ export const useGridFocus = (apiRef: GridApiRef, props: Pick<GridComponentProps,
 
   React.useEffect(() => {
     const doc = ownerDocument(apiRef.current.rootElementRef!.current as HTMLElement);
-    doc.addEventListener('click', handleDocumentClick, true);
+    const handleDocumentClickAsync = (event) => {
+      // If the state is changed during the capture phase, onChange will not be fired
+      // By delaying the state update, there is time for other callbacks to be called first
+      // See https://github.com/facebook/react/issues/13424
```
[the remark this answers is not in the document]
m4theushw: > We rarely use event listeners in the capture phase in the core components. I'm always suspicious when we use them. So, I tried without, I couldn't help myself. The CI is green, the demos seems to work. What am I missing?

True, we don't need to add the listener in the capture phase now. Probably I did this because the code from this PR is actually from #2098, which at the time it was opened #2179 was not merged yet, so stopping the propagation in a `onCellClick` would also break the focus. 

> Off-topic. I personally find the current approach a bit backward. We set the focus state, then the state calls .focus().

The tricky part for me is the Portal support. At the beginning we were doing this approach but we inverted.

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -145,13 +145,17 @@ export const GridCell = React.memo(function GridCell(props: GridCellProps) {
   };
 
   React.useLayoutEffect(() => {
-    const doc = ownerDocument(apiRef!.current.rootElementRef!.current as HTMLElement);
+    if (!hasFocus) {
+      return;
+    }
+
+    if (cellMode === 'edit' && isEditable) {
```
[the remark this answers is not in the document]
m4theushw: Updated. Answering your question, if `apiRef.current.setCellMode` is called in a cell that cannot be edited it will turn to edit mode. We need to add a check there.

packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx
```diff
@@ -33,9 +36,15 @@ export function GridEditInputCell(props: GridCellParams & InputBaseProps) {
     setValueState(value);
   }, [value]);
 
+  useEnhancedEffect(() => {
+    if (hasFocus) {
+      inputRef.current!.focus();
```
flaviendelangle, line 41: Are we sure that `hasFocus` is false 100% of the time on the 1st render since it's a `useLayoutEffect` ?

packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
```diff
@@ -90,47 +102,52 @@ export const useGridFocus = (apiRef: GridApiRef, props: Pick<GridComponentProps,
...
         GRID_CELL_FOCUS_OUT,
-        apiRef.current.getCellParams(cell.id, cell.field),
+        apiRef.current.getCellParams(focusedCell.id, focusedCell.field),
         event,
       );
+
+      if (cellParams) {
+        apiRef.current.setCellFocus(cellParams.id, cellParams.field);
+      } else {
+        setGridState((previousState) => ({
```
flaviendelangle, line 143: To follow https://github.com/mui-org/material-ui-x/pull/2249

```suggestion
        setGridState((state) => ({
```

packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx
m4theushw: I didn't understad. `hasFocus` could be false or true in the first render.

flaviendelangle: My bad
I thought that in `useLayoutEffect` first run, the ref was not yet initialized
But I tested, and it is

oliviertassinari: > I thought that in useLayoutEffect first run, the ref was not yet initialized

@flaviendelangle ref runs with a higher priority than use layout effects :)

oliviertassinari (commented):
Looks great

flaviendelangle (approved):

Issue #1412, opened 2021-04-13
[DataGrid] Allow to control column visibility


tusharchutani:
I have a requirment to show or hide a column based on click of button. On the click of the button I am setting the hide property to true or false of the columns (which is in a state).

I have set up a CodeSandbox to show case this: https://codesandbox.io/s/material-demo-forked-l004j?file=/demo.tsx

I am not sure if this is supported yet if it is not is there a work around this?

m4theushw:
The [columns definitions](https://material-ui.com/components/data-grid/columns/#column-definitions) should never change after the first mount. Storing them in a state and updating it has no effect. The XGrid provides the method `apiRef.current.updateColumn` to update the columns at any time. I've updated your example showing how to use this API.

oliviertassinari:
Should we close? I'm not sure how we could leverage this feedback in any way to improve the docs or the product.

Well, unless we want to freeze the provided props, but it might be a bit extreme?

m4theushw:
A warning when the `columns` prop changes between renders sounds enough. What do you think?

tusharchutani:
@m4theushw the apiRef functionality is not available on DataGrid right (only on XGrid) right?

m4theushw:
`apiRef` is only available in XGrid. Storing the columns in a state variable is not recommended because each time it's updated, the reference changes, which causes another rerender.

oliviertassinari:
> A warning when the columns prop changes between renders sounds enough. What do you think?

@m4theushw We have allowed developers to change the `columns` prop, sure it will reset a few states, but this is working. If in the future, they even control the state, they could mitigate the side-effects, I personally think that it's fine.

IMHO the issue with the author is that he mutates he data structure, this is wrong in the first place, it breaks react principles

oliviertassinari:
@tusharchutani please don't edit the original reproduction with the proposed fix. It's makes it hard for us to find root cause solutions.

dtassone:
ATM the hide attribute is tightly binded to the column definition model so the only way to hide a column programmatically would be by using the react state and resetting the column prop of the Data-Grid component, as below (copied from codesandbox)

``` tsx
export default function ApiRefRowsGrid() {
  const [columns, setColumns] = useState(defaultColumns);

  const change = () => {
    defaultColumns[2].hide = !columns[2].hide;
    setColumns([...defaultColumns]);
  };
  return (
    <div style={{ height: 400, width: "100%" }}>
[...]

PR #2241, opened 2021-07-30
[core] Split `useGridColumns` into `useGridColumns` and `useGridVisibleColumns`

flaviendelangle:
Part of #1412 

Preliminary work for https://github.com/mui-org/material-ui-x/issues/1412#issuecomment-860935241

____

@mui-org/x I have two options for the file structure:

- set `useGridVisibleColumns` in its own folder with its own selector file. In that case all selectors relying on visible columns must be in this folder (the PR takes this approach)
- set `useGridVisibleColumns` in the same folder as `useGridColumns`, they share the same selector file. In that case we could let generic behaviors like `getColumnPosition` in `useGridColumns`

EDIT : I went for the 2nd one as its a lot lighter in term of modifications.


DanailH:
> @mui-org/x I have two options for the file structure:
> 
> * set `useGridVisibleColumns` in its own folder with its own selector file. In that case all selectors relying on visible columns must be in this folder (the PR takes this approach)
> * set `useGridVisibleColumns` in the same folder as `useGridColumns`, they share the same selector file. In that case we could let generic behaviors like `getColumnPosition` in `useGridColumns`

I think we can go with option 2. We have other cases where we keep multiple feature hooks in the same folder as they operate in the same underlying domain, in this case, the columns logic. If it makes sense we can create a separate selectors file if there are cases where a specific selector is needed for either hook?

packages/grid/_modules_/grid/components/containers/GridRoot.tsx
```diff
@@ -6,7 +6,7 @@ import { GridRootPropsContext } from '../../context/GridRootPropsContext';
 import { optionsSelector } from '../../hooks/utils/optionsSelector';
 import { GridRootContainerRef } from '../../models/gridRootContainerRef';
 import { useStyles } from './GridRootStyles';
-import { visibleGridColumnsLengthSelector } from '../../hooks/features/columns/gridColumnsSelector';
+import { visibleGridColumnsLengthSelector } from '../../hooks/features/columnVisibility/gridVisibleColumnsSelector';
```
DanailH, line 9: As you suggested it makes more sense to keep the new hook under `/columns` as well.

On another note. I was about to ask why did we name the selector `visibleGridColumnsLengthSelector` but quick git blame shows that I am responsible for this 😅. Now that I think about it to keep it consistent with the other selectors maybe we should rename it to `gridVisibleColumnsLengthSelector`?

flaviendelangle: :smile: yes no problem for renaming it

If we keep both under the same folder, would you still move the methods like `getColumnsMeta`, `getColumnPosition` in `useGridVisibleColumns`

Basically, should `useGridVisibleColumns` handle everything that is related to the visible columns or just the hide / show  behavior ?

packages/grid/_modules_/grid/hooks/features/columnVisibility/gridVisibleColumnsSelector.ts
```diff
@@ -0,0 +1,32 @@
...
+    }, 0);
+
+    return { totalWidth, positions };
+  },
+);
+
+export const visibleGridColumnsLengthSelector = createSelector<GridState, GridColumns, number>(
+  visibleGridColumnsSelector,
+  (visibleColumns: GridColumns) => visibleColumns.length,
+);
```
DanailH, line 28:
```suggestion
);

```

packages/grid/_modules_/grid/hooks/features/columnVisibility/useGridVisibleColumns.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  const getVisibleColumns = React.useCallback(
+    () => visibleGridColumnsSelector(apiRef.current.getState()),
+    [apiRef],
+  );
+
+  const getColumnsMeta: () => GridColumnsMeta = React.useCallback(() => columnsMeta, [columnsMeta]);
+
+  const getColumnIndex = React.useCallback(
```
DanailH, line 48: That one is interesting. Not sure if it makes more sense to be here or in the `useGridColumns`?

packages/grid/_modules_/grid/hooks/features/columnVisibility/useGridVisibleColumns.ts
```diff
@@ -0,0 +1,75 @@
...
+    (field: string, useVisibleColumns: boolean = true): number =>
+      useVisibleColumns
+        ? visibleColumns.findIndex((col) => col.field === field)
+        : allColumns.findIndex((col) => col.field === field),
+    [allColumns, visibleColumns],
+  );
+
+  const getColumnPosition: (field: string) => number = React.useCallback(
```
DanailH, line 56: Same as the `getColumnIndex`

packages/grid/_modules_/grid/hooks/features/columnVisibility/useGridVisibleColumns.ts
flaviendelangle: When splitting in two folders I had to move it due to circular imports (the selector from each file importing the selector from the other file)

But if we keep a single folder (ie an single selector file), the question is worth asking.

https://github.com/mui-org/material-ui-x/pull/2241#discussion_r679795569

packages/grid/_modules_/grid/components/containers/GridRoot.tsx
DanailH: Mmm, these methods sound like something that is generic for all columns regardless if it is visible or not. I would probably keep them in the `useGridColumns` but let's see what the others have to say about it. I don't really have a strong personal preference.

oliviertassinari:
> @mui-org/x I have two options for the file structure:

If we envision the column visibility as a feature on its own, then `/hooks/features/columnVisibility/`, like column resizing, column rendering. I don't think that there is a debate do have around it.
If we don't envision this as different features, I would personally keep everything under `/hooks/features/columns/`, but we can split too.

I would raise the same question regarding column width, does it need to be controlled?

---

Do we need to work on #1412 now? From what I understand, it's not required to get the stable release happen nor to get one of the most requires features happen.

packages/grid/x-grid/src/useXGridComponent.tsx
```diff
@@ -4,6 +4,7 @@ import { useGridColumnMenu } from '../../_modules_/grid/hooks/features/columnMen
 import { useGridColumnReorder } from '../../_modules_/grid/hooks/features/columnReorder/useGridColumnReorder';
 import { useGridColumnResize } from '../../_modules_/grid/hooks/features/columnResize/useGridColumnResize';
 import { useGridColumns } from '../../_modules_/grid/hooks/features/columns/useGridColumns';
+import { useGridVisibleColumns } from '../../_modules_/grid/hooks/features/columns/useGridVisibleColumns';
```
oliviertassinari, line 7: How about we only allow one top-level hook feature per folder? Either the feature can be used on its own and

```suggestion
import { useGridVisibleColumns } from '../../_modules_/grid/hooks/features/columnVisibility/useGridVisibleColumns';
```

or it can't it should already be part of `useGridColumns.ts`

```suggestion
```

oliviertassinari (changes requested):

packages/grid/x-grid/src/useXGridComponent.tsx
flaviendelangle: You mean be a part of `useGridColumns.ts` like `usGrideVirtualColumns` is called from `useGridVirtualRows` ?

flaviendelangle:
@oliviertassinari I updated the PR with a proposal to split big hooks without multiplying the top level hooks called from the outside.

DanailH (approved):

packages/grid/_modules_/grid/models/api/gridColumnApi.ts
```diff
@@ -68,3 +57,22 @@ export interface GridColumnApi {
...
+   * @param {string} field The column to change visibility.
+   * @param {boolean} isVisible Pass `true` to show the column, or `false` to hide it.
+   */
+  setColumnVisibility: (field: string, isVisible: boolean) => void;
+}
+
+/**
+ * The column API interface that is available in the grid [[apiRef]].
+ */
+export type GridColumnApi = GridColumnUpdateApi & GridColumnInformationApi & GridVisibleColumnApi;
```
oliviertassinari, line 78: Prefer using an interface, it creates a flat structure, not a union. TypeScript processes it faster
```suggestion
export interface GridColumnApi extends GridColumnUpdateApi, GridColumnInformationApi, GridVisibleColumnApi;
```

packages/grid/_modules_/grid/hooks/features/columns/useGridColumnInformation.ts
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { GridApiRef } from '../../../models/api/gridApiRef';
+import { GridColumnInformationApi } from '../../../models/api/gridColumnApi';
+
```
oliviertassinari, line 4:
```suggestion
```

oliviertassinari (commented):
I personally think that useGridColumns.ts is still a small file (less than 500 LOCs), I would have let it grown to 2k before considering breaking it down. But no objections. My only hope is that it will help refactoring

packages/grid/_modules_/grid/models/api/gridColumnApi.ts
```diff
@@ -1,27 +1,19 @@
...
-/**
- * The column API interface that is available in the grid [[apiRef]].
- */
-export interface GridColumnApi {
+export interface GridColumnInformationApi {
   /**
-   * Returns the [[GridColDef]] for the given `field`.
+   * Returns the [[GridStateColDef]] for the given `field`.
    * @param {string} field The column field.
    * @returns {{GridStateColDef}} The [[GridStateColDef]].
```
m4theushw, line 7:
```suggestion
   * @returns {GridStateColDef} The [[GridStateColDef]].
```

packages/grid/_modules_/grid/hooks/features/columns/useGridVisibleColumns.ts
```diff
@@ -0,0 +1,52 @@
...
+        colDef: updatedCol,
+        api: apiRef,
+        isVisible,
+      });
+    },
+    [apiRef, forceUpdate],
+  );
+
+  const getVisibleColumns = React.useCallback(
+    () => visibleGridColumnsSelector(apiRef.current.getState()),
```
m4theushw, line 36:
```suggestion
    () => visibleGridColumnsSelector(apiRef.current.state),
```

flaviendelangle:
The more I move forward on this PR the less convinced I am
I think that a `useGridVisibleColumns` would be nice when we have controlled visible columns.
But to do so I have to make to many ugly things due to a lack of clear separation between what's in `useGridVisibleColumns` and `useGridColumns`

I'm closing this PR until I have a better idea or if the file continue to grow :+1:

PR #2242, opened 2021-07-30
[DataGrid] Fix renderCell returning false-ish values

### LINKED ISSUE #1849: [DataGrid] Fix renderCell false-ish value rendering

## Current Behavior 😯

Using `formattedValue` in a custom `renderCell` will render nothing on a boolean typed cell.

## Expected Behavior 🤔

Should render a ❌ or ✔ based on the value. Other value formatters are working correctly (date, dateTime). See example.

## Steps to Reproduce 🕹

https://codesandbox.io/s/bitter-forest-sd9tp?file=/src/index.tsx

Steps:

1. Add custom `renderCell` function to column definition which renders via `params.formattedValue`
2. Add boolean type to column definition
3. Add data

## Context 🔦

I want to change the rendering of a given cell, but want to use the default formatting in it.

Also tried to use the "old" `renderCell` function of the given column but this also renders nothing.

## Your Environment 🌎

<details>
  
```
  System:
    OS: macOS 11.3.1
  Binaries:
    Node: 14.17.0 - /usr/local/opt/node@14/bin/node
    npm: 6.14.13 - /usr/local/opt/node@14/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Edge: Not Found
    Firefox: Not Found
    Safari: 14.1
  npmPackages:
    @material-ui/lab: ^4.0.0-alpha.58 => 4.0.0-alpha.58 
    @material-ui/styles:  4.11.4 
    @material-ui/system:  4.11.3 
    @material-ui/types:  5.1.0 
    @material-ui/utils:  4.11.2 
    @material-ui/x-grid: ^4.0.0-alpha.30 => 4.0.0-alpha.30 
    @types/react: ^17.0.6 => 17.0.6 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    styled-components: ^5.3.0 => 5.3.0 
    typescript: ^4.2.4 => 4.2.4 

```

Tested on latest chrome and firefox dev
</details>

siriwatknp:
closes #1849

- fix render `0` from renderCell. (`renderCell: () => 0` now show 0 on the screen)
- add default formatter to string for column type boolean. (`renderCell: params => params.formattedValue` now show `'true'` | `'false'` on the screen.
- add tests

### DIFF
--- a/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
+++ b/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
@@ -4,6 +4,14 @@ import { renderBooleanCell } from '../../components/cell/GridBooleanCell';
 import { renderEditBooleanCell } from '../../components/cell/GridEditBooleanCell';
 import { gridNumberComparer } from '../../utils/sortingUtils';
 import { getGridBooleanOperators } from './gridBooleanOperators';
+import { GridValueFormatterParams } from '../params/gridCellParams';
+
+function gridBooleanFormatter({ value }: GridValueFormatterParams) {
+  if (typeof value === 'boolean') {
+    return value.toString();
+  }
+  return value;
+}
 
 export const GRID_BOOLEAN_COL_DEF: GridColTypeDef = {
   ...GRID_STRING_COL_DEF,
@@ -14,4 +22,5 @@ export const GRID_BOOLEAN_COL_DEF: GridColTypeDef = {
   renderEditCell: renderEditBooleanCell,
   sortComparator: gridNumberComparer,
   filterOperators: getGridBooleanOperators(),
+  valueFormatter: gridBooleanFormatter,
 };
--- a/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
+++ b/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
@@ -4,14 +4,6 @@ import { renderBooleanCell } from '../../components/cell/GridBooleanCell';
 import { renderEditBooleanCell } from '../../components/cell/GridEditBooleanCell';
 import { gridNumberComparer } from '../../utils/sortingUtils';
 import { getGridBooleanOperators } from './gridBooleanOperators';
-import { GridValueFormatterParams } from '../params/gridCellParams';
-
-function gridBooleanFormatter({ value }: GridValueFormatterParams) {
-  if (typeof value === 'boolean') {
-    return value.toString();
-  }
-  return value;
-}
 
 export const GRID_BOOLEAN_COL_DEF: GridColTypeDef = {
   ...GRID_STRING_COL_DEF,
@@ -22,5 +14,4 @@ export const GRID_BOOLEAN_COL_DEF: GridColTypeDef = {
   renderEditCell: renderEditBooleanCell,
   sortComparator: gridNumberComparer,
   filterOperators: getGridBooleanOperators(),
-  valueFormatter: gridBooleanFormatter,
 };
--- a/packages/grid/data-grid/src/tests/cells.DataGrid.test.tsx
+++ b/packages/grid/data-grid/src/tests/cells.DataGrid.test.tsx
@@ -63,4 +63,66 @@ describe('<DataGrid /> - Cells', () => {
     );
     expect(getCell(0, 0)).to.have.class('foobar');
   });
+
+  describe('type `boolean`', () => {
+    it('able to render number from provided `renderCell`', () => {
+      render(
+        <div style={{ width: 300, height: 500 }}>
+          <DataGrid
+            autoHeight={isJSDOM}
+            columns={[
+              {
+                field: 'number',
+                headerName: 'Number',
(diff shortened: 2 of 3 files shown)


packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
```diff
@@ -4,6 +4,14 @@ import { renderBooleanCell } from '../../components/cell/GridBooleanCell';
 import { renderEditBooleanCell } from '../../components/cell/GridEditBooleanCell';
 import { gridNumberComparer } from '../../utils/sortingUtils';
 import { getGridBooleanOperators } from './gridBooleanOperators';
+import { GridValueFormatterParams } from '../params/gridCellParams';
+
+function gridBooleanFormatter({ value }: GridValueFormatterParams) {
```
DanailH, line 9: I think for now we can keep that one here. If in future we need it in more than 1 place we can move to the `/utils`.

DanailH:
🙏 for adding tests.

DanailH (approved):

flaviendelangle (approved):

packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
```diff
@@ -4,6 +4,14 @@ import { renderBooleanCell } from '../../components/cell/GridBooleanCell';
...
+  if (typeof value === 'boolean') {
+    return value.toString();
```
m4theushw, line 11: This value should be translated. We already have `booleanCellTrueLabel` and `booleanCellFalseLabel` that are used to set the title of the icon.

https://github.com/mui-org/material-ui-x/blob/363a7a1cff3f4cb1dba20bb17264acf7d80d89f6/packages/grid/_modules_/grid/components/cell/GridBooleanCell.tsx#L31

m4theushw (changes requested):

packages/grid/data-grid/src/tests/cells.DataGrid.test.tsx
```diff
@@ -63,4 +63,66 @@ describe('<DataGrid /> - Cells', () => {
...
+              },
+            ]}
+          />
+        </div>,
+      );
+      expect(getCell(0, 0)).to.have.text('0');
+      expect(getCell(1, 0)).to.have.text('1');
+    });
+
+    it('formattedValue to string by default', () => {
```
m4theushw, line 98: I think this test could be moved to this file: 

https://github.com/mui-org/material-ui-x/blob/363a7a1cff3f4cb1dba20bb17264acf7d80d89f6/packages/grid/x-grid/src/tests/export.XGrid.test.tsx#L94

There, the value formatter is used by default and we don't need to provide a `renderCell`.

packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
siriwatknp: Is it a separate thing? the `GridBooleanCell` works as expected if no `renderCell` is provided, this fix is for custom `renderCell`.

packages/grid/data-grid/src/tests/cells.DataGrid.test.tsx
siriwatknp: Oh, but the fix is basic feature, not just for x-grid. should I move it?

packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
m4theushw: `GridBooleanCell` is not working quite well. When the data is exported to CSV, the `renderCell` is not rendered but the formatted value is used instead. Since there was no default formatter before this PR, it would cast the boolean to string, generating "true" or "false". My point is that if we're going to add a default value formatter, it should translate the resulting value according to the user's locale. That being said, the value should be "True" or "False", when the English locale is used. Also, we don't need to check if `typeof value` is boolean, any value that evaluates to true is fine.

siriwatknp: Okay, that's a good insight.

packages/grid/data-grid/src/tests/cells.DataGrid.test.tsx
m4theushw: Yeah, I think we can move it. This file is for tests that use the imperative API, but, since this fix is for both DataGrid and XGrid and we don't have another file to use it, this one is fine.

packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
oliviertassinari: If it's simpler. We can undo the change of the value formater. Fixing renderCell to correctly process react nodes is already great.

m4theushw:
I went one step back here and removed the value formatter. We can add it back in another PR with support for translations too.

oliviertassinari (approved):

PR #2244, opened 2021-07-30
[core] Use type inference in selectors

flaviendelangle:
No behavior change
I'm trying to reduce the amount of imports between the main parts of the codebase and letting TypeScript infer the types helps a lot.

___

```ts
export const scrollStateSelector = createSelector<
  GridState,
  InternalRenderingState,
  GridScrollParams
>(renderStateSelector, (renderingState) => renderingState.realScroll);
```

Is equivalent to 

```ts
export const scrollStateSelector = createSelector(
  renderStateSelector,
  (renderingState) => renderingState.realScroll,
);
```

We only have to type the root selector param

```ts
export const renderStateSelector = (state: GridState) => state.rendering;
```

Sometime we even had the double type explicit definition

```ts
export const selectedGridRowsCountSelector: OutputSelector<
  GridState,
  number,
  (res: GridSelectionState) => number
> = createSelector<GridState, GridSelectionState, number>(
  gridSelectionStateSelector,
  (selection) => selection.length,
);
```

Which is equivalent to

```ts
export const selectedGridRowsCountSelector = createSelector(
  gridSelectionStateSelector,
  (selection) => selection.length,
);
```


oliviertassinari:
It seems fair. On a related topic, we try to systematically type the return value of the functions to speed up TypeScript inference.

oliviertassinari:
Also, it would feel safer to do the change if we turned https://github.com/mui-org/material-ui-x/blob/6899f66057236becc823959d70344076fdfc38fe/tsconfig.json#L17-L18

flaviendelangle:
> Also, it would feel safer to do the change if we turned
> 
> https://github.com/mui-org/material-ui-x/blob/6899f66057236becc823959d70344076fdfc38fe/tsconfig.json#L17-L18

@oliviertassinari I'm 100% in favor of doing this rule. I tried a bit but the event from the core often don't have the properties we use.

With that being said, the current situation for the selector is not safe either, as part of them have generics and other don't.
I'm not sure waiting for the `noImplicitAny`  is worth.

flaviendelangle:
> It seems fair. On a related topic, we have tried so far to systematically type the return value of the functions to speed up TypeScript inference.

@oliviertassinari If you have resources about what's slowing up TypeScript type inference
Because I suppose it's not getting the type of a function like `(state: GridState) => state.pagination`
I don't think the selectors are where the true gain is done.

Maybe more when generating an object like in https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L39 where saying what's return really helps.

oliviertassinari:
> resources about what's slowing up TypeScript type inference

@flaviendelangle Sebastian taught me this https://github.com/microsoft/TypeScript/wiki/Performance#using-type-annotations.

Regarding these changes, I personally don't really care, happy no matter what the outcome of the PR is.

flaviendelangle:
Thanks for the link
I propose we move forward with selectors then :+1:

PR #2246, opened 2021-07-31
[DataGrid] Fix filter with extended columns

### LINKED ISSUE #2225: [DataGrid] Using filterModel to filter extended columns throws an error
The demo is displaying an error message when viewing it in Opera GX and Google Chrome.

Extended filter operators:
Error: Material-UI: No filter operators found for column 'totalPrice'.

https://material-ui.com/components/data-grid/filtering/#extend-filter-operators

## Steps to Reproduce 🕹

Steps:

1.
2.
3.
4.

## Your Environment 🌎

<details>
  
```
```
</details>

m4theushw:
Fixes #2225 
Preview: https://deploy-preview-2246--material-ui-x.netlify.app/components/data-grid/filtering/#extend-filter-operators

The fix is the same idea of #2213: use the props instead of the options.

The bug started to happen because in #1720 the filter hook was changed to get the `filterModel` from the props, instead of the options:

https://github.com/mui-org/material-ui-x/blob/4f52c84e104dd598e454e5d8d0c395c5aa3b7d54/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts#L323-L326

However, the hydration of the extended columns was still using the options:

https://github.com/mui-org/material-ui-x/blob/4f52c84e104dd598e454e5d8d0c395c5aa3b7d54/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L304-L306

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -88,7 +88,7 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number): GridCo
 
 function hydrateColumns(
   columns: GridColumns,
-  columnTypes: GridColumnTypesRecord = {},
+  columnTypes: GridColumnTypesRecord,
   withCheckboxSelection: boolean,
   logger: Logger,
   getLocaleText: <T extends GridTranslationKeys>(key: T) => GridLocaleText[T],
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -22,7 +22,6 @@ import { Logger } from '../../../models/logger';
 import { GridColumnOrderChangeParams } from '../../../models/params/gridColumnOrderChangeParams';
 import { mergeGridColTypes } from '../../../utils/mergeUtils';
 import { useGridApiMethod } from '../../root/useGridApiMethod';
-import { optionsSelector } from '../../utils/optionsSelector';
 import { useLogger } from '../../utils/useLogger';
 import { useGridSelector } from '../core/useGridSelector';
 import { GridLocaleText, GridTranslationKeys } from '../../../models/api/gridLocaleTextApi';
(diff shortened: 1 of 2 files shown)


packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -318,8 +319,8 @@ export function useGridColumns(
     logger,
     apiRef,
     props.columns,
-    options.columnTypes,
-    options.checkboxSelection,
+    props.columnTypes,
```
oliviertassinari, line 322: **Off-topic**. Would it make sense to use the "rootProps" terminology for these props and systematically?

oliviertassinari:
For those that wonder why the CI didn't break when we have introduced this regression, it's because https://github.com/mui-org/material-ui-x/blob/6899f66057236becc823959d70344076fdfc38fe/test/regressions/index.js#L21

which is normal. We had about 45 days between the introduction of the regression and the feedback from the developers. We could try using Sentry out if we want to shorten this feedback loop. I personally think that it's as good as it can get n the current tradeoff/bandwidth we have available.

oliviertassinari:
This line looks strange:

https://github.com/mui-org/material-ui-x/blob/6899f66057236becc823959d70344076fdfc38fe/packages/grid/_modules_/grid/models/gridOptions.tsx#L531

should we remove it? To me, it doesn't make any sense. `columnTypes` is meant to extend the default column type, why is it not empty by default?

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -89,7 +88,7 @@ function updateColumnsWidth(columns: GridColumns, viewportWidth: number): GridCo
 
 function hydrateColumns(
   columns: GridColumns,
-  columnTypes: GridColumnTypesRecord,
+  columnTypes: GridColumnTypesRecord = {},
```
flaviendelangle, line 91: I don't think the prop can ever be undefined
We have a default prop.

m4theushw: Reverted.

m4theushw:
> This line looks strange:
> 
> https://github.com/mui-org/material-ui-x/blob/6899f66057236becc823959d70344076fdfc38fe/packages/grid/_modules_/grid/models/gridOptions.tsx#L531
> 
> should we remove it? To me, it doesn't make any sense. `columnTypes` is meant to extend the default column type, why is it not empty by default?

@oliviertassinari Yes, we can remove it in another PR. Since we started to merge with the default columns in https://github.com/mui-org/material-ui-x/pull/950/files#diff-9a22a60c7d67b9dce7363291e06bc6d3845431c01d4035ff22e15366adb8b43dR61, there's no more need of this line.

flaviendelangle (approved):

oliviertassinari:
> Yes, we can remove it in another PR

Yeah fair enough. The logic was already here.

PR #2247, opened 2021-07-31
[XGrid] Document virtualization APIs in virtualization section

### LINKED ISSUE #1103: [XGrid] Document how to use the scroll feature

## Summary 💡

This feature is already implemented but the documentation doesn't cover it. From the eye of a developer, it's as if we had never implemented it. I propose we add a section for it. 

## Examples 🌈

- https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--scroll-to-row
- https://www.ag-grid.com/documentation/react/grid-api/#reference-scrolling

I'm not sure we need an actual demo for the feature. I think that the main problem is that there is no way to use this feature without diving deep into the source. 

## Motivation 🔦

One use case example: "as a developer, I'm streaming updates to the rows in the data grid. When an update is pushed, I would like to scroll the data grid to make it visible."

ZeeshanTamboli:
One chunk of #1103 

We already auto generate document of scroll API here in GridApi: https://material-ui.com/api/data-grid/grid-api/. But as per discussion it should also be documented as part of virtualization like filtering apiRefs https://material-ui.com/components/data-grid/filtering/#apiref for ease of access.

Follow conversation from https://github.com/mui-org/material-ui-x/issues/1103#issuecomment-885695797

Preview: https://deploy-preview-2247--material-ui-x.netlify.app/components/data-grid/virtualization/#apiref


docs/pages/api-docs/data-grid/grid-virtualization-api.json
```diff
@@ -0,0 +1,26 @@
...
+  "properties": [
+    {
+      "name": "scroll",
+      "description": "Triggers the viewport to scroll to the given positions (in pixels).",
+      "type": "(params: Partial<GridScrollParams>) => void"
+    },
+    {
+      "name": "scrollToIndexes",
+      "description": "Triggers the viewport to scroll to the cell at indexes given by <code>params</code>. Returns <code>true</code> if the grid had to scroll to reach the target.",
+      "type": "(params: Optional<GridCellIndexCoordinates, 'rowIndex'>) => boolean"
```
ZeeshanTamboli, line 13: I notice `Optional` is a custom utility type we created:
https://github.com/mui-org/material-ui-x/blob/4f52c84e104dd598e454e5d8d0c395c5aa3b7d54/packages/grid/_modules_/grid/utils/utils.ts#L67

And we removed exporting from `@material-ui/data-grid` in #2233 . How do we document this now? 

Should we do:
```diff
- (params: Optional<GridCellIndexCoordinates, 'rowIndex'>) => boolean
+ (params: Pick<Partial<GridCellIndexCoordinates>, 'rowIndex'> & Omit<GridCellIndexCoordinates, 'rowIndex'>) => boolean
```

It's in other places in docs as well.

oliviertassinari: We can update the type, partial is enough

oliviertassinari (approved):

oliviertassinari:
@ZeeshanTamboli A good step forward with the docs 👍

PR #2269, opened 2021-08-01
[DataGrid] Add Sudanese Arabic (arSD) locale

YassinHussein:
add Arabic Localization

### DIFF
--- a/packages/grid/_modules_/grid/locales/arSD.ts
+++ b/packages/grid/_modules_/grid/locales/arSD.ts
@@ -1,3 +1,4 @@
+// TODO v5: import from arSD
 import { enUS as enUSCore } from '@material-ui/core/locale';
 import { GridLocaleText } from '../models/api/gridLocaleTextApi';
 import { getGridLocalization, Localization } from '../utils/getGridLocalization';
--- a/packages/grid/_modules_/grid/locales/arSD.ts
+++ b/packages/grid/_modules_/grid/locales/arSD.ts
@@ -1,0 +1,110 @@
+import { arSD as arSDCore } from '@material-ui/core/locale';
+import { GridLocaleText } from '../models/api/gridLocaleTextApi';
+import { getGridLocalization, Localization } from '../utils/getGridLocalization';
+
+const arSDGrid: Partial<GridLocaleText> = {
+  // Root
+  noRowsLabel: 'لا توجد صفوف',
+  noResultsOverlayLabel: 'لم يتم العثور على نتائج.',
+  errorOverlayDefaultLabel: 'حدث خطأ.',
+
+  // Density selector toolbar button text
+  toolbarDensity: 'كثافة',
+  toolbarDensityLabel: 'كثافة',
+  toolbarDensityCompact: 'مدمج',
+  toolbarDensityStandard: 'المعيار',
+  toolbarDensityComfortable: 'مريح',
+
+  // Columns selector toolbar button text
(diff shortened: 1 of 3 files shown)


flaviendelangle:
Thanks for your help !

https://app.circleci.com/pipelines/github/mui-org/material-ui-x/6515/workflows/b13f8c6d-c8d4-41a4-8bd4-980e11ef19f9/jobs/34895

You need to run `yarn prettier` to reformat your code

https://app.circleci.com/pipelines/github/mui-org/material-ui-x/6515/workflows/b13f8c6d-c8d4-41a4-8bd4-980e11ef19f9/jobs/34891

The core do not have the `arSD` locale. 
@m4theushw do we agree that we need the core translations in order to add the grid one
Or do we accept to use the english for the core ?

packages/grid/_modules_/grid/locales/arSD.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  // Checkbox selection text
+  checkboxSelectionHeaderName: 'تحديد',
+
+  // Boolean cell text
+  booleanCellTrueLabel: 'صحيح',
+  booleanCellFalseLabel: 'خاطئة',
+
+  // Used core components translation keys
+  MuiTablePagination: {},
```
m4theushw, line 107:
```suggestion
```

packages/grid/_modules_/grid/locales/arSD.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+
+export const arSD: Localization = getGridLocalization(arSDGrid, arSDCore);
```
m4theushw, line 110: This locale doesn't exist in the [core](https://material-ui.com/guides/localization/#supported-locales) yet. Do you want to open another PR there to add the translations for the pagination or is it fine to use the English as default?

```suggestion
export const arSD: Localization = getGridLocalization(arSDGrid, enUSCore);
```

packages/grid/_modules_/grid/locales/arSD.ts
YassinHussein: updated

packages/grid/_modules_/grid/locales/arSD.ts
YassinHussein: updated

oliviertassinari:
The change in the core won't be accessible to us until we drop v4 support, looking at #27588, we did a plain copy and paste of the arEG locale… so we could do the same here

m4theushw:
> The change in the core won't be accessible to us until we drop v4 support. Looking at [mui-org/material-ui#27588](https://github.com/mui-org/material-ui/pull/27588), we did a plain copy and paste of the arEG locale… so we could do the same here

@oliviertassinari My idea would be to extend the core arEG or enUS locale, but use arSD if available.

packages/grid/_modules_/grid/locales/arSD.ts
```diff
@@ -0,0 +1,105 @@
+import { enUS as enUSCore } from '@material-ui/core/locale';
```
oliviertassinari, line 1:
```suggestion
// TODO v5: import from arSD
import { enUS as enUSCore } from '@material-ui/core/locale';
```

oliviertassinari (approved):
> My idea would be to extend the core arEG or enUS locale, but use arSD if available.

@m4theushw This logic would be great, but AFAIK it's not possible to implement. We still need to retrain the tree-shaking property.

PR #2271, opened 2021-08-01
[core] rebaseWhen=auto does not seem to work

oliviertassinari:
As soon as one renovate PR is merged, all the other ones are rerun. It's not scaling with our CI. Try to force the value.


oliviertassinari:
doesn't wor

Issue #2231, opened 2021-07-29
[RFC] Improve the Grid state management


flaviendelangle:
- [x] Replace the options with direct prop reads #2292 
- [x] Each feature hook should initialize synchronously its state #2293

flaviendelangle:
# 1. Remove the `options` pattern in favor of a direct prop read

## Goals

1. Avoid re-renders caused by props-to-state patterns

2. Avoid having synchronization issues between the values from the `options` and the values from the `props` inside hooks and components

When setting a prop, you have a render with the new prop but the old option. So if some parts of the code uses the options but other uses the props it causes bugs.

## Proposal

We can the options by a preprocessing of the props that accomplish the same things as `useOptionsProp` but synchronously

```ts
const DEFAULT_GRID_PROPS = {
  columnBuffer: 2,
  columnTypes: getGridDefaultColumnTypes(),
  density: GridDensityTypes.Standard,
  filterMode: GridFeatureModeConstant.client,
  headerHeight: 56,
  paginationMode: GridFeatureModeConstant.client,
  rowHeight: 52,
  rowsPerPageOptions: [25, 50, 100],
  scrollEndThreshold: 80,
  sortingMode: GridFeatureModeConstant.client,
  sortingOrder: ['asc' as const, 'desc' as const, null],
  logger: console,
  logLevel: process.env.NODE_ENV === 'production' ? 'error' : 'warn',
};

const useGridPropsWithDefaultValues = (inProps: GridComponentInputProps) => {
  const localeText = React.useMemo(() => ({ 
    ...inProps.localeText, 
    ...GRID_DEFAULT_LOCALE_TEXT 
  }), [inProps.localeText])

  const classes = React.useMemo(() => composeClasses(
    {
      root: [GRID_ROOT_CSS_CLASS_SUFFIX],
      columnHeader: [GRID_COLUMN_HEADER_CSS_CLASS_SUFFIX],
      row: [GRID_ROW_CSS_CLASS_SUFFIX],
      cell: [GRID_CELL_CSS_CLASS_SUFFIX],
    },
    getDataGridUtilityClass,
    inProps.classes as Record<string, string>,
  ), [inProps.classes])

  const props = React.useMemo(() => ({
    ...GRID_DEFAULT_PROPS,
    ...inProps,
    localeText,
    classes,
  })
}


// DataGrid.tsx

const DataGrid = (inProps: DataGridProps) => {
[...]

PR #2275, opened 2021-08-02
[core] Reduce options usage in feature hooks

flaviendelangle:
I'm doing this migration in several steps to avoid large PRs :+1: 

- [x] Stop using `options` in `useGridKeyboardNavigation` 
- [x] Stop using `options` in `useGridEditRows` (needed to run the `useEffect` that update the state from the prop after the control registration, its done that way in the other hooks)
- [x] Stop using `options` in `useGridRows` 
- [x] Stop using `options` in `useGridComponents` 
- [x] Stop using `options` in `useErrorHandler` 
- [x] Stop using `options` in `useLogger`
 
I'm removing the prop decomposition in the parameters to follow https://github.com/mui-org/material-ui-x/issues/6
And because it make it obvious that we are accessing the props and not the state for cases where both make sense.

Part of #2231


packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -63,9 +63,11 @@ const getNextColumnHeaderIndexes = (key: string, indexes: GridColumnHeaderIndexC
   }
 };
 
-export const useGridKeyboardNavigation = (apiRef: GridApiRef): void => {
+export const useGridKeyboardNavigation = (
+  apiRef: GridApiRef,
+  props: Pick<GridComponentProps, 'pagination'>,
```
DanailH, line 68: Should we do it like this when there are only one or two props used?
```suggestion
  { pagination }: Pick<GridComponentProps, 'pagination'>,
```

packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
```diff
@@ -83,14 +86,14 @@ export const useGridRows = (
...
-        getRowId,
+        props.rows,
+        props.rowCount,
+        props.getRowId,
       );
 
       return { ...state, rows: internalRowsState.current };
     });
-  }, [getRowId, rows, setGridState]);
+  }, [props.getRowId, props.rows, props.rowCount, setGridState]);
```
DanailH, line 96: I think you can have it as just `props` here since you are depending on all 3.

```suggestion
  }, [props, setGridState]);
```

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
flaviendelangle: > I'm removing the prop decomposition in the parameters to follow #6
And because it make it obvious that we are accessing the props and not the state for cases where both make sense.

I like the clarity to have every prop usage being very explicit.

packages/grid/_modules_/grid/hooks/utils/useErrorHandler.ts
```diff
@@ -3,8 +3,9 @@ import { GRID_COMPONENT_ERROR } from '../../constants';
 import { GridApiRef } from '../../models/api';
 import { useGridState } from '../features/core/useGridState';
 import { useGridApiEventHandler } from '../root/useGridApiEventHandler';
+import { GridComponentProps } from '../../GridComponentProps';
 
-export function useErrorHandler(apiRef: GridApiRef, props) {
+export function useErrorHandler(apiRef: GridApiRef, props: Pick<GridComponentProps, 'error'>) {
```
DanailH, line 8:
```suggestion
export function useErrorHandler(apiRef: GridApiRef, { error }: Pick<GridComponentProps, 'error'>) {
```

packages/grid/_modules_/grid/hooks/utils/useLogger.ts
```diff
@@ -49,19 +50,21 @@ let factory: LoggerFactoryFn | null;
 
 export function useLoggerFactory(
   apiRef: any,
-  { logger, logLevel }: { logger?: Logger; logLevel?: string | false },
+  props: Pick<GridComponentProps, 'logger' | 'logLevel'>,
```
DanailH, line 53: Maybe we can have it as a rule if we all agree - if there are more than 2 then have it as `props`?

```suggestion
 { logger, logLevel }: Pick<GridComponentProps, 'logger' | 'logLevel'>,
```

packages/grid/_modules_/grid/hooks/utils/useErrorHandler.ts
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2275#discussion_r682585408

packages/grid/_modules_/grid/hooks/utils/useLogger.ts
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2275#discussion_r682585408

I agree that we should agree on a common rule :+1:

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
DanailH: Ok, I'm ok with this but then let's just follow the rule that regardless we always have it as `props`, because there are places where you kept the prop decomp.

flaviendelangle: In which places ? I probably made a mistake

oliviertassinari: +1 for avoiding destructuring when we can. I'm also annoyed by the selectors that makes it hard to know when refactoring what's used where.

In the core we do theme.palette.primary.main with the full path systematically to make it easier to work on the theme's structure.

DanailH: @flaviendelangle here for example https://github.com/mui-org/material-ui-x/blob/3ca9645d7df64a44c51df69e78ff578e5a677c69/packages/grid/_modules_/grid/hooks/features/useGridComponents.tsx#L67

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -341,4 +343,12 @@ export function useGridEditRows(
       changeEvent: GRID_EDIT_ROWS_MODEL_CHANGE,
     });
   }, [apiRef, props.editRowsModel, props.onEditRowsModelChange]);
+
+  React.useEffect(() => {
+    const oldEditRowsModel = apiRef.current.getState().editRows;
```
oliviertassinari, line 348: How about we reverse the dichotomy. What we call old is the current value, 

```suggestion
    const currentEditRowsModel = apiRef.current.getState().editRows;
```

oliviertassinari (commented):
Look like a good step forward

PR #2276, opened 2021-08-02
[DataGrid] Don't break the page if `valueOptions` is missing from the `ColDef`

### LINKED ISSUE #2199: [DataGrid] Specifying a filterModel prop when using singleSelect column throws fatal error
In the DataGrid component, setting `filterModel` in combination with using a `singleSelect` typed column type causes the filter form to crash when the user tries to change the filter to a column with `type = 'singleSelect'`

## Current Behavior 😯

When setting the `filterModel`, the initial table filtering functions as expected, but when the user tries to change the filtered column to a column that is `singleSelect`, the table fails with a fatal error.

## Expected Behavior 🤔

Setting a `filterModel` prop with a `singleSelect` column should function similarly to all of the other column type (`string`, `number`, `date`, etc).

## Steps to Reproduce 🕹

Forked from the `singleSelect` demo from the MaterialUI docs:
https://codesandbox.io/s/material-demo-forked-078r2?file=/demo.js

Steps:

1. Using the three-dot menu or the filter icon, open the filter options dropdown
2. Change the "Column" field to the `country` field (which is a `singleSelect` column)
3. DataGrid throws an error ('t is not iterable or t is undefined' depending on browser).

## Context 🔦

I was trying to create a table containing multiple column filters with `singleSelect` and have the table retain its `filterModel` when changing between different views.

## Your Environment 🌎
 Using Chrome: 92.0.4515.107, also reproduced the error on FireFox.
<details>

```
  System:
    OS: macOS 11.4
  Binaries:
    Node: 14.12.0 - ~/.nvm/versions/node/v14.12.0/bin/node
    Yarn: Not Found
    npm: 6.14.8 - ~/.nvm/versions/node/v14.12.0/bin/npm
  Browsers:
    Chrome: 92.0.4515.107
    Edge: Not Found
    Firefox: 87.0
    Safari: 14.1.1
  npmPackages:
    @material-ui/core: ^4.11.3 => 4.11.3
    @material-ui/data-grid: ^4.0.0-alpha.34 => 4.0.0-alpha.34
    @material-ui/styles:  4.11.3
    @material-ui/system:  4.11.3
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @types/react:  17.0.3
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
```
</details>

DanailH:
Fixes #2199


m4theushw (commented):
We can keep the fallback when the `valueOptions` is not set, but we missed the root cause of this bug. If the `filterModel` prop is used it must not update the state if the prop didn't change, however it does:

TrvMEZRSGN

The `GridFilterForm` keeps an internal state, so when the column is changed it updates its state but the prop value continues the same. That's the reason why it crashes with the `singleSelect`.

https://github.com/mui-org/material-ui-x/blob/9f336333ef1e537c9ac56b81defa3634f294642a/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx#L100-L103

I noticed this problem in https://github.com/mui-org/material-ui-x/pull/1909#pullrequestreview-703446864 too.

DanailH:
@m2mathew I noticed that as well but I want able to reliably reproduce the issue. If you try to set the `filterModel` in another demo, like the basic filter demo on the `/filtering` page then it works fine (or at least it was working fine when I tested it).

m4theushw:
The problem is that you shouldn't be able to change the column or operator when the `filterModel` prop is set but `onFilterModelChange` doesn't. It's keeping an internal state, so this state is updated no matter the prop is set or not. We should get rid of the internal state.

oliviertassinari:
@m4theushw Oh yeah nice. These two should have never been here:

https://github.com/mui-org/material-ui-x/blob/6ce7f0395a904c1d950dbe31cdae8d0570e688f7/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx#L78

https://github.com/mui-org/material-ui-x/blob/6ce7f0395a904c1d950dbe31cdae8d0570e688f7/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx#L84

DanailH:
@m2mathew Ok I think I understood the problem. I made some updated.

But still, if you look at this example https://deploy-preview-2276--material-ui-x.netlify.app/components/data-grid/columns/#column-types (I'll revert that change)
and this one https://deploy-preview-2276--material-ui-x.netlify.app/components/data-grid/filtering/#basic-filter they should behave the same, no? The only difference is that the second one uses the `useDemoData` hook.

oliviertassinari:
> they should behave the same

@DanailH I don't think so. I think that https://github.com/mui-org/material-ui-x/pull/2276#pullrequestreview-721535195 draw the correct expected behavior; One demo has its filter model set, another one hasn't, how the data is feed into it, is not relevant, AFAIK.

DanailH:
> > they should behave the same
> 
> @DanailH I don't think so. I think that [#2276 (review)](https://github.com/mui-org/material-ui-x/pull/2276#pullrequestreview-721535195) draws the correct expected behavior. One demo has its filter model set, another one hasn't. How the data is feed into it (useDemoData or not) is not relevant, AFAIK.

No, I reverted the change in the demo. Both had the filter model set but only one was behaving like explained in https://github.com/mui-org/material-ui-x/pull/2276#pullrequestreview-721535195 and the only difference was that the misbehaving one uses the `useDemoData` hook. Removing that hook and passing data directly soves the problem.

I had a chat yesterday with @m4theushw about this. It seems that there is a problem with the `useDemoData`. That can be explored in a separate PR.

oliviertassinari:
@DanailH Oh ok, I was looking at the Netlify preview, but if the setup is different, then ignore my previous comment. I guess a reproduction is required at this point to continue the conversation.

m4theushw:
After rebasing it onto HEAD both examples from https://github.com/mui-org/material-ui-x/pull/2276#issuecomment-893303561 behaved equally and correctly, that means, the prop was always respected. I suppose #2208 fixed something that we weren't aware of.

oliviertassinari:
Ok so the next steps are?

1. Rebase on HEAD
2. Revert the changes in the demos
3. Keep the removal of the harmful/useless useStates
4. Add a test case to make we cover the pain in the issue

And in parallel, we can see if we can reproduce the other strange behavior?

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -90,15 +91,26 @@ export function GridFilterForm(props: GridFilterFormProps) {
       currentColumn.filterOperators?.find((operator) => operator.value === item.operatorValue) ||
       null
     );
-  });
+  }, [item, getCurrentColumn]);
+
+  const CurrentOperatorInputComponent = React.useCallback(() => {
```
m4theushw, line 96: Since it creates a different component each time a dependency is changed, the focus is lost after the debounce. I would revert this part.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -67,6 +69,7 @@ export function GridFilterInputValue(props: GridTypeFilterInputValueProps & Text
   React.useEffect(() => {
     return () => {
       clearTimeout(filterTimeout.current);
+      setIsApplying(false);
```
m4theushw, line 72:
```suggestion
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -58,7 +61,6 @@ export function GridFilterInputValue(props: GridTypeFilterInputValueProps & Text
       setIsApplying(true);
       filterTimeout.current = setTimeout(() => {
         applyValue({ ...item, value });
-        setIsApplying(false);
```
m4theushw, line 61: I was wondering how it kept working after removing this.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
DanailH: I was playing with this idea yesterday. The reason why it didn't lose focus before is that the `currentOperator` was kept in the state. and now every time the value changes the `getCurrentOperator()` is called again which results in input loosing focus.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
DanailH: This needs to be there because there is no more local state in the `GridFilterForm`. Before the `return` was only called once, now its every time the value changes.

m4theushw:
I think we can keep the previous version, we had removed the internal state to relay on the grid's state. The problem was just it creating a new component on each render. I pushed 1b626e9 to make it work without losing focus, which doesn't occur now. Could you review?

m4theushw (approved):

PR #2278, opened 2021-08-03
[docs] Sort events alphabetically

flaviendelangle:
I'm trying to gather the events into a single enum without loosing the documentation (see  #2227 )
The events will be sorting alphabetically, I do this PR to sort them alphabetically with the current approach to help compare the output.

PR #2279, opened 2021-08-03
[DataGrid] Group events into a single enum

flaviendelangle:
Part of #2227 

With #2278 there is no change in the documentation output.

Several benefits here :

- Improve the clarity of the public API by gathering dozens of string variables
- Simplify the imports in our codebase
- Help the future typing of `publishEvent` / `subscribeEvent` (they will be able to import the enum instead of having floating strings as keys

____

Do we want the user to use directly the string (like in https://github.com/mui-org/material-ui-x/blob/master/docs/src/pages/components/data-grid/events/SubscribeToEvents.tsx#L16) or do we want them to import the constant (like in https://github.com/mui-org/material-ui-x/blob/master/docs/src/pages/components/data-grid/editing/CatchEditingEventsGrid.tsx#L25)

If we want him to import the enum, could we type  `publishEvent` / `subscribeEvent` first param accordingly ?


packages/grid/_modules_/grid/hooks/root/useEvents.ts
```diff
@@ -99,7 +77,7 @@ export function useEvents(
   useNativeEventListener(
     apiRef,
     apiRef.current.rootElementRef!,
-    GRID_KEYDOWN,
-    getHandler(GRID_KEYDOWN),
+    GridEvents.keydown,
```
oliviertassinari, line 80: It's a native event handler, it doesn't make sense to use the grid events, AFAIK

```suggestion
    'keydown',
```

packages/grid/_modules_/grid/constants/eventsConstants.ts
```diff
@@ -1,443 +1,376 @@
...
- * Fired when a column visibility changes. Called with a [[GridColumnVisibilityChangeParams]] object.
- * @event
- */
-export const GRID_COLUMN_VISIBILITY_CHANGE = 'columnVisibilityChange';
+export enum GridEvents {
+  /**
+   * Fired when the grid is resized. Called with a [[GridResizeParams]] object.
+   */
+  resize = 'resize',
+
```
oliviertassinari, line 6: For consistency, we can remove all these blank lines

```suggestion
```

x the other

oliviertassinari (commented):
It looks like a great idea. I have merged HEAD into te PR to leverage #2278, showing the events.json has no diffs.

docs/src/pages/components/data-grid/editing/CatchEditingEventsGrid.tsx
```diff
@@ -22,7 +21,7 @@ export default function CatchEditingEventsGrid() {
 
   React.useEffect(() => {
     return apiRef.current.subscribeEvent(
-      GRID_CELL_EDIT_START,
+      GridEvents.cellEditStart,
```
m4theushw, line 24: We should be using the string value here. 

```suggestion
      'cellEditStart',
```

The same for `GridEvents.cellEditStop` below.

packages/grid/_modules_/grid/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -82,7 +82,7 @@ export const useGridClipboard = (apiRef: GridApiRef): void => {
     [apiRef],
   );
 
-  useGridApiEventHandler(apiRef, GRID_KEYDOWN, handleKeydown);
+  useGridApiEventHandler(apiRef, GridEvents.keydown, handleKeydown);
```
m4theushw, line 85: I'm fine with not adding native events to the enum.

```suggestion
  useGridApiEventHandler(apiRef, 'keydown', handleKeydown);
```

This `keydown` is only used in this hook and here:

https://github.com/mui-org/material-ui-x/blob/09b936fd34d9c36b92b24fbdacd6cc4b0d45186c/packages/grid/_modules_/grid/hooks/root/useEvents.ts#L99-L104

oliviertassinari: AFAIK, it's not a native event here, not as in https://github.com/mui-org/material-ui-x/pull/2279#discussion_r684542859

m4theushw: This event handler is weird. We listen for the native event to publish a grid's event. For the scroll this doesn't happen (example below), the native listener is added in the same hook where it's used. I propose we don't add the native event to the enum and in a second iteration we move the `useNativeEventListener` to the clipboard hook.

https://github.com/mui-org/material-ui-x/blob/09b936fd34d9c36b92b24fbdacd6cc4b0d45186c/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts#L387

flaviendelangle: @m4theushw which events would you remove from the enum ? 
"keydown" or are there others ?

m4theushw: Yeah, just remove "keydown". It's already marked as ignored so no problem removing it.

flaviendelangle: But their is a listener on it

```ts
useGridApiEventHandler(apiRef, GridEvents.keydown, handleKeydown);
```

You would replace it like this ?

```ts
useGridApiEventHandler(apiRef, 'keydown', handleKeydown);
```

m4theushw: Yes, I would leave `'keydown'`. The scroll listener is also hardcoded.

flaviendelangle: For the scroll it's using`useNativeEventListener` not `useGridApiEventListener`

```ts
  useNativeEventListener(apiRef, windowRef, 'scroll', handleScroll, { passive: true });
```

m4theushw: > By the way, we never publishes publishEvent(GridEvents.keydown, ...) or publishEvent('keydown', ...

That's the problem I was mentioning in https://github.com/mui-org/material-ui-x/pull/2279#discussion_r685487183. We add the native listener to publish a grid's event.

> Is this useGridApiEventHandler(apiRef, GridEvents.keydown, handleKeydown); even called ?

It's used for the clipboard feature.

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2279/commits/ee602b614b38d300aad2b2d4971eade5fe03208b

I'm now using directly `useNativeEventListener` in the clipboard feature to totally remove the event from the grid events structure :+1:

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
```diff
@@ -112,7 +107,7 @@ export const useGridKeyboard = (apiRef: GridApiRef): void => {
       }
 
       if (isNavigationKey(event.key) && !isSpaceKey(event.key) && !event.shiftKey) {
-        apiRef.current.publishEvent(GRID_COLUMN_HEADER_NAVIGATION_KEY_DOWN, params, event);
+        apiRef.current.publishEvent(GridEvents.cellNavigationKeyDown, params, event);
```
oliviertassinari, line 110: It should have been:

```suggestion
        apiRef.current.publishEvent(GridEvents.columnHeaderNavigationKeyDown, params, event);
```

The keyboard navigation is no longer working on the header https://codesandbox.io/s/material-demo-forked-git2x

PR #2281, opened 2021-08-04
[DataGrid] Drop support of v4 below `4.12.0`

DanailH:
Found some errors being thrown in the console on some pages in the docs. This PR aims to fix those.

The main one is the deprecated `createMuiTheme` and since the docs use the latest version of the core this results in errors in the console.

The other one is an error related to having `key` being passed to the `GridToolbarFilterButton`.

---

###  Breaking changes

Drop support of v4 below `"@material-ui/core": "4.12.0"`


docs/src/pages/components/data-grid/style/StylingRowsGrid.tsx
```diff
@@ -1,14 +1,14 @@
 import * as React from 'react';
 import { DataGrid } from '@material-ui/data-grid';
 import { useDemoData } from '@material-ui/x-grid-data-generator';
-import { createMuiTheme, darken, lighten, Theme } from '@material-ui/core/styles';
+import { createTheme, darken, lighten, Theme } from '@material-ui/core/styles';
```
oliviertassinari, line 4: If we do this change, we need to change https://github.com/mui-org/material-ui-x/blob/30512081208da6b47922950206bc72cf18254c24/packages/grid/data-grid/package.json#L48

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
```diff
@@ -76,9 +76,9 @@ export const GridToolbarFilterButton = React.forwardRef<
       <div>
         {apiRef!.current.getLocaleText('toolbarFiltersTooltipActive')(counter)}
         <ul className={classes.list}>
-          {activeFilters.map((item) => ({
+          {activeFilters.map((item, index) => ({
             ...(lookup[item.columnField!] && (
-              <li key={item.id}>
+              <li key={index}>
```
oliviertassinari, line 81: +1 for reverting this change that Damien did recently.

docs/src/pages/components/data-grid/columns/ColumnTypesGrid.js
```diff
@@ -35,32 +35,31 @@ const rows = [
   },
 ];
 
+const columns = [
```
oliviertassinari, line 38: Why this change?

DanailH: I'll rever that, its from another PR.

docs/src/pages/components/data-grid/style/StylingRowsGrid.tsx
DanailH: I just noticed the problem and since the docs use the latest version I thought it is safe to make the change.

oliviertassinari: Ok, then I would label this as a BC: drop support of v4 < 4.12.0, and update the peer dependency scope.

DanailH: Updated

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
DanailH: So keep that change right?

packages/grid/data-grid/package.json
```diff
@@ -45,8 +45,8 @@
     "reselect": "^4.0.0"
   },
   "peerDependencies": {
-    "@material-ui/core": "^4.9.12 || ^5.0.0-beta.0",
-    "@material-ui/styles": "^4.9.10 || ^5.0.0-beta.0",
+    "@material-ui/core": "^4.12.0 || ^5.0.0-beta.0",
+    "@material-ui/styles": "^4.12.0 || ^5.0.0-beta.0",
```
m4theushw, line 49: The lowest @material-ui/styles required by the core v4 is 4.12.0.

https://github.com/mui-org/material-ui/blob/53a1655143aa4ec36c29a6063ccdf89c48a74bfd/packages/material-ui/package.json#L53

```suggestion
    "@material-ui/styles": "^4.11.4 || ^5.0.0-beta.0",
```

oliviertassinari:
It looks great. I think that we can keep the logic that supports < v4.12.0 as long as the overhead for us is low, it seems to be the case, so far.

packages/grid/_modules_/grid/components/toolbar/GridToolbarFilterButton.tsx
oliviertassinari: Yeap

PR #2282, opened 2021-08-04
[DataGrid] Allow non-integer column width for flex columns

flaviendelangle:
Part of #2011

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -73,8 +73,7 @@ function getStateColumns(
 
   // Compute the width of flex columns
   if (totalFlexUnits && widthToAllocateInFlex > 0) {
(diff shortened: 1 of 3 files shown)


oliviertassinari (dismissed):
I think that we miss a snapshot and a functional test case, you can find them in https://github.com/mui-org/material-ui-x/pull/2110/files

> Part of #2011

Should it be "Fix #2011" instead?

flaviendelangle:
@oliviertassinari it does not fix entirely the issue
See https://github.com/mui-org/material-ui-x/issues/2011#issuecomment-873189677

I'll add the tests

DanailH (approved):

flaviendelangle:
I have to fix the test because Safari does not round the value in the same way

flaviendelangle:
@oliviertassinari how can I update the snapshots and merge ? 
The diff is normal here

image

oliviertassinari:
Argos seems to have bugged

PR #2285, opened 2021-08-05
[DataGrid] Add German (deDE) translation for export and isEmpty operator

ChristopherBussick:
Add the missing deDE locale values for exporting and using the "is empty" and "is not empty" operators.


m4theushw (approved):
Thanks. 👍

Issue #1562, opened 2021-05-04
Checkbox column appended instead of prepended in concurrent mode


eps1lon:
If you can point me to the code that's responsible for inserting the column I can take a look to understand what's going wrong if you want.

## Current Behavior 😯

When using `<DataGrid checkboxSelection />` the column containing the checkboxes is the last column in concurrent react.

![screenshot of the current behavior](https://user-images.githubusercontent.com/12292047/116969261-436a5700-acb6-11eb-97cb-c8df5b276428.png)

## Expected Behavior 🤔

It should be the first column

![screenshot of the expected behavior](https://user-images.githubusercontent.com/12292047/116969237-38172b80-acb6-11eb-97af-4b8ec74b3331.png)

## Steps to Reproduce 🕹

https://codesandbox.io/s/datagrid-legacy-vs-concurrent-root-84z9o?file=/demo.js

Steps:

1. Use `react@experimental` (currently `0.0.0-experimental-79740da4c`)
2. render `<DataGrid rows={rows} columns={columns} pageSize={5} checkboxSelection />`

## Context 🔦

Reproducible in our docs: https://60906bee09db3300076bbe72--material-ui.netlify.app/components/tables/#DataTable (https://github.com/mui-org/material-ui/pull/26107)

## Your Environment 🌎

See codesandbox. For the docs see the details below.
<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
  Binaries:
    Node: 12.20.0 - ~/.nvm/versions/node/v12.20.0/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.20.0/bin/npm
  Browsers:
    Chrome: 90.0.4430.72
    Firefox: 88.0
  npmPackages:
    @emotion/react: ^11.0.0 => 11.1.5 
    @emotion/styled: ^11.0.0 => 11.3.0 
    @material-ui/codemod:  5.0.0-alpha.30 
    @material-ui/core:  5.0.0-alpha.32 
    @material-ui/data-grid:  4.0.0-alpha.27 
    @material-ui/docs:  5.0.0-alpha.31 
    @material-ui/envinfo:  1.1.6 
    @material-ui/icons:  5.0.0-alpha.29 
    @material-ui/lab:  5.0.0-alpha.32 
[...]

PR #2286, opened 2021-08-05
[core] Avoid useless columns preprocessing on Grid mount

flaviendelangle:


oliviertassinari:
It seems to fix #1562, proof https://codesandbox.io/s/datagrid-legacy-vs-concurrent-root-forked-4x35b?file=/demo.js

flaviendelangle:
I was not the main goal but nice :smile:

flaviendelangle:
@oliviertassinari if #2344 has the same behavior, it is indeed simpler and I'm OK to close this one

oliviertassinari:
@flaviendelangle I'm not exactly sure it has the same behavior. I looked at it because I was suspicious we could have a simpler logic. I don't plan to push #2344 much further. If you continue the exploration, and pick whoever works best, I think that it would be great.

oliviertassinari:
Alright, at least, it easy to follow. Note that we have a few tests in the codebase that assert we don't render too many times. We could consider doing the same if we see regressions.

PR #2290, opened 2021-08-06
[core] useGridVirtualRows is using outdated pagination state

flaviendelangle:
This PR does not clean every state issue on the `useGridVirtualRows` hook. We still have potential inconsistencies between ref-based state access and selector-based state access.
I spent some time trying deeper changes but it's a rabbit hole and I want this PR to remain light.
In this PR I'm only removing some useless code and fixing the inconsistencies around the pagination variables since it's the one #2286 needs.

### Description

When removing a useless rerender in #2286, I hit a point where the last run of `getRenderState` was not having access to the last pagination state because the selector was still having the initial value for the `pageSize`.

The test failing in #2286 without this PR changes : https://app.circleci.com/pipelines/github/mui-org/material-ui-x/6726/workflows/2a19cd06-c747-4ca4-b318-ad8d4ea7a605/jobs/36258

Before #2286, on a test with `autoPageSize`, we had the following behavior

**Render n°1**
The `state.pagination.page` is initialized
=> The `useEffect` tracking the `state.pagination.page` from the selector if triggered 
=>The `updateViewport` method updates the `renderState` with `pageSize: 0`

**Render n°2**
The `state.viewportSizes` is initialized
=> The `state.rows.totalRowCount` + `state.pagination.pageSize` are initialized
=> The `useEffect` tracking the `state.rows.totalRowCount` from the selector is triggered
=> The `updateViewport` method updates the `renderState` with `pageSize: 0` (the selector still has the previous value)

**Render n°3**
The `state.scrollBar` + `state.containerSizes` + `state.viewportSizes` are updated (due to a bug fixed in #2286, they are all updated twice)
=> The `useEffect` tracking `containerSizes` is triggered but `updateViewport` is not call because `rendering.renderedSizes` has been set from the ref so it already has the new value (we should really drop the selectors for this scenarios). T
=> The `useEffect` tracking the `state.rows.totalRowCount` also have `state.scrollBar`, `state.containerSizes` and `state.viewportSizes` in its deps even if it's never using it, so it  runs again
=> The `updateViewport` method updates the `renderState` with `pageSize: 20`

With #2286 we are removing the 3rd rerender by removing a bug causing a useless call to `refreshContainerSizes`.
So their is never this re-run of `updateViewport` after the `pageSize` is initialized in the selector.

However, on the last "legit" run (2nd rerender), `apiRef.current.getState().pageSize` already have the good `pageSize` since its initialized synchronously with `totalRowCount`. So if we stop using the selector, it works.

My proposal is therefore to stop using the selector. The main consequence is that we must replace the `useEffect` bases on the selector with an event listener and a `useEffect` based on the prop (see below why this `useEffect`).

This hook would require some refacto to be more explicit about why it updates when some element changes.
But I want to keep the current change readable :+1: 

### Behavior changes

- Stop using the pagination selector to access `page` and `pageSize`. That means replacing the `useEffect` on the `paginationState.page` from the selector with a listener on the `PAGE_SIZE_CHANGE` and a `useEffect` on the `props.page`. It allows the hook to access the initialized page one render sooner.
- `useGridVirtualRows` has to be after `useGridPage` and `useGridPageSize` to have the correct pagination without selectors.
- Use `props` instead of `options` to avoid synchronization issues
- Remove `scrollBar` from the `useEffect` tracking the `totalRowCount` updates (am I missing something here ? If the scrollbar changes, it should change `containerSizes` right ?)
- Remove the `useEffect` tracking the `containerSizes` updates since its already in the `useEffect` tracking the `totalRowCount` 

### Small refacto

- `apiRef.current.updateViewport` can never be undefined in callbacks / effects of this hook, the existence checks are not needed
- The pagination part of `GridRenderContextProps` is never undefined even if pagination is not enabled (we have default values) 
- The `getRenderState` method do not return a partial version of the `GridRenderContextProps` but a full version, the `renderedColDef` is always defined.
- Merge `getRenderRowsProps` and `getRenderingState`, this hook has a lot of method only used in one place (`getRenderRowsProps` only used in `getRenderingState` which is only used in `reRender` which is only used in `updateViewport`). I understand this benefit of having small functions, but following the callback deps to understand why everything is running can be tricky. I also think the new versions is more explicit on the return type of `getRenderingState` and why the `Partial` is not needed here.

___

Since the page is controlled, I have to listen to the page change event AND to the prop change. 
We should probably have a unified way to listen to a control value change, not caring if it's currently being controlled or not.
___

All the tests are passing, do you see any edge cases that could be missing in the tests ?

___

Small label question : for me it was a bug, but an internal one, the problematic situation never occurred due to other behaviors in the codebase (an additional re-render that did have the correct pagination in the selector).
Do we put the `bug` label or not for those kind of situations ?


oliviertassinari (commented):
I wouldn't care much about what the virtualization hooks are doing at this point. I think that it would be better to reevaluate the whole implementation by going after the issues users reported with it. A better state refactoring can come along the way, at this point, we might lose this benefit as soon as we resume work on it.

---

Regarding the changes themselves, it's not clear to me that it's better to rely on the event than the state to listen to the page change. Reacting to state changes seems more idiomatic to React than reacting to events.

flaviendelangle:
> I wouldn't care much about what the virtualization hooks are doing at this point. 

It blocks those optimizations https://github.com/mui-org/material-ui-x/pull/2286/files#diff-b067e2d006d18d8c1214e487d80d2a267f8bc6e8b163055fb0c381317db7010f

I can remove them from #2286 to only merge the optimization in `useGridColumns` and do a new PR that focuses on removing the useless render fixes in the diff I linked above.
The CI will fail until we clean `useGridVirtualRows`. But it can remain "on hold" and be a good indicator of the viability of the new implementation.

oliviertassinari:
@flaviendelangle Ok, no real preference. I guess worse case we will have regressions and have the opportunity to add more tests.

PR #2295, opened 2021-08-06
[DataGrid] Remove dependency on x-license

### LINKED ISSUE #2291: [DataGrid] @material-ui/x-license is now required for DataGrid to load.
As of the current release, the open-source DataGrid requires @material-ui/x-license to be installed to load. This appears to have been introduced in #2176.

## Current Behavior 😯

DataGrid does not work unless you manually install @material-ui/x-license.

## Expected Behavior 🤔

DataGrid either has @material-ui/x-license installed as a dependency, or, since it's open source, don't even try to verify the license in DataGrid and just don't include the stuff that isn't open source with it. 

## Steps to Reproduce 🕹

https://codesandbox.io/s/n8nk9

Steps:

1. Open any of the examples in [the docs](https://material-ui.com/components/data-grid/) in CodeSandbox.
2. Marvel at the sight of it not loading because it can't find x-license since npm doesn't know its a dependency.

## Context 🔦

I'm trying to play with some examples in CodeSandbox so I don't have to pollute my local project. 

## Your Environment 🌎

CodeSandbox.

(If you want me to open a PR to add x-license as a dependency, I can do that in like 5 seconds. But I figured there's some regression I'm not seeing with my only mild familiarity with the codebase)

oliviertassinari:
I fail to see the use case for putting the license in the state, removing it. Closes #2291. 

Proof it works:

- DataGrid: https://codesandbox.io/s/zxtn5?file=/demo.js
- XGrid: https://codesandbox.io/s/dh3hr

For the test side of things, I'm out of ideas.


flaviendelangle:
I don't know what was the goal of putting it in the state
So if it works without it, it's great for me.

For the tests, once we isolate the DataGrid codebase into its own folder, we could use this ESLint rule on the DataGrid folter https://eslint.org/docs/rules/no-restricted-imports

flaviendelangle:
I checked the bundle with this PR and the DataGrid is not importing the license package anymore so perfect :+1:

flaviendelangle (approved):

oliviertassinari:
eslint `no-restricted-imports` is a great idea! We use it in the main repo for similar issues. We could already make it happen by moving `Watermark.tsx` into the x-license or x-grid package. I don't have the bandwidth to do it. If it happens a second time, I'm 💯  for it.

PR #2296, opened 2021-08-07
[DataGrid] Translate booleans when exporting to CSV

m4theushw:
A quick follow-up from https://github.com/mui-org/material-ui-x/pull/2242#discussion_r679944486

Before: https://codesandbox.io/s/material-demo-forked-4bdf7?file=/demo.js
Export to CSV and check the "Is admin" column. It's "true" and "false" when it should be translated.

After: https://codesandbox.io/s/material-demo-forked-rmgvq?file=/demo.js
Export to CSV and check the "Is admin" column. It's correctly translated.

### DIFF
--- a/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
+++ b/packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
@@ -7,7 +7,9 @@ import { getGridBooleanOperators } from './gridBooleanOperators';
 import { GridValueFormatterParams } from '../params/gridCellParams';
 
 function gridBooleanFormatter({ value, api }: GridValueFormatterParams) {
-  return api.getLocaleText(value ? 'booleanCellTrueLabel' : 'booleanCellFalseLabel');
+  return value
+    ? api.getLocaleText('booleanCellTrueLabel')
+    : api.getLocaleText('booleanCellFalseLabel');
 }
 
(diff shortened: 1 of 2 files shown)


packages/grid/_modules_/grid/models/colDef/gridBooleanColDef.tsx
```diff
@@ -4,6 +4,11 @@ import { renderBooleanCell } from '../../components/cell/GridBooleanCell';
 import { renderEditBooleanCell } from '../../components/cell/GridEditBooleanCell';
 import { gridNumberComparer } from '../../utils/sortingUtils';
 import { getGridBooleanOperators } from './gridBooleanOperators';
+import { GridValueFormatterParams } from '../params/gridCellParams';
+
+function gridBooleanFormatter({ value, api }: GridValueFormatterParams) {
+  return api.getLocaleText(value ? 'booleanCellTrueLabel' : 'booleanCellFalseLabel');
```
oliviertassinari, line 10: Would it make more sense so we can more easily grep all the usage of getLocaleText?

```suggestion
  return value ? api.getLocaleText('booleanCellTrueLabel') : api.getLocaleText('booleanCellFalseLabel');
```

oliviertassinari (commented):
Niiiice

flaviendelangle (approved):

PR #2299, opened 2021-08-07
[DataGrid] Remove private API from the export

oliviertassinari:
A quick follow-up on #2236


packages/grid/_modules_/grid/hooks/root/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from './useApi';
-export * from './useGridApiEventHandler';
+export type { GridCallbackDetails } from './useGridApiEventHandler';
```
oliviertassinari, line 2: Where the potential BC is. I don't think that we should document it as a BC in the changelog because we don't document what's exported here.

flaviendelangle: I'll remove all those `export *` for the hooks at one point to stop exporting the hooks we do not want to expose.
I don't think stop exporting `useGridApiEventHandler` is an issue

m4theushw (approved):
Looks good. Is it still a BC? I don't think people import from `useGridApiEventHandler.ts` to make us list when something is not exported.

Issue #820, opened 2021-01-08
[DataGrid] Add methods to save and restore the state


dheeraj1447:
Currently, no column re-order events are exposed on the tables both (DataGrid & XGrid). Need this feature as we can handle custom events while the column reordering happens.

- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

When column A is dragged to column B. I should be able to track it.

## Examples 🌈

- There is already a `ColumnReorderApi` present as part of the table but need to expose those as event handlers.

## Motivation 🔦

- This will help the users to track column as when the page reloads we still show the same view without reordering the columns to the initial state.

## Benchmark

- https://www.ag-grid.com/react-grid/column-state/#save-and-apply

DanailH:
Hi @dheeraj1447, quick question. Did you try doing `apiRef.current.subscribeEvent(COL_REORDER_START, yourHandlerHere);` ?

There is a list of events that are being fired when columns are reordered, you can check them [HERE](https://material-ui.com/components/data-grid/columns/#column-reorder)

dheeraj1447:
Yes @DanailH I tried that.. It would be lot better If that can be exposed on the table itself. Because `ApiRef` can only be accessible through functional components, If I'm not wrong.

DanailH:
Yes @dheeraj1447 `ApiRef` comes out of `useApiRef` which is a hook.

Maybe I'm not understanding the problem correctly, can you help me understand what do you call a table, is it the rows of the grid?

Or maybe you are reordering columns using the API, in which case these events won't be fired?

dheeraj1447:
Since `ApiRef` can only be used in a functional components, It would be better If the column re-order events are exposed as `XGrid` or `DataGrid's` declarative API. Take the custom event handler as a prop on these just like how you take event handlers `onPageChange`. `OnSelectionChange` etc.

For example,

`<XGrid {...props} onColReorderStart={customHandler} />`

@DanailH hope that gives you a clear picture on what I'm suggesting for. Thanks!

oliviertassinari:
@dheeraj1447 Ok so we don't need to solve the reorder event problem. What we need to solve is how to store the state of the grid between two sessions:

> we still show the same view without reordering the columns to the initial state.

We have started to work on it in #533, however, we aren't here yet. I'm at least aware of the problem problems to solve:

1. What's inside the public state:
[...]

PR #2300, opened 2021-08-07
[core] Remove use of getState

oliviertassinari:
A follow-up on https://github.com/mui-org/material-ui-x/pull/2208#discussion_r680924729. I don't get why we even have such `getState` method. For #820, I think that we could restructure the state, and maybe have `getState` returns a subset of the state that is public.


flaviendelangle:
@m2mathew do you remember exactly what was the purpose of `.getState()` and if it makes sense to use it in our codebase ?
I agree that it's lighter to just access the property, but I may be missing something here.

oliviertassinari:
We had a getState(subpart: string) API at some point 🤷‍♂️, happy with whatever direction we take here.

m4theushw (approved):
We already removed in #2141 the ability to return a subset of the state because it was never used. I'm OK with removing this method or making it private.

packages/grid/_modules_/grid/models/api/gridStateApi.ts
```diff
@@ -8,8 +8,7 @@ export interface GridStateApi {
   state: GridState;
   /**
    * Returns the state of the grid.
-   * @param {string} stateId The part of the state to be returned.
-   * @returns {any} The state of the grid.
+   * @returns {GridState} The state of the grid.
```
m4theushw, line 11: Should we not document it for now?

```suggestion
   * @returns {GridState} The state of the grid.
   * @ignore - do not document.
```

flaviendelangle:
If we stop using it internally, we can just remove it.

packages/grid/_modules_/grid/models/api/gridStateApi.ts
flaviendelangle: If we stop using it internally, we can just remove it.

m4theushw:
@oliviertassinari I've resolved the conflicts, removed a few remaining `getState` and made it private.

packages/grid/_modules_/grid/models/api/gridStateApi.ts
oliviertassinari: @flaviendelangle Removing the method would be a breaking change. We can move more iteratively.

oliviertassinari:
@m4theushw Thanks

Issue #1755, opened 2021-05-24
[DataGrid] Allow theme customizations


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Extending the style of the data grid components in the theme creates TypeScript errors.

## Expected Behavior 🤔

No TypeScript errors when customizing the data grid.

## Steps to Reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/material-demo-forked-2l2iu?file=/demo.tsx

```jsx
const theme = createMuiTheme({
  overrides: {
    MuiDataGrid: {
      root: {
        backgroundColor: "red"
      }
    }
  }
});
```

2. See the error 💥

<img width="520" alt="Screenshot 2021-05-24 at 12 13 55" src="https://user-images.githubusercontent.com/3165635/119332926-8b631500-bc89-11eb-9395-707aafe57d23.png">

## Context 🔦

Reported by @LeoLozes in https://github.com/mui-org/material-ui/issues/19427#issuecomment-846828746.

## Your Environment 🌎

v4.0.0-alpha.29

oliviertassinari:
For the solution, we can reproduce the solution used by `@material-ui/lab` to extend the core: https://next.material-ui.com/components/about-the-lab/#typescript. I think that it would also be good to explore how to make this built-in.

Primajin:
Hey there - I'm having a question regarding the overrides - are they actually read besides root? Putting the TypeError aside I am trying to use global overrides rather than using local classes and assign them to XGrid. (might be a duplicate of #919)

```json
    cell: {
      background: 'red !important',
      content: 'hello!',
      padding: 0,
    },
    root: {
      borderColor,
      borderRadius: 0,
      borderWidth: '1px 0 0',
    },
```
while I can see the styles from root applied - I can not see the ones from cell applied 🤔

oliviertassinari:
@Primajin Thanks for raising the duplicate. I have closed it. This is a simple problem to fix. But I think that we should wait for depending on v5 only to have a built-in fix. Developers can extend the theme manually for now, as long as we export the right type, which we could add now.

Regarding the available classes, you only have the `root` available. 
[...]

PR #2307, opened 2021-08-08
[DataGrid] TypeScript module augmentation for theme

ZeeshanTamboli:
Part of #1755 

Point 1 of https://github.com/mui-org/material-ui-x/issues/1755#issuecomment-881963942

We are extending the core styles overrides and theme props.

Added typescript tests for overrides and props. 

Proof that it works:
- Before: https://codesandbox.io/s/material-demo-forked-2eb6y
- After: https://codesandbox.io/s/material-demo-forked-obucn


packages/grid/_modules_/grid/themeAugmentation/overrides.ts
```diff
@@ -0,0 +1,9 @@
+import { GridClassKey } from '../models/gridClasses';
+
+export interface GridComponentNameToClassKey {
+  MuiDataGrid: GridClassKey;
+}
+
+declare module '@material-ui/core/styles/overrides' {
```
oliviertassinari, line 7: Would it work if we stay at the public API level?
```suggestion
declare module '@material-ui/core/styles' {
```

packages/grid/data-grid/src/tests/overrides.test.ts
```diff
@@ -0,0 +1,15 @@
+import { createTheme } from '@material-ui/core';
+
+createTheme({
+  overrides: {
```
oliviertassinari, line 4: This is the API of v4, which we do no longer optimize for. Could we test the v5 API?

oliviertassinari (changes requested):
- We also need to cover the default props theme override https://github.com/mui-org/material-ui/tree/next/packages/material-ui-lab/src/themeAugmentation 
- In the lab, we ask the developers to do the module augmentation manually. It seems that we could revisit the approach in the core as we could have pushed more for option 4. in https://github.com/mui-org/material-ui/pull/21279#issuecomment-639008993

ZeeshanTamboli:
> * We also need to cover the default props theme override https://github.com/mui-org/material-ui/tree/next/packages/material-ui-lab/src/themeAugmentation

@oliviertassinari  For props, how do we support `XGrid` props as well but not let it available for `DataGrid` users?

oliviertassinari:
> how do we support XGrid props as well but not make it available for DataGrid users?

@ZeeshanTamboli By scoping them per package, it can't be inside `/packages/grid/_modules_/`

packages/grid/data-grid/src/tests/overrides.test.ts
ZeeshanTamboli: The `components` ThemeOptions isn't available in `createTheme` on `@material-ui/core` under `packages/grid/data-grid`.  Following v5 api: https://next.material-ui.com/customization/theme-components/#global-style-overrides
Do we support v5 here?

packages/grid/_modules_/grid/themeAugmentation/overrides.ts
ZeeshanTamboli: For `overrides` it does not work. But for `props` it works. Only `'@material-ui/core/styles'` is fine for theme props.

packages/grid/data-grid/src/tests/overrides.test.ts
```diff
@@ -0,0 +1,15 @@
+import { createTheme } from '@material-ui/core';
```
m4theushw, line 1: TypeScript test files are named using the `.spec.ts` suffix.

packages/grid/_modules_/grid/themeAugmentation/overrides.ts
m4theushw: We can't stay at the public API level because the `overrides` definition is not exported in https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/styles/index.d.ts. In v5 it's exported.

ZeeshanTamboli:
As far as I understand, I think we will have to change these tests here once we add support for v5 of core styles in the grid.

oliviertassinari (commented):
The bundle infrastructure diverges from the main repo, we can't do https://github.com/mui-org/material-ui/pull/21279#discussion_r435748054 yet. Otherwise, I think that it's a good step forward

ZeeshanTamboli:
I have pushed https://github.com/mui-org/material-ui-x/pull/2307/commits/125b8c015a4984ab3f9b78568756abb5de5bb0b3 for docs.

Preview:

- DataGrid: https://deploy-preview-2307--material-ui-x.netlify.app/api/data-grid/data-grid/#component-name
- XGrid: https://deploy-preview-2307--material-ui-x.netlify.app/api/data-grid/x-grid/#component-name

oliviertassinari:
@ZeeshanTamboli Great 👍

m4theushw (approved):

PR #2211, opened 2021-07-26
[core] Add tests for column resizing

flaviendelangle:
Split from #2181

PR #2308, opened 2021-08-09
[DataGrid] Make resized column not flexible

flaviendelangle:


oliviertassinari (commented):
Seem to work: https://codesandbox.io/s/material-demo-forked-655lx?file=/demo.js, clean solution, 👍 for a test case as it's not straightforward to notice

flaviendelangle:
@oliviertassinari I just forgot to uncomment the test I did in #2211

PR #2312, opened 2021-08-09
[DataGrid] Remove `apiRef` from `options` params

### LINKED ISSUE #1301: [DataGrid] Standardize/Reconcile all event handlers and params

### LINKED ISSUE #1821: [DataGrid] Should not expose apiRef

DanailH:
Fixes #1821
Fixes #1301

This is part 2 of 2 for #1821. As we discussed during the call the `api` is removed from all `options` callbacks. It is only present in what callbacks are in the `GridColDef`.

---

### Breaking changes

All `DataGrid` `props` callbacks don't provide access to `api`  anymore. To access `api` you need to use `XGrid` and access it from the `details` parameter.

```
<XGrid 
  onColumnResize={(params, event, details) => console.log(details.api)}
/>
```

- `GridStateChangeParams` is removed, instead `onStateChange` first param is `GridState`

```diff
- onStateChange?: (params: GridStateChangeParams, event: MuiEvent<{}>, details: any) => void; 
+ onStateChange?: (state: GridState, event: MuiEvent<{}>, details: any) => void; 
```

```diff
export interface GridColumnHeaderParams {
   * The column of the current header component.
   */
  colDef: GridStateColDef;
-  /**
-  * API ref that let you manipulate the grid.
-  */
- api: any;
}
```

```diff
export interface GridColumnOrderChangeParams {
   * The old column index.
   */
  oldIndex: number;
-  /**
-   * API ref that let you manipulate the grid.
-   */
-  api: any;
}
```

```diff
export interface GridColumnResizeParams {
   * The column of the current header component.
   */
  colDef: GridStateColDef;
-  /**
-   * API ref that let you manipulate the grid.
-   */
-  api: any;
```

```diff
export interface GridColumnVisibilityChangeParams {
   * The column of the current header component.
   */
  colDef: GridStateColDef;
-  /**
-   * API ref that let you manipulate the grid.
-   */
-  api: any;
```

```diff
export interface GridRowParams {
   * All grid columns.
   */
  columns: GridColumns;
-  /**
-   * GridApiRef that let you manipulate the grid.
-   */
-  api: any;
```

```diff
export interface GridRowScrollEndParams {
   * The grid visible columns.
   */
  visibleColumns: GridColumns;
-  /**
-   * API ref that let you manipulate the grid.
-   */
-  api: any;
```


docs/src/pages/components/data-grid/style/StylingAllCells.tsx
```diff
@@ -37,7 +41,7 @@ export default function StylingAllCells() {
       <DataGrid
         rows={rows}
         columns={columns}
-        getCellClassName={(params: GridCellParams) => {
+        getCellClassName={(params: GridCellOptionsParams) => {
```
flaviendelangle, line 44: Do we want to explicitly set the `params` type here ?
TypeScript can infer it from the `getCellClassName` signature.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
```diff
@@ -90,7 +90,9 @@ export const GridRowCells = React.memo(function GridRowCells(props: RowCellsProp
     }
 
     if (getCellClassName) {
-      classNames.push(getCellClassName(cellParams));
+      const classNameParams = { ...cellParams };
```
flaviendelangle, line 93:
```suggestion
      const classNameParams = { api, ...cellParams };
```

This makes the `delete classNameParams.api` unnecessary.

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -240,15 +240,17 @@ export function useGridEditRows(
       }
 
       const isEditMode = cellMode === 'edit';
-
       const isModifierKeyPressed = event.ctrlKey || event.metaKey || event.altKey;
+      const stripedParams = { ...params };
```
flaviendelangle, line 244: Same as above

packages/grid/x-grid/src/tests/state.XGrid.test.tsx
```diff
@@ -50,9 +50,9 @@ describe('<XGrid /> - State', () => {
     render(<Test />);
     const header = screen.getByRole('columnheader', { name: 'brand' });
     fireEvent.click(header);
-    expect(onStateParams.api).to.equal(apiRef.current);
-    expect(onStateParams.state).to.equal(apiRef.current.state);
-    expect(onStateParams.state).to.not.equal(undefined);
+    // expect(onStateParams.api).to.equal(apiRef.current);
```
flaviendelangle, line 53: Shouldn't we remove this line altogether ?

docs/src/pages/components/data-grid/style/StylingAllCells.tsx
DanailH: It's docs, since this interface is exposed people can also see what is inside it.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2312/files#diff-b58ef2608987589ff71f621df6cc9458c607e9b080c9a2c9be7dcaf920539f67R111

In that case we may want to use delete here

packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -108,10 +108,10 @@ export function useGridEditRows(
 
   // TODO it's returning undefined when colDef.editable is undefined
```
m4theushw, line 109: This was already fixed by another PR.

```suggestion
```

packages/grid/_modules_/grid/hooks/utils/useResizeContainer.ts
```diff
@@ -19,9 +19,10 @@ export function useResizeContainer(
   const resizeFn = React.useCallback(() => {
     gridLogger.debug(`resizing...`);
 
-    apiRef.current.publishEvent(GridEvents.debouncedResize, {
-      containerSize: apiRef.current.getState().containerSizes?.windowSizes,
-    });
+    apiRef.current.publishEvent(
+      GridEvents.debouncedResize,
+      apiRef.current.getState().containerSizes?.windowSizes,
```
m4theushw, line 24: #2300

```suggestion
      apiRef.current.state.containerSizes?.windowSizes,
```

m4theushw (commented):
It seems weird to me to remove `api` from `GridRowParams` but keep it in `GridCellParams`. I would remove from both and for those places where it needs to be passed we add a child interface including it: e.g. `ValueParserParams`.

I found `GridCellOptionsParams` a bit confusing, because options and params are the same thing. Maybe by doing the first proposal we don't need it.

DanailH:
> It seems weird to me to remove `api` from `GridRowParams` but keep it in `GridCellParams`. I would remove from both and for those places where it needs to be passed we add a child interface including it: e.g. `ValueParserParams`.
> 
> I found `GridCellOptionsParams` a bit confusing, because options and params are the same thing. Maybe by doing the first proposal we don't need it.

Ok, the problem with that aproach is that by keeping the `GridCellParams` without `api` in it and changing the types to all `ColDef` callback to `ValueParserParams` which includes the `api` the typing won't be descriptive. Basically, the `renderCell`  will get params of type `ValueParserParams`?

I would keep it as it is but maybe change the name of the interface `GridCellOptionsParams`. I agree there are better names for it so I'm open to suggestions.

m4theushw:
> Basically, the renderCell will get params of type ValueParserParams?

No, we would create dedicated interfaces for these callbacks. In the case of `renderCell` it could be `RenderCellParams`. We're already doing that for some callbacks:

https://github.com/mui-org/material-ui-x/blob/8deeab3b2400f0a91b90f9cea4b2a82fe2d8f6d8/packages/grid/_modules_/grid/models/colDef/gridColDef.ts#L91
https://github.com/mui-org/material-ui-x/blob/8deeab3b2400f0a91b90f9cea4b2a82fe2d8f6d8/packages/grid/_modules_/grid/models/colDef/gridColDef.ts#L97
https://github.com/mui-org/material-ui-x/blob/8deeab3b2400f0a91b90f9cea4b2a82fe2d8f6d8/packages/grid/_modules_/grid/models/colDef/gridColDef.ts#L127

The reason I'm against keeping the `api` is that it's not consistent and we need to remember to always delete it when passing the params as argument.

DanailH:
But you would still need to pass in the `api` for those callbacks.

m4theushw:
The `api` would still be passed but it would be in a dedicated interface only for the params related to cell rendering.  

This component could use `RenderCellParams` instead of `GridCellParams`.

https://github.com/mui-org/material-ui-x/blob/8deeab3b2400f0a91b90f9cea4b2a82fe2d8f6d8/packages/grid/_modules_/grid/components/cell/GridBooleanCell.tsx#L5

One of our technical debits is that we used `GridCellParams` a lot only to pass the id and the field and to remove it now it will be a pain.

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -67,6 +67,7 @@ export const GridCell = React.memo(function GridCell(props: GridCellProps) {
       }
 
       const params = apiRef!.current.getCellParams(rowId, field || '');
+      delete params.api;
```
oliviertassinari, line 70: Could we do the opposite? Only provide the api where it's absolutely needed?

DanailH: Will work to create different interfaces for where the `api` is needed.

DanailH:
@m4theushw @oliviertassinari I've updated the `GridCellParams` so now it doesn't contain `api` anymore and `api` is explicitly added to the `GridColDef` methods that need it

oliviertassinari:
I have resolved the conflict that I introduced with #2340 (I work on this after seeing the same in this PR).

packages/grid/_modules_/grid/models/gridOptions.tsx
```diff
@@ -546,12 +546,12 @@ export interface GridOptions {
   onSortModelChange?: (model: GridSortModel, details?: any) => void;
   /**
    * Callback fired when the state of the grid is updated.
-   * @param params
+   * @param state
```
m4theushw, line 549:
```suggestion
   * @param state The new state.
```

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -110,12 +111,12 @@ export interface GridColDef {
...
    * @param params
    */
-  renderCell?: (params: GridCellParams) => React.ReactNode;
+  renderCell?: (params: GridRenderCellParams) => React.ReactNode;
   /**
    * Allows to override the component rendered in edit cell mode for this column.
    * @param params
    */
-  renderEditCell?: (params: GridCellParams) => React.ReactNode;
+  renderEditCell?: (params: GridRenderCellParams) => React.ReactNode;
```
m4theushw, line 119: We can reuse `GridRenderCellParams` but it also need to include `GridEditCellProps` or the user won't be able to access the additional props.

https://github.com/mui-org/material-ui-x/blob/019c7ccb19a8ee31214da7936cdd4c899921aa3c/packages/grid/_modules_/grid/models/gridEditRowModel.ts#L3-L6

```suggestion
  renderEditCell?: (params: GridRenderEditCellParams) => React.ReactNode;
```

m4theushw (approved):
Nice work!

Issue #2292, opened 2021-08-06
[core] Replace the options with direct prop reads


flaviendelangle:
Part of #2231 

## Goals

1. Avoid re-renders caused by props-to-state patterns

2. Avoid having synchronization issues between the values from the `options` and the values from the `props` inside hooks and components

When setting a prop, you have a render with the new prop but the old option. So if some parts of the code uses the options but other uses the props it causes bugs.

## Proposal

We can replace the options by a preprocessing of the props that accomplish the same things as `useOptionsProp` but synchronously

```ts
// useGridPropsWithDefaultValues.ts

const GRID_DEFAULT_PROPS: POCGridOptions = {
  columnBuffer: 2,
  columnTypes: getGridDefaultColumnTypes(),
  density: GridDensityTypes.Standard,
  localeText: GRID_DEFAULT_LOCALE_TEXT,
};

/**
 * Properties with a default value that can be overridden with a prop
 */
interface POCGridOptions {
  /**
   * Number of columns rendered outside the grid viewport.
   * @default 2
   */
  columnBuffer: number;
  /**
   * Extend native column types with your new column types.
   */
  columnTypes: GridColumnTypesRecord;
  /**
   * Set the density of the grid.
   */
  density: GridDensityTypes;
  /**
   * Set the locale text of the grid.
   * You can find all the translation keys supported in [the source](https://github.com/mui-org/material-ui-x/blob/HEAD/packages/grid/_modules_/grid/constants/localeTextConstants.ts) in the GitHub repository.
   */
  localeText: Partial<GridLocaleText>;
}

/**
 * Properties that do not have a default value
 */
interface POCGridOtherProps {
  /**
   * The ref object that allows grid manipulation. Can be instantiated with [[useGridApiRef()]].
   */
  apiRef?: GridApiRef;
  /**
   * The label of the grid.
   */
  'aria-label'?: string;
  /**
   * Override or extend the styles applied to the component.
   */
  classes?: GridClasses;
  /**
   * Set of columns of type [[GridColumns]].
   */
  columns: GridColumns;
  /**
   * Set of rows of type [[GridRowsProp]].
   */
  rows: GridRowsProp;
}

type POCGridInputComponentProps = POCGridOtherProps & Partial<POCGridOptions>;

type POCGridComponentProps = POCGridOtherProps & POCGridOptions;

export const useGridPropsWithDefaultValues = (
  inProps: POCGridInputComponentProps,
): POCGridComponentProps => {
[...]

PR #2318, opened 2021-08-10
[core] Reduce options usage (3)

flaviendelangle:
Part of #2292 

Remove all the options from our codebase that are not used for `classes` or `localText` or `scrollbarSize`

Next:

- Preprocess the props to merge the `classes` and `localText` with default values
- Stop using the `options.scrollbarSize` 
- Replace `options` with `rootProps` in `useGridSlotComponentProps`


oliviertassinari (commented):
It looks great. We could maybe rebase his PR, not sure? Flavien is off next week too.

m4theushw (approved):

Issue #1519, opened 2021-04-29
[DataGrid] Only three columns are rendered in jsdom


mattcarlotta:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

With the inclusion of #1361, jsdom appears to only be partially rendering columns. The first two (with checkbox) columns within the array are being rendered, while the rest are hidden.

## Expected Behavior 🤔

All columns should be rendered and visible.

## Steps to Reproduce 🕹

Repo [here](https://github.com/mattcarlotta/datagrid-jsdom-bug).

Instructions can be found [here](https://github.com/mattcarlotta/datagrid-jsdom-bug/blob/main/README.md).

## Context 🔦

Since not all columns are rendered, I can't test secondary actions (my table actions) nor can I achieve full columns coverage.

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>

```
System:
    OS: Linux 5.8 Linux Mint 20.1 (Ulyssa)
  Binaries:
    Node: 14.16.1 - /usr/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 6.14.12 - /usr/bin/npm
  Browsers:
    Chrome: 90.0.4430.85
    Firefox: 88.0
  npmPackages:
    @emotion/react: ^11.1.5 => 11.1.5 
    @emotion/styled: ^11.3.0 => 11.3.0 
    @material-ui/core: ^4.11.4 => 4.11.4 
    @material-ui/data-grid: https://pkg.csb.dev/mui-org/material-ui-x/commit/0c03787f/@material-ui/data-grid => 4.0.0-alpha.26 
    @material-ui/styles:  4.11.4 
    @material-ui/system:  4.11.3 
    @material-ui/types:  5.1.0 
    @material-ui/utils:  4.11.2 
    @types/react:  17.0.4 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    typescript: ^4.2.4 => 4.2.4
```
</details>

oliviertassinari:
@mattcarlotta Thanks for reporting this limitation. I can confirm with this test case:

```diff
diff --git a/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx b/packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
[...]

PR #2326, opened 2021-08-12
[DataGrid] Add `disableVirtualization` to disable virtualization completely

### LINKED ISSUE #1781: [DataGrid] Allow to disable virtualization
It should be possible to disable row virtualization.

Reasons why you would want to disable row virtualization:

1. It introduces *microstutter* when scrolling, especially on mobile
2. It breaks <kbd>Ctrl</kbd>+<kbd>F</kbd> for users
3. It doesn't provide much benefit if the page size is small

m4theushw:
Preview https://deploy-preview-2326--material-ui-x.netlify.app/components/data-grid/virtualization/#disable-virtualization

Closes #1781

I created a separate hook to not mix concerns and to touch as little as possible the virtualization logic.

- cf7ac773 moves the `scrollToIndexes` and `scroll` to a dedicated hook because they're not tied to the virtualization
- 624352ee better name the hook
- fe60f827 adds a `useGridNoVirtualization` hook to sync the state and handle the scroll when virtualization is off

Preview: https://deploy-preview-2326--material-ui-x.netlify.app/storybook/?path=/story/x-grid-tests-rows--disable-virtualization&globals=measureEnabled:false


oliviertassinari:
Note that we still have 1. because of how scroll needs to be sync with translate. I think that we should look into dropping this pattern for #1933. We are the only one in the benchmark to not directly rely on the native scroll feature rendering.

m4theushw:
There's a bug in this PR. Filter and try to navigate with the keyboard, it doesn't work.

---

> We are the only one in the benchmark to not directly rely on the native scroll feature rendering.

@oliviertassinari Yeah, I'll investigate if it's easier to use the native scroll instead. Basically, we need to remove the two divs between the window and the rendering zone.

image

> Do we disable column virtualization?

Yes

> Should we add documentation about it in https://material-ui.com/components/data-grid/virtualization/?

I added a small section in https://deploy-preview-2326--material-ui-x.netlify.app/components/data-grid/virtualization/#disable-virtualization

docs/src/pages/components/data-grid/virtualization/virtualization.md
```diff
@@ -33,9 +33,13 @@ By default, 2 columns are rendered outside of the viewport. You can change this
 
 You can disable column virtualization by setting the column buffer to a higher number than the number of rendered columns, e.g. with `columnBuffer={columns.length}` or `columnBuffer={Number.MAX_SAFE_INTEGER}`.
 
-## apiRef [<span class="pro"></span>](https://material-ui.com/store/items/material-ui-pro/)
+## Disable virtualization
 
-{{"demo": "pages/components/data-grid/virtualization/VirtualizationApiNoSnap.js", "bg": "inline", "hideToolbar": true}}
```
oliviertassinari, line 38: Why did you remove this section?

m4theushw: One of the changes is to move the scroll methods to a separated hook. Since this interface is empty now I removed it. Maybe we should add a page for scrolling with a demo: https://github.com/mui-org/material-ui-x/issues/1103#issuecomment-885695797. In this new page we could document the API available.

flaviendelangle:
Probably also closes #1519

packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts
```diff
@@ -0,0 +1,81 @@
...
+  const visibleColumns = useGridSelector(apiRef, visibleGridColumnsSelector);
+
+  const syncState = React.useCallback(() => {
+    if (!gridState.containerSizes || !windowRef?.current) {
+      return;
+    }
+
+    let firstRowIdx = 0;
+    const { page, pageSize } = paginationState;
+    if (props.pagination && pageSize != null && props.paginationMode === 'client') {
```
flaviendelangle, line 30: `pageSize` can never be null here
See https://github.com/mui-org/material-ui-x/pull/2433/files/0eb2f693687f04990c2b318706cb4923ac25c91c#r695592921

flaviendelangle (changes requested):

DanailH (approved):

Issue #2315, opened 2021-08-09
[DataGrid] `autoHeight` + `hideFooter` results in double bottom border


lgeiger:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Enabling both `autoHeight` and `hideFooter` in the data grid renders a double border on the bottom of data grid:
<img width="579" alt="Screenshot 2021-08-10 at 00 28 45" src="https://user-images.githubusercontent.com/13285808/128787595-2521cce8-3862-4c36-9fa8-b8d6edf05147.png">

This is very minor, but it looks like the bottom border of the last grid row shouldn't be rendered in this case to avoid this.

## Expected Behavior 🤔

No double border should appear on the bottom of the grid.

## Steps to Reproduce 🕹

https://codesandbox.io/s/material-demo-forked-3ndjy?file=/demo.tsx

Steps:

1. Zoom in and look at the bottom of the data grid

## Context 🔦

This is a small, but noticeable layout bug. This is especially noticeable if the border size of the columns doesn't match the root border color of the grid.

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
  System:
    OS: macOS 11.5.1
  Binaries:
    Node: 16.6.1 - /usr/local/bin/node
    Yarn: Not Found
    npm: 7.20.3 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
  npmPackages:
    @emotion/react: ^11.4.0 => 11.4.0
    @emotion/styled: ^11.3.0 => 11.3.0
    @material-ui/core: ^5.0.0-beta.2 => 5.0.0-beta.2
    @material-ui/data-grid: ^4.0.0-alpha.35 => 4.0.0-alpha.35
    @material-ui/icons: ^5.0.0-beta.1 => 5.0.0-beta.1
    @material-ui/private-theming:  5.0.0-beta.2
    @material-ui/styled-engine:  5.0.0-beta.1
    @material-ui/styles: ^5.0.0-beta.2 => 5.0.0-beta.2
    @material-ui/system:  5.0.0-beta.2
    @material-ui/types:  6.0.1
    @material-ui/unstyled:  5.0.0-alpha.41
    @material-ui/utils:  5.0.0-beta.1
    @types/react: 17.0.15 => 17.0.15
[...]

PR #2327, opened 2021-08-12
[DataGrid] Remove unused row CSS classes

ZeeshanTamboli:
### Breaking changes

- [DataGrid] The CSS classes `.Mui-odd` and `.Mui-even` were removed from the row. Use the `:nth-child(odd|even)` selector to replace them.

---

While looking at bug #2315 which is unexpectedly on a large scope due to virutalization case to be considered for last row, I noticed this line is not being used. 

So removed `Mui-even` and `Mui-odd` classes.

Perhaps it is added so that it can be targeted via CSS like `& .Mui-even`, `& .Mui-odd`. If yes, I think we can rename it to `MuiDataGrid-odd` and `MuiDataGrid-even` and also document it. 

If its relevant, we can close this PR.

### DIFF
--- a/packages/grid/_modules_/grid/components/GridRow.tsx
+++ b/packages/grid/_modules_/grid/components/GridRow.tsx
@@ -11,13 +11,12 @@ import { optionsSelector } from '../hooks/utils/optionsSelector';
 export interface GridRowProps {
   id: GridRowId;
   selected: boolean;
-  className: string;
   rowIndex: number;
   children: React.ReactNode;
 }
 
 export function GridRow(props: GridRowProps) {
-  const { selected, id, className, rowIndex, children } = props;
+  const { selected, id, rowIndex, children } = props;
   const ariaRowIndex = rowIndex + 2; // 1 for the header row and 1 as it's 1 based
   const apiRef = useGridApiContext();
   const rowHeight = useGridSelector(apiRef, gridDensityRowHeightSelector);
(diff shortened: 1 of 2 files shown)


m4theushw (approved):
I marked it as a BC because we removed CSS classes. I think the alternative is good but would be nice to do https://github.com/mui-org/material-ui-x/pull/2327#pullrequestreview-728769169 as well. If users still want the classes they can re-add them.

ZeeshanTamboli:
> I think the alternative is good

Nice, didn't knew there is this `:nth-child(odd|even)` selector.

>  but would be nice to do [#2327 (review)](https://github.com/mui-org/material-ui-x/pull/2327#pullrequestreview-728769169) as well. If users still want the classes they can re-add them.

Sure, I can try if it is useful.

oliviertassinari:
> Use the :nth-child(odd|even) selector to replace them.

This doesn't work in practice, because the rows are virtualized.

m4theushw:
It works to some extent. https://codesandbox.io/s/material-demo-forked-u4z7j?file=/demo.js

When it changes the rows it will invert the styling: odd -> even.

oliviertassinari:
> When it changes the rows it will invert the styling: odd -> even.

Right, it's my concern, when end-users scroll, it's 🎄

Capture d’écran 2021-08-13 à 18 51 38
Capture d’écran 2021-08-13 à 18 51 33

oliviertassinari:
I have open https://github.com/mui-org/material-ui-x/issues/2334 to anticipate the next steps

ZeeshanTamboli:
> I have open #2334 to anticipate the next steps

Yeah, thanks and sorry for merging too quickly. I could revert this PR if needed soon.

shoxter:
Is there any fix to the breaking change of .Mui-even and .Mui-odd? I'm forced to revert back to an alpha where this PR hasn't been merged because of the breaking change.

flaviendelangle:
@shoxter do we agree that you are looking for the index on the current page (considering the whole filtered list as a single page if youi do not paginate) ?

m4theushw:
Moving the discussion to #2334

PR #1289, opened 2021-03-25
[DataGrid] Make column header cells focusable

DanailH:
Fixes #1022 

Preview: https://deploy-preview-1289--material-ui-x.netlify.app/components/data-grid/

**A general observation - this was trickier than expected and is more of a "nice to have" feature, but at least navigating the grid with the keyboard is now possible**

This PR aims to provide the ability for users to navigate to the column headers using the keyboard. In addition, focusing on a column header cell and pressing `enter` will trigger the sorting, and pressing `ctrl`/`metaKey` + `enter` will open the 3dot 
column menu and focus it so that pressing the arrows will navigate within the column menu. Pressing `escape` when the 3dot column menu is opened will close it and return the focus to the last focused column header cell.

Things that are missing in this PR:
 - currently, there is a bug that when you navigate with the keyboard to the last rendered column header cell and then try to scroll to the next column header cell the scroll is broken and doesn't calculate the positions correctly.
 - as far as I could work out there is no way to navigate to focus a cell in the grid, meaning that when changing focus using `tab` the focus will move through what is in the toolbar and then be lost - IMO it should focus the first cell or column header (needs to be decided)

To be added:
- [x] tests
- [x] probably it will be a good idea to document the new keyboard interactions

PR #2336, opened 2021-08-13
[DataGrid] Use rowIndex relative to visible rows on keyboard navigation

### LINKED ISSUE #2335: [DataGrid] Keyboard navigation not working when rows are filtered

## Current Behavior 😯

Navigating with the keyboard when the rows are filtered doesn't work.

## Expected Behavior 🤔

It should allow to navigate between cells when the rows are filtered.

## Steps to Reproduce 🕹

Steps:

1. Open https://material-ui.com/components/data-grid/demo/
2. Add a filter to a column
3. Use the keyboard to navigate between cells

## Context 🔦

This is a regression from #1289
Noticed while working in https://github.com/mui-org/material-ui-x/pull/2326#issuecomment-898010271
We also had a comment about this in https://github.com/mui-org/material-ui-x/issues/1822#issuecomment-877096194

m4theushw:
Fixes #2335 

Preview: https://deploy-preview-2336--material-ui-x.netlify.app/components/data-grid/filtering/#basic-filter

The `getRowIndex` can't be used since it returns the index relative to all rows. When the rows are filtered the index might be different. It was working before #1289 because `setCellFocus` required indexes instead of the id of the cell. Since it needs the id now, we first need to get the correct id corresponding to the computed next index.


DanailH (approved):

oliviertassinari:
This change seems to trigger :

> Uncaught TypeError: Invalid attempt to destructure non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
    at r (90a709d476ef7b5c93e6f792e6832025104dadea.307fb0f35e0b173d3dd8.js:1)
    at l (90a709d476ef7b5c93e6f792e6832025104dadea.307fb0f35e0b173d3dd8.js:1)
    at a50aa100c893fbfa86a65ea271ac39b20e5e6fce.f321f07a39172bfdb4ee.js:1
    at n.<anonymous> (a50aa100c893fbfa86a65ea271ac39b20e5e6fce.f321f07a39172bfdb4ee.js:1)
    at n.value (a50aa100c893fbfa86a65ea271ac39b20e5e6fce.f321f07a39172bfdb4ee.js:1)
    at Object.publishEvent (a50aa100c893fbfa86a65ea271ac39b20e5e6fce.f321f07a39172bfdb4ee.js:1)
    at n.a.useCallback.e.current.s.forEach.e.current.hasOwnProperty.e.current.<computed> [as publishEvent] 
    
When trying to use CTRL + END on a filtered list (the key combination that moves to the last row/last column).

Issue #210, opened 2020-08-22
[DataGridPro] Implement Tree data


oliviertassinari:
### Benchmark

- https://material-table.com/#/docs/features/tree-data
- https://ant.design/components/table/#components-table-demo-expand-children
- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/tree-data/
- https://www.ag-grid.com/javascript-grid-tree-data/
- https://www.telerik.com/kendo-react-ui/components/treelist/
- https://ej2.syncfusion.com/react/demos/#/material/treegrid/treegrid-overview
- http://tabulator.info/docs/4.5/tree
- https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--tree-view
- https://adazzle.github.io/react-data-grid/#/tree-view

teenguyen:
Hi! Are there any updates to how this feature is going? I've noticed it's been pushed back a few times now and trying to manually implement this has been fairly difficult. Thanks!

flaviendelangle:
@teenguyen, you have the `Project` section on the right pannel that tells when we expect to build the feature.
It's not guaranteed of course but it can give you an idea of when we intend to do it :+1:

george-thomas-hill:
Upvote!

I'm glad to see that this feature is coming soon!

matandro:
The linked benchmarks focus on child rows / columns
Would this feature include a basic expendable description of the row (single div expending from row containing)?
Example https://material-table.com/#/docs/features/detail-panel

oliviertassinari:
@matandro Yes, this is pretty much what this feature is about.

JonasRothmann:
This is the only feature I need to move from react-table to XGrid

oliviertassinari:
@JonasRothmann It's on our roadmap for Q3 2021. Hopefully, we can start working on it next sprint, in two weeks.

paulprinke:
Looking forward to this release!  Is it still progressing for a Q3 release?

flaviendelangle:
I wrote the 1st line of code this afternoon.
It will be a large feature to build so it won't come that soon, but it's now in progress :+1:

bitcoinbullbullbull:
Is there an estimate on release for this feature?

Seddy24:
@flaviendelangle 

Is it possible to render the groupingColDef in a different spot than the first column?

flaviendelangle:
Not currently, what would be your usecase ?

Seddy24:
[...]

PR #2337, opened 2021-08-13
[DataGrid] Remove state.isScrolling

m4theushw:
One of #820

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts
@@ -45,26 +45,12 @@ export const useGridColumnMenu = (apiRef: GridApiRef): void => {
   );
 
   const handleColumnResizeStart = React.useCallback(() => {
-    setGridState((state) => {
-      if (state.columnMenu.open) {
-        return {
-          ...state,
-          columnMenu: {
-            ...state.columnMenu,
-            open: false,
-          },
-        };
-      }
+    hideColumnMenu();
+  }, [hideColumnMenu]);
 
-      return state;
-    });
-  }, [setGridState]);
-
-  React.useEffect(() => {
-    if (gridState.isScrolling) {
-      hideColumnMenu();
-    }
-  }, [gridState.isScrolling, hideColumnMenu]);
+  const handleRowsScroll = React.useCallback(() => {
+    hideColumnMenu();
+  }, [hideColumnMenu]);
 
   useGridApiMethod(
     apiRef,
@@ -77,4 +63,5 @@ export const useGridColumnMenu = (apiRef: GridApiRef): void => {
   );
 
   useGridApiEventHandler(apiRef, GridEvents.columnResizeStart, handleColumnResizeStart);
+  useGridApiEventHandler(apiRef, GridEvents.rowsScroll, handleRowsScroll);
 };
--- a/packages/grid/_modules_/grid/hooks/features/core/gridState.ts
+++ b/packages/grid/_modules_/grid/hooks/features/core/gridState.ts
@@ -38,7 +38,6 @@ export interface GridState {
   editRows: GridEditRowsModel;
   pagination: GridPaginationState;
   options: GridOptions;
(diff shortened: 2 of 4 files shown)


packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts
```diff
@@ -77,4 +63,5 @@ export const useGridColumnMenu = (apiRef: GridApiRef): void => {
   );
 
   useGridApiEventHandler(apiRef, GridEvents.columnResizeStart, handleColumnResizeStart);
+  useGridApiEventHandler(apiRef, GridEvents.rowsScroll, handleRowsScroll);
```
oliviertassinari, line 66: Would it make sense to simplify even futher?
```suggestion
  useGridApiEventHandler(apiRef, GridEvents.columnResizeStart, hideColumnMenu);
  useGridApiEventHandler(apiRef, GridEvents.rowsScroll, hideColumnMenu);
```

oliviertassinari (commented):
> One of #820

For this issue, I'm not sure that we need to clean up the state. Would it work if we segment it instead? Meaning, have different top-level sections:

- `state`: the root
 - `public`: the part that developers can store in their DB
 - `private`: where we store private things
 
Another alternative is to add a `getPublicState` method that returns a subset. One downside with this issue is that it doesn't force us to think upfront about what's public and what's private.

m4theushw:
> For this issue, I'm not sure that we need to remove things in the state.

@oliviertassinari I think we have two things to have in mind here. The first one being to solve #820 and another one which is the opportunity to clean the state. Besides the first one, removing data not widely used from the state might be worth it to reduce complexity. Now we have a good view of what kind of information must be in the state and what's used only in one hook. For everything that is not shared between hooks, we could use internal state. 

> Would it work if we segment it instead? Meaning, have different top-level sections:

My idea would be to have a single state. Having a private state seems that we're storing derived data. With a single state, developers only need to serialize the state and save in their DBs. To restore the session, we just replace the entire state with the newer one, doing some sanity check before (e.g. a column might not exist anymore).

> If we think about it, what would developers want to store? Based on the duplicates of #820:

I can also think about pagination, selection and with features like #210 we might want to store other stuff (opened/closed nodes).

oliviertassinari:
@m4theushw Cleaning the state sound great. For the "private state", my proposal was to use the public state as the source of truth and to use the private state to store intermediary or short-lived state there, stuff that could be recomputed at any time, the same way of how useMemo behaves.

packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts
```diff
@@ -76,5 +54,6 @@ export const useGridColumnMenu = (apiRef: GridApiRef): void => {
     'ColumnMenuApi',
   );
 
-  useGridApiEventHandler(apiRef, GridEvents.columnResizeStart, handleColumnResizeStart);
+  useGridApiEventHandler(apiRef, GridEvents.columnResizeStart, hideColumnMenu);
```
flaviendelangle, line 57: `hideColumnMenu` is also resetting `state.columnMenu.field`, is it changing the behavior ?

```tsx
// handleColumnResizeStart
columnMenu: { ...state.columnMenu,  open: false }

// hideColumnMenu
columnMenu: { ...state.columnMenu, open: false, field: undefined },
```

flaviendelangle:
Yeah it's a great idea to remove this kind of data from the state, it will help a lot when trying to change the implementation of the state.

packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts
m4theushw: There's no problem in not resetting the field. When the menu is open it always changes the field:

https://github.com/mui-org/material-ui-x/blob/8a8828dca0c9c0c1264b955ab76a84b566fd54a6/packages/grid/_modules_/grid/hooks/features/columnMenu/useGridColumnMenu.ts#L18

PR #2341, opened 2021-08-14
[DataGrid] Move packages to `@mui` scope

### LINKED ISSUE #2313: Rename the packages
We are making a change to the terminology used in the project to support the long-term direction. The high-level bit is that we are going to use `x-` as a prefix for the rich/advanced components and `-pro` as a suffix for the paid professional version.

Current:

- `@material-ui/data-grid` DataGrid MIT
- `@material-ui/x-grid XGrid` Commercial

Target:

- `@mui/x-data-grid` DataGrid MIT 
- `@mui/x-data-grid-pro` DataGridPro Commercial
- `@mui/x-tree-view` TreeView MIT 
- `@mui/x-tree-view-pro` TreeViewPro Commercial

This is the result of a longer discussion that happened in https://www.notion.so/mui-org/Rearranging-the-packages-2f700caf8b5e48559fe86ebb648ed91a.

#### 1. MUI over Material-UI

- A large part of the community is confusing Material-UI with Material Design, or even Google, MUI distinguishes ourselves
- `@mui` is shorter to write, read, and pronounce.
- The advanced components, while only currently providing a Material Design UI are meant to be heavily customizable. The features matter more than the design

#### 2. The X prefix

Pros:

- x allows the product to live on its own. Bootstrap didn't need any of the advanced components to strive, nor does the core team need them. We want to compete with the DX for instance in the core. In X we are competing with other standalone open-source UI components, react-table, Kendo-UI, AG Grid, etc. 
- x allows signaling to the developers on which repository they should seek support. We have seen too many developers opening issues on https://github.com/mui-org/material-ui while it should be here.
- x makes it easy to avoid package name collisions between the different repositories of MUI. Say x needs a codemode, `@material-ui/codemod` is already taken, what should we use?

Cons:

- Longer to write and read.
- It clearly isolates the component, it does no longer feel part of the exact same product.

#### 3. The Pro suffix 

- pro makes it clear what's not MIT. We saw too many developers confused about what is X about, is it about the paid version, or is it a different product?

## The execution steps

We will make the first step ASAP

- [x] Grant org access of the members to `@mui`
- [x] Rename the packages #2341
- [x] Rename the components: XGrid -> DataGridPro #2382 
- [x] Release beta
- [x] We deprecate the latest version of the older packages with migration instruction to check 
- [x] Rearrange docs according to https://www.notion.so/mui-org/Sitemap-and-URL-structure-rearrange-ca604f628c17438db2d987961fcb8ffe#be64fe17a857402f9b0712e5a98e3e23
- [ ] Update the demos in the core to reflect the change of the package names 
- [ ] Move the GitHub repository: https://github.com/mui-org/material-ui-x > https://github.com/mui/mui-x cc @mbrookes could you help me with this one?
- [ ] Move the folders to be consistent with the main repository: x-license -> mui-x-license-pro

cc @mui-org/x

oliviertassinari:
### Breaking changes

- [DataGrid] Move packages to `@mui` scope (#2341) @oliviertassinari 

  ```diff
  -import { DataGrid } from '@material-ui/data-grid';
  +import { DataGrid } from '@mui/x-data-grid';
  ```

  ```diff
  -import { LicenseInfo } from '@material-ui/x-grid';
  +import { LicenseInfo } from '@mui/x-data-grid-pro';
  ```
---

One part of #2313 

The next steps (different PRs to minimize conflicts):

- [ ] Rename the components: XGrid -> DataGridPro
- [ ] Move the folders to be consistent with the main repository: x-license -> mui-x-license-pro
- [ ] Move the GitHub repository: https://github.com/mui-org/material-ui-x > https://github.com/mui/mui-x cc @mbrookes could you help me with this one?


mbrookes:
Would losing the "- in "data-grid" work, i.e. `@mui/x-datagrid-pro`? It's slightly shorter, and reads a tiny bit better.

Also, seeing it practice, I'm wondering about the need for the x prefix? I know there's been much discussion on the package renaming internally and otherwise, but we only realistically get one shot at it, so we should be sure about it before we pull the trigger.

What do we gain by the x prefix here? It's a separate package, so is explicitly not part of core (or whatever that will now be called)...

oliviertassinari:
> Would losing the "- in "data-grid" work,

If we do this change, I think that `<DataGrid>` => `<Datagrid>` in React and "data grid" => "datagrid" in the markdown should happen too. It wouldn't be consistent otherwise.

>  I'm wondering about the need for the x prefix?

The purpose would be to clearly indicate to the developers where they should search for the details on this package, which product it's part of (core or x?). For instance, will they find the GitHub issue tracker/the source on mui/mui or mui/mui-x.

mbrookes:
> If we do this change, I think that <DataGrid> => <Datagrid> in React and "data grid" => "datagrid" in the markdown should happen too. It wouldn't be consistent otherwise.

I'm on the fence then (and probably easily pushed off 😄 ), as I'm not so sure about some of these related changes.

> which product it's part of (core or x?)

If it isn't in `@mui/core`, it isn't part of core? 🤷🏼‍♂️ 

It hasn't been a concern so far for other packages that aren't part of core (icons, for example).

oliviertassinari:
> I'm not so sure about some of these related changes.

Fair enough, my reasoning is that, we pick the name of the component, and we derive everything else from it:

- React: PascalCase
- URL, kebab care
- npm, kebab care

> If it isn't in @mui/core, it isn't part of core? 🤷🏼‍♂️

`@mui` is the namespace of the company. What would you do with the low-code solution packages?

mbrookes:
> Fair enough, my reasoning is that, we pick the name of the component, and we derive everything else from it

I think you might have misunderstood. I was leaning towards agreeing with you.

>> It hasn't been a concern so far for other packages that aren't part of core (icons, for example).

Crickets...

oliviertassinari:
> It hasn't been a concern so far for other packages that aren't part of core (icons, for example).

We used to have a single repository, we have two now.
I can relate to one difference that x is likely going to have: one package per component. It's likely that we will have more components.

mbrookes:
> Having a prefix to enforce no name collision might be interesting.

Having no name collision to begin with seems like the better option.

oliviertassinari:
Actually, I wonder if we shouldn't add a `@material-ui/x-codemod` package here.

PR #2342, opened 2021-08-14
[DataGrid] Fix keyboard navigation header regression

oliviertassinari:
See https://github.com/mui-org/material-ui-x/pull/2279/files#r688988720 for why. I didn't add a test case, but if somebody want to, please go ahead.


m4theushw (approved):
I've added two tests.

PR #2343, opened 2021-08-14
[core] Clean up grid column

### LINKED ISSUE #2333: Error: "n" is read-only
Hi!

- [Yes ] The issue is present in the latest release.
- [ Yes ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯
when i deploy the project to GOCD ubuntu distro I got this mensaje when i m try to render de x-grid, is strange because i can't reproduce this in local, only in the server. this happen after update @material-ui/x-grid "4.0.0-alpha.24" to "4.0.0-alpha.36.
update : this happen only in 4.0.0-alpha.36

image

image

## Your Environment 🌎

<details>
  
```
  System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.15.5 - C:\Program Files\nodejs\node.EXE       
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1023.0), Chromium (92.0.902.67)
  npmPackages:
    @emotion/styled:  10.0.27
    @material-ui/styles:  4.11.4
    @material-ui/system:  4.12.1
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @material-ui/x-grid: ^4.0.0-alpha.37 => 4.0.0-alpha.37
    @material-ui/x-grid-data-generator: ^4.0.0-alpha.37 => 4.0.0-alpha.37
    @material-ui/x-license:  4.0.0-alpha.37
    @types/react:  17.0.0
    react: ^17.0.1 => 17.0.1
    react-dom: ^17.0.1 => 17.0.1
    styled-components: ^5.2.1 => 5.2.1
```
</details>

oliviertassinari:
I have tried to simplify the logic of this hook. It's mostly about renaming the variable to minimize the number of different terms we use. I have also skipped a few branches that seem to be unreachable.

I did this while I was looking into #2344

Closes #2333

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -136,7 +136,7 @@ export function useGridColumns(
   const allColumns = useGridSelector(apiRef, allGridColumnsSelector);
   const visibleColumns = useGridSelector(apiRef, visibleGridColumnsSelector);
 
-  const getGridColumnsState = React.useCallback(
+  const setGridColumnsState = React.useCallback(
     (columnsState: GridColumnsState, emit = true) => {
       logger.debug('Updating columns state.');
 
@@ -201,9 +201,9 @@ export function useGridColumns(
         }, {}),
       };
 
-      getGridColumnsState(columnState, emit);
+      setGridColumnsState(columnState, emit);
     },
-    [apiRef, logger, getGridColumnsState],
+    [apiRef, logger, setGridColumnsState],
   );
 
   const updateColumns = React.useCallback(
@@ -256,9 +256,9 @@ export function useGridColumns(
 
       const updatedColumns = [...gridState.columns.all];
       updatedColumns.splice(targetIndexPosition, 0, updatedColumns.splice(oldIndexPosition, 1)[0]);
-      getGridColumnsState({ ...gridState.columns, all: updatedColumns });
+      setGridColumnsState({ ...gridState.columns, all: updatedColumns });
     },
-    [apiRef, gridState.columns, logger, getGridColumnsState],
+    [apiRef, gridState.columns, logger, setGridColumnsState],
   );
 
   const setColumnWidth = React.useCallback(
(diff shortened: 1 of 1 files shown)


packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
```diff
@@ -145,15 +136,15 @@ export function useGridColumns(
   const allColumns = useGridSelector(apiRef, allGridColumnsSelector);
   const visibleColumns = useGridSelector(apiRef, visibleGridColumnsSelector);
 
-  const updateState = React.useCallback(
-    (newState: GridColumnsState, emit = true) => {
+  const getGridColumnsState = React.useCallback(
```
m4theushw, line 139: This function doesn't return anything.

```suggestion
  const updateColumnsState = React.useCallback(
```

We could try, in another PR, to remove it and use only `setColumnsState`.

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
oliviertassinari: Right, I meant to use 

```suggestion
  const setGridColumnsState = React.useCallback(
```

flaviendelangle:
I think we should stop using the term `hydrate` for columns processings
As Sebastian told, we are using it in a very different way as React which is confusing.

oliviertassinari:
Happy to use a different terminology. I was using it coming from an backend perspective, for instance https://stackoverflow.com/questions/6991135/what-does-it-mean-to-hydrate-an-object (in my case it was from a python library not a Java one but it's the same concern I believe). Do you have an alternative in mind?

Mike-Loffland:
When is this slated for release?

oliviertassinari:
Today

PR #2346, opened 2021-08-14
[docs] Remove legacy logic

oliviertassinari:
Since #2281, the demos break on `createTheme()` for developers that uses core < v4.12.0 so we might as well remove the rest.


oliviertassinari:
My bad, we didn't make this change in the core.

PR #2378, opened 2021-08-16
[core] Update `doesSupportTouchActionNone` implementation using `CSS.supports`

DanailH:
Based on this comment https://github.com/mui-org/material-ui/pull/27724#issuecomment-898614730

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
+++ b/packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
@@ -20,25 +20,24 @@ import { useGridState } from '../core/useGridState';
 import { useNativeEventListener } from '../../root/useNativeEventListener';
 import { GridComponentProps } from '../../../GridComponentProps';
 
-let cachedSupportsTouchActionNone = false;
-
 // TODO: remove support for Safari < 13.
 // https://caniuse.com/#search=touch-action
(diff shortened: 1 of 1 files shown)


DanailH:
Can I merge this?

oliviertassinari:
I think that we could wait on what are the final supported browsers for v5, maybe we could remove the whole method.

m4theushw:
First things first. Support for IE 11 is completely broke because we're compiling using ES6 which is not supported:

https://github.com/mui-org/material-ui-x/blob/aa9922f68b987ce0fe721fccd93c7227278ea098/tsconfig.json#L3

Then, we have this import which is not ES5-complaint: 

https://github.com/mui-org/material-ui-x/blob/aa9922f68b987ce0fe721fccd93c7227278ea098/packages/grid/data-grid/src/DataGrid.tsx#L3

We can use the legacy version of it available in the same package.

I managed to run in IE 11 changing the files above, but there're a bunch of things not working, including the resize.

---

For this PR we can merge it, but we need to evaluate if we want to support, even partially, IE in v4. Fortunately, nobody complained about.

m4theushw (approved):

oliviertassinari:
> if we want to support, even partially, IE in v4

@m4theushw It resonates with https://github.com/mui-org/material-ui-x/issues/775#issuecomment-750079555

PR #2382, opened 2021-08-17
[DataGridPro] Rename XGrid to DataGridPro

m4theushw:
### Breaking changes

- [DataGridPro] Rename XGrid to DataGridPro

  ```diff
  -import { XGrid } from '@mui/x-data-grid-pro';
  +import { DataGridPro } from '@mui/x-data-grid-pro';
  ```

---

Part of #2313

### DIFF
--- a/packages/grid/x-grid/src/DataGridPro.tsx
+++ b/packages/grid/x-grid/src/DataGridPro.tsx
@@ -12,10 +12,10 @@ import {
   useGridApiRef,
 } from '../../_modules_/grid';
 import { GridContextProvider } from '../../_modules_/grid/context/GridContextProvider';
-import { useXGridComponent } from './useXGridComponent';
+import { useDataGridProComponent } from './useDataGridProComponent';
 import { Watermark } from '../../_modules_/grid/components/Watermark';
-import { XGridProps } from './XGridProps';
-import { useXGridProps } from './useXGridProps';
+import { DataGridProProps } from './DataGridProProps';
+import { useDataGridProProps } from './useDataGridProProps';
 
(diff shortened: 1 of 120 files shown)


packages/grid/x-grid/src/DataGridPro.tsx
```diff
@@ -12,10 +12,10 @@ import {
   useGridApiRef,
```
DanailH, line 12: Can we also rename the folders? 

From `packages/grid/x-grid/src/XGrid.tsx` -> `packages/grid/data-grid-pro/src/DataGridPro.tsx`

Same for the `x-grid-data-generator` -> `data-grid-data-generator`.

DanailH:
Thanks, @m4theushw! I was planning to do this today to enable the `beta` release.

packages/grid/x-grid/src/DataGridPro.tsx
m4theushw: In #2313 there's a "Move the folders to be consistent with the main repository" item. I didn't move anything here because I thought it would be done in this task.

DanailH: That also works.

DanailH (approved):

oliviertassinari:
I believe we miss a 301 for https://master--material-ui-x.netlify.app/api/data-grid/x-grid/ it should redirect to https://master--material-ui-x.netlify.app/api/data-grid/data-grid-pro/. It can be added here: https://github.com/mui-org/material-ui-x/blob/44e659b4abd1939eadf5cefa144303b082b0dafb/docs/public/_redirects#L11

PR #2384, opened 2021-08-18
[core] Update `.browserslistrc`

DanailH:
Based on this PR https://github.com/mui-org/material-ui/pull/27788

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -7,46 +7,48 @@ node 14
 
 # Default/Fallback
 # `npx browserslist --mobile-to-desktop "> 0.5%, last 2 versions, Firefox ESR, not dead, not IE 11"` when the last major is released.
-# On update check all #stable-snapshot markers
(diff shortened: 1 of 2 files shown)


oliviertassinari (approved):
Since we buifd with TypeScript, and the builder doesn't handle the browserlist file, these changes have almost no impacts (it's a different story in the main repository).

flaviendelangle:
Couldn't we use the file from the core ?

In package.json

```json
  "browserslist": [
    "extends @material-ui/monorepo/.browserslistrc"
  ],
```

DanailH:
> Since we build with TypeScript here, and the builder doesn't handle the browserlist file, these changes have almost no impacts (it's a different story in the main repository).

Should we then just ignore the changes and close the PR?

oliviertassinari:
> Couldn't we use the file from the core ?

It seems supported https://github.com/browserslist/browserslist#shareable-configs but I wonder. If we intend to cover multiple versions of the core, we might have to compile with the older versions of the browsers?

PR #2390, opened 2021-08-19
[docs] Fix onCellEditCommit param type

ArthurPedroti:
A little fix of docs, a think this type was updated and was missed to update the docs:

Current docs:
image

Current type at "data-grid.d.ts":
image


m4theushw (approved):
Thanks!

PR #2391, opened 2021-08-19
[docs] Fix arguments in onXXX props

m4theushw:
Noticed while reviewing #2390


flaviendelangle:
I think that we should be able to type correctly this `details` argument.
The `any` feels wrong.

PR #2395, opened 2021-08-20
[core] Generate propTypes

m4theushw:
One step towards #1826

I'm already working to generate the documentation for the components based on the propTypes.


docs/scripts/generateProptypes.ts
```diff
@@ -0,0 +1,90 @@
...
+  const formatted = fixBabelGeneratorIssues(prettified);
+  const correctedLineEndings = fixLineEndings(sourceContent, formatted);
+
+  await fse.writeFile(sourceFile, correctedLineEndings);
+}
+
+async function run() {
+  const files = [
+    path.resolve(__dirname, '../../packages/grid/data-grid/src/DataGrid.tsx'),
+    path.resolve(__dirname, '../../packages/grid/x-grid/src/DataGridPro.tsx'),
```
oliviertassinari, line 53: I think that we should do it for all the exported components

packages/grid/data-grid/src/DataGrid.tsx
```diff
@@ -40,113 +40,184 @@ DataGridRaw.defaultProps = DEFAULT_GRID_PROPS_FROM_OPTIONS;
...
-          `Material-UI: \`apiRef\` is not a valid prop.`,
-          'GridApiRef is not available in the MIT version.',
-          '',
-          'You need to upgrade to the DataGridPro component to unlock this feature.',
-        ].join('\n'),
-      );
-    }
-    return null;
-  }),
-  checkboxSelectionVisibleOnly: chainPropTypes(PropTypes.bool, (props: any) => {
```
oliviertassinari, line 58: We are losing the custom prop-types. If necessary, we can use this comment https://github.com/mui-org/material-ui/blob/a4e2ed82c76fffb5875ba96ab8c0efd5bcb424ef/packages/material-ui/src/Popper/Popper.js#L314 

x the others

oliviertassinari (commented):
It looks like we are on the right track. It feels like it uses a lot of the code of the main repository, which feels great.

packages/grid/data-grid/src/DataGrid.tsx
m4theushw: The comment doesn't work in this situation. I'm already using it on other props. The prop-type for this prop was removed because `checkboxSelectionVisibleOnly` is omitted from the grid's props. The generator doesn't allow to have more prop-types than the props specified in the TS definition.

https://github.com/mui-org/material-ui-x/blob/1a0ad36f7380ba6bff04fa8f2b6366e001988734/packages/grid/data-grid/src/DataGridProps.ts#L11

docs/scripts/generateProptypes.ts
m4theushw: Done! Now it's generating the prop-types for any exported component.

package.json
```diff
@@ -61,6 +62,7 @@
     "@material-ui/core": "^4.9.12",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.54",
+    "@material-ui/utils": "^5.0.0-beta.4",
```
m4theushw, line 65: I don't know why but without adding it here, the build fails. The dependency is already added to the DataGrid's package.json.

packages/grid/data-grid/src/DataGrid.tsx
flaviendelangle: What is the conclusion here ?
Do we have a way to improve the generator to keep those prop types ?

babel.config.js
```diff
@@ -44,7 +44,12 @@ module.exports = {
       },
     ],
   ],
-  ignore: [/@babel[\\|/]runtime/], // Fix a Windows issue.
+  ignore: [
+    // Fix a Windows issue.
+    /@babel[\\|/]runtime/,
+    // Fix const foo = /{{(.+?)}}/gs; crashing.
+    /prettier/,
```
oliviertassinari, line 51: It seems to work without, is it still needed?

```suggestion
```

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -642,22 +641,13 @@ describe('<DataGrid /> - Layout & Warnings', () => {
   });
 
   describe('warnings', () => {
-    before(() => {
-      PropTypes.resetWarningCache();
-    });
```
oliviertassinari, line 647: Seeing this diff made me suspicious, so I gave it a try to confirm it: When running the tests in watch mode, the first run is green, but the second one is red. What motived the change of approach?

oliviertassinari (commented):
1. Saw a new warning opening http://0.0.0.0:3001/components/data-grid/columns/:

Screenshot 2021-09-04 at 16 18 28

2. I expect the new prop-types to significantly increase the bundle size. But we didn't bring the bundle size checker from the main repo yet so we are blind. Also, we are no using the babel remove prop-type helpers to work around the limitation. What should do about it. Would it make more sense to only generate the prop-types for `DataGrid` and `DataGridPro` and hold on for the other components?

packages/grid/data-grid/src/DataGrid.tsx
m4theushw: There's no way to include more props than those allowed in the interface of the component. In theory, if the prop is not in the interface it shouldn't be used in the component. We could add a warning in `useDataGridProps` if the user tries to use a prop from the Pro version.

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
m4theushw: To test the warnings we were using `PropTypes.checkPropTypes` and feeding it with the prop-types. However, to get the prop-types they had to be attached to the exported component, which is a memo. Memoized components can't have prop-types so the generator attaches them to the `DataGridRaw`.

https://github.com/mui-org/material-ui-x/blob/2cd18f9cc5f4723a20f4ddec2ef3864e398e726b/packages/grid/data-grid/src/DataGrid.tsx#L38-L41

Since there's no more way to access the prop-types directly, I had to render the component to be able to test. The reason for seeing the second run red in watch mode: https://github.com/facebook/react/issues/18251#issuecomment-607567390

babel.config.js
m4theushw: Here it crashes if I remove.

image

m4theushw:
> I expect the new prop-types to significantly increase the bundle size.

@oliviertassinari  It increased 5%. I'll open another PR to remove during the build the prop-types from internal components.

image

packages/grid/data-grid/src/DataGrid.tsx
flaviendelangle: Would you do this switch to `useDataGridProps` / `useDataGridProProps` in this PR ?
It would be sad to have a release without those warnings.

packages/grid/_modules_/grid/components/cell/GridRowCells.tsx
```diff
@@ -131,4 +132,88 @@ export const GridRowCells = React.memo(function GridRowCells(props: RowCellsProp
...
+  cellClassName: PropTypes.string,
+  cellFocus: PropTypes.shape({
+    field: PropTypes.string.isRequired,
+    id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+  }),
+  cellTabIndex: PropTypes.shape({
+    field: PropTypes.string.isRequired,
+    id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+  }),
+  columns: PropTypes.arrayOf(
```
flaviendelangle, line 153: Do we want to have the details for this type ? It's quite heavy
Same thing for the other `columns`, `currentColumn` or `colDef`

packages/grid/data-grid/src/DataGrid.tsx
m4theushw: > Would you do this switch to useDataGridProps / useDataGridProProps in this PR ?

In another PR.

PR #2396, opened 2021-08-21
[docs] Add `Row & Cell editing` in features list

ZeeshanTamboli:
Maybe not that important, but since we are completed with row and cell editing, thought we can promote `Editing` as a feature.

In docs rearranging (part of  #2313), we need to change these as well or maybe there will be a re-direct rule.

Preview: https://deploy-preview-2396--material-ui-x.netlify.app/components/data-grid/#features

### DIFF
--- a/docs/src/pages/components/data-grid/overview/overview.md
+++ b/docs/src/pages/components/data-grid/overview/overview.md
@@ -71,6 +71,7 @@ We provide three options:
 - High performance 🚀
 - [Filtering](/components/data-grid/filtering/) and [multi-filtering](/components/data-grid/filtering/#multi-column-filtering) <span class="pro"></span>
 - [Pagination](/components/data-grid/pagination/)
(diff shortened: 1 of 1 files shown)


docs/src/pages/components/data-grid/overview/overview.md
```diff
@@ -71,6 +71,7 @@ We provide three options:
 - High performance 🚀
 - [Filtering](/components/data-grid/filtering/) and [multi-filtering](/components/data-grid/filtering/#multi-column-filtering) <span class="pro"></span>
 - [Pagination](/components/data-grid/pagination/)
+- [Editing](/components/data-grid/editing/)
```
ZeeshanTamboli, line 74:
```suggestion
- [Row & Cell editing](/components/data-grid/editing/)
```

oliviertassinari:
Good idea, it's an important feature

PR #2431, opened 2021-08-22
[DataGrid]: Add zhCN localization 

wlf100220:
I use DataGrid in chinese, but there is lack of zhCN in the package.

### DIFF
--- a/packages/grid/_modules_/grid/locales/zhCN.ts
+++ b/packages/grid/_modules_/grid/locales/zhCN.ts
@@ -1,0 +1,106 @@
+import { enUS as enUSCore } from '@material-ui/core/locale';
+import { GridLocaleText } from '../models/api/gridLocaleTextApi';
+import { getGridLocalization, Localization } from '../utils/getGridLocalization';
+
+export const zhCNGrid: Partial<GridLocaleText> = {
+  // Root
+  noRowsLabel: "没有数据。",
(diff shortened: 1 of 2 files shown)


flaviendelangle:
Thanks for the contribution
Could you add a line for `zhCN` on this file ? https://github.com/mui-org/material-ui-x/blob/next/docs/src/pages/components/data-grid/localization/localization.md

wlf100220:
> Thanks for the contribution
> Could you add a line for `zhCN` on this file ? https://github.com/mui-org/material-ui-x/blob/next/docs/src/pages/components/data-grid/localization/localization.md

Yes, with pleasure. And it's done.

packages/grid/_modules_/grid/locales/zhCN.ts
```diff
@@ -0,0 +1,106 @@
+import { enUS as enUSCore } from '@material-ui/core/locale';
```
flaviendelangle, line 1: The core has a `zhCN` local, we should probably use it instead of `enUS`

wlf100220: Ahh~, it's a mistake. Thank you very much.

flaviendelangle (approved):

flaviendelangle:
Looks fine to me
I'm letting it open if someone else has something to say, but it will be merged for next release :+1:

PR #2433, opened 2021-08-23
[core] Replace the options with direct prop reads

### LINKED ISSUE #2292: [core] Replace the options with direct prop reads

### LINKED ISSUE #2439: Data Grid default props for paginationMode
- [X ] The issue is present in the latest release.
- [X ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Setting paginationMode in the default theme properties does not work, this may apply to other properties as well. Some properties are picked up however.

## Expected Behavior 🤔

Setting paginationMode to server in the default theme properties should be used in DataGrid instances.

## Steps to Reproduce 🕹

https://codesandbox.io/s/material-data-grid-default-props-p7bw4?file=/demo.tsx

Steps:

1. Go to the second page of results, the result table is blank (the DataGrid instance did not use the default paginationMode set in the theme)
2. Set the paginationMode="server" on the DataGrid instance, the second page of results is no longer blank.

## Context 🔦

The DataGrid appears to ignore the default property setting as of 4.0.0-alpha.36 but it works in 4.0.0-alpha.35

## Your Environment 🌎
This is after a rollback to 4.0.0-alpha.35. Just updating datagrid causes the problem to come back.

<details>
npx @material-ui/envinfo
npx: installed 2 in 7.601s

  System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.17.4 - /usr/local/opt/node@14/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.14.14 - /usr/local/opt/node@14/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Edge: 92.0.902.78
    Firefox: 89.0.2
    Safari: 14.1.2
  npmPackages:
    @material-ui/data-grid: ^4.0.0-alpha.35 => 4.0.0-alpha.35
    @material-ui/styles:  4.11.4
    @material-ui/system:  4.12.1
    @material-ui/types:  5.1.0
    @material-ui/utils:  4.11.2
    @types/react: ^17.0.18 => 17.0.18
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.3.5 => 4.3.5
  
```
```
</details>

flaviendelangle:
### Breaking change

```diff
-const { options } = useGridSlotComponentProps();
+const rootProps = useGridRootProps();
``` 

---

Closes #2292 
Fixes #2439

- [x] Replace `state.options` and `.defaultProps` with a step of processing of the props (`useProcessedProps`)
- [x] Replace `apiRef.current.components` and `apiRef.current.componentProps` by `rootProps.components` and `rootProps.componentProps`
- [x] Distinguish typing before and after preprocessing to avoid `rootProps.myProp!` after the preprocessing
- [x] Add missing `forceUpdate` in `setSelectionModel` (there was probably a side effect rerender due to options update, the bug occured after I removed the options)
- [x] nullable entries of `rootProps.components` should always have a null check before instantiating them and entries with a null check before instantiation should be nullable in the typing
___

### Improve the typings

We now distinguish the input value and the value after the props processing.

For the props with a simple default value
On the `DataGrid`  / `DataGridPro` components properties, they are optional (`headerHeight?: number`) 
But when used in the hooks / components inside the Grid, they are mandatory (`headerHeight: number`) because we applied the default value

For the props with a merge behavior (<=> the user can pass a partial value which will be merged with a full default value).
On the `DataGrid` / `DataGridPro` components properties, they are optional and partial (`localeText?: Partial<GridLocaleText>`)
But when used in the hooks / components inside the Grid, they are mandatory and complete (`localeText: GridLocaleText`)

### Have synchronous access to the processed values of the props

When using `options.x`, `x` has the default value during the 1st render.
Using a synchronous processing of the props instead of storing in the state allow to access the overridden value of `options.x` all the time.

### Clearly distinguish nullable and non-nullable components

Some components like the `Toolbar` or `ColumnUnsortedIcon` can be null (in which case we don't render them)
But as the internal typing of `GridSlotsComponent` had every key optional, we were not clearly distinguishing them from the non-nullable components (`ErrorOverlay` for instance) which crash if you set them at null.

Now the internal typing should reflect the fact that some components can be null (in which case the code calling them must check if they are not null) and some can't (in which case the test is not usefull).

## New wording

If you have ideas to make it as clear as possible

### `DataGridProps` / `DataGridProProps` 
The interfaces of the `DataGrid` and `DataGridPro` components 

- do not have the prop `signature`
- the `DataGridProps` do not have the paid-only props
- all the prop with a default value are optional `headerHeight?: number`
- all the prop with a merge behavior are optional and partial (`localeText?: Partial<GridLocaleText>`)

### `GridComponentInputProps`
The interface given to `useProcessedProps`

- have the prop `signature`
- all the prop with a default value are optional (`headerHeight?: number`)
- all the prop with a merge behavior are optional and partial (`localeText?: Partial<GridLocaleText>`)

### `GridComponentProps`
The interface returned by `useProcessedProps`
It is accessible via `useGridRootProps` and passed for every hooks called in `useDataGridComponent` / `useDataGridProComponent`

- have the prop `signature`
- all the prop with a default value are mandatory since we applied the default value (`headerHeight: number`)
- all the prop with a merge behavior are mandatory and complete since we merged with the default value (`localeText: GridLocaleText`)


docs/pages/api-docs/data-grid/grid-api.md
```diff
@@ -10,85 +10,83 @@ import { GridApi } from '@mui/x-data-grid-pro';
```
flaviendelangle, line 10: Switch to riche comparison to have something readable
I don't know why Git is marking every line as modified

Their are two changes

- Removal of `components` and `componentProps` keys (use `useGridRootProps` instead)
- Order change in type disjunction for two callbacks, I don't know why but it's not a real change

flaviendelangle:
@m4theushw @DanailH 
I'd be interesting in feedbacks about the approach I took
It's mainly on those two files :

https://github.com/mui-org/material-ui-x/pull/2433/files#diff-02afbf689cabc39856e95f2e3d013f06dd141a10ce1ce2f2db4f9ed8d760b67b => the new hook processing the props
https://github.com/mui-org/material-ui-x/pull/2433/files#diff-ac062a61f7ba3a7517eefeb144fbb489ad5228d76e8143ace2832a8d5e222aae => the hook calling the prop processing hook for DataGrid

There are a lot of changes due to the removal of `apiRef.current.components` and `state.options` but first we should agree on the basic idea :+1:

m4theushw (commented):
The `options` key is present in https://deploy-preview-2433--material-ui-x.netlify.app/components/data-grid/components/#getting-props yet.

image

m4theushw (approved):

packages/grid/_modules_/grid/components/GridFooter.tsx
```diff
@@ -31,21 +29,17 @@ export const GridFooter = React.forwardRef<HTMLDivElement, GridFooterContainerPr
...
+      rootProps.components.Pagination && (
+        <rootProps.components.Pagination {...rootProps.componentsProps?.pagination} />
+      );
 
     return (
       <GridFooterContainer ref={ref} {...props}>
         {SelectedRowCountElement}
         {RowCountElement}
-        {PaginationElement}
+        {paginationElement}
```
oliviertassinari, line 42: Great call, elements are camelCase, not PascalCase.

packages/grid/_modules_/grid/hooks/features/useGridSlotComponentProps.tsx
```diff
@@ -1,30 +1,25 @@
...
   const columns = useGridSelector(apiRef, visibleGridColumnsSelector);
   const [state] = useGridState(apiRef);
 
-  const slotComponentProps: GridSlotComponentProps = React.useMemo(
+  return React.useMemo<GridSlotComponentProps>(
     () => ({
       state,
       rows,
       columns,
-      options,
```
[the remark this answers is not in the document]
oliviertassinari: Could we update the PR's description to document the migration for once we have to generate the changelog, and document the migration path for developer's that rely on it.

packages/grid/_modules_/grid/components/GridFooter.tsx
flaviendelangle: Their are some others to rename, I'll to a quick standalone PR later

Issue #1329, opened 2021-04-02
XGrid - Allow strong typing for GridColDef data (perhaps with Generics)


jamshally:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

When implementing a `renderCell` function on a `GridColDef` - the `GridCellParams` property is not generic. It would be very useful if this type was generic, so that the `params.value` and `params.row` properties could be strongly typed

### CURRENT
```
   {
        field: 'createdOn',
        headerName: 'Created',
        renderCell: (params: GridCellParams) => (
                         // 🠑🠑 params.value & params.row are note strongly typed
        ),
    },
```

### SUGGESTED
(note the generic 
```
   {
        field: 'createdOn',
        headerName: 'Created',
        renderCell: (params: GridCellParams<MyType>) => (
                         // 🠑🠑 params.value & params.row are strongly typed!
        ),
    },
```

## Motivation 🔦

Without strong typing, the implementation of renderCell methods requires dealing with params as `any` type, leaving a lot of room for introducing runtime errors

## Order id 💳
21793

oliviertassinari:
Thoughts:

1. I don't get why we don't have an any in

https://github.com/mui-org/material-ui-x/blob/25edafedf67bea45f6c6ed15281f3ba8f9b6b65e/packages/grid/_modules_/grid/models/gridCell.ts#L9

2. `params.value` is the output of `valueGetter`. So `params.value` and `params.row` would need two different arguments. Maybe a first optional argument for GridRowModel, and an optional second argument that uses the field name and GridRowModel to return GridCellValue could work.

jamshally:
@oliviertassinari:
 1 - I'm not sure either
 2 - Sounds good to me

Haaxor1689:
I've just got around to upgrading past the beta5 version with generics, and I feel like the implementation isn't complete. `GridColDef` is still not generic so you either need to type all of the functions like `renderCell` or `valueGetter` manually or just use type casts as before. Neither props of DataGrid like `onRowClick` are generic, so again the usefulness of adding generics is totally lost. Is this the final version of generics?

flaviendelangle:
Could you give me concrete example of what the code would look like on your side with full generic support ?
We did a first light version of the generics where you have to set the type manually on the callbacks

For instance: 

```tsx
const columns: GridColumns = [
  { ... },
  { ... },
  {
    field: 'age',
[...]

Issue #545, opened 2020-11-06
Generic TypeScript Types


BrentFarese:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

It would be great if the `XGrid`, `DataGrid` and other components utilized TS generics for the type system.  For instance, in [DataGrid](https://material-ui.com/api/data-grid/#datagrid-api), the type for `onRowSelected` is `(param: RowSelectedParams) => void`.  `RowSelectedParams` contains a `data` prop that has a type as follows:

```
interface RowData extends ObjectWithId {
    [key: string]: any;
}
```

A developer should really be able to pass in a type parameter to either `RowSelectedParams` like `RowSelectedParams<MyRowDataShape>` so that the `data` prop comes back with a proper type in TS.  There are other instances of the type system for MUI-X that could greatly benefit from generics.

It would be super useful for TS users!  Right now, I am just using type assertions to coerce the type of things (like `data` above) into the proper shape.

## Examples 🌈

Examples above.

## Motivation 🔦

Motivation above.  A better type system is just overall an improvement for TS users.

Thanks!

PR #2436, opened 2021-08-23
[DataGrid] Add option for generic types in `GridRowParams`, `GridCellParams`, `GridRenderCellParams`

ZeeshanTamboli:
Exploring generics for some row and cell types.

Related to issue #1329 (maybe closes this?), issue #545

By passing generic types, we can type `value`, `formattedValue`, `getValue` return type and `row` in `GridRenderCellParams`.

For onRowXXXX handlers, row is typed:

Screenshot (7)

Need early feedback and how we can go further.

**TODO:**

- [ ] I tried auto generate docs, it created wrong docs with generics. I think we need to support generics in script.

### DIFF
--- a/packages/grid/_modules_/grid/models/gridRows.ts
+++ b/packages/grid/_modules_/grid/models/gridRows.ts
@@ -1,16 +1,20 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-export type GridRowData = { [key: string]: any };
+interface ID {
+  id: any;
+}
+interface RowIndexSignature<T> {
+  [key: string]: T;
+}
+export type GridRowData<T = any> = RowIndexSignature<T> | ID;
 
 /**
  * The key value object representing the data of a row.
  */
-export type GridRowModel = GridRowData;
+export type GridRowModel<T = any> = GridRowData<T>;
 
 export type GridUpdateAction = 'delete';
 
-export interface GridRowModelUpdate extends GridRowData {
-  _action?: GridUpdateAction;
-}
+export type GridRowModelUpdate = GridRowData & { _action?: GridUpdateAction };
 
 /**
  * The type of Id supported by the grid.
--- a/packages/grid/_modules_/grid/models/gridRows.ts
+++ b/packages/grid/_modules_/grid/models/gridRows.ts
@@ -1,5 +1,4 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-
 export type GridRowData<T = any> = { [key: string]: T };
 
 /**
@@ -9,7 +8,9 @@ export type GridRowModel<T = any> = GridRowData<T>;
 
 export type GridUpdateAction = 'delete';
 
-export type GridRowModelUpdate = GridRowData & { _action?: GridUpdateAction };
+export interface GridRowModelUpdate extends GridRowData {
+  _action?: GridUpdateAction;
+}
 
 /**
  * The type of Id supported by the grid.
--- a/packages/grid/_modules_/grid/models/gridRows.ts
+++ b/packages/grid/_modules_/grid/models/gridRows.ts
@@ -1,10 +1,10 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-export type GridRowData<T = any> = { [key: string]: T };
+export type GridRowData<RowType = any> = RowType;
 
 /**
  * The key value object representing the data of a row.
  */
-export type GridRowModel<T = any> = GridRowData<T>;
+export type GridRowModel<RowType = any> = GridRowData<RowType>;
 
 export type GridUpdateAction = 'delete';
 
--- a/packages/grid/_modules_/grid/models/gridRows.ts
+++ b/packages/grid/_modules_/grid/models/gridRows.ts
@@ -1,11 +1,6 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-interface ID {
-  id: any;
-}
-interface RowIndexSignature<T> {
-  [key: string]: T;
-}
-export type GridRowData<T = any> = RowIndexSignature<T> | ID;
+
+export type GridRowData<T = any> = { [key: string]: T };
 
 /**
  * The key value object representing the data of a row.
--- a/packages/grid/_modules_/grid/models/gridRows.ts
+++ b/packages/grid/_modules_/grid/models/gridRows.ts
@@ -1,5 +1,5 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-export type GridRowData<RowType = any> = RowType;
+export type GridRowData<RowType = { [key: string]: any }> = RowType;
 
 /**
  * The key value object representing the data of a row.
--- a/packages/grid/_modules_/grid/models/params/gridCellParams.ts
+++ b/packages/grid/_modules_/grid/models/params/gridCellParams.ts
@@ -6,7 +6,7 @@ import { GridEditCellProps } from '../gridEditRowModel';
 /**
  * Object passed as parameter in the column [[GridColDef]] cell renderer.
  */
-export interface GridCellParams<T = GridCellValue> {
+export interface GridCellParams<T = GridCellValue, RowType = any> {
   /**
    * The grid row id.
    */
@@ -26,7 +26,7 @@ export interface GridCellParams<T = GridCellValue> {
   /**
    * The row model of the row that the current cell belongs to.
    */
-  row: GridRowModel<T>;
+  row: GridRowModel<RowType>;
   /**
    * The column of the row that the current cell belongs to.
    */
@@ -58,7 +58,7 @@ export interface GridCellParams<T = GridCellValue> {
 /**
  * GridCellParams containing api.
  */
-export interface GridRenderCellParams<T = any> extends GridCellParams<T> {
+export interface GridRenderCellParams<T = any, RowType = any> extends GridCellParams<T, RowType> {
   /**
    * GridApi that let you manipulate the grid.
    */
--- a/packages/grid/_modules_/grid/models/params/gridCellParams.ts
+++ b/packages/grid/_modules_/grid/models/params/gridCellParams.ts
@@ -6,7 +6,7 @@ import { GridEditCellProps } from '../gridEditRowModel';
 /**
  * Object passed as parameter in the column [[GridColDef]] cell renderer.
  */
-export interface GridCellParams {
+export interface GridCellParams<T = GridCellValue> {
   /**
    * The grid row id.
    */
@@ -18,15 +18,15 @@ export interface GridCellParams {
   /**
    * The cell value, but if the column has valueGetter, use getValue.
    */
-  value: GridCellValue;
+  value: T;
   /**
    * The cell value formatted with the column valueFormatter.
    */
-  formattedValue: GridCellValue;
+  formattedValue: T;
   /**
    * The row model of the row that the current cell belongs to.
    */
-  row: GridRowModel;
+  row: GridRowModel<T>;
   /**
    * The column of the row that the current cell belongs to.
    */
@@ -52,13 +52,13 @@ export interface GridCellParams {
    * @param id
    * @param field
    */
-  getValue: (id: GridRowId, field: string) => GridCellValue;
+  getValue: (id: GridRowId, field: string) => T;
 }
 
 /**
  * GridCellParams containing api.
  */
-export interface GridRenderCellParams extends GridCellParams {
+export interface GridRenderCellParams<T = any> extends GridCellParams<T> {
   /**
    * GridApi that let you manipulate the grid.
    */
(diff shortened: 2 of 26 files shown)


packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -18,15 +18,15 @@ export interface GridCellParams {
   /**
    * The cell value, but if the column has valueGetter, use getValue.
    */
-  value: GridCellValue;
+  value: T;
   /**
    * The cell value formatted with the column valueFormatter.
    */
-  formattedValue: GridCellValue;
+  formattedValue: T;
```
flaviendelangle, line 25: I don't think the formattedValue has always the same type as the value

flaviendelangle: https://github.com/final-form/react-final-form/blob/2d9e7129f40efd39450c1d581956e8cfce17a20a/typescript/index.d.ts#L90

Final form decides not to care about the formatted value in its interface

Another option would be to do `GridCellParams<T = GridCellValue, FormattedT = T, RowType = any>`
But I'm not sure it's worth it

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -6,7 +6,7 @@ import { GridEditCellProps } from '../gridEditRowModel';
 /**
  * Object passed as parameter in the column [[GridColDef]] cell renderer.
  */
-export interface GridCellParams {
+export interface GridCellParams<T = GridCellValue, RowType = any> {
```
flaviendelangle, line 9: We should limit the valid types


```suggestion
-export interface GridCellParams<T = GridCellValue, RowType = any> {
+export interface GridCellParams<T extends GridCellValue = GridCellValue, RowType = any> {
```

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -52,13 +52,13 @@ export interface GridCellParams {
...
    */
-  getValue: (id: GridRowId, field: string) => GridCellValue;
+  getValue: (id: GridRowId, field: string) => T;
 }
 
 /**
  * GridCellParams containing api.
  */
-export interface GridRenderCellParams extends GridCellParams {
+export interface GridRenderCellParams<T = any, RowType = any> extends GridCellParams<T, RowType> {
```
flaviendelangle, line 61:
```suggestion
-export interface GridRenderCellParams<T = any, RowType = any> extends GridCellParams<T, RowType> {
+export interface GridRenderCellParams<T extends GridCellValue = GridCellValue, RowType = any> extends GridCellParams<T, RowType> {
```

packages/grid/_modules_/grid/models/params/gridRowParams.ts
```diff
@@ -3,17 +3,17 @@ import { GridRowId, GridRowModel } from '../gridRows';
 import type { GridColumns } from '../colDef';
 
 /**
- * Object passed as parameter in the column [[GridColDef]] cell renderer.
+ * Object passed as parameter in the row callbacks.
  */
-export interface GridRowParams {
+export interface GridRowParams<RowDataType = any> {
```
flaviendelangle, line 8: @m4theushw do we support row that are not objects ?

If not, we should limit `RowDataType` 

```tsx
type GridRowData = { [key: string]: any }

export interface GridRowParams<RowDataType extends GridRowData = GridRowData> {
```

Same in other usages of `row`

packages/grid/_modules_/grid/models/params/gridCellParams.ts
ZeeshanTamboli: Yes, nice. So what should be the generic type for handlers like `getCellClassName`?
```
getCellClassName?: (params: GridCellParams, details?: any) => string;
```
I am assuming `GridCellValue`. Like:
```
getCellClassName?: (params: GridCellParams<GridCellValue>, details?: any) => string;
```

`GridCellValue` does not support `any`.

ZeeshanTamboli: Ignore above comment, after limiting the valid types by applying above code suggestion when I add generic to params of `getCellClassName`, why am I getting this error?

Screenshot (10)

flaviendelangle: `getCellClassName` can be called with any cell of the grid

```tsx
<DataGrid
  rows={rows}
  columns={[{ field: 'id,  type: 'number }, { field: 'name', type: 'string }]
  getCellClassName={(params) => typeof params.value} // string or number depending on the column
 />
```

So I don't think users should override the param types like this

They could use a type guard function

```ts
const isNumberCell = (params: GridCellParams): params is GridCellParams<number> => params.field !== 'city'
```

Or cast with `as`

ZeeshanTamboli: I believe same happens for other methods, as cell value can be anything. And `GridCellParams` is used in callbacks only I think. So is extending with `GridCellValue` beneficial here? 
I think we need to support `any` then? as suggested here : https://github.com/mui-org/material-ui-x/issues/1329#issuecomment-814478259.

ZeeshanTamboli: Just saw this, if its relevant: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/78d27175abe1296b87089e44fedd7493abed9dca/types/react-table/index.d.ts#L311-L317

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/78d27175abe1296b87089e44fedd7493abed9dca/types/react-table/index.d.ts#L857

packages/grid/_modules_/grid/models/gridRows.ts
```diff
@@ -1,10 +1,11 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-export type GridRowData = { [key: string]: any };
+export type GridRowType = { [key: string]: any };
+export type GridRowData<RowType = GridRowType> = RowType;
 
 /**
  * The key value object representing the data of a row.
  */
-export type GridRowModel = GridRowData;
+export type GridRowModel<RowType = any> = GridRowData<RowType>;
```
flaviendelangle, line 8: I think we should avoid doing ` export type GridRowData<RowType = GridRowType> = RowType;` 
Because we would have 3 types basically doing the same thing `GridRowData`, `GridRowType` and `GridRowModel`

I would do the following:

```tsx
export type GridRowDefaultData = { [key: string]: any };

export type GridRowModel<RowType extends GridRowDefaultData = GridRowDefaultData> = GridRowDefaultData
```

flaviendelangle:
In another PR, we should add the `RowType` generic to `GridApi`

packages/grid/_modules_/grid/models/gridRows.ts
ZeeshanTamboli: Yeah thanks, I was also thinking about what to name it. `GridRowDefaultData` sounds good.

packages/grid/_modules_/grid/models/params/gridRowParams.ts
ZeeshanTamboli: I limited row as only objects.

ZeeshanTamboli:
I have to still work on fixing the script to correctly generate the correct docs for generic types. Will be working on it.

ZeeshanTamboli:
> I have to still work on fixing the script to correctly generate the correct docs for generic types. Will be working on it.

Looks like `typedoc` does not support documenting generic types in the way we do. I looked into the `buildApi` script and it seems we have the script for generics but it does not work as expected.

eg: For `GridCellParams` with row supporting generics to type row data:

```js
 export interface GridCellParams<
  T = GridCellValue,
  RowType extends GridRowDefaultData = GridRowDefaultData,
> {
 --
 --
  value: T;
  /**
   * The cell value formatted with the column valueFormatter.
   */
  formattedValue: T;
  /**
   * The row model of the row that the current cell belongs to.
   */
  row: GridRowModel<RowType>;
 --
 --
}
```

Script which executes for generic type:
https://github.com/mui-org/material-ui-x/blob/e6cae75d7a49e889d063dba7af741503e2b240d6/docs/scripts/buildApi.ts#L49-L56

The `type.type` here for `row` is a reference which makes sense. But it is not detecting `type.typeArguments` for `GridRowModel<RowType>`.  It is undefined. 

The script is instead generating,
| <span class="prop-name">row</span>                                                           | <span class="prop-type">RowType</span>                                | The row model of the row that the current cell belongs to.       |

Same is happening for other types and for some it is picking only default generic type in the docs. 

I checked into core repo for reference but there is nothing I can find to help me.

@m4theushw Do you have any idea? You can try running `docs:api` script in this PR and see the generated docs.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -1,12 +1,15 @@
...
 import type { GridStateColDef } from '../colDef';
 import { GridEditCellProps } from '../gridEditRowModel';
 
 /**
  * Object passed as parameter in the column [[GridColDef]] cell renderer.
  */
-export interface GridCellParams {
+export interface GridCellParams<
+  T = GridCellValue,
+  RowType extends GridRowDefaultData = GridRowDefaultData,
```
m4theushw, line 11: I would reuse `GridRowModel` as much as possible.

```suggestion
  RowType extends GridRowModel = GridRowModel,
```

packages/grid/_modules_/grid/models/gridRows.ts
```diff
@@ -1,10 +1,12 @@
 export type GridRowsProp = Readonly<GridRowData[]>;
-export type GridRowData = { [key: string]: any };
+export type GridRowDefaultData = { [key: string]: any };
```
m4theushw, line 2: It's better to keep it private than to export it and had to create a breaking change in the future to remove it. 

```suggestion
type GridRowDefaultData = { [key: string]: any };
```

packages/grid/_modules_/grid/models/params/gridCellParams.ts
m4theushw: Yes, the formatted value can have other type than `T`. I agree with adding one more generic, like @flaviendelangle suggested.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -52,13 +55,16 @@ export interface GridCellParams {
    * @param id
    * @param field
    */
-  getValue: (id: GridRowId, field: string) => GridCellValue;
+  getValue: (id: GridRowId, field: string) => T;
```
m4theushw, line 58: The same situation from the `formattedValue`. Calling `getValue` allows to get the value from other fields, whose type may differ from `T`. We could add another generic, but I think there're enough. Reverting this line is fine.

```suggestion
  getValue: (id: GridRowId, field: string) => GridCellValue;
```

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -52,13 +55,16 @@ export interface GridCellParams {
...
 }
 
 /**
  * GridCellParams containing api.
  */
-export interface GridRenderCellParams extends GridCellParams {
+export interface GridRenderCellParams<
+  T = any,
```
m4theushw, line 65:
```suggestion
  T = GridCellValue,
```

m4theushw (commented):
It's good first step to improve the typings. Could we name the generics with a single letter, like React does? With so abstract names, it's hard to remember what it's a generic and what it's a type.

packages/grid/_modules_/grid/models/gridRows.ts
m4theushw: The reason for having `GridRowModel` and `GridRowData` is that, in the [past](https://github.com/mui-org/material-ui-x/pull/1377/files#diff-7f232d14dba7e4e30d0acb33b58b964bf8a8f9dd2c38b47981d98d976a336b4aL7), the first one would always contain the row id. The latter one could be removed.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
ZeeshanTamboli: If I apply this, typescript complains for `renderCell` callbacks, see https://github.com/mui-org/material-ui-x/pull/2436#discussion_r694634035 thread for more details. 

But if we keep `any` type, then [tests fail](https://app.circleci.com/pipelines/github/mui-org/material-ui-x/7753/workflows/aeec0160-2c39-4738-a417-36eab53d5890/jobs/42873) for `GridActionsCell` `getActions` after regenerating propTypes for `formattedValue` and `value` fields. 

What should we do? Should we make `formattedValue` and `value` optional in `GridActionsCellProps` props?

PR #2441, opened 2021-08-24
[docs] Use same infra from v5

DanailH:
Based on this commit https://github.com/mui-org/material-ui-x/commit/50a16a1575b790effb0991c006d972cd9f5f7cb1

Preview: https://deploy-preview-2441--material-ui-x.netlify.app/components/data-grid/


oliviertassinari:
I think that we need to CI to be green before we can review the changes.

oliviertassinari:
Also, the title of the PR could likely be improved fixing Algolia doesn't seem to be the main purpose, it seems to be mostly about updating the docs infra from master to next, what @flaviendelangle wanted to focus on?

oliviertassinari:
If we only want to fix Algolia, we would need to unwrap https://github.com/mui-org/material-ui-x/blob/master/docs/pages/_document.js

DanailH:
I was trying to make the docs work, compile. So this is that effort

m4theushw:
Finally the docs is using the same infra from the core v5. There're still some small things that I'll fix in a follow-up. This PR is already giant.

- Trying to edit the Filled Quantity in https://deploy-preview-2441--material-ui-x.netlify.app/components/data-grid/demo/ crashes because it can't find the element to focus
- Since the size of the IconButton increased in v5, the title of the columns on some demos is cutted: https://github.com/mui-org/material-ui-x/pull/2515#issuecomment-912521097
- Although it runs the demos with v5, opening them in CodeSandBox will still use the old @material-ui/core

docs/next.config.js
```diff
@@ -16,11 +16,17 @@ const workspaceRoot = path.join(__dirname, '../');
...
-console.log(`Using React '${reactMode}' mode.`);
+const reactStrictMode = true;
+if (reactStrictMode) {
+  // eslint-disable-next-line no-console
+  console.log(`Using React.StrictMode.`);
+}
 
 module.exports = {
+  eslint: {
+    // TODO: https://github.com/mui-org/material-ui/issues/25966
```
oliviertassinari, line 27: The issue was closed, we do no longer need this comment. 

```suggestion
```

https://nextjs.org/docs/api-reference/next.config.js/ignoring-eslint

> This is not recommended unless you already have ESLint configured to run in a separate part of your workflow (for example, in CI or a pre-commit hook).

We already run eslint, no need to do it twice. I will remove this comment too in the core.

docs/next.config.js
```diff
@@ -60,32 +67,36 @@ module.exports = {
...
+    //     if (typeof external !== 'function') return external;
+    //     return (ctx, req, cb) => {
+    //       return includesMonorepo.find((include) =>
+    //         req.startsWith('.') ? include.test(path.resolve(ctx, req)) : include.test(req),
+    //       )
+    //         ? cb()
+    //         : external(ctx, req, cb);
+    //     };
+    //   });
+    // }
```
oliviertassinari, line 81:
```suggestion
```

docs/package.json
```diff
@@ -9,7 +9,7 @@
     "build:clean": "rimraf .next && yarn build",
     "build-sw": "node ./scripts/buildServiceWorker.js",
     "dev": "rimraf ./node_modules/.cache/babel-loader && next dev --port 3001",
-    "deploy": "git push upstream master:release",
+    "deploy": "git push upstream next:release",
```
oliviertassinari, line 12: I assume we need a new branch for the docs of for v5:
```suggestion
    "deploy": "git push upstream next:docs-v5",
```

docs/pages/_document.js
```diff
@@ -1,3 +1,3 @@
-import MyDocument from '@material-ui/monorepo/docs/pages/_document';
+import MyDocument from 'docs/pages/_document';
```
oliviertassinari, line 1: What do you think about using `'docs/'` as little as possible (only to avoid import crashes for the import from the main repo). I find it confusing that a file import itself (it seems so)

```suggestion
import MyDocument from '@mui/monorepo/docs/pages/_document';
```

docs/pages/index.js
```diff
@@ -1,3 +1,3 @@
-import LandingPage from '@material-ui/monorepo/docs/pages/index';
+import LandingPage from 'docs/pages/index';
```
oliviertassinari, line 1: Same comment as above regarding self referring imports

oliviertassinari (commented):
Looks like a great first iteration. This one required to dive deep into the structure of the repositories, well done

oliviertassinari:
Also, it's awesome that we finally reconcile the docs of MUI X with MUI Core 🎉. A couple of feedback to the core:

1. I think that we have the same side nav issue as with the date picker. It's not very clear what "Demo" refers to. cc @danilo-leal I didn't follow what was the conclusion to it:

Screenshot 2021-09-04 at 13 20 49

https://deploy-preview-2441--material-ui-x.netlify.app/components/data-grid/demo/

2. Regarding the not correctly themed links in https://deploy-preview-2441--material-ui-x.netlify.app/components/data-grid/events/#catalog-of-events. Maybe we need to expose a theme provider to wrap this demo cc @mnajdova, @siriwatknp.

Screenshot 2021-09-04 at 13 26 38

danilo-leal:
> 1. I think that we have the same side nav issue as with the date picker. It's not very clear what "Demo" refers to. cc @danilo-leal I didn't follow what was the conclusion to it:

Let me quickly separate the answer into two topics:
1. **Visual**: in regards to the visual conclusion of nested levels, we haven't yet implemented it yet but I left [a simple suggestion](https://github.com/mui-org/material-ui/pull/27956#issuecomment-909452788) on the previous Markdown polish PR for addressing it.
2. **Content**: we decided to create a new PR and not use the Markdown polish PR for rearranging the grouping of links on the side nav. We also haven't done it yet. But reflecting specifically about the Data Grid Demo page, I'm not sure if we need to have it stand-alone, it seems weird. I'd think that it makes more sense to have a Demo section probably on the Getting Started page or something like that. I went to GA looking for pageviews and, in the last month, Demo had 18k and Getting Started had 16k. They're not too far away so I suppose it wouldn't be that much of a problem to merge them?!

Anyway, I think we should probably create a second Docs polishments PR for addressing these two (visual solution for nested levels on the sidebar and content regrouping) and other small details left open from the first PR. We can move these discussions further there. It would be nice to do that before v5 stable. How do you feel about it? @mnajdova @siriwatknp

docs/pages/_app.js
```diff
@@ -43,12 +53,9 @@ ponyfillGlobal.muiDocConfig = {
...
     const newDeps = { ...deps };
 
-    if (
-      newDeps['@material-ui/x'] ||
-      newDeps['@mui/x-data-grid-pro'] ||
-      newDeps['@mui/x-data-grid']
-    ) {
+    if (newDeps['@mui/x-data-grid-pro'] || newDeps['@mui/x-data-grid']) {
       newDeps['@material-ui/core'] = versions['@material-ui/core'];
+      newDeps['@material-ui/styles'] = versions['@material-ui/styles'];
```
flaviendelangle, line 58: Should we switch to `@mui` for core repo on this file since we installed them ?

https://github.com/mui-org/material-ui-x/pull/2441/files#diff-adfa337ce44dc2902621da20152a048dac41878cf3716dfc4cc56d03aa212a56R28

docs/package.json
```diff
@@ -72,6 +74,10 @@
     "redux": "^4.1.1",
```
flaviendelangle, line 74: I think we can remove Redux

packages/eslint-plugin-material-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-material-ui",
-  "version": "4.0.0-alpha.37",
+  "version": "5.0.0-beta.0",
```
flaviendelangle, line 3: Aren't we supposed to only modified these versions just before releasing ?

docs/pages/_app.js
```diff
@@ -212,6 +224,18 @@ function Analytics() {
...
+let reloadInterval;
+
+// Avoid infinite loop when "Upload on reload" is set in the Chrome sw dev tools.
+function lazyReload() {
+  clearInterval(reloadInterval);
+  reloadInterval = setInterval(() => {
+    if (document.hasFocus()) {
+      window.location.reload();
+    }
+  }, 100);
```
DanailH, line 236: 0 will probably also work since you just want the current stack to be empy.

m4theushw: I just copied this file from the core and pasted it here. To be honest, I don't know what this function does.

packages/eslint-plugin-material-ui/package.json
m4theushw: I reverted all these version bumps.

docs/pages/_document.js
m4theushw: I replaced in all files that were already modified. The remaining will be updated latter.

flaviendelangle:
image

Why did we rename the "Overview" page "Overview" ? It's creating visual redundancy.

PR #2442, opened 2021-08-24
[core] Allow to create one logger per Grid instance

flaviendelangle:
- [x] Two Grids should be able to have different log level and logger
- [x] The logger factory should not be called on every render

### DIFF
--- a/packages/grid/_modules_/grid/hooks/utils/useGridLogger.ts
+++ b/packages/grid/_modules_/grid/hooks/utils/useGridLogger.ts
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { Logger } from '../../models/logger';
 import { localStorageAvailable } from '../../utils/utils';
 import { GridComponentProps } from '../../GridComponentProps';
+import { GridApiRef } from '../../models/api/gridApiRef';
+import { GridLoggerApi } from '../../models/api/gridLoggerApi';
 
 const forceDebug = localStorageAvailable() && window.localStorage.getItem('DEBUG') != null;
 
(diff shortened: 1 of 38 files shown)


packages/grid/_modules_/grid/hooks/utils/useGridLogger.ts
```diff
@@ -39,35 +41,35 @@ function getAppender(name: string, logLevel: string, appender: Logger = console)
...
-  apiRef: any,
+export function useGridLoggerFactory(
+  apiRef: GridApiRef,
   props: Pick<GridComponentProps, 'logger' | 'logLevel'>,
 ) {
-  if (forceDebug) {
-    factory = defaultFactory('debug');
-    return;
-  }
+  apiRef.current.getLogger = React.useCallback<GridLoggerApi['getLogger']>(
```
flaviendelangle, line 48: I can't use `useGridApiMethod` because it's using `useGridLogger` so it's initialized after.
We can either 

- remove the logger usage in `useGridApiMethod`

- add manually the method to `apiRef` (like in this PR)

- do not use `apiRef` to store the callback (don't like this one)

```ts
const loggers: { [gridId: number]: LoggerFactoryFn } = {}

const useGridLoggerFactory = (apiRef, props) => {
  const id = useUniqId()

  if (!loggers[id]) {
    loggers[id] = createLogger()
    apiRef.id = id
  }
}

const useGridLogger = (apiRef, name) => {
  const getLogger = loggers[apiRef.id]
  ...
}
``` 

- create the logger in a component above `GridComponent` and expose it through a React context. This would add a React component to the tree but would allow to reuse the same logger system for other components.

flaviendelangle:
:roll_eyes: I merged it on master...
Doing a cherry-pick on next

PR #2443, opened 2021-08-25
[DataGridPro] Only apply checkboxSelectionVisibleOnly when pagination is enabled

### LINKED ISSUE #2386: [DataGridPro] When checkboxSelectionVisibleOnly is enabled, select all checkbox selection is limited to the first 100 rows.
We notice since upgrading from alpha.31, that now when we use the select all when checkboxSelectionVisibleOnly is enabled, it'll only select the first 100 rows.  Demo attached...

## Current Behavior 😯

Only the first 100 rows will be selected.

## Expected Behavior 🤔

All rows (up to 100000) should be selected.

## Steps to Reproduce 🕹

https://codesandbox.io/s/material-demo-100-row-selection-bug-7w0lk

Steps:

1. Select all rows using the checkbox in the header.

## Context 🔦

I'm trying to select all rows or all visible rows after filtering and if there are more then 100 this is problematic. 

## Your Environment 🌎

System:
    OS: macOS 11.2.2
    CPU: (8) x64 Apple M1
    Memory: 84.03 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.1 - ~/.nvm/versions/node/v14.15.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.15.1/bin/npm
  Managers:
    Composer: 1.6.5 - /usr/local/bin/composer
    Homebrew: 2.7.7 - /usr/local/bin/brew
    pip3: 21.0.1 - /usr/local/bin/pip3
  Utilities:
    CMake: 3.19.4 - /usr/local/bin/cmake
    Make: 3.81 - /usr/bin/make
    GCC: 4.2.1 - /usr/bin/gcc
    Git: 2.30.0 - /usr/local/bin/git
    Clang: 12.0.5 - /usr/bin/clang
    Ninja: 1.10.2 - /usr/local/bin/ninja
  Servers:
    Apache: 2.4.46 - /usr/local/bin/apachectl
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
  IDEs:
    Atom: 1.58.0
    Nano: 2.0.6 - /usr/bin/nano
    Vim: 8.2 - /usr/bin/vim
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Perl: 5.28.2 - /usr/bin/perl
    PHP: 8.0.1 - /usr/local/bin/php
    Python3: 3.8.2 - /usr/bin/python3
    Ruby: 2.6.3 - /usr/bin/ruby
  Databases:
    MySQL: 10.5.6 (MariaDB) - /usr/local/bin/mysql
    SQLite: 3.32.3 - /usr/bin/sqlite3
  Browsers:
    Chrome: 92.0.4515.159
    Chrome Canary: 95.0.4612.2
    Firefox: 87.0
    Safari: 14.0.3

## Order id 💳

Order #25887

flaviendelangle:
Closes #2386


packages/grid/_modules_/grid/hooks/features/pagination/gridPaginationSelector.ts
```diff
@@ -2,15 +2,21 @@ import { createSelector } from 'reselect';
...
 export const gridPaginationSelector = (state: GridState): GridPaginationState => state.pagination;
 
 export const gridPaginatedVisibleSortedGridRowIdsSelector = createSelector(
   gridPaginationSelector,
   visibleSortedGridRowIdsSelector,
-  (pagination, visibleSortedRows) => {
+    gridContainerSizesSelector,
+    (pagination, visibleSortedRows, containerSizes) => {
+      if (!containerSizes) {
+          return []
```
flaviendelangle, line 15: I can throw if you prefer
The selector is never used in a render, but if it was I prefer not to have a buggy behavior like `containerSizes!.virtualRowsCount`

m4theushw (commented):
Now, having `checkboxSelectionVisibleOnly` true or false doesn't any make difference. It selects all 100k rows: https://codesandbox.io/s/material-demo-100-row-selection-bug-forked-jy731?file=/package.json

I don't think the behavior of this prop is right. When the pagination is disabled, it should select only the portion of the rows that are really visible, the virtual page is irrelevant here. The description should reflect that. You could use `containerProps.viewportPageSize` to know how many rows fit in.

https://github.com/mui-org/material-ui-x/blob/8eabe16e899e3eaed75325442c1bdceeadd7da32/packages/grid/_modules_/grid/models/gridOptions.tsx#L54

flaviendelangle:
For a grid with pagination and scroll, you would also select only the ones in the viewport ?

m4theushw:
Looking at issue #605 that was the motivation to introduce `checkboxSelectionVisibleOnly` in #1998, I noticed that we only cared about the scenario where the pagination is enabled. Nothing was done for when the pagination is disabled. The user in https://github.com/mui-org/material-ui-x/issues/2386#issue-974199924 asked that all the row set should be selected, but this can already be achieved with `checkboxSelectionVisibleOnly={false}`. 

Since I didn't find any other grid that implement the behavior I proposed in https://github.com/mui-org/material-ui-x/pull/2443#pullrequestreview-742081266, I think we just need to improve the prop description and not use it when the pagination is disabled. We could add a warning in the propTypes. We could also ask the user if just `checkboxSelectionVisibleOnly={false}` is not enough for his use case. To better picture https://github.com/mui-org/material-ui-x/pull/2443#pullrequestreview-742081266, I was proposing the following:

Pagination enabled
- `checkboxSelectionVisibleOnly={true}` selects all rows in the current page
- `checkboxSelectionVisibleOnly={false}` selects all rows in the all pages

Pagination disabled
- `checkboxSelectionVisibleOnly={true}` selects only the rows fully visible in the viewport
- `checkboxSelectionVisibleOnly={false}` selects all rows

flaviendelangle:
It seems a good approach to me
I modified my PR

oliviertassinari (commented):
Does `checkboxSelectionVisibleOnly` impact the selection behavior when filtering is applied?

Edit: no, related issue https://github.com/mui-org/material-ui-x/issues/1141.

Edit 2: I personally find the next path forward blurry. Maybe we need to:

- Rename the `checkboxSelectionVisibleOnly` prop to be more specific: `checkboxSelectionActivePageOnly`
- Add a warning when the prop is used without pagination
- Add a `checkboxSelectionFilteredRowsOnly` to allow to differentiate the behavior for filtering.

packages/grid/x-grid/src/tests/selection.DataGridPro.test.tsx
```diff
@@ -333,5 +334,31 @@ describe('<DataGridPro /> - Selection', () => {
...
+
+    it('should select all the lines when using infinite scroll', () => {
+      const SelectAllTest = () => {
+        apiRef = useGridApiRef();
+        const data = useData(1000, 2);
+        return (
+          <div style={{ height: 300, width: 300 }}>
+            <DataGridPro
+              {...data}
+              loading={data.rows.length === 0}
```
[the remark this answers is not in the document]
flaviendelangle: It was a hack to have a synchronous initialization of the data, but I did better in my last commit :+1: 

Pagination enabled

- `checkboxSelectionVisibleOnly`={true} selects all rows in the current page
- `checkboxSelectionVisibleOnly`={false} selects all rows in all pages

Pagination disabled
- `checkboxSelectionVisibleOnly`={} selects all rows in all pages
- `checkboxSelectionVisibleOnly`={false} selects all rows in all pages

packages/grid/x-grid/src/tests/selection.DataGridPro.test.tsx
```diff
@@ -70,9 +69,35 @@ describe('<DataGridPro /> - Selection', () => {
...
+          checkboxSelectionVisibleOnly={false}
+          rowLength={100}
+        />,
+      );
+
+      const selectAll = screen.getByRole('checkbox', {
+        name: /select all rows checkbox/i,
+      });
+      fireEvent.click(selectAll);
+      expect(apiRef.current.getSelectedRows().size).to.equal(100);
```
m4theushw, line 88:
```suggestion
      expect(apiRef.current.getSelectedRows()).to.have.length(100);
```

packages/grid/x-grid/src/tests/selection.DataGridPro.test.tsx
```diff
@@ -70,9 +69,35 @@ describe('<DataGridPro /> - Selection', () => {
       );
       const selectAllCheckbox = document.querySelector('input[type="checkbox"]');
       fireEvent.click(selectAllCheckbox);
-      expect(getRow(0)).to.have.class('Mui-selected');
-      fireEvent.click(screen.getByRole('button', { name: /next page/i }));
-      expect(getRow(1)).not.to.have.class('Mui-selected');
+      expect(Array.from(apiRef.current.getSelectedRows().keys())).to.deep.equal([0]);
```
m4theushw, line 72:
```suggestion
      expect(apiRef.current.getSelectedRows()).to.have.keys([0]);
```

packages/grid/x-grid/src/tests/selection.DataGridPro.test.tsx
```diff
@@ -70,9 +69,35 @@ describe('<DataGridPro /> - Selection', () => {
...
+      );
+
+      const selectAll = screen.getByRole('checkbox', {
+        name: /select all rows checkbox/i,
+      });
+      fireEvent.click(selectAll);
+      expect(apiRef.current.getSelectedRows().size).to.equal(100);
+    });
+
+    it('should select all the lines if checkboxSelectionVisibleOnly = true and pagination is not enabled', () => {
```
m4theushw, line 91: This title doesn't reflect what the test is asserting.

```suggestion
    it('should throw a console error if checkboxSelectionVisibleOnly is used without pagination', () => {
```

flaviendelangle: Oh I forgot to rename this one right

m4theushw (commented):
We need to update the prop description in https://deploy-preview-2443--material-ui-x.netlify.app/api/data-grid/data-grid-pro/.

m4theushw (approved):

packages/grid/x-grid/src/DataGridPro.tsx
```diff
@@ -57,4 +57,12 @@ export const DataGridPro = React.memo(DataGridProRaw);
 DataGridProRaw.propTypes = {
   columns: PropTypes.array.isRequired,
   rows: PropTypes.array.isRequired,
+  checkboxSelectionVisibleOnly: chainPropTypes(PropTypes.bool, (props: any) => {
+    if (!props.pagination && props.checkboxSelectionVisibleOnly) {
+      return new Error(
+        'Material-UI: The `checkboxSelectionVisibleOnly` prop has no effect when the pagination is not enabled.',
```
oliviertassinari, line 63: A detail. I would advocate that a great warning starts by explaining what's wrong, and then explain how to solve the issue. We did the first one, but not the second. For instance, see how [React handles it](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/scripts/error-codes/codes.json#L209).

ungurnicoleta:
@flaviendelangle this one is still not working for me, even if I've enabled `pagination` and `checkboxSelection`

flaviendelangle:
The fix was applied in `5.0.0-beta.0`  :+1:

ungurnicoleta:
Is it stable? I see that the lastest on npm is `4.0.0-alpha.35`

flaviendelangle:
We published the `v4.0.0` which is the last version of the DataGrid compatible with the ecosystem Material-UI v4 
We are now focusing on the `v5.0.0` version (currently `v5.0.0-beta.1`) which is as stable as `v4.0.0` but we want to rework a few important parts of the codebase before launching the `v5.0.0` stable version, to avoid future breaking change.

If you are using the Material-UI (now MUI) ecosystem in v5, I would suggest you migration to the `v5.0.0-beta.1` from DataGrid.
If you are using the Material-UI ecosystem in v4, you are stuck with `v4.0.0` of the DataGrid. We will provide fixes on this version (this `checkboxSelectionVisibleOnly` could be applied on `v4.0.0` is their is a demand) but no major change.

Hope this will make it clearer.

flaviendelangle:
@ungurnicoleta you should have the warning in the version `v4.0.1` of the DataGrid :+1:

zxh19890103:
Would @material-ui/data-grid always be wrong code running on user's end?

flaviendelangle:
What do you mean by "wrong code running" ?

zxh19890103:
> What do you mean by "wrong code running" ?

Sorry for my rude words. Actually, I mean that, in general, users would expect the click on header checkbox leads only those rows which are visible being checked. But DataGrid checks all items it has in state. It's not wrong, but i feel it's not natural, at least for me.  I'm using "@material-ui/data-grid": "^4.0.0-alpha.31", i tried migrating it to alpha.41 which is the last version on 4.0.0, but this causes many `errors of typescript`. So I will still use alpha.31 for a long time, which means I must implement this `checkboxSelectionVisibleOnly` feature by my own.

This makes me unhappy. ~~

flaviendelangle:
If you are blocked in an alpha version and can't migrate to a more recent alpha version, you won't have access to the features released after.

I hope your migration is not that difficult, it is unfortunately the drawback of an alpha phase, we changed a lot of things to provide the best product possible when releasing it as stable.

PR #2456, opened 2021-08-26
[DataGrid] Allow to select range of rows with shift key

### LINKED ISSUE #1858: [DataGrid] Extend selection with Shift + click

## Summary 💡

Selecting rows while pressing <kbd>Shift</kbd> should select all rows between the last selected and the clicked one.

## Examples 🌈

https://marmelab.com/react-admin-demo/

Ze0AFwvmb0

## Motivation 🔦

https://github.com/mui-org/material-ui-x/pull/1813#discussion_r646925340

flaviendelangle:
- [x] Handle keyboard selection expansion / reduction with `Shift`
- [x] Handle click selection expansion / reduction with `Shift`
- [x] New method `apiRef.current.selectRowRange` to handle in a single place the keyboard and mouse range selection behavior

Closes #1858

Preview: https://deploy-preview-2456--material-ui-x.netlify.app/components/data-grid/selection


packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +167,43 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
+      if (canHaveMultipleSelection && event.shiftKey) {
+        apiRef.current.selectRowRange(
+          {
+            startId: lastRowToggledByClick.current ?? params.id,
+            endId: params.id,
+          },
+          !apiRef.current.isRowSelected(params.id),
         );
+
+        event.stopPropagation();
```
m4theushw, line 179: Was it added to prevent the row highlighting during the click? I believe that with the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection/removeAllRanges) we can do that.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -112,6 +114,35 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
     [apiRef, isRowSelectable, logger, canHaveMultipleSelection],
   );
 
+  const selectRowRange = React.useCallback<GridSelectionApi['selectRowRange']>(
+    (
+      {
+        startId,
+        endId,
+      }: {
```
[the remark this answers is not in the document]
flaviendelangle: It was to be consistent with the other methods that accepts the IDs
But we could switch to the index on this one, or accept both

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: `removeAllRanges` works great if applied on `cellKeyDown`
If applied on `rowClick`, we have a selection blink during the click.

image

The compatibility seems quite good

flaviendelangle: See https://github.com/mui-org/material-ui-x/pull/2456/commits/89d1567649254fd73dc476c7b2ceace23d2e6a1c

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +167,43 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
-
-      if (resetSelection) {
-        apiRef.current.selectRow(
-          params.id,
-          hasCtrlKey || checkboxSelection ? !apiRef.current.isRowSelected(params.id) : true,
-          true,
+      if (canHaveMultipleSelection && event.shiftKey) {
+        apiRef.current.selectRowRange(
+          {
+            startId: lastRowToggledByClick.current ?? params.id,
```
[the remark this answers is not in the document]
flaviendelangle: My implementation is really simple for now.
To have something similar to excel (or even the GDrive home which is my testing page), I would need to now if the row I'm clicking on is inside the same selection range as the last row clicked

If true, then we unselect the rows from the last selected row to the row after the clicked row (in the direction of the reduction).

If false, then we expand the range to the clicked row (whether or not it is selected)

But I think it make sense to 1st release a simple version of the range selection and then do a 2nd run to improve the exact behavior.

m4theushw (commented):
We should also extend selection when doing <kbd>Shift</kbd> + click with `checkboxSelection`.

rFIbpGAQzY

https://material-ui.com/components/data-grid/accessibility/#selection can be updated as well to mention the new shortcuts.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +167,53 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
+    (_, event: React.MouseEvent) => {
+      if (canHaveMultipleSelection && event.shiftKey) {
+        window.getSelection()?.removeAllRanges();
+      }
+    },
+    [canHaveMultipleSelection],
+  );
+
   useGridApiEventHandler(apiRef, GridEvents.rowClick, handleRowClick);
+  useGridApiEventHandler(apiRef, GridEvents.rowMouseDown, preventSelectionOnShift);
```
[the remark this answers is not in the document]
flaviendelangle: Are the cell covering 100% of the rows in any circumstances ?
If so we can, I did not want to have an edge case where you click in the only pixel of the rows not in any cell and it does not reset the browser selection.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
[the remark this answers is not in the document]
flaviendelangle: There are probably edge cases but it works a lot better than my implementation :+1:

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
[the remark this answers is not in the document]
m4theushw: Yes, there's no way to click the row and not the cell. Any padding or border is always applied to the cell. The row has now "visible" styles.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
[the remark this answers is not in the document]
m4theushw: Maybe it's time to also accepts indexes in `selectRowRange`. We could reuse the logic for the keyboard navigation.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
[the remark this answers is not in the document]
flaviendelangle: Done :+1:

flaviendelangle:
@mui-org/x can we merge this PR once the release is done ?

m4theushw:
> @mui-org/x can we merge this PR once the release is done ?

I didn't approve it yet because the <kbd>Shift</kbd> + click was not working with `checkboxSelection`. Do you plan to work on https://github.com/mui-org/material-ui-x/pull/2456#pullrequestreview-754251554?

flaviendelangle:
> > @mui-org/x can we merge this PR once the release is done ?
> 
> I didn't approve it yet because the <kbd>Shift</kbd> + click was not working with `checkboxSelection`. Do you plan to work on https://github.com/mui-org/material-ui-x/pull/2456#pullrequestreview-754251554?

I was convinced I fixed this one...my bad if it is not. I will take a look

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
```diff
@@ -33,12 +33,18 @@ const GridCellCheckboxForwardRef = React.forwardRef<HTMLInputElement, GridCellPa
...
-      apiRef.current.selectRow(id, event.target.checked, false);
+      if (shiftRef.current) {
+        apiRef.current.expandRowRangeSelection(id);
+      } else {
+        apiRef.current.selectRow(id, event.target.checked);
+      }
     };
 
     const handleClick = (event: React.MouseEvent<HTMLInputElement>) => {
+      shiftRef.current = event.shiftKey;
```
flaviendelangle, line 47: The `onChange` does not have access to `shiftKey`
The `onClick` seems to always be called before `onChange` but if you have an alternative I'm interested. I did not wanted to listen to `keyDown` on every checkbox of the Grid.

flaviendelangle:
@m4theushw https://github.com/mui-org/material-ui-x/pull/2456#pullrequestreview-754251554 is done :+1:

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
```diff
@@ -21,44 +20,34 @@ import { GridCellModes } from '../../../models/gridEditRowModel';
...
 
-      const currentRowIndex = Number(rowEl.getAttribute('data-rowindex'));
-      let selectionFromRowIndex = currentRowIndex;
-
-      // TODO Refactor here to not use api call
-      const selectedRowsIds = [...apiRef.current.getSelectedRows().keys()];
-      if (selectedRowsIds.length > 0) {
-        const selectedRowsIndex = selectedRowsIds.map((id) => apiRef.current.getRowIndex(id));
+      const startRowIndex = Number(rowEl.getAttribute('data-rowindex'));
+      const startId = apiRef.current.getRowIdFromRowIndex(startRowIndex);
```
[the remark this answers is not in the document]
m4theushw: Selecting filtered rows with the keyboard is not working. The ID has to come from the selector, not from `apiRef.current.getRowIdFromRowIndex`.

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
```diff
@@ -33,12 +33,18 @@ const GridCellCheckboxForwardRef = React.forwardRef<HTMLInputElement, GridCellPa
 
     const handleRef = useForkRef(checkboxElement, ref);
     const element = apiRef.current.getCellElement(id, field);
+    const shiftRef = React.useRef(false);
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-      apiRef.current.selectRow(id, event.target.checked, false);
+      if (shiftRef.current) {
+        apiRef.current.expandRowRangeSelection(id);
```
m4theushw, line 40: I don't get why we should expose this method in the public API. You could listen for the `cellClick` event in `useGridSelection` and check if the clicked cell is the one with the checkbox.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +192,44 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
-      const resetSelection = !canHaveMultipleSelection || isMultipleSelectionDisabled;
-
-      if (resetSelection) {
-        apiRef.current.selectRow(
-          params.id,
-          hasCtrlKey || checkboxSelection ? !apiRef.current.isRowSelected(params.id) : true,
-          true,
-        );
+      if (event.shiftKey && (canHaveMultipleSelection || checkboxSelection)) {
+        apiRef.current.expandRowRangeSelection(params.id);
```
m4theushw, line 196: This method could be only available inside the hook.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +192,44 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
+        } else {
+          apiRef.current.selectRow(params.id, !apiRef.current.isRowSelected(params.id), false);
+        }
       }
     },
     [apiRef, canHaveMultipleSelection, disableSelectionOnClick, checkboxSelection],
   );
 
+  const preventSelectionOnShift = React.useCallback(
+    (_, event: React.MouseEvent) => {
```
m4theushw, line 217:
```suggestion
    (params: GridCellParams, event: React.MouseEvent) => {
```

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
m4theushw: My alternative is to listen for the cell click event in the hook. https://github.com/mui-org/material-ui-x/pull/2456#discussion_r718031822

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
flaviendelangle: We stop the propagation in the checkbox component.
Probably to avoid double call when clicking on the checkbox
And if we just keep the cell click and remove the checkbox, then it would not work with the other types of interaction that triggers the `onChange`.
We could replicate them (the `Enter` keydown for instance) but I don't think it's a good pattern.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: As explained in https://github.com/mui-org/material-ui-x/pull/2456#discussion_r718209180, I did not find a clean way to keep it inside the `useGridSelection` hook

I could remove it from the documentation.
But for me it's not that strange to have a method saying "apply the range selection logic between this row and the last interacted row".

docs/src/pages/components/data-grid/getting-started/getting-started.md
```diff
@@ -149,7 +149,7 @@ If you have an enterprise grid running with an expired or missing license key, t
 
 The following table summarizes the features available in the community `DataGrid` and enterprise `DataGridPro` components.
 All the features of the community version are available in the enterprise one.
-The enterprise components come in two plans: Pro and Premium.
+The enterprise components come in two plans: Pro and Premium.https://www.twitch.tv/directory/game/Art
```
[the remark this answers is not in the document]
flaviendelangle: :D That was totally intented

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
flaviendelangle: We could also remove the `stopPropagation` from the checkbox `onClick` handler and prevent the double call in another way. But my ideas to do it are hacky.

m4theushw: > Add a rowSelectionCheckboxChange event and just fire the event in the checkbox component, all the behavior could then be handled in useGridSelection

This one sounds better. It avoids adding another method to the API and keeps everything related to selection in one place.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
m4theushw: The problem is that its use case is too specific.

packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
flaviendelangle: PR updated with an event-based implementation

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
```diff
@@ -52,15 +51,12 @@ const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnHeaderPa
     const isChecked = (totalSelectedRows > 0 && totalSelectedRows === totalRows) || isIndeterminate;
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-      const checked = event.target.checked;
+      const params: GridHeaderSelectionCheckboxParams = {
+        value: event.target.checked,
+        indeterminate: event.target.indeterminate,
```
m4theushw, line 56: It's always false.

image

The `indeterminate` prop is not apply to the input: https://github.com/mui-org/material-ui/blob/3e679ac9e368aeb170d564d206d59913ceca7062/packages/mui-material/src/Checkbox/Checkbox.js#L79

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -136,29 +209,107 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
...
+    document.addEventListener('keyup', handleKeyUp);
+
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown);
+      document.removeEventListener('keyup', handleKeyUp);
+    };
+  }, []);
+
+  const handleRowSelectionCheckboxChange = React.useCallback(
+    (params: GridRowSelectionCheckboxParams) => {
```
m4theushw, line 263: You can know if <kbd>Shift</kbd> is pressed using `event.nativeEvent.shiftKey`.

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
flaviendelangle: I copy pasted the old behavior.
But I can switch to `isIndeterminate`

flaviendelangle: In https://github.com/mui-org/material-ui-x/issues/2672, if you have `shouldLimitSelectionToCurrentPage = false` and `props.pagination = true`, you are still using  `gridPaginatedVisibleSortedGridRowIdsSelector`

m4theushw: There's a small regression here. With `checkboxSelectionVisibleOnly=false` it should display the header checkbox in an indeterminate state when going to the second page. Could you revert the part where you're only considering the rows in the current page when deciding whether is indeterminate or not? I forgot about this in https://github.com/mui-org/material-ui-x/issues/2672#issuecomment-928132290. 

Also, I guess we need to always pass true to the 3rd argument of `apiRef.current.selectRows` to unselect all rows in all pages when `checkboxSelectionVisibleOnly=false`. #2672 will fix the behavior when it's `true`.

hCYScjeaBL

flaviendelangle: @m4theushw I think with my latest commit it works without the 3rd argument.
If you have a better name for `currentSelectionChunkSelectedCount`, my idea is to compute the list of the IDs that the checkbox can toggle and then only base the isChecked / isIndeterminate on this list.

m4theushw: As discussed during the checkpoint meeting, let's keep the old behavior https://github.com/mui-org/material-ui-x/pull/2456#discussion_r720479373, even though it's not working as it should. Later, in #2672 it can be fixed.

flaviendelangle: @m4theushw done

m4theushw (approved):

PR #2320, opened 2021-08-10
[DataGrid] Replace hard coded classes with gridClasses

m4theushw:
Closes #1950

The tests will be updated in a second PR, including the removal of the `cssClassesContants.ts` file.

PR #2458, opened 2021-08-26
[DataGrid] Add missing keys to the `classes` prop

### LINKED ISSUE #420: [DataGrid] Missing keys in classes prop
As shown on the image below, all rule names (except `root`) for [DataGrid](https://material-ui.com/api/data-grid/#css) and [XGrid](https://material-ui.com/api/x-grid/#css) are missing.

image

This is causing me some consternation since I'm using the docs to automatically create F# Fable bindings for MUI. Seems like a quick fix, though. :)

m4theushw:
### Breaking changes

- [DataGrid] Rename the `.Mui-resizing` CSS class to `.MuiDataGrid-columnSeparator`.

---

Preview: https://deploy-preview-2458--material-ui-x.netlify.app/api/data-grid/data-grid/#css

Closes #420 

The docs generator uses the `classes` prop to extract information for the [CSS section](https://material-ui.com/api/data-grid/data-grid/#css). Since we were supporting only a few keys, we would not be able to document all the classes available. This PR adds these missing keys. In components where multiple classes are applied depending on conditions, I implemented the `useUtilityClasses` pattern used in the core, but in most cases I went with `clsx`.

Part of this PR will conflict with #2433, specially where we replace "__" with "--". #2433 introduced a single hook to allow to pre-process the prop values, but here I duplicated the code between DataGrid and DataGridPro.

I renamed `.MuiDataGrid-columnSeparator--dragging` to `.MuiDataGrid-columnHeader--dragging` because in #2320 I used the wrong name: https://github.com/mui-org/material-ui-x/pull/2320/files#diff-0fae50bd4900beba39da8e79eedc2a3655445f14682f842c89b65b49e1a622dfL67


packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -17,12 +17,31 @@ export interface GridRowProps {
   children: React.ReactNode;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 20: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/GridScrollArea.tsx
```diff
@@ -13,6 +14,16 @@ interface ScrollAreaProps {
   scrollDirection: 'left' | 'right';
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 17: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -32,10 +34,23 @@ export interface GridCellProps {
   tabIndex: 0 | -1;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 37: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/columnHeaders/ColumnHeaderFilterIcon.tsx
```diff
@@ -5,15 +5,31 @@ import Tooltip from '@material-ui/core/Tooltip';
...
-import { gridClasses } from '../../gridClasses';
+import { getDataGridUtilityClass } from '../../gridClasses';
+import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
+import { composeClasses } from '../../utils/material-ui-utils';
 
 export interface ColumnHeaderFilterIconProps {
   counter?: number;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 16: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/columnHeaders/ColumnHeaderMenuIcon.tsx
```diff
@@ -13,10 +14,24 @@ export interface ColumnHeaderMenuIconProps {
   iconButtonRef: React.RefObject<HTMLButtonElement>;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 17: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -31,13 +32,38 @@ interface GridColumnHeaderItemProps {
   tabIndex: 0 | -1;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 35: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSeparator.tsx
```diff
@@ -1,22 +1,39 @@
...
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
+import { composeClasses } from '../../utils/material-ui-utils';
 
 export interface GridColumnHeaderSeparatorProps extends React.HTMLAttributes<HTMLDivElement> {
   resizable: boolean;
   resizing: boolean;
   height: number;
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 13: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaders.tsx
```diff
@@ -11,10 +11,22 @@ import { GridColumnHeadersItemCollection } from './GridColumnHeadersItemCollecti
...
 import { gridColumnReorderDragColSelector } from '../../hooks/features/columnReorder/columnReorderSelector';
 import { gridContainerSizesSelector } from '../../hooks/root/gridContainerSizesSelector';
-import { gridClasses } from '../../gridClasses';
+import { getDataGridUtilityClass } from '../../gridClasses';
+import { composeClasses } from '../../utils/material-ui-utils';
+import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
 
 export const gridScrollbarStateSelector = (state: GridState) => state.scrollBar;
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 20: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/containers/GridRoot.tsx
```diff
@@ -2,40 +2,51 @@ import * as React from 'react';
...
 import { useGridState } from '../../hooks/features/core/useGridState';
 import { useGridApiContext } from '../../hooks/root/useGridApiContext';
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
-import { gridClasses } from '../../gridClasses';
+import { getDataGridUtilityClass } from '../../gridClasses';
+import { composeClasses } from '../../utils/material-ui-utils';
 
 export type GridRootProps = React.HTMLAttributes<HTMLDivElement>;
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 17: Can we type the `ownerState` ?

packages/grid/_modules_/grid/components/containers/GridWindow.tsx
```diff
@@ -7,13 +7,25 @@ import {
...
-import { gridClasses } from '../../gridClasses';
+import { getDataGridUtilityClass } from '../../gridClasses';
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
+import { composeClasses } from '../../utils/material-ui-utils';
 
 export interface GridWindowProps extends React.HTMLAttributes<HTMLDivElement> {
   size: { width: number; height: number };
 }
 
+const useUtilityClasses = (ownerState) => {
```
flaviendelangle, line 18: Can we type the `ownerState` ?

packages/grid/data-grid/src/useDataGridProps.ts
```diff
@@ -10,6 +10,16 @@ export const useDataGridProps = (inProps: DataGridProps): GridComponentProps =>
 
   const themedProps = useThemeProps({ props: inProps, name: 'MuiDataGrid' });
 
+  const classes = React.useMemo(() => {
+    if (!inProps.classes) {
+      return {};
+    }
+    return Object.entries(inProps.classes).reduce((acc, [key, value]) => {
```
flaviendelangle, line 17: Can we comment why we need to replace `__` by `--` ?

packages/grid/_modules_/grid/components/GridRowCount.tsx
```diff
@@ -25,7 +27,11 @@ export const GridRowCount = React.forwardRef<HTMLDivElement, GridRowCountProps>(
         : rowCount.toLocaleString();
 
     return (
-      <div ref={ref} className={clsx(gridClasses.rowCount, className)} {...other}>
+      <div
+        ref={ref}
+        className={clsx(gridClasses.rowCount, rootProps.classes?.rowCount, className)}
```
flaviendelangle, line 32: Is it impossible to just process the `class` prop and have `rootProps.classes?.XXX` always contains `gridClasses.XXX` and `inputProps.classes?.XXX` ?

To avoid all these merge in the hooks / components.

flaviendelangle (changes requested):

packages/grid/_modules_/grid/components/GridRowCount.tsx
m4theushw: It's possible, but this same pattern is used in the core. They chose to be verbose.

packages/grid/data-grid/src/useDataGridProps.ts
m4theushw: It's ugly to use two dashes in an interface member. See https://github.com/mui-org/material-ui-x/issues/420#issuecomment-896359219. Now the replace is being done in `generateUtilityClass` and this pre-process doesn't exist anymore.

packages/grid/_modules_/grid/components/GridRow.tsx
m4theushw: All `ownerState` are properly typed now.

packages/grid/_modules_/grid/components/GridRowCount.tsx
oliviertassinari: In the core, I believe we do as Flavien suggests https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Alert/Alert.js#L167, composeClasses merge the global classes with the classes prop.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -151,9 +151,6 @@ export const useStyles = makeStyles(
           whiteSpace: 'nowrap',
           fontWeight: theme.typography.fontWeightMedium,
         },
-        [`& .${gridClasses['columnHeader--moving']}`]: {
-          backgroundColor: theme.palette.action.hover,
```
oliviertassinari, line 155: Where did this style go?

packages/grid/_modules_/grid/components/GridRowCount.tsx
m4theushw: From what I understood, https://github.com/mui-org/material-ui-x/pull/2458#discussion_r698295681 is proposing to merge in a single place (e.g. useDataGridProps) the `classes` that the user is providing with the default grid's classes. By doing that, `composeClasses` will include each class twice:

`<DataGrid classes={{ cell: 'foobar' }} />`

`rootProps.classes.cell` -> `"MuiDataGrid-cell foobar"`

`composeClasses(['cell'], ..., rootProps.classes)` -> `"MuiDataGrid-cell MuiDataGrid-cell foobar"`

The alternative is to implement the `useUtilityClasses` pattern in all the places.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
m4theushw: Reverted. My bad.

packages/grid/_modules_/grid/components/GridRowCount.tsx
flaviendelangle: Yes my proposal is what @m4theushw is describing.
If it can be done without downsides, I think it's cleaner.
But if it duplicates data, it may not be a good idea.

flaviendelangle:
Seems good to me

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -64,21 +90,16 @@ export function GridRow(props: GridRowProps) {
   };
 
   const rowClassName =
-    isFunction(rootProps.getRowClassName) &&
+    typeof rootProps.getRowClassName === 'function' &&
```
oliviertassinari, line 93: 👍

packages/grid/_modules_/grid/gridClasses.ts
```diff
@@ -1,62 +1,304 @@
...
+  'menuIconButton',
   'menuOpen',
-  'columnHeaderWrapper',
-  'editInputCell',
-  'editBooleanCell',
-  'booleanCell',
-  'checkboxInput',
+  'overlay',
+  'renderingZone',
+  'root',
```
oliviertassinari, line 288: I wonder if we should keep the classes in a DOM order rather than an alphabetical one. For instance, *root* is the first one people will see in the DOM. In the core, we use the DOM order, but I would understand that it might not work well when they are many classes.

packages/grid/_modules_/grid/utils/material-ui-utils.ts
```diff
@@ -105,7 +105,7 @@ const globalPseudoClassesMapping: Record<string, string> = {
 
 export function generateUtilityClass(componentName: string, slot: string): string {
   const globalPseudoClass = globalPseudoClassesMapping[slot];
-  return globalPseudoClass || `${componentName}-${slot}`;
+  return globalPseudoClass || `${componentName}-${slot}`.replace('__', '--');
```
oliviertassinari, line 108: I got a bit confused about the usage of `__` for the JS classes API, and `--` for the CSS class name API. Is the long-term plan to unify around `__`? I recall seeing an issue but I can't find it back. Maybe to open a new one.

oliviertassinari (commented):
Looks like we are close, I have left a couple of questions

packages/grid/_modules_/grid/utils/material-ui-utils.ts
m4theushw: I reverted this part because I didn't receive many positive feedbacks about it in https://github.com/mui-org/material-ui-x/issues/420#issuecomment-896359219. We could re-evaluate if the BEM syntax is something we want to use in the future. I was trying to avoid having to use this syntax: https://github.com/mui-org/material-ui-x/pull/2458/files#diff-daa7ce589b380e02bf73ba2d9b08377cc7571039f62c5e057e027f3f0393779aR86

packages/grid/_modules_/grid/gridClasses.ts
m4theushw: DOM order would make more sense. I'll reorder them in #2465. For now, the alphabetical order is fine.

PR #2462, opened 2021-08-27
[DataGrid] Reduce specificity of GridToolbarContainer styles

### LINKED ISSUE #2177: [DataGrid] Flatten CSS specificity of GridToolbarContainer

## Current Behavior 😯

The styles of `GridToolbarContainer` have a CSS specificity of 2. This means that when developers try to customize the style, they will get surprises. For instance, this demo: https://material-ui.com/components/data-grid/export/#csv-export.

I have experienced the pain firsthand in #2149 trying to customize the demo. It broke my flow. It took me at least 5 minutes to open my dev tool to figure out why my flex alignment wasn't working. 

## Expected Behavior 🤔

Flat specificity. I think that we can simply have `GridToolbarContainer` own its styles.

## Steps to Reproduce 🕹

I assume that the bug description is clear enough, but I can provide a reproduction if asked. The high order bit is:

```jsx
// .my-class doesn't win
<GridToolbarContainer className="my-class" />
```

## Context

I couldn't find more cases like this.

## Your Environment 🌎

v4.0.0-alpha.33

michaldudak:
This reduces the specificity of the styles of GridToolbarContainer by one.

I added the `nameof` utility to be able to use class keys in type-safe way.

Fixes #2177


m4theushw (commented):
I think that this component should use `makeStyles`, like others. 

https://github.com/mui-org/material-ui-x/blob/4eb92a6dcae63f6fbb0bb9d195a8adde805a356f/packages/grid/_modules_/grid/components/panel/GridPanelHeader.tsx#L14

m4theushw:
I reworked this PR to solve the problem like other components do, that means adding `makeStyles` to them.

DanailH (approved):

PR #2463, opened 2021-08-27
[DataGrid] Pressing `Escape` should close `GridColumnHeaderMenu` without errors

DanailH:
Found a regression that is a quick win.

Currently pressing the `Escape` key when the `GridColumnHeaderMenu` is opened results in an error in the console and the menu doesn't close. This PR fixes the broken behavior.

### DIFF
--- a/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
+++ b/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
@@ -39,7 +39,7 @@ const useStyles = makeStyles(
 export interface GridMenuProps extends Omit<PopperProps, 'onKeyDown'> {
   open: boolean;
   target: React.ReactNode;
-  onClickAway: (event: React.SyntheticEvent<Document, MouseEvent>) => void;
+  onClickAway: (event: React.MouseEvent<Document> | React.TouchEvent) => void;
   position?: MenuPosition;
 }
 
--- a/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
+++ b/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
@@ -39,7 +39,7 @@ const useStyles = makeStyles(
 export interface GridMenuProps extends Omit<PopperProps, 'onKeyDown'> {
   open: boolean;
   target: React.ReactNode;
-  onClickAway: (event?: React.MouseEvent<Document, MouseEvent>) => void;
+  onClickAway: (event: React.SyntheticEvent<Document, MouseEvent>) => void;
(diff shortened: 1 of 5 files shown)


packages/grid/_modules_/grid/components/menu/columnMenu/GridColumnHeaderMenu.tsx
```diff
@@ -25,7 +25,7 @@ export function GridColumnHeaderMenu({
   const currentColumn = apiRef.current.getColumn(field);
 
   const hideMenu = React.useCallback(
-    (event) => {
+    (event: React.SyntheticEvent<any, any>) => {
```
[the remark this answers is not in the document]
DanailH: This needs to be `React.SyntheticEvent<any>` because that function `hideMenu` is passed to the `GridMenu` and that on the other hand passes it to the `<ClickAwayListener onClickAway={hideMenu}>`. The `onClickAway` expects the event to be `React.SyntheticEvent<Document, MouseEvent>`

packages/grid/_modules_/grid/components/menu/GridMenu.tsx
```diff
@@ -39,7 +39,7 @@ const useStyles = makeStyles(
 export interface GridMenuProps extends Omit<PopperProps, 'onKeyDown'> {
   open: boolean;
   target: React.ReactNode;
-  onClickAway: (event?: React.MouseEvent<Document, MouseEvent>) => void;
+  onClickAway: (event: React.SyntheticEvent<Document, MouseEvent>) => void;
```
oliviertassinari, line 42: This is not a synthetic event AFAIK. In v5, I see

```tsx
onClickAway: (event: MouseEvent | TouchEvent) => void;
```

in the source of ClickAwayListener. So I guess?
```suggestion
  onClickAway: (event: MouseEvent | TouchEvent) => void;
```

DanailH: Yes but in v4 (which we are still using) is `onClickAway: (event: React.MouseEvent<Document>) => void;`

https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/ClickAwayListener/ClickAwayListener.d.ts#L20

For now I can type it to satisfy the linter and later when we fully migrate to v5 fix the typing (it will inevitably fail on the typescript step so there is no way we will miss it).

oliviertassinari: Ok, sounds fair

flaviendelangle (approved):

PR #2464, opened 2021-08-27
[DataGrid] Don't consider unselectable rows when selectionModel is used

m4theushw:
Fixes https://github.com/mui-org/material-ui-x/pull/2457#discussion_r697617709

If `selectionModel` is used, its value goes directly to the internal state. By doing that, the state may contain unselectable rows. This PR fixes a situation where the "Select All" checkbox would appear in an indeterminate state, although there're no selectable rows selected.

Before: https://codesandbox.io/s/material-demo-forked-9qx02?file=/demo.js
After: https://codesandbox.io/s/material-demo-forked-lrb2u?file=/demo.js


flaviendelangle:
With this approach, sometimes the `useGridSelection` removes the unselectable items from the internal state and sometime it does not.
Wouldn't it be better to always filter before injecting in the state, or to never filter and always let the components filter ?

And if it is in the component, we probably want to use a `useMemo` because with large selection amount if causes a lot of computation on every render.

flaviendelangle (approved):

PR #2500, opened 2021-08-29
[core] Do not run test_browsers on Renovate PRs

flaviendelangle:
The Renovate PRs are coming in large groups every week.
So the `test_browser` step is failing 100% of the time and rerun it PR by PR sequentially to avoid timeout would take hours.
I think we can drop it, as we are merging them anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,6 +218,10 @@ workflows:
       - test_browser:
           requires:
             - checkout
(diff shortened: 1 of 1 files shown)


oliviertassinari:
I would propose we put this PR on hold for the resolution of https://github.com/mui-org/material-ui/issues/27991.

PR #2501, opened 2021-08-30
[core] Drop context folder

flaviendelangle:
@mui-org/x the `OutsideColumnsPanel` story is not working in `next` because it tries to use the preference panel outside of the Grid, but the panel uses `useGridRootProps` which is not available.

To be able to access `apiRef`, it instanciates `GridApiContext.Provider` manually, which I would like to avoid.
But anyway, for `rootProps` I don't see how we could correctly handle it.
It we just export the context, users would not have the (`DataGridProps` | `DataGridProProps`) => `GridComponentProps` processing so the `rootProps` would not be complete.

Do we really want to allow people to instanciate our components outside of the Grid context.
It seems to be a lot of trouble and very easy to break in the future.


oliviertassinari:
I guess we have moved the discussion to https://github.com/mui-org/material-ui-x/issues/2522.

m4theushw:
Could we close this one or do you want to push it forward?

flaviendelangle:
@m4theushw I'll close it

PR #2468, opened 2021-08-28
[core] Improve the feedback loop from developers

oliviertassinari:
This pull request started out of my frustration of having developers send emails to get support on the Pro plan. This feels inefficient because: 

1. We split the place where we collect user's pain into two different locations. It makes it harder to consolidate the problems, note a great product feedback loop.
Instead, by moving the questions to GitHub, we can for instance: link duplicates, link PRs that are fixing the root causes. A note on StackOverflow. This is a pain we have with it too, the main difference is that the community is answering these support questions. It's a tradeoff we have made to leverage the gamification they have in place to incentive contributions. With the Pro plan, it's us that have to answer, I doubt that the community will have the incentive to contribute much.
2. On the MIT side, each member asking a question and us answering is adding value to the whole user base. We should be able to reproduce this network effect with our paid customers. An instant self-serving experience is often better than waiting 2 days to get an answer.

But while I was working on it, I realized that GitHub had a neat new feature in beta: [Issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms). I went on leveraging it.

You can preview the exact impacts of these changes at https://github.com/mui-org/test-issue-template/issues/new/choose. For instance:

Screenshot 2021-08-29 at 00 26 53

The aim is to backport these changes into @mui-org/core if the PR is merged.

PR #2503, opened 2021-08-30
[core] Polish issue template

oliviertassinari:
A follow-up on #2468, based on https://github.com/mui-org/material-ui/pull/28038

### DIFF
--- a/.github/ISSUE_TEMPLATE/1.bug.yml
+++ b/.github/ISSUE_TEMPLATE/1.bug.yml
@@ -56,14 +56,14 @@ body:
   - type: textarea
     attributes:
       label: Your environment 🌎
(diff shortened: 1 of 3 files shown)


oliviertassinari:
I also wonder if we shouldn't make the Order ID field more explicit around its optionality

m4theushw (approved):
I pushed 8f57500 with a fix for a small typo I noticed outside this PR. It looks much better now with the form inputs. 👌

oliviertassinari:
Nice thanks for the fix, other issues were spotted on the same PR in the core repo, I will keep the two in sync and merge once both approved.

oliviertassinari:
Nice thanks for the fix, other issues were spotted on the same PR in the core repo, I will keep the two in sync and merge once both approved.

Issue #479, opened 2020-10-20
[core] Enforce noImplicitAny


oliviertassinari:
We should aim to enforce `noImplicitAny` in the repository. This is a best practice we encourage in [the documentation](https://material-ui.com/guides/typescript/). I have tried to enforce it in #448 but that was too much of an effort. I had to give up, I don't have the bandwidth for it.

simplenotezy:
What about projects partially migrating to noImplicitAny? Disabling this, would cause quite a lot of work to migrate to typescript: https://mui.com/guides/typescript/#minimum-configuration

flaviendelangle:
Not sure to understand your point here.

simplenotezy:
We're migrating a site to Typescript. I believe if we enable `noImplicitAny` we're forced to migrate all code to Typescript if we wish to avoid errors, or am I wrong? The documentation at https://mui.com/guides/typescript/#minimum-configuration tells us we need to enable it.

flaviendelangle:
OK I understand
This issue talks about our codebase which do not have `noImplicitAny` yet.

For your codebase, I don't think that `noImplicitAny` is mandatory. The doc example could be clearer.
I will check with the team responsible to update it if they confirm it works without it.

mnajdova:
@simplenotezy these options are recommended, not required. The wording that was chosen there is a bit misleading. I will update the documentation.

PR #2505, opened 2021-08-30
[core] Prepare x-grid-data-generator for noImplicitAny

flaviendelangle:


oliviertassinari:
Do you know how big of an effort it will be to get #479 in? I guess the longer it wait, the worse it will be. On the other hand, I wonder if we will find bugs thanks to it.

flaviendelangle:
I think it should not take too long, and yes I agree that waiting is making it worse.

> On the other hand, I wonder if we will find bugs thanks to it.

Perhaps
It's also improving the readability of functions

PR #2515, opened 2021-08-31
[DataGrid] Drop v4 support

m4theushw:
### Breaking changes

- The module augmentation is not enabled by default. This change was done to prevent conflicts with projects using `DataGrid` and `DataGridPro` together. In order to still be able to do overrides at the theme level, add the following imports to your project:

  ```diff
  +import type {} from '@mui/x-data-grid/themeAugmentation';
  ```
  
  ```diff
  +import type {} from '@mui/x-data-grid-pro/themeAugmentation';
  ```

---

One of #1902, Preview: https://deploy-preview-2515--material-ui-x.netlify.app/components/data-grid/

The previous module augmentation was using the `MuiDataGridPro` key for overrides:

https://github.com/mui-org/material-ui-x/blob/afb858cb986c7b86494be7261353fde9fe537b90/packages/grid/x-grid/src/themeAugmentation/props.ts#L4

However, this is not right because even in the Pro version the overrides are retrieved from the `MuiDataGrid` key, so I updated the definitions to use the same key. As consequence, we can't export these definitions in the index because if a project uses both versions (e.g.  `@mui/x-data-grid-generator`) they will conflict and not compile. That's the reason for the breaking change.

https://github.com/mui-org/material-ui-x/blob/afb858cb986c7b86494be7261353fde9fe537b90/packages/grid/x-grid/src/useDataGridProProps.ts#L11

I didn't update the build config to generate the `themeAugmentation` folder. I plan to do it in a follow-up.


packages/grid/data-grid/src/index.ts
```diff
@@ -2,6 +2,3 @@ export * from '../../_modules_';
 export * from './DataGrid';
 export * from './DataGridProps';
 export * from './useDataGridComponent';
-// TODO v5, remove this should be handled userland
-// https://github.com/mui-org/material-ui/pull/21279#discussion_r435748054
-export * from './themeAugmentation';
```
oliviertassinari, line 7: AFAIK, we can't remove it yet. Develpers need to be able to import deep `@mui/x-data-grid/themeAugmentation `

test/regressions/index.js
```diff
@@ -21,6 +21,7 @@ const blacklist = [
   'docs-components-data-grid-filtering/ColumnTypeFilteringGrid.png', // Needs interaction
   'docs-components-data-grid-filtering/CustomRatingOperator.png', // Needs interaction
   'docs-components-data-grid-filtering/ExtendNumericOperator.png', // Needs interaction
+  'docs-components-data-grid-components/CustomFooter.png', // Needs @material-ui/lab v4
```
oliviertassinari, line 24: Could you detail the issue here?

oliviertassinari (commented):
Looks great

packages/grid/x-grid/src/themeAugmentation/props.ts
```diff
@@ -1,9 +1,21 @@
...
+    styleOverrides?: ComponentsOverrides['MuiDataGrid'];
+  };
 }
 
 declare module '@material-ui/core/styles' {
-  interface ComponentsPropsList extends DataGridProComponentProps {}
+  interface ComponentsPropsList extends DataGridProComponentsPropsList {}
+}
+
+declare module '@material-ui/core/styles/components' {
```
oliviertassinari, line 19: It's a private path:

```suggestion
declare module '@material-ui/core/styles' {
```

packages/grid/data-grid/src/themeAugmentation/props.ts
```diff
@@ -1,9 +1,21 @@
...
+    styleOverrides?: ComponentsOverrides['MuiDataGrid'];
+  };
+}
+
 declare module '@material-ui/core/styles' {
-  interface ComponentsPropsList extends DataGridComponentProps {}
+  interface ComponentsPropsList extends DataGridComponentsPropsList {}
+}
+
+declare module '@material-ui/core/styles/components' {
```
oliviertassinari, line 19: Private 

```suggestion
declare module '@material-ui/core/styles' {
```

flaviendelangle:
image

It's sad to loose the full label of the header of the column `Desk` (same on other headers)

Otherwise, the Argos diffs seems logical

flaviendelangle:
https://github.com/mui-org/material-ui-x/issues/2527

@mui-org/x  should we switch to the release candidate of the core directly on this PR to avoid having this problem ?
I suppose users already on v5 will quickly upgrade to the release candidate, and thus we should support the new package names rather than the old ones.

m4theushw:
@flaviendelangle No, this PR is only to remove the v4 dependency. To support the new package name we have to rename all files, so it's better to do in another one.

m4theushw:
@flaviendelangle It's sad to loose the full label of the header of the column Desk (same on other headers)

The IconButton in v5 is 28x28 while it's 24x24 in v4. We can increase the column width in the x-data-grid-generator.

test/regressions/index.js
m4theushw: I upgraded all @material-ui/xxx dependencies in the root's package.json to v5, so this regression test will now run with @material-ui/lab v5. However, the Rating component in v5 is not in the lab anymore, then it throws an "The Rating component was moved from the lab to the core" error in the console. Since no test can output to the console it fails.

https://github.com/mui-org/material-ui-x/blob/e214b3a43f639980fde82ff9d62eb40612aa335f/test/regressions/index.test.js#L107

packages/grid/x-grid/src/themeAugmentation/props.ts
m4theushw: The `Components` interface is not exported. We will be able to remove the private path when @mui/material 5.0.0-rc.0 became the minimum version. The RC includes https://github.com/mui-org/material-ui/pull/28078.

packages/grid/data-grid/src/themeAugmentation/props.ts
m4theushw: https://github.com/mui-org/material-ui-x/pull/2515#discussion_r702179651

packages/grid/x-grid-data-generator/src/renderer/renderEditCountry.tsx
```diff
@@ -66,7 +66,7 @@ function EditCountry(props: GridRenderEditCellParams) {
       open
       classes={classes}
       disableClearable
-      renderOption={(option) => (
+      renderOption={(liProps, option) => (
```
oliviertassinari, line 69: `liProps` needs to be applied on a DOM node, to replace the React fragment.

test/regressions/index.js
oliviertassinari: OK, so maybe we should do
```suggestion
  // TODO import the Rating from @mui/material, not the lab.
  'docs-components-data-grid-components/CustomFooter.png',
```

oliviertassinari (changes requested):
Looks good

flaviendelangle:
> > It's sad to loose the full label of the header of the column Desk (same on other headers)
> 
> @flaviendelangle The IconButton in v5 is 28x28 while it's 24x24 in v4. We can increase the column width in the x-data-grid-generator.

OK
I think it would be a good idea indeed.

PR #2519, opened 2021-09-01
[DataGrid] Print export

### LINKED ISSUE #200: [DataGrid] Implement Print export
### Benchmark

- https://www.ag-grid.com/javascript-grid-for-print/
- https://ej2.syncfusion.com/react/demos/#/material/grid/print
- https://www.telerik.com/kendo-react-ui/components/grid/pdf-export/
- https://demos.telerik.com/kendo-ui/grid/pdf-export
- http://tabulator.info/docs/4.7/print & pdf

DanailH:
Fixes #200 

This is the initial version of the Print Export feature. I'll keep the PR as a Draft for now as tests and docs are missing from the PR and I would like to discuss the approach and features before adding the tests and docs and converting it to normal PR.

**Preview**: TO BE ADDED...

### Explonation:

In the `Export` drop-down menu in the `GridToolbar` is now an additional button called `Print`. The remaining API is part of the `useGridPrintExport` hook. The setup is identical to the CSV Export feature.

Options available for the Print Export:

```
export interface GridPrintExportOptions {
  /**
   * The string used as the print window title.
   * @default 'Print Window'
   */
  fileName?: string;
  /**
   * The columns to be printed.
   * This should only be used if you want to restrict the columns exports.
   */
  fields?: string[];
  /**
   * If `true`, the hidden columns will also be printed.
   * @default false
   */
  allColumns?: boolean;
  /**
   * If `true`, the toolbar is removed for when printing.
   * @default false
   */
  hideToolbar?: boolean;
  /**
   * If `true`, the footer is removed for when printing.
   * @default false
   */
  hideFooter?: boolean;
  /**
   * If `false`, all <style> and <link type="stylesheet" /> tags from the <head> will not be copied
   * to the print window.
   * @default true
   */
  copyStyles?: boolean;
  /**
   * One or more classes passed to the print window.
   */
  bodyClass?: string;
  /**
   * Provide Print specific styles to the print window.
   */
  pageStyle?: string | Function;
}
```

These options can also be passed to the `GridToolbarExport` component (same as the CSV export).

---

ToDo:
- [ ] Add tests
- [ ] Add docs


m4theushw (commented):
I had a quick look and it's working fantastically. 👌

1. Did you think about exposing the function below as a `disableVirtualization` API call? It could be used instead of updating the state.

https://github.com/mui-org/material-ui-x/blob/4eb92a6dcae63f6fbb0bb9d195a8adde805a356f/packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts#L23

2. We could add a few CSS rules with `@media print` to make the grid print-safe. The left and right borders could be removed, the same for the padding.

3. In the full-featured demo, when I close the print window it shows columns that were hidden before, like the ID.

DanailH:
> I had a quick look and it's working fantastically. 👌
> 
> 1. Did you think about exposing the function below as a `disableVirtualization` API call? It could be used instead of updating the state.
> 
> https://github.com/mui-org/material-ui-x/blob/4eb92a6dcae63f6fbb0bb9d195a8adde805a356f/packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts#L23
> 
> 1. We could add a few CSS rules with `@media print` to make the grid print-safe. The left and right borders could be removed, the same for the padding.
> 2. In the full-featured demo, when I close the print window it shows columns that were hidden before, like the ID.

@m4theushw thanks for the review. I'll wait to see what the others have to say and fix the issues.

For the `disableVirtualization` - when worked on the implementation that wasn't available. I can have a look and see about exposing it.

I was thinking of adding the CSS rules. We can add a few but that's why I added the option for people to provide their own riles directly.

I'll check about the hidden columns, that shouldn't happen

flaviendelangle:
Concerning the rename of the CSV option interface

We will need to make a PR in the core (adding also the print at the same time I suppose) and then delete this file (and the `js` part).

https://github.com/mui-org/material-ui-x/blob/2dbaca1a24fb554408e6b8ec225bf612d8e445a9/docs/pages/api-docs/data-grid/grid-export-csv-options.md

oliviertassinari (commented):
It looks like a great start. We are going further than what I was expecting for the first iteration, which is cool. I would propose we offer two different levels of solution:

1. You have built an API to turn the rendering of the grid in a print-ready mode. I would propose that we double down on it:
  - There is a rendering bug on http://0.0.0.0:3001/components/data-grid/export/#csv-export, only one row is visible in the print version
  - There is a rendering bug on http://0.0.0.0:3001/components/data-grid/export/#csv-export, the restored state is not identical to the initial one. Related to https://github.com/mui-org/material-ui-x/pull/2519#discussion_r701137959
  - We could expose this as a public API. It would make it easy for developers that are used to the AG Grid API. It would allow more customizations for the developers, say they want to print chart at the same time or do some other transformations before calling `print` (Pro only?). https://github.com/mui-org/material-ui-x/pull/2519#discussion_r701140409
  - There is likely something to do about > 1,000 rows export. It prints on 20 pages, it feels that beyond, it's pointless. Same feedback for the `autoHeight` prop.
2. The new export menu option is cool. It's what I mean by "going further than what I was expecting". I didn't notice something specific about it, outside of a few changes, some breaking, on the existing CSV feature that maybe should be done standalone (not sure, I assume we fixed stuff not directly related to the print).

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const logger = useLogger('useGridPrintExport');
+  const [, setGridState, forceUpdate] = useGridState(apiRef);
+  const visibleSortedRows = useGridSelector(apiRef, visibleSortedGridRowsSelector);
+  const columns = useGridSelector(apiRef, allGridColumnsSelector);
+  const doc = React.useRef<Document | null>(null);
+  const previousGridState = React.useRef<any>(null);
+  const previousHiddenColumns = React.useRef<string[]>([]);
+
+  React.useEffect(() => {
+    doc.current = ownerDocument(apiRef!.current.rootElementRef!.current as HTMLElement);
```
flaviendelangle, line 35: The non-null operator is not necessary here

```suggestion
    doc.current = ownerDocument(apiRef.current.rootElementRef!.current!);
```

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+  }, [apiRef]);
+
+  const prepGridForPrint = React.useCallback(
+    (fields?: string[], allColumns?: boolean): void => {
+      // Show only wanted columns
+      if (fields && fields.length) {
+        apiRef!.current.updateColumns(
```
flaviendelangle, line 42: Same

```suggestion
        apiRef.current.updateColumns(
```

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+      // `onload` callback. This ensures that it is only called once.
+      printWindow.onload = null;
+
+      const printDoc = printWindow.contentDocument || printWindow.contentWindow?.document;
+
+      if (!printDoc) {
+        return;
+      }
+
+      const gridRootElement = apiRef!.current.rootElementRef!.current;
```
flaviendelangle, line 114: Same

```suggestion
      const gridRootElement = apiRef.current.rootElementRef!.current;
```

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+    [apiRef, doc, visibleSortedRows.size, props.rowHeight],
+  );
+
+  const handlePrintWindowAfterPrint = React.useCallback(
+    (printWindow: HTMLIFrameElement): void => {
+      // Remove the print iframe
+      doc.current!.body.removeChild(printWindow);
+
+      // Revert columns to their original state
+      apiRef!.current.updateColumns(
```
flaviendelangle, line 196: Same

```suggestion
      apiRef.current.updateColumns(
```

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+      previousHiddenColumns.current = [];
+    },
+    [previousGridState, previousHiddenColumns, columns, apiRef, setGridState, forceUpdate],
+  );
+
+  const exportDataAsPrint = React.useCallback(
+    (options?: GridPrintExportOptions): void => {
+      logger.debug(`Export data as Print`);
+
+      if (!apiRef!.current.rootElementRef!.current) {
```
flaviendelangle, line 225: Same

```suggestion
      if (!apiRef.current.rootElementRef!.current) {
```

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+type PrintWindowOnLoad = (
+  printWindow: HTMLIFrameElement,
+  options?: Pick<
+    GridPrintExportOptions,
+    'copyStyles' | 'bodyClass' | 'pageStyle' | 'hideToolbar' | 'hideFooter'
+  >,
+) => void;
+
+export const useGridPrintExport = (
```
flaviendelangle, line 22: Can you please list the feature hooks that this hook requires ?
So that we can correctly assert the order in `useDataGridComponent` and `useDataGridProComponent`

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+      if (normalizeOptions.bodyClass) {
+        printDoc.body.classList.add(...normalizeOptions.bodyClass.split(' '));
+      }
+
+      if (normalizeOptions.copyStyles) {
+        const headStyleElements = doc.current!.querySelectorAll("style, link[rel='stylesheet']");
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < headStyleElements.length; i++) {
```
flaviendelangle, line 147:
```suggestion
        for (let i = 0; i < headStyleElements.length; i+=1) {
```

Otherwise we should remove the linting rule if we start disabling it.
I'm not a fan of this `no plus plus` rule but as you want :+1:

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+              height: visibleSortedRows.size * props.rowHeight!,
+            },
+            renderingZonePageSize: visibleSortedRows.size,
+          },
+          viewportSizes: {
+            ...state.viewportSizes,
+            height: visibleSortedRows.size * props.rowHeight!,
+          },
+        };
+      });
```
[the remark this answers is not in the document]
DanailH: Ok, I took a bit more time to check this more in-depth. The problem is that this logic doesn't really do enable `autoHeight` neither it disables the virtualization. What it does is force the grid to render all rows (have all the HTML for the rows be in the dom but still keep the scroll so that the content on the page doesn't jump). 

For example @m4theushw suggestion to reuse his logic from the `useGridNoVirtualization` won't work because as far as can see there are other places where that prop is directly red and used (same as the `autoHeight` , which was my original idea). I'll see if moving that chunk into `useGridNoVirtualization` will make sense and expose it as an API from there.

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const handlePrintWindowAfterPrint = React.useCallback(
+    (printWindow: HTMLIFrameElement): void => {
+      // Remove the print iframe
+      doc.current!.body.removeChild(printWindow);
+
+      // Revert columns to their original state
+      apiRef!.current.updateColumns(
+        columns.map((column) => {
+          if (previousHiddenColumns.current.includes(column.field)) {
```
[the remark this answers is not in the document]
DanailH: Unfortunately, it won't work unless I deep clone it.

flaviendelangle:
@DanailH https://github.com/mui-org/material-ui-x/blob/next/docs/src/pages/components/data-grid/getting-started/getting-started.md

We can update the `Print` line on this page

DanailH:
Ok, I finally made the test work (locally, will see if it works in here). I would really appreciate it if you guys can take a look at the `playwrite` test (it's ugly but I've added comments on why it's like this). There is probably room for improvement but I would prefer to move forward and merge this.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts
```diff
@@ -69,6 +77,25 @@ export const useGridNoVirtualization = (
...
   ]);
 
+  const disableVirtualization = React.useCallback((): void => {
+    setGridState((state) => ({
+      ...state,
+      containerSizes: {
+        ...state.containerSizes!,
+        renderingZone: {
+          ...state.containerSizes!.renderingZone,
+          height: visibleRowCount * props.rowHeight!,
```
m4theushw, line 87: Same thing as before. Using `props.rowHeight` won't consider different densities. The selector has to be used.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts
```diff
@@ -69,6 +77,25 @@ export const useGridNoVirtualization = (
...
+        },
+        renderingZonePageSize: visibleRowCount,
+      },
+      viewportSizes: {
+        ...state.viewportSizes,
```
m4theushw, line 92: The width must also be updated to avoid clipping the content during print. You can get the total width occupied by the columns using `gridColumnsMetaSelector`.

https://github.com/mui-org/material-ui-x/blob/50aae6943c4f865d4658e56adda815dca1386d13/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualization.ts#L212

image

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -35,7 +30,52 @@ export interface GridExportCsvOptions {
   allColumns?: boolean;
 }
 
+/**
+ * The options to apply on the Print export.
+ */
+export interface GridPrintExportOptions {
+  /**
+   * The string used as the print window title.
+   * @default 'Print Window'
```
m4theushw, line 39:
```suggestion
   * The value to be used as the print window title.
   * @default The title of the page.
```

test/regressions/index.test.js
```diff
@@ -95,13 +101,52 @@ async function main() {
...
+          let printGrid;
+          // Click the export button in the toolbar.
+          await page.$eval(`button[aria-label="Export"]`, (exportButton) => {
+            exportButton.click();
+          });
+          // Click the print export option from the export menu in the toolbar.
+          await page.$eval(`li[role="menuitem"]:last-child`, (printButton) => {
+            printButton.click();
+          });
+          // Because by default the iframe doesn't have a src the `onload` won't be fired in Chromium
```
m4theushw, line 121: Outside the test environment, the print button is also not working in Chromium / Chrome. We need a better solution.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridNoVirtualization.ts
DanailH: Updated - forcing the `width` on the clone directly solves the issue

test/regressions/index.test.js
DanailH: Updated - should be fixed now - adding `src` to the `iframe` seems to fix the issue. An additional step was needed tho to remove the contents of the `body` inside the `iframe` to prevent duplicated content.

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,277 @@
...
+      logger.debug(`Export data as Print`);
+
+      if (!apiRef.current.rootElementRef!.current) {
+        throw new Error('No Grid Root element available');
+      }
+
+      previousGridState.current = gridState;
+
+      if (props.pagination) {
+        apiRef.current.setPageSize(visibleRowCount);
```
m4theushw, line 249: This won't work if the user is controlling the page size. Forcing the state also doesn't work, because it will require a `forceUpdate()` to propagate the changes, and when this runs, all effects will also run again and reset the value back to the prop value. I see some options:

1. Add a escape hatch to temporarily disable the state update constraint. https://github.com/mui-org/material-ui-x/blob/95fed89847c5f1489ddb63dafedcf0d907d46354/packages/grid/_modules_/grid/hooks/features/core/useGridControlState.ts#L40
2. Add a note to the docs mentioning that the print export doesn't work if `pageSize` or `page` is present.
3. Render the grid inside the iframe and disable the virtualization: 

```jsx
ReactDOM.render(<DataGrid {...props} disableVirtualization />, iframeEl)
```

There're going to be a dependency cycle with 3., which I believe we could solve by using `require()` instead of `import`.

m4theushw (commented):
If the dark mode is used it will print with the styles from the dark mode. Since we only have the dark styles to copy there's no simple solution.

image

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -0,0 +1,277 @@
...
+        );
+        resolve();
+      }),
+    [columns, apiRef],
+  );
+
+  const buildPrintWindow = React.useCallback((title?: string): HTMLIFrameElement => {
+    const iframeEl = document.createElement('iframe');
+
+    iframeEl.id = 'grid-print-window';
```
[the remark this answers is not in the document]
DanailH: The `id` is only used for the tests. If we are going with your approach, setting the `id` is not needed.

packages/grid/_modules_/grid/hooks/features/export/useGridPrintExport.tsx
DanailH: I would go with option 2. -> the reason is that 3. for me will be too expensive in terms of the time it will take to render.
Also, this feature (along with a couple of other ones) doesn't work with server-side pagination which is controlled. If a developer is controlling the pagination the print export has an API that they can call manually.

This looks to me like a corner case - we can see what the feedback will be and if it is problem then we can spend time fixing it.

Issue #1865, opened 2021-06-10
[XGrid] Move license content to a standalone page


oliviertassinari:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Move the EULA to a standalone page at https://material-ui.com/x/license/.

## Examples 🌈

https://docs.google.com/document/d/1nkd5JP5wCefo6UwpVi6PnbTaw8DxhstkCZlfHV5ZgfI/edit

## Motivation 🔦

So far, we have used Google Drive as it was making it easy to have private collaboration with @mbrookes on the content. However: 

- Using Google Drive to render the EULA feels cheap.
- Using Google Drive to render prevents a few organizations to display it. I have forwarded an internal discussion with one example (we saw more in the past)

We depend on the work of the rebranding to make enough progress before the X team can work on this, but it shouldn't too far off down the road.

For ongoing collaboration on the content, I propose that we keep a duplicate of the file internally, to sync from time to time.

PR #2525, opened 2021-09-02
[docs] Clarify confusion

oliviertassinari:
The license is for MUI X Pro, not MUI X.

### DIFF
--- a/packages/x-license/LICENSE.md
+++ b/packages/x-license/LICENSE.md
@@ -1,7 +1,7 @@
-# [Material-UI X](https://material-ui.com/x/)
+# [Material-UI X Pro](https://material-ui.com/x/)
 
 This is commercial software.
-To use it, you need to agree to the [**End User License Agreement for Material-UI X**](https://material-ui.com/x/license/).
+To use it, you need to agree to the [**End User License Agreement for Material-UI X Pro**](https://material-ui.com/x/license/).
 If you do not own a commercial license, this file shall be governed by the trial license terms.
 
 All available Material-UI commercial licenses may be obtained at https://material-ui.com/store/items/material-ui-pro/.
--- a/packages/x-license/LICENSE.md
+++ b/packages/x-license/LICENSE.md
@@ -1,4 +1,4 @@
-# [Material-UI X Pro](https://material-ui.com/x-advanced/)
+# [Material-UI X Pro](https://material-ui.com/x/)
 
 This is commercial software.
 To use it, you need to agree to the [**End User License Agreement for Material-UI X Pro**](https://material-ui.com/x/license/).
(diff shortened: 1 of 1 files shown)


oliviertassinari:
Off-topic, the review might feel weird:

Screenshot 2021-09-02 at 21 33 58

The pakcage is @mui/x-license-pro but the folder is x-license. I propose we rename it to mui*-x-license-pro*.

oliviertassinari:
@danilo-leal Regarding the URLs, what should I use? Are we going with http://mui.com/x/ or http://mui.com/x-advanced/? And what should be the URL of the license page #1865?

mbrookes (approved):

danilo-leal:
> @danilo-leal Regarding the URLs, what should I use? Are we going with http://mui.com/x/ or http://mui.com/x-advanced/? And what should be the URL of the license page #1865?

I believe that the best way to go, after [revisiting this discussion on the man website pages refinement PR](https://github.com/mui-org/material-ui/pull/28081#discussion_r701443239), is `mui.com/x`. We can always use the _"advanced"_ word as a main characteristic/attribute of the components, but as far as the name of the product goes, it seems we're better off naming it as MUI X. The benefit comes whenever we have to mention the paid version of the components (MUI X _[Pro or Premium]_), which is way clearer and less redundant than MUI X Advanced Pro (or Premium).

No issues with `/mui-x-license-pro/`, as proposed, for the license, though. Looks good to me!

oliviertassinari:
> No issues with /mui-x-license-pro/, as proposed, for the license, though. Looks good to me!

Haha, yeah, it's for the name of the folder on GitHub 😁

> mui.com/x

Ok, no objections, I was wondering about either it would be too short for an URL or not. I assume that we will only try to get this page rank for the MUI X query, so we should be good.

packages/x-license/LICENSE.md
```diff
@@ -1,7 +1,7 @@
-# [Material-UI X](https://material-ui.com/x/)
+# [Material-UI X Pro](https://material-ui.com/x-advanced/)
```
oliviertassinari, line 1:
```suggestion
# [Material-UI X Pro](https://material-ui.com/x/)
```

packages/x-license/LICENSE.md
```diff
@@ -1,7 +1,7 @@
-# [Material-UI X](https://material-ui.com/x/)
+# [Material-UI X Pro](https://material-ui.com/x/)
```
oliviertassinari, line 1: Actually, this might be even clearer

```suggestion
# [Material-UI X Pro](https://material-ui.com/pricing/)
```

danilo-leal: Yeah, maybe! `/x` will still exist though, for being the marketing page for the advanced components. But if we were meaning to have one page especially for them in the paid version, I guess linking right to the pricing one might do the trick yes!

PR #2530, opened 2021-09-03
[core] Publish GridEvents.rowsSet when the rows state is modified after props.rows is updated

flaviendelangle:
I'm starting to work on the tree data and I found it weird to listen to `props.rows` in a `useEffect` to trigger behavior that is based on `state.rows`.
We rely on the fact that when the `useEffect` of `useGridFilter` or `useGridSorting`, the update of the state of `useGridRows` has already been done.

Additionally, the description of the events was suggesting that it was trigger of rows update, not matter why it was updated

```
  /**
   * Fired when the rows are updated.
   * @ignore - do not document.
   */
  rowsSet = 'rowsSet',
```

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
+++ b/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
@@ -21,9 +21,7 @@ import { useGridApiEventHandler } from '../../root/useGridApiEventHandler';
 import { useGridApiMethod } from '../../root/useGridApiMethod';
 import { useGridLogger } from '../../utils/useGridLogger';
 import { allGridColumnsSelector } from '../columns/gridColumnsSelector';
-import { useGridSelector } from '../core/useGridSelector';
 import { useGridState } from '../core/useGridState';
-import { gridRowCountSelector } from '../rows/gridRowsSelector';
 import { sortedGridRowIdsSelector, sortedGridRowsSelector } from './gridSortingSelector';
 
 /**
@@ -45,7 +43,6 @@ export const useGridSorting = (
   const logger = useGridLogger(apiRef, 'useGridSorting');
 
   const [gridState, setGridState, forceUpdate] = useGridState(apiRef);
-  const rowCount = useGridSelector(apiRef, gridRowCountSelector);
 
   const upsertSortModel = React.useCallback(
     (field: string, sortItem?: GridSortItem): GridSortModel => {
@@ -297,13 +294,6 @@ export const useGridSorting = (
     applySorting,
   };
   useGridApiMethod(apiRef, sortApi, 'GridSortApi');
-
-  React.useEffect(() => {
-    if (rowCount > 0) {
-      logger.debug('row changed, applying sortModel');
-      apiRef.current.applySorting();
-    }
-  }, [rowCount, apiRef, logger]);
 
   React.useEffect(() => {
     apiRef.current.updateControlState<GridSortModel>({
--- a/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
+++ b/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
@@ -219,30 +219,6 @@ export const useGridSorting = (
     [upsertSortModel, setSortModel, createSortItem, props.disableMultipleColumnsSorting],
   );
 
-  const handleColumnHeaderClick = React.useCallback(
-    ({ colDef }: GridColumnHeaderParams, event: React.MouseEvent) => {
-      const allowMultipleSorting = event.shiftKey || event.metaKey || event.ctrlKey;
-      sortColumn(colDef, undefined, allowMultipleSorting);
-    },
-    [sortColumn],
-  );
-
-  const handleColumnHeaderKeyDown = React.useCallback(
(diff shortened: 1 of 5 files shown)


packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
```diff
@@ -299,11 +298,6 @@ export const useGridSorting = (
...
   useGridApiMethod(apiRef, sortApi, 'GridSortApi');
 
-  React.useEffect(() => {
-    // When the rows prop change, we re apply the sorting.
-    apiRef.current.applySorting();
-  }, [apiRef, props.rows]);
-
   React.useEffect(() => {
     if (rowCount > 0) {
       logger.debug('row changed, applying sortModel');
```
oliviertassinari, line 303: I wonder if we can get ride of this effect too

```suggestion
```

oliviertassinari (commented):
It looks like the right direction to go in. I recall trying to make the logic that applies the sorting and the filtering more canonical. If there is an event for when the rows change, shouldn't it matter regardless of why the rows changed?

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
flaviendelangle: Why were we reapplying the sort when the `rowCount` selector changes AND when `props.rows` changes ?
For me we can remove it.

All the test are passing.

With this last `useEffect` removal, `useGridFilter` and `useGridSorting` both have 2 `useEffect` :

- the control state registration (that will become synchronous with the sync state initialization)
- the prop to state (that could become synchronous but with some issues that I must address 1st)

It would be cool to have two major pieces of our architecture that have 100% or their state updates synchronous.
We may not succeed in having all our state updates synchronous (some edge cases are hard to solve).
But if when the user updates its rows, our internal state have directly the new sorted and filtered rows ready it would be cool (that needs some work after this PR of course).

flaviendelangle: I added a commit removing the `useEffect` you linked.

packages/grid/x-grid/src/tests/state.DataGridPro.test.tsx
```diff
@@ -62,6 +62,7 @@ describe('<DataGridPro /> - State', () => {
           ...prev,
           sorting: { ...prev.sorting, sortModel: [{ field: 'brand', sort: 'asc' }] },
         }));
+        apiRef.current.applySorting();
```
oliviertassinari, line 65: Why do we need to add this imperative action? I was under the assumption that the sorting should be applied when the state changes, e.g. with a controller prop or imperative call.

flaviendelangle: As I understand it, the sorting and filter are not applied systematically when their state change.
We have a manual `applySorting` / `applyFilter` in the update functions.

And I think before, it was doing something like this

1. Render 1 of the Grid
2. useEffect of the Grid in useGridRows, the rows are stored in state + useEffect of the Grid in useGridSorting listening to props.rows, the sorting is applied
3. useEffect of the test runs
4. useEffect of the Grid in useGridSorting listening to the rowCount selector due to the state rows change, the sorting is reapplied

With my last commit, I removed the 2nd sorting application that was after you update the state in the test

One way to confirm is to delay the state update of the test, and it fails on master

image


That's typically why I want to reduce as much as possible these async state update cascades, they make things work for the wrong reason in some tests.

flaviendelangle: Do we want the user to update the state manually ? Is it something we see as normal or is it something they should do at their own risk ?

Because most hooks have a method to cleanly update, method that can rerun other behaviors (`applySorting` for instance) and `forceRender`.

By default, a `apiRef.current.setState` does not
- rerender
- apply side effects like `applySorting`
- apply the control state constraints

For me we should at the very least migrate the `setGridState` code of `useGridState` inside `apiRef.current.setState`
`apiRef.current.setState` feels like legacy to me in its current shape

oliviertassinari: Good points, I don't think that my initial question challenges that we can merge it.

I think that the main question for me is should the application of the sorting directly depends on the state to update, or should it depends on higher level signals, like the fact that a `sortModel` prop changes.

m4theushw (commented):
It's a good step towards removing the timeout. I think, in the future, we should have a single `rowsChange` event for users to listen to. The separation between set / update doesn't explain their purposes correctly

Could you add some tests checking if the event was published?

packages/grid/x-grid/src/tests/state.DataGridPro.test.tsx
flaviendelangle: >  I think what this test is doing is a "DataGrid anti-pattern".

I agree on this
I'll do another PR

flaviendelangle: For the merge of `rowsSet` and `rowsUpdate`, I agree, that can be done later one :+1:

flaviendelangle:
I'm adding more tests in a follow up PR :+1:

PR #2557, opened 2021-09-06
[core] Synchronously subscribe to events in useGridApiEventHandler

flaviendelangle:
- [x] Subscribe to the event in the render instead of waiting for the `useEffect`. This will allow us to get the events triggered during the render phase
- [x] Improve the typing of the event publishing process
- [x] Do not call `api.emit` for the Grid unmount, it was skipping the event default value setting (add a test for it)


packages/grid/_modules_/grid/hooks/features/rows/useGridEditRows.ts
```diff
@@ -422,11 +422,10 @@ export function useGridEditRows(
   );
 
   const handleCellEditStop = React.useCallback(
-    (params: GridCellParams, event?: React.SyntheticEvent) => {
+    (params: GridCellParams, event: React.SyntheticEvent | {}) => {
```
flaviendelangle, line 425: @m4theushw do you agree that we now have a default value so the event is never null ?
Otherwise it would crash in `useGridApIEventHandler`

docs/src/pages/components/data-grid/events/SubscribeToEvents.tsx
```diff
@@ -13,11 +17,14 @@ export default function SubscribeToEvents() {
...
-    return apiRef.current.subscribeEvent('columnResize', (params) => {
-      setMessage(
-        `Column ${params.colDef.headerName} resized to ${params.computedWidth}px.`,
-      );
-    });
+    return apiRef.current.subscribeEvent(
+      'columnResize',
+      (params: GridColumnResizeParams) => {
+        setMessage(
+          `Column ${params.colDef.headerName} resized to ${params.width}px.`,
```
m4theushw, line 24: Good catch. 👍

packages/grid/_modules_/grid/utils/eventEmitter/GridEventEmitter.ts
```diff
@@ -1,15 +1,27 @@
 import * as React from 'react';
 import { EventEmitter } from './EventEmitter';
+import { MuiEvent } from '../../models/muiEvent';
+import { GridCallbackDetails } from '../../models/api/gridCallbackDetails';
 
-// TODO replace any with Generics
-export type GridListener = (params: any, event?: React.SyntheticEvent) => void;
+export type GridValidEvent = MuiEvent<
```
m4theushw, line 6: What do you think about reusing `MuiEvent`? Two types for events might be confusing.

```tsx
export type MuiEvent<E = React.SyntheticEvent | DocumentEventMap[keyof DocumentEventMap] | {}> =
  E & {
    defaultMuiPrevented?: boolean;
  };
```

docs/src/pages/components/data-grid/events/SubscribeToEvents.tsx
flaviendelangle: I was the one introducing it though :grimacing:

m4theushw (approved):

PR #2558, opened 2021-09-07
[DataGrid] Prevent scroll when selecting rows

### LINKED ISSUE #2508: [question] Issue with jumping x-grid starting from alpha.37
### Order id 💳

25229

### Duplicates

### Latest version

### Summary 💡

The grid jumps, starting with alpha.37 (alpha.36 is fine). When clicking a row and then scrolling until this row is not visible anymore, the grid jumps.

### The problem in depth 🔍

Video of what happens in alpha.36:
Peek 2021-08-31 12-20

Video of what happens in alpha.37 and later:
Peek 2021-08-31 12-20a

### Context 🔦

_No response_

### Your Environment 🌎

Chrome: Version 92.0.4515.159 (Official Build) (64-bit)

  System:
    OS: Linux 5.11 Ubuntu 21.04 (Hirsute Hippo)
  Binaries:
    Node: 16.3.0 - ~/.nvm/versions/node/v16.3.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v16.3.0/bin/yarn
    npm: 7.20.5 - ~/.nvm/versions/node/v16.3.0/bin/npm
  Browsers:
    Chrome: 92.0.4515.159
    Firefox: 91.0.2
  npmPackages:
    @material-ui/codemod: ^4.5.0 => 4.5.0 
    @material-ui/lab: 4.0.0-alpha.60 => 4.0.0-alpha.60 
    @material-ui/styles:  4.11.4 
    @material-ui/system:  4.12.1 
    @material-ui/types:  5.1.0 
    @material-ui/utils:  4.11.2 
    @material-ui/x-grid: ^4.0.0-alpha.37 => 4.0.0-alpha.37 
    @material-ui/x-license:  4.0.0-alpha.37 
    @types/react: ^17.0.19 => 17.0.19 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    typescript: ^4.3.5 => 4.3.5

m4theushw:
Fixes #2508 
Before: https://codesandbox.io/s/material-demo-forked-5u9pt?file=/demo.js
After: https://codesandbox.io/s/material-demo-forked-sw3r0?file=/demo.js

This bug is pretty interesting. Let's imagine that the first row is already selected. The selection and the focus are both listing to the click event. The user clicks into a different row and a rerender is triggered. While that, the focus state is also updated, which triggers another rerender. When the cell of the first selected row renders (it's not selected anymore) it still thinks that it has focus so it will call `.focus()`. The problem is that if this first cell is not visible, the browser will try to scroll to it. My fix is to pass an [option](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#focus_with_focusoption) to `.focus` to disable the scroll. This doesn't work in all browsers, so we use our own API to provide an alternative. I don't see a problem disabling the scroll, in this [demo](https://material-ui.com/components/data-grid/editing/#control-with-external-buttons) we're already not relaying on it but on the scroll APIs.


test/e2e/index.test.ts
```diff
@@ -194,5 +194,20 @@ describe('e2e', () => {
...
+      await page.evaluate(() =>
+        document.querySelector('[role="cell"][data-rowindex="3"]')!.scrollIntoView(),
+      );
+      expect(
+        await page.evaluate(() => document.querySelector('.MuiDataGrid-window')!.scrollTop!),
+      ).not.to.equal(0);
+      await page.click('[role="cell"][data-rowindex="3"]');
+      expect(
+        await page.evaluate(() => document.querySelector('.MuiDataGrid-window')!.scrollTop!),
+      ).not.to.equal(0);
```
flaviendelangle, line 210: Should the `document.querySelector('.MuiDataGrid-window')!.scrollTop!` be equal between the two `expect` ?
If so, we may want to check more strictly if the two are equal, and not just if they are not equal to 0

flaviendelangle (approved):

PR #2564, opened 2021-09-08
[DataGrid] Make `hideFooterRowCount` prop available only for DataGridPro

### LINKED ISSUE #2509: [DataGrid] The `hideFooterRowCount` prop should only be available for DataGridPro
### Duplicates

### Latest version

### Current Behavior 😯

Does not hide the row count

### Expected Behavior 🤔

Should hide the row count

### Steps to Reproduce 🕹

Add `hideFooterRowCount` to the props, and see it doesn't hide.

or click here - https://codesandbox.io/s/compassionate-murdock-sv1me?file=/src/Demo.tsx

Just FYI I'm using the pro datagrid, but the reproduction above is copied from a demo that uses the MIT version, which still presents the issue.

### Context 🔦

I'm building against an endpoint that won't expose exactly how much data it has, and along with that I'm doing cursor based paging. 

So my intention is to set the rowCount to the largest number possible, and then when there is less than the page size returned in data, set the rowCount to currentPage * pageSize + rows.length, which should currently disable the next page button. Ideally I would just like to access the underlying implementation (I.E `onNextPage` `onPreviousPage` `disableNextPage` `disablePreviousPage`) but alas I work with what I have for now.

### Order id 💳

28357

ZeeshanTamboli:
Closes #2509 

I added a test which is not related to the issue, but just ensures that it works.

Also, is it a breaking change? 🤔  Because for `DataGrid` typescript users typescript will complain now if they had passed the prop but actually it does not work.


flaviendelangle:
Can you add `hideFooterRowCount: false` to the `useDataGridProps` ?
If we don't do it, TypeScript complains, but a user could still set the value to `true`.

m4theushw:
@flaviendelangle I pushed a few improvements. Could you review them?

packages/grid/data-grid/src/DataGridProps.ts
```diff
@@ -14,11 +14,13 @@ export type DataGridProps = Omit<
...
   | 'disableMultipleSelection'
+  | 'hideFooterRowCount'
   | 'options'
   | 'onRowsScrollEnd'
   | 'onViewportRowsChange'
   | 'scrollEndThreshold'
   | 'signature'
 > & {
   pagination?: true;
+  hideFooterRowCount?: false;
```
flaviendelangle, line 25: Why setting `hideFooterRowCount` here instead of just removing it from the DataGrid interface ?

m4theushw: I followed the pagination example. Should we remove it?

flaviendelangle: https://github.com/mui-org/material-ui-x/pull/2176#issuecomment-887710864

I don't find it great to put prop that you can't do anything with in the interface.
So I removed all of them in #2176.

Except for `pagination` because it allows us to write

```tsx
<DataGrid pagination {...data} />
```

So the user cancopy/paste the code from a `DataGrid` demo into their `DataGridPro` and have the same pagination behavior.

But it is 100% equivalent to 

```tsx
<DataGrid {...data} />
```

But I still think that it's underlines a deeper problem, which is the different default value of `pagination` between `DataGrid` and `DataGridPro`.

I'll add a comment on the `pagination?: true` line :+1:

m4theushw:
```suggestion
```

flaviendelangle:
Looks good to me, just need to rerun the prop types genetion

flaviendelangle (approved):

PR #2566, opened 2021-09-08
[core] Only run Prettier on files different from 'next' instead of 'master'

flaviendelangle:
The `next` branch is the default one on the config of Prettier in the core

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "storybook:export": "yarn workspace storybook export",
     "deduplicate": "node scripts/deduplicate.js",
     "stylelint": "stylelint '**/*.js' '**/*.ts' '**/*.tsx'",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "storybook:export": "yarn workspace storybook export",
     "deduplicate": "node scripts/deduplicate.js",
     "stylelint": "stylelint '**/*.js' '**/*.ts' '**/*.tsx'",
-    "prettier": "node ./scripts/prettier.js --branch master",
+    "prettier": "node ./scripts/prettier.js",
```
m4theushw, line 20: I would keep it explicit. If the core changes from next to master, it will break us.

```suggestion
    "prettier": "node ./scripts/prettier.js --branch next",
```

PR #2572, opened 2021-09-09
[core] Each hook should initialize its state synchronously

### LINKED ISSUE #2293: Initialize the state synchronously with the values of the props

flaviendelangle:
Fixes #2293 
Part of #924 

Is currently rebased upon #2723

## Overview

- [x] Each hook is now responsible for its state initialization
- [x] The components should never import selectors from optional hook like `useGridColumnReorder`, `useGridColumnResize` and `useGridInfiniteLoader`.  Instead, the feature hook should inject the props in `props.componentsProps?.XXX`
- [x] The state initialization should be synchronous  except for DOM-related information
- [x] Unify model update behaviors : always check if the new model equals the old one before storing it in the state 

I plan to rework the folder structure in a future PR to put back the pro-only features inside the `data-grid-pro` folder to be able to ban the imports from `data-grid` to `data-grid-pro` with ESLint.

## Breaking changes

Merge `state.visibleRows` and `state.filter` into `state.filter` 

```diff
- state.filter
+ state.filter.filterModel
+ gridFilterModelSelector(state) // preferred method
```

```diff
- state.visibleRows.visibleRowsLookup
+ state.filter.visibleRowsLookup
+ gridVisibleRowsLookupSelector(state) // preferred method
```

## Detailed structure of a feature hook

I did not reorder the elements of every feature hooks because the order does not always cause bugs.
For instance, if you instantiate the api methods after the events listeners, it only cause problems if one of the event listeners is an api method (ex: `useGridApiEventHandler(apiRef, GridEvents.rowsSet, apiRef.current.applyFilters);`).

But we should use the following order as much as possible to avoid any issue.

Concerning the hook order, the following hook should be after `useGridSomeOtherFeature` as it uses its state in the render phase.
If possible it should be after `useGridRows` as it depends on it, but if for some reason you have to put it after, it would work.
The dependance is asynchronous (the `GridEvents.rowsSet` is published through a `useEffect`) so `useGridExample` would already be initialized for the 1st event publication even if its after in the feature list).

```ts
/**
 * @requires useGridSomeOtherFeature (state)
 * @requires useGridRows (event)
 */
export const useGridExample = (apiRef: GridApiRef, props: GridComponentProps) => {
  // 1. LOGGER
  const logger = useGridLogger(apiRef, 'useGridExample');

  // 2. STATE INITIALISATION
  useGridStateInit(apiRef, (state) => ({
    ...state,
    example: {
      exampleModel: props.exampleModel ?? getDefaultExampleModel(),
    },
  }));

  // 3. STATE SELECTORS AND UPDATER
  const [, setGridState, forceUpdate] = useGridState(apiRef);
  const someOtherFeatureSubState = useGridSelector(apiRef, gridSomeOtherFeatureSubSelector);


  // 4. SYNCHRONOUS CONTROL STATE REGISTRATION
  useGridRegisterControlState(apiRef, {
    stateId: 'example',
    propModel: props.exampleModel,
    propOnChange: props.onExampleModelChange,
    stateSelector: gridExampleModelSelector,
    changeEvent: GridEvents.exampleChange,
  });

  // 5. API METHODS
  const setExampleModel = React.useCallback<GridExampleApi['setExampleModel']>(
    (model) => {
      const currentModel = gridExampleModelSelector(apiRef.current.state);

      if (currentModel !== model) {
        logger.debug('Setting example model');
  
        setGridState((state) => ({
          ...state,
          example: {
            ...state.example,
            exampleModel: model,
          },
        }));

        // If we have complex post update logic to run after model update, we must call it here
        // If not, remember to call `forceUpdate` here
        apiRef.current.applyExample();
      }
    },
    [apiRef, setGridState, logger],
  );

  const applyExample = React.useCallback<GridExampleApi['applyExample']>(
    () => { /* Some code to run whenever the model or the rows are updated */ },
    []
  )

  const exampleApi: GridExampleApi = {
    setExampleModel,
    applyExample,
  };

  useGridApiMethod(apiRef, exampleApi, 'GridExampleApi');

  // 6. UPDATE EFFECTS
  React.useEffect(() => {
    if (props.exampleModel != null) {
      apiRef.current.setExampleModel(props.exampleModel);
    }
  }, [apiRef, props.exampleModel]);

  // 7. FIRST RUN COMPLEX STATE INITIALIZATION
  // Some logic for the state initialization requires to much of the hook codebase to be placed inside `useGridStateInit`.
  //  We add a callback here to run it synchronously during the 1st render, but after the initialization of the api methods.
  useFirstRender(() => apiRef.current.applyExample());

  // 8. EVENT CALLBACKS
  const handleColumnHeaderClick = React.useCallback(
    () => { /* Some code to run when the user clicks on a column header */ },
    [],
  );

  useGridApiEventHandler(apiRef, GridEvents.rowsSet, this.apiRef.applyExample);
  useGridApiEventHandler(apiRef, GridEvents.columnHeaderClick, handleColumnHeaderClick);
};
```

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
+++ b/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
@@ -17,7 +17,7 @@ import { GridPreferencePanelsValue } from '../preferencesPanel/gridPreferencePan
 import { sortedGridRowsSelector } from '../sorting/gridSortingSelector';
 import { getInitialGridFilterState } from './gridFilterModelState';
 import { GridFilterModel } from '../../../models/gridFilterModel';
-import { visibleSortedGridRowsSelector } from './gridFilterSelector';
+import { gridFilterModelSelector, visibleSortedGridRowsSelector } from './gridFilterSelector';
 import { getInitialVisibleGridRowsState } from './visibleGridRowsState';
 
 /**
@@ -278,11 +278,11 @@ export const useGridFilter = (
 
   const onColUpdated = React.useCallback(() => {
     logger.debug('onColUpdated - GridColumns changed, applying filters');
-    const filterState = apiRef.current.state.filter;
+    const filterModel = gridFilterModelSelector(apiRef.current.state);
     const columnsIds = filterableGridColumnsIdsSelector(apiRef.current.state);
     logger.debug('GridColumns changed, applying filters');
 
-    filterState.items.forEach((filter) => {
+    filterModel.items.forEach((filter) => {
       if (!columnsIds.find((field) => field === filter.columnField)) {
         apiRef.current.deleteFilter(filter);
       }
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -114,7 +114,7 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
 
   const setSelectionModel = React.useCallback<GridSelectionApi['setSelectionModel']>(
     (model) => {
-      const currentModel = apiRef.current.state.selection;
+      const currentModel = gridSelectionStateSelector(apiRef.current.state);
       if (currentModel !== model) {
         setGridState((state) => ({ ...state, selection: model }));
         forceUpdate();
--- a/packages/grid/_modules_/grid/components/containers/GridRoot.tsx
+++ b/packages/grid/_modules_/grid/components/containers/GridRoot.tsx
@@ -6,10 +6,10 @@ import { GridRootContainerRef } from '../../models/gridRootContainerRef';
 import { useStyles } from './GridRootStyles';
 import { visibleGridColumnsLengthSelector } from '../../hooks/features/columns/gridColumnsSelector';
 import { useGridSelector } from '../../hooks/features/core/useGridSelector';
-import { useGridState } from '../../hooks/features/core/useGridState';
 import { useGridApiContext } from '../../hooks/root/useGridApiContext';
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
 import { gridClasses } from '../../gridClasses';
+import { gridRowCountSelector } from '../../hooks/features/rows/gridRowsSelector';
 
 export type GridRootProps = React.HTMLAttributes<HTMLDivElement>;
 
@@ -22,7 +22,7 @@ export const GridRoot = React.forwardRef<HTMLDivElement, GridRootProps>(function
   const rootProps = useGridRootProps();
   const { children, className: classNameProp, ...other } = props;
   const visibleColumnsLength = useGridSelector(apiRef, visibleGridColumnsLengthSelector);
-  const [gridState] = useGridState(apiRef);
+  const totalRowCount = useGridSelector(apiRef, gridRowCountSelector);
   const rootContainerRef: GridRootContainerRef = React.useRef<HTMLDivElement>(null);
   const handleRef = useForkRef(rootContainerRef, ref);
 
@@ -44,7 +44,7 @@ export const GridRoot = React.forwardRef<HTMLDivElement, GridRootProps>(function
         )}
         role="grid"
         aria-colcount={visibleColumnsLength}
-        aria-rowcount={gridState.rows.totalRowCount}
+        aria-rowcount={totalRowCount}
         aria-multiselectable={!rootProps.disableMultipleSelection}
         aria-label={rootProps['aria-label']}
         aria-labelledby={rootProps['aria-labelledby']}
--- a/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
+++ b/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
@@ -1,24 +1,22 @@
 import * as React from 'react';
 import Button from '@mui/material/Button';
-import { useGridState } from '../../../hooks/features/core/useGridState';
 import { GridFilterItem, GridLinkOperator } from '../../../models/gridFilterItem';
 import { useGridApiContext } from '../../../hooks/root/useGridApiContext';
-import { GridAddIcon } from '../../icons/index';
+import { GridAddIcon } from '../../icons';
 import { GridPanelContent } from '../GridPanelContent';
 import { GridPanelFooter } from '../GridPanelFooter';
 import { GridPanelWrapper } from '../GridPanelWrapper';
 import { GridFilterForm } from './GridFilterForm';
 import { useGridRootProps } from '../../../hooks/utils/useGridRootProps';
+import { useGridSelector } from '../../../hooks/features/core/useGridSelector';
+import { gridFilterModelSelector } from '../../../hooks/features/filter/gridFilterSelector';
 
 export function GridFilterPanel() {
   const apiRef = useGridApiContext();
-  const [gridState] = useGridState(apiRef);
   const rootProps = useGridRootProps();
+  const filterModel = useGridSelector(apiRef, gridFilterModelSelector);
 
-  const hasMultipleFilters = React.useMemo(
-    () => gridState.filter.items.length > 1,
-    [gridState.filter.items.length],
-  );
+  const hasMultipleFilters = filterModel.items.length > 1;
 
   const applyFilter = React.useCallback(
     (item: GridFilterItem) => {
@@ -46,15 +44,15 @@ export function GridFilterPanel() {
   );
 
   React.useEffect(() => {
-    if (gridState.filter.items.length === 0) {
+    if (filterModel.items.length === 0) {
       addNewFilter();
     }
-  }, [addNewFilter, gridState.filter.items.length]);
+  }, [addNewFilter, filterModel.items.length]);
 
   return (
     <GridPanelWrapper>
       <GridPanelContent>
-        {gridState.filter.items.map((item, index) => (
+        {filterModel.items.map((item, index) => (
           <GridFilterForm
             key={item.id == null ? index : item.id}
             item={item}
@@ -62,7 +60,7 @@ export function GridFilterPanel() {
             deleteFilter={deleteFilter}
             hasMultipleFilters={hasMultipleFilters}
             showMultiFilterOperators={index > 0}
-            multiFilterOperator={gridState.filter.linkOperator}
+            multiFilterOperator={filterModel.linkOperator}
             disableMultiFilterOperator={index !== 1}
             applyMultiFilterOperatorChanges={applyFilterLinkOperator}
           />
--- a/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
+++ b/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts
@@ -16,12 +16,13 @@ import { GridColumnOrderChangeParams } from '../../../models/params/gridColumnOr
 import { mergeGridColTypes } from '../../../utils/mergeUtils';
 import { useGridApiMethod } from '../../root/useGridApiMethod';
 import { useGridLogger } from '../../utils/useGridLogger';
-import { useGridSelector } from '../core/useGridSelector';
 import { GridLocaleText, GridTranslationKeys } from '../../../models/api/gridLocaleTextApi';
 import { useGridState } from '../core/useGridState';
 import {
+  allGridColumnsFieldsSelector,
   allGridColumnsSelector,
   gridColumnsMetaSelector,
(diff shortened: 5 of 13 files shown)


flaviendelangle:
@mui-org/x I think its ready enough to start reviewing

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
```diff
@@ -340,4 +351,23 @@ export const useGridColumnResize = (
...
+      ...props.componentsProps,
+      columnsHeader: {
+        ...props.componentsProps?.columnsHeader,
+        resizeCol: resizingColumnField,
+      },
+    }),
+    [props.componentsProps, resizingColumnField],
+  );
+
+  return React.useMemo(
```
m4theushw, line 366: Why to return props if they are not used in `useDataGridProComponent`?

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -11,20 +11,36 @@ import { useGridApiEventHandler } from '../../root/useGridApiEventHandler';
...
+import { useGridRegisterControlState } from '../../utils/useGridRegisterControlState';
+import { useGridStateInit } from '../../utils/useGridStateInit';
+import { useFirstRender } from '../../utils/useFirstRender';
+
+const checkFilterModelValidity = (model: GridFilterModel) => {
+  if (model.items.length > 1) {
+    const hasItemsWithoutIds = model.items.find((item) => item.id == null);
+    if (hasItemsWithoutIds) {
+      throw new Error(
+        "The 'id' field is required on filterModel.items when you use multiple filters.",
```
m4theushw, line 33: Why is it required?

```suggestion
        "MUI: The 'id' field is required on `filterModel.items` when you use multiple filters.",
```

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -39,6 +39,19 @@ export const useGridSelection = (apiRef: GridApiRef, props: GridComponentProps):
     return [props.selectionModel];
   }, [props.selectionModel]);
 
+  useGridStateInit(apiRef, (state) => ({ ...state, selection: propSelectionModel ?? [] }));
+
+  const [, setGridState, forceUpdate] = useGridState(apiRef);
+  const rowsLookup = useGridSelector(apiRef, gridRowsLookupSelector);
+
+  const isStateControlled = useGridRegisterControlState(apiRef, {
```
m4theushw, line 47: This new `useGridRegisterControlState` hook has two concerns: registering the state and checking if the prop is controlled. I would call `updateControlState` as before and check the prop to see if it is controlled. Its second purpose is only used once and here.

packages/grid/_modules_/grid/models/gridSlotsComponentsProps.ts
```diff
@@ -1,10 +1,13 @@
+import type { GridColumnsHeaderProps } from '../components/columnHeaders/GridColumnHeaders';
+
 /**
  * Overrideable components props dynamically passed to the component at rendering.
  */
 export interface GridSlotsComponentsProps {
   checkbox?: any;
   columnMenu?: any;
   columnsPanel?: any;
+  columnsHeader?: GridColumnsHeaderProps;
```
m4theushw, line 10: Why did you add a new prop? There's no respective slot for it.

m4theushw (commented):
I didn't look through all files. I'll do a deeper check later. This PR could have been divided in two: the synchronous state initialization and a second one with the `state.key` -> `keySelector(state)` migration. It would be easier to review with less noise. 😁

packages/grid/_modules_/grid/hooks/features/columnResize/useGridColumnResize.tsx
flaviendelangle: Oh I forgot to use them, thanks !

packages/grid/_modules_/grid/models/gridSlotsComponentsProps.ts
flaviendelangle: It's related to 

>  The components should never import selectors from optional hook like useGridColumnReorder, useGridColumnResize and useGridInfiniteLoader. Instead, the feature hook should inject the props in props.componentsProps?.XXX

m4theushw: It's also related to https://github.com/mui-org/material-ui-x/issues/2293#issuecomment-914266340.

The purpose of `componentsProps` was to allow to change the props of inner components without having to recreate the entire grid. However, the `GridColumnsHeader` has no prop for the user to change. I agree to add new options to `GridSlotsComponentsProps` for those components that could be relevant to allow users to change their props, but not for internal components that only we need to change.

To provide an alternative, we can use events to get the current column being dragged/resized. In these two cases, I was already advocating that these keys could be removed from the state and replaced with events. See https://github.com/mui-org/material-ui-x/issues/820#issuecomment-897906608 and https://github.com/mui-org/material-ui-x/pull/2337

For the column resizing, the selector is first called in `GridColumnHeadersItemCollection` but its value is only used in `GridColumnHeaderSeparator`. The tree is: 

```
GridColumnHeadersItemCollection -> GridColumnHeaderItem -> GridColumnHeaderSeparator
```

For column dragging, it doesn't change too much. The selector is called in `GridColumnsHeader`. Then, its value is used in this first component and passed to `GridColumnHeaderItem`. There's one component in the middle that doesn't need it. It's not a complex task to replace with events. The three is:

```
GridColumnsHeader -> GridColumnHeadersItemCollection -> GridColumnHeaderItem
```

What do you think about using events?

flaviendelangle: I think its worth taking a look at the future pro-only features and see if the information we want to pass to the components can easily be implemented via events.
To avoid having to find another mechanism in a few weeks.

But on paper, events are probably a good approach.
As long as we accept that the event enum is on the `x-data-grid` version, even for paid-only events.
Which won't be a problem for bundle size.
And for separation of concern, it's not perfect, but if we wanted to really put all the code related to a paid-only feature inside the pro package, then we would have the same discussions for the localization and I don't think we want to split it.

I'll try an event-based implementation tomorrow to see how it looks.

m4theushw: Another PR only removing the selector and listening to the event is better. There's no need to remove the key from the state right now.

flaviendelangle: To merge this PR, I need to remove the selector from the component because the state is never set on the DataGrid, only on the DataGridPro now that its the hook which initialize its own state.

But I can do it on another PR, it just need to be merged before this one.

flaviendelangle: See #2716

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: What do you mean by "call updateControlState as before"

I can remove the 2nd purpose to simplify the understanding of this hook.
The important point of this hook is that it registers synchronously the controlled state rather than in an effect.

m4theushw: I was saying that we could remove `useGridRegisterControlState` and call `apiRef.current.updateControlState` directly, but in a synchronous way. The new hook seems to be only a wrapper.

flaviendelangle: I thought that it would eventually have more logic inside it, but the PR is ready and there is none so I suppose we can indeed remove it :+1:

PR #2574, opened 2021-09-09
[DataGrid] Force scrollEndThreshold to undefined

flaviendelangle:


m4theushw:
Although the infinite loader is a Pro feature, its hook is added to the DataGrid:

https://github.com/mui-org/material-ui-x/blob/305affa79b0723a8005cabb31026dcb9eadffdc2/packages/grid/data-grid/src/useDataGridComponent.tsx#L62

DanailH (approved):

DanailH:
> Although the infinite loader is a Pro feature, its hook is added to the DataGrid:
> 
> https://github.com/mui-org/material-ui-x/blob/305affa79b0723a8005cabb31026dcb9eadffdc2/packages/grid/data-grid/src/useDataGridComponent.tsx#L62

I checked yesterday - we would need to audit the hooks that are added to the `DataGrid`. I guess that one is in both places because it was created before the split happened and it was duplicated in both places. The same goes for the `useColumnResize` one

flaviendelangle:
@m4theushw what in the ` useGridInfiniteLoader` code stop it from working on the free version ?

flaviendelangle:
I added an object with a strict typing so the TypeScript complains if we forget to set a forced prop.

DanailH:
> @m4theushw what in the ` useGridInfiniteLoader` code stop it from working on the free version ?
> We can remove it from `useDataGridComponent`, I only removed `useGridColumnsReorder` but it their are other to remove we should indeed do it.

`useGridInfiniteLoader` is a paid feature only so I guess it makes sense to only be included in the `useDataGridProComponent`.

m4theushw (approved):

PR #2571, opened 2021-09-09
[core] Support for @mui packages

m4theushw:
Fixes #2527

PR #2575, opened 2021-09-09
[core] Remove v4 conditional code

flaviendelangle:
I wait for #2571 before doing those who are more than just remove some branches (for instance replace the `muiStyleAlpha` function by a direct call to `alpha` from the core

Preview: https://deploy-preview-2575--material-ui-x.netlify.app/components/data-grid/


packages/grid/_modules_/grid/utils/utils.ts
```diff
@@ -15,23 +15,6 @@ export function isObject(value: any): value is Record<string, any> {
...
-  }
-
-  if ('fade' in styles && 'alpha' in styles) {
-    return 'v4.12';
-  }
-
-  return 'v4';
-}
-
 export function muiStyleAlpha(color: string, value: number): string {
```
m4theushw, line 18: This helper can be removed now.

packages/grid/_modules_/grid/components/GridPagination.tsx
```diff
@@ -60,26 +60,12 @@ export const GridPagination = React.forwardRef<
   );
 
   const handlePageChange = React.useCallback(
```
m4theushw, line 62:
```suggestion
  const handlePageChange = React.useCallback<TablePaginationProps['onPageChange']>(
```

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -1,13 +1,13 @@
...
+import { muiStyleAlpha, createTheme } from '../../utils/utils';
 import { gridClasses } from '../../gridClasses';
 
 const defaultTheme = createTheme();
 export const useStyles = makeStyles(
   (theme: Theme) => {
     const borderColor =
-      getThemePaletteMode(theme.palette) === 'light'
+      theme.palette.mode === 'light'
         ? lighten(muiStyleAlpha(theme.palette.divider, 1), 0.88)
```
m4theushw, line 11: `alpha` from `@mui/material/styles` can be used now.

```suggestion
        ? lighten(alpha(theme.palette.divider, 1), 0.88)
```

packages/grid/_modules_/grid/utils/utils.ts
flaviendelangle: Yes some changes needed you migration to `@mui/`, I'll complet this PR :+1:

packages/grid/_modules_/grid/components/GridPagination.tsx
flaviendelangle: A lot better indeed

m4theushw: It won't work with `onRowsPerPageChange`, I tried. Maybe the typings in the core are wrong.

oliviertassinari (commented):
It looks like it's going in the right direction.

flaviendelangle:
@oliviertassinari it's always a pleasure to remove this kind of temporary code

flaviendelangle:
Fixing this comment https://github.com/mui-org/material-ui-x/blob/next/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx#L51 causes the related test to fail

I'll check and remove the comment if the v5 still needs the `|| isIndeterminate`

flaviendelangle:
I think we can move on without https://github.com/mui-org/material-ui-x/pull/2575#issuecomment-919251610 and fix the more complex one later

PR #2579, opened 2021-09-10
[DataGridPro] Remove apiRef getState method

flaviendelangle:


m4theushw (approved):

oliviertassinari:
I have updated a bit the PR's description for the changelog and the curious developers. From what I understood in the weekly meeting:

- apiRef.current.state is currently public
- we plan to make changes to this state, which will be breaking changes, in theory, and in practice might not impact a lot of projects.
- we are in the search for a way to balance maintainability and flexibility in how the state is accessed, where we have a bunch of options:
  - only use state selectors
  - only use api methods getter
  - ban direct state reads, internally and externally
  - remove documentation on,`apiRef.current.state` consider it private 
  - hybrid in the above.

PR #2581, opened 2021-09-10
[DataGrid] Apply `valueFormatter` to `singleSelect` select options

### LINKED ISSUE #2514: [DataGrid] `valueFormatter` should be applied to `valueOptions` in select menu
### Duplicates

### Latest version

### Summary 💡

When using the `singleSelect` column type in a DataGrid, if a `valueFormatter` is specified, it should act on the menu items in the select component, not just the currently displayed value.

### Examples 🌈

Suppose I have a list of user roles that are stored as integers in the database, but I want to display them as text. I'd write something like this:
```javascript
{
  field: "role",
  editable: true,
  valueOptions: [0, 1, 2],
  valueFormatter: ({ value }) => ["User", "Admin", "Superuser"][value]
}
```
However, while the selected menu item shows up correctly in the table, when I try to edit the role I see the raw values in the menu instead of the formatted values.

Not editing | Editing (current behavior) | Editing (expected behavior)
----------- | -------------------------- | ---------------------------
Not editing | Menu with raw values | Menu with formatted values

### Motivation 🔦

This can currently be accomplished manually by providing an array of objects with label and value fields to `valueOptions`. However, applying the value formatter when an array of strings is passed as value options makes sense as a default, since the displayed text should always match the corresponding text in the list of options. Indeed, this is how the `<Select>` component behaves. Additionally, it would save users several lines of code by cutting back on repetition.

### Order id 💳

_No response_

DanailH:
Fixes #2514 

---

### Breaking changes:

```diff
-export type GridValueFormatterParams = Omit<GridRenderCellParams, 'formattedValue' | 'isEditable'>;
+export interface GridValueFormatterParams {
+  /**
+   * The column field of the cell that triggered the event
+   */
+  field: string;
+  /**
+   * The cell value, but if the column has valueGetter, use getValue.
+   */
+  value: GridCellValue;
+  /**
+   * GridApi that let you manipulate the grid.
+   */
+  api: any;
+}
```


packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -40,6 +40,19 @@ function GridEditSingleSelectCell(props: GridRenderEditCellParams & SelectProps)
...
+  let valueOptionsFormatted = colDef.valueOptions;
+
+  if (colDef.valueFormatter) {
+    valueOptionsFormatted = colDef.valueOptions.map((option) =>
+      // valueFormatter is only applied to primitive type valueOptions
+      typeof option === 'object'
+        ? option
+        : {
+            value: option,
+            label: colDef.valueFormatter({ ...props, value: option }),
```
m4theushw, line 52:
```suggestion
            label: String(colDef.valueFormatter({ ...props, value: option }),
```

m4theushw (commented):
We miss applying the `valueFormatter` to the filter component too:

image

---

If you apply the following diff, you'll get more meaningful types. Specially about the parameters to pass when calling the value formatter.

```diff
diff --git a/packages/grid/_modules_/grid/models/params/gridCellParams.ts b/packages/grid/_modules_/grid/models/params/gridCellParams.ts
index 369a19a6..31ceedc8 100644
--- a/packages/grid/_modules_/grid/models/params/gridCellParams.ts
+++ b/packages/grid/_modules_/grid/models/params/gridCellParams.ts
@@ -69,7 +69,7 @@ export interface GridRenderCellParams extends GridCellParams {
 /**
  * GridEditCellProps containing api.
  */
-export interface GridRenderEditCellParams extends GridEditCellProps {
+export interface GridRenderEditCellParams extends GridEditCellProps, GridCellParams {
   /**
    * GridApi that let you manipulate the grid.
    */
```

DanailH:
@m4theushw regarding the 2 points. The second one  - for extending the `GridRenderEditCellParams` with `GridCellParams ` - I would keep it as is for now because changing this creates other type errors.

For the point in applying the `valueFormatter` to the filter - that is also questionable as the `valueFormatter` requires cell-specific information that does not exist in the `GeidFilterInputValue` component and I can't get it since the filter is not opened for a specific cell.

m4theushw:
> For the point in applying the valueFormatter to the filter - that is also questionable as the valueFormatter requires cell-specific information that does not exist in the GeidFilterInputValue component and I can't get it since the filter is not opened for a specific cell.

The problem I see is that only applying the `valueFormatter` to the cell solves 50% of the issue. Once the user opens the filter, the options will not be formatted. 

Looking at the definition of `GridValueFormatterParams` there are only a few properties tied to cell. The only things we can't provide is the `id`, ` row` and some accessory information. We have the `colDef` and `value` is the filter value. Since we created a dedicated type for the value formatter, it doesn't have to have the same members from `GridCellParams`. What do you think about making the following properties optional?

```diff
-id: GridRowId;
field: string;
value: GridCellValue; // The filter value that the user has selected
-row: GridRowModel;
colDef: GridStateColDef;
-cellMode: GridCellMode;
-hasFocus: boolean;
-tabIndex: 0 | -1;
getValue: (id: GridRowId, field: string) => GridCellValue;
api: any;
```

DanailH:
@m4theushw I was thinking of going one step further and only passing in the value, field, and API. I don't think the rest are needed.

packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -40,6 +40,19 @@ function GridEditSingleSelectCell(props: GridRenderEditCellParams & SelectProps)
...
+  let valueOptionsFormatted = colDef.valueOptions;
+
+  if (colDef.valueFormatter) {
+    valueOptionsFormatted = colDef.valueOptions.map((option) =>
+      // valueFormatter is only applied to primitive type valueOptions
+      typeof option === 'object'
+        ? option
+        : {
+            value: option,
+            label: String(colDef.valueFormatter({ ...props, value: option })),
```
m4theushw, line 52: We could type the `valueFormatter` params argument with:

```diff
diff --git a/packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx b/packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
index 114676df..f3bfbf4f 100644
--- a/packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridEditSingleSelectCell.tsx
@@ -43,15 +43,19 @@ function GridEditSingleSelectCell(props: GridRenderEditCellParams & SelectProps)
   let valueOptionsFormatted = colDef.valueOptions;
 
   if (colDef.valueFormatter) {
-    valueOptionsFormatted = colDef.valueOptions.map((option) =>
+    valueOptionsFormatted = colDef.valueOptions.map((option) => {
       // valueFormatter is only applied to primitive type valueOptions
-      typeof option === 'object'
-        ? option
-        : {
-            value: option,
-            label: String(colDef.valueFormatter({ ...props, value: option })),
-          },
-    );
+      if (typeof option === 'object') {
+        return option;
+      }
+
+      const params: GridValueFormatterParams = { field, api, value: option };
+
+      return {
+        value: option,
+        label: String(colDef.valueFormatter(params)),
+      };
+    });
   }
 
   const handleChange = (event) => {
```

m4theushw (commented):
Looks good to me. Since you removed properties from `GridValueFormatterParams`, this is a breaking change. Could you update the description to mention that?

The `valueFormatter` is still being called with all the removed properties. I think this may cause confusion between the BC and what users receive as argument. We could also update below to only call with the correct params.

https://github.com/mui-org/material-ui-x/blob/68ae5d5d477500ba56564df10a215eb7ad2e22d9/packages/grid/_modules_/grid/hooks/features/rows/useGridParamsApi.ts#L118

PR #2520, opened 2021-09-01
[core] Provide theme augmentation as separate module

m4theushw:
See #2515 for context.

Preview: https://codesandbox.io/s/create-react-app-with-typescript-forked-z59p4?file=/src/index.tsx

PR #2582, opened 2021-09-10
[docs] Explain how to use theme augmentation

ZeeshanTamboli:
Since we dropped support for v4 in #2515 and use the same theme overrides key for DataGridPro, I renamed it in the docs as well to `MuiDataGrid`.

Also since we now tell to instead manually import `themeAugmentation` folder, we add the docs for it similarly like https://next.material-ui.com/components/about-the-lab/#typescript. I added it in the `Getting Started` page.

I think we want to wait for v5 docs infra support #2441 before we merge this.

Also maybe for #2520.

Preview:

- https://deploy-preview-2582--material-ui-x.netlify.app/api/data-grid/data-grid-pro/#component-name
- https://deploy-preview-2582--material-ui-x.netlify.app/components/data-grid/getting-started/#typescript


ZeeshanTamboli:
Now that #2441 and #2520 is merged, I think we can consider this for next release. Or maybe wait for https://github.com/mui-org/material-ui/pull/28357 v5 release for the doc hyperlinks in this PR to work?

m4theushw:
I'll merge this one after today's release. We only need to make the example clear, but I can take care of that. Thanks for updating the docs!

m4theushw (approved):

PR #2602, opened 2021-09-13
[docs] Add section for controlled selection and server-side pagination

### LINKED ISSUE #1782: [DataGrid] Prevent the grid from cleaning up the `selectionModel` when `selectionModel` prop is provided

DanailH:
Fixes #1782 

Updated the docs with the explanation of why the `selectionModel` might be cleared when new rows are loaded on page change when `paginationMode="server"`.


docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -62,8 +62,16 @@ In the demo below only rows with quantity above 50000 can be selected:
 
 ## Controlled selection
 
+Use the `selectionModel` prop to control the grid selection. Each time the `selectionModel` changes the `onSelectionModelChange` callback is called with the new selection model value.
+
 {{"demo": "pages/components/data-grid/selection/ControlledSelectionGrid.js", "bg": "inline"}}
 
+### Controlled selection and server-side pagination
+
+Using controlled selection when `paginationMode="server"` can result in the `DataGrid` clearing the provided `selectionModel` on page change. Depending on your implementation of the server-side pagination, when the page changes and there are no rows in the grid with ids equal to the `GridRowId`s provided to the `selectionModel` the grid will clear the `selectionModel`. To prevent this either save the `selectionModel` and restore it later or append the newly loaded grid rows to the existing rows.
```
flaviendelangle, line 71: > and restore it later

Some people will probably struggle to implement it.
Maybe we could also add an example for this part, or just take for granted that appending the rows is the way we want to encourage.

rjv: > Maybe we could also add an example for this part

This would be greatly appreciated if you can accommodate.

rjv:
> append the newly loaded grid rows to the existing rows

Doesn't this break the functionality of the "select all" checkbox in the grid header? Using [this demo](https://codesandbox.io/s/material-demo-forked-ok333?file=/demo.tsx), try the following:

1. Let the data for page 1 load
2. Browse to page 2
3. Click the "select all" checkbox in the grid header
4. Observe that there are **10 rows selected** and not the expected 5

docs/src/pages/components/data-grid/selection/selection.md
DanailH: I'm opened to suggestions here. It's unclear what the desired behaviour should be. For me, the way it is looks correct but so does the initial implementation in the ticket (not appending the rows). Because the selection is controlled the only solution I can see is either to leave it as it is or prevent the grid from cleaning up the `selectionModel` if the rows are missing.

DanailH:
> > append the newly loaded grid rows to the existing rows
> 
> Doesn't this break the functionality of the "select all" checkbox in the grid header? Using [this demo](https://codesandbox.io/s/material-demo-forked-ok333?file=/demo.tsx), try the following:
> 
> 1. Let the data for page 1 load
> 2. Browse to page 2
> 3. Click the "select all" checkbox in the grid header
> 4. Observe that there are **10 rows selected** and not the expected 5

This is how the Select All feature works when the pagination is client-side.

rjv:
> This is how the Select All feature works when the pagination is client-side.

This feels unintuitive for the `paginationMode="server"` use case. Having elements from `rows` hidden _only_ because of the page size is a work-around at best. This introduces another gotcha that I think developers should at least be made aware of, especially if you're making an "official" recommendation in the docs.

docs/src/pages/components/data-grid/selection/selection.md
DanailH: We should probably also take this https://github.com/mui-org/material-ui-x/pull/2602#issuecomment-919173644 into consideration. If we want to fix that as well we will need to change the way the `selectionModel` behaves as the Select All feature is not really server-side friendly at the moment.

flaviendelangle: Indeed

m4theushw, line 71: I tried to make it less technical.

```suggestion
Using the controlled selection with `paginationMode="server"` may result in selected rows being lost when the page is changed.
This happens because the grid cross-checks with the `rows` prop and only calls `onSelectionModelChange` with existing row IDs.
Depending on your server-side implementation, when the page changes and the new value for the `rows` prop does not include previously selected rows, the grid will call `onSelectionModelChange` with an empty value.
To prevent this unwanted behavior, there are two ways:

- Save the `selectionModel` **before the page is changed** and restore it later
- Append the newly loaded rows to the existing rows

The following demo shows to implement the second solution:
```

m4theushw (commented):
We could add a note in https://mui.com/components/data-grid/pagination/#server-side-pagination linking to the new section.

docs/src/pages/components/data-grid/selection/selection.md
m4theushw: Looking again and https://github.com/mui-org/material-ui-x/pull/2602#issuecomment-919045908 is right, the Select All is broke. Adding `checkboxSelectionVisibleOnly` doesn't help and it still selects the appended rows. We can try the first solution which is a bit more hacky or we investigate why the prop didn't work.

DanailH: True, but I would treat that as a separate issue and investigate it in an independent PR. This one should be to just add clarity around the relationship between the server-side pagination and the selection model.

DanailH:
@m4theushw @flaviendelangle if we agree to investigate the select all issue in a separate PR should we proceed with merging this one?

flaviendelangle:
> @m4theushw @flaviendelangle if we agree to investigate the select all issue in a separate PR should we proceed with merging this one?

@DanailH the select all issue is not caused by this PR so no problem for fixing it on a following one.

I take a last look at the texts of this PR and I approve :+1:

flaviendelangle (approved):

rjv:
I devised a crude work-around that attempts to fix the issue of "select all" selecting more than the current page by using `isRowSelectable`: https://codesandbox.io/s/material-demo-forked-f0mxc?file=/demo.tsx. Still thinking through edge cases but it's a working premise.

It doesn't automatically select rows on any _new_ pages that user hasn't already browsed, so I'd have to keep track of an "all selected" state independent of the data grid and trigger `apiRef.current.selectRows(apiRef.current.getAllRowIds())` after new data loads.

DanailH:
For now, we can go with the option @m4theushw proposed (with saving the `selectionModel` on page change and then restoring it). Both options are listed as potential solutions, although they behave differently. Either way, a fix for this will be provided in a separate PR.

docs/src/pages/components/data-grid/pagination/pagination.md
```diff
@@ -43,6 +43,8 @@ Finally, you need to handle the `onPageChange` callback to load the rows for the
 
 {{"demo": "pages/components/data-grid/pagination/ServerPaginationGrid.js", "bg": "inline"}}
 
+**Note**: For more information regarding server-side pagination in combination with controlled selection check [here](/components/data-grid/selection/#usage-with-server-side-pagination)
```
oliviertassinari, line 46: Two feedback on this one. A sentence should have a closing dot. "here" is considered to be avoided as link description: https://www.w3.org/QA/Tips/noClickHere. Maybe?

```suggestion
**Note**: For more information regarding server-side pagination in combination with controlled selection check [the dedicated section](/components/data-grid/selection/#usage-with-server-side-pagination).
```

PR #2606, opened 2021-09-14
[DataGrid] Free up column header space when icons are not visible

DanailH:
Based on this https://github.com/mui-org/material-ui-x/issues/2460

A small improvement that frees up column header space when column icons are not visible.

You can check Argos for more visual representation 😉

Before:
before

After:
after

### DIFF
--- a/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
+++ b/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
@@ -102,7 +102,8 @@ export const useStyles = makeStyles(
           alignItems: 'center',
         },
         [`& .${gridClasses['columnHeader--sorted']} .${gridClasses.iconButtonContainer}`]: {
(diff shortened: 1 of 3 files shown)


packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -123,6 +127,8 @@ export const useStyles = makeStyles(
         },
         [`& .${gridClasses.iconButtonContainer}`]: {
           display: 'flex',
+          visibility: 'hidden',
```
m4theushw, line 130: Why to change the visibility if we have `opacity: 0`?

DanailH: Ah you are right. Only `width: 0` will work.

m4theushw (approved):

PR #2669, opened 2021-09-22
[docs] Add type to `onChange` event argument

jayariglesias:


docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -116,7 +116,7 @@ export default function QuickFilteringGrid() {
         componentsProps={{
           toolbar: {
             value: searchText,
-            onChange: (event) => requestSearch(event.target.value),
+            onChange: (event: any) => requestSearch(event.target.value),
```
m4theushw, line 119: Thanks for your contribution. What do you think about pushing a little bit further?

```suggestion
            onChange: (event: React.ChangeEvent<HTMLInputElement>) => requestSearch(event.target.value),
```

m4theushw (approved):

PR #2673, opened 2021-09-23
[DataGrid] New virtualization implementation

m4theushw:


DanailH:
Ok, check the implementation and I can really add much as I haven't researched the topic. From what I see after playing with the grid it looks much smoother than before. I also tested some of the features I know are dependant on the virtualizations and this is what I found:
- `onRowsScrollEnd` - it is called when the scroll reaches the bottom. The problem is that it is also called in the beginning when the grid loads with the first and last row both equal to 0.
- `onViewportRowsChange` - is not called at all. Before when the `renderState` changes then a check was made to see if the first and last rows are the same and if not then `GridEvents.viewportRowsChange` event was emitted that called the `onViewportRowsChange`. I guess now we don't need to keep track of the `renderState` but rather the `rowBuffer` and do the same (it will be actually simpler). This is the part that is responsible for that logic -> https://github.com/mui-org/material-ui-x/blob/next/packages/grid/_modules_/grid/hooks/features/infiniteLoader/useGridInfiniteLoader.ts#L94
On another hand, it seems that the test we have doesn't really work, because if it did then it would have failed... 😅

flaviendelangle:
image

Not sure this diff is intended
Looks like the last row now want to be fully visible, causing the padding of the pagination to be lost.

PR #2686, opened 2021-09-24
[Charts] PR for the purpose of a deploy preview

mbrookes:
however it doesn't build locally, so will be amazed if it does on Netlify


oliviertassinari:
I was about to close this PR as the Hackathon is over, but I see we keep pushing commits 😁

mbrookes:
> I was about to close this PR as the Hackathon is over, but I see we keep pushing commits 😁

They seem mutually exclusive - we can push commits even if you close it, you can close it even if we keep pushing commits 😜

PR #2712, opened 2021-09-27
[DataGrid] Add Persian (faIR) locale

devlifeX:


flaviendelangle:
Hey,

Could you please add the core translations ? They seems to be available in persian
You have an equivalent here to see how it works  https://github.com/mui-org/material-ui-x/blob/next/packages/grid/_modules_/grid/locales/frFR.ts

devlifeX:
> Hey,
> Thanks for your contribution !
> 
> Could you please add the core translations ? They seems to be available in persian
> You have an equivalent here to see how it works https://github.com/mui-org/material-ui-x/blob/next/packages/grid/_modules_/grid/locales/frFR.ts

Hey, I'm glad to contribute!
Core translations added.

flaviendelangle:
Nice
Last thing, you can add the locale in this [file](https://github.com/mui-org/material-ui-x/blob/next/docs/src/pages/components/data-grid/localization/localization.md) so it appears in the documentation

devlifeX:
> Nice Last thing, you can add the locale in this [file](https://github.com/mui-org/material-ui-x/blob/next/docs/src/pages/components/data-grid/localization/localization.md) so it appears in the documentation

Done.

flaviendelangle (approved):

PR #2715, opened 2021-09-27
[DataGrid] Fix `undefined` in filter panel

### LINKED ISSUE #2687: [DataGrid] Filter panel shows "undefined" when is open for the first time
### Duplicates

### Latest version

### Current behavior 😯

image

### Expected behavior 🤔

The filter value field should start as empty.

### Steps to reproduce 🕹

Steps:

2. Add a filter

### Context 🔦

Regression from #2581.

The OR operator should be inside the cast.

https://github.com/mui-org/material-ui-x/blob/609248ea030cf6ca657bfae92f23bea3ba868c2b/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx#L81

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

DanailH:
Fixes #2687


m4theushw (commented):
Could you add a test? We should have caught this if we had one.

m4theushw (approved):
I pushed one more commit to improve the assertion. We could go even further dropping the `state` prop, but there're other tests also using it.

flaviendelangle:
Replacing the `state` prop with an `initialState` that only contains the same format as a future `exportState` is something I'd like to do soon.

But I will wait for the sync state PR to avoid collisions.

PR #2722, opened 2021-09-28
[core] Remove useless state update in useGridColumnMenu

flaviendelangle:


m4theushw (approved):
I had noticed this in https://github.com/mui-org/material-ui-x/pull/2673/files#diff-d4eaec74be51801926e1aa6c726a9e9db60bd5c9dc170724673002aeb125d39eR35 too.

Issue #211, opened 2020-08-22
[DataGrid] Implement Master detail/panel


oliviertassinari:
### Docs

https://mui.com/components/data-grid/group-pivot/#master-detail

### Benchmark

- https://material-table.com/#/docs/features/detail-panel
- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/detail-row/
- https://shlomiassaf.github.io/ngrid/features/built-in-plugins/detail-row
- https://www.ag-grid.com/javascript-grid-master-detail/
- https://ej2.syncfusion.com/react/demos/#/material/grid/master-detail
- https://www.telerik.com/kendo-react-ui/components/grid/advanced-features/hierarchy/

oliviertassinari:
@ak-mid In theory, this will be inside the Pro plan. Virtualization doesn't make it a simple problem to solve.

ak-mid:
@oliviertassinari Thanks for clarification! I already thought that this won't be an easy task.

wmhartl:
Another example in case it helps: https://autodesk.github.io/react-base-table/examples/detail-view

tlmader:
Hello! We are interested in using `<XGrid>` with the Pro plan for our project, and we require master/detail panel. I don't currently see this in the [Material-UI X public roadmap](https://github.com/mui-org/material-ui-x/projects/1). Do you have any ETAs regarding this feature?

oliviertassinari:
@tlmader We will update the public roadmap in the coming 7 days. We reach the end of Q1, time to reprioritize :).

tlmader:
@oliviertassinari Thank you for the reply!

davidhernon:
@oliviertassinari Hello, we are in the same boat as @tlmader. I just checked the roadmap (seeing it was updated 2 days ago) but I'm not seeing anything about this feature! Are there any plans for this feature at this time? Thank you!

oliviertassinari:
@davidhernon Hard to tell. We currently aim to improve the quality of the current feature implemented, so existing users get a great experience with. We know that we want to fix #417 and #438. Once done, this should be trivial to handle.

davidhernon:
@oliviertassinari Thanks for the info!

arheinjohnson:
Same as @tlmader and @davidhernon this is a highly valuable feature for us

allidoisace:
Hi there, what would be the difference between this and the Tree Data feature?

wmhartl:
[...]

PR #2725, opened 2021-09-28
[DataGridPro] Allow to automatically group the rows in a Tree Data

### LINKED ISSUE #210: [DataGridPro] Implement Tree data

flaviendelangle:
Closes #210 (I will create new issues for the following features in the Tree Data)

[Preview doc](https://deploy-preview-2725--material-ui-x.netlify.app/components/data-grid/group-pivot/#tree-data)

## Merging plan

1. Merge #2782  

2. Extract independent features in stand-alone PRs
- Columns pre-processing (and apply it for selection) #2839 
- Row grouping pre-processing (but do not apply it) #2840 
- Filtering rework #2870 
- CSV export cleaning #2941

3. Extract the selector breaking changes #2942

3. Extract the list => tree migration in `useGridRows`, `useGridFilter`, `useGridSorting` and rendering #2996 

4. Extract the data generator updates #2978

5. Merge the `useGridTreeData` and directly related features + its documentation

1-3 should be done before the major release because 3. is a breaking change.


packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
```diff
@@ -0,0 +1,149 @@
...
+  const groupingColDef = React.useMemo<GridColDef>(() => {
+    const propGroupingColDef = props.groupingColDef;
+
+    const baseColDef: GridColDef = {
+      ...GRID_TREE_DATA_GROUP_COL_DEF,
+      headerName: apiRef.current.getLocaleText('treeDataGroupingHeaderName'),
+    };
+    let colDefOverride: Partial<GridColDef>;
+
+    if (isFunction(propGroupingColDef)) {
```
m4theushw, line 43: The idea is to remove these one-line helpers.

```suggestion
    if (typeof propGroupingColDef === 'function') {
```

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
```diff
@@ -0,0 +1,149 @@
...
+        ...params,
+        defaultGroupingExpansionDepth: props.defaultGroupingExpansionDepth,
+      });
+    };
+
+    return apiRef.current.unstable_registerRowGroupsBuilder('treeData', groupRows);
+  }, [apiRef, props.getTreeDataPath, props.treeData, props.defaultGroupingExpansionDepth]);
+
+  useFirstRender(() => {
+    updateColumnsPreProcessing();
```
m4theushw, line 105: I created a hook to abstract this logic of registering pre-processors. You can `git checkout` from https://github.com/mui-org/material-ui-x/pull/2946/files#diff-333cd6c7803e1d6c4ea1f4af0205935a3e6768856016f078abe513b6e19399d7.

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
```diff
@@ -0,0 +1,149 @@
...
+    };
+  }, [apiRef, props.groupingColDef]);
+
+  const updateColumnsPreProcessing = React.useCallback(() => {
+    const addGroupingColumn = (columns: GridColumns) => {
+      if (!props.treeData) {
+        return columns;
+      }
+
+      const index = columns[0].type === 'checkboxSelection' ? 1 : 0;
```
m4theushw, line 65: Would checking for the `field` be more reliable?

```suggestion
      const index = columns[0].field === '__check__' ? 1 : 0;
```

docs/src/pages/components/data-grid/group-pivot/SetRowExpansionTreeData.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const rowId = apiRef.current.getRowIdFromRowIndex(0);
+    apiRef.current.unstable_setRowExpansion(
+      rowId,
+      !apiRef.current.unstable_getRowNode(rowId)?.expanded,
+    );
+  };
+
+  return (
+    <Stack style={{ width: '100%' }} alignItems="flex-start" spacing={2}>
+      <Button onClick={toggleFirstRow}>Toggle 1st row expansion</Button>
```
m4theushw, line 129: It does nothing when I click on it.

m4theushw (commented):
I didn't go too deep. I want to do a better review this week yet. 

Something is wrong here:

TJGdP6qtq2

docs/src/pages/components/data-grid/group-pivot/SetRowExpansionTreeData.tsx
flaviendelangle: I left the action on the 1st line, but `Sarah` have not children... I'm moving it to the 2nd one :+1:

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
flaviendelangle: I started checking the type instead of the field, because on  the row grouping by column you can have several grouping columns and therefore the field is not static, it becomes something like `__row_group_by_column_${originalFIeld}`
So I am checking the type as it would be more reliable in this scenario.
But for the checkbox column, I suppose both are equivalent no ?

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
flaviendelangle: I copy pasted the file :+1:

packages/grid/_modules_/grid/components/cell/GridTreeDataGroupingCell.tsx
```diff
@@ -0,0 +1,191 @@
...
+    root: ['treeDataGroupingCell'],
+    toggle: ['treeDataGroupingCellToggle'],
+  };
+
+  return composeClasses(slots, getDataGridUtilityClass, classes);
+};
+
+const GridTreeDataGroupingCellRoot = styled(Box, {
+  name: 'MuiDataGrid',
+  slot: 'GridTreeDataGroupingCell',
```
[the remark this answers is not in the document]
flaviendelangle: Done :+1:

flaviendelangle:
https://github.com/mui-org/material-ui-x/pull/2725#pullrequestreview-803227586

I was not hiding nodes when their ancestors further away than their parents were collapsed.
This should be fixed

https://github.com/mui-org/material-ui-x/pull/2725/commits/f0f14c56c0518df595f47a0f851096b848fa2307

packages/grid/_modules_/grid/utils/rowTreeUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+        nodeNameConfig = { id: nodeId, children: {} };
+        nodeNameToIdSubTree[nodeName] = nodeNameConfig;
+      } else {
+        nodeId = nodeNameConfig.id;
+      }
+      nodeNameToIdSubTree = nodeNameConfig.children;
+
+      if (depth < row.path.length - 1) {
+        let node = tempTree[nodeId] ?? null;
+        if (!node) {
```
m4theushw, line 63: Could you comment the code of this function? I spent a good time to understand that here it's filling the missing values.

Example: https://github.com/dowjones/react-dropdown-tree-select/blob/26371757d8f861f534c4b4f23167d16730c091ac/src/tree-manager/flatten-tree.js#L9

We never did that but this helper function could have unit tests to make easier to see the input and output.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -172,7 +172,7 @@ export const useGridFilter = (
         const filterTreeNode = (
```
m4theushw, line 172: The filtering and sorting logic for the tree data could be moved to a pre-processor to avoid shipping them as MIT.

In https://github.com/mui-org/material-ui-x/blob/2409df34aa924b4dd0d2ef80d59f2812974a1171/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts#L231

Would be smarter to exit the loop as soon a `node.depth > 0` is found? The tree nodes are ordered from the lowest depth to the highest.

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
m4theushw: For now both appear to yield the same result so keep as it is. However, in the future, checking the `field` seems to be better since it's unique while the type can repeat.

packages/grid/_modules_/grid/components/cell/GridTreeDataGroupingCell.tsx
```diff
@@ -0,0 +1,170 @@
...
+    }
+    if (isNavigationKey(event.key) && !event.shiftKey) {
+      apiRef.current.publishEvent(GridEvents.cellNavigationKeyDown, props, event);
+    }
+  };
+
+  const handleClick = (event) => {
+    apiRef.current.unstable_setRowExpansion(id, !value.expanded);
+    apiRef.current.setCellFocus(id, field);
+    event.stopPropagation();
```
m4theushw, line 58: Stopping the propagation causes side effects. See https://github.com/mui-org/material-ui-x/issues/3158#issuecomment-966644362

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
flaviendelangle: I think we should remain vigilant with pre-processor. We can have the temptation of answering any custom code with them, but they can make the global understanding of the codebase very complex.

For these two specific algorithm, I think it is better to start by in-lining them in `useGridFilter` and `useGridSorting` and then improve it (with pre-processor or another solution) when we want to use them in the row grouping feature.


> The tree nodes are ordered from the lowest depth to the highest.

Not necessarily.
Your auto generated nodes can be added after nodes of higher depth.

packages/grid/_modules_/grid/utils/rowTreeUtils.ts
flaviendelangle: I agree that this kind of function would be worth testing.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
m4theushw: > I think we should remain vigilant with pre-processor. We can have the temptation of answering any custom code with them, but they can make the global understanding of the codebase very complex.

I have a different vision. I found easier to understand when keeping everything related to a feature inside the same hook. The filtering hook is already quite long with 440 LoC in total, where 65 LoC are for the tree-data.

> For these two specific algorithm, I think it is better to start by in-lining them in useGridFilter and useGridSorting and then improve it (with pre-processor or another solution) when we want to use them in the row grouping feature.

Is it not just to copy and paste into a pre-processor? I think if we move to a pre-processor there's no other way then the pre-processor completely overriding the filtering logic.

flaviendelangle: > I found easier to understand when keeping everything related to a feature inside the same hook

We agree on that point
My problem is more about when the behavior is triggered and how many time.

> where 65 LoC are for the tree-data.

The Tree Data or the Row Grouping, or the Aggregation (partially).
I did not put it in the Tree Data hook because it can be applied to any feature using a tree.
Same reason why it is `useGridRows` which store the tree structure.
My PR of the row grouping is very simple for the sorting / filtering / rows management because all the logic is compatible.

And when we will launch the row grouping for instance, the pre-processing change will potentially trigger several times if we change a shared prop (`props.groupingColDef` for instance) even if one of them is actually not enabled (`props.treeData = false`). And in that scenario, we will have to make sure that the filtering / sorting does not run several times.

A solution could be to introduce a `useTreeRows` which registers pre-processors only once is one of the tree features is enabled. I don't have a clear vision on that part yet.

flaviendelangle: I think the pre-processing is a powerfull tool, but I also think we need more distance.
Things like "re-run all the pre-processors of a group whenever one processor said he has changed" is probably not suitable for extensive use.

packages/grid/_modules_/grid/utils/rowTreeUtils.ts
flaviendelangle: I added a few tests

docs/src/pages/components/data-grid/group-pivot/DisableChildrenFilteringTreeData.tsx
```diff
@@ -0,0 +1,160 @@
...
+    field: 'recruitmentDate',
+    headerName: 'Recruitment Date',
+    type: 'date',
+    width: 150,
+  },
+];
+
+const getTreeDataPath = (row) => row.hierarchy;
+
+export default function DisableChildrenSortingTreeData() {
```
m4theushw, line 119:
```suggestion
export default function DisableChildrenFilteringTreeData() {
```

docs/src/pages/components/data-grid/group-pivot/DisableChildrenFilteringTreeData.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [disableChildrenFiltering, setDisableChildrenFiltering] =
+    React.useState(true);
+  const [filterModel, setFilterModel] = React.useState<GridFilterModel>({
+    linkOperator: GridLinkOperator.Or,
+    items: [
+      {
+        columnField: 'recruitmentDate',
```
m4theushw, line 126: ID missing.

docs/src/pages/components/data-grid/group-pivot/DisableChildrenFilteringTreeData.tsx
```diff
@@ -0,0 +1,160 @@
...
+    <Stack style={{ width: '100%' }} alignItems="flex-start" spacing={2}>
+      <FormGroup>
+        <FormControlLabel
+          control={
+            <Checkbox
+              checked={disableChildrenFiltering}
+              onChange={(event) => setDisableChildrenFiltering(event.target.checked)}
+            />
+          }
+          label="Enable `disableChildrenFiltering`"
```
m4theushw, line 143: This formatting doesn't work in the demos.

```suggestion
          label={
            <Typography component="span">
              Enable <code>disableChildrenFiltering</code>
            </Typography>
          }
```

docs/src/pages/components/data-grid/group-pivot/SetRowExpansionTreeData.tsx
```diff
@@ -0,0 +1,141 @@
...
+  },
+];
+
+const getTreeDataPath = (row) => row.hierarchy;
+
+export default function SetRowExpansionTreeData() {
+  const apiRef = useGridApiRef();
+
+  const toggleFirstRow = () => {
+    const rowId = apiRef.current.getRowIdFromRowIndex(1);
```
m4theushw, line 120: Once the user filters, this method will return the wrong id.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,113 @@ title: Data Grid - Group & Pivot
...
+{{"demo": "pages/components/data-grid/group-pivot/CustomGroupingColumnTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Group expansion
+
+Use the `defaultGroupingExpansionDepth` prop to expand all the groups up to a given depth when loading the data.
+If you want to expand the whole tree, set `defaultGroupingExpansionDepth = -1`
+
+{{"demo": "pages/components/data-grid/group-pivot/DefaultGroupingExpansionDepthTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+Use the `unstable_setRowExpansion` method on `apiRef` to programmatically set the expansion of a row.
```
m4theushw, line 73: Is not weird to tell users to use an unstable method? About the naming, it may cause confusion when #211 is implemented.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -288,6 +290,16 @@ export const GridRootStyles = styled('div', {
...
     },
+    [`& .${gridClasses.treeDataGroupingCell}`]: {
+      display: 'flex',
+      alignItems: 'center',
+      width: '100%',
+    },
+    [`& .${gridClasses.treeDataGroupingCellToggle}`]: {
+      flex: '0 0 28px',
+      alignSelf: 'stretch',
+      marginRight: 16,
```
m4theushw, line 301:
```suggestion
      marginRight: theme.spacing(2),
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,113 @@ title: Data Grid - Group & Pivot
...
 
-## 🚧 Tree data [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
+## Tree Data [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
 
-> ⚠️ This feature isn't implemented yet. It's coming.
->
-> 👍 Upvote [issue #210](https://github.com/mui-org/material-ui-x/issues/210) if you want to see it land faster.
+Tree Data allows to display data with parent / child relationships.
+
+### General behavior
```
m4theushw, line 13: I don't think we need this title.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,113 @@ title: Data Grid - Group & Pivot
...
+
+By default, the filtering is applied to every depth of the tree.
+You can limit the filtering to the top level rows with the `disableChildrenFiltering` prop.
+
+{{"demo": "pages/components/data-grid/group-pivot/DisableChildrenFilteringTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Sorting
+
+By default, the sorting is applied to every depth of the tree.
+You can limit the filtering to the top level rows with the `disableChildrenSorting` prop.
```
m4theushw, line 98:
```suggestion
You can limit the sorting to the top level rows with the `disableChildrenSorting` prop.
```

m4theushw (commented):
Clicking on Select All, with all children closed, only selects the top level rows. However, if one parent is open, its children are selected too. Should we add a section about selection in the docs mentioning this?

Related to https://ag-grid.com/javascript-data-grid/tree-data/#selection

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: Yes it's probably not great to prefix it and document it.
As it is a very basic method that will probably never change, I suppose we can make it public.
I had in mind we would use the same expansion boolean and method for boths as discussed on Monday. But as we won't, I can rename it`setRowChildrenExpansion`. In which case I would rename the `node.expanded` into `node.childrenExpanded`

docs/src/pages/components/data-grid/group-pivot/SetRowExpansionTreeData.tsx
flaviendelangle: Good catch
I really don't like this method :smile:

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: Done

flaviendelangle:
https://github.com/mui-org/material-ui-x/pull/2725#pullrequestreview-809408192

Good question...  For you what should be the intended behavior here ?

packages/grid/_modules_/grid/components/cell/GridTreeDataGroupingCell.tsx
```diff
@@ -0,0 +1,170 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { unstable_composeClasses as composeClasses } from '@mui/core';
```
m4theushw, line 3: `@mui/core` is not a dependency.

```suggestion
import { unstable_composeClasses as composeClasses } from '@mui/material';
```

m4theushw:
I've created a demo to try to use the tree data to manage product categories: https://codesandbox.io/s/basictreedata-material-demo-forked-vtxd8?file=/demo.tsx

It crashes if the row with ID = 5 is uncommented. Also, it shows "Photography (1)" when there's actually 3 children. Is this a bug or I'm doing something wrong? If I'm wrong, it shouldn't crash because rows may not have a parent.

m4theushw:
> Good question... For you what should be the intended behavior here ?

Selecting one row shouldn't select its children in the tree data. On the Select All I think we could have a prop to decide to include the hidden nodes or not. We can wait for the community feedback, but this will be certainly a question.  

For the row group, I think selecting the parent should instead also include the children since the parent row is not a real row.

flaviendelangle:
@m4theushw `Note that all paths must contain at least one element.`
In the Tree Data, the last element of the path is the grouping value of the row so a path of length 1 has no parent.
It is the way AG-Grid works and for me it make sense.

So same thing for your other issue, all your rows are missing one path step.

The user can of course concatenate in `getTreeDataPath` if there data is structured like yours with the structure of the parents.

m4theushw:
> The user can of course concatenate in getTreeDataPath if there data is structured like yours with only the hierarchy starting with the parent.

If I use this trick, it doesn't use "Photography" as the parent for all children. Instead, there're two top level rows: "Photography" and "Digital Cameras". I'm trying to simulate how the tree data would be used with data coming from a database table. Usually, only the immediate parent is saved, not the entire path.

```js
{
  breadcrumb: ["Photography"], // Group is "Photography/Digital Cameras"
  name: "Digital Cameras",
  id: 0
},
{
  breadcrumb: ["Digital Cameras"], // Group is "Digital Cameras/Mirrorless Cameras"
  name: "Mirrorless Cameras",
  id: 2
},
```

---

Trying to make the grouping colDef editable doesn't work. It shows "[object Object]". We should not allow the following:

```ts
const groupingColDef: DataGridProProps["groupingColDef"] = {
  headerName: "Parent",
  editable: true
};
```

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
m4theushw: If there are no columns it crashes here. I could decide to have only the grouping column.

flaviendelangle:
In the tree data the path is not just the parent it's all the ancestors.
I updated your Codesandbox to make it work without touching the data to show that it can be done when your path only goes from the top level ancestor to the parent and does not include the current row.
But the use case is to have a path on each row that goes from to top level row to the current row.
On the example below, the `Mirrorless Cameras` must explicitely put `Photography`.
If we only register the parent, it's is a very different use case.

```ts
{
  breadcrumb: ["Photography"], // Group is "Photography/Digital Cameras"
  name: "Digital Cameras",
  id: 0
},
{
  breadcrumb: ["Photography", "Digital Cameras"], // Group is "Photography/Digital Cameras/Mirrorless Cameras"
  name: "Mirrorless Cameras",
  id: 2
},
```

___

For the forced value of the grouping col def, I agree that we should remove them from the typing and spread them after to avoid override.

flaviendelangle:
Fix for the `editable` :+1: 
I try to keep it generic to be able to apply the same override logic to the checkbox column or the row grouping.

m4theushw:
There's still this one: https://github.com/mui-org/material-ui-x/pull/2725/files#r752488746

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
flaviendelangle: Fixed

flaviendelangle:
@m4theushw Done :+1:

sanjinbecirovic:
Is there an ETA when this feature will be released?

flaviendelangle:
@sanjinbecirovic we don't have a fixed date but depending on the feedback it could be for the 02/11.
We prefer to review thoroughly this one to avoid bad surprises and breaking changes to fix bugs / unintended behaviors :+1:

alexfauquette:
Small design detail, the arrow icon should be in the other direction when children are open. See the "Thomas" row in the DataGrid and the "Component API" in the doc menu

image
image

m4theushw:
> Small design detail, the arrow icon should be in the other direction when children are open. See the "Thomas" row in the DataGrid and the "Component API" in the doc menu

Is not the sidebar that is wrong? In most of the demos in https://mui.com/components/accordion/#customization, the arrow is pointing up when the item is open. Only https://mui.com/components/accordion/#customization reproduces the same pattern as the sidebar. cc @danilo-leal

m4theushw (commented):
The data generator is generating duplicated paths. This causes unwanted behaviors if users open the demo in CodeSandbox and change the number of rows to something big, like 2k rows. See this example: https://codesandbox.io/s/treedatafullexample-material-demo-forked-hkm7q?file=/demo.tsx

image

I think we could also send an error to the console if the paths are not unique, similar to what happens with the `key` prop. It took me some time to figure out they weren't unique.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,111 @@ title: Data Grid - Group & Pivot
...
+By default, the sorting is applied to every depth of the tree.
+You can limit the sorting to the top level rows with the `disableChildrenSorting` prop.
+
+{{"demo": "pages/components/data-grid/group-pivot/DisableChildrenSortingTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+**Note**: If you are using `sortingMode="server"`, you need to always set the children of a row after this row.
+For instance:
+
+```ts
+// The row A.A is just after its parent
```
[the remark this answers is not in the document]
flaviendelangle: Fixed, it's much more readable that way

flaviendelangle:
@m4theushw concerning the non-uniq paths, I indeed only applied the logic to the non-seeded generator, so below 1000 lines.
I will add an error when using the generator with tree data and more than 1000 lines for now :+1: 

For the unicity test, same thing I think it can be done in a folow up PR. And probably in dev mode only to avoid overhead

docs/src/pages/components/data-grid/group-pivot/DisableChildrenSortingTreeData.tsx
```diff
@@ -0,0 +1,151 @@
...
+    <Stack style={{ width: '100%' }} alignItems="flex-start" spacing={2}>
+      <FormGroup>
+        <FormControlLabel
+          control={
+            <Checkbox
+              checked={disableChildrenSorting}
+              onChange={(event) => setDisableChildrenSorting(event.target.checked)}
+            />
+          }
+          label="Enable `disableChildrenSorting`"
```
m4theushw, line 134: The formatting is different from https://deploy-preview-2725--material-ui-x.netlify.app/components/data-grid/group-pivot/#filtering.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,111 @@ title: Data Grid - Group & Pivot
...
+By default, the sorting is applied to every depth of the tree.
+You can limit the sorting to the top level rows with the `disableChildrenSorting` prop.
+
+{{"demo": "pages/components/data-grid/group-pivot/DisableChildrenSortingTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+> If you are using `sortingMode="server"`, you need to always set the children of a row after this row.
+> For instance:
+
+```ts
+// The row A.A is immediately after its parent
```
m4theushw, line 104: The code should be inside the warning.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,111 @@ title: Data Grid - Group & Pivot
...
+const validRows = [{ path: ['A'] }, { path: ['A', 'A'] }, { path: ['B'] }];
+
+// The row A.A is not immediately after its parent
+const invalidRows = [{ path: ['A'] }, { path: ['B'] }, { path: ['A', 'A'] }];
```
m4theushw, line 108: Add emojis to make it clear what is the wrong pattern. Two examples: https://mui.com/guides/minimizing-bundle-size/#option-1 and https://mui.com/customization/how-to-customize/#main-content

m4theushw (approved):
Great work! I don't have anything to add. 💯

My only concern is that the code for sorting and filtering tree data nodes shouldn't be shipped as MIT, however, this can be addressed later. 

Merging with `master` should fix the screenshot below:

image

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,13 +6,111 @@ title: Data Grid - Group & Pivot
...
+{{"demo": "pages/components/data-grid/group-pivot/DisableChildrenFilteringTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Sorting
+
+By default, the sorting is applied to every depth of the tree.
+You can limit the sorting to the top level rows with the `disableChildrenSorting` prop.
+
+{{"demo": "pages/components/data-grid/group-pivot/DisableChildrenSortingTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+
+> If you are using `sortingMode="server"`, you need to always set the children of a row after this row.
```
m4theushw, line 100: A small detail. It's not clear what is "this row".

```suggestion
> If you are using `sortingMode="server"`, you need to always put the children of a row after its parent.
```

sanjinbecirovic:
> @sanjinbecirovic we don't have a fixed date but depending on the feedback it could be for the 02/11. We prefer to review thoroughly this one to avoid bad surprises and breaking changes to fix bugs / unintended behaviors 👍

Thank you for the reply, 02/11 is a release or? I've seen the code has been merged? but not released?

flaviendelangle:
02/12 not 02/11 my bad.
Yes it should be release on the 02/12, or in the days just after if the release takes more time to prepare.
In any  case, it will be in the next release and we have a release planned tomorrow.

PR #2733, opened 2021-09-28
v4.0.1

m4theushw:


CHANGELOG.md
```diff
@@ -3,13 +3,43 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 4.0.1
+
+_Sep 29, 2021_
+
+This version includes bugfixes and features already released in v5.x.
```
oliviertassinari, line 10: Would it help to frame the message as, v5 is active, and we have cherry-picked what we could from it to the legacy v4 version?

```suggestion
This version backports bug fixes and features from the v5.x active release line.
```

m4theushw: Should we make clear that might not have other v4 releases due to the complexity of cherry-picking?

```suggestion
This version backports bug fixes and some of the features from the v5.x active release line.
As the development of v5 evolves, newer fixes might not be cherry-picked to the legacy version.
To have access to the latest features it is encouraged to upgrade to MUI X v5 and MUI Core v5.
```

---

I created a [CodeSandbox](https://codesandbox.io/s/datagriddemo-material-demo-forked-drz2n?file=/src/demo.js) using MUI Core v4 and v5 and it seems to work fine. The only drawback is having to install both versions, but the benefit of getting new features pays off, I think. We could add a section to the docs explaining how to configure the prefixes.

oliviertassinari: Yeah, agree, a second about how to use MUI Core v4 would help. I have simplified/improved the codesandbox: https://codesandbox.io/s/datagriddemo-material-demo-forked-8myv0?file=/src/demo.js.

- No more emotion cache provider
- Use createGenerateClassName from the v4 package.

flaviendelangle:
By the way, do we want to cherry-pick these changelog in the `next` `CHANGELOG.md` file ?

DanailH:
> By the way, do we want to cherry-pick this changelog in the `next` `CHANGELOG.md` file?

I don't think that is needed. At some point, the 2 branches will diverge significantly (when you might need to fix something specifically for v4 for example). I think that changelog should only include v4 related releases (but then again the other way also makes sense).

CHANGELOG.md
```diff
@@ -3,13 +3,43 @@
...
+
+Big thanks to the 3 contributors who made this release possible. Here are some highlights ✨:
+
+- 👁 Allow to disable virtualization with the `disableVirtualization` prop (#2728) @m4theushw
```
DanailH, line 14: Do we need this feature in v4?

oliviertassinari: I would personally encourage we make this one the last of the v4 release line. Unless there is a major issue: Chrome regression, Node.js regression, webpack regression, etc. Basically, use the policy we had for Core v4 when we were working on Core v5. There are 1-2 years of inertia with open source. v4 keep growing during all the time we were working on v5, while we were almost not making any changes.

oliviertassinari:
> By the way, do we want to cherry-pick these changelog in the next CHANGELOG.md file?

When we browse the changelog of React do we have to look at different branches to see the whole release history?  https://github.com/facebook/react/bblog /main/CHANGELOG.md. As far as I know, no, this would be confusing. I do think that we should have the changelog in HEAD to include the whole history. So yes 👍 for a cherry-pick.

CHANGELOG.md
flaviendelangle: If MUI X v5 + MUI Core v4 is viable (= if users do not find major issues that we would have missed, like weird style collision) I am 100% in favor of this.

m4theushw: > Do we need this feature in v4?

An user asked for it and it was simple to cherry-pick: https://github.com/mui-org/material-ui-x/issues/1902#issuecomment-912017532

DanailH (approved):

PR #2737, opened 2021-09-29
[docs] Add redirection from doc home page to DataGrid home page

flaviendelangle:


flaviendelangle:
@oliviertassinari do you confirm that the Netlify redirection will not break the doc deployed on https://mui.com/ ?

.gitignore
```diff
@@ -11,6 +11,7 @@ __diff_output__
 /coverage
 /docs/.next
 /docs/export
+/docs/pages/playground.js
```
oliviertassinari, line 14: I don't think that it would woof .gitignore prevent untracked files from being added. This file is already tracked by git and we want it to be tracked so we can quickly get started. 

```suggestion
```

docs/public/_redirects
```diff
@@ -10,5 +10,7 @@
 
 # Redirections
 
+/ /components/data-grid
```
oliviertassinari, line 13: AFAIK this would only impact the Netlify preview, it should work.

```suggestion
/ /components/data-grid/
```

docs/next.config.js
```diff
@@ -154,4 +154,10 @@ module.exports = {
       { source: '/api/:rest*', destination: '/api-docs/:rest*' },
     ];
   },
+  redirects:
+    process.env.NODE_ENV !== 'production'
+      ? async () => {
+          return [{ source: `/:lang(${LANGUAGES.join('|')})?/:rest*`, destination: '/:rest*' }];
```
oliviertassinari, line 160: The content here seems off topic, did you intent to redirect to the data grid component?

docs/public/_redirects
oliviertassinari: Once we have a second component, we could point to the MUI X introduction page

flaviendelangle: Yes the redirection page is the best one as of today but at one point we should have a X home page

docs/next.config.js
flaviendelangle: :roll_eyes: I commited one of my tests copied from the core...
It should be the same redirection as in the Netlify config but for the local env

docs/pages/index.js
```diff
@@ -1,5 +0,0 @@
-import * as React from 'react';
-
-export default function Home() {
```
oliviertassinari, line 3: I guess we should rebase? It's already gone on HEAD.

PR #2753, opened 2021-10-02
[DataGrid] Add row and cell slots

### LINKED ISSUE #2138: [DataGrid / X-Grid] Add possibility to pass in component for row and cell

m4theushw:
Closes #2138 

In the new virtualization, I found that reducing the number of attributes and listeners added to each row / cell can improve drastically the performance. In this context, I plan to remove a bunch of events that are emitted but which are not used internally, e.g. `cellEnter`. This PR will allow users that depend on these events to re-add them as needed.


docs/src/pages/components/data-grid/components/components.md
```diff
@@ -11,7 +11,8 @@ title: Data Grid - Components
 As part of the customization API, the grid allows you to override internal components with the `components` prop.
 The prop accepts an object of type [`GridSlotsComponent`](/api/data-grid/data-grid/#slots).
 
-If you wish to pass additional props in a component slot, you can do it using the `componentsProps` prop. This prop is of type `GridSlotsComponentsProps`.
+If you wish to pass additional props in a component slot, you can do it using the `componentsProps` prop.
+This prop is of type `GridSlotsComponentsProps`.
```
flaviendelangle, line 15: This interface would be very interesting to add to the API documentation in the future
The comments in it are important to know each key impact on the UI.

packages/grid/data-grid/src/tests/components.DataGrid.test.tsx
```diff
@@ -49,6 +50,70 @@ describe('<DataGrid /> - Components', () => {
...
+    });
+
+    it('should render the row with the component given in components.Row', () => {
+      render(
+        <div style={{ width: 300, height: 500 }}>
+          <DataGrid
+            {...baselineProps}
+            hideFooter
+            disableVirtualization
+            components={{ Row: ({ rowIndex }) => <span role="row" data-rowindex={rowIndex} /> }}
```
flaviendelangle, line 109: Do we agree that if somebody overrides `row` or `cell`, he must to copy / paste the code in`GridRow`, otherwise the basic event listeners won't be created for instance.

This mean that every modification we do on this component is technically a breaking change.
I think it makes the API very fragile. Someone overriding this kind of component risks to have broken behavior a soon as we add new features that requires to add code to those components.

What is our approach on this subject ?

- the `components` props is for advanced application and those who use it must be aware that this part may not respect semver ?
- we respect semver on those components and will never add any required code to them between major (don't think it's viable)
- 3rd option ?

m4theushw: > Do we agree that if somebody overrides row or cell, he must to copy / paste the code inGridRow, otherwise the basic event listeners won't be created for instance.

Yes, that's the purpose of the prop. It's for extreme use cases, when the row or cell needs to be completely overridden. For simple cases, using `componentsProps` is sufficient.

> This mean that every modification we do on this component is technically a breaking change.

Yes, but the row and cell are quite stable. We didn't made big changes on both in the past.

> What is our approach on this subject ?

Option 2 maybe. These components are dumb and don't have any complex functionality, everything comes via props. I don't see a feature which will require breaking changes on them, besides the virtualization.

m4theushw: > We have to modify these components if we want to add a class to the DOM, as we did when we introducted the edition mode for instance.

In this case, adding a class is not a breaking change.

> But if you think that these components are stable enough, I just think that it would be nice to add a clear warning here https://mui.com/components/data-grid/components/#components, to avoid complains due to complex migrations (even if they only come in the next major).

That can be done latter once we decide what to do after the stable release. For now, any breaking change to the row or cell component should be mentioned.

flaviendelangle: Indded
Adding a `data-attr` that we use in the rest of the codebase like `data-id` would.


> That can be done latter once we decide what to do after the stable release. For now, any breaking change to the row or cell component should be mentioned.

OK :+1:

packages/grid/data-grid/src/tests/components.DataGrid.test.tsx
```diff
@@ -49,6 +50,70 @@ describe('<DataGrid /> - Components', () => {
...
+  describe('components', () => {
+    it('should render the cell with the component given in components.Cell', () => {
+      render(
+        <div style={{ width: 300, height: 500 }}>
+          <DataGrid
+            {...baselineProps}
+            hideFooter
+            disableVirtualization
+            components={{
+              Cell: ({ rowIndex, colIndex }) => (
```
DanailH, line 92: Maybe a relevant question but what is the difference between providing a cell like this and doing it via the `renderCell` in the `colDef`? What if you do both?

m4theushw: We can wait for the feedback from the community. In reality, I don't see an use case for having to override the cell or row (maybe to render `<tr>` and `<td>`), even on the demo I didn't have to do this. This PR is more for the `componentsProps` which will be used to re-add events I want to remove later.

flaviendelangle: OK so for you the idea is to theoratically handle the `props.components.row` and `props.components.cell` but to talk little about it in the doc because it is really an edge case ?

m4theushw: Yes, I added `components.Cell` for convention but users will rarely need to replace the cell.

PR #2784, opened 2021-10-05
[DataGrid] Remove `@mui/styles` dependency

### LINKED ISSUE #1234: Replace @mui/styles with @mui/system

DanailH:
Fixes #1234 

I'll keep the PR as a draft for now.

I would really appreciate the feedback on this first iteration, I'm not sure that I've done it as well as it can be. Once the initial feedback is there I'll iterate and open it as a normal PR.

What's not included in this PR is the migration of the docs, storybook, x-data-generation package. Not sure how important is the last one or we should postpone them.

One test is failing - no idea why - the structure seems the same as before.

### DIFF
--- a/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
+++ b/packages/grid/_modules_/grid/components/menu/GridMenu.tsx
@@ -4,9 +4,9 @@ import ClickAwayListener, { ClickAwayListenerProps } from '@mui/material/ClickAw
 import Grow from '@mui/material/Grow';
 import Paper from '@mui/material/Paper';
 import Popper, { PopperProps } from '@mui/material/Popper';
-import { Theme, createTheme } from '@mui/material/styles';
-import { makeStyles } from '@mui/styles';
+import { styled } from '@mui/material/styles';
 import { HTMLElementType } from '@mui/utils';
+import { gridClasses } from '../../gridClasses';
 
 type MenuPosition =
   | 'bottom-end'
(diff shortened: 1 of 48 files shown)


packages/grid/_modules_/grid/components/menu/GridMenu.tsx
```diff
@@ -54,7 +49,7 @@ const GridMenu = (props: GridMenuProps) => {
   const { open, target, onClickAway, children, position, ...other } = props;
   const prevTarget = React.useRef(target);
   const prevOpen = React.useRef(open);
-  const classes = useStyles();
+  // const classes = useStyles();
```
Rupert-com, line 52: Why?
It's not used...

packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx
```diff
@@ -16,28 +16,27 @@ import { GridPanelWrapper } from './GridPanelWrapper';
...
-      marginRight: 4,
-    },
-    dragIcon: {
-      justifyContent: 'flex-end',
-    },
-  },
-  { name: 'MuiDataGridColumnsPanel' }, // TODO rename to MuiGridColumnsPanel
-);
+const GridColumnsPanelRoot = styled('div', {
+  name: 'MuiGridColumnsPanel',
```
Rupert-com, line 20: `{ name: 'MuiDataGridColumnsPanel' }, // TODO rename to MuiGridColumnsPanel`
are dependencies renamed too?

DanailH: What do you mean by dependencies? This is to match our naming convetion.

DanailH:
> image
> 
> Is this diff wanted ?

This is the conscience of using the spacing from the theme. I replaced where I could. that is where the misalignment of a couple of pixels is coming from.

PR #2802, opened 2021-10-07
[DataGridPro] Stop drag event propagation

### LINKED ISSUE #2680: [DataGridPro] Column reorder does not work when `DataGridPro` is a child of react-dnd-html5-backend
### Duplicates

### Latest version

### Current behavior 😯

When the component tree, which contains the DataGrid, is wrapped with `react-dnd-html5-backend`, the column reorder functionality breaks.

### Expected behavior 🤔

Reordering columns should work, even when the DataGrid is wrapped with a third party dnd library like `react-dnd`.

### Steps to reproduce 🕹

Steps:

1. Go to https://codesandbox.io/s/kind-pine-tostw
2. Try to reorder the columns by Drag n Drop.

### Context 🔦

The problem is, that the HTML5Backend of `react-dnd` hijacks all out of context `draggable` elements: https://github.com/react-dnd/react-dnd/issues/3304. And since the grid leverages this API, its column reorder feature breaks when used with `react-dnd`.

One solution that seems to work is described here https://github.com/react-dnd/react-dnd/issues/2868:

It suggests to stop event propagation in `onDragStart`, `onDragEnd` and `onDrag` event listeners.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

22075

DanailH:
Fixes #2680

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx
+++ b/packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx
@@ -78,6 +78,9 @@ export const useGridColumnReorder = (
       }
 
       logger.debug(`Start dragging col ${params.field}`);
+      // Prevent drag events propagation.
+      // For more information check here https://github.com/mui-org/material-ui-x/issues/2680.
+      event.stopPropagation();
 
       dragColNode.current = event.currentTarget;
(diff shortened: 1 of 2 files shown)


m4theushw:
This solution didn't solve the problem: https://codesandbox.io/s/nice-ellis-di0vm?file=/package.json

PR #2810, opened 2021-10-08
[DataGrid] Remove props not used and redundant DOM attributes

m4theushw:
### Breaking changes

- [DataGrid] The following props were removed. If you depend on them, use `componentsProps.row` and `componentsProps.cell` to pass custom props to the row or cell.

  - onCellBlur
  - onCellOver
  - onCellOut
  - onCellEnter
  - onCellLeave
  - onRowOver
  - onRowOut
  - onRowEnter
  - onRowLeave
  
  For more information, check [this page](https://next--material-ui-x.netlify.app/components/data-grid/components/#row). Example:

  ```diff
  -<DataGrid onRowOver={handleRowOver} />;
  +<DataGrid
  +  componentsProps={{
  +    row: { onMouseOver: handleRowOver },
  +  }}
  +/>;
  ```
 
- [DataGrid] The `data-rowindex` and `data-rowselected` attributes were removed from the cell element. Equivalent attributes can be found in the row element.

- [DataGrid] The `data-editable` attribute was removed from the cell element. Use the `.MuiDataGrid-cell--editable` CSS class.

- [DataGrid] The `data-mode` attribute was removed from the cell element. Use the `.MuiDataGrid-cell--editing` CSS class.

---

The purpose of this PR is to remove a few props and attributes that are applied to the row and cell, but which are not used internally. We can improve the performance a bit by freeing React from having to update the DOM each time they change. Most of the props removed are related to very specific mouse events that I believe not many users relay onto. In cases they need, the slot prop can be used as explained in the breaking changes. The `React.useMemo` was removed because it's not useful, as each time the row renders it has a different ID so it needs to output a different value. Even though it's suppose to improve performance, there's a cost to check if the dependencies changed.

### DIFF
--- a/packages/grid/_modules_/grid/components/GridRow.tsx
+++ b/packages/grid/_modules_/grid/components/GridRow.tsx
@@ -19,6 +19,8 @@ export interface GridRowProps {
   selected: boolean;
   rowIndex: number;
   children: React.ReactNode;
+  onClick?: React.MouseEventHandler<HTMLDivElement>;
+  onDoubleClick?: React.MouseEventHandler<HTMLDivElement>;
 }
 
 type OwnerState = GridRowProps & {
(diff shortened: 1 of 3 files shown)


flaviendelangle:
A single example in the *Breaking change* section would be nice for the CHANGELOG

```diff
<DataGrid
-   onRowOver={handleRowOver}
+  componentsProps={{
+    row: onMouseOver: handleRowOver,
+  }}
/>
/>
```

PR #2862, opened 2021-10-14
[DataGrid] Don't select the row when clicking on an item in the `actions` column type

m4theushw:
Fixes https://github.com/mui-org/material-ui-x/issues/891#issuecomment-926429837

Before: https://codesandbox.io/s/columntypesgrid-material-demo-forked-6codp?file=/demo.tsx

After: https://codesandbox.io/s/columntypesgrid-material-demo-forked-nrrs5?file=/demo.tsx

Click on any item from the last column and check the console.

### DIFF
--- a/packages/grid/_modules_/grid/components/cell/GridActionsCell.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridActionsCell.tsx
@@ -26,7 +26,10 @@ const GridActionsCell = (props: GridActionsCellProps) => {
     throw new Error('MUI: Missing the `getActions` property in the `GridColDef`.');
   }
 
-  const showMenu = () => setOpen(true);
+  const showMenu = (event: React.MouseEvent) => {
+    event.stopPropagation();
+    setOpen(true);
+  };
 
   const hideMenu = () => setOpen(false);
 
--- a/packages/grid/_modules_/grid/components/cell/GridActionsCellItem.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridActionsCellItem.tsx
@@ -13,18 +13,26 @@ export type GridActionsCellItemProps = {
 );
 
 const GridActionsCellItem = (props: GridActionsCellItemProps) => {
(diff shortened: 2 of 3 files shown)


flaviendelangle:
I'm trying to do the same thing with the expand / collapse column of the Tree Data
And you seem to have the same limitation.
Try to open the action menu then close it by clicking outside
You are then in a state were can can't navigate with the keyboard.
The DataGrid state tells that the cell is focused but the focused DOM element is not the cell (it's the whole page in your case), to the `onKeyDown` of `GridCell` is not called when pressing the arrow keys.

NemanjaTck:
Data Grid Pro version 5.4.0 introduced this problem again. Both row click event is activated together with the showMenu. Please take a look into this. We had to lock version in our package.json to 5.2.1 to make it work.

m4theushw:
@NemanjaTck In #3492 we changed to not stop the propagation of click events. This change was to fix #3489. The problem that #2862 fixes is to not mark a row as selected when clicking in action. Despite this change, `onRowClick` will still be called, same for `onCellClick`. You can check the `event.currentTarget` provided to this callback and if the DOM element has `data-field="actions"` then do nothing. If this doesn't work I would suggest to open another issue so we can discuss better your use case.

https://github.com/mui-org/material-ui-x/blob/cbe570cad533397324d7589e6a7e584d54e5ed4d/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L338-L342

kodeschooler:
@NemanjaTck Hi, can you tell me if #3492 is in the latest version of mui? I'm having an issue with an action cell that launches a dialog with a text field on it. It's losing focus when you click into the text field. I found your comment in another thread. I notice in the code you referenced above in useGridSelection.ts (line 352) where it just returns if the column type is an action. I'm currently on v5.1.1. If I upgrade will this help? I see v5.5.1 is out, but all the ticket numbers seem to be 3100 or lower, and you indicate 3492 above. Thanks

PR #2874, opened 2021-10-14
[DataGrid] Add `is any of` filter operator

### LINKED ISSUE #2437: [DataGrid] Add operator to filter for rows that contain any of the specified values
hello, 

There is a lack of functionality for the xGrid, it would be necessary that when we apply filters on several columns we can also use several logical operators.

For example being able to do:

```
columnA === "valueA"   OR   columnA === "valueB"  AND   columnB === "valueX"
```

Have a nice day

alexfauquette:
Closes #2437

The idea is to add the "is one of" filtering condition.

For now, filtering has been imagined as containing a unique value. I'm modifying the pipeline to allow an array of values.

- [x] Column Type to support
	- [x] text
	- [x] number
	- [x] ~~date~~ (relevant?)
	- [x] ~~datetime-local~~ (relevant?)
	- [x] singleSelect
- [x] tests
- [x] doc

To test the `is any of`: 
https://deploy-preview-2874--material-ui-x.netlify.app/components/data-grid/filtering/#predefined-filters

The documentation about how to create custom filters operators with multiple values:
https://deploy-preview-2874--material-ui-x.netlify.app/components/data-grid/filtering/#multiple-values-operator


alexfauquette:
Because `Autocomplete` distinguishes `onChange` and `onInputChange`, the filtering is not applied while the user is typing. Filters are applied each time the value change (Chip added or removed)

Do you have an opinion on the expected behavior? Should the filter be applied or not when the user is typing?

flaviendelangle:
> Do you have an opinion on the expected behavior? Should the filter be applied or not when the user is typing?

I think it should not, for performances reason.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValueProps.ts
```diff
@@ -0,0 +1,8 @@
+import { GridFilterItem } from '../../../models/gridFilterItem';
+
+export interface GridFilterInputMultipleValueProps {
```
m4theushw, line 3: The props should be inside `GridFilterInputMultipleValue`. `GridFilterInputValue` is not a good example.

packages/grid/_modules_/grid/components/menu/columnMenu/GridColumnsMenuItem.tsx
```diff
@@ -48,8 +48,9 @@ GridColumnsMenuItem.propTypes = {
     filterOperators: PropTypes.arrayOf(
       PropTypes.shape({
         getApplyFilterFn: PropTypes.func.isRequired,
-        InputComponent: PropTypes.elementType,
+        InputComponent: PropTypes.func,
```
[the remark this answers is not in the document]
alexfauquette: I think it is because the two JSX elements have different props and then are considered as different objects. So it provides the common property: they are both function

alexfauquette:
I added some tests, don't hesitate if you see more to add

packages/grid/_modules_/grid/models/colDef/gridNumericOperators.ts
```diff
@@ -109,4 +110,20 @@ export const getGridNumericColumnOperators = (): GridFilterOperator[] => [
       };
     },
   },
+  {
+    label: 'is any of',
+    value: 'isAnyOf',
+    isArrayValue: true,
+    getApplyFilterFn: (filterItem: GridFilterItem) => {
+      if (filterItem.value == null || filterItem.value.length === 0) {
```
m4theushw, line 118: Checking if it's an array seems to be more reliable.

```suggestion
      if (!Array.isArray(filterItem.value) || filterItem.value.length === 0) {
```

packages/grid/_modules_/grid/models/colDef/gridSingleSelectOperators.ts
```diff
@@ -35,4 +36,23 @@ export const getGridSingleSelectOperators: () => GridFilterOperator[] = () => [
...
+    getApplyFilterFn: (filterItem: GridFilterItem) => {
+      if (!filterItem.value || filterItem.value.length === 0) {
+        return null;
+      }
+
+      return ({ value }): boolean => {
+        if (typeof value === 'object') {
+          return filterItem.value.includes(value);
+        }
+        return filterItem.value.includes(value);
```
m4theushw, line 52: The return values are the same.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,109 @@
...
+      clearTimeout(filterTimeout.current);
+      setFilterValueState(value.map((x) => String(x)));
+
+      if (type !== 'singleSelect' && value === []) {
+        setIsApplying(false);
+        return;
+      }
+
+      setIsApplying(true);
+      // TODO singleSelect doesn't a debounce
```
m4theushw, line 49: If the type is `singleSelect` we don't need to debounce. You can call `applyValue` immediately in this situation.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,109 @@
...
+      }, SUBMIT_FILTER_STROKE_TIME);
+    },
+    [applyValue, item, type],
+  );
+
+  const InputProps = applying ? { endAdornment: <GridLoadIcon /> } : {};
+
+  return (
+    <Autocomplete
+      {...others}
```
m4theushw, line 62: The rest of the props should be applied at the end.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export interface GridTypeFilterInputMultipleValueProps extends GridFilterInputMultipleValueProps {
+  type?: 'text' | 'number' | 'singleSelect';
+}
+
+function GridFilterInputMultipleValue(
+  props: GridTypeFilterInputMultipleValueProps &
+    Omit<AutocompleteProps<any[], true, false, true>, 'options' | 'renderInput'>,
+) {
+  const { item, applyValue, type, apiRef, ...others } = props;
```
m4theushw, line 21:
```suggestion
  const { item, applyValue, type, apiRef, ...other } = props;
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const InputProps = applying ? { endAdornment: <GridLoadIcon /> } : {};
+
+  return (
+    <Autocomplete
+      {...others}
+      multiple
+      freeSolo
+      options={
+        type === 'singleSelect' ? apiRef.current.getColumn(item.columnField).valueOptions : []
```
m4theushw, line 66: Instead of an empty array, we could use the text that the user is typing as the first option. It would be closer to https://github.com/mui-org/material-ui-x/issues/2437#issuecomment-903868956. At the first moment, I didn't know I had to press <kbd>Enter</kbd> to filter. BTW, pressing <kbd>Tab</kbd> should also add the typed value as option.

m4theushw (commented):
Looks good for an initial version. We only need to polish some things to avoid the need to modify  the core filtering logic to add this new filter. Why did you add `isArrayValue`?

alexfauquette:
@m4theushw the problem solved by `isArrayValue` is in the `changeOperator` method. When you change from "contains" friter to "isAnyOf" the `changeOperator` has to format the value. Whereas if you change from "contains" to "equals", there is no need to format the value.

The `isArrayValue` is here to know if the value type is the same or if we have to change it.

alexfauquette:
I corrected all the feedback except, removing `isArrayValue`. 

I'm still thinking about how to avoid it, but it seems to be mandatory, especially if developers use array in the value such as proposed in issue #3131 . In such a case it seems impossible to infer if the filter is an array of values or if the value is an array. I think developers have to specify that the filter input takes an array of values.

alexfauquette:
During the weekend I get the idea to keep `isArrayValue` logic but rename it `multiple` like the `Autocomplete` component: You can get an array of values if you set `multiple=true`

m4theushw:
I feel that `isArrayValue` was added to detour from something that is actually a bug. The filter value should not be reused when changing to an operator which uses a different filter component. Imagine that you're using the "contains" operator. Changing to "equals" should reuse the value because it's the same component. However, if on "is any of" operator and "contains" is selected, then the value should be reset since the types are different (array vs string).

flaviendelangle:
If the doc is not ready, can we keep this PR as a draft ?

alexfauquette:
The doc has been added here: deploy-preview-2874--material-ui-x.netlify.app/components/data-grid/filtering/#multiple-values-operator

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { item, applyValue, type, apiRef, focusElementRef, ...other } = props;
+  const [filterValueState, setFilterValueState] = React.useState(item.value || []);
+  const id = useId();
+
+  React.useEffect(() => {
+    const itemValue = item.value ?? [];
+    setFilterValueState(itemValue.map(String));
+  }, [item.value]);
+
+  const onFilterChange = React.useCallback(
```
m4theushw, line 29: To keep with the React convention.

```suggestion
  const handleChange = React.useCallback(
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
```diff
@@ -0,0 +1,95 @@
...
+    (event, value) => {
+      const parsedValue = value.map(String);
```
m4theushw, line 31: Why to convert to string?

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const id = useId();
+
+  const currentColumn = item.columnField ? apiRef.current.getColumn(item.columnField) : null;
+  const currentValueOptions = React.useMemo(() => {
+    return typeof currentColumn.valueOptions === 'function'
+      ? currentColumn.valueOptions({ field: currentColumn.field })
+      : currentColumn.valueOptions;
+  }, [currentColumn]);
+
+  const filterValueOptionFormater = getSingleSelectOptionFormatter(
```
m4theushw, line 42:
```suggestion
  const filterValueOptionFormatter = getSingleSelectOptionFormatter(
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,148 @@
...
+    [],
+  );
+
+  React.useEffect(() => {
+    let itemValue;
+
+    if (currentValueOptions !== undefined) {
+      // sanitize if valueOptions are provided
+      const parsedCurrentValueOptions = currentValueOptions.map(filterValueParser);
+      itemValue = item.value
```
m4theushw, line 58: If `item.value` is not an array, e.g. `undefined`, it crashes.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,148 @@
...
+        .map(filterValueParser)
+        .filter((element) => parsedCurrentValueOptions.includes(element));
+      if (itemValue !== item.value) {
```
m4theushw, line 61: You're comparing the reference, but the references will never be the same because `Array.filter` outputs a new array. This causes a infinite loop when I try to filter for a value. Either use `isDeepEqual` or compare the number of non-existing options found.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -982,6 +1017,57 @@ describe('<DataGrid /> - Filter', () => {
         expect(getColumnValues()).to.deep.equal(['Nike', 'Adidas', 'Puma']);
       });
 
+      it('should allow operator isAnyOf', () => {
+        const { setProps } = render(<TestCase value={[]} operatorValue="isAnyOf" />);
+        setProps({
+          field: 'country',
+          operatorValue: 'isAnyOf',
+
```
m4theushw, line 1025: These blank lines between the operator and the value could be removed.

docs/src/pages/components/data-grid/filtering/filtering.md
```diff
@@ -116,6 +116,35 @@ In this demo, you can see how to create a completely new operator for the Rating
 
 {{"demo": "pages/components/data-grid/filtering/CustomRatingOperator.js", "bg": "inline", "defaultCodeOpen": false}}
 
+### Multiple values operator
+
+To create a custom operator which require multiple value, you can provide an array of values to the `value` property of `filterItem`.
```
m4theushw, line 121:
```suggestion
To create a custom operator which accepts multiple values, you can provide an array of values to the `value` property of the `filterItem`.
```

docs/src/pages/components/data-grid/filtering/filtering.md
```diff
@@ -116,6 +116,35 @@ In this demo, you can see how to create a completely new operator for the Rating
...
+In this case, the `valueParser` of the `GridColDef` will be applied to each item of the array.
+
+The filtering function `getApplyFilterFn` must be adapted to handle `filterItem.value` has an array.
+Here is an example for a "between" operator, applied on the quantity column.
```
m4theushw, line 125:
```suggestion
Below is an example for a "between" operator, applied on the "Quantity" column.
```

m4theushw (commented):
It's taking shape. Could you pull the changes from `master`? The last merge was 9 days ago.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleValue.tsx
alexfauquette: Because the original component was converting value to string before doing the update, so I applied the same logic on the multipleValue component.

https://github.com/mui-org/material-ui-x/blob/61e27becdc716a27205d5afda19d0e0c3dbd94de/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx#L56-L67

alexfauquette: I did not found any interest in doing it, so I removed all the String() from my code

alexfauquette:
I corrected a bug that is present from  long time ago.

Here is a deployment of PR from September. Take the demo "Commodity" and filter a country, the table will be empty, even if there exist rows with this country. https://deploy-preview-2519--material-ui-x.netlify.app/components/data-grid/demo/

The reason: in commodity, coldef, the value parser transform the value `"GH"` to `{value: "GH", name: "Ghana", ...}` But the `gridSingleSelectOperators.ts` assumes that row values contain only the value.

The quick fix is commit "fix commodity demo"

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -253,7 +252,7 @@ function GridFilterForm(props: GridFilterFormProps) {
           ))}
         </Select>
       </FormControl>
-      <FormControl variant="standard" sx={{ width: 190 }}>
+      <FormControl variant="standard" sx={{ minWidth: 190 }}>
```
m4theushw, line 255: I'm not sure if removing the width is correct, maybe we should set `max-width` too. cc @oliviertassinari @danilo-leal 

image

m4theushw (commented):
I noticed an inconsistency. The selected options are being formatted differently between "is" and "is any of". I guess `filterItem.value` should contain an array of objects, not the values, or we normalize the options passed to the `value` prop of the Autocomplete.

GvUfcuTLLV

---

In https://deploy-preview-2874--material-ui-x.netlify.app/components/data-grid/filtering/#change-the-input-component we should remove the "is any of" operator since a custom component is provided for all operators.

https://github.com/mui-org/material-ui-x/blob/739a86b64be66f5cdf7a8266d503af09e90201d0/docs/src/pages/components/data-grid/filtering/ExtendNumericOperator.tsx#L62-L65

Being able to select the new operator is weird, since it doesn't work here.

image

alexfauquette:
To be consistent with the other `singleSelect`, the `item.value` only contains the value, such that the `valueParser` gets the same input format. And the local state of the component contains the objects if `valueOptions` is made of objects.

m4theushw:
@alexfauquette Then the alternative seems to be to map the values provided to the chips, so it's always an object instead of a value.

alexfauquette:
@m4theushw I've been unclear in my previous message. The formating problem is solved. I was briefly describing how I change the state management such that Chip formatting works.

So with the new commits, we have:
- value provided to the `filterModel`: an array of value 
- value provided to the `autoComplete`: an array of object

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+    } else if (currentValueOptions !== undefined) {
+      // sanitize if valueOptions are provided
+      const parsedCurrentValueOptions = currentValueOptions.map(getValueFromOption);
+      const itemValueIndexes = item.value.filter((element) => {
+        const formatedElement = getValueFromOption(element);
+        const index = parsedCurrentValueOptions.findIndex(
+          (formatedOption) => formatedOption === formatedElement,
+        );
+
+        return index;
```
m4theushw, line 65: `Array.filter` should return a boolean value.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+      });
+      itemValue = itemValueIndexes
+        .filter((index) => index >= 0)
+        .map((index) => currentValueOptions[index]);
+
+      if (itemValue.length !== item.value.length) {
+        // remove filtered values
+        applyValue({ ...item, value: parsedCurrentValueOptions });
```
m4theushw, line 73: This is causing an infinite loop. Maybe we're missing a test case covering the selection of items to filter with a click, not only changing props.

alexfauquette: Yes, tests were made only on the operator "is", so I added similar ones on "isAnyOf" to verify if the values are cleaned when changing the column.

I also refactorized how value is sanitized. The AutoComplete requires to have coherent values with options, so I sanitize the value with a useMemo. Such that the sanitization is performed before rendering the AutoComplete component.

If sanitization removes some elements, the DataGrid state is updated by a dedicated hook.

I spend too much time on this little piece of code. I'm taking a break on it, and I come back tomorrow to check on the deployment if it is working as expected, or if bugs are remaining

alexfauquette: I played with filtering, no trivial bug was found this time

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+  getValueFromOption(option) === getValueFromOption(value);
+
+const filter = createFilterOptions<any>();
+
+function GridFilterInputMultipleSingleSelect(props: GridFilterInputMultipleSingleSelectProps) {
+  const { item, applyValue, type, apiRef, focusElementRef, ...other } = props;
+  const id = useId();
+
+  const currentColumn = item.columnField ? apiRef.current.getColumn(item.columnField) : null;
+  const currentValueOptions = React.useMemo(() => {
```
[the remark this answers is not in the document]
alexfauquette: Yes, it will be more logic, since data are resolved from the apiRef. I've done the modification for the following namings:
- resolvedColumn
- resolvedValueOptions
- resolvedFormattedValueOptions

alexfauquette:
I will wait for #3437 to adapt the documentation

flaviendelangle:
#3437 is merged, we can fix the conflicts and merge this one

flaviendelangle (approved):

PR #2907, opened 2021-10-18
[internal] Revert `36124c6`

DanailH:
Part of https://github.com/mui-org/material-ui-x/issues/1902
Based on this comment https://github.com/mui-org/material-ui-x/pull/2393#pullrequestreview-741985582

~~Will keep it in draft, for now, to test if the original issue persists under v5.~~

Seems to be working -> https://codesandbox.io/s/react-forked-v8mur?file=/src/App.tsx


DanailH:
I don't think it's needed. I also don't think it brings any performance optimizations.

Issue #1313, opened 2021-03-29
[docs] Cover how to parse string to a different type


Bluefitdev:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

Using `datetime` or `date` column type does not work as expected. Filter is not working, also it shows as raw data from db. 

For instance, my json api request:
`created_at: 2021-03-23T06:23:44.000000Z`

## Expected Behavior 🤔

I suspected that you only expect `Date` type of value in the cell (as it shown in every example in the website), but after inspecting the commit [here](https://github.com/mui-org/material-ui-x/pull/411/commits/3dc2b24846d15cee60ebec082ad5abe24db77a26), apparently filter should also accept string.

Also I thought date should be type should be formatted if taken in `Date` object of js? Right now i'm solving it like this:

```
    {
      field: "created_at",
      headerName: "Requested day",
      type: "date",
      valueFormatter: (params) => dayjs(params.value).format("DD MMM YYYY"),
      filterable: false,
    },
```

However filter is still not working

## Steps to Reproduce 🕹

Here's the [sandbox](https://codesandbox.io/s/datagrid-bug-demo-3tmw7?file=/src/App.js), however I'm using XGrid. I don't want to put my license in the sandbox so i used datagrid instead.

## Your Environment 🌎
Browser: Chrome

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
npx: installed 2 in 4.577s

  System:
    OS: macOS 11.2.3
  Binaries:
    Node: 12.20.0 - ~/.nvm/versions/node/v12.20.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v12.20.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.20.0/bin/npm
  Browsers:
    Chrome: 89.0.4389.90
    Edge: Not Found
    Firefox: 76.0.1
    Safari: 14.0.3
  npmPackages:
    @material-ui/core: ^4.11.2 => 4.11.2
    @material-ui/icons: ^4.11.2 => 4.11.2
    @material-ui/lab: ^4.0.0-alpha.57 => 4.0.0-alpha.57
    @material-ui/pickers: 3.2.10 => 3.2.10
    @material-ui/styles:  4.11.2
    @material-ui/system:  4.11.2
[...]

PR #3003, opened 2021-10-27
[docs] Explain how to use valueGetter to transform type

alexfauquette:
Solves #1313

add a paragraph in "column types" section to explain that column of type `"dateTime"` should not be stored with a string value.

I added subsections to separate clearly the two topics:
 - the additional properties
 - data type

preview: https://deploy-preview-3003--material-ui-x.netlify.app/components/data-grid/columns/#column-types

### DIFF
--- a/docs/src/pages/components/data-grid/columns/columns.md
+++ b/docs/src/pages/components/data-grid/columns/columns.md
@@ -280,7 +280,7 @@ The following are the native column types:
 
 ### Converting types
 
-Default methods such as filtering or sorting assume that `type` corresponds to the type of the stored values.
+Default methods, such as filtering and sorting, assume that the type of the values will match the type of the column specified in `type` (e.g. the values of a `number` column will be numbers).
 For example, values of column with `type: 'dateTime'` are expecting to be stored as a `Date()` objects.
 If for any reason, your data type is not the correct one, you can use `valueGetter` to parse the value to the correct type.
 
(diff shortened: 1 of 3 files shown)


docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -282,6 +278,22 @@ The following are the native column types:
 - `'singleSelect'`
 - `'actions'`
 
+### Converting types
+
+Default methods such as filtering or sorting assume that `type` corresponds to the type of the stored values.
+For example, values of column with `type:'dateTime'` are expecting to be stored as a `Date()` objects.
```
m4theushw, line 284:
```suggestion
For example, values of column with `type: 'dateTime'` are expecting to be stored as a `Date()` objects.
```

docs/src/pages/components/data-grid/columns/columns.md
```diff
@@ -282,6 +278,22 @@ The following are the native column types:
 - `'singleSelect'`
 - `'actions'`
 
+### Converting types
+
+Default methods such as filtering or sorting assume that `type` corresponds to the type of the stored values.
```
m4theushw, line 283:
```suggestion
Default methods, such as filtering and sorting, assume that the type of the values will match the type of the column specified in `type` (e.g. the values of a `number` column will be numbers).
```

oliviertassinari (approved):

flaviendelangle (approved):

PR #3004, opened 2021-10-27
[DataGrid] Focus on the last value when a filter is added or removed

### LINKED ISSUE #1783: [DataGrid] Input in filter panel should be focus/selected when open

## Current Behavior 😯

The focus is moved to the filter panel container when open.

## Expected Behavior 🤔

The focus is moved to the last textbox in the filter panel when opened. If the textbox is not empty, all the text should be selected.

## Steps to Reproduce 🕹

2. 
focus

## Context 🔦

I have noticed the UX opportunity looking into #1777. Airtable focus as soon as the filter panel is open, it's really great:

Capture d’écran 2021-05-28 à 20 57 26

## Your Environment 🌎

v4.0.0-alpha.29

alexfauquette:
Preview: https://deploy-preview-3004--material-ui-x.netlify.app/components/data-grid/demo/

Fixes #1783

When the filter panel is open, or the number of filter changes, the focus is set on the last filter value field.

`autoFocus` props is not enough for the situation where the last filter is removed. In such a case, the autoFocus is moved to the last remaining filter. But this one is already mounted, so it will not be focussed.

To overcome this problem, I used a useEffect, such that each time `autoFocus` switches to `true` the input is focussed.


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -47,7 +48,11 @@ function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldPr
           ),
         }
       : {};
-
+  React.useEffect(() => {
```
flaviendelangle, line 51: I think we should use the same behavior / naming convention as in the grid cells

`autoFocus` => `hasFocus` (to keep the distinction `autoFocus` = set focus on mount and `hasFocus` = set focus when `true`)
`React.useEffect` => `useEnhancedEffet` (to set the focus sooner)

alexfauquette:
When testing, I found a bug:

When changing the column, the value type can switch from "string" to "boolean", and so the valueInput is a new component, triggering the focus.

To solve that bug, I moved the focusing logic in the `GridFilterForm` component, such that modifying the form does not trigger the focus. And it should also work with custom filter inputs if they implement the `inputRef` prop

DanailH (approved):
Seems to be working as expected. Maybe we can add a simple test to cover that behaviour?

m4theushw:
I noticed two bugs introduced in this PR.

When a filter is added the page jumps:

https://user-images.githubusercontent.com/42154031/140537107-c43b2cad-c8ed-408b-862b-02aa4dc67649.mp4

Once a filter, which is not the last, is removed the focus doesn't return to the immediate below filter input.

https://user-images.githubusercontent.com/42154031/140537468-0d614e9c-6089-45ab-9864-a7b5edad3ed7.mp4

alexfauquette:
For the scroll on focus, this can be removed by `{ preventScroll: true })` But It seems there is a deeper cause of this page jumps. Probably due to the absolute position of the filter panel. But I do not know how to investigate it.

I found a similar bug in the doc: 
[https://mui.com/components/data-grid/filtering/#change-the-input-component](https://mui.com/components/data-grid/filtering/#change-the-input-component) The first time I open filter pannel of the example "change the input component" it scrolls to the top of the page.

For the focus management, Is a bit tricky, because the props of the last FilterForm do not change when we remove a line. I see two solutions:
- To pass a prop `toFocus` which is an object `{value: true}` if the Form should focus on the value input. This would allow triggering the focus again when updating the object (event is value is still to `true`)
- To set the focus on the delete button of the line just after the removed one, such as the user can remove another line. This implies keeping in memory the item Ids to know which one has been removed.

Does anyone have preferences between these solutions?

m4theushw:
`preventScroll` is not supported on Safari: https://caniuse.com/mdn-api_htmlelement_focus_options_preventscroll_parameter

I took a look at the root cause of this bug. The problem is that the filter is mounted immediately at the bottom of the page after the Popper (the MUI component) is open. Then, `.focus()` is called and it tries to scroll there. Only after the Popper instance (the plugin) is created it goes to the correct position. To detour from the limitation imposed by Safari, this bug will require a more complex solution. I was thinking about publishing an event when the Popper (the plugin) is fully initialized. Listening to this event combined to controlling which component has focus should fix the bug. You could have a `hasFocus` state in `GridFilterPanel` with the index of the filter item to focus. This would also allow to focus on the next filter once one is deleted.

```diff
diff --git a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
index 79be43418..3a64db6dd 100644
--- a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
@@ -82,6 +82,14 @@ const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>((props, ref)
           name: 'flip',
           enabled: false,
         },
+        {
+          name: 'publishEvent',
+          phase: 'main',
+          enabled: true,
+          fn: () => {
+            // publish event
+          },
+        },
       ]}
       {...other}
     >
```

Docs: https://popper.js.org/docs/v2/modifiers/#example-modifier

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
```diff
@@ -82,12 +85,22 @@ const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>((props, ref)
           name: 'flip',
           enabled: false,
         },
+        {
+          name: 'isPlaced',
+          enabled: true,
+          phase: 'main',
+          fn: ({ state }) => {
+            if (state.placement === GRID_PANEL_POSITION) {
```
m4theushw, line 93: If the placement is not `bottom-start` the filter panel content is never visible. Could we make it to work on any placement?

```suggestion
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
```diff
@@ -49,6 +50,11 @@ export function GridFilterPanel() {
     }
   }, [addNewFilter, filterModel.items.length]);
 
+  React.useEffect(() => {
+    // use an object instead of a boolean to trigger the focus of GridFilterForm by modifying the object.
+    setHasFocus({ value: true });
```
m4theushw, line 55: I don't understand why to use an object. Wouldn't be better to use `filterItem.id`? In this sense, we keep which filter form to focus on. Once a filter is removed, we change to focus the next filter.

Notice that it's still jumping when opening the panel. The value of `filterModel.items.length` is 0 initially, then the first filter is added and it becomes 1.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -23,6 +27,7 @@ export interface GridFilterFormProps {
   showMultiFilterOperators?: boolean;
   multiFilterOperator?: GridLinkOperator;
   disableMultiFilterOperator?: boolean;
+  toFocus?: { value: boolean };
```
m4theushw, line 30: To keep with the convention: https://github.com/mui-org/material-ui-x/blob/0fd81307c6ffdc1031726680a5f302bb280058c4/packages/grid/_modules_/grid/components/cell/GridCell.tsx#L27

```suggestion
  hasFocus?: { value: boolean };
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -132,6 +140,16 @@ function GridFilterForm(props: GridFilterFormProps) {
 
   const currentOperator = getCurrentOperator();
 
+  useEnhancedEffect(() => {
+    if (toFocus && toFocus.value) {
+      if (valueInputRef.current) {
+        valueInputRef.current!.focus();
+      } else {
+        filterSelectorRef.current!.focus();
+      }
```
m4theushw, line 149: If the user provided a custom component (e.g. [Slider](https://mui.com/components/slider/)), which doesn't support `inputRef` as well as the control to focus is a inner element, calling `filterSelectorRef.current!.focus()` won't work.

If think we could add an imperative method `focus()` on each filter component so the task of focusing the right element is delegated to them. As an example: https://github.com/mui-org/material-ui/blob/cd247a197b772b74d81db70e1c3dc5d0a019c447/packages/mui-material/src/Select/SelectInput.js#L156-L166

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
alexfauquette: If I have 2 Items with ID1 and ID2 and `hasFocus='ID2'` Now, I remove the filter with ID1.

The focus disappears because it has moved to the delete button of FilterForm with ID1 which is now removed by React.

So I want to set again focus on the last FilterItem which is still ID2. Then I get `hasFocus='ID2'`  which is the same as before, and the FilterForm props `hasFocus` is still `true` so it does not trigger a next `focus()`. 

Being able to call `focus()` from the FilterPannel seems to be the good solution to handle this logic

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: `filterSelectorRef` is a ref on the selector of the filter type which is a `Select` component provided by ourselves. I did not think about custom components, but about filters without input like "is empty"

image

flaviendelangle (approved):

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
m4theushw: With custom filter operators that don't hold `inputRef` it doesn't work correctly:

https://deploy-preview-3004--material-ui-x.netlify.app/components/data-grid/filtering/#create-a-custom-operator

_Ignore the jump when opening the panel, it's a different bug._

s8J3hWrClL

I was proposing to add the imperative method on the filter's `InputComponent` so users can set which element to focus. 

For the situation where a filter may not require a value, you can check if it has a `InputComponent` defined and then move the focus to the operator selector instead.

https://github.com/mui-org/material-ui-x/blob/a2b1f5dc4b031cd041e9ce55953308076362221a/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts#L104

alexfauquette:
@m4theushw I don't see what you want to do with `InputComponent`. as it is, developers can have the focus working by setting the `inputRef` prop on the component they want to focus.

m4theushw:
I would like to reproduce the same pattern used in the core for custom components: https://mui.com/components/text-fields/#integration-with-3rd-party-input-libraries

The `inputRef` is commonly used for components that have a `<input />` inside and expose it has a ref. The Slider doesn't expose its input as a ref.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValueProps.ts
```diff
@@ -1,8 +1,10 @@
+import React from 'react';
```
m4theushw, line 1:
```suggestion
import * as React from 'react';
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputBoolean.tsx
```diff
@@ -32,11 +35,12 @@ export function GridFilterInputBoolean(props: GridFilterInputValueProps & TextFi
       InputLabelProps={{
         shrink: true,
       }}
+      inputRef={ref}
```
m4theushw, line 38: The ref prop should be forwarded to the root element: https://mui.com/guides/api/#ref

What about using `focusElementRef` as the ref of the element to focus?

docs/src/pages/components/data-grid/filtering/CustomRatingOperator.js
```diff
@@ -17,7 +17,16 @@ const useStyles = makeStyles({
...
 function RatingInputValue(props) {
   const classes = useStyles();
-  const { item, applyValue } = props;
+  const { item, applyValue, focusElementRef } = props;
+
+  const ratingRef = React.useRef(null);
+  React.useImperativeHandle(focusElementRef, () => ({
+    focus: () => {
+      ratingRef.current
+        .querySelector(`input[value="${Number(item.value) || ''}"]`)
```
alexfauquette, line 26: the `|| ''` allows focusing on the input field if the `item.value` is undefined

docs/src/pages/components/data-grid/filtering/CustomRatingOperator.js
```diff
@@ -72,7 +88,7 @@ export default function CustomRatingOperator() {
   const { data } = useDemoData({ dataSet: 'Employee', rowLength: 100 });
   const columns = [...data.columns];
   const [filterModel, setFilterModel] = React.useState({
-    items: [{ columnField: 'rating', value: '3.5', operatorValue: 'from' }],
+    items: [{ id: 1, columnField: 'rating', value: '3.5', operatorValue: 'from' }],
```
alexfauquette, line 91: Without the id, the first value update create a rerender, and the lost of the focus in the demo

m4theushw (approved):
Looks good. You only need to fix the build and merge.

Issue #2836, opened 2021-10-10
[DataGrid] Fails to edit field of type `singleSelect` when paired with `editRowsModel`


TeoTN:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Given that there's an editable column defined with type `singleSelect`, 
and the DataGrid has controlled edit model (uses `editRowsModel` and `onEditRowsModelChange` props), 
when user double clicks on the field to edit the cell and selects an option from the dropdown
then the cell is not updated

(see below for reproduction) 

### Expected behavior 🤔

Cell should be editable

### Steps to reproduce 🕹

Minimal reproduction: https://codesandbox.io/s/mui-datagrid-bug-option-editing-updny?file=/demo.tsx
Copied an example from docs page and added single select column to demonstrate editing this cell won't work.

### Context 🔦

I'd like to have client-side validation when some fields are enums

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

  System:
    OS: Linux 4.19 Ubuntu 20.04.2 LTS (Focal Fossa)
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.15 - ~/.yarn/bin/yarn
    npm: 7.10.0 - ~/.nvm/versions/node/v14.16.0/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @emotion/react: 11.4.1 => 11.4.1
    @emotion/styled: 11.3.0 => 11.3.0
    @mui/core:  5.0.0-alpha.49
    @mui/icons-material: 5.0.1 => 5.0.1
    @mui/lab: 5.0.0-alpha.49 => 5.0.0-alpha.49
    @mui/material: 5.0.2 => 5.0.2
    @mui/private-theming:  5.0.1
    @mui/styled-engine:  5.0.1
    @mui/styles: 5.0.1 => 5.0.1
    @mui/system:  5.0.2
    @mui/types:  7.0.0
    @mui/utils:  5.0.1
    @mui/x-data-grid: 5.0.0-beta.3 => 5.0.0-beta.3
    @types/react: 17.0.3 => 17.0.3
    react: 17.0.2 => 17.0.2
    react-dom: 17.0.2 => 17.0.2
    typescript: 4.3.5 => 4.3.5
</details>

TeoTN:
[...]

PR #3006, opened 2021-10-27
[DataGrid] Add API for synchronous validation

m4theushw:
Solve the validation use case of #2836 

Preview: https://deploy-preview-3006--material-ui-x.netlify.app/components/data-grid/editing/#client-side-validation

This PR adds an `onEditCellPropsChange` prop to the `GridColDef` to allow users to synchronously transform the props that are saved in the internal state. The most common use case for it is to do validation. Currently, the way we teach users to do validation, by controlling the `editRowsModel` prop, is error-prone and it won't work in the long term as it requires the grid to re-render to get the updated value. For now, there will be two `onEditCellPropsChange` props (there's one tied to the respective event). In v6, the event prop will be removed.


packages/grid/_modules_/grid/hooks/features/editRows/useGridEditRows.ts
```diff
@@ -266,24 +305,55 @@ export function useGridEditRows(
...
+      const row = apiRef.current.getRow(id)!;
+
+      const commitParams: GridCellEditCommitParams = {
+        ...params,
+        value: editCellProps.value,
+      };
+
+      let hasError = !!editCellProps.error;
+      if (!hasError && typeof column.onEditCellPropsChange === 'function') {
+        const result = column.onEditCellPropsChange({ id, row, props: editCellProps });
```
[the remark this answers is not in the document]
m4theushw: I renamed it to `preProcessEditCellProps`.

docs/src/pages/components/data-grid/editing/ConditionalValidationGrid.js
```diff
@@ -73,33 +47,61 @@ const rows = [
...
+    },
+    {
+      field: 'paymentMethod',
+      headerName: 'Payment method',
+      type: 'singleSelect',
+      valueOptions: ['Credit card', 'Wire transfer', 'Cash'],
+      width: 160,
+      editable: true,
+      onEditCellPropsChange: (params) => {
+        const editRowsModel = apiRef.current.getEditRowsModel();
```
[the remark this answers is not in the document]
oliviertassinari: The discussion has moved to https://www.notion.so/mui-org/Free-version-Pro-version-scopes-c81b2e6fee4d404880fad5bd25836739

m4theushw:
> Does it? I get that if fixes most of the pain, but it seems that the behavior with the controlled prop is still wrong. For the solution, could it be the deprecation of the editRowsModel controlled prop?

@oliviertassinari There's nothing we can do to fix it while using the controlled prop. We need the latest value without waiting for a re-render. The removal of the controlled prop will be done in v6. For now, this PR is introducing a new API and promoting it as the solution for validation. While the `apiRef` is not public, the controlled prop is necessary if someone wants to put a cell in edit mode not using the default triggers.

packages/grid/_modules_/grid/models/api/gridEditRowApi.ts
```diff
@@ -64,12 +64,12 @@ export interface GridEditRowApi {
    * @param {React.SyntheticEvent} event The event to pass forward.
    * @returns {boolean} A boolean indicating if there is an error.
    */
-  commitCellChange: (params: GridCommitCellChangeParams, event?: any) => boolean;
+  commitCellChange: (params: GridCommitCellChangeParams, event?: any) => boolean | Promise<Boolean>;
```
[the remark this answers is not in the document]
m4theushw: Updated.

docs/src/pages/components/data-grid/editing/ConditionalValidationGrid.js
```diff
@@ -73,33 +47,61 @@ const rows = [
...
-  const handleEditRowsModelChange = React.useCallback((newModel) => {
-    const updatedModel = { ...newModel };
-    Object.keys(updatedModel).forEach((id) => {
-      const hasError =
-        updatedModel[id].isPaid.value && !updatedModel[id].paidAt.value;
-      updatedModel[id].paidAt = { ...updatedModel[id].paidAt, error: hasError };
-    });
-    setEditRowsModel(updatedModel);
-  }, []);
+  const columns = [
```
flaviendelangle, line 58: We should memoize it to give a good example to users.
Non-memoized rows and columns tends to create unwanted behaviors.

x other cases

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -77,23 +77,36 @@ The following demo shows how these two functions can be used:
 
 ### Client-side validation
 
-To validate the value in the cells, use `onEditRowsModelChange` to set the `error` attribute of the respective field when the value is invalid.
-If this attribute is true, the value will never be committed.
-This prop is invoked when a change is triggered by the edit cell component.
+To validate the value in the cells, first add an `preProcessEditCellProps` callback to the [column definition](/api/data-grid/grid-col-def/) of the respective field to validate.
```
flaviendelangle, line 80:
```suggestion
To validate the value in the cells, first add a `preProcessEditCellProps` callback to the [column definition](/api/data-grid/grid-col-def/) of the field to validate.
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -174,19 +187,24 @@ You can handle the `onEditRowsModelChange` callback to control the `GridEditRows
 
 {{"demo": "pages/components/data-grid/editing/RowEditControlGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
-### Conditional validation
+### Conditional validation [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
 
 Having all cells of a row in edit mode allows validating a field based on the value of another one.
-To accomplish that, set the `onEditRowsModelChange` prop and return a new model with the `error` attribute of the invalid field set to `true`.
-Use the other fields available in the model to check if the validation should run or not.
+To accomplish that, start by adding the `onEditCellPropsChange` as explained in the [cell editing](#client-side-validation).
```
flaviendelangle, line 193:
```suggestion
To accomplish that, start by adding the `preProcessEditCellProps` as explained in the [cell editing](#client-side-validation).
```

docs/src/pages/components/data-grid/editing/ConditionalValidationGrid.js
m4theushw: Not always. `React.useMemo` will only be useful if the component re-renders, e.g. as result of a state change. Besides that we're introducing the solution without first having the problem. We only recommend memoizing the columns, not the rows: https://mui.com/components/data-grid/columns/#column-definitions. In this specific demo, memoization is not necessary because it doesn't have state.

flaviendelangle: > We only recommend memoizing the columns, not the rows: https://mui.com/components/data-grid/columns/#column-definitions.

We should recommend it for the rows, it re-applies the filter / sorting / selection so in real applications it is an heavy task/

>  In this specific demo, memoization is not necessary because it doesn't have state.

I think it is simpler to do it on all our demos.
Until now, we always generated them outside of the component, even if the component did not re-render. It was not necessary for the component to work correctly. Of course if the Grid never re-renders, it is not actually doing anything.
But it makes it very clear for our user that we rely on the rows and columns reference for updates. Whereas personally if I saw a demo passing an array generated in the render method without memoization, I would probably take for granted that it can change between renders without any problem.

flaviendelangle (approved):

PR #3012, opened 2021-10-28
[core] Reduce styles complexity

DanailH:
Reorganizing the grid styles. Due to performance reasons the styles for the column headers, cells and rows are kept in the `GridRootStyles`. All other styles that aren't nested in one of those components have been moved out of this file and into their respective component.

This PR doesn't have any breaking changes.


packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -76,6 +77,82 @@ const useUtilityClasses = (ownerState: OwnerState) => {
   return composeClasses(slots, getDataGridUtilityClass, classes);
 };
 
+const GridCellRoot = styled('div', {
```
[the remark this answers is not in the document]
DanailH: Reverted the changes for the mentioned components.

packages/storybook/src/stories/grid-toolbar.stories.tsx
```diff
@@ -66,6 +66,7 @@ export const Export = () => {
   return (
     <div style={{ height: 600 }}>
       <DataGridPro
+        disableVirtualization
```
flaviendelangle, line 69: Why do we disable it on this story ?

flaviendelangle:
Looks good to me

packages/grid/data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -731,4 +731,43 @@ describe('<DataGrid /> - Layout & Warnings', () => {
...
+
+    render(
+      <ThemeProvider theme={theme}>
+        <div style={{ width: 300, height: 300 }}>
+          <DataGrid {...baselineProps} />
+        </div>
+      </ThemeProvider>,
+    );
+
+    const gridRoot = document.querySelector('[role="grid"]');
```
m4theushw, line 765:
```suggestion
    const gridRoot = screen.getByRole('grid');
```

packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx
```diff
@@ -20,6 +21,19 @@ const useUtilityClasses = (ownerState: OwnerState) => {
   return composeClasses(slots, getDataGridUtilityClass, classes);
 };
 
+const InputBaseStyled = styled(InputBase, {
```
m4theushw, line 24: @mui-org/x Do we agree on this convention? The styled components are always named to match the parent component name?

```suggestion
const GridEditInputCellRoot = styled(InputBase, {
```

m4theushw (commented):
I didn't find any performance issues. 💯

packages/storybook/src/stories/grid-toolbar.stories.tsx
DanailH: Ou no idea - reverted.

packages/grid/_modules_/grid/components/cell/GridEditInputCell.tsx
DanailH: I'm ok with this. I used `Styled` as that seemed to be the naming convention used in the core but I can see how what you are suggesting also makes sense in our case. I've updated this and all the other places where we had this name construct.

m4theushw (approved):

packages/grid/_modules_/grid/components/containers/GridFooterContainer.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import { unstable_composeClasses as composeClasses } from '@mui/material';
```
oliviertassinari, line 3: This looks problematic. We bundle the whole `@mui/material` dependency in dev mode.

packages/grid/_modules_/grid/components/containers/GridFooterContainer.tsx
```diff
@@ -19,13 +20,26 @@ const useUtilityClasses = (ownerState: OwnerState) => {
...
+}));
+
 export const GridFooterContainer = React.forwardRef<HTMLDivElement, GridFooterContainerProps>(
   function GridFooterContainer(props: GridFooterContainerProps, ref) {
     const { className, ...other } = props;
     const rootProps = useGridRootProps();
     const ownerState = { classes: rootProps.classes };
     const classes = useUtilityClasses(ownerState);
 
-    return <div ref={ref} className={clsx(classes.root, className)} {...other} />;
```
oliviertassinari, line 29: Something is off with the `classes`. `gridClasses` has `footerContainer` and yet, here, we create the class name a second time.

I think that we should either:

1. have `<DataGrid classes={{ footerContainer: 'foo' }} />` to actually work and use `gridClasses.footerContainer` here
2. or change the name of the styled component: `name: 'MuiDataGrid',` -> `name: 'GridFooterContainer',`
  `slot: 'FooterContainer',` -> `slot: 'Root',`

oliviertassinari (commented):
Great to see this one in. A few notes, looking at what was already here

packages/grid/_modules_/grid/components/containers/GridFooterContainer.tsx
flaviendelangle: Fixed https://github.com/mui-org/material-ui-x/pull/3094

packages/grid/_modules_/grid/components/containers/GridFooterContainer.tsx
DanailH: I would stick to the first option since that is the convention we are currently following. IMO it will look off to have `GridFooterContainer`  because that way we would need to go through the other components and update them as well.

m4theushw: @oliviertassinari  I didn't understand the problem and solution proposed in the first option.

- The value of `gridClasses.footerContainer` is `MuiDataGrid-footerContainer` which breaks `composeClasses`.

- `<DataGrid classes={{ footerContainer: 'foo' }} />` works correctly: 

    image

oliviertassinari: Ok, it seems that we are good.

PR #3018, opened 2021-10-28
[DataGrid] Fix filter not working after deleting the value

### LINKED ISSUE #2940: [DataGrid] onFilterModelChange is not fired when last character in value field is deleted
### Duplicates

### Latest version

### Current behavior 😯

If we have a single character written in filter value input box, and then deleted it, the `filterModelChange` event is not fired, thus, the `onFilterModelChange` callback is not executed.

This has been tested on latest Chrome, Firefox and Edge browsers, and the behaviour is the same on all of them.

### Expected behavior 🤔

One would expect that the event is being fired since there actually is a change in the filter model.

### Steps to reproduce 🕹

Steps:

1. Open filter
2. Enter a character in the input value field
3. filterModelChange event is fired
4. Delete the character entered in the step 2
5. filterModelChange is not fired

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
   System:
    OS: Windows 10 10.0.19042
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 1.19 GB / 15.74 GB
  Binaries:
    Node: 14.15.1 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD
  Utilities:
    Git: 2.20.1. - /mingw64/bin/git
  Languages:
    Bash: 4.4.23 - C:\Program Files\Git\usr\bin\bash.EXE
    Java: 1.8.0_201 - /c/Program Files/Java/jdk1.8.0_201/bin/javac
    Perl: 5.26.2 - C:\Program Files\Git\usr\bin\perl.EXE
  Databases:
    PostgreSQL: 10.6 - C:\PostgreSQL\pg10\bin\postgres.EXE
  Browsers:
    Chrome: 94.0.4606.81
    Edge: Spartan (44.19041.1266.0), Chromium (94.0.992.50)
    Internet Explorer: 11.0.19041.1202
```
</details>

### Order ID 💳 (optional)

#23200

### LINKED ISSUE #3010: [DataGrid] Deleting all text in filter value field doesn't show all results.
### Duplicates

Looks related to https://github.com/mui-org/material-ui-x/issues/2940 , the problem there is probably the cause of this issue.

### Latest version

- [X] I have tested the latest version. Happens in both beta5 locally and beta4 in CodeSandBox version.

### Current behavior 😯

When all contents from the input value field is removed, either one by one via Backspace button or just selecting and deleting/cutting out all text in the field, the search filter isn't reset but keeps showing the previous search result.

It seems like the filter doesn't refresh when there is no content in in. (which is the state it ends in after you delete the last letter/symbol in the input field)

### Expected behavior 🤔

Whenever the input value field is empty, all results should be shown. 

### Steps to reproduce 🕹

https://codesandbox.io/s/basicfilteringgrid-material-demo-forked-ri89f?file=/demo.tsx

Check out this forked version of the Filtering demo.

1. Open the filter. It already has the text "rice" written in it. Select the entire word "rice", delete it using the Backspace button on your keyboard. The table still shows no results.

2. Open the filter. Click the X button to clear results. Select Trader Name again. Enter the number 2. Backspace to delete the number 2. Still shows "No Results".

### Context 🔦

I try to search for a name, but I press the number 2 by mistake. I click backspace but I still see no results and it looks like the solution is broken.
Looks very unintuitive.

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3010 
Fixes #2940 

Preview: https://deploy-preview-3018--material-ui-x.netlify.app/

Side node: We need more tests testing the filtering logic with real interaction, not setting the `filterModel` prop.


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -66,14 +66,13 @@ function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldPr
           .find((optionValue) => String(optionValue) === value);
       }
 
+      if (type === 'number' && value === '') {
+        value = NaN; // Prevents the value parser from converting '' to 0 (zero)
+      }
```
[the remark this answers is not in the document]
m4theushw: > I wonder, it feels like we add custom logic for a specific column type.

Yes, but this problem is bigger than that. `GrildFilterInputValue` already contains custom logic for `singleSelect`. We need to split it into different filter components.

> It feels like we have the same bug with the edit feature

I'm aware of this. However, nobody complained about it so I don't if someone is already treating it as a feature (empty the field to get zero) or everybody is adding a custom value formatter. 🤷‍♀️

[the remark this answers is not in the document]
oliviertassinari: > GrildFilterInputValue already contains custom logic for singleSelect

It doesn't make it right or justify to do more of it if it's wrong 😁.

> nobody complained about it

Can you envision a case where it could be right? I don't personally. If I empty the input, I want an empty value, not 0. I can input 0 if I want this value.

packages/grid/data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -312,6 +322,42 @@ describe('<DataGrid /> - Filter', () => {
...
+              },
+            ]}
+          />,
+        );
+        expect(getColumnValues()).to.deep.equal(['Asics', 'RedBull', 'Hugo']);
+        fireEvent.change(screen.getByRole('combobox', { name: 'Operators' }), {
+          target: { value: operatorValue },
+        });
+        const input = screen.getByRole('textbox', { name: 'Value' });
+        fireEvent.change(input, { target: { value: 'abc' } });
```
oliviertassinari, line 352: Instead of using the UI controls, did you consider using the control prop for the filter? Maybe this will be more resilient to future UI changes? For example as in https://codesandbox.io/s/basiceditinggrid-material-demo-forked-ttl15?file=/demo.tsx

m4theushw: We need a balance between testing the whole user interaction and using the control prop. For instance, #2940 was caused because we lacked testing the first. In theory we should be testing both. I don't know if we should have an exact % of each category of test.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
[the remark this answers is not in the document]
m4theushw: I moved this part to the value parser: https://github.com/mui-org/material-ui-x/pull/3018/files#diff-8e275a43cf9db1cb50555f033653ff9cab146383e1b06aba6f3fbb0ebd365735R13

> Can you envision a case where it could be right? I don't personally. As an end-user, if I empty the input, I want an empty value, not 0. I can input 0 if I want this value.

It could be used to force a number when the user empties the input. If the field doesn't allow null values, defaulting to 0 is something desirable.

packages/grid/_modules_/grid/models/colDef/gridNumericColDef.ts
```diff
@@ -10,7 +10,7 @@ export const GRID_NUMERIC_COL_DEF: GridColTypeDef = {
   align: 'right',
   headerAlign: 'right',
   sortComparator: gridNumberComparer,
-  valueParser: (value) => Number(value),
+  valueParser: (value) => (value === '' ? null : Number(value)),
```
flaviendelangle, line 13:
```suggestion
  valueParser: (value) => Number.isNaN(value) ? null : Number(value),
```

Would this be more flexible ?

m4theushw: I don't think so. An empty string becomes 0.

```js
Number.isNaN('') ? null : Number('') // 0
```

flaviendelangle: ... `''` is a valid number for JavaScript I forgot ^^

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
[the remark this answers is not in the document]
DanailH: > Yes, but this problem is bigger than that. `GrildFilterInputValue` already contains custom logic for `singleSelect`. We need to split it into different filter components.

At the time when the `singleSelect` was created it was ok to have it reuse the `GridFilterInputValue` but I agree, it's not ideal and if we introduce `multiSelect` it will become too complex to manage. We can have a separate ticket to track the effort for this tech debt, too much DRY.

flaviendelangle (approved):

PR #3022, opened 2021-10-29
[DataGridPro] Type the event publisher and listener

flaviendelangle:
Preview: https://deploy-preview-3022--material-ui-x.netlify.app/components/data-grid/events/#catalog-of-events

The goal is to have a strict typing on our event publisher / listener arguments. This have several benefits:
- hep us tracking errors in our code
- have a single source of truth to link event name, params and event object
- improve 3rd party developer DX when using `subscribeEvent`
- allow to auto-generate documentation basic examples

## Changes

- [x] Use the lookup to generate the description of the event name instead of hardcoding it 
- [x] Use the lookup to define the prop callback type (`props.onRowClick`)
- [x] Type `useGridApiEventHandler` 
- [x] Type `apiRef.current.publishEvent`
- [x] Type `apiRef.current.subscribeEvent`
- [x] Accept `GridEvents.someEvent` and `"someEvent"` everywhere

We still have the signature of the prop callbacks hardcoded in the comment because otherwise the documentation is not generated correctly from the prop-types.
We could autogen the comments of the TS file at some point but I don't want to over-engineer.

I improved the event table in the doc with a small example of subscription than shows the types and the params available.
This part can be moved to a follow up PR if you want.

image

## Typing of the `event` params

- [x] Type as narrow as possible if the event comes from the DOM / the internal code
- [x] Accept `MuiBaseEvent` if the event potentially come from a `apiRef` method (ex: `rowEditCommit` is triggered via `commitRowChange`)

## Bug fixes

I found a few bugs / behavior inconsistencies after typing the events

- The `cellNavigationKeyDown` event can receive a `GridRowParams` as `params` argument when fired by `handleRowEditStop` in `useGridEditRows`. In this scenario it does not have a `field` and should pick the 1st column. It was working but was not safe since it called `getColumnIndex` with `field=undefined`.

## Public API changes :

- [x] `onCellValueChange` removed because never called
- [x] `cellDragStart` and `cellDragEnd` events removed because never published 

I don't think any of this count as breaking change, it's only type modification and bug fixes.

## Changelog 

- 🎁 Add strict typing to the event publisher and listener #3022 (@flaviendelangle)

  `apiRef.current.subscribeEvent` / `apiRef.current.publishEvent` and `useGridApiEventHandler` are now fully typed and gives you the correct arguments based on the event you are listening to or emitting.

  ```ts
  const handleRowClick: GridEventListener<'rowClick'> = (
    params, // has type `GridRowParams`
    event, // has type `MuiEvent<React.MouseEvent<HTMLElement>>
    details, // has type `GridCallbackDetails
  ) => {
    /* ... */
  };

  // with string event name
  apiRef.current.subscribeEvent('rowClick', handleRowClick)
  useGridApiEventHandler(apiRef, 'rowClick', handleRowClick)

  // or with enum event name
  apiRef.current.subscribeEvent(GridEvents.rowClick, handleRowClick)
  useGridApiEventHandler(apiRef, GridEvents.rowClick, handleRowClick)
  ```


packages/grid/_modules_/grid/models/events/gridEventLookup.ts
```diff
@@ -0,0 +1,204 @@
...
+
+export interface GridRowEventLookup {
+  rowClick: { params: GridRowParams; event: React.MouseEvent<HTMLElement> };
+  rowDoubleClick: { params: GridRowParams; event: React.MouseEvent<HTMLElement> };
+}
+
+export interface GridColumnHeaderEventLookup {
+  columnHeaderClick: {
+    params: GridColumnHeaderParams;
+    event: React.MouseEvent<HTMLElement>;
```
m4theushw, line 34: The type of the event is always `MuiEvent`. In this case: `MuiEvent<React.MouseEvent<HTMLElement>>`

packages/grid/x-data-grid-pro/src/DataGridPro.tsx
```diff
@@ -359,26 +359,26 @@ DataGridProRaw.propTypes = {
...
    * @param {GridCellParams} params With all properties from [[GridCellParams]].
-   * @param {MuiEvent<React.SyntheticEvent>} event The event that caused this prop to be called.
+   * @param {MuiEvent<React.KeyboardEvent | React.MouseEvent>} event The event that caused this prop to be called.
    */
   onCellEditStart: PropTypes.func,
   /**
    * Callback fired when the cell turns to view mode.
    * @param {GridCellParams} params With all properties from [[GridCellParams]].
-   * @param {MuiEvent<React.SyntheticEvent>} event The event that caused this prop to be called.
+   * @param {MuiEvent<MuiBaseEvent>} event The event that caused this prop to be called.
```
m4theushw, line 375: Why `MuiEvent<MuiBaseEvent>` here but `MuiEvent<React.KeyboardEvent | React.MouseEvent>` in the previous one?

I would prefer to be explicit in the generic than to add another layer of complexity.

docs/scripts/buildApi.ts
```diff
@@ -312,16 +312,39 @@ function findProperties(reflection: TypeDoc.DeclarationReflection) {
...
+
   allEvents.forEach((event) => {
     const description = linkify(event.comment?.shortText, apisToGenerate, 'html');
-
+    const eventProperties = eventLookup[event.escapedName!];
     events.push({
       name: event.escapedName!,
       description: renderMarkdownInline(description),
+      params: eventProperties.params,
+      event: eventProperties.event,
```
m4theushw, line 347: You can use the `linkify` function and save the link to the interface instead of only the name.

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -10,33 +22,112 @@ const useStyles = makeStyles({
...
+              <span
+                dangerouslySetInnerHTML={{
+                  __html: event.event,
+                }}
+              />
+            </div>
+          )}
+        </TableCell>
+      </TableRow>
+      <TableRow>
```
m4theushw, line 101: Always rendering the child causes a double-border:

image

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -10,33 +22,112 @@ const useStyles = makeStyles({
...
+    ${args.join('\n    ')}
+  ) => {...}    
+  
+// Imperative subscription    
+apiRef.current.subscribeEvent(
+  GridEvents.${event.name},
+  onEvent,
+);
+
+// Hook subscription (only available inside the Grid scope)
```
m4theushw, line 49:
```suggestion
// Hook subscription (only available inside the scope of the grid)
```

m4theushw (commented):
The demo in https://deploy-preview-3022--material-ui-x.netlify.app/components/data-grid/events/#catalog-of-events is not responsive.

We could on each event a demo using the `onXXX` if available. Related to https://github.com/mui-org/material-ui-x/pull/2179#discussion_r675235435

packages/grid/_modules_/grid/models/events/gridEventLookup.ts
flaviendelangle: Yes, in the lookup you have the non-wrapper version of the event
That way you can type the `publishEvent` properly.

And on the listeners you have the lookup value but wrapped in `MuiEvent`.

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
flaviendelangle: I followed the demo on the Core documentation, I'll see how to improve it

packages/grid/x-data-grid-pro/src/DataGridPro.tsx
flaviendelangle: Because in `commitPropsAndExit` you receive a `MuiBaseEvent` and you publish a `cellEditStop` with it.
It is a `MuiBaseEvent` there because the `cellFocusOut` passes a `MuiBaseEvent` when triggered by `setColumnHeaderFocus`.

I did not wanted to change the signature of `setColumnHeaderFocus` but I can if it makes sense.

What I did for those events is to find the most restrictive set of events that matches all our calls.

packages/grid/_modules_/grid/models/events/gridEventLookup.ts
m4theushw: If it's wrapped later that's okay, but on the docs it's not consistent.

image

packages/grid/_modules_/grid/hooks/features/editRows/useGridEditRows.ts
```diff
@@ -106,24 +100,28 @@ export function useGridEditRows(
...
-    commitPropsAndExit(params, nativeEvent);
-  });
+  const handleColumnHeaderDragStart: GridEventListener<GridEvents.columnHeaderDragEnter> =
+    useEventCallback(() => {
+      const { cell } = apiRef.current.state.focus;
+      if (!cell) {
+        return;
+      }
+      const params = apiRef.current.getCellParams(cell.id, cell.field);
+      commitPropsAndExit(params, {});
```
flaviendelangle, line 123: @m4theushw I have an interesting case here.

From what I understand, you mis-used `useEventCallback`. The event is on the 2nd argument since it is still one of our listeners. So we were passing the `GridColumnHeaderParams` as event to `commitPropsAndExit`.

I did a 1st fix by passing the actual event  

```tsx
    useEventCallback((dragParams, nativeEvent) => {
      const { cell } = apiRef.current.state.focus;
      if (!cell) {
        return;
      }
      const params = apiRef.current.getCellParams(cell.id, cell.field);
      commitPropsAndExit(params, nativeEvent);
```

But it seems that the event has `isPropagationStopped() === true` in some cases.
So the event in `commitPropsAndExit` is swallowed and we never leave the edit mode.

To bypass this issue I stopped passing the event to `commitPropsAndExit`, which works but is radical.

packages/grid/_modules_/grid/models/events/gridEventLookup.ts
flaviendelangle: Yes the doc should wrap them indeed

docs/scripts/buildApi.ts
flaviendelangle: What do you want me to linkify here ?
The description is already linkified.

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
flaviendelangle: Fixed

packages/grid/_modules_/grid/models/events/gridEventLookup.ts
flaviendelangle: Done

flaviendelangle:
> We could on each event a demo using the onXXX if available. Related to #2179 (comment)

I propose we do it on a follow PR
I do not have the link `eventName` => `propName` yet but it is totally doable.

packages/grid/_modules_/grid/hooks/features/editRows/useGridEditRows.ts
m4theushw: > From what I understand, you mis-used useEventCallback. The event is on the 2nd argument since it is still one of our listeners.

Yes, it was wrong.

> But it seems that the event has isPropagationStopped() === true in some cases.

Do you have an example? We should avoid stopping the propagation of events because it prevent users from listening to them. I could only find https://github.com/mui-org/material-ui-x/blob/e126eb9179c2643fb2bc98619909273e54bdccc5/packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx#L89

However, calling `event.stopPropagation()` above only stops further listeners that are binded to different events. The listener from the editing hook should still be called.

flaviendelangle: In my PR if you pass the `nativeEvent` on the `commitPropsAndExit` of line 123 like so:

```tsx
  const handleColumnHeaderDragStart: GridEventListener<GridEvents.columnHeaderDragEnter> =
    useEventCallback((_, nativeEvent) => {
      const { cell } = apiRef.current.state.focus;
      if (!cell) {
        return;
      }
      const params = apiRef.current.getCellParams(cell.id, cell.field);
      commitPropsAndExit(params, nativeEvent);
    });
```

You have `should save changes when a column header is dragged` failing
Because the cell is never leaving edit mode.

The guilty `stopPropagation` is here: https://github.com/mui-org/material-ui-x/blob/master/packages/grid/_modules_/grid/hooks/features/columnReorder/useGridColumnReorder.tsx#L89

m4theushw: Now I understood. It's calling `commitPropsAndExit` but `cellEditStop` is never published because its event has propagation stopped. Not passing an event seems to be the only alternative. In v6 we should remove the `stopPropagation` and handle the integration between React DnD differentely.

docs/scripts/api/buildInterfacesDocumentation.ts
```diff
@@ -124,18 +126,39 @@ function findProperties(reflection: TypeDoc.DeclarationReflection) {
...
+
   allEvents.forEach((event) => {
     const description = linkify(event.comment?.shortText, documentedInterfaces, 'html');
+    const eventProperties = eventLookup[event.escapedName!];
 
     events.push({
       name: event.escapedName!,
       description: renderMarkdownInline(description),
+      params: escapeHTML(linkify(eventProperties.params, documentedInterfaces, 'html')),
+      event: escapeHTML(`MuiEvent<${eventProperties.event ?? '{}'}>`),
```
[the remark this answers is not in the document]
flaviendelangle: Fixed, I did the escape on the docs instead of doing it during the build

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -10,33 +23,118 @@ const useStyles = makeStyles({
...
+    }
+    if (event.params) {
+      args.unshift(`params, // ${event.params}`);
+    }
+
+    return `
+const onEvent: GridEventListener<GridEvents.${event.name}> =
+ (
+    ${args.join('\n    ')}
+  ) => {...}
```
m4theushw, line 46: The indentation is not the same as https://deploy-preview-3022--material-ui-x.netlify.app/components/data-grid/events/#subscribing-to-events

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -10,33 +23,118 @@ const useStyles = makeStyles({
...
+            {open ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
+          </IconButton>
+        </TableCell>
+        <TableCell style={{ borderBottom: 'unset' }}>
+          <code>{event.name}</code>
+        </TableCell>
+        <TableCell style={{ borderBottom: 'unset' }}>
+          <div
+            dangerouslySetInnerHTML={{
+              __html: escapeHTML(event.description),
```
m4theushw, line 77: It's not formatting correctly.

image

flaviendelangle: Fixed

docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -10,33 +23,117 @@ const useStyles = makeStyles({
...
 
+function escapeHTML(value) {
+  return value.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+}
+
+const EventRow = ({ event }) => {
+  const [open, setOpen] = React.useState(false);
+
+  const example = React.useMemo(() => {
+    const args = ['details // GridCallbackDetails'];
```
m4theushw, line 34: When multiline, all arguments have a trailing comma.

```suggestion
    const args = ['details, // GridCallbackDetails'];
```

m4theushw (approved):

PR #3056, opened 2021-10-31
[DataGrid] Fix transitive dependency

oliviertassinari:
We have this line https://github.com/mui-org/material-ui-x/blob/3ad0e791be440de82aeadc9acd5b861887c6838a/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts#L1

and yet no direct dependency on the package. Yarn 2 would fail on a JS module (here it's a TS one, I don't really know what happens). In the future, with the objective to support multiple design systems, we will likely do:

```diff
-import { darken, lighten, alpha, styled } from '@mui/material/styles';
+import { darken, lighten, alpha, styled } from '@mui/system';
```

### DIFF
--- a/packages/grid/data-grid/package.json
+++ b/packages/grid/data-grid/package.json
@@ -37,6 +37,7 @@
     "directory": "packages/grid/data-grid"
   },
   "dependencies": {
+    "@mui/system": "^5.0.1",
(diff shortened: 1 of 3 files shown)


oliviertassinari:
No time for this

PR #3057, opened 2021-10-31
[core] Improve playground

oliviertassinari:
A new tradeoff, if you want to use the playground, you can copy the example file. I hope it will help more people 😁 

I was previously using Next.js's page over Storybook simply because I felt that I could go from a problem to trying a change out a lot faster (spoiler alert, I hate the DX in Storybook). With https://www.youtube.com/watch?v=dMBYI7pTR4Q&t=517s, this seems to be even better now.

---

On the topic of productivity, I'm upgrading my laptop with an M1 chip, it seems to be 100% faster on a single-core benchmark, even with the MacBook Air 13" of 2020 compared to my current MacBook Pro 13" of 2019.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __diff_output__
 /.nyc_output
 /coverage
 /docs/.next
+/docs/pages/playground.tsx
 /docs/export
 /test/regressions/screenshots
 build
--- a/docs/pages/playground.example.tsx
+++ b/docs/pages/playground.example.tsx
@@ -1,0 +1,10 @@
+import * as React from 'react';
+
+export default function Playground() {
+  return (
+    <div>
+      A playground for a fast iteration development cycle. You can copy this file and name it{' '}
+      <code>playground.tsx</code>
+    </div>
+  );
+}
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -10,6 +10,7 @@ __diff_output__
 /.nyc_output
 /coverage
 /docs/.next
+/docs/pages/playground.tsx
```
m4theushw, line 13: This file was deleted, no?

```suggestion
/docs/pages/playground.example.tsx
```

flaviendelangle: I think the idea is to have `playground.example.tsx` on Github
But that if you create a `playground.tsx` on your machine, it should not be committed.

m4theushw: > But that if you create a playground.tsx on your machine, it should not be committed.

Then I have to create a new file every time I need to test something. It's easier to use `docs/pages/components/data-grid/demo.js` and not commit it.

flaviendelangle: Why would you need to create a new one every time ?

m4theushw: I have an alias that does `git restore -- . && git clean -f && git checkout next && git pull` to always have the latest changes from HEAD. I was thinking that it would delete the file but it didn't so I'm okay with pushing this forward.

m4theushw (approved):

flaviendelangle (approved):

PR #3059, opened 2021-10-31
[DataGrid] Revert year change in the MIT license

oliviertassinari:
Apply https://github.com/mui-org/material-ui-x/pull/2676/files#r720686645

### DIFF
--- a/packages/grid/data-grid/LICENSE
+++ b/packages/grid/data-grid/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Material-UI SAS
+Copyright (c) 2020 Material-UI SAS
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/packages/grid/x-grid-data-generator/LICENSE
+++ b/packages/grid/x-grid-data-generator/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Material-UI SAS
+Copyright (c) 2020 Material-UI SAS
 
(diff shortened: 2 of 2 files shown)


mbrookes:
There appears to be no consensus as to whether the original year or a range of years is appropriate (the FSF supports using a range, for example), but in any case:
1. It seems better to at least use the starting year - some code will fall out of copyright earlier than the rest. It anchors the date of first creation.
2. Since by the time the copyright expires on the initial code, it will all be largely irrelevant (I hope!), so having a later date isn't important.
3. Copyright is rather meaningless with a permissive license. It's mostly about attribution if I understand correctly (the literature is weak on this point).

Worth a read, even though it doesn't address this point specifically (or arguably gets it wrong): https://writing.kemitchell.com/2016/09/21/MIT-License-Line-by-Line.html

mbrookes (approved):
There appears to be no consensus as to whether the original year or a range of years is appropriate (the FSF supports using a range, for example), but in any case:
1. It seems better to at least use the starting year - some code will fall out of copyright earlier than the rest. It anchors the date of first creation.
2. Since by the time the copyright expires on the initial code, it will all be largely irrelevant (I hope!), so having a later date isn't important.
3. Copyright is rather meaningless with a permissive license. It's mostly about attribution if I understand correctly (the literature is weak on this point).

Worth a read, even though it doesn't address this point specifically (or arguably gets it wrong): https://writing.kemitchell.com/2016/09/21/MIT-License-Line-by-Line.html

oliviertassinari:
Yeah, I think that it's a small detail. My main objective is to:

1. If there is no clear winners, then better not change it: revert
2. Match core, consistency 
3. Use the first year, no burden to update it every year.

PR #3062, opened 2021-11-01
[test] Fix flaky e2e tests

alexfauquette:
Some e2e tests fail with
```
Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

I do not see any other way than increasing the timeout parameter.


m4theushw:
We can wait a bit more before merging this one. The CI seems to be starting to have more greens than reds again.

image

PR #3070, opened 2021-11-02
[DataGrid] Fix filtering of nullish numeric cells

### LINKED ISSUE #3019: [DataGrid] Filtering for value 0 show rows that are null
Once using DataGrid filter for fields with numeric value '0', the filter just shows all values and ignores 0's.

flaviendelangle:
Fixes #3019

### DIFF
--- a/packages/grid/_modules_/grid/models/colDef/gridNumericOperators.ts
+++ b/packages/grid/_modules_/grid/models/colDef/gridNumericOperators.ts
@@ -2,6 +2,14 @@ import { GridFilterInputValue } from '../../components/panel/filterPanel/GridFil
 import { GridFilterItem } from '../gridFilterItem';
 import { GridFilterOperator } from '../gridFilterOperator';
 
+const parseNumericValue = (value: string | number | null) => {
+  if (value == null) {
+    return null;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/grid/_modules_/grid/models/colDef/gridNumericOperators.ts
```diff
@@ -2,6 +2,14 @@ import { GridFilterInputValue } from '../../components/panel/filterPanel/GridFil
 import { GridFilterItem } from '../gridFilterItem';
 import { GridFilterOperator } from '../gridFilterOperator';
 
+const parseNumericValue = (value: string | number | null) => {
+  if (value == null) {
```
alexfauquette, line 6: For my own learning: Are you using `==` to catch `null` and `undefined`?

alexfauquette (approved):

packages/grid/_modules_/grid/models/colDef/gridNumericOperators.ts
flaviendelangle: Yes, and I tested both on the test file.
It's to avoid the `Number(null) === 0`

PR #3074, opened 2021-11-02
[DataGrid] Keyboard navigation broken on page > 0

flaviendelangle:
Introduced by #2994, was never released, I will merge it before the next release if no review :+1: 

The variable wording was approximate and I did not test enough.
`rowCount` was refering to the index of the last element of the page + 1

A few examples:
- `rows.length: 100, page: 0, pageSize: 10` => `rowCount: 10`
- `rows.length: 100, page: 1, pageSize: 10` => `rowCount: 20`
- `rows.length: 14, page: 1, pageSize: 10` => `rowCount: 14`

I will probably clean those methods later to add explicit comments on each instruction. 
Because `// we go to the current row, first col, or last col!` is not very explicit to me.

### DIFF
--- a/packages/grid/_modules_/grid/utils/utils.ts
+++ b/packages/grid/_modules_/grid/utils/utils.ts
@@ -32,5 +32,5 @@ export function escapeRegExp(value: string): string {
   return value.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
 }
 
-export const clipNumber = (value: number, min: number, max: number) =>
+export const clamp = (value: number, min: number, max: number) =>
   Math.min(max, Math.max(min, value));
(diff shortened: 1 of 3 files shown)


packages/grid/_modules_/grid/utils/utils.ts
```diff
@@ -31,3 +31,6 @@ export function localStorageAvailable() {
 export function escapeRegExp(value: string): string {
   return value.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
 }
+
+export const setInInterval = (value: number, min: number, max: number) =>
```
alexfauquette, line 35: I find `set` prefix confusing because it is not a setter. `get`, or `transform` could make sense. NumPy calls this function [clip](https://numpy.org/doc/stable/reference/generated/numpy.clip.html)

alexfauquette (approved):

packages/grid/_modules_/grid/utils/utils.ts
m4theushw: A more commonly used name: `clamp`.

https://github.com/mui-org/material-ui/blob/eb01f34ed9b680f4b893cacd73b846c2284a429a/packages/mui-system/src/colorManipulator.js#L10

https://github.com/mui-org/material-ui/blob/eb01f34ed9b680f4b893cacd73b846c2284a429a/packages/mui-core/src/SliderUnstyled/SliderUnstyled.js#L25

PR #3077, opened 2021-11-02
[DataGrid] Fix selecting 100k rows takes too long

### LINKED ISSUE #3076: [DataGrid] Selecting all 100k rows takes too much time
### Duplicates

### Latest version

### Current behavior 😯

Selecting all 100k is taking too much time.

### Expected behavior 🤔

It should be faster. In https://v4.mui.com/components/data-grid/#commercial-version it's almost immediately.

### Steps to reproduce 🕹

Steps:

1. Open https://mui.com/components/data-grid/#commercial-version
2. Select all 100k rows

### Context 🔦

It started in #2781

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3076 

The logic of `currentSelectionSize` is O(n^2). Now it's O(n).

### DIFF
--- a/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
+++ b/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
@@ -52,10 +52,17 @@ const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnHeaderPa
 
     // All the rows that could be selected / unselected by toggling this checkbox
     const selectionCandidates = React.useMemo(() => {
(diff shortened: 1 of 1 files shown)


flaviendelangle:
A lot better !

flaviendelangle (approved):

oliviertassinari:
Isn't it the second time we have a performance regression on this feature? Maybe we should have a sanity performance regression test. Or add the case to the benchmark

m4theushw:
@oliviertassinari I want to do that. The unit-based performance tests are only for sorting and filtering. Also, these tests only use a small subset so a regression like #3076 wouldn't be noticed.

PR #3078, opened 2021-11-02
[DataGrid] Move column headers virtualization to hook

m4theushw:
### Breaking changes

- [DataGrid] The following CSS classes were renamed to follow the internal convention

  - `MuiDataGrid-columnsContainer` was renamed to `MuiDataGrid-columnHeaders`
  - `MuiDataGrid-columnHeaderWrapper` was renamed to `MuiDataGrid-columnHeadersInner`
  - The `scroll` class applied to `MuiDataGrid-columnHeaderWrapper` was renamed to `MuiDataGrid-columnHeadersInner--scrollable`

---

Part of #2946

The column pinning is a Pro feature, which means that code related to it shouldn't be shipped as MIT. For this feature, the column headers will have to work as it works now but it also needs to render two additional groups of columns pinned to left or right. The main goal of this PR then is to move the common logic responsible for the column headers to a separate hook. This way, we can create a Pro's column headers component with the part that renders the pinned columns added.

To summarize, this PR does:
1. Moves the logic inside `GridColumnsHeader` to `useGridColumnHeaders`.
2. Normalizes component names: `GridColumnsHeader` -> `GridColumnHeaders`.
3. Normalizes class names to follow the component tree.
4. Adds `DataGridColumnHeaders` which renders the column headers in the MIT version.
5. Adds `DataGridProColumnHeaders` which renders the column headers in the Pro version. In #2946 it will also render the pinned columns so I'm already creating it here.
6. Incorporates part of #3012


flaviendelangle:
I suppose this PR have to be merged before the stable release since it is breaking.

m4theushw:
@flaviendelangle Yes, this is the only breaking change required for the column pinning.

flaviendelangle (approved):

PR #3079, opened 2021-11-03
[DataGrid] Move virtualization logic to hook

m4theushw:
Part of #2946 

The same idea of #3078 but with the main virtualization.


packages/grid/_modules_/grid/components/virtualization/GridVirtualScrollerContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+    root: ['virtualScrollerContent'],
+  };
+
+  return composeClasses(slots, getDataGridUtilityClass, classes);
+};
+
+const VirtualScrollerContentRoot = styled('div', {
+  name: 'MuiDataGrid',
+  slot: 'VirtualScrollerContent',
+  overridesResolver: (props, styles) => styles.virtualScrollerContent,
```
m4theushw, line 24: On HEAD:
https://github.com/mui-org/material-ui-x/blob/cca95ce54d26ec5a4c5d491daf069212480010b9/packages/grid/_modules_/grid/components/GridVirtualScroller.tsx#L55 

This doesn't match the docs:

image

flaviendelangle:
Do we agree that it is only a copy / paste on the logic and 2-3 function wrappers ?
Le global logic seems good to me.
I'm just not a huge fan of creating function like `getRootProps` if they just return variables that are computed during the render anyway.

m4theushw:
> Do we agree that it is only a copy / paste on the logic and 2-3 function wrappers ?

@flaviendelangle Yes, the code of the hook is a literal copy and paste.

> I'm just not a huge fan of creating function like getRootProps if they just return variables that are computed during the render anyway. I don't see the gain compared to simplify return rootProps.

This is a pattern heavily used in the core for unstyled components: https://github.com/mui-org/material-ui/blob/eb01f34ed9b680f4b893cacd73b846c2284a429a/packages/mui-core/src/AutocompleteUnstyled/useAutocomplete.js#L1028

Downshift and other libs also used it: https://github.com/downshift-js/downshift#prop-getters

For now, it could return a normal object. However, returning a function could be useful to sanitize the props that are passed to the component.

flaviendelangle (approved):

PR #3080, opened 2021-11-03
[DataGrid] Improve the filter panel behaviors

### LINKED ISSUE #1258: [DataGrid] Filter Column does not change on choosing filter from new column
When an initial filter is set, clicking on the filter column menu of a different column shows the filter popup but with the previously selected column, not the new column.

### LINKED ISSUE #1777: [DataGrid] The column menu filtering UX is K.O.

## Current Behavior 😯

When filtering multiple columns, if I click in the Filter button of a column that has an active filter an empty filter is added. It doesn't occur in the column that is the last in the filter list.

duplicate-filter

## Expected Behavior 🤔

It shouldn't add an filter if there's already one for the same field. 

## Steps to Reproduce 🕹

Steps:

1. Navigate to https://material-ui.com/components/data-grid/filtering/#multi-column-filtering
2. Open the menu of the Commodity column.
3. Click in the Filter button

I suppose the fix is simple. We have to change the logic to check if there's a filter added for the same field in the whole list and not just the last item:

https://github.com/mui-org/material-ui-x/blob/0cd94899863a2827ee31718105733fbca63b841a/packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts#L216-L224

flaviendelangle:
Closes #1777
Closes #1258

[Docs](https://deploy-preview-3080--material-ui-x.netlify.app/components/data-grid/filtering/)

### Limit the amount of `setGridState` in `useGridFilter`

- Use `apiRef.current.setFilterModel` in `apiRef.current.deleteFilterItem`, `apiRef.current.upsertFilterItem` and `apiRef.current.setFilterLinkOperator`

- Do not re-render when deleting a non existing item or setting the link operator to the same value as before

### Do not store in state the empty filter line displayed in the filter panel

This has several benefics

- No flickering when deleting the last filter item, the placeholder is not generated in a `useEffect` on the panel but directly in a `useMemo`

- No filter item in the state without a `columnField`

- No call on `onFilterModelChange` for those incomplete filter item,


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -73,20 +74,20 @@ function GridFilterForm(props: GridFilterFormProps) {
...
-  const getCurrentColumn = React.useCallback(() => {
+  const currentColumn = React.useMemo<GridStateColDef | null>(() => {
     if (!item.columnField) {
       return null;
     }
     return apiRef.current.getColumn(item.columnField)!;
   }, [apiRef, item]);
 
-  const getCurrentOperator = React.useCallback(() => {
-    const currentColumn = getCurrentColumn();
```
DanailH, line 84: Am I missing something ... `currentColumn` is now a function right? How is it being called to get the target column `ColDef`?

flaviendelangle:
I would be interested in functional test in the UI to see if their are scenarios where you find the new behaviors non-intuitive.
#1777 showed that it is a subjective topic but I would like the new approach to be at least better in all scenarios if not perfect.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -73,20 +74,20 @@ function GridFilterForm(props: GridFilterFormProps) {
   const ownerState = { classes: rootProps.classes };
   const classes = useUtilityClasses(ownerState);
 
-  const getCurrentColumn = React.useCallback(() => {
+  const currentColumn = React.useMemo<GridStateColDef | null>(() => {
```
DanailH, line 77: I'm not sure we actually need to memoize that since where it is being called it is already is memoized and they both depend on `item`.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
```diff
@@ -43,16 +72,12 @@ export function GridFilterPanel() {
     [apiRef],
   );
 
-  React.useEffect(() => {
-    if (filterModel.items.length === 0) {
-      addNewFilter();
-    }
-  }, [addNewFilter, filterModel.items.length]);
+  const hasMultipleFilters = items.length > 1;
```
DanailH, line 75: Can we organize it a bit better - maybe put this line after the `items` useMemo since this is on what it depends. The others are useCallbacks or just normal arrow functions.

alexfauquette (approved):

DanailH (approved):

PR #3081, opened 2021-11-03
[docs] Fix inline previews

### LINKED ISSUE #2990: [docs] Inline previews are no longer working in `DataGrid` pages
### Duplicates

### Latest version

### Current behavior 😯

Compare https://v4.mui.com/components/data-grid/editing/#control-cell-editability with https://mui.com/components/data-grid/editing/#control-cell-editability

### Expected behavior 🤔

Previews shod work the same in v5 as in v4.

DanailH:
Fixes #2990 

Based on https://github.com/mui-org/material-ui/pull/28215

The diff is so big because the previews need to be generated.

Preview -> https://deploy-preview-3081--material-ui-x.netlify.app/components/data-grid/columns/#column-definitions

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -13,8 +13,8 @@
     "export": "rimraf docs/export && next export --threads=3 -o export && yarn build-sw",
     "icons": "rimraf public/static/icons/* && node ./scripts/buildIcons.js",
     "typescript": "tsc -p tsconfig.json",
-    "typescript:transpile": "node scripts/formattedTSDemos",
-    "typescript:transpile:dev": "node scripts/formattedTSDemos --watch"
+    "typescript:transpile": "cross-env BABEL_ENV=development babel-node --extensions \".tsx,.ts,.js\" scripts/formattedTSDemos",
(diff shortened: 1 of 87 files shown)


docs/package.json
```diff
@@ -13,8 +13,8 @@
     "export": "rimraf docs/export && next export --threads=3 -o export && yarn build-sw",
     "icons": "rimraf public/static/icons/* && node ./scripts/buildIcons.js",
     "typescript": "tsc -p tsconfig.json",
-    "typescript:transpile": "node scripts/formattedTSDemos",
-    "typescript:transpile:dev": "node scripts/formattedTSDemos --watch"
+    "typescript:transpile": "cross-env BABEL_ENV=development babel-node --extensions \".tsx,.ts,.js\" scripts/formattedTSDemos",
```
eps1lon, line 16: Why do you need to switch to `babel-node` here? I would recommend to stick to just node until it becomes absolutely necessary.

eps1lon (approved):

alexfauquette (approved):
I tested some pages on preview, seems to work well :)

docs/package.json
DanailH: If I leave it as it was the script fails. I checked how it's done in the core and copied the approach.

PR #3088, opened 2021-11-04
[core] Rename directories to match new packages new names

flaviendelangle:
I also tested the `release:build` script and the output seems fine


DanailH (approved):
Resolving conflicts after this is merged will be interesting.

flaviendelangle:
@DanailH As long as you don't create a file it is fine from what I saw since I opened this PR.
But yeah, created files are not handled correctly, we will have to move them manually from the old directory to the new one.

.codesandbox/ci.json
```diff
@@ -2,16 +2,16 @@
   "buildCommand": "build:codesandbox",
   "node": "12",
   "packages": [
-    "packages/x-license",
-    "packages/grid/data-grid",
-    "packages/grid/x-grid",
-    "packages/grid/x-grid-data-generator"
+    "packages/x-license-pro",
+    "packages/grid/x-data-grid",
```
oliviertassinari, line 6: Killing `/grid/` next would be 👌 . As a side note, in Core, we include the full name of the npm package: (adding `mui-)

PR #3089, opened 2021-11-04
v5.0.0-beta.7

alexfauquette:

PR #3090, opened 2021-11-04
[core] Use 'unstable_' prefix instead of 'unsafe_'

flaviendelangle:
This wrong naming was introduced in https://github.com/mui-org/material-ui-x/pull/2985

### DIFF
--- a/packages/grid/_modules_/grid/hooks/core/useGridControlState.ts
+++ b/packages/grid/_modules_/grid/hooks/core/useGridControlState.ts
@@ -9,7 +9,7 @@ import { useGridApiMethod } from '../utils/useGridApiMethod';
 export function useGridControlState(apiRef: GridApiRef, props: GridComponentProps) {
   const controlStateMapRef = React.useRef<Record<string, GridControlStateItem<any>>>({});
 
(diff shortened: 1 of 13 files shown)


alexfauquette:
# Breaking change

- `unsafe_updateControlState` rename to `unstable_updateControlState`
- `unsafe_applyControlStateConstraint` rename to `unstable_applyControlStateConstraint`
- `unsafe_copySelectedRowsToClipboard` rename to `unstable_copySelectedRowsToClipboard`
- `unsafe_applyFilters` rename to `unstable_applyFilters`

flaviendelangle:
@alexfauquette for me it is not a breaking change because it was not in the public API.

m4theushw:
Do we really need this for the release? These methods were already not documented.

flaviendelangle:
@m4theushw as you want, as it is private and not in the doc, I don't think it matters a lot to include it or not.

m4theushw (approved):

m4theushw:
@flaviendelangle I would merge after the release to avoid having to update #3089 to include this.

oliviertassinari:
Nice, happy to see the same convention used as in React and MUI Core :)

PR #3092, opened 2021-11-04
[core] Make useGridColumnsPreProcessing generic

m4theushw:
On the column pinning, the calculation of the position to scroll in `scrollToIndexes` is [different](https://github.com/mui-org/material-ui-x/pull/2946/files#diff-719fae9618e984ed1754b665caa6a98d8e6cd2bc8bf81a5a3adc7490a13e2f98R16). It has to take into account that the pinned columns might be covering the cell. To not have to create a separate `useGridScroll` only for DataGridPro I thought about reusing the idea of `useGridColumnsPreProcessing`  to allow other hooks to have the chance to pre-processes a value before it's used. This hook is implementing the pipe-filter design pattern and it can be used for other things too. This PR reshapes it to allow to group the pre-processors. I also changed to unregister the pre-processor on the cleanup. This removes the need to name every pre-processor, we only need to provide its group. If we were to keep the old approach we would need to provide a name for the group and another name for the pre-processor.

Another interesting use for this hook will be to allow other hooks to register buttons on the column menu or toolbar. I think this API once public will be more powerful than the events API.

For a future PR: the logic of registering during the first render and re-registering on subsequent re-renders could be abstracted to a hook.


flaviendelangle:
For me the cleanup function is weird and I did not put one because of it.
In the cleanup function, you change the list of the pre-processing but you do not fire the event.

The following example feels right and would probably be what I would implement if I did not look at the internals of `unstable_registerPreProcessor`.
But it would be buggy, since when switching `prop.someFeature` from true to false, it would not call any event and therefore not remove the column from the list.

```tsx
React.useEffect(() => {
  if (props.someFeature) {
    return apiRef.current.unstable_registerPreProcessor(
      GridPreProcessingGroup.hydrateColumns,
      addSomeColumn,
    );
  }
});
```

And if you do fire the event, you run the pre-processing application twice every time, which is really not acceptable, so I am not proposing to fire the event.

I understand the pain of passing a name argument.
But I think it's better to avoid risking this unwanted behavior.

m4theushw:
> The following example feels right and would probably be what I would implement if I did not look at the internals of unstable_registerPreProcessor.

Yeah, I didn't notice that. I tested your example and although it shouldn't, it works correctly. Changing `checkboxSelection` to false removed the column. There's a bug here which causes the strange behavior. The following `useEffect` is firing always so it re-applies the pre-processors again:

https://github.com/mui-org/material-ui-x/blob/fbdea4e152a677d713213bc8d3912218e8773dd9/packages/grid/_modules_/grid/hooks/features/columns/useGridColumns.ts#L326

> And if you do fire the event, you run the pre-processing application twice every time, which is really not acceptable, so I am not proposing to fire the event.

I don't find it so unacceptable. This is how React works, it calls the cleanup even when it's going to call the effect again. https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

> But I think it's better to avoid risking this unwanted behavior.

Not unregistering the pre-processor once it's not more useful (e.g. `checkboxSelection` becomes `false`) might also be an unwanted behavior.

flaviendelangle:
>  I tested your example and although it shouldn't, it works correctly

I would not be surprise if it's caused by the multiple re-renders we currently have and that it could break in the future (after #3007 for instance).

> I don't find it so unacceptable. This is how React works, it calls the cleanup even when it's going to call the effect again.

My problem is if it causes a double re-render due to the double `setGridState`.

> Well, if we're going to name the callback let's at least use a random ID and hide this inside a hook.

That could be a solution.

> Not unregistering the pre-processor once it's not more useful (e.g. checkboxSelection becomes false) might also be an unwanted behavior. 

This is why on my 1st version the pre-processor could be null.
That way you were doing :

```ts
React.useEffect(() => {
  if (props.someFeature) {
    apiRef.current.unstable_registerPreProcessor(
      GridPreProcessingGroup.hydrateColumns,
      addSomeColumn,
    );
  } else {
    apiRef.current.unstable_registerPreProcessor(
      GridPreProcessingGroup.hydrateColumns,
      null,
    );  }
}));
```


But we changed it for an alternative where it is `addSomeColumn` which checks `props.someFeature`.

m4theushw:
I went one step back and re-added the argument to name the pre-processors. However, the cleanup function is still returned but not used for the column hydration. For the column pinning, I don't need to run the pre-processors when something changes, only when the user triggers an action, so I'm going to use the cleanup to unregister the callback once no column is pinned.

packages/grid/_modules_/grid/hooks/core/preProcessing/useGridPreProcessing.ts
```diff
@@ -0,0 +1,56 @@
...
+        const { [id]: removedPreProcessor, ...otherProcessors } = preProcessorsRef.current[group]!;
+        preProcessorsRef.current[group] = otherProcessors;
+        apiRef.current.publishEvent(GridEvents.preProcessorUnregister, group);
+      };
+    },
+    [apiRef],
+  );
+
+  const applyPreProcessors = React.useCallback<GridPreProcessingApi['unstable_applyPreProcessors']>(
+    (name, value, params) => {
```
flaviendelangle, line 37:
```suggestion
    (group, value, params) => {
```

flaviendelangle (approved):

PR #3094, opened 2021-11-05
[core] Remove imports from `@mui/material`

flaviendelangle:
Fixes https://github.com/mui-org/material-ui-x/pull/3012#discussion_r743052273

### DIFF
--- a/packages/grid/_modules_/grid/gridClasses.ts
+++ b/packages/grid/_modules_/grid/gridClasses.ts
@@ -1,4 +1,4 @@
-import { generateUtilityClasses, generateUtilityClass } from '@mui/material';
+import { generateUtilityClasses, generateUtilityClass } from '@mui/material/utils';
 
 export interface GridClasses {
   /**
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { unstable_composeClasses as composeClasses } from '@mui/material';
+import { unstable_composeClasses as composeClasses } from '@mui/material/utils';
 import { GridEvents } from '../../../constants/eventsConstants';
 import { GridComponentProps } from '../../../GridComponentProps';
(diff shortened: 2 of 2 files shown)


DanailH:
The reason why I imported it from `@mui/material` is because if you do it from `@mui/material/utils` it complains that there is no exported method.
Also `@mui/material` is used on a lot of other places to import either `unstable_composeClasses` or `generateUtilityClasses`.

flaviendelangle:
@DanailH I saw that after creating the PR. 
I asked Olivier if he add something in mind because as it it I don't see how to avoid the root import.

m4theushw:
You can import `generateUtilityClass` from `@mui/core`:

https://github.com/mui-org/material-ui/blob/4a67b3ce32a1a6e89d1c36b2781db460159c9f57/packages/mui-material/src/Accordion/accordionClasses.ts#L1

oliviertassinari:
Regarding why it matters, I believe it's https://github.com/mui-org/material-ui/pull/24147.

Regarding how to fix it. I imagine we could either

1. add `@mui/core` as a dependency. 
or 
2. expose the module from `@mui/material/utils`. 

However, it won't truly solve the root cause. We are increasingly adding new modules to `@mui/core`. cc @michaldudak. I assume that ideally, we could do:

```jsx
import composeClasses from '@mui/core/composeClasses';
// or
import { composeClasses } from '@mui/core/utils';
```

DanailH:
> Regarding how to fix it. I imagine we could either
> 
> 1. add `@mui/core` as a dependency.
>    or
> 2. expose the module from `@mui/material/utils`.

Maybe I misunderstood but there is a discussion in the process to rename `@mui/core` right? If that is true and we go with option 1. then we need to change our dep. and that will be a breaking change.

oliviertassinari:
https://github.com/mui-org/material-ui/pull/29585

m4theushw:
Either option 1 or 2 will be a breaking change for us. Although it might not be for MUI Core. If we add `@mui/base` as dependency, then DataGrid can only be used with MUI Core 5.1.x. I think this change will have to be postponed to v6 or v5.1.

flaviendelangle:
I'm closing this PR since we don't have a quick solution

PR #3095, opened 2021-11-05
[DataGrid] Block multi-rows updates in `apiRef.current.updateRows` on the free plan

flaviendelangle:
Preparing for the unlocking of the `apiRef` on the free plan.

I did not create a `disableMultiRowUpdates` prop with a forced value on the `DataGrid` like for `disableMultipleColumnsFiltering` because I don't see any situation where a user of the pro version would want to limit the updates to one row.
So for me it would only flood the prop list.

We could also add the prop only on `GridComponentProps` and set a forced value at `true` on `DataGrid` and `false` on `DataGridPro`.
But I think checking directly the signature is simpler.

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
@@ -238,7 +238,7 @@ export const useGridRows = (
         // TODO: Add test with direct call to `apiRef.current.updateRows` in DataGrid after enabling the `apiRef` on the free plan.
         throw new Error(
           [
-            "MUI: You can't update several rows at once in `apiRef.current.updateRows` on the free plan",
+            "MUI: You can't update several rows at once in `apiRef.current.updateRows` on the DataGrid.",
             'You need to upgrade to the DataGridPro component to unlock this feature.',
           ].join('\n'),
         );
--- a/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
+++ b/packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
@@ -297,7 +297,7 @@ export const useGridRows = (
 
       throttledRowsChange(state, true);
     },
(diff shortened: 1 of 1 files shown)


DanailH (approved):

packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
```diff
@@ -231,6 +234,16 @@ export const useGridRows = (
 
   const updateRows = React.useCallback<GridRowApi['updateRows']>(
     (updates) => {
+      if (props.signature === GridSignature.DataGrid && updates.length > 1) {
+        // TODO: Add test with direct call to `apiRef.current.updateRows` in DataGrid after enabling the `apiRef` on the free plan.
+        throw new Error(
+          [
+            "MUI: You can't update several rows at once in `apiRef.current.updateRows` on the free plan",
```
m4theushw, line 241:
```suggestion
            "MUI: You can't update several rows at once in `apiRef.current.updateRows` on the DataGrid.",
```

packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
```diff
@@ -231,6 +234,16 @@ export const useGridRows = (
...
+            'You need to upgrade to the DataGridPro component to unlock this feature.',
```
m4theushw, line 242:
```suggestion
            'You need to upgrade to the DataGridPro to unlock this feature.',
```

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
flaviendelangle: We use the `DataGridPro component` on the other error messages (on the `chainPropTypes`)

m4theushw: Ok, fine. +1 for keeping as it is.

PR #3123, opened 2021-11-08
[docs] Fix example in 'data-grid'

Himself65:


flaviendelangle:
Thanks for your contribution.
I think we should rather use `paginationState` for the row count.

flaviendelangle (approved):

PR #3124, opened 2021-11-08
[docs] Row and cell editing new examples for server-side persistence

### LINKED ISSUE #3101: [DataGrid] Add documentation on how to persist edited cells/rows
### Duplicates

### Latest version

### Summary 💡

The nominal use case of the data grid edit feature is to modify the data and then to **persists** the changes in a DB. And, yet, none of the demos in https://mui.com/components/data-grid/editing/ clearly describes how to do this. This seems to be a major opportunity.

### Examples 🌈

@m4theushw Put a demo together: https://codesandbox.io/s/basiceditinggrid-material-demo-forked-m140n?file=/demo.tsx

### Motivation 🔦

This started from the email Tyler McKee from United Airlines sent us:

> A large use case we have is the need to update a row or cell and have it send a request to the backend. While playing around with this I am having trouble figuring out a way to do this that doesn't seem overly complicated. Is there a demonstration of this or just simply updating state somewhere in the docs? I can't justify this purchase without that capability.

I couldn't help myself to think that the large majority of the developers of the people will face the same pain. For the next steps, here is one proposal:

- [ ] 1. We add a "### Server-side persistence" just before https://mui.com/components/data-grid/editing/#start-editing
- [ ] 2. We add a "### Server-side persistence" just before https://mui.com/components/data-grid/editing/#start-editing-2.

---

And to seriously consider:

- [ ] 3. We change the API of `onRowEditCommit(id :string, …)`. Is only returning the id of the row good enough? It forces the MIT users to upgrade to the pro plan, this is not what we advertise: we say that row edit is MIT, but it's not really. I believe the API of `onCellEditCommit(params: GridCellEditCommitParams, …)` is a lot better in this regard. Developers can persist the new cell's value without any apiRef call. Now, we also have [a discussion](https://www.notion.so/mui-org/Free-version-Pro-version-scopes-c81b2e6fee4d404880fad5bd25836739#2ea50470c6b041e7a2f8ed4827b6f54a) about moving some of the apiRef's methods to the public space, so maybe this will solve the problem.
- [ ] 4. We consider the use case of a new `editMode:? "server" | "client"`. Currently, the changes are optimistically shown as persisted in the database. But what if a developer wants to opt out of the optimistic update? Then I believe an edit more, as we do with [`paginationMode`](https://mui.com/components/data-grid/pagination/#server-side-pagination), `sortingMode`, etc. would help.
- [ ] 5. We rename the current `editMode` prop to something else, it could be confusing for developers that are using `paginationMode`, `sortingMode`, etc.

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Closes #3101


docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
```diff
@@ -0,0 +1,139 @@
...
+      new Promise((resolve) =>
+        setTimeout(() => {
+          resolve(user);
+        }, 200),
+      ),
+    [],
+  );
+};
+
+export default function CellEditServiceSidePersistence() {
```
m4theushw, line 38:
```suggestion
export default function CellEditServerSidePersistence() {
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -17,6 +17,14 @@ Set the `editable` property in the `GridColDef` object to `true` to allow editin
 
 {{"demo": "pages/components/data-grid/editing/BasicEditingGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
+### Service-side persistence [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
```
m4theushw, line 20:
```suggestion
### Server-side persistence [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -17,6 +17,14 @@ Set the `editable` property in the `GridColDef` object to `true` to allow editin
...
+
+If you want to send the edited data to your server, you can use the `onCellEditCommit` which is fired just before committing the new row value to the grid.
```
m4theushw, line 22:
```suggestion
If you want to send the updated data to your server, you can use the `onCellEditCommit` which is fired just before committing the new cell value to the grid.
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -17,6 +17,14 @@ Set the `editable` property in the `GridColDef` object to `true` to allow editin
...
+
+You can then decide if you want to send the whole row or manually check the modified fields and only send those.
```
m4theushw, line 24:
```suggestion
You can then decide if you want to send the whole row or only the modified fields.
```

docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
```diff
@@ -0,0 +1,139 @@
...
+    [apiRef, mutateRow],
+  );
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGridPro
+        apiRef={apiRef}
+        rows={rows}
+        columns={columns}
+        editMode="row"
```
m4theushw, line 72: This is cell editing.

```suggestion
```

docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
```diff
@@ -0,0 +1,139 @@
...
+    AlertProps,
+    'children' | 'severity'
+  > | null>(null);
+
+  const handleCloseSnackbar = () => setSnackbar(null);
+
+  const handleCellEditCommit = React.useCallback(
+    async (params: GridCellEditCommitParams) => {
+      // Get the row old value before committing
+      const oldRow = apiRef.current.getRow(params.id)!;
```
m4theushw, line 51: You can get the row from the `rows` prop and make this demo Community-compatible.

docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+      try {
+        // Make the HTTP request to save in the backend
+        await mutateRow({ id: params.id, [params.field]: params.value });
+        setSnackbar({ children: 'User successfully saved', severity: 'success' });
+      } catch (error) {
+        setSnackbar({ children: 'Error while saving user', severity: 'error' });
+        // Restore the row in case of error
+        apiRef.current.updateRows([oldRow]);
```
m4theushw, line 60: The `rows` prop can be updated instead.

docs/src/pages/components/data-grid/editing/RowEditServerSidePersistence.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const handleRowEditCommit = React.useCallback(
+    async (id: GridRowId) => {
+      const model = apiRef.current.getEditRowsModel(); // This object contains all rows that are being edited
+      const newRow = model[id]; // The data that will be committed
+
+      // The new value entered
+      const name = newRow.name.value as string;
+      const age = newRow.age.value as number;
+      const lastLogin = newRow.age.lastLogin as Date;
+      const dateCreated = newRow.age.dateCreated as Date;
```
m4theushw, line 56:
```suggestion
      const lastLogin = newRow.lastLogin.value;
      const dateCreated = newRow.dateCreated.value;
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -142,6 +150,14 @@ To enable it, change the edit mode to `"row"` using the `editMode` prop, then se
 
 {{"demo": "pages/components/data-grid/editing/BasicRowEditingGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
+### Service-side persistence [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+If you want to send the edited data to your server, you can use the `onRowEditCommit` which is fired just before committing the new row value to the grid.
```
m4theushw, line 155:
```suggestion
If you want to send the updated data to the server, you can use the `onRowEditCommit` prop which is fired just before committing the new row value to the grid.
To access the new values for the row, use `apiRef.current.getEditRowsModel` to get all rows in edit mode, then use the id provided to get only the values for the row that was committed.
```

flaviendelangle:
OK I was way to short on that one sorry, I'll fix those

docs/src/pages/components/data-grid/editing/editing.md
flaviendelangle: If I stop using the `apiRef` in the demo, then we should probably add a "If you are using the pro plan" here.

docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
flaviendelangle: Done

docs/src/pages/components/data-grid/editing/CellEditServerSidePersistence.tsx
flaviendelangle: Done

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -155,6 +163,16 @@ To enable it, change the edit mode to `"row"` using the `editMode` prop, then se
 
 {{"demo": "pages/components/data-grid/editing/BasicRowEditingGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
+### Server-side persistence [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+If you want to send the updated data to your server, you can use the `onRowEditCommit` which is fired just before committing the new cell value to the grid.
+
+To access the new values for the row, use `apiRef.current.getEditRowsModel` to get all rows in edit mode, then use the id provided to get only the values for the row that was committed.
+
+You can then decide if you want to send the whole row or only the modified fields.
```
m4theushw, line 172:
```suggestion
You can then decide if you want to send the whole row or only the modified fields, by checking it against the previous row values.
```

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -17,6 +17,14 @@ Set the `editable` property in the `GridColDef` object to `true` to allow editin
 
 {{"demo": "pages/components/data-grid/editing/BasicEditingGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
+### Server-side persistence
```
m4theushw, line 20: It should not be the first section. The user has just opened the editing page to learn how it works. I think it could be below "Server-side validation".

image

flaviendelangle: Done

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -219,6 +227,16 @@ Saving columns that make use of `valueGetter` can be done adding a `valueSetter`
 The same [approach](/components/data-grid/editing/#saving-nested-structures) from the cell editing mode can be used here.
 Note that the `valueSetter` will be called for each field.
 
+### Server-side persistence [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+If you want to send the updated data to your server, you can use the `onRowEditCommit` which is fired just before committing the new cell value to the grid.
+
+To access the new values for the row, use `apiRef.current.getEditRowsModel` to get all rows in edit mode, then use the id provided to get only the values for the row that was committed.
+
+You can then decide if you want to send the whole row or only the modified fields, by checking it against the previous row values.
```
m4theushw, line 236:
```suggestion
You can then decide if you want to send the whole row or only the modified fields, by checking them against the previous row values.
```

m4theushw (approved):

PR #3145, opened 2021-11-09
[test] Use interval for `waitFor` in test:e2e

m4theushw:
Another attempt to fix `test:e2e`.


alexfauquette (approved):
From what I understood, you try to pass the test each 50ms up to success or 1s of failing

If it does not succeed to solve the problem, `lastError` can be replaced by `errorStack` in which you push all the errors obtained to check if it evolves with time or not.

m4theushw:
@alexfauquette I tried to record a video (#3026) of the test running in CircleCI. I know that it works correctly, but for some reason only on this environment it fails to assert.

PR #3146, opened 2021-11-09
[DataGrid] Use color-scheme to set dark mode on native components

### LINKED ISSUE #1805: [DataGrid] Remove custom dark mode scrollbar in v5

## Summary 💡

The current custom dark mode scrollbar implementation suffers from the same issue as https://github.com/mui-org/material-ui/issues/23520. It forces a width on the scrollbar on macOS, it also doesn't respect the native dark mode style. 
A better solution was found in https://github.com/mui-org/material-ui/issues/25016#issuecomment-849659928.

## Motivation 🔦

We should remove:

https://github.com/mui-org/material-ui-x/blob/40ebf9b4079dd7433200b7de00b6959fd25a50ec/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts#L353-L378

once https://github.com/mui-org/material-ui/issues/25016 is fixed and we have updated the repo to dependant on v5. 

See #1613 for a related issue.

alexfauquette:
Fixes #1805

Instead of modifying the style of the scroll bar, we set `color-scheme` property on the root div according to the palette mode.

### Doc modification

- For users with @mui/material (v5.1.0)
	They have to remove the `darkScrollbar()` from their theme and use `enableColorScheme` prop. The  `color-scheme` property will be defined twice (once in the `<html>` and once in the `DataGrid` root `<div>`) but It is not a problem.

- For user bellow @mui/material (v5.1.0)
	I'm not sure. They can avoid using `darkScrollbar()` in the app style to avoid overriding the native scrollbar style. But this would lead to a page scrollbar unadapted. I'm wondering if they could not simply use
```jsx
MuiCssBaseline: {
	styleOverrides: {
		html: {
			colorScheme: theme.palette.mode 
		},
	},
},
```
instead of
```jsx
MuiCssBaseline: {
	styleOverrides: {
		body: theme.palette.mode === 'dark' ? darkScrollbar() : null,
	},
},
```

### DIFF
--- a/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
+++ b/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
@@ -50,7 +50,7 @@ export const GridRootStyles = styled('div', {
       ? lighten(alpha(theme.palette.divider, 1), 0.88)
       : darken(alpha(theme.palette.divider, 1), 0.68);
 
-  let gridStyle: CSSInterpolation = {
+  const gridStyle: CSSInterpolation = {
     flex: 1,
     boxSizing: 'border-box',
     position: 'relative',
@@ -62,6 +62,7 @@ export const GridRootStyles = styled('div', {
     height: '100%',
     display: 'flex',
     flexDirection: 'column',
+    colorScheme: theme.palette.mode,
     [`&.${gridClasses.autoHeight}`]: {
       height: 'auto',
     },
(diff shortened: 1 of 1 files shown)


packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -62,6 +62,7 @@ export const GridRootStyles = styled('div', {
     height: '100%',
     display: 'flex',
     flexDirection: 'column',
+    colorScheme: theme.palette.mode,
```
m4theushw, line 65: We can leave a note for developers to upgrade to MUI Core 5.1 or use https://mui.com/components/css-baseline/#scrollbars

```suggestion
```

m4theushw, line 65: We can leave a note for developers to upgrade to MUI Core 5.1 or use https://mui.com/components/css-baseline/#scrollbars

```suggestion
```

alexfauquette: Is it possible that DataGrid is used for itself (not in MUI Core)? If yes, we should provide an explanation of how to set dark mode for DataGrid itself.

DanailH: I guess you meant `@mui/material`? It's a peer dep so I think it's ok to have a case like that.

m4theushw: > Is it possible that DataGrid is used for itself (not in MUI Core)?

Yes, but `@mui/material` will always be installed. What changes is that they may not add the `CssBaseline`. In this case, other parts of the their applications will also look wrong without setting `color-scheme`. Overriding the color scheme only for the DataGrid

m4theushw (approved):

Primajin:
Great! Could you provide a screenshot from OSX and Windows what to expect how it will look like? That'd be great 👍🏻

alexfauquette:
@Primajin You can see how it will look like [here](https://deploy-preview-3146--material-ui-x.netlify.app/components/data-grid/demo/)

Primajin:
@alexfauquette Ohh perfect - this looks way more natural now 😍

PR #3167, opened 2021-11-12
[docs] move sentence about disabling multi rows selection

alexfauquette:
Some explanations in the doc seems to be wrong/out dated

- Disabling multi rows selection is not necessary to unselect row with ctrl key
- To select a range of rows, "Shift + Arrow Up/Down" is enough. No need of "Space"
- Some key for sorting columns was in the table about "Selection" instead of "Sorting"

### DIFF
--- a/docs/src/pages/components/data-grid/selection/selection.md
+++ b/docs/src/pages/components/data-grid/selection/selection.md
@@ -28,7 +28,6 @@ Row selection can be performed with a simple mouse click, or using the [keyboard
 
 Single row selection is enable by default with the `DataGrid` component.
 To unselect a row, hold the <kbd class="key">CTRL</kbd> key and click on it.
-For the `DataGridPro`, you need to disable multiple row selection with `disableMultipleSelection={true}`.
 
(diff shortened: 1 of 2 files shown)


flaviendelangle:
I will have a look
I worked a lot on the selection in august / september and maybe I unlocked somethings for the community plan by error :+1:

docs/src/pages/components/data-grid/selection/selection.md
```diff
@@ -38,6 +37,7 @@ On the `DataGridPro` component, you can select multiple rows in two ways:
 
 - To select multiple independent rows, hold the <kbd class="key">CTRL</kbd> key while selecting rows.
 - To select a range of rows, hold the <kbd class="key">SHIFT</kbd> key while selecting rows.
+- To disable multiple row selection use `disableMultipleSelection={true}`.
```
flaviendelangle, line 40:
```suggestion
- To disable multiple row selection, use `disableMultipleSelection={true}`.
```

flaviendelangle (approved):

Issue #3093, opened 2021-11-05
[DataGridPro] Expanding a Row


lebaphi:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

I see we have an example of the Sticky Header, Filter and Sorting here https://codesandbox.io/s/datatable-material-demo-forked-2zw8u?file=/demo.tsx:1452-1460
And an example with the expanding row at https://codesandbox.io/s/6plud?file=/demo.tsx:4765-4844.
So I have a demand to combine these feature together in 1 table. Can we do that?

flaviendelangle:
Your 2nd example is using the `Table` component not the `DataTable`.
But in the `DataTable`, this feature is not available yet. It is called #211 and should be developed this quarter.

m4theushw:
Please subscribe to #211 to be notified when it gets released.

I'll close this one.

PR #3187, opened 2021-11-15
[core] Improve testing of the keyboard navigation

flaviendelangle:
Before opening my PR about the keyboard navigation improvements, I tried to improve the tests.

- [x] `getActiveColumnHeader` should be 0-based for consistency with the rest
- [x] All the keyboard navigation keys should be correctly tested with filters, on page > 0 and when reaching the limits (ArrowDown on last row, ArrowUp on 1st row, ArrowLeft on 1st col, etc...)

If you want to take a look at the rework: #3093


alexfauquette (approved):
Looks more unified than before

PR #1572, opened 2021-05-06
[docs] Add description for all events

m4theushw:
### Breaking changes

- [XGrid] Event `scrolling:rows` renamed to `rowsScroll` for consistency.
- [XGrid] Event `scroll:rowEnd` renamed to `rowsScrollEnd` for consistency.

---

Closes #1527

PR #3193, opened 2021-11-16
[DataGrid] Rework keyboard navigation

flaviendelangle:
### Goals

Remove abstractions whenever possible to make the code easier to read

- No key mapping
- No clamping afterwards, we are doing the checks during the switch
- Explicit variables `viewportPageSize` / `firstRowIndexInPage` / `lastRowIndexInPage` / `firstColIndex` / `lastColIndex` to make the switch statement as clear as possible
- A single `goToCell` and `goToHeader` method to abstract the scroll and focus behaviors

### Follow up

On v6 I think we should replace the `cellNavigationKeyDown` and `columnHeaderNavigationKeyDown` with `apiRef` methods.
For me events should  be "hey I did X, do whatever you want with this information" not "hey, could someone listen to me and do X".

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
+++ b/packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
@@ -1,20 +1,8 @@
 import * as React from 'react';
 import { GridEvents } from '../../../constants/eventsConstants';
 import { GridApiRef } from '../../../models/api/gridApiRef';
-import {
-  GridCellIndexCoordinates,
-  GridColumnHeaderIndexCoordinates,
-} from '../../../models/gridCell';
 import { GridCellParams } from '../../../models/params/gridCellParams';
 import { GridColumnHeaderParams } from '../../../models/params/gridColumnHeaderParams';
-import {
-  isArrowKeys,
-  isEnterKey,
-  isHomeOrEndKeys,
-  isPageKeys,
-  isSpaceKey,
-  isTabKey,
-} from '../../../utils/keyboardUtils';
 import { visibleGridColumnsLengthSelector } from '../columns/gridColumnsSelector';
 import { useGridSelector } from '../../utils/useGridSelector';
 import { useGridLogger } from '../../utils/useGridLogger';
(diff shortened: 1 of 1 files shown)


alexfauquette (approved):

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -78,135 +31,190 @@ export const useGridKeyboardNavigation = (
...
+          goToHeader(lastColIndex, event);
+          break;
+        }
+      }
     },
-    [apiRef, colCount, logger, currentPage.rows],
+    [apiRef, colCount, currentPage, goToCell, goToHeader],
   );
 
   useGridApiEventHandler(apiRef, GridEvents.cellNavigationKeyDown, navigateCells);
```
m4theushw, line 220: Since this file is being reworked.

```suggestion
  useGridApiEventHandler(apiRef, GridEvents.cellNavigationKeyDown, handleCellNavigationKeyDown);
```

Same for `GridEvents.columnHeaderNavigationKeyDown`.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -78,135 +31,190 @@ export const useGridKeyboardNavigation = (
...
-        if (!isCtrlPressed) {
-          // we go to the current row, first col, or last col!
-          nextCellIndexes = { colIndex: colIdx, rowIndex };
-        } else {
-          // In that case we go to first row, first col, or last row last col!
-          let newRowIndex = 0;
-          if (colIdx === 0) {
-            newRowIndex = currentPage.range.firstRowIndex;
+      const viewportPageSize = apiRef.current.unstable_getViewportPageSize();
+      const colIndexBefore = params.field ? apiRef.current.getColumnIndex(params.field) : 0;
```
m4theushw, line 64: Why the ternary operator?

m4theushw (commented):
Much more simple to understand. 👍

> On v6 I think we should replace the cellNavigationKeyDown and columnHeaderNavigationKeyDown with apiRef methods.

Yeah, it's weird to have a keyboard event that is published only when a specific set of keys are pressed but even when these keys are pressed it's ignored, e.g. on cell editing. This is one reason I made it private in #1572. I think `cellKeyDown` should be used for everything.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
flaviendelangle: Because their is one case where this event is triggered after comiting a row edit and we don't have a field.
It was working because `getColumnIndex` returned `-1` and the clamp behavior set it back  to 0, but it's not great calling it that way.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
flaviendelangle: Better indeed, done !

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
m4theushw: Makes sense. The event below shouldn't be published.

https://github.com/mui-org/material-ui-x/blob/0065e2f89b661070c7536b391544b7e82191698c/packages/grid/_modules_/grid/hooks/features/editRows/useGridEditRows.ts#L460

m4theushw (approved):

PR #3198, opened 2021-11-16
[DataGrid] Filter value are conserved when possible

### LINKED ISSUE #3188: [DataGrid] The value of the filter should not be reset when changing the column
### Duplicates

### Latest version

### Current behavior 😯

When changing the column to filter to another of the same type, the filter value is lost.

### Expected behavior 🤔

The value should be reused if the filter input of the next column is the same. The inverse is also true. Changing to a column or operator which uses a different filter input than the current one, then the value should be cleaned.

See how Airtable handles that:

qdrzVsi7Fk

### Steps to reproduce 🕹

Steps:

2. Filter for a value on the "Quantity" column
3. Change the filtered column to "Unit Price"

### Context 🔦

I noticed this while reviewing https://github.com/mui-org/material-ui-x/pull/2874.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3188

### The behavior

If column type changes, we try to keep the same operator (if available in the new column)
If the operator changes, we reset the filter Value if Input components are different.

### Limitation

The singleSelect type is special because the input component is the same as the text input, but it behaves differently. That's why it has a specific case in column cahnge. Moreover, it implies to check if the value is available in the value options

### DIFF
--- a/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
+++ b/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
@@ -74,6 +74,7 @@ function GridFilterInputSingleSelect(props: GridFilterInputSingleSelectProps) {
       itemValue = getValueFromValueOptions(item.value, currentValueOptions);
       if (itemValue !== item.value) {
         applyValue({ ...item, value: itemValue });
+        return;
       }
     } else {
       itemValue = item.value;


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -87,27 +87,89 @@ function GridFilterForm(props: GridFilterFormProps) {
...
+        applyFilterChanges({
+          ...item,
+          columnField,
+          operatorValue: newOperator.value,
+          value: eraseItemValue ? undefined : item.value,
+        });
+      } else {
+        applyFilterChanges({
+          ...item,
+          columnField,
```
m4theushw, line 142: The user could have changed to a column with the same type, however, with less operators than the current one. We need to reset the value in this case and save the new operator.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -87,27 +87,89 @@ function GridFilterForm(props: GridFilterFormProps) {
...
-        value: undefined,
-        columnField,
-        operatorValue: newOperator.value,
-      });
+      // try to keep the same operator when column change
+      const newOperator =
+        column.filterOperators!.find((operator) => operator.value === item.operatorValue) ||
+        column.filterOperators![0];
+
+      if (column.type === 'singleSelect') {
```
m4theushw, line 96: We shouldn't have code specific to one column type here. Why it was added?

m4theushw (commented):
We need a couple of test cases since the filtering is a critical part.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: For two reasons:
- Its InputComponent is the same as the InputComponent for string or number. Then the comparison of InputComponent will not detect if we switch from a "string" to a "singleSelect"
- It has an additional parameter: `valueOptions` so I need a specific case to check if the value is still available in the new column

m4theushw: > Its InputComponent is the same as the InputComponent for string or number.

One more reason to extract into another component the logic for `singleSelect`. https://github.com/mui-org/material-ui-x/pull/3018#discussion_r739439422

> It has an additional parameter: valueOptions so I need a specific case to check if the value is still available in the new column

`GridFilterInputValue` should sanitize the value it receives. This includes, if it's a `singleSelect`, checking if the value is included in the options.

alexfauquette: Ok, splitting components makes sense. I'm opening another PR for that

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -1155,6 +1251,64 @@ describe('<DataGrid /> - Filter', () => {
       render(<TestCase field="brand" operatorValue="contains" />);
       expect(screen.getByRole('textbox', { name: 'Value' }).value).to.equal('');
     });
+    it('should keep filter operator and value if available', async () => {
```
m4theushw, line 1254: Please separate each test case with a blank line.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -1155,6 +1251,64 @@ describe('<DataGrid /> - Filter', () => {
...
+          value="Pu"
+          imperativeFilterModel={false}
+        />,
+      );
+      expect(screen.getByRole('textbox', { name: 'Value' }).value).to.equal('Pu');
+      expect(screen.getByRole('combobox', { name: 'Operators' }).value).to.equal('contains');
+      expect(getColumnValues()).to.deep.equal(['Puma']);
+
+      setOperatorValue('isEmpty');
+      setOperatorValue('contains');
```
m4theushw, line 1306: One of the two can be removed.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -1155,6 +1251,64 @@ describe('<DataGrid /> - Filter', () => {
...
+
+      expect(getColumnValues()).to.deep.equal(['Nike', 'Adidas', 'Puma']);
+      expect(screen.getByRole('combobox', { name: 'Operators' }).value).to.equal('contains');
+      expect(screen.getByRole('textbox', { name: 'Value' }).value).to.equal('');
```
m4theushw, line 1310: Asserting that the last call to `onFilterModelChange` contains `value: undefined` may be easier on this case.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -71,30 +139,58 @@ describe('<DataGrid /> - Filter', () => {
...
-            items: [
-              {
-                columnField: field,
-                value,
-                operatorValue,
-              },
-            ],
-          }}
+          filterModel={
+            imperativeFilterModel
```
m4theushw, line 161: We never did that. You could check the `initialState` prop and if it's defined, don't set the model.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -71,30 +139,58 @@ describe('<DataGrid /> - Filter', () => {
...
+                }
+              : undefined
+          }
           disableColumnFilter={false}
           initialState={{
             preferencePanel: {
               open: true,
               openedPanelValue: GridPreferencePanelsValue.filters,
             },
+            ...(imperativeFilterModel
```
m4theushw, line 179: Spread the `initialState` prop here.

m4theushw (commented):
It's working correctly. My only comments are to improve the tests. We could add a test case using `singleSelect` since it does a bit of additional work to sanitize the options.

alexfauquette:
I had to remove the sanitization for `singleSelect` because we have a test "should work if valueOptions is not provided". So if I make sanitization, it will be a kind of breaking change.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -1155,6 +1251,64 @@ describe('<DataGrid /> - Filter', () => {
...
+      expect(screen.getByRole('combobox', { name: 'Operators' }).value).to.equal('contains');
+      expect(getColumnValues()).to.deep.equal(['Puma']);
+
+      setColumnValue('slogan');
+
+      expect(getColumnValues()).to.deep.equal([]);
+      expect(screen.getByRole('combobox', { name: 'Operators' }).value).to.equal('from');
+      expect(screen.getByTestId('customInput').value).to.equal('');
+    });
+    it('should reset value if the new operator has no input component', async () => {
```
m4theushw, line 1292: `async` can be removed if the test doesn't use promises.

```suggestion
    it('should reset value if the new operator has no input component', () => {
```

x other tests

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -71,30 +139,36 @@ describe('<DataGrid /> - Filter', () => {
       operatorValue?: string;
       value?: any;
       field?: string;
+      imperativeFilterModel?: boolean;
```
m4theushw, line 142: This can be removed.

packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx
```diff
@@ -1,10 +1,46 @@
...
+  GridFilterInputValue,
+} from '@mui/x-data-grid';
 import { getColumnValues } from 'test/utils/helperFn';
 
+function setColumnValue(columnValue) {
+  fireEvent.change(screen.getByRole('combobox', { name: 'Columns' }), {
+    target: { value: columnValue },
+  });
+}
+function setOperatorValue(operatorValue) {
```
m4theushw, line 20:
```suggestion

function setOperatorValue(operatorValue) {
```

m4theushw (commented):
Removing the sanitization created one regression:

prnWnAbljb

> I had to remove the sanitization for singleSelect because we have a test "should work if valueOptions is not provided". So if I make sanitization, it will be a kind of breaking change.

This is test is meant to ensure that the grid won't crash if the `valueOptions` is undefined. Although, if it's defined and there's no corresponding value it can still be sanitized.

alexfauquette:
@m4theushw Nice catch this regression :)

I did not notice that the test was for the special case of `undefined`. I made it such that the sanitization is applied if the `valueOptions` is different from undefined, And adding tests, to prevent the regression you found

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
```diff
@@ -36,29 +45,44 @@ function GridFilterInputSingleSelect(props: GridFilterInputSingleSelectProps) {
...
 
   React.useEffect(() => {
-    const itemValue = item.value ?? '';
+    let itemValue;
+
+    if (currentValueOptions !== undefined) {
+      // sanitize if valueOptions are provided
+      itemValue = getValueFromValueOptions(item.value, currentValueOptions);
+      if (itemValue !== item.value) {
+        applyValue({ ...item, value: itemValue });
```
m4theushw, line 76: Just a small detail. For a brief moment, `filterValueState` will contain an invalid value if we don't return.

```suggestion
        applyValue({ ...item, value: itemValue });
        return;
```

m4theushw (approved):

PR #3201, opened 2021-11-16
[core] Add script to sync translation files

m4theushw:


flaviendelangle (approved):

alexfauquette (approved):

oliviertassinari:
@m4theushw Cooool, for further automation, we could maybe make it part of the CI, so we never forget about running the script. 

@mui-org/x speaking of CI tools and how we can make our work as maintainers easier. Would it make sense to dedicate some of the weekly meetings to maintain a backlog of this? Effectively, we could collaborate on the next opportunities. I will propose this topic in the next retrospective, maybe it will resonate with the team, if not, an idea to hurry.

PR #3206, opened 2021-11-17
[core] `EventManager` should not run listeners removed after the beginning of the emission

flaviendelangle:
This occurred for the 1st time in #3007.
The more synchronous our behaviors become (which is a good thing), the more it became likely that we would end up with this kind of scenario.

1. The event `X` is emitted which has 2 listeners `A` and `B`
2. The listener `A` is ran, it triggers a behavior that causes the removal of the listener `B` (potentially by emitting another event `Y`, as long as it is synchronous, the removal can be very far in the run sequence, it can even be caused on another render if the render is synchronous)
3. The listener `B` is ran

On #3007 it was causing `B` to run after the unmount of `GridOverlay`. I think it was caused when `props.loading` was switched to false at the same time as the new `props.rows` arrived. We wanted to update the dimensions of the Grid for the new rows AND remove the overlay. But the overlay listens to the dimensions to resize itself. 

I kept the exact same execution order (the `isFirst` listeners are executed from the most recent to the oldest). Not sure it make sense but I did not want to break any user code.
I switched to a `Map` implementation to avoid having to do an `include` on the listener list before running each listener in `emit`.


m4theushw:
> On #3007 it was causing B to run after the unmount of GridOverlay. I think it was caused when props.loading was switched to false at the same time as the new props.rows arrived. 

Doesn't https://github.com/mui-org/material-ui-x/pull/3217#issuecomment-974578348 fix this?

Is there an issue to link here? Or an example using real events? It's hard to review the code without a concrete example.

flaviendelangle:
You can try removing the `isMounted` logic here: https://github.com/mui-org/material-ui-x/pull/3206/files#diff-29eacbb7ce682747bc3a8bf3633af42149164a97bcd7b2ed0dd55a333413382aL65

React should warn you that you are trying to update the state after the unmount of `GridOverlay`

And for me #3217 does not change the issue.

flaviendelangle:
You can try removing the `isMounted` logic here: https://github.com/mui-org/material-ui-x/pull/3206/files#diff-29eacbb7ce682747bc3a8bf3633af42149164a97bcd7b2ed0dd55a333413382aL65

React should warn you that you are trying to update the state after the unmount of `GridOverlay`

And for me #3217 does not change the general issue.
It seems to fix this specific issue (not sure why but I saw that you removed the `isMounted` on #3217 and the tests pass). 

But you can look at https://codesandbox.io/s/datagridpro-v5-quick-start-forked-g55f4?file=/src/App.tsx
If you subscribe then run, the `callbackB` is ran after it has been unsubscribed.

packages/grid/_modules_/grid/utils/EventManager.ts
```diff
@@ -38,30 +60,36 @@ export class EventManager {
...
+    if (!collection) {
+      return;
+    }
+
+    const highPriorityListeners = Array.from(collection.highPriority.keys());
+    const regularListeners = Array.from(collection.regular.keys());
+
+    for (let i = highPriorityListeners.length - 1; i >= 0; i -= 1) {
+      const listener = highPriorityListeners[i];
+      if (collection.highPriority.has(listener)) {
```
m4theushw, line 84: Could you add a test resuming what you said in https://github.com/mui-org/material-ui-x/pull/3206#issuecomment-976678291?

Besides the separation of the listeners in two groups, the biggest value of this PR on this line (and the equivalent for regular events).

m4theushw (commented):
Running the tests in StrictMode brought the "Can't perform a React state update on an unmounted component." error again. #3217 should fix this.

flaviendelangle:
> Running the tests in StrictMode brought the "Can't perform a React state update on an unmounted component." error again. #3217 should fix this.

This PR does not fix the double test registration in render, so yes #3217 is still very much needed :+1:

packages/grid/_modules_/grid/utils/EventManager.ts
flaviendelangle: Yes, this PR was supposed to be a one liner if I did not have to change to a Map.
I'll add a few tests on the `EventManager` :+1:

flaviendelangle: Done

m4theushw (approved):

PR #3209, opened 2021-11-17
[docs] Improve rows documentation

flaviendelangle:
- [x] Remove strange wordings / repetitions like 

> Grid rows can be defined with the `rows` prop.
`rows` expects an array of objects.
Rows should have this type: `GridRowModel[]`.

- [x] Improve the `updateRows` demo to display all the basic cases (I could add a "delete" + "update" to show that we can put several type of actions into one update, but not sure it is worth it)

- [x] New `prop.rows` update demo with the same features as the `updateRows` one

- [x] Section order change to put the infinite loading section after the `updateRows` (to keep the two basic row updating sections next to eachother)

- [x] New warning to encourage people to memoize their rows

- [x] New link to the density demo to avoid having people updating the row height when they can do otherwise


docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -53,6 +43,14 @@ Alternatively, if you would like to delete a row, you would need to pass an extr
 apiRef.current.updateRows([{ id: 1, _action: 'delete' }]);
 ```
 
+### Infinite loading [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+The grid provides a `onRowsScrollEnd` prop that can be used to load additional rows when the scroll reaches the bottom of the viewport area.
+
+In addition, the area in which the callback provided to the `onRowsScrollEnd` is called can be changed using `scrollEndThreshold`.
```
alexfauquette, line 50:
```suggestion
In addition, the area in which `onRowsScrollEnd` is called can be changed using `scrollEndThreshold`.
```

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -72,6 +70,8 @@ To change the row height for the whole grid, set the `rowHeight` prop:
 
 {{"demo": "pages/components/data-grid/rows/DenseHeightGrid.js", "bg": "inline"}}
 
+If you want to create a more / less compact grid and not only set the row height, take a look at our [Density documentation](/components/data-grid/accessibility/#density-selector)
```
alexfauquette, line 73: I would put this before the demo. Otherwise, it is easy to miss the info

alexfauquette (approved):

PR #3217, opened 2021-11-18
[core] Replace @mui/core with @mui/base

### LINKED ISSUE #3157: Update `@mui/core` to `@mui/base` on grid-generator
### Duplicates

### Latest version

### Current behavior 😯

Due to https://github.com/mui-org/material-ui/pull/29585.

~`@mui/core` should not be a dependency because~
~- the package is not used directly~
~- `@mui/material` is already a peer dependency (`@mui/core` is a dependency of this package)~

~I believe that this is a mistake from the migration when we changed the package name.~

### Expected behavior 🤔

~`@mui/core` is not listed as a dependency.~

### Steps to reproduce 🕹

~https://github.com/mui-org/material-ui-x/blob/56610e553ef32afab69ac169c78674430aa2a4ce/packages/grid/x-data-grid-generator/package.json?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L31~

### Context 🔦

~I found it in yarn.lock on this PR. https://github.com/mui-org/material-ui/pull/29464#discussion_r746424792~

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Closes #3157

Related to https://github.com/mui-org/material-ui/pull/29536


yarn.lock
```diff
@@ -2418,7 +2425,20 @@
...
+  dependencies:
+    "@babel/runtime" "^7.16.3"
+    "@emotion/is-prop-valid" "^1.1.1"
+    "@mui/utils" "^5.1.1"
+    "@popperjs/core" "^2.4.4"
+    clsx "^1.1.1"
+    prop-types "^15.7.2"
+    react-is "^17.0.2"
+
+"@mui/core@5.0.0-alpha.54":
```
eps1lon, line 2441: Something is still using `@mui/core`.

yarn.lock
```diff
@@ -1353,6 +1353,13 @@
   dependencies:
     "@emotion/memoize" "^0.7.4"
 
+"@emotion/is-prop-valid@^1.1.1":
```
eps1lon, line 1356: Can you deduplicate that package?

m4theushw:
This PR became more than a simple change of packages. To be able to build the docs, I had to also update the commit we're pointing the monorepo to. Because of this, it brought https://github.com/mui-org/material-ui/pull/29589 which made all tests to run with StrictMode enabled by default causing a few tests to start to fail.

#### Why they broke?

https://github.com/mui-org/material-ui-x/blob/e45cf31723303ccc20cf21cccb6ef1e975364bed/packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts#L26

In #2557, we started to register events synchronously, and re-register them, if necessary, in an effect. This change created a memory leak. With StrictMode, each component is rendered twice so two events are registered synchronously (React resets the ref that keeps the subscription in the second run). Once an effect runs, it unsubscribes the **last** registered event. The very first registered event is never removed, so it stays there forever. Even though we had StrictMode in the docs, this was working okayish up until now. I saw the memory-leak error once but never again, so I ignored. Running tests with StrictMode, however, made the problem clearly visible, since `GridOverlay` was updating the state after the component is unmounted: https://github.com/mui-org/material-ui-x/blob/e45cf31723303ccc20cf21cccb6ef1e975364bed/packages/grid/_modules_/grid/components/containers/GridOverlay.tsx#L66

To solve it, I removed the code that registers events synchronously. I don't know if we should try to restore it. This is an anti-pattern, it's even in the React docs (https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects - search for "idempotent").

eps1lon:
> Because of this, it brought mui-org/material-ui#29589 which made all tests to run with StrictMode enabled by default causing a few tests to start to fail.

@m4theushw  Note that mui-org/material-ui#29589 was only a regression fix. Tests were run in StrictMode by default for the longest time. StrictMode was accidentally disabled for just a few days.

You can always disable StrictMode either per renderer via `createRenderer({ strict: false })` or per render `render(<div />, { strict: false })` until you fixed those issues.

Looking forward to this PR getting merged since it's hard to keep the lockfile in check in mui-org/material-ui so that we don't pull in duplicate modules resulting in unexpected behavior.

m4theushw:
I've reverted some of the changes I did and disabled StrictMode on the problematic tests. Although removing the synchronous event registering worked in general, it generated one difference. I'll open another PR with a proposal to fix the event system while making it work with StrictMode.

flaviendelangle (approved):

Issue #2144, opened 2021-07-14
[DataGrid] Allow to add margin between rows


vitalist82:
When applying custom style to rows, e.g. specifying row height and margin-bottom to achieve extra space between rows, the last rows of the grid are not visible, because the parent container has a height that seems to be calculated without taking the extra margin into account.

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

In the example on https://codesandbox.io/s/quizzical-meitner-cuff9 I styled the rows to have an extra green border after being selected and also 5px margin on the bottom so that the space between the rows is bigger. Without the `disableExtendRowFullWidth` attribute the right border isn't visible probably also because of the precalculated size of the parent containers. When using `disableExtendRowFullWidth` attribute, I am able to work around this issue with right border by specifying fixed size of the grid's parent element, but since I am trying to make the whole thing responsive, this isn't the best solution.

## Expected Behavior 🤔

I would like to be able to customize the height, borders and spacing of rows without breaking the layout.

## Steps to Reproduce 🕹

https://codesandbox.io/s/quizzical-meitner-cuff9

Steps:

1. Select a row or two
2. Scroll down to the end of the grid

## Context 🔦

We're trying to adapt the component as close to the UI/UX requirements as possible.

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>

Browser used: Chrome..
  
System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 14.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 7.19.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1023.0), Chromium (91.0.864.67)

</details>

## Order id 💳

Still doing POC without having a license.

oliviertassinari:
@vitalist82 Thanks for the bug reports. It gives us another interesting test case on how the current rendering logic is wrong. It's related to #1911

carl-meyer-paxton:
Hi, I am also busy trying to add some spacing between rows. 

I am using ` @mui/x-data-grid": "^5.0.0-beta.3` and still need to use makeStyles to style the data grid (not using new emotion yet?)

I have applied the following styles:
```
const useStyles = makeStyles((theme: Theme) => ({
  root: {
[...]

PR #3218, opened 2021-11-18
[DataGrid] Variable row height

### LINKED ISSUE #438: [DataGrid] Variable row height
This is a sequel of #417 to focus on the variable height problem. Variable height is a simpler problem than dynamic height. 

### Benchmark

- https://react-window.now.sh/#/examples/list/variable-size `itemSize` (function) 
- https://bvaughn.github.io/react-virtualized/#/components/Grid `rowHeight` (function)
- https://www.ag-grid.com/javascript-grid-row-height/#getrowheight-callback `getRowHeight` (function)
- https://github.com/tannerlinsley/react-virtual#options `estimateSize`.

DanailH:
Fixes #438 

Preview -> https://deploy-preview-3218--material-ui-x.netlify.app/components/data-grid/rows/#variable-row-height

ToDo:
- [x] Tests
- [x] Docs


DanailH:
Ok ... I would really like a review of what has been done so far. The functionality is almost ready but I'm struggling with fixing the failing tests. I'm probably missing a key interaction because this is closely tight to the virtualization and is impacted by every functionality that changes the rows in a way (sorting, filtering, pagination, treeData, density, etc). If someone can spare a pair of eyes to check it I would really appreciate it.


I skipped 2 tests (both are the same, related to the translation using the Theme) - for some reason it enters an infinite loop there.

Another thing that I don't really get is how are the tests from `yarn test:karma` passing and the ones from `yarn test:coverage` failing?

packages/grid/_modules_/grid/hooks/features/rowHeight/useGridRowHeight.ts
```diff
@@ -0,0 +1,84 @@
...
+  const [gridState, setGridState, forceUpdate] = useGridState(apiRef);
+
+  const hydrateRowsMeta = React.useCallback(() => {
+    logger.debug(`Hydrating all visible rows`);
+    // console.log('Hydrating rows...')
+    // if (!gridState.pagination) {
+    //   return;
+    // }
+
+    const { rows } = getCurrentPageRows(apiRef.current.state, props);
```
m4theushw, line 38: Here's the reason of why the skipped tests are failing. `props` changes every time and makes `hydrateRowsMeta` to be called. To fix, change to:

```ts
const { rows } = getCurrentPageRows(apiRef.current.state, {
  pagination: props.pagination,
  paginationMode: props.paginationMode,
});
```

m4theushw (commented):
I didn't look into the functionality yet. However, I'm already using some parts in https://github.com/mui-org/material-ui-x/pull/3387.

I don't think we should treat variable row height as a feature, the column variable width is not too.

For the remaining tests that are failing, you can fix them by using `disableVirtualization`. They fail in JSDOM because `rootRef.current!.clientHeight!` is 0.

packages/grid/_modules_/grid/hooks/features/rowHeight/useGridRowHeight.ts
DanailH: Thank you for checking the PR. I tried the suggestion but it's still failing the 2 skipped tests. The problem is coming from the `React.useEffect` used here but the other way is to subscribe/create events to react to the state changes listed in the `useEffect`.

Regarding not being a separate feature - true but then the way things are hooked up the variable row height depends on the pagination.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -107,12 +107,11 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
...
-    containerWidth,
-    rootProps.autoHeight,
+    columnsMeta.positions,
+    rowsMeta.positions,
     disableVirtualization,
-    rowHeight,
     currentPage.rows.length,
+    containerWidth,
     visibleColumns.length,
   ]);
```
m4theushw, line 117: Another finding doing #3387. If `rowsMeta.totalHeight` changes in a way where it causes the vertical scrollbar to be visible e.g. click a button to change the row height, the container width needs to be recalculated. I did this in https://github.com/mui-org/material-ui-x/pull/3387/files#diff-7fb7dbb1d7fe2c699d9fc5cfc5c6a56b2489af3b46dc1a2a8eaa3404575cacbdR136

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,109 @@
...
+      setGridState((state) => {
+        const positions: number[] = [];
+        const totalHeight = rows.reduce((acc: number, row) => {
+          positions.push(acc);
+          let targetRowHeight = state.density?.rowHeight;
+
+          if (getRowHeight) {
+            // Default back to base rowHeight if getRowHeight returns null or undefined.
+            targetRowHeight =
+              getRowHeight({ ...row.model, densityFactor: state.density?.factor }) ||
```
flaviendelangle, line 46: I would advocate against spreading the model like that.
I would prefer the following:

```suggestion
              getRowHeight({ ...row, densityFactor: state.density?.factor }) ||
```

And 

```ts
export interface GridRowHeightParams extends GridRowEntry {
  densityFactor: number;
}
```

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,109 @@
...
+            meta: { totalHeight, positions },
+          },
+        };
+      });
+      forceUpdate();
+    },
+    [apiRef, props.pagination, props.paginationMode, setGridState, forceUpdate],
+  );
+
+  const handlePreProcessorRegister = React.useCallback<
```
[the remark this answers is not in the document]
m4theushw: https://github.com/mui-org/material-ui-x/pull/3387/files#diff-6a287bb8e507d976894a4b18fa26fda06b80cdeaed90b825b527dfabb65e371fR157

It's not necessary for this PR but #3387 and #2144 needs it.

oliviertassinari (commented):
I didn't see this feedback yet, so leaving: the keyboard handling is not scrolling correctly, the focused cell is only partially revealed:

keyboard
 
https://deploy-preview-3218--material-ui-x.netlify.app/storybook/?path=/story/datagridpro-test-rows--variable-row-height

m4theushw (commented):
`useGridDimensions` also needs to consider variable row height.

https://github.com/mui-org/material-ui-x/blob/b7fb0169f74cec58817e85ff167d197a244d6652/packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts#L95

packages/grid/_modules_/grid/hooks/features/density/useGridDensity.tsx
```diff
@@ -25,18 +25,21 @@ const getUpdatedDensityState = (
         value: newDensity,
         headerHeight: Math.floor(newHeaderHeight * COMPACT_DENSITY_FACTOR),
         rowHeight: Math.floor(newRowHeight * COMPACT_DENSITY_FACTOR),
+        factor: COMPACT_DENSITY_FACTOR,
       };
     case GridDensityTypes.Comfortable:
       return {
         value: newDensity,
         headerHeight: Math.floor(newHeaderHeight * COMFORTABLE_DENSITY_FACTOR),
```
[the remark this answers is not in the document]
DanailH: I'll add a comment to not forget to check this function.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,109 @@
...
+ * @requires useGridPage (method)
+ */
+export const useGridRowsHydration = (
+  apiRef: GridApiRef,
+  props: Pick<GridComponentProps, 'getRowHeight' | 'pagination' | 'paginationMode'>,
+): void => {
+  const [gridState, setGridState, forceUpdate] = useGridState(apiRef);
+  const rowsHeightCollection = React.useRef(new Map<GridRowId, number>());
+
+  const hydrateRowsMeta = React.useCallback(
```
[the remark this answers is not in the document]
DanailH: I tried that approach but the thing is that it's the same logic as the columns. It's a different data structure now, so instead of having just a static `rowHeight` we extract the row height by checking the positions of the different rows.

DanailH:
> I didn't see this feedback yet, so leaving it: the keyboard handling is not scrolling correctly, the focused cell is only partially revealed:
> 
> keyboard
> 
> https://deploy-preview-3218--material-ui-x.netlify.app/storybook/?path=/story/datagridpro-test-rows--variable-row-height

Fixed

DanailH:
> `useGridDimensions` also needs to consider variable row height.
> 
> https://github.com/mui-org/material-ui-x/blob/b7fb0169f74cec58817e85ff167d197a244d6652/packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts#L95

Updated

DanailH:
Ok there is one pagination test failing which is weird because it does pass locally.

packages/grid/_modules_/grid/hooks/features/rows/useGridRows.ts
```diff
@@ -41,6 +42,11 @@ interface GridRowsInternalCacheState {
    * It is used to avoid processing several time the same set of rows
    */
   rowsBeforePartialUpdates: GridRowsProp;
+
```
m4theushw, line 45:
```suggestion
```

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (getRowHeight) {
+          // Default back to base rowHeight if getRowHeight returns null or undefined.
+          targetRowHeight =
+            getRowHeight({ ...row, densityFactor: gridDensityFactorSelector(state) }) ||
+            gridDensityRowHeightSelector(state);
+        }
+
+        const heights = apiRef.current.unstable_applyPreProcessors(
+          GridPreProcessingGroup.rowHeight,
+          { base: targetRowHeight }, // We use an object to make simple to check if a size was already added or not
```
m4theushw, line 58: Could you remove the `rowHeight` pre-processing? It's not needed for this PR.

packages/grid/_modules_/grid/models/api/gridRowsHydrationApi.ts
```diff
@@ -0,0 +1,13 @@
+import { GridRowId } from '../gridRows';
+
+/**
+ * The Row Hydration API interface that is available in the grid `apiRef`.
+ */
+export interface GridRowsHydrationApi {
+  /**
+   * Get target row height.
```
m4theushw, line 8:
```suggestion
   * Gets target row height.
```

packages/grid/x-data-grid/src/tests/rows.DataGrid.test.tsx
```diff
@@ -6,6 +6,7 @@ import Portal from '@mui/material/Portal';
 import { DataGrid, DataGridProps, GridActionsCellItem } from '@mui/x-data-grid';
 import { getColumnValues, getRow } from 'test/utils/helperFn';
 import { getData } from 'storybook/src/data/data-service';
+import { COMPACT_DENSITY_FACTOR } from 'packages/grid/_modules_/grid/hooks/features/density/useGridDensity';
```
m4theushw, line 9: Could we import from `@mui/x-data-grid`?

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -71,6 +71,14 @@ To change the row height for the whole grid, set the `rowHeight` prop:
 
 {{"demo": "pages/components/data-grid/rows/DenseHeightGrid.js", "bg": "inline"}}
 
+## Variable row height
+
+If you need different rows to have different row heights this can be achieved using the `getRowHeight` prop. The prop is called for each visible row and if the return value is `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
+
+{{"demo": "pages/components/data-grid/rows/VariableRowHeightGrid.js", "bg": "inline"}}
+
+Changing the `DataGrid` density doesn't affect the rows with variable row height. You can access the `DataGrid` density factor from the `getRowHeight` params.
```
m4theushw, line 80: This could be a warning.

```suggestion
> Changing the `DataGrid` density does not affect the rows with variable row height.
> You can access the density factor from the params provided to the `getRowHeight` prop.
```

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -71,6 +71,14 @@ To change the row height for the whole grid, set the `rowHeight` prop:
 
 {{"demo": "pages/components/data-grid/rows/DenseHeightGrid.js", "bg": "inline"}}
 
+## Variable row height
+
+If you need different rows to have different row heights this can be achieved using the `getRowHeight` prop. The prop is called for each visible row and if the return value is `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
+
+{{"demo": "pages/components/data-grid/rows/VariableRowHeightGrid.js", "bg": "inline"}}
```
m4theushw, line 78: This demo could have the toolbar so users can play with the density. Also, the preview doesn't show how to calculate the value based on the factor.

m4theushw:
I just merged #3558. Could you check if tests that were failing with "Can't perform a React state update on an unmounted component." now pass? Also, calling `clock.runToLast` shouldn't be necessary anymore, unless to advance a transition.

docs/src/pages/components/data-grid/rows/VariableRowHeightGrid.tsx
```diff
@@ -0,0 +1,44 @@
...
+    rowLength: 100,
+    maxColumns: 6,
+  });
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGrid
+        {...data}
+        getRowHeight={() => {
+          isRowIndexOdd.current = !isRowIndexOdd.current;
```
flaviendelangle, line 30: This logic may be a little weird for users at 1st.
If we could find another row height criteria than the odd / even index based on a mutable ref.

Non blocking

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -71,6 +71,15 @@ To change the row height for the whole grid, set the `rowHeight` prop:
 
 {{"demo": "pages/components/data-grid/rows/DenseHeightGrid.js", "bg": "inline"}}
 
+## Variable row height
+
+If you need different rows to have different row heights this can be achieved using the `getRowHeight` prop. The prop is called for each visible row and if the return value is `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
```
flaviendelangle, line 76: Just to avoid repetitions.

```suggestion
If you need some rows to have different row heights this can be achieved using the `getRowHeight` prop. This function is called for each visible row and if the return value is a `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
```

packages/grid/_modules_/grid/hooks/features/density/useGridDensity.tsx
```diff
@@ -13,6 +13,7 @@ import { isDeepEqual } from '../../../utils/utils';
 export const COMPACT_DENSITY_FACTOR = 0.7;
 export const COMFORTABLE_DENSITY_FACTOR = 1.3;
 
+// TODO: revise keeping headerHeight and rowHeight in state for v6
```
flaviendelangle, line 16:
```suggestion
// TODO v6: revise keeping headerHeight and rowHeight in state
```

Seems to be the most common form for v6 comment in the codebase, will help searching for them.

packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts
```diff
@@ -88,18 +89,15 @@ export function useGridDimensions(
...
-    const currentPage = getCurrentPageRows(apiRef.current.state, {
-      pagination: props.pagination,
-      paginationMode: props.paginationMode,
-    });
-    const pageScrollHeight = currentPage.rows.length * rowHeight;
     const viewportOuterSize: ElementSize = {
       width: rootDimensionsRef.current.width,
-      height: props.autoHeight ? pageScrollHeight : rootDimensionsRef.current.height - headerHeight,
+      height: props.autoHeight
+        ? rowsMeta.totalHeight
```
flaviendelangle, line 95: At 1st I thought it was the sum for all rows

```suggestion
        ? rowsMeta.currentPageTotalHeight
```

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,81 @@
...
+} from '../density/densitySelector';
+import { gridFilterStateSelector } from '../filter/gridFilterSelector';
+import { gridPaginationSelector } from '../pagination/gridPaginationSelector';
+import { gridSortingStateSelector } from '../sorting/gridSortingSelector';
+
+/**
+ * @requires useGridPageSize (method)
+ * @requires useGridPage (method)
+ */
+export const useGridRowsHydration = (
```
flaviendelangle, line 21: I try to have 1 feature hook = 1 sub state to simplify the codebase understanding.

Could we either

1. Move `state.rows.meta` into some `state.rowDimensions` for instance.
2. Call `useGridRowsHydratation` inside `useGridRows`. Which seems to require more changes due to the dependances of this hook, so maybe not worth it.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,81 @@
...
+      const positions: number[] = [];
+      const totalHeight = rows.reduce((acc: number, row) => {
+        positions.push(acc);
+        let targetRowHeight = gridDensityRowHeightSelector(state);
+
+        if (getRowHeight) {
+          // Default back to base rowHeight if getRowHeight returns null or undefined.
+          targetRowHeight =
+            getRowHeight({ ...row, densityFactor: gridDensityFactorSelector(state) }) ||
+            gridDensityRowHeightSelector(state);
```
flaviendelangle, line 48: Even if memoized, we could move the selectors outside of the `reduce` to only call them once.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,81 @@
...
+    apiRef.current.setState((state) => {
+      const positions: number[] = [];
+      const totalHeight = rows.reduce((acc: number, row) => {
+        positions.push(acc);
+        let targetRowHeight = gridDensityRowHeightSelector(state);
+
+        if (getRowHeight) {
+          // Default back to base rowHeight if getRowHeight returns null or undefined.
+          targetRowHeight =
+            getRowHeight({ ...row, densityFactor: gridDensityFactorSelector(state) }) ||
```
flaviendelangle, line 47:
```suggestion
            getRowHeight({ ...row, densityFactor: gridDensityFactorSelector(state) }) ??
```

Based on my understanding of the doc `0` should be treated as a valid value here

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,81 @@
...
+      };
+    });
+    apiRef.current.forceUpdate();
+  }, [apiRef, pagination, paginationMode, getRowHeight]);
+
+  const getTargetRowHeight = (rowId: GridRowId): number =>
+    rowsHeightLookup.current[rowId] || rowHeight;
+
+  // The effect is used to build the rows meta data - totalHeight and positions.
+  // Because of variable row height this is needed for the virtualization
```
flaviendelangle, line 71: I don't understand that comment line

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,81 @@
...
+  React.useEffect(() => {
```
flaviendelangle, line 72: It would be nice to be able to update the row height state synchronously instead that in an effect, to avoid flickering if sometimes a render slides between the visible rows update and the rows meta update.
With the naive example below, we might have several computation when we update the visible rows AND the page at the same time for instance. It would be interesting to have an event for `pageRowsChange` I think. As it is not a trivial change, I think we can keep it for later.

```suggestion
  React.useEffect(() => {
    hydrateRowsMeta();
  }, [hydrateRowsMeta]);

  // The code for the api method

  useGridApiEventHandler(apiRef, GridEvents.visibleRowsSet, hydrateRowsMeta);
  useGridApiEventHandler(apiRef, GridEvents.pageChange, hydrateRowsMeta);
  useGridApiEventHandler(apiRef, GridEvents.pageSizeChange, hydrateRowsMeta);
```

packages/grid/_modules_/grid/models/api/gridRowsHydrationApi.ts
```diff
@@ -0,0 +1,13 @@
+import { GridRowId } from '../gridRows';
+
+/**
+ * The Row Hydration API interface that is available in the grid `apiRef`.
+ */
+export interface GridRowsHydrationApi {
+  /**
+   * Gets target row height.
+   * @param {GridRowId} id The id of the row.
+   * @returns {number} The target row height.
```
flaviendelangle, line 10:
```suggestion
   * @returns {number} The target row height.
   * @ignore - do not document.
```

flaviendelangle:
I am adding @joserodolfofreitas to the review, for the doc page :+1:

docs/src/pages/components/data-grid/rows/VariableRowHeightGrid.tsx
DanailH: If this is too weird I can just make a static example with custom rows and check their id?

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
DanailH: In general, I agree with the approach of having 1 feature hook and 1 substate. In this case tho it is a bit different because the row dimensions do make sense to the part of the row state as you can't separate them (same logic as for the columns). Ideally, the `useGridRowsHydration` should be part of the `useGridRows` (as it was in the beginning) but then there is a problem because of the pagination.

I would suggest keeping it as is for now and maybe reviewing it already took more time than planned for this feature to be done.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
DanailH: Without this the rows a basically broken, not rendered.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
DanailH: Agreed as a follow-up. I was looking into a solution like this in the beginning but we need to add events for density and `rowHeight` as well in order for this to work. The difficulty would be to add the events emit on all the relevant places and not miss a case.

flaviendelangle: I feel like at one point we may need a hooks that runs a state updater when a selector value has changed.

Something like:

```tsx
export const useGridDerivedState = <V>(
  apiRef: GridApiRef,
  selector: (state: GridState) => V, 
  updater: (state: GridState) => GridState,
) => {
  const prevValueRef = React.useRef<V>();
  
  useFirstRender(() => {
    prevValueRef.current = selector(apiRef.current.state);
  });

  const handleStateChange = React.useCallback(() => {
    const newValue = selector(apiRef.current.state);
    if (newValue !== prevValueRef.current) {
      prevValueRef.current = newValue;
      apiRef.current.setState(updater);
    }
  }, [apiRef]);

  useGridApiEventHandler(apiRef, GridEvents.stateChange, handleStateChange);
}
```

Or even with a single `setState` with some deeper change.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
flaviendelangle: The only problem in doing it later is that, as long as we export selectors like `gridRowsStateSelector` (which I would be in favor of dropping), the `state.rows` is public API so we won't be able to change it before v6.

docs/src/pages/components/data-grid/rows/VariableRowHeightGrid.tsx
flaviendelangle: I think it would make the example easier to read.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
DanailH: I moved the Meta state outside of the rows state and renamed the files accordingly.

packages/grid/_modules_/grid/hooks/features/rows/gridRowsMetaState.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * The grid rows total height and row possitions.
```
flaviendelangle, line 2:
```suggestion
 * The grid rows total height and row positions.
```

packages/grid/_modules_/grid/hooks/features/rows/gridRowsMetaState.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export interface GridRowsMetaState {
+  /**
+   * The sum of of all visible grid rows in the current rows.
+   */
+  currentPageTotalHeight: number;
+  /**
+   * The grid rows possitions.
```
flaviendelangle, line 10:
```suggestion
   * The grid rows positions.
```

flaviendelangle:
You can remove the `idCounter` from the demo :+1:

flaviendelangle (commented):
`unstable_getViewportPageSize` has not been updated.
This mean for instance that pressing `PageDown` on a `DataGridPro` won't go down by the expected amount of rows.

Maybe we should stop using it and use `getIndexFromScroll` instead for the keyboard navigation.
And for the `handleRowsScrollEnd`, if you can fix it it would be great.
If not, at least add a comment in the JSDoc saying that it does not work well with variable row height.

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -71,6 +71,15 @@ To change the row height for the whole grid, set the `rowHeight` prop:
 
 {{"demo": "pages/components/data-grid/rows/DenseHeightGrid.js", "bg": "inline"}}
 
+## Variable row height
```
m4theushw, line 74: Should it be a sub-section of "Row height"?

```suggestion
### Variable row height
```

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -71,6 +71,15 @@ To change the row height for the whole grid, set the `rowHeight` prop:
...
+
+If you need some rows to have different row heights this can be achieved using the `getRowHeight` prop. This function is called for each visible row and if the return value is a `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
+
+{{"demo": "pages/components/data-grid/rows/VariableRowHeightGrid.js", "bg": "inline"}}
+
+> Changing the `DataGrid` density does not affect the rows with variable row height.
+> You can access the density factor from the params provided to the `getRowHeight` prop
+
```
m4theushw, line 82: Could you add a warning recommending `getRowHeight` to be be memoized. See https://deploy-preview-3387--material-ui-x.netlify.app/components/data-grid/group-pivot/#master-detail for an example.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsMeta.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * @requires useGridPageSize (method)
+ * @requires useGridPage (method)
+ */
+export const useGridRowsMeta = (
+  apiRef: GridApiRef,
+  props: Pick<DataGridProcessedProps, 'getRowHeight' | 'pagination' | 'paginationMode'>,
+): void => {
+  const { getRowHeight, pagination, paginationMode } = props;
```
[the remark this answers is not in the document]
DanailH: I need all 3 to be in the dep array and when I add them TS complains that I should just add `props` and when that happens I get an infinite loop. So this is to trick TS basically.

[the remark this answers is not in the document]
m4theushw: Oh I see. I faced this problem once and destructuring is the only solution.

packages/grid/x-data-grid-pro/src/tests/pagination.DataGridPro.test.tsx
```diff
@@ -83,17 +81,18 @@ describe('<DataGridPro /> - Pagination', () => {
...
               onPageSizeChange={(newPageSize) => setPageSize(newPageSize)}
               pagination
+              disableVirtualization
             />
           </div>
         );
       };
 
       render(<GridTest />);
+      clock.runToLast();
```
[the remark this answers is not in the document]
DanailH: This one no, the other one seems to be.

packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts
```diff
@@ -149,6 +145,8 @@ export function useGridDimensions(
   );
 
   const getViewportPageSize = React.useCallback(() => {
+    // Aways null
+    console.log(apiRef.current.unstable_getGridRenderContext());
```
flaviendelangle, line 149: WIP I suppose

packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts
```diff
@@ -149,21 +145,15 @@ export function useGridDimensions(
   );
 
   const getViewportPageSize = React.useCallback(() => {
-    const dimensions = apiRef.current.getRootDimensions();
-
-    if (!dimensions) {
-      return 0;
-    }
+    const renderContext = apiRef.current.unstable_getRenderContext();
```
flaviendelangle, line 148: I'm not a huge fan of exposing virtualization internals upper in the DataGrid component tree.
But I don't have a good solution either.

If @m4theushw has a better idea.
Otherwise, I'm fine with merging as is and taking time later to improve.

m4theushw: Actually the keyboard navigation won't work as expected if `disableVirtualization=true`. With no virtualization, the last row index in the render context points to the last row in the entire row set. I'm waiting for #3667 to be merged to able to test if there're no regressions. In the worst scenario, I would keep the old behavior if `getRowHeight` is not being used.

DanailH: I tried calculating it without the `renderContext` -> the problem is that because of the variable row height you need to know the indexes of the first and last rows in the viewport to calculate the total number of visible rows. If anyone have any ideas I'm open to them.

cherniavskii:
I've resolved a conflict caused by merging https://github.com/mui-org/material-ui-x/pull/3667

packages/grid/_modules_/grid/hooks/features/dimensions/useGridDimensions.ts
m4theushw: There's a regression here. Pressing <kbd>Page Up</kbd> is jumping directly to the column header while not on the first row yet. I don't mind if we ship this small bug only affecting those users using `getRowHeight`, but keeping the old behavior for those who don't use it. I think the correct way would be to base on `rowsMeta.positions` and the current scroll position.

msedge_srHruiStdO

DanailH: Updated!

m4theushw (approved):
Well done.

DanailH:
Merged. I'll update the PR description for the release.

m4theushw:
Although we merged, there's still room for improvement. There're 3 things we could do in the future:

- Add an `apiRef.current.setRowHeight` method to the API to imperatively change the row height, similar to the one available to control the column width

- Fix keyboard navigation, e.g. <kbd>Page Up</kbd>, to use the scroll position and the position of each row

- Add a resize handle to the rows to allow to change their height. It's the same thing we have for the columns. We need `apiRef.current.setRowHeight` for that.

flaviendelangle:
I created an issue to keep track of those subjects
https://github.com/mui-org/material-ui-x/issues/3737

oliviertassinari:
A thought on the `getRowHeight` API signature: https://github.com/mui/material-ui/pull/32766#issuecomment-1154373656

PR #3219, opened 2021-11-18
[core] Split cell/row editing into different hooks

m4theushw:
One item of https://github.com/mui-org/material-ui-x/issues/3101#issuecomment-963107684

This PR splits the logic to handle cell/row editing into three hooks:

- `useGridRowEditing` contains only the API methods and handlers used by the row editing
- `useGridCellEditing` contains only the API methods and handlers used by the cell editing
- `useGridEditing` contains the common logic and calls `useGridRowEditing` and `useGridCellEditing`.

If a event was used by both row and cell editing in the past, now it's published twice. Although, only one handler is called, depending on the edit mode.


flaviendelangle (commented):
It's a lot better that way !

flaviendelangle:
Good luck for the conflits :grimacing: 
I'll approve after, for me everything is good.

packages/grid/_modules_/grid/models/api/gridEditingApi.ts
```diff
@@ -20,19 +24,28 @@ export interface GridEditRowApi {
...
+  /**
+   * @ignore - do not document.
+   */
+  unstable_setEditCellProps: (params: GridEditCellPropsParams) => void;
+}
+
+/**
+ * The row editing API interface.
+ */
+export interface GridRowEditingApi extends GridEditingSharedApi {
```
m4theushw, line 48: It extends `GridEditingSharedApi` because I want to create a separate page for row editing and another for cell editing. That way, I can add the API widget with all methods that make sense for row editing. The same for cell editing.

packages/grid/_modules_/grid/models/api/gridEditingApi.ts
```diff
@@ -69,10 +82,22 @@ export interface GridEditRowApi {
...
+   * @param {string} field The field to get the mode.
+   * @returns {GridCellMode} Returns `"edit"` or `"view"`.
+   */
+  getCellMode: (id: GridRowId, field: string) => GridCellMode;
 }
+
+/**
+ * The editing API interface that is available in the grid `apiRef`.
+ */
+export interface GridEditingApi extends GridCellEditingApi, GridRowEditingApi {}
```
flaviendelangle, line 103: Just a slight follow up on the typing breaking change
This is technically a typing breaking change for users importing the old interface.
We can keep exporting a deprecated alias `export type GridEditingApi = GridEditingApi` or consider that this is a very minor BC and do nothing more than your current PR.

packages/grid/_modules_/grid/models/api/index.ts
```diff
@@ -3,7 +3,7 @@ export * from './gridParamsApi';
 export * from './gridCoreApi';
 export * from './gridColumnApi';
 export * from './gridDensityApi';
-export * from './gridEditRowApi';
+export * from './gridEditingApi';
```
flaviendelangle, line 6: Could we not export `GridEditingSharedApi` / `GridRowEditingApi` / `GridCellEditingApi`

```suggestion
export { GridEditingApi } from './gridEditingApi';
```

We can still access the private api with the full import path.

flaviendelangle (approved):

PR #3220, opened 2021-11-19
[DataGrid] Fix focus when selecting option with Enter in the `singleSelect`

### LINKED ISSUE #2880: [DataGrid] Select inside cell doesn't trigger keyboard navigation when an item is selected via keyboard
### Duplicates

### Latest version

### Current behavior 😯

Selecting an item with <kbd>Enter</kbd> doesn't close the Select and the focus is not moved to the correct cell.

MOzZnhjFK9

### Expected behavior 🤔

In https://v4.mui.com/components/data-grid/demo

d5NlwGP7PK

It should move the focus to the cell underneath when an item is selected with <kbd>Enter</kbd>.

### Steps to reproduce 🕹

It can be reproduced in https://mui.com/components/data-grid/demo/.

### Context 🔦

This check is wrong: https://github.com/mui-org/material-ui-x/blob/998b4024d2793b673baa9f410e8b4c08c6dd7811/packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts#L58-L60

When using MUI Core v4, a `keyDown` event fired by a Select has a fake target, which doesn't satisfy the above condition: https://github.com/mui-org/material-ui/blob/c545ccab7edfdf4a44d4ec2f4bf10ebc7fd00259/packages/material-ui/src/Select/SelectInput.js#L194

In v5, it was changed to not override the original target, which made it to fall into the condition: https://github.com/mui-org/material-ui/blob/81578f0b52b469e0b0d2cb9df2ebf6820e05e6c6/packages/mui-material/src/Select/SelectInput.js#L269-L271

Flavien raised a similar issue in https://github.com/mui-org/material-ui-x/pull/2862#issuecomment-944094261.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #2880 

The solution is a hack involving manually publishing `cellNavigationKeyDown` to trigger the keyboard navigation. Another alternative would be to call `setCellFocus`, but it's too cumbersome to compute the next id to focus on. The best thing to do would be to not ignore events coming from a portal and adopt a solution like https://github.com/mui-org/material-ui-x/issues/3158#issuecomment-966644362. I'll not do this for now because it may cause regressions.


packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
```diff
@@ -53,9 +53,8 @@ export const useGridKeyboard = (apiRef: GridApiRef): void => {
 
   const handleCellKeyDown = React.useCallback(
     (params: GridCellParams, event: React.KeyboardEvent) => {
-      // The target is not an element when triggered by a Select inside the cell
-      // See https://github.com/mui-org/material-ui/issues/10534
-      if ((event.target as any).nodeType === 1 && !isGridCellRoot(event.target as Element)) {
+      // Ignore portal
```
alexfauquette, line 56: It is not really just about blocking portal. For example, keyboard navigation on ActionCell is blocked because of this condition
```suggestion
      // Do not apply shortcuts if the focus is not on the cell root component
```

alexfauquette (approved):

packages/grid/x-data-grid-generator/src/renderer/renderEditCountry.tsx
```diff
@@ -46,10 +46,8 @@ function EditCountry(props: GridRenderEditCellParams) {
   const handleChange = React.useCallback(
     (event, newValue) => {
       api.setEditCellValue({ id, field, value: newValue }, event);
-      if (!event.key) {
-        api.commitCellChange({ id, field });
-        api.setCellMode(id, field, 'view');
-      }
+      api.commitCellChange({ id, field });
+      api.setCellMode(id, field, 'view');
```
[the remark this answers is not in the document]
m4theushw: Good catch.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboard.ts
m4theushw: The implementation of `isGridCellRoot` is wrong. It's not considering that the element might not be an immediate child of the parent. That's why the keyboard navigation doesn't work in the actions cell. I'll fix this in another PR.

PR #3222, opened 2021-11-19
[DataGrid] Fix DatePicker bug by limiting years to 4 digits

### LINKED ISSUE #3221: [DataGrid]DatePickers bugs after saving a year with 5 digits
### Duplicates

### Latest version

### Current behavior 😯

If a user enters a date with 5 digits, the date picker becomes unusable.

### Expected behavior 🤔

Two options:
- The year is limited to 4 digits
- The components is still editable with more than 4 digits

### Steps to reproduce 🕹

Steps:

1. Go to https://mui.com/components/data-grid/demo/
2. Edit last column cell to set 20211
3. Validate the editing
4. Edit the same cell. It will be impossible the set a new date

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3221

HTML input allows to set more than 4 digits in a year but it seems to be a problem when setting the value so I propose to fix the maximal year to 9999. Probably need to change it around year 8000 ;)

It would probably be possible to avoid the bug without limiting years to 4 digits because filters can accept such dates. But I do not see any value for that and would take more time to find where does the bug comes from.

demo: https://deploy-preview-3222--material-ui-x.netlify.app/components/data-grid/demo/


sebastianfrey:
I have noticed, that it is also possible in the input component for filtering date fields.

alexfauquette:
I've noticed it too, but it does not make the input component buggy. Do you feel it is a bug to be able to set years with 5 digits?

sebastianfrey:
@alexfauquette Indeed, it does not make the input component buggy. But I think it does not feel natural and intuitive, when users can enter 5+ digits for year.

From my perspective it might be a bug, because in my use case I have to adopt multiple server backends using the grid component. Each of those server backends handles dates differently. That means, some of them may not fail when querying a date field with a date with more than 4 year digits, and some of them will fail.

flaviendelangle:
We have edge cases where dates with 5 digits can make sense (if I want the starting year of my favorite SF cycles for instance :smile: )
But I suppose people can always create custom `date-custom` col types with a renderer calling `GridEditDateCell` with additional props.

m4theushw:
For consistency I'm in favor of also not allowing 5-digit years in the filter value. With #3227, it will be easier to add the `max` prop.

sebastianfrey:
@flaviendelangle In my opinion the special case is 5 or more digits per year. So if someone has such a requirement, than he/she should create a custom date component. But it's up to you guys, what direction this is going.

flaviendelangle:
> For consistency I'm in favor of also not allowing 5-digit years in the filter value. With #3227, it will be easier to add the max prop.

Should it be done on this PR ?

alexfauquette:
@flaviendelangle  Il will merge #3227 and do the max digits in this one

alexfauquette:
@flaviendelangle My bad, I did it in the other PR. So filter limitation is already on master. But I did it wrong. I did not notice the distinction between `InputProps` and `inputProps`. That's corrected  in last commit

https://github.com/mui-org/material-ui-x/blob/1a4324cbc24723b768ef0544a72b09b272094c16/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx#L50

packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
```diff
@@ -117,6 +117,9 @@ export function GridEditDateCell(props: GridRenderEditCellParams & InputBaseProp
       fullWidth
       className={classes.root}
       type={isDateTime ? 'datetime-local' : 'date'}
+      inputProps={{
+        max: isDateTime ? '9999-12-31T23:59' : '9999-12-31',
```
m4theushw, line 121: Should the `inputProps` be spread here?

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -47,7 +47,10 @@ function GridFilterInputDate(props: GridFilterInputDateProps) {
 
   const InputProps = {
     ...(applying ? { endAdornment: <GridLoadIcon /> } : {}),
-    max: type === 'datetime-local' ? '9999-12-31T23:59' : '9999-12-31',
+    inputProps: {
+      max: type === 'datetime-local' ? '9999-12-31T23:59' : '9999-12-31',
+      ...inputProps,
```
m4theushw, line 52: The filter input is based on the `TextField` so users except that the props will be forwarded to the root component.

Here you're spreading the top-level `inputProps` inside `InputProps`. It's like:

`<GridFilterInputDate inputProps={{ min: '...' }} />`
-> `<TextField InputProps={{ inputProps: { min: '...' } }} />` (this PR)
-> `<TextField inputProps={{ min: '...' }} />` (how it should behave)

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -64,8 +59,15 @@ function GridFilterInputDate(props: GridFilterInputDateProps) {
         shrink: true,
       }}
       inputRef={focusElementRef}
+      InputProps={{
+        ...(applying ? { endAdornment: <GridLoadIcon /> } : {}),
+        inputProps: {
+          max: type === 'datetime-local' ? '9999-12-31T23:59' : '9999-12-31',
+          ...inputProps,
```
m4theushw, line 66: The top-level `inputProps` is still being applied inside `InputProps`. The following diff should fix this

```diff
diff --git a/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx b/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
index 7e061a7ec..e8b4f1006 100644
--- a/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
+++ b/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
@@ -11,7 +11,7 @@ export type GridFilterInputDateProps = GridFilterInputValueProps &
 export const SUBMIT_FILTER_DATE_STROKE_TIME = 500;
 
 function GridFilterInputDate(props: GridFilterInputDateProps) {
-  const { item, applyValue, type, apiRef, focusElementRef, inputProps, InputProps, ...other } =
+  const { item, applyValue, type, apiRef, focusElementRef, InputProps, ...other } =
     props;
 
   const filterTimeout = React.useRef<any>();
@@ -61,11 +61,11 @@ function GridFilterInputDate(props: GridFilterInputDateProps) {
       inputRef={focusElementRef}
       InputProps={{
         ...(applying ? { endAdornment: <GridLoadIcon /> } : {}),
+        ...InputProps,
         inputProps: {
           max: type === 'datetime-local' ? '9999-12-31T23:59' : '9999-12-31',
-          ...inputProps,
+          ...InputProps?.inputProps,
         },
-        ...InputProps,
       }}
       {...other}
     />
```

m4theushw (approved):

PR #3224, opened 2021-11-19
[DataGrid] Should not crash when a sort item uses a non-existing column

### LINKED ISSUE #3020: [DataGrid] Changing `columns` when using `sortModel` throws exception if the previously sorted column is missing
### Duplicates

### Latest version

### Current behavior 😯

When using the `<DataGrid>`, it is expected and understood that changing the `columns` prop may throw an exception if the `sortModel` prop is applying a sort on a column that no longer exists. You would see the following exception
```
Error sorting: column with field 'age' not found. 
```

However, if both the `columns` and `sortModel` props are updated at the same time to ensure the `sortModel` is appropriate for the new `columns` definition, it still throws the "error sorting" exception which is unexpected.

I suspect that the `columns` update is taking precedence and it is still trying to apply the previous `sortModel`.

### Expected behavior 🤔

Render the new columns and sort.

### Steps to reproduce 🕹

 I have created a repro test case here https://codesandbox.io/s/datagriddemo-material-demo-forked-ft5j6?file=/demo.tsx

This repro has two column definitions as `Columns1` and `Columns2` with some overlapping columns (`ID`, `First name`), but the last column is different (either `Last name` or `Age`).

Steps:
1. Start with "Column1"
2. Click on the last column "Last name" to apply a sort
3. Then click on "Set Column2"
4. See exception

I've discovered a workaround which uses `setTimeout()` to force updating the `sortModel` first and then `columns` in a separate render cycle.

```typescript
    // set sort model first
    setSortModel(DefaultSort);
    // at next render, update columns
    setTimeout(() => {
      setColumns(columns);
    });
```

### Context 🔦

I'm creating a dashboard which allows users to switch between different datasets (with different columns) in the same DataGrid.

### Your environment 🌎

<details>

```
  System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 16.9.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 7.21.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 95.0.4638.54
    Edge: Spartan (44.19041.1266.0), Chromium (95.0.1020.30), ChromiumDev (96.0.1054.2)
  npmPackages:
    @emotion/react: ^11.4.1 => 11.4.1
    @mui/icons-material: ^5.0.4 => 5.0.4
    @mui/material: ^5.0.4 => 5.0.4
    @mui/styled-engine:  5.0.1
    @mui/utils:  5.0.1
    @mui/x-data-grid: ^4.0.1 => 4.0.1
    @types/react: 17.0.27 => 17.0.27
    react: 17.0.2 => 17.0.2
    react-dom: 17.0.2 => 17.0.2
    typescript: 4.4.3 => 4.4.3
```
</details>

Using Chrome 95.0.4638.54

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #3020
It unifies the behavior with the filter. If we have an item that uses a non-existing column, we just ignore it.
This is necessary when the `sortModel` is controlled. When we change the columns, we automatically clean the `sortModel`. But when it is controlled, the update can be async and occurs after the `applySorting` triggered by the row change. So we have an `applySorting` with the new columns but the old `sortModel`, which causes a crash.

Before: https://codesandbox.io/s/datagriddemo-material-demo-forked-r27yc?file=/demo.tsx

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
+++ b/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
@@ -9,11 +9,11 @@ import { GridFeatureModeConstant } from '../../../models/gridFeatureMode';
 import { GridColumnHeaderParams } from '../../../models/params/gridColumnHeaderParams';
 import { GridRowId, GridRowTreeNodeConfig } from '../../../models/gridRows';
 import {
-  GridFieldComparator,
   GridSortItem,
   GridSortModel,
   GridSortDirection,
   GridSortCellParams,
+  GridComparatorFn,
 } from '../../../models/gridSortModel';
 import { isDesc, nextGridSortDirection } from '../../../utils/sortingUtils';
 import { isEnterKey } from '../../../utils/keyboardUtils';
@@ -30,6 +30,8 @@ import {
 import { gridRowIdsSelector, gridRowTreeDepthSelector, gridRowTreeSelector } from '../rows';
 import { useGridStateInit } from '../../utils/useGridStateInit';
 import { useFirstRender } from '../../utils/useFirstRender';
+
+type GridFieldComparator = { field: string; comparator: GridComparatorFn };
 
 /**
  * @requires useGridRows (state, event)
--- a/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
+++ b/packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
@@ -9,7 +9,7 @@ import { GridFeatureModeConstant } from '../../../models/gridFeatureMode';
 import { GridColumnHeaderParams } from '../../../models/params/gridColumnHeaderParams';
 import { GridRowId, GridRowTreeNodeConfig } from '../../../models/gridRows';
 import {
-  GridFieldComparatorList,
+  GridFieldComparator,
   GridSortItem,
   GridSortModel,
   GridSortDirection,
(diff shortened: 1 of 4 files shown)


scripts/exportsSnapshot.json
```diff
@@ -164,7 +164,6 @@
   { "name": "GridEnrichedColDef", "kind": "Type alias" },
   { "name": "GridExportFormat", "kind": "Type alias" },
   { "name": "GridFeatureMode", "kind": "Type alias" },
-  { "name": "GridFieldComparatorList", "kind": "Type alias" },
```
flaviendelangle, line 167: It is an internal type never used in anything exported.

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
```diff
@@ -150,22 +152,26 @@ export const useGridSorting = (
...
-          : column.sortComparator!;
-        return { field: column.field, comparator };
-      });
+    (sortModel: GridSortModel): GridFieldComparator[] => {
+      const comparators = sortModel
+        .map((item) => {
+          const column = apiRef.current.getColumn(item.field);
+          if (!column) {
+            return null;
+          }
```
alexfauquette, line 161: I'm wondering if a warning could help developers to catch when they have a bug in their code. But this will be kind of spam for those who control the sort model. What do you think?

alexfauquette (approved):

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
flaviendelangle: Yes I think it would spam in valid use-cases.
But more globally, we have logger system that should answer those kind of issues and that is totally unusable because it spams way to much (I don't even add logging messages anymore). I think we could address this issue when cleaning the logger.

PR #3225, opened 2021-11-19
NOT FOR MERGE - Investigate React 18 compat

flaviendelangle:
I will checkout the infra evolution from the core on another PR


m4theushw:
Do you plan to continue this effort or I can give a try? I wanted to see if React 18 Automatic Batching solves #3516.

flaviendelangle:
@m4theushw I don't think I will have the bandwidth, I let you play with it.

For #3516, probably, but we will have to support React 17 for some time so we probably won't be able to use the batching to its full capacity.

PR #3227, opened 2021-11-19
[core] Create dedicated `InputComponent`s for single select and date filters

alexfauquette:
Context: https://github.com/mui-org/material-ui-x/pull/3198#discussion_r752817538

I have a doubt about the second commit. If the `GridFilterInputValue` component is used by developers, removing the support for `singleSelect` type will be a breaking change


m4theushw:
> If the GridFilterInputValue component is used by developers, removing the support for singleSelect type will be a breaking change

We should keep the logic for the `singleSelect` inside it to not create a breaking change. However, we can add a deprecation message to redirect users to use the new component, only if they use it with a column with type `singleSelect`.

As example: https://github.com/mui-org/material-ui/blob/857c7ad547207b647094c97be69385c7d921ddc2/packages/mui-material/src/styles/adaptV4Theme.js#L4

alexfauquette:
I have put Date inputs in a dedicated component because I was looking to modify their InputProps [#3222](https://github.com/mui-org/material-ui-x/pull/3222#issuecomment-974613345)

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -33,6 +33,17 @@ export interface GridTypeFilterInputValueProps extends GridFilterInputValueProps
 
 function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldProps) {
   const { item, applyValue, type, apiRef, focusElementRef, ...others } = props;
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    ['date', 'datetime-local', 'singleSelect'].includes(type as string)
+  ) {
+    console.warn(
+      [
+        'MUI: the use of GridFilterInputValue is deprecated for singleSelect column',
```
m4theushw, line 42:
```suggestion
       `MUI: Using GridFilterInputValue with a ${type} column is deprecated.`,
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -33,6 +33,17 @@ export interface GridTypeFilterInputValueProps extends GridFilterInputValueProps
 
 function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldProps) {
   const { item, applyValue, type, apiRef, focusElementRef, ...others } = props;
+  if (
```
m4theushw, line 36: We only show the warning once.

https://github.com/mui-org/material-ui-x/blob/8ffe04e6efe09b17daece526d116df5eff5aa0bd/packages/grid/_modules_/grid/hooks/features/rows/useGridParamsApi.ts#L141-L142

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </option>
+    ) : (
+      <option key={option} value={option}>
+        {valueFormatter && option !== '' ? valueFormatter({ value: option, field, api }) : option}
+      </option>
+    ),
+  );
+};
+
+export interface GridTypeFilterInputSingleSelectProps extends GridFilterInputValueProps {
```
m4theushw, line 28: We have been using the `[ComponentName]Props` convention.

```suggestion
export interface GridFilterInputSingleSelectProps extends GridFilterInputValueProps {
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { GridFilterInputValueProps } from './GridFilterInputValueProps';
+
+export interface GridTypeFilterInputDateProps extends GridFilterInputValueProps {
+  type?: 'date' | 'datetime-local';
+}
+
+export const SUBMIT_FILTER_STROKE_TIME = 500;
+
+function GridFilterInputDate(props: GridTypeFilterInputDateProps & TextFieldProps) {
+  const { item, applyValue, type, apiRef, focusElementRef, ...others } = props;
```
m4theushw, line 15:
```suggestion
  const { item, applyValue, type, apiRef, focusElementRef, ...other } = props;
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,87 @@
...
+      setIsApplying(true);
+      filterTimeout.current = setTimeout(() => {
+        applyValue({ ...item, value });
+        setIsApplying(false);
+      }, SUBMIT_FILTER_STROKE_TIME);
+    },
+    [applyValue, item],
+  );
+
+  React.useEffect(() => {
```
m4theushw, line 38: We miss cleaning the timeout during unmount.

https://github.com/mui-org/material-ui-x/blob/8ffe04e6efe09b17daece526d116df5eff5aa0bd/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx#L82-L86

packages/grid/_modules_/grid/models/colDef/gridSingleSelectOperators.ts
```diff
@@ -16,7 +17,8 @@ export const getGridSingleSelectOperators: () => GridFilterOperator[] = () => [
         return filterItem.value === value;
       };
     },
-    InputComponent: GridFilterInputValue,
+    // InputComponent: GridFilterInputValue,
+    InputComponent: GridFilterInputSingleSelect,
     InputComponentProps: { type: 'singleSelect' },
```
m4theushw, line 22: We don't need to pass the type anymore.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,93 @@
...
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+import { unstable_useId as useId } from '@mui/material/utils';
+import { GridLoadIcon } from '../../icons/index';
+import { GridFilterInputValueProps } from './GridFilterInputValueProps';
+
+export interface GridFilterInputDateProps extends GridFilterInputValueProps {
+  type?: 'date' | 'datetime-local';
+}
+
+export const SUBMIT_FILTER_STROKE_TIME = 500;
```
flaviendelangle, line 12: I think we can re-use the variable `SUBMIT_FILTER_STROKE_TIME` of `GridFilterInputValue`
Or rename this one.
To avoid having two variables with the same name, one exported an not the other.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+function GridFilterInputDate(props: GridFilterInputDateProps & TextFieldProps) {
```
flaviendelangle, line 14: Could we export this file in `filterPanel/index.ts` ?

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { item, applyValue, type, apiRef, focusElementRef, ...other } = props;
+
+  const filterTimeout = React.useRef<any>();
+  const [filterValueState, setFilterValueState] = React.useState(item.value ?? '');
+  const [applying, setIsApplying] = React.useState(false);
+  const id = useId();
+
+  const onFilterChange = React.useCallback(
+    (event) => {
```
flaviendelangle, line 23:
```suggestion
    (event: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+function GridFilterInputSingleSelect(props: GridFilterInputSingleSelectProps & TextFieldProps) {
+  const { item, applyValue, type, apiRef, focusElementRef, ...others } = props;
+  const [filterValueState, setFilterValueState] = React.useState(item.value ?? '');
+  const [applying, setIsApplying] = React.useState(false);
+  const id = useId();
+
+  const onFilterChange = React.useCallback(
+    (event) => {
```
flaviendelangle, line 39:
```suggestion
    (event: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+        typeof column.valueOptions === 'function'
+          ? column.valueOptions({ field: column.field })
+          : column.valueOptions;
+      value = columnValueOptions
+        .map((option) => (typeof option === 'object' ? option.value : option))
+        .find((optionValue) => String(optionValue) === value);
+
+      setFilterValueState(String(value));
+
+      setIsApplying(true);
```
flaviendelangle, line 53: What is the use of `setIsApplying` is we set it to `true` then to `false` synchronously ?

alexfauquette: None. In my mind, `applyValue` could be time-consuming, and then be asynchronous. It's not the case so Yes, the setValue is useless here

flaviendelangle: If it would be async then there is an `await` missing here.
But no, `applyValue` is just running `upsertFilterItem`, which can be heavy but is always synchronous according to it signature :+1:

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </option>
+    ) : (
+      <option key={option} value={option}>
+        {valueFormatter && option !== '' ? valueFormatter({ value: option, field, api }) : option}
+      </option>
+    ),
+  );
+};
+
+export interface GridFilterInputSingleSelectProps extends GridFilterInputValueProps {
```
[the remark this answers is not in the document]
alexfauquette: Seems that we can not extend with a union of type, so we cannot do 
```ts
export interface GridFilterInputSingleSelectProps extends GridFilterInputValueProps, TextFieldProps {
  type?: 'singleSelect';
}
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
alexfauquette: I renamed it, because we are splitting components to create more independence between them

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputSingleSelect.tsx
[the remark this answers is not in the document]
flaviendelangle: In that case I think we should use a type

```tsx
export type GridFilterInputSingleSelectProps = GridFilterInputValueProps & TextFieldProps & { type?: 'singleSelect' }
```

It's not great, but exporting an incomplete interface seems worse to me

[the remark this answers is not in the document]
alexfauquette: I corrected it for the two new components. I can not do it for the`GridFilterInputValue` because the type `GridFilterInputValueProps` is already exported. I add it as a comment in the breaking change issue

flaviendelangle (approved):

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputDate.tsx
```diff
@@ -0,0 +1,92 @@
...
+      placeholder={apiRef.current.getLocaleText('filterPanelInputPlaceholder')}
+      value={filterValueState}
+      onChange={onFilterChange}
+      type={type || 'text'}
+      variant="standard"
+      InputLabelProps={{
+        shrink: true,
+      }}
+      inputRef={focusElementRef}
+      {...other}
```
m4theushw, line 67: This is not a problem here, but per convention `other` should always be last prop to be spread.

Issue #3211, opened 2021-11-17
Help us to translate


m4theushw:
### 🇸🇩 arSD
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/arSD.ts#L105
### 🇧🇬 bgBG
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/bgBG.ts#L8
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/bgBG.ts#L57
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/bgBG.ts#L58
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/bgBG.ts#L84
### 🇩🇪 deDE
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/deDE.ts#L95
### 🇪🇸 esES
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/esES.ts#L8
### 🇰🇷 koKR
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/koKR.ts#L91
### 🇳🇱 nlNL
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/nlNL.ts#L8
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/nlNL.ts#L31
https://github.com/mui-org/material-ui-x/blob/ae6b73459488d7ee96ef4d9ce0dfe10b522b419c/packages/grid/_modules_/grid/locales/nlNL.ts#L32
[...]

PR #3232, opened 2021-11-20
[l10n] Improve ko_KR locale

zzossig:
I believe `/` is the best translation. #3211

### DIFF
--- a/packages/grid/_modules_/grid/locales/koKR.ts
+++ b/packages/grid/_modules_/grid/locales/koKR.ts
@@ -88,8 +88,8 @@ const koKRGrid: Partial<GridLocaleText> = {
   footerTotalRows: '총 행수:',
 
   // Total visible rows footer text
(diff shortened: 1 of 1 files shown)


m4theushw:
@zzossig Thanks for your help. After syncing with the English locale, I noticed that there're still a couple of translations missing. I've updated #3211. If you want to do a follow-up PR we're gonna love.

https://github.com/mui-org/material-ui-x/blob/d826fc758ee01cb2a599042f7d19cfc173a123a9/packages/grid/_modules_/grid/locales/koKR.ts#L33
https://github.com/mui-org/material-ui-x/blob/d826fc758ee01cb2a599042f7d19cfc173a123a9/packages/grid/_modules_/grid/locales/koKR.ts#L96
https://github.com/mui-org/material-ui-x/blob/d826fc758ee01cb2a599042f7d19cfc173a123a9/packages/grid/_modules_/grid/locales/koKR.ts#L103

PR #3235, opened 2021-11-21
[core] Reflect the change of default branch

oliviertassinari:
A follow-up on the change of the default branch from *next* to *master*. Asked in https://mui-org.slack.com/archives/G011VC970AW/p1637339352041900

Note that I also had to update the baseline branch of Argos-CI manually in the DB:

Screenshot 2021-11-21 at 00 52 08

### DIFF
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -41,7 +41,8 @@ module.exports = {
     REACT_STRICT_MODE: reactStrictMode,
     // Set by Netlify
     GRID_EXPERIMENTAL_ENABLED: process.env.PULL_REQUEST === 'false' ? 'false' : 'true',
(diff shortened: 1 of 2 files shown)


docs/next.config.js
```diff
@@ -41,7 +41,8 @@ module.exports = {
     REACT_STRICT_MODE: reactStrictMode,
     // Set by Netlify
     GRID_EXPERIMENTAL_ENABLED: process.env.PULL_REQUEST === 'false' ? 'false' : 'true',
-    SOURCE_CODE_ROOT_URL: 'https://github.com/mui-org/material-ui-x/blob/next',
+    // #default-branch-switch
```
oliviertassinari, line 44: This is a breadcrumb we leave so the next time we change the name of the active branch, we can search it and replace all the instances blindly.

m4theushw (approved):

PR #3262, opened 2021-11-22
[test] Row checkbox selection is handled by click and space key

### LINKED ISSUE #1431: [DataGrid] Select all checkboxes doesn't support keyboard

## Current Behavior 😯

The <kbd>Enter</kbd> and <kbd>Space</kbd> key on the column header don't have any action, space is swallowed.

## Expected Behavior 🤔

The <kbd>Enter</kbd> and <kbd>Space</kbd> key on the column header triggers the select all feature

## Steps to Reproduce 🕹

1. Use <kbd>Tab</kbd> to reach a data grid
1. Use the arrow to focus the checkbox all selection
1. Use <kbd>Enter</kbd> or <kbd>Space</kbd>, nothing happens

## Context

This is a follow-up on https://github.com/mui-org/material-ui-x/pull/1289.

## Your Environment 🌎

v4.0.0-alpha.25

alexfauquette:
Fix #1431

Add tests to verify the row can be selected by pressing `Space`. This test must be in e2e test, because the onChange of the checkbox is only fired by click in react-testing-library (https://github.com/testing-library/react-testing-library/issues/156).

I also removed `Space` from the keys entering edit mode, because it was overriding two accessibility keys presented in the [Doc](https://mui.com/components/data-grid/accessibility/#selection)

keys | action
:--: | --
<kbd>Space</kbd> | Navigate to the next scrollable page
<kbd>Shift</kbd> + <kbd>Space</kbd> | Select the current row

### DIFF
--- a/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
+++ b/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
@@ -339,6 +339,26 @@ describe('<DataGrid /> - Selection', () => {
       expect(checkbox).to.have.attribute('tabindex', '0');
       expect(checkboxCell).to.have.attribute('tabindex', '-1');
     });
+
+    it('should select/unselect all rows when pressing space', async () => {
+      render(<TestDataGridSelection checkboxSelection />);
+
+      const selectAllCell = document.querySelector(
+        '[role="columnheader"][data-field="__check__"] input',
+      ) as HTMLElement;
+      selectAllCell.focus();
+
+      fireEvent.keyDown(selectAllCell, {
+        key: ' ',
+      });
+
+      expect(getSelectedRowIds()).to.deep.equal([0, 1, 2, 3]);
+      fireEvent.keyDown(selectAllCell, {
+        key: ' ',
+      });
+
+      expect(getSelectedRowIds()).to.deep.equal([]);
+    });
   });
 
   describe('props: isRowSelectable', () => {


flaviendelangle:
Is there something blocking us to test this in unit testing instead of e2e ?

alexfauquette:
Does not seem to be possible with `fireEvent`

Here a codesandbox to test it https://codesandbox.io/s/react-codesandbox-forked-72eg6?file=/src/__tests__/index.js

flaviendelangle:
What do we want to test here ?
If we want to test the `Shift` + `Space` on a row selects / unselects the row, then we must fire the event on the row, not the input right ?

```ts
    it('should unselect from last clicked cell to cell after clicked cell if clicking inside a selected range', () => {
      render(<TestDataGridSelection checkboxSelection />);
      getCell(1, 0).focus()
      fireEvent.keyDown(getCell(1, 0), { key: ' ', shiftKey: true });
      expect(getSelectedRowIds()).to.deep.equal([1]);
      fireEvent.keyDown(getCell(1, 0), { key: ' ', shiftKey: true });
    });
```

For me this works

alexfauquette:
> The <kbd>Space</kbd> key on the column header triggers the select all feature

It appears that this has been a regression, so to close the ticket, I try to put a test on it.

The <kbd>Shift</kbd> + <kbd>Space</kbd> test  works, because the event listener is the row prop `onKeyDown` which dispatches the event to the useSelection hook.

For the checkboxes, the selection by <kbd>Click</kbd> or <kbd>Space</kbd> is catched by the `onChange` prop which is not triggered by the `fireEvent.keyDown`. Only fireEvent.click can change the value of a component `<input type="checkbox" />`

alexfauquette:
Because we modified how the "select all" checkbox handle key down, I can now use `fireEvent.keyDown` to toggle the selection

flaviendelangle (approved):

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -339,6 +339,26 @@ describe('<DataGrid /> - Selection', () => {
       expect(checkbox).to.have.attribute('tabindex', '0');
       expect(checkboxCell).to.have.attribute('tabindex', '-1');
     });
+
+    it('should select/unselect all rows when pressing space', async () => {
```
oliviertassinari, line 343: No need for asnyc here

```suggestion
    it('should select/unselect all rows when pressing space', () => {
```

alexfauquette: Will be corrected in #3675

PR #3264, opened 2021-11-22
[core] Rework columns state management

flaviendelangle:
- [x] Simplify the preparation of the column state (apply custom types + apply flex width + apply pre-processors + merge with previous state).

- [x] Reduces the amount of loop on the full column list 

- [x] Apply the columns pre-processing on the raw state instead of the list. This is needed for the row grouping to avoid linear complexity when switching from single to multi grouping column.

- [x] Fix small bug: We were reseting the columns to their prop value when applying pre-processing so we were loosing the updates via `setColumnWidth`, `setColumnIndex`, `updateColumns`, ...


alexfauquette:
I do not understand what are `all` and `lookup` in the columnsState. In which feature are they used?

flaviendelangle:
`all` is the list of the column fields in the order they are displayed
`lookup` is object with the fields as key and the colDef as value.
They are the main source of truth for the columns.

With `allGridColumnsSelector` we build back an array of colDef
But thanks to the lookup format we have constant complexity to access a colDef given a field.

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -377,19 +377,26 @@ export const useGridSelection = (
   }, [apiRef, isRowSelectable, isStateControlled]);
 
   const updateColumnsPreProcessing = React.useCallback(() => {
-    const addCheckboxColumn = (columns: GridColumns) => {
-      if (!props.checkboxSelection) {
-        return columns;
-      }
-
-      const groupingColumn: GridColDef = {
+    const addCheckboxColumn = (columnsState: GridColumnsRawState) => {
```
alexfauquette, line 380: It is a bit confusing to call "add..." a function that can remove the checkbox column "update" or "manage" could be more appropriate

alexfauquette (approved):

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
flaviendelangle: Indeed, fixed :+1:

PR #3266, opened 2021-11-23
[docs] Always use auto-generated apiRef documentation

flaviendelangle:
- [x] Add auto-generated API docs paragraph for pagination / sorting
- [x] Add the `Only use this API when you have no alternative. Always start from the declarative API that the grid exposes.` to all `apiRef` paragraph (I'm not sure we want this warning, but I think we should either have it everywhere or nowhere)
- [x] Create a `GridPaginationApi` interface, I will create a `useGridPagination` which calls `useGridPageSize` and `useGridPage` like for the row / cell edit split. That way outside of the pagination hook, we don't care about the split.


packages/grid/_modules_/grid/hooks/features/pagination/gridPaginationInterfaces.ts
```diff
@@ -0,0 +1,33 @@
+export interface GridPaginationState {
```
flaviendelangle, line 1: I'm not sure how we should name this file
But I like the idea of having a single file with all the interfaces of a given feature hook.

docs/src/pages/components/data-grid/pagination/pagination.md
```diff
@@ -66,18 +66,9 @@ The following demo displays 200 rows per page:
...
 > ⚠️ Only use this API when you have no alternative. Always start from the declarative API that the grid exposes.
 
-- `setPageSize`: Set the number of rows in one page.
-- `setPage`: Set the displayed page.
-- `onPageChange`: Callback fired after a new page has been displayed.
-- `onPageSizeChange`: Callback fired after the page size was changed.
-
-Below is an example of how you can reset the page using the imperative `setPage` method.
-
-{{"demo": "pages/components/data-grid/pagination/ApiRefPaginationGrid.js", "bg": "inline", "disableAd": true}}
```
flaviendelangle, line 80: I removed this example because I don't think it brings a lot of value.

docs/src/pages/components/data-grid/editing/editing.md
```diff
@@ -248,6 +248,8 @@ The following events can be imported and used to customize the row edition:
 
 ## apiRef [<span class="pro"></span>](https://mui.com/store/items/material-ui-pro/)
 
+> ⚠️ Only use this API when you have no alternative. Always start from the declarative API that the grid exposes.
```
m4theushw, line 251:
```suggestion
> ⚠️ Only use this API as the last option. Give preference to the props to control the grid.
```

x other warnings.

flaviendelangle: Done :+1:

packages/grid/_modules_/grid/hooks/features/pagination/gridPaginationInterfaces.ts
```diff
@@ -0,0 +1,33 @@
...
+  setPageSize: (pageSize: number) => void;
+}
+
+/**
+ * The apiRef methods handled by `useGridPage`
+ */
+export interface GridPageApi {
+  /**
+   * Sets the displayed page to the value given by `page`.
+   * @param {number} page The new page number
```
m4theushw, line 25:
```suggestion
   * @param {number} page The new page number.
```

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/pagination/gridPaginationInterfaces.ts
flaviendelangle: Done :+1:

PR #3270, opened 2021-11-23
[DataGrid] Export option can be disable

### LINKED ISSUE #3064: [DataGrid] Allow export options to be removed from `GridToolbarExport`
### Order ID 💳

#31706

### Duplicates

### Latest version

### Summary 💡

In the datagrid we would like to enable csv export only and disable the print export (it only confuses our users). I've created a custom toolbar, with the `<GridToolbarExport />`, but can't find any options to select which exports I want to enable. Is there a way?

### The problem in depth 🔍

_No response_

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
 System:
    OS: macOS 11.6
  Binaries:
    Node: 15.13.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 7.7.6 - /usr/local/bin/npm
  Browsers:
    Chrome: 95.0.4638.54
    Edge: Not Found
    Firefox: 87.0
    Safari: 14.1.2
  npmPackages:
    @emotion/react: ^11.4.1 => 11.5.0 
    @mui/core:  5.0.0-alpha.51 
    @mui/icons-material: ^5.0.4 => 5.0.4 
    @mui/material: ^5.0.4 => 5.0.4 
    @mui/private-theming:  5.0.1 
    @mui/styled-engine:  5.0.1 
    @mui/styles: ^5.0.1 => 5.0.1 
    @mui/system:  5.0.4 
    @mui/types:  7.0.0 
    @mui/utils:  5.0.1 
    @mui/x-data-grid-pro: ^5.0.0-beta.5 => 5.0.0-beta.5 
    @mui/x-license-pro:  5.0.0-beta.5 
    @types/react: ^17.0.25 => 17.0.31 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    styled-components: ^5.3.1 => 5.3.3 
    typescript: ^4.4.3 => 4.4.4 

```
</details>

alexfauquette:
Fix #3064


flaviendelangle:
Nice
Could you also update the condition of the `null` export inn `GridToolbar` which totally misses the export part. The current behavior is kind of broken but now we can fix it.

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -34,6 +34,12 @@ export interface GridCsvExportOptions {
    * @default true
    */
   includeHeaders?: boolean;
+
```
m4theushw, line 37: Could you also remove the blank line after `allColumns`?

```suggestion
```

packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
```diff
@@ -85,6 +85,10 @@ const GridToolbarExport = React.forwardRef<HTMLButtonElement, GridToolbarExportP
       }
```
m4theushw, line 85: Could you clean a bit the types in this type? There're some redundant interfaces. 

I was thinking in this direction:

```diff
diff --git a/packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx b/packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
index 4c6d88235..e9e5558f4 100644
--- a/packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
+++ b/packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
@@ -15,15 +15,10 @@ import {
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
 import { gridClasses } from '../../gridClasses';
 
-interface GridExportFormat {
-  format: ExportTypes;
-  formatOptions?: GridCsvExportOptions;
-}
-
-type GridExportFormatOption = GridExportFormat;
-
-type GridExportOption = GridExportFormatOption & {
+type GridExportOption = {
   label: React.ReactNode;
+  format: ExportTypes;
+  formatOptions?: GridCsvExportOptions | GridPrintExportOptions;
 };
 
 export interface GridToolbarExportProps extends ButtonProps {
@@ -64,10 +59,10 @@ const GridToolbarExport = React.forwardRef<HTMLButtonElement, GridToolbarExportP
     const handleExport = (option: GridExportOption) => () => {
       switch (option.format) {
         case 'csv':
-          apiRef.current.exportDataAsCsv(option.formatOptions);
+          apiRef.current.exportDataAsCsv(option.formatOptions as GridCsvExportOptions);
           break;
         case 'print':
-          apiRef.current.exportDataAsPrint(option.formatOptions);
+          apiRef.current.exportDataAsPrint(option.formatOptions as GridPrintExportOptions);
           break;
         default:
           break;
```

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -79,6 +85,11 @@ export interface GridPrintExportOptions {
    * Provide Print specific styles to the print window.
    */
   pageStyle?: string | Function;
+  /**
+   * If `true`, the Print export option will be removed from the GridToolbarExport menu.
+   * @default false
+   */
+  disable?: boolean;
```
m4theushw, line 92: Maybe it's too premature, but this attribute could be extract into an interface so every export option must inherit it and support `disable`.

alexfauquette: Done in this commit: https://github.com/mui-org/material-ui-x/pull/3270/commits/9cbc962c80c04174791a5b5489f8d0b4f5964e7b

I took the common options: 
- fields
- allColumns
- disable

The only problem is for documentation comment, in which I have to replace "Print export" and "CSV export" by "this export"

packages/grid/_modules_/grid/models/gridExport.ts
```diff
@@ -1,7 +1,28 @@
...
+   * This should only be used if you want to restrict the columns exports.
+   */
+  fields?: string[];
+  /**
+   * If `true`, the hidden columns will also be exported.
+   * @default false
+   */
+  allColumns?: boolean;
+  /**
+   * If `true`, this export option will be removed from the GridToolbarExport menu.
```
flaviendelangle, line 16: What should be the behavior if the user calls `exportDataAsCsv` with `disable: true` ? (same for print)
Should be raise an error ? Ignore it as you currently do ? Skip the export ?

The most coherent behavior would probably be to split the interfaces between the imperative method and the React component, to only have `disable` on  the React component.

This would be an argument in favor of `csvOptions={false}` instead of `csvOptions={{ disable: false }}`

alexfauquette: Yes, if `GridCsvExportOptions` is used by imperative methods, it does not make sens to put user interface customisation in it.

`disable` could be defined in type `GridExportOption`. But it does not make sens to add an option `disable` that will disable all the other options.

Does someone else has an argument in favour of the `disable` logic?

m4theushw: I'm not very in favor of `csvOptions={false}`, since the prop was meant to be an object initially. We have been using slots to allow simple customizations without forcing to override an entire component. For instance, the options of the column menu can be easily changed doing that: https://mui.com/components/data-grid/components/#column-menu. The same approach could be done here maybe. It would even allow users to add their export options.

m4theushw: My idea is not exactly as you proposed. We add a new `toolbarExportButton` slot which has by default:

```tsx
<GridToolbarExportContainer>
  <GridCsvExportItem options={...} />
  <GridPrintExportItem options={...} />
  <MyCustomExportItem options={...} />
</GridToolbarExportContainer>
```

If the user wants to disable one of item, he removes the respective option. My argument against passing `false` is that the options prop is meant to be object. We're changing its meaning depending on its value. If the slot approach is too overkill, we could also add an `allowedFormats` prop to the `GridToolbarExportContainer`.

m4theushw: Or `disableToolbarButton`

packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
```diff
@@ -15,20 +15,25 @@ import {
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
 import { gridClasses } from '../../gridClasses';
 
-interface GridExportFormat {
-  format: ExportTypes;
-  formatOptions?: GridCsvExportOptions;
-}
-
-type GridExportFormatOption = GridExportFormat;
+type GridExportDisplayOptions = {
```
m4theushw, line 18: Prefer interfaces always.

```suggestion
interface GridExportDisplayOptions {
```

Same for `GridExportOption`.

packages/grid/_modules_/grid/components/toolbar/GridToolbarExport.tsx
```diff
@@ -15,20 +15,25 @@ import {
...
+  disableToolbarButton?: boolean;
+};
 
-type GridExportOption = GridExportFormatOption & {
+type GridExportOption = {
   label: React.ReactNode;
+  format: ExportTypes;
+  formatOptions?:
+    | (GridCsvExportOptions & GridExportDisplayOptions)
+    | (GridPrintExportOptions & GridExportDisplayOptions);
```
m4theushw, line 31: It fits in one line.

```suggestion
  formatOptions?: (GridCsvExportOptions | GridPrintExportOptions) & GridExportDisplayOptions;
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -52,6 +52,14 @@ You can provide a [`valueFormatter`](/components/data-grid/columns/#value-format
 />
 ```
 
+### Remove export button
```
m4theushw, line 55:
```suggestion
### Remove the export button
```

docs/src/pages/components/data-grid/export/export.md
```diff
@@ -106,6 +114,14 @@ You can provide a [`valueFormatter`](/components/data-grid/columns/#value-format
 />
 ```
 
+### Remove print button
```
m4theushw, line 117:
```suggestion
### Remove the print button
```

m4theushw (approved):
I left some minor suggestions. Otherwise, looks good.

Issue #2991, opened 2021-10-26
[docs] Remove dependency on @mui/styles


DanailH:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Update the documentation examples in all `DataGrid` pages as well as in the `x-grid-data-generator` package, storybook examples, and test cases to use `styled` (the new way of customizing styles from v5) instead of `makeStyles`.

Remove all `@mui/styles` imports.

https://github.com/mui-org/material-ui-x/blob/b7fb0169f74cec58817e85ff167d197a244d6652/package.json#L72

https://github.com/mui-org/material-ui-x/blob/6efc5f9f6f49e9a0566e048b2358cf204abc5753/docs/src/pages/components/data-grid/events/CatalogOfEventsNoSnap.js#L13

PR #3274, opened 2021-11-24
[docs] Replace @mui/styles on demos

m4theushw:
Related to #2991

This PR replaces `@mui/styles` from all demos and uses the `sx` prop or styled components. I'll open other PRs to migrate `x-grid-data-generator` and other parts still using the legacy styling solution.

The differences pointed by argos are actually the correct screenshots now. They were wrong in the past because the custom styles, which were based on jss, were being inserted before the DataGrid styles, based on emotion, so the emotion styles always won. Now, everything is based on emotion so custom styles work correctly.

Before: _(from the quick filter demo)_

image

After: 

image


docs/src/pages/components/data-grid/columns/RenderExpandCellGrid.js
```diff
@@ -1,29 +1,11 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
+import Box from '@mui/material/Box';
 import Typography from '@mui/material/Typography';
 import Paper from '@mui/material/Paper';
 import Popper from '@mui/material/Popper';
-import { createStyles, makeStyles } from '@mui/styles';
 import { DataGrid } from '@mui/x-data-grid';
 
-const useStyles = makeStyles(() =>
```
mnajdova, line 9: I would have just converted this to a styled component, but this works too. I usually made this decision but the length of the `sx` prop, if it doesn't fit in one line, I would use styled().

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
```diff
@@ -1,23 +1,15 @@
 import * as React from 'react';
-import { makeStyles } from '@mui/styles';
+import Box from '@mui/material/Box';
 import { DataGrid, useGridApiContext, useGridState } from '@mui/x-data-grid';
 import { useDemoData } from '@mui/x-data-grid-generator';
 import Pagination from '@mui/material/Pagination';
 
-const useStyles = makeStyles({
-  root: {
-    display: 'flex',
```
mnajdova, line 9: Was this not needed?

docs/src/pages/components/data-grid/demo/FullFeaturedDemo.js
```diff
@@ -306,16 +274,33 @@ export default function FullFeaturedDemo() {
     });
   };
 
+  const DataGridComponent = isAntDesign ? AntDesignStyledDataGridPro : DataGridPro;
+
   return (
-    <div className={classes.root}>
+    <Box
+      sx={{
```
mnajdova, line 281: Again I would use here `styled()` as the `sx` prop value is too long.

docs/src/pages/components/data-grid/editing/ConditionalValidationGrid.js
```diff
@@ -95,14 +71,22 @@ export default function ConditionalValidationGrid() {
...
-    <div style={{ height: 400, width: '100%' }}>
-      <DataGridPro
-        apiRef={apiRef}
-        className={classes.root}
-        rows={rows}
-        columns={columns}
-        editMode="row"
-      />
-    </div>
+    <Box
```
mnajdova, line 74: Prefer `styled()`

docs/src/pages/components/data-grid/editing/ValidateServerNameGrid.js
```diff
@@ -70,15 +48,28 @@ export default function ValidateServerNameGrid() {
   }, []);
 
   return (
-    <div style={{ height: 400, width: '100%' }}>
+    <Box
```
mnajdova, line 51: Prefer `styled()`

docs/src/pages/components/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -78,8 +56,22 @@ function QuickSearchToolbar(props: QuickSearchToolbarProps) {
             </IconButton>
           ),
         }}
+        sx={{
+          width: {
+            xs: 1,
+            sm: 'auto',
+          },
+          m: (theme) => theme.spacing(1, 0.5, 1.5),
```
mnajdova, line 64:
```suggestion
          mt: 1,
          mx: 0.5,
          mb: 1.5,
```
One alternative, not sure if it is worthed.

mnajdova (approved):
Looks good, I left some comments on places where I think `styled()` is more suitable, but it's not necessary to be changed

docs/src/pages/components/data-grid/components/CustomPaginationGrid.js
m4theushw: It does nothing. In other demos we don't use `display: 'flex'`.

m4theushw:
@mnajdova Thanks for the suggestions.

PR #3277, opened 2021-11-24
[DataGridPro] Allow to group rows based on column value

### LINKED ISSUE #212: [DataGridPro] Implement Row Grouping by Column
### Benchmark

- https://www.ag-grid.com/javascript-grid-grouping/
- https://material-table.com/#/docs/features/grouping
- https://www.telerik.com/kendo-react-ui/components/grid/grouping/
- https://ej2.syncfusion.com/react/demos/#/material/grid/aggregate-group
- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/grouping/
- http://tabulator.info/docs/4.5/group
- https://www.infragistics.com/products/ignite-ui-react/react/components/data-grid-row-grouping.html
- https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--grouping
- https://adazzle.github.io/react-data-grid/docs/examples/basic-grouping

flaviendelangle:
Closes #212

[Docs preview](https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#row-grouping)

## Feature summary

- Use `props.experimentalFeatures.groupingColumns: true` to unlock the feature

- Use `props.groupingColumnsModel` / `props.onGroupingColumnsModelChange` to control

- Use `props.initialState.groupingColumns.model` to init (I did not store the model directly in `initialState.groupingColumns` because I feel like we will have additional stuff in this sub-state one day)

- Use the column menu to group / ungroup columns

- Use `props.groupingColumnMode?: 'single' | 'multiple'` to create one grouping column with all the grouping criteria or one grouping column per grouping criteria (if you have an idea for a better name to avoid confusion with our `xxxMode?: 'client' | 'server'`. I don't want a boolean because AG-Grid has a 3rd mode with no grouping col and with a boolean we would be blocked)

- Use `props.groupingColDef.leafField` to render a column cell on the leafs of the grouping column
- Use `props.disableGroupingColumns` to fully disable the feature

- Use `colDef.canBeGrouped: false` to block grouping on some columns

- Use `colDef.keyGetter` to transform your cell value (post `valueGetter`) into a serializable value

## Changelog

- 🚀 Introduce the [grouping columns](https://mui.com/components/data-grid/group-pivot/#grouping-columns feature (#3277) @flaviendelangle

  **Add GIF**

## Extracted pull requests

- #3264
- #3314 
- #3315 
- #3316
- #3318
- #3366
- #3368
- #3380


packages/grid/_modules_/grid/hooks/features/groupingColumns/gridGroupingColumnsUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+  GridValueGetterSimpleParams,
+} from '../../../models';
+import { GridFilterState } from '../filter';
+import { GridComponentProps } from '../../../GridComponentProps';
+
+export const GROUPING_COLUMN_SINGLE = '__row_group_by_columns_group__';
+
+export const GROUPING_COLUMNS_FEATURE_NAME = 'grouping-columns';
+
+export const getGroupingColDefFieldFromGroupingCriteriaField = (
```
flaviendelangle, line 9: Should we export this utility to avoid users hard-coding there grouping column field for `sortModel` and `filterModel` ?

packages/grid/_modules_/grid/hooks/features/groupingColumns/useGridGroupingColumns.tsx
```diff
@@ -0,0 +1,480 @@
...
+        groupingColumnsModel.forEach((groupingCriteria) => {
+          const { key } = getCellGroupingCriteria({
+            row,
+            id: rowId,
+            colDef: columnsLookup[groupingCriteria],
+          });
+          const groupingFieldsDistinctKeys = distinctValues[groupingCriteria];
+
+          if (key != null && !groupingFieldsDistinctKeys.map[key.toString()]) {
+            groupingFieldsDistinctKeys.map[key.toString()] = true;
```
flaviendelangle, line 161: We use a map to avoid the linear complexity of looking if the key is already registered.

packages/grid/_modules_/grid/hooks/features/groupingColumns/useGridGroupingColumns.tsx
```diff
@@ -0,0 +1,483 @@
...
+        groupingColumnsModel.forEach((groupingCriteria) => {
+          const { key } = getCellGroupingCriteria({
+            row,
+            id: rowId,
+            colDef: columnsLookup[groupingCriteria],
+          });
+          const groupingFieldsDistinctKeys = distinctValues[groupingCriteria];
+
+          if (key != null && !groupingFieldsDistinctKeys.lookup[key.toString()]) {
+            groupingFieldsDistinctKeys.lookup[key.toString()] = true;
```
flaviendelangle, line 162: We use a lookup to avoid the linear complexity of looking if the key is already registered.

alexfauquette:
That's an impressive feature 🥳️

I played a bit with the preview:

- I see paragraphs about how to disable the grouping. But none to explain how to group by column. Is there a button I missed?
- The second example of "[Grouping column customization](https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#grouping-column-customization)" seems to have a problem: When clicking on one of the chip the grouping update but not the header. I have to press a second time on the chip to have the correct header

https://user-images.githubusercontent.com/45398769/145402941-fd0ca73c-e012-4bd7-bfa8-40298392cf85.mp4

- The sorting button does not keep the focus the first time it is cliked

https://user-images.githubusercontent.com/45398769/145401474-c9887057-78ee-4a86-9cba-b6fefcfb9f68.mp4

flaviendelangle:
@alexfauquette There is probably a paragraph missing
You can change the grouping cols with the column menu, just like the pinning column.

mbrookes:
1. I was confused by the name "grouping columns" when I first viewed the docs. I thought perhaps it was meant to say "column grouping", but then realised its grouping rows, not columns. 

After looking at the demo, I released that it refers to the column in which the grouping tree appears, but that feels more like an implementation concern than core to what the functionality delivers.

Looking at the examples linked from the related issue, it seems "row grouping" or just "grouping" seems more common. "Group by column" or "Group by column values" might also work for the feature name, but not necessarily good for the API.

I appreciate changing this may have implications for API naming, but better to get it right now rather than change it later, especially since we're declaring this as stable in the docs.

2. I noticed that the icon shown when a group is open is upside-down compared to what is typical for a tree-view, including ours. I guess this is coming from the Tree feature, so happy to open a separate issue or PR for it.

3. It was discussed early on (regarding filtering?), but looking at this I have the same thought, namely that I'm not sure if the column menu is the most efficient way to select or deselect multiple columns for grouping, especially for a table with a large number of columns, since you have to scroll and scan for the relevant header. 

Airtable has a menu from which you can select multiple columns to group. Perhaps like filter, once the user has created a group, they could add more columns to group from the grouping column drop-down menu?

4. Filtering at the same time as grouping returns no results.

5. Unrelated bug, but while I notice it: You can't scroll the table when hovering the column headings.

6. If you group rows by a column off to the right, you get an empty table:
image

Should the table scroll back left? Or should the grouping column be pinned by default? Or something else?

7. When you stop grouping, the grouping column remains:

image

8. If you try and group in this state the grouping column disappears.

9. Grouping through `initialState` hides the grouped columns. If you create a group through the UI, the grouped column remains as a duplicate:

image

10. If you add or remove a group the unfolded state of the existing group(s) is lost.

11. Unrelated bug, but while I notice it: The keyboard focus indicator (blue outline) is shown when you click on a cell. (You don't have to tab into it).

12. Should it be possibly to select child rows of a group with all their data intact? (The row value for the grouped column is blank, so can't be selected.)


Numbered for ease of reference in any reply, not to enumerate the number of issues :D

flaviendelangle:
> I appreciate changing this may have implications for API naming, but better to get it right now rather than change it later,

100%
Technically, the Tree Data is also a way to group rows, that's why I called it `grouping columns`, the exact name would be `group rows by column values` or something like that but it's too long.
AG-Grid calls it *Row Grouping* so no problem for me to take that name.

> I noticed that the icon shown when a group is open is upside-down compared to what is typical for a tree-view, including ours. I guess this is coming from the Tree feature, so happy to open a separate issue or PR for it.

I think it's our current behavior on the Tree View which is not the most common one.
But in either case, it's very easy to modify.

> It was discussed early on (regarding filtering?), but looking at this I have the same thought, namely that I'm not sure if the column menu is the most efficient way to select or deselect multiple columns for grouping, especially for a table with a large number of columns, since you have to scroll and scan for the relevant header.

I have a panel (similar to https://www.ag-grid.com/javascript-data-grid/grouping-group-panel/) which is almost ready but I removed it from this PR to avoid making this PR more huge that it already is.

For me we have 3 main use case:

- Grids with a single meaningful grouping combination: the developer provides it when init the Grid
- Grids with a lot of  meaningful grouping combination: the developer enables the Panel and let the user play with it
- Grids where grouping is not a major feature but users could want to group things: the developer let the user use the column menu 

In the same way, we could add a future menu in the Toolbar for grouping, if that's what you had in mind.

> Filtering at the same time as grouping returns no results.

Could you give me an exact reproduction case please ?

> If you group rows by a column off to the right, you get an empty table

Not sure it's a big deal, but we could scroll
Pinning by default I'm not a huge fan. We have a similar subject with hiding the grouped column, I let the user do it to avoid limiting the use-cases. I could add a demo where the grouping columns are automatically pinned.

flaviendelangle:
> The second example of "Grouping column customization" seems to have a problem: When clicking on one of the chip the grouping update but not the header. I have to press a second time on the chip to have the correct header

@alexfauquette It is the same bug as #3328 
I'll do a PR :+1:

flaviendelangle:
> 10. If you add or remove a group the unfolded state of the existing group(s) is lost.

Yes, adding / removing a group fully regenerates the tree, so we lost any expansion information.

It is somehow related to #3377, I need to improve the tree management to allow partial update, not sure how hard it will be for the groups.

> 12. Should it be possibly to select child rows of a group with all their data intact? (The row value for the grouped column is blank, so can't be selected.)

Not sure to understand this one, could you elaborate please ?

mbrookes:
> Technically, the Tree Data is also a way to group rows

Yeah, I realised later. I guess the APIs need to be consistent between them, so...

> I think it's our current behavior on the Tree View which is not the most common one.

At some point I benchmarked it thoroughly (sorry, can't find it on GitHub), so I'm confident that it is. Here are a couple of examples that I see in front of me without looking further:

### VS Code:
image

#### Finder:
image

> But in either case, it's very easy to modify.

Yep, happy to fix it if you agree.

> we could add a future menu in the Toolbar for grouping

I think that's what Airtable does, but open to alternatives. I didn't fully grok your suggestion – I guess I'd need to see it (or a mockup).

>> Filtering at the same time as grouping returns no results.

> Could you give me an exact reproduction case please ?

Sure, just go to the [first example](https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#grouping-columns-definition), and apply a filter:

image










"Total rows" shows the number of groups, not the number of rows.

mbrookes:
>> Should it be possibly to select child rows of a group with all their data intact? (The row value for the grouped column is blank, so can't be selected.)

> Not sure to (sic) understand this one, could you elaborate please ?

Sure:
image

No company or director values in the selected data...

mbrookes:
image

Total rows reflects the number of groups, not the number of rows.

mbrookes:
image

Should the column title for multiple groupings in a single column reflect the "path", e.g. "Company > Director" or "Company / Director" rather than just "Group"?

flaviendelangle:
> At some point I benchmarked it thoroughly (sorry, can't find it on GitHub), so I'm confident that it is. Here are a couple of examples that I see in front of me without looking further:

OK I forgot to sync with the Tree Data, your behavior is the one I have implemented on the Tree Data, I'll fix it :+1: 

> "Total rows" shows the number of groups, not the number of rows.

This is the "expected" behavior in my code.
The whole pagination is based on top level rows (<=> group of the 1st grouping critieria for this PR)
But it's a very subjective behavior, and maybe the expected behavior is not the same for Tree Data and Grouping Columns on that one.

mbrookes:
> In the demo I manually hide them because I feel like lots of people will want to hide them.

Gotcha.

> But maybe we should not hide them on the first demos to display a use-case as close as possible from the default one.

Sounds like a plan.

> Do you think we can release without it (like for the drag&drop panel) ? To avoid making this PR heavier.

Yes, it can be left for future consideration.

mbrookes:
> AG-Grid calls it Row Grouping so no problem for me to take that name.

I think that would be better

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,178 @@ title: Data Grid - Group & Pivot
 
 <p class="description">Use grouping, pivoting, and more to analyze the data in depth.</p>
 
+## Row grouping [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
+
+Use grouping columns to group the rows according to one or several columns value</p>
```
mbrookes, line 11:
```suggestion
Use row grouping to group the rows according to one or several column values</p>
```
Thanks for changing the name. There are a few more leftovers like this, so I won't try to suggest them all.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,178 @@ title: Data Grid - Group & Pivot
 
 <p class="description">Use grouping, pivoting, and more to analyze the data in depth.</p>
 
+## Row grouping [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
```
mbrookes, line 9:
```suggestion
## Row grouping [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
```
Since this is a sort of superset of the tree view, should it come after that section? Or perhaps have its own page? (The page is quite long now).

flaviendelangle: I don't see the row grouping as a superset of the Tree Data but as a different way of grouping
On the Tree Data, each rows gives it's position in the Tree, the groups are real rows.
On the Row Grouping, the position is deduced from the other rows, the groups are fake rows.
[AG Grid](https://www.ag-grid.com/javascript-data-grid/grouping/) is putting "Row Grouping" before, I think because it is the most complete / generic feature and the Tree Data doc refers a lot to the "Row Grouping" doc for shared behavior.

For the split, I agree that the page is long. I propose we wait for https://github.com/mui-org/material-ui/issues/30091 and then we split the big pages like the Filtering / Columns / Group Pivot pages.
Unless it's really to long for you in which case we can split it now. Just we won't have the additional depth on the sidebar to create a "Group & Pivot" section containing all the new pages extracted from the `group-pivot` page.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: I forgot most of those on the doc page indeed.
Done :+1:

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -0,0 +1,475 @@
...
+  });
+
+  const isFirstRender = React.useRef(true);
+  React.useEffect(() => {
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
+      return;
+    }
+
+    updateRowGrouping();
```
m4theushw, line 211: I noticed the same in `useTreeData`.

```suggestion
    return updateRowGrouping();
```

m4theushw (commented):
I didn't dive too deep into the code but overall it's working correctly. Good job!

I noticed a small bug in https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#grouping-columns:

LJV4ObcXR3

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -83,6 +84,11 @@ export interface GridColDef {
    * @default false
    */
   editable?: boolean;
+  /**
+   * If `true`, the rows can be grouped based on this column values (pro-plan only).
+   * @default true
+   */
+  canBeGrouped?: boolean;
```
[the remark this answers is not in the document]
flaviendelangle: https://dictionary.cambridge.org/spellcheck/english/?q=groupable

The word don't exist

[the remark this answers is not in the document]
m4theushw: Most of these "able" words are not in this dictionary too. But people still use them. For instance, "groupable" is used by [Kendo UI](https://docs.telerik.com/kendo-ui/api/javascript/ui/grid/configuration/groupable#groupable). Many [books](https://books.google.com/ngrams/graph?content=groupable&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Cgroupable%3B%2Cc0#t1%3B%2Cgroupable%3B%2Cc0) also use this word.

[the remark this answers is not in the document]
mbrookes: +1 for "groupable". Consistency wins.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
mbrookes: Perhaps superset was the wrong word, but the point was that row grouping builds on the features of the tree view, so it makes sense to introduce that first to build foundational knowledge. (Not that people necessarily read the content in a linear manner, but at least to guide the reader if they happen to.

mbrookes: Happy to wait for the restructure to decide what and where to split.

flaviendelangle: Do we agree that what you call "tree view" is the Tree Data in the Data Grid doc ?

packages/grid/_modules_/grid/hooks/features/sorting/gridSortingUtils.ts
```diff
@@ -39,6 +39,7 @@ const parseSortItem = (sortItem: GridSortItem, apiRef: GridApiRef): GridParsedSo
   const getSortCellParams = (id: GridRowId): GridSortCellParams => ({
     id,
     field: column.field,
+    rowNode: apiRef.current.getRowNode(id)!,
```
flaviendelangle, line 42: I added the `rowNode` property to `GridSortCellParams` to be able to base the sorting on the `groupingField` of the row.

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
flaviendelangle: What is the advantage of the return here ?
There is no un-subscribe behavior on the row grouping.
If I add the `return`, eslint complains that the return are inconsistent.
I could return `() => {}` inside the `if` above or disable the lint rule for this line but I don't think it's interesting to do it.

packages/grid/_modules_/grid/models/gridIconSlotsComponent.ts
```diff
@@ -87,4 +87,12 @@ export interface GridIconSlotsComponent {
    * Icon displayed on the tree data toggling column when the children are expanded
    */
   TreeDataCollapseIcon: React.JSXElementConstructor<any>;
+  /**
+   * Icon displayed on the grouping column when the children are collapsed
```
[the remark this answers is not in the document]
flaviendelangle: Done

packages/grid/_modules_/grid/components/menu/columnMenu/GridRowGroupingMenuItems.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    return (
+      <React.Fragment>
+        <Divider />
+        {renderUnGroupingMenuItem(getGroupingCriteriaFieldFromGroupingColDefField(column.field)!)}
+      </React.Fragment>
+    );
+  }
+
+  if (isGrouped) {
```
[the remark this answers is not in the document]
flaviendelangle: image


The 1st column is the grouping column and the 2nd column is currently grouped because its field is being used as a grouping criteria by a grouping column.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
mbrookes: Yes, sorry, mixing my terminology

packages/grid/_modules_/grid/components/menu/columnMenu/GridRowGroupingMenuItems.tsx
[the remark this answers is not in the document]
flaviendelangle: But if you have ideas to improve the wording (either variable names or comment) I'm always interested

mbrookes (approved):
This is a tentative approval, which is to say, the issues I raised have been addressed either directly or indirectly (thanks @flaviendelangle!), but I'll defer to others for final approval.

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -0,0 +1,481 @@
...
+    [apiRef, props.rowGroupingColumnMode],
+  );
+
+  const checkGroupingColumnsModelDiff = React.useCallback<
+    GridEventListener<GridEvents.columnsChange>
+  >(() => {
+    const rowGroupingModel = gridRowGroupingSanitizedModelSelector(apiRef.current.state);
+    const lastGroupingColumnsModelApplied = sanitizedModelOnLastRowPreProcessing.current;
+
+    if (!isDeepEqual(lastGroupingColumnsModelApplied, rowGroupingModel)) {
```
m4theushw, line 460: The `isDeepEqual` shouldn't be necessary if reselect allowed to check only if part of the state has changed. Currently, it checks if `state === prevState` which is always false. I tried https://github.com/reduxjs/reselect#customize-equalitycheck-for-defaultmemoize to avoid an infinite loop in the variable row height but didn't get nothing concrete.

packages/grid/_modules_/grid/locales/ptBR.ts
```diff
@@ -116,6 +116,11 @@ const ptBRGrid: Partial<GridLocaleText> = {
   // treeDataGroupingHeaderName: 'Group',
   // treeDataExpand: 'see children',
   // treeDataCollapse: 'hide children',
+
+  // Grouping columns
+  // groupingColumnHeaderName: 'Group',
+  // groupColumn: name => `Group by ${name}`,
+  // unGroupColumn: name => `Stop grouping by ${name}`,
```
m4theushw, line 123:
```suggestion
  groupingColumnHeaderName: 'Grupo',
  groupColumn: name => `Agrupar por ${name}`,
  unGroupColumn: name => `Parar agrupamento por ${name}`,
```

packages/grid/x-data-grid-generator/src/useMovieData.ts
```diff
@@ -0,0 +1,284 @@
...
+  gross: number;
+  director: string;
+  company: string;
+  year: number;
+  composer: { name: string };
+  cinematicUniverse?: string;
+};
+
+const COLUMNS: GridColumns = [
+  { field: 'title', headerName: 'Title', width: 200, groupable: false },
```
m4theushw, line 14: Could this new dataset be an option of `useDemoData` like "commodities" and "employees"? I mean to have a single entry point to get sample data.

m4theushw (commented):
How did you solve the bug from https://github.com/mui-org/material-ui-x/pull/3277#pullrequestreview-834749868? If I use `checkboxSelection` I still can reorder the grouping column.

On this same topic, should we block the auto-generated rows from being selected? They are not real rows but clicking in Select All selects them too. Example: https://codesandbox.io/s/rowgroupinginitialstate-material-demo-forked-2ebgv?file=/demo.js

packages/grid/_modules_/grid/components/menu/columnMenu/GridRowGroupingMenuItems.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <React.Fragment>
+          <Divider />
+          {rowGroupingModel.map(renderUnGroupingMenuItem)}
+        </React.Fragment>
+      );
+    }
+
+    return (
+      <React.Fragment>
+        <Divider />
```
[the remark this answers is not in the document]
flaviendelangle: If I add it in the pre-processor, then I have a divider even when the `GridRowGroupingMenuItems` renders 0 item, which is not what we want.

packages/grid/_modules_/grid/components/menu/columnMenu/GridRowGroupingMenuItems.tsx
```diff
@@ -0,0 +1,116 @@
...
+  if (isGrouped) {
+    return (
+      <React.Fragment>
+        <Divider />
+        {renderUnGroupingMenuItem(column.field)}
+      </React.Fragment>
+    );
+  }
+
+  if (!column.groupable) {
```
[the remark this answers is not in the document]
flaviendelangle: We render nothing if the column is neither groupable or a grouping column.
We could do this check in the pre-processing function. But that case I would probably split the component in two:
- The items of the grouping columns
- The items of the groupable columns

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -0,0 +1,481 @@
...
+  /**
+   * PRE-PROCESSING
+   */
+  const getGroupingColDefs = React.useCallback(
+    (columnsState: GridColumnsRawState) => {
+      if (props.disableRowGrouping) {
+        return [];
+      }
+
+      const propGroupingColDef = props.groupingColDef;
```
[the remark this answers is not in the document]
flaviendelangle: OK I'm also modifying it is `useGridTreeData`

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
flaviendelangle: I think it was usefull when the model was not stored in state but derived from the columns.
Because then, modifying a column not related to grouping was causing a new reference of the selector and therefore a new useless application of the grouping process.
But with the model now stored in state it should not be useful.

packages/grid/x-data-grid-generator/src/useMovieData.ts
flaviendelangle: I moved it to another entry point to differentiate auto-generated data and fine-tuned datasets written by hand.

packages/grid/_modules_/grid/components/menu/columnMenu/GridRowGroupingMenuItems.tsx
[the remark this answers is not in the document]
m4theushw: Yeah, moving the check to the pre-processor and splitting the menu items is good. The logic is already a bit complex to render completely different items if the column is being grouped.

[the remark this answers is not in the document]
flaviendelangle: Done

flaviendelangle:
@m4theushw concerning the re-ordering
I improved my constraint in `useGridColumnReorder` to block the droping of a column between two non-reorderable columns
See https://github.com/mui-org/material-ui-x/pull/3277/commits/5dee1fb3e4850d780d51eefd12e86e036ecbd37d

I think it is better than a pre-processor here because it handles out of the box scenarios like 1 grouping column + 1 checkbox column.
Or 1 grouping column + 1 user custom non-reorderable column.

alexfauquette:
When a row is open, the width of the grouping cell is the default one, even if this width has been modified

https://user-images.githubusercontent.com/45398769/149117456-708e11b1-2df5-466f-913f-e92bb171a52d.mp4

flaviendelangle:
@alexfauquette interesting one
I have the exact same issue on the Tree Data

flaviendelangle:
https://user-images.githubusercontent.com/3309670/149138730-f383c571-fba4-4d87-beb9-f75b8a3e7e42.mp4

Fixed :+1:

alexfauquette:
It seems that `valueFormatter` is called in created rows. It does not apears in demos because they return nothing. But for developers indicating when a value is unknown, it creates a strange behavior

```js
// used in demo
valueFormatter: ({ value }) => {
  if (!value || typeof value !== 'number') {
    return value;
  }
  return `${value.toLocaleString()}$`;
}

// The one I use
valueFormatter: ({ value }) => value ? currencyFormatter.format(value) : "N/A"
```

image

flaviendelangle:
@alexfauquette I think this is the wanted behavior.
I make sure that the built-in cells do not render anything for autogenerated rows.
But the `renderCell`, `valueFormatter` are still called if user want to display something in those cells.
The behavior is the same for the Tree Data.

Does it make sense ?

alexfauquette:
Yes, that makes perfectly sens :)

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingLeafWithValue.js", "bg": "inline", "defaultCodeOpen": false}}
+
+#### Hide the descendant count
+
+You can use the `hideDescendantCount` property of the `groupingColDef` to hide the amount of descendant of a grouping row.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingHideDescendantCount.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Complex grouping value
```
joserodolfofreitas, line 104: ### Using 'keyGetter' for complex grouping value

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+
+In most scenarios, when you need to handle complex values, you provide a `valueGetter` property to your column definition.
```
joserodolfofreitas, line 106: The `keyGetter` is a new property on the column definition object.
It's analogous to the known `valueGetter`, which is used to handle objects and other complex values.

The difference between the two is that the `keyGetter` is designed to provide an specific value for grouping, without requiring you to change the displayed value (`valueGetter`).

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+    field: 'composer',
+    keyGetter: (params) => params.value.name,
+  },
+  // ...
+];
+```
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingKeyGetter.js", "bg": "inline", "defaultCodeOpen": false}}
+
+If your column also have a `valueGetter` property, the value passed to the `keyGetter` property will be the one returned by `valueGetter`.
```
joserodolfofreitas, line 123: *as a note: if your column also have a `valueGetter` property, its value is used as the parameter for the `keyGetter` method.

joserodolfofreitas (commented):
I used the documentation as my first contact with the new feature and it's looking good.
I left a few suggestions inline where I think we could explain it a bit clearer.

Besides that I'd suggest:
* Add a "single column group" as the first example.
* move the "Grouping columns" section above the "Disable the row grouping".

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+You can use the `hideDescendantCount` property of the `groupingColDef` to hide the amount of descendant of a grouping row.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingHideDescendantCount.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Complex grouping value
+
+In most scenarios, when you need to handle complex values, you provide a `valueGetter` property to your column definition.
+But sometimes, you need to keep the object format for the `renderCell` property and thus can not convert it to a serializable value in `valueGetter`.
+
+You can then provide a `keyGetter` property in your column definition to convert this object into a serializable value.
```
m4theushw, line 109:
```suggestion
In this case, pass a `keyGetter` property to the column definition to convert this object into a serializable value.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingKeyGetter.js", "bg": "inline", "defaultCodeOpen": false}}
+
+If your column also have a `valueGetter` property, the value passed to the `keyGetter` property will be the one returned by `valueGetter`.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingKeyGetterValueGetter.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Rows with missing groups
+
+If the grouping key of a grouping criteria is `null` or `undefined` for a row, the grid will consider that this row do not have a value for this group. and will inline it for those groups.
```
m4theushw, line 129:
```suggestion
If the grouping key of a grouping criteria is `null` or `undefined` for a row, the grid will consider that this row does not have a value for this group. and will inline it for those groups.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
 
 <p class="description">Use grouping, pivoting, and more to analyze the data in depth.</p>
 
+## Row grouping [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
+
+Use row grouping to group the rows according to one or several columns value</p>
```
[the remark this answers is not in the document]
m4theushw: @joserodolfofreitas To improve the review process, use the "suggestion" button to suggest text that should be replaced. That way it's clear if we should use the comment text as replacement or not. 😛

msedge_6C62YObfuk

alexfauquette:
When using it I found a limitation that could be done in another PR

The use case is the next one. I want to group films by the order of magnitude of the box office. So I define a `keyGetter` that returns for example `>$10,000,000`. The problem is that this key is used for sorting, and the order will be defined by comparing `","` and `"0"`.

So I have two solution
- let `keyGetter` return a number, which is less beautiful, but it avoids me to rewrite sorting methods
- let `keyGetter` as it is, and define the `sortComparator` to deal with these keys.

It could be interesting to provide a `keyFormatter` such that in this scenario, the `keyGetter` returns a number (allowing sorting to work prefect) and the `keyFormatter` transforms `10000000` to `>$10,000,000`.

flaviendelangle:
@alexfauquette 
You should be able to pass a custom `renderCell` on the grouping col def to have the sortable data returned by `keyGetter` and the formatting in `renderCell`.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -53,13 +238,11 @@ const rows: GridRowsProp = [
...
 
-{{"demo": "pages/components/data-grid/group-pivot/BasicTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+{{"demo": "pages/components/data-grid/group-pivot/TreeDataSimple.js", "bg": "inline", "defaultCodeOpen": false}}
 
 ### Custom grouping column
 
-Use the `groupingColDef` prop to customize the rendering of the grouping column.
-
-{{"demo": "pages/components/data-grid/group-pivot/CustomGroupingColumnTreeData.js", "bg": "inline", "defaultCodeOpen": false}}
+Same behavior as for the [Row grouping](##grouping-column-customization) except for the `leafField` and `mainGroupingCriteria` which are not applicable for the Tree Data.
```
alexfauquette, line 245: This link is pointing to himself. The title you want to refer to has the same name but is `h4`. So it does not have ancor link.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,191 @@ title: Data Grid - Group & Pivot
...
+In the example below, the `director` column can not be grouped. And in all example, the `title` and `gross` columns can not be grouped.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingColDefCanBeGrouped.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Grouping columns
+
+#### Single grouping column
+
+By default, the grid will create only one grouping column even if you have several grouping criteria.
+
```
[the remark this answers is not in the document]
flaviendelangle: Maybe:

```
By default, the grid will create only one grouping column.
If you have multiple grouping criteriaes, this column header name will be set to "Group".
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
alexfauquette: I'm not sure it is a good idea to replace [the tree demo](https://mui.com/components/data-grid/group-pivot/#custom-grouping-column) where developers learn how to completely override the grouping cell by a link to [row grouping](https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#grouping-columns) named "Custom grouping column" in which only the `headerName` prop is replaced

packages/grid/x-data-grid-generator/src/useMovieData.ts
```diff
@@ -0,0 +1,284 @@
...
+    director: 'Pierre Coffin & Kyle Balda',
+    company: 'Universal Pictures',
+    year: 2015,
+    composer: {
+      name: 'Heitor Pereira',
+    },
+  },
+];
+
+const getRowId = (row: any) => row.title;
```
m4theushw, line 276: Using the `title` as the identifier may not be a good idea. If I set `editable=true` in the Title column, every time I change a movie title it gets duplicated. Adding a hidden id column is better.

m4theushw (commented):
I'm testing the interaction between the row grouping and the editing feature. I noticed that when I change a cell value, after commit it, it closes all nodes. See this [CodeSandbox](https://codesandbox.io/s/rowgroupinginitialstate-material-demo-forked-mb97r?file=/demo.js).

msedge_CvhjxsB4PL

packages/grid/x-data-grid-generator/src/useMovieData.ts
flaviendelangle: I did not think about the editing while building it but you are right

flaviendelangle:
@m4theushw 

> I'm testing the interaction between the row grouping and the editing feature. I noticed that when I change a cell value, after commit it, it closes all nodes. See this CodeSandbox.

I suspect #3604 will fix this issue, I'll test tomorrow

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: Yeah you are probably right

packages/grid/x-data-grid-generator/src/useMovieData.ts
flaviendelangle: Done :+1:

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: I added the demo back and pointed the anchor to he h3

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: I rephrased with @m4theushw proposal first, which has the same content but with a better formulation.
I think we need to decide for this `valueGetter` => `keyGetter` chaining or not before writing the final sentence.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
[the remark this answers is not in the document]
flaviendelangle: Not sure what to replace with this sentence indeed :+1:

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: What do you mean by "chaining"?

A different topic: isn't `keyGetter` a too broad name? Putting `valueGetter` aside and there's no clue of the difference between them. Should it have "group" in the name?

flaviendelangle: We had a discussion with @joserodolfofreitas about whether the value passed to `keyGetter` should be the value returned by `valueGetter` or not.

We can rename it `groupingKeyGetter` if it seems clearer.

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -121,6 +127,12 @@ export interface GridColDef {
    * @returns {GridCellValue} The cell value.
    */
   valueGetter?: (params: GridValueGetterParams) => GridCellValue;
+  /**
+   * Function that transform a complex cell value into a key that be used for grouping the rows.
```
m4theushw, line 131:
```suggestion
   * Function that transforms a complex cell value into a key to be used for grouping the rows.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,197 @@ title: Data Grid - Group & Pivot
...
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingKeyGetter.js", "bg": "inline", "defaultCodeOpen": false}}
+
+**Note**: If your column also have a `valueGetter` property, its value is used as the parameter for the `keyGetter` method.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingKeyGetterValueGetter.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Rows with missing groups
+
+If the grouping key of a grouping criteria is `null` or `undefined` for a row, the grid will consider that this row does not have a value for this group. and will inline it for those groups.
```
m4theushw, line 135: We might need to change this behavior for #213. I see rows with `null` or `undefined` as another group, often times called "N.A.", "Empty", "Other", etc.

DevExpress groups them not matter the value. See https://codesandbox.io/s/gracious-ride-fo79o?file=/demo-data/generator.js

flaviendelangle: https://www.ag-grid.com/javascript-data-grid/grouping-unbalanced-groups/

AG-Grid does not and suggest to use an arbitrary value to represent this groupable nullish values.


What would be blocking for #213 ?

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: > Suppose we have a "City" column composed of city + state (e.g. "Los Angeles, CA"), then it could be grouped either by the city name or the state.

It adds an additional layer of complexity.
Nothing impossible to manage, we could add a `groupingCriteria` property on `GridColDef`.
But I would wait for users needs for that one.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: > What would be blocking for #213 ?

I suppose that the aggregated values will be displayed in the group row or at the footer of the current group. If these rows don't belong to a group, there's no place to display the aggregated values.

m4theushw (approved):

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: Not sure to follow you here.
My point was just that AG-Grid consider `null` and `undefined` as non-groupable values. Which enable the creation of unbalanced group (cf my link). In the Tree Structure, they are still in a group, the group with "no value" for this criteria.

For instance if you group by year and director and some movies don't have a director.
Then the movies with a director are in the groups:
- `[{ groupingField: 'year', groupingValue: 2020 }, { groupingField: 'director', groupingValue: 'A' }]`
- `[{ groupingField: 'year', groupingValue: 2020 }, { groupingField: 'director', groupingValue: 'B' }]`
- ...
- 
And the movies without a director are in the groups:
- `[{ groupingField: 'year', groupingValue: 2020 }]
- `[{ groupingField: 'year', groupingValue: 2019 }]
- ...

And for Aggregation, for me we could store the aggregated values on the node of the parent
Which would make us create a parent node for the top level rows (which don't exist right now but could easily be added).

As for where those values should be rendered, as you said AG-Grid has 2 places:
- The footer: https://www.ag-grid.com/javascript-data-grid/aggregation/, which looks like a row below the last row of the group
- The grouping row: https://www.ag-grid.com/javascript-data-grid/aggregation/

But as every row is in a group, I don't think unbalanced groups are an issue here.

m4theushw: In the link you added it doesn't group rows with `null` or `undefined`. In the "sum(Val1)" column for these rows there's the raw values instead. Anyway, it's a small detail that I would expect that grid would handle automatically, but we can keep the same behavior as AG Grid.

packages/grid/_modules_/grid/hooks/features/rowGrouping/useGridRowGrouping.tsx
m4theushw: I don't have a solution yet. What I noticed is that any selector, that has a result function, the reference to its value will change on each rerender, because the reference to the whole state (`apiRef.current.state`) has changed.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
joserodolfofreitas: I agree the keyGetter as a name could be improved with "groupingKeyGetter", or even better with "groupingValueGetter", so the correlation with the valueGetter usage is more clear.

flaviendelangle: In which case I guess we are removing the chaining and instead passing the raw value to `groupingValueGetter` right ?

joserodolfofreitas: Yeah, I think that would make the most sense, the chaining doesn't feel like an intuitive assumption to make, in the sense that users who didn't read a specific line on the docs might get caught off guard.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,194 @@ title: Data Grid - Group & Pivot
 
 <p class="description">Use grouping, pivoting, and more to analyze the data in depth.</p>
 
+## Row grouping [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
+
+For when you need to group rows based on repeated column values, and/or custom functions.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingSingleGroupingCol.js", "bg": "inline", "defaultCodeOpen": false}}
```
joserodolfofreitas, line 13: @flaviendelangle Can we have the first demo with a single criteria?
I'd then add a simple description for it like this:
"On the following example, we're grouping all movies based on their production **company**"

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,194 @@ title: Data Grid - Group & Pivot
...
+#### Single grouping column
+
+By default, the grid will display a single column holding all grouped columns.
+If you have multiple grouped columns, this column name will be set to "Group".
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingSingleGroupingCol.js", "bg": "inline", "defaultCodeOpen": false}}
+
+#### Multiple grouping column
+
+To display a column for each grouping criterion, set the `rowGroupingColumnMode` prop to `multiple.
```
joserodolfofreitas, line 62:
```suggestion
To display a column for each grouping criterion, set the `rowGroupingColumnMode` prop to `multiple`.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -6,6 +6,194 @@ title: Data Grid - Group & Pivot
...
+
+By default, the grouped rows display no value on their grouping columns' cells. We're calling those cells "leaves".
+
+If you want to display some value, you can provide a `leafField` property to the `groupingColDef`.
+
+{{"demo": "pages/components/data-grid/group-pivot/RowGroupingLeafWithValue.js", "bg": "inline", "defaultCodeOpen": false}}
+
+#### Hide the descendant count
+
+Use the `hideDescendantCount` property of the `groupingColDef to hide the number of descendants of a grouping row.
```
joserodolfofreitas, line 90:
```suggestion
Use the `hideDescendantCount` property of the `groupingColDef` to hide the number of descendants of a grouping row.
```

joserodolfofreitas (commented):
The docs are looking very solid.
I've noticed just a couple details and I'm suggesting a single column as the first example because:
1) it's a gentler introduction
2) It doesnt replace the headerName with "Group" in contrast with the next example that does it. The explanation for this difference comes then right on the next section "Grouping Columns", which feels like a good introduction sequence.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
flaviendelangle: You are right, done :+1:

mbrookes:
👏  👏  👏

PR #3281, opened 2021-11-24
[DataGrid] Add support for `sx` prop

m4theushw:
We forgot to add support for the `sx` prop in #1902.

Preview: https://deploy-preview-3281--material-ui-x.netlify.app/components/data-grid/style/#using-the-sx-prop


flaviendelangle (approved):
Nothing to add here

PR #3289, opened 2021-11-27
[DataGrid] Fix column menu position when closing

m4theushw:
Preview: https://deploy-preview-3289--material-ui-x.netlify.app/components/data-grid/#commercial-version

Related to https://github.com/mui-org/material-ui-x/issues/1712#issuecomment-963601382

Before:

ZzE2NNiKJM

After:

V3ioEHabt8


packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -150,12 +151,22 @@ function GridColumnHeaderItem(props: GridColumnHeaderItemProps) {
     ariaSort = sortDirection === 'asc' ? 'ascending' : 'descending';
   }
 
+  React.useEffect(() => {
+    if (!showColumnMenuIcon) {
+      setShowColumnMenuIcon(columnMenuOpen);
```
flaviendelangle, line 156: To only trigger this effect when the `columnMenuOpen` prop changes, should we remove this `!showColumnMenuIcon`
Updating from `true` to `true` is not a problem no ?

packages/grid/_modules_/grid/components/menu/GridMenu.tsx
```diff
@@ -79,6 +80,19 @@ const GridMenu = (props: GridMenuProps) => {
     prevTarget.current = target;
   }, [open, target]);
 
+  const handleExited = React.useCallback(
```
flaviendelangle, line 83: The `useCallback` is not usefull here since we use it as a HOF.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx
m4theushw: The purpose of this effect is: `showColumnMenuIcon` should always reflect the value of `columnMenuOpen` UNLESS `showColumnMenuIcon=true`.

It could be rewrote as `setShowColumnMenuIcon(value => !value ? columnMenuOpen : value);`. By the time this effect run, `columnMenuOpen` will already be false.

> Updating from true to true is not a problem no ?

No, React won't rerender the component if the value didn't change.

flaviendelangle (approved):

PR #3292, opened 2021-11-27
[core] Fix yarn prettier script

oliviertassinari:
I got the following error:

Screenshot 2021-11-27 at 14 55 33

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "deduplicate": "node scripts/deduplicate.js",
     "l10n": "babel-node -x .ts ./scripts/l10n.ts",
     "stylelint": "stylelint '**/*.js' '**/*.ts' '**/*.tsx'",
-    "prettier": "node ./scripts/prettier.js --branch next",
+    "prettier": "node ./scripts/prettier.js --branch master",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "deduplicate": "node scripts/deduplicate.js",
     "l10n": "babel-node -x .ts ./scripts/l10n.ts",
     "stylelint": "stylelint '**/*.js' '**/*.ts' '**/*.tsx'",
-    "prettier": "node ./scripts/prettier.js --branch next",
+    "prettier": "node ./scripts/prettier.js --branch master",
```
oliviertassinari, line 22: master is the default but I'm forcing it in case the main mono-repo changes its base branch.

flaviendelangle: We should add a `Switch main branch` paragraph in the `scripts` README.
We already forgot to switch that one when doing `master`  =>  `next` a few months ago.

flaviendelangle (approved):

package.json
oliviertassinari: Yeah, OK, doing a new PR for this.

PR #3302, opened 2021-11-29
[DataGrid] Avoid page break inside a row in the Print Export

flaviendelangle:
[Example before](https://github.com/mui-org/material-ui-x/files/7619365/master.version.pdf)
Screenshot 2021-11-30 at 12 16 23

[Example after](https://github.com/mui-org/material-ui-x/files/7619368/avoid.break.inside.pdf)
Screenshot 2021-11-30 at 12 16 01

See the first/last rows of each page

https://stackoverflow.com/questions/907680/css-printing-avoiding-cut-in-half-divs-between-pages/907719


packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -186,6 +186,7 @@ export const GridRootStyles = styled('div', {
     [`& .${gridClasses.row}`]: {
       display: 'flex',
       width: 'fit-content',
+      breakInside: 'avoid',
```
oliviertassinari, line 189: Should we leave a comment on why?

```suggestion
      breakInside: 'avoid', // Avoid the row to be broken in two different print pages.
```

flaviendelangle: If the CSS property seems to obscure to be self explanatory then we can add a comment yes :+1:

PR #3303, opened 2021-11-29
[core] Update monorepo

m4theushw:


m4theushw:
@mui-org/x I removed the hardcoded license key to avoid leaking it on GitHub. To remove the watermark during development, add an `NEXT_PUBLIC_MUI_LICENSE` env var with the value from https://github.com/mui-org/material-ui-x/pull/3303/files#diff-41c21e2dc42b8c223fc2fa17580e58041afa590fe34d9fbe9a1ffbe5e17d318bL6.

Issue #2854, opened 2021-10-13
[DataGrid] Cell edit with date(or datetime) types not working properly


zzossig:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When I double click a cell that is date(dateTime) type, key inputs are scramble.

### Expected behavior 🤔

strings should be typed correctly.

### Steps to reproduce 🕹

MUI document also doesn't work correctly. See and try edit a cell your self(with keyboard).
https://mui.com/components/data-grid/editing/#cell-editing

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

### Order ID 💳 (optional)

#27772

m4theushw:
What do you mean by "key inputs are scramble"? Also, which date format you use, e.g. American style (yyyy-mm-dd)?

flaviendelangle:
@zzossig Could you please confirm that my fix is correcting your failing use case ? 
https://deploy-preview-2918--material-ui-x.netlify.app/components/data-grid/editing/#cell-editing

zzossig:
It looks good for me only if edit cell accept just `date`.
But when I try to edit `Last Login` cell, which is accept not only `date` but also `time`,
transition to time input not going smoothly.
That means, the year part(`yyyy`) accept 6 character instead of 4

flaviendelangle:
I agree that it feels weird
But it's the HTML input that is doing it, I don't think we should try to override it
See https://developer.mozilla.org/fr/docs/Web/HTML/Element/Input/datetime-local#valeur

m4theushw:
@flaviendelangle Can you record again the video but also showing the key strokes? Which locale and date format did you use?

https://user-images.githubusercontent.com/42154031/139683549-563fac89-539c-4b16-bbda-a610588c9f01.mp4

flaviendelangle:
Without my PR:

https://user-images.githubusercontent.com/3309670/139706271-456fbf72-b484-4557-b27a-19ec6413e3d2.mp4

https://user-images.githubusercontent.com/3309670/139706296-43b0393d-e12b-4d0e-88d8-882002edd1d7.mp4

flaviendelangle:
With my PR: 

[...]

PR #1652, opened 2021-05-11
[DataGrid] Fix incorrect date selection

aTmb405:
Fixes #1573

PR #3307, opened 2021-11-30
[DataGrid] Fix DateInput bug on parser

### LINKED ISSUE #3305: [DataGrid] Can't pick February in `date` column
### Duplicates

### Latest version

### Current behavior 😯

If you have a "date" type column for the DataGrid, selecting any day in February (for any year), returns the same day in March instead.

Basically the same as this issue: #3061 
I posted a comment there, but I'm not sure if "Closed" issue actually receive any attention so I'm making a new issue.

### Expected behavior 🤔

Selecting a date returns the selected date.

### Steps to reproduce 🕹

Steps:

1. Add a "date" type column to any DataGrid.
2. Try picking a day in February

Check out [this](https://codesandbox.io/s/sweet-pasteur-dtjx6?file=/src/App.js) codesandbox.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
  Tested on Safari & Chrome.

  System:
    OS: macOS 12.0.1
  Binaries:
    Node: 14.17.5 - /usr/local/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 8.1.4 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 96.0.4664.55
    Edge: Not Found
    Firefox: Not Found
    Safari: 15.1
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3305

The bug was the following: If you are the 30th you can not set February, and if you are the 31th you can not set a month with only 30 days.


oliviertassinari (commented):
Do we have enough tests on this date input logic? I'm asking because it seems that this problem is really hard to get right, it's our 5th iteration on it:

- bug #1567, PR #1570
- bug #1573, PR #1652
- bug #1920, PR #2051
- bug #2854, PR #2918
- bug #3305, PR #3307

oliviertassinari:
Could this be a regression? I see an offset of 1 minute between what I select and what I can see when I enter again the edit mode:

offset

flaviendelangle:
> Could this be a regression

:grimacing: yes clearly
Date are such a pain... I'm in favor of adding a test file on `GridEditDateCell` to more easily unit test it.

oliviertassinari:
Regarding the approach itself, to be aware of:

Screenshot 2021-11-30 at 18 21 51

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

and how dateFns approach it: https://github.com/date-fns/date-fns/blob/4544f0cb1d724bfa7ee846b762bacb8fc303a961/src/parseISO/index.js#L75

m4theushw:
We could add an e2e test. Unit testing the date input can't fully reproduce how the browser behaves.

On the long term, I'm in favor of using the date picker component on this field. The native input has many problems, as well as, it's not consistent visually across different browsers.

alexfauquette:
@oliviertassinari I do not manage to reproduce the minute shift

test/e2e/index.test.ts
```diff
@@ -257,5 +257,63 @@ describe('e2e', () => {
...
+      // set 06/25/1986
+      await page.dblclick('[role="cell"][data-field="birthday"]');
+      await page.keyboard.press('0');
+      await page.keyboard.press('6');
+      await page.keyboard.press('2');
+      await page.keyboard.press('5');
+      await page.keyboard.press('1');
+      await page.keyboard.press('9');
+      await page.keyboard.press('8');
+      await page.keyboard.press('6');
```
oliviertassinari, line 280: Maybe it would be easier to read with https://playwright.dev/docs/api/class-page#page-type

test/e2e/fixtures/DataGrid/KeyboardEditDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+      lastConnexion: new Date(2022, 1, 20, 6, 50, 0),
+    },
+  ],
+  columns: [
+    { field: 'birthday', type: 'date', editable: true, width: 100 },
+    { field: 'lastConnexion', type: 'dateTime', editable: true, width: 100 },
+  ],
+};
+
+export default function KeyboardNavigationFocus() {
```
m4theushw, line 18:
```suggestion
export default function KeyboardEditDate() {
```

test/e2e/fixtures/DataGrid/KeyboardEditDate.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import { DataGrid } from '@mui/x-data-grid';
+
+const baselineProps = {
+  rows: [
+    {
+      id: 0,
+      birthday: new Date(1984, 1, 29),
+      lastConnexion: new Date(2022, 1, 20, 6, 50, 0),
```
m4theushw, line 9:
```suggestion
      lastConnection: new Date(2022, 1, 20, 6, 50, 0),
```

m4theushw (commented):
Is there a recording of #3305? When I type "30/02/2020" on HEAD it empties the field. I don't know if this is the bug that we're fixing here.

alexfauquette:
@m4theushw I don't have a recording.

The 30/02/2020 empty the field because it does not exist. Probably we should not commit the new cell value in such a case, but that's another feature.

The bug was when you open the date picker calendar, and you pick a valid date in February, the date saved was in march, because the date was initialized to the current date, and the month was set before the day, leading to an unexisting date if you are the 30th of the month.

About tests, I added, they are not related to this bug. It would imply to set the environment at a date after the 30th?

m4theushw (approved):
I can't reproduce the initial bug but I also didn't notice any regression with the change proposed.

> About tests, I added, they are not related to this bug. It would imply to set the environment at a date after the 30th?

I didn't understand this.

alexfauquette:
@m4theushw The bug was because we used some wear `new Date()` which initialize the date to now. And this was a problem only if the current date is the 30th or the 31th of the month.

So to make a test that prevents the issue to occurs again, we should fake the browser clock such that `new Date()` always returns the 30th of January for example.

PR #3310, opened 2021-12-01
[DataGrid] Unsubscribe event listeners registered in uncommitted renders

m4theushw:
This PR is a follow-up on https://github.com/mui-org/material-ui-x/pull/3217#issuecomment-979682863.

To be able to get event listeners being fired as soon as possible, we started to register them synchronously (outside an effect). This ensured that we wouldn't depend on the order that React calls the effects. In a perfect world, these listeners would be cleaned latter once the cleanup of the effect ran. However, with StrictMode / Concurrent Mode, a component might be called multiples times and their effects might not even run. When this occurs, the listener is never unsubscribed, and if this listener is called later it may update the state of an unmounted component. To solve that, this PR implements a pattern, also used by [other](https://github.com/facebook/react/issues/15317#issuecomment-721383797) libs, of trying to guess when the component is called without running its effects. When this occurs, it unsubscribes the listener also outside the effect. Two techniques are employed: 

- [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)-based implementation - This is a new API supported on all major browsers (not IE, obviously), which allows to run a callback once a value/reference is not retained in memory anymore.
- Timer-based implementation - For each call to `useGridApiEventHandler`, we start a timer (1000ms for now). If a effect runs within this time, it means that the cleanup function will also be called and we stop the timer. If no effect ran, we unsubscribe the listener when the timer times-out. This is the fallback if the above API is not available.

When running the tests, the timer-based implementation is always used. I made this decision because, if the fake clock is also enabled, is easier to avoid one test leaking into another. This is also one trade-off of this solution: all tests should, in theory, use the fake clock. I say "in theory" because leaving some running with the real clock didn't cause any warning, but it might. @mui-org/x be aware that if an unrelated test starts to fail weirdly it can be caused by not using a fake clock.

Wants a practical use-case?

- Open https://codesandbox.io/s/fullfeatureddemo-material-demo-forked-lrngb?file=/index.tsx
- Add a filter in the "Commodity" column with the value "foobar" (something that won't be found)
- Look at the "Can't perform a React state update on an unmounted component." error in the console.

In the same demo using this PR (https://codesandbox.io/s/fullfeatureddemo-material-demo-forked-xkh6q) the error is less frequent. It may still occur because we don't know when FinalizationRegistry will run but we know that it will be called at some point in the future.


packages/grid/_modules_/grid/hooks/core/useGridApiInitialization.ts
```diff
@@ -60,7 +60,6 @@ export function useGridApiInitialization(
     return () => {
       logger.info('Unmounting Grid component. Clearing all events listeners.');
       api.publishEvent(GridEvents.unmount);
-      api.unstable_eventManager.removeAllListeners();
```
m4theushw, line 63: Each component/hook should be responsible for unsubscribing only its own listeners. Removing all listeners at once could hide something.

packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts
```diff
@@ -12,30 +15,26 @@ export enum GridSignature {
...
-    subscription.current();
-    subscription.current = null;
-  }
+  return function useGridApiEventHandler<E extends GridEventsStr>(
+    apiRef: GridApiRef,
+    eventName: E,
+    handler?: GridEventListener<E>,
+    options?: EventListenerOptions,
+  ) {
+    const [objectRetainedByReact] = React.useState(new ObjectToBeRetainedByReact());
```
flaviendelangle, line 32:
```suggestion
    const [objectRetainedByReact] = React.useState(() => new ObjectToBeRetainedByReact());
```

May be totally negligible though.

packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts
```diff
@@ -44,15 +43,69 @@ export function useGridApiEventHandler<E extends GridEventsStr>(
       };
 
       subscription.current = apiRef.current.subscribeEvent(eventName, enhancedHandler, options);
-    }
 
-    return () => {
-      subscription.current?.();
+      cleanupTokensCounter += 1;
+      cleanupTokenRef.current = { cleanupToken: cleanupTokensCounter };
```
flaviendelangle, line 48: Compared to https://github.com/Bnaya/use-dispose-uncommitted/blob/main/src/finalization-registry-based-impl.ts, why don't you have the `committed` key ?

flaviendelangle:
Not a lot to say here, I checked the React Github Issue and the hook you are using as a base, your implementation seems fine.
It could be interesting to be able to force the time mode for debugging purposes via an env variable for instance.

DanailH:
Not sure if applicable but is this related to this one https://github.com/mui-org/material-ui-x/issues/3276?

flaviendelangle:
You did not answer my review comments

packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts
m4theushw: The hook that I based on has the option to pass an `onUncommittedCallback` callback. The `committed` key seems to be used to update the reference to this callback when the component rerenders before running the effects. It's a bit similar to https://github.com/mui-org/material-ui-x/blob/0b23cdd82d671fa7be52f9622de8cfe0ef1cd747/packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts#L59

m4theushw:
> It could be interesting to be able to force the time mode for debugging purposes via an env variable for instance.

@flaviendelangle For debugging, there's always the option of commenting the line that checks if `FinalizationRegistry` is available: https://github.com/mui-org/material-ui-x/blob/0b23cdd82d671fa7be52f9622de8cfe0ef1cd747/packages/grid/_modules_/grid/hooks/utils/useGridApiEventHandler.ts#L102

> You did not answer my review comments

Sorry, this PR has been sitting for too long.

PR #3313, opened 2021-12-01
[core] Generate CHANGELOG from GitHub API

alexfauquette:
The goal is to get a script that will not break when GitHub updates its font-end


m4theushw:
Did you consider that the results might be paginated? In #3312 it has 2 pages in the GitHub UI.

In the core repo, there's this script which you could take some inspiration: https://github.com/mui-org/material-ui/blob/master/scripts/releaseChangelog.js

It uses https://github.com/octokit/octokit.js instead of manually fetching the REST API.

alexfauquette:
The script is now able to look for PR description containing `# changelog`. All the text after this key will be added to the changelog generated. On the same idea, we could add break changes when we will prepare v6.

alexfauquette:
@m4theushw if you have a github token, the l10n can automatically update the issue when running `yarn l10n --report`. The token can be provided by adding `GITHUB_TOKEN`in your env variables or adding the option `--github_token` 🎁

m4theushw (commented):
Finally, it's the end for the inconsistencies in the CHANGELOG between each release. 😁

scripts/l10n.ts
```diff
@@ -212,12 +213,34 @@ async function generateReport(
...
+async function updateIssue(githubToken, newMessage) {
+  // Initialize the API client
+  const octokit = new Octokit({
+    auth: githubToken,
+  });
+
+  await octokit.request('PATCH /repos/{owner}/{repo}/issues/{issue_number}', {
+    owner: 'mui-org',
+    repo: 'material-ui-x',
+    issue_number: 3211,
```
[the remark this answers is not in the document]
alexfauquette: What about defining them as constants at the beginning of the file

```
GIT_ORGANIZATION = 'mui-org';
GIT_REPO = 'material-ui-x';
L10N_ISSUE_ID = 3211;
```

My concern is that someone debugging this code in 2 years will have every information set in a single file.

scripts/l10n.ts
```diff
@@ -212,12 +213,34 @@ async function generateReport(
   return lines.join('\n');
 }
 
+async function updateIssue(githubToken, newMessage) {
```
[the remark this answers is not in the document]
alexfauquette: Ok, I will do the split latter

alexfauquette:
@m4theushw The dedicated PR for translation is #3588

scripts/releaseChangelog.js
```diff
@@ -0,0 +1,243 @@
...
+
+    const sortedCommits = commitsList.sort((a, b) => {
+      const aTags = parseTags(a.commit.message);
+      const bTags = parseTags(b.commit.message);
+      if (aTags === bTags) {
+        return a.commit.message < b.commit.message ? -1 : 1;
+      }
+      return aTags.localeCompare(bTags);
+    });
+    return `${header}
```
m4theushw, line 173: It should not add the header if it's an empty string. It adds unnecessary line breaks before inserting the other commits.

image

scripts/releaseChangelog.js
```diff
@@ -0,0 +1,243 @@
...
+
+A big thanks to the ${
+    authors.length
+  } contributors who made this release possible. Here are some highlights ✨:
+
+### \`@mui/x-data-grid@v\` / \`@mui/x-data-grid-pro@v\`
+
+TODO INSERT HIGHLIGHTS
+
+${changeLogMessages.join('\n')}
```
m4theushw, line 201: The highlights come before the package names.

m4theushw: This was not addressed.

image

scripts/releaseChangelog.js
```diff
@@ -0,0 +1,242 @@
...
+        return;
+      }
+
+      const changelogMotif = '# changelog';
+      const changelogIndex = bodyMessage.toLowerCase().indexOf(changelogMotif);
+      if (changelogIndex >= 0) {
+        changeLogMessages.push(
+          `From https://github.com/${GIT_ORGANIZATION}/${GIT_REPO}/pull/${
+            searchPullRequestId[1]
+          }\n${bodyMessage.slice(changelogIndex + changelogMotif.length)}`,
```
m4theushw, line 117: If the PR description doesn't have "## changelog" as the last section it will include everything as the example below:

image

I tried to use the regex `/## changelog(.*?)(?:(##|$))/is` to capture only the part after the title and before the next title. I'm not sure if we should care about this but here it's.

scripts/releaseChangelog.js
alexfauquette: My bad, I moved the "TODO INSERT HIGHLIGHTS" and forget the `changeLogMessages`

alexfauquette:
I merge it because I'm doing the release this week. That will be a perfect occasion for a first test ;)

PR #3314, opened 2021-12-01
[DataGrid] Deprecate the `getValue` param for the `valueGetter` callback

### LINKED ISSUE #1564: `InternalError too much recursion` only when `GridColDef.valueGetter` is included

## Current Behavior 😯

This is a component that I am using in my Next.js app:

```tsx
import React from 'react'
import {
  Button,
  Grid,
  makeStyles,
  TextField,
  Theme,
  Typography
} from '@material-ui/core'
import {
  DataGrid,
  GridColDef,
  GridValueGetterParams
} from '@material-ui/data-grid'
import { format } from 'date-fns-tz'
import { Client } from './models'

export interface FollowupListProps {
  client: Client
}

const useStyles = makeStyles((theme: Theme) => ({
  head: {
    backgroundColor: `${theme.palette.primary.main}`,
    color: '#ffffff'
  },
  table: {
    minWidth: 650
  },
  modal: {
    display: 'flex',
    alignItems: 'center',
    justifyContent: 'center',
    minWidth: 800
  }
}))

export default function FollowupList({
  client
}: FollowupListProps): React.ReactElement<FollowupListProps> {
  const classes = useStyles()
  const { followupHistory: followups } = client
  const [searchText, setSearchText] = React.useState('')

  const formatter = (key: string, value: any): string => {
    const option = { timeZone: 'Australia/Sydney' }
    if (key === 'createdOn')
      return format(value as Date, 'dd/MM/yyyy hh:mm:ss', option)
    if (key === 'followupDate')
      return format(value as Date, 'dd/MM/yyyy', option)
    return value
  }

    { field: 'id', headerName: 'S.No.', width: 20 },
    {
      field: 'createdOn',
      headerName: 'Created On',
      width: 200,
      valueGetter: ({ getValue }: GridValueGetterParams) =>
        `${formatter('createdOn', getValue('createdOn'))}`
    },
    { field: 'createdBy', headerName: 'Created By', width: 120 },
    { field: 'createdFrom', headerName: 'Created From', width: 80 },
    { field: 'note', headerName: 'Note', width: 500 },
    {
      field: 'followupDate',
      headerName: 'Followup Date',
      width: 200,
      valueGetter: ({ getValue }: GridValueGetterParams) =>
        `${formatter('followupDate', getValue('followupDate'))}`
    }
  ]

  return (
    <div style={{ marginTop: '10px' }}>
      <Grid
        container
        direction="row"
        alignItems="center"
        className={classes.head}
      >
        <Grid item>
          <Typography variant="h6">Followup List</Typography>
        </Grid>
      </Grid>
      <fieldset>
        <legend>
          <Typography variant="h6" color="textSecondary">
            Followup History
          </Typography>
        </legend>
        <div
          style={{
            display: 'flex',
            justifyContent: 'center',
            alignItems: 'center'
          }}
        >
          <Button variant="contained" color="primary">
            Add Followup
          </Button>
        </div>
        {followups.length ? (
          <div>
            <div
              style={{
                display: 'flex',
                flexDirection: 'row',
                justifyContent: 'right',
                marginBottom: '10px'
              }}
            >
              <TextField
                variant="outlined"
                label="Search"
                aria-label="Search followups"
                value={searchText}
                onChange={(e) => setSearchText(e.target.value)}
              />
            </div>
            <div style={{ height: 1000, width: '100%' }}>
              <DataGrid
                rows={followups}
                columns={columns}
                pageSize={15}
                checkboxSelection
              />
            </div>
          </div>
        ) : (
          <div
            style={{
              display: 'flex',
              justifyContent: 'center',
              alignItems: 'center'
            }}
          >
            <Typography>No followup</Typography>
          </div>
        )}
      </fieldset>
    </div>
  )
}
```

When I use the `valueGetter` property inside the items of `columns`, it shows the following error:

- on Edge 90:

  ```
  RangeError
  Maximum call stack size exceeded
  ```

- on Firefox 88:
  ```
  InternalError
  too much recursion
  ```

The grid renders all right and no error is shown when I remove all the `valueGetter` properties.

## Expected Behavior 🤔

The grid should render all right, with the dates properly formatted.

## Steps to Reproduce 🕹

Link to the CodeSandbox: https://codesandbox.io/s/festive-grothendieck-fk0lv?file=/src/Demo.tsx

I am using the component in a Next.js app, but the error also appears when the component is used in a CRA template.

Steps:

1. Add the packages `@material-ui/data-grid`, `date-fns` and `date-fns-tz`
2. Add the models `Client` and `Followup`
3. Add some dummy data of `Client` and `Followup`
4. Pass the data on to the component defined above

## Context 🔦

I am developing a front-end project using Next.js and Material-UI.

## Your Environment 🌎

<details>

  I am using Firefox 88.0 as my primary browser. The error also appears on Edge browser, as mentioned before.

```
System:
  OS: Windows 10 10.0.19041
Binaries:
  Node: 16.0.0 - C:\Program Files\nodejs\node.EXE
  npm: 7.11.1 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Chrome: Not Found
  Edge: Spartan (44.19041.906.0), Chromium (90.0.818.51)
npmPackages:
  @emotion/react: ^11.1.5 => 11.1.5
  @emotion/styled:  10.0.27
  @material-ui/core: ^4.11.3 => 4.11.3
  @material-ui/data-grid: ^4.0.0-alpha.26 => 4.0.0-alpha.26
  @material-ui/styles:  4.11.3
  @material-ui/system:  4.11.3
  @material-ui/types:  5.1.0
  @material-ui/utils:  4.11.2
  @types/react: ^17.0.1 => 17.0.3
  react: ^17.0.1 => 17.0.2
  react-dom: ^17.0.1 => 17.0.2
  typescript: ^4.1.3 => 4.2.4
```

</details>

flaviendelangle:
Fixes #1564 
Extracted from #3277

For #3277 I need to call the `valueGetter` before the column state is initialized to use its value as a grouping key.
But at that point I can't use `getBaseCellParams` in `useGridParamsApi` because many api methods / states are not ready yet.
I added a `GridValueGetterSimpleParams` interface and make `valueGetter` callable with either the full params or this new interface.

It is not a breaking change since the new incomplete interface is only called for new code in #3277.
But I migrated all our examples to be compatible with #3277 and the main friction point is the `getValue` param which is an `apiRef` method that I can't use before `useGridParamsApi` initialization. 

We are using this property in 2 callbacks:
- `getRowClassName`:  I can understand the interest of having a wrapper that given a field, handles the potential `valueGetter`.

- `valueGetter`: I don't really see the point of using this property instead of a direct row access. The main difference is that if a col A uses `getValue('B')` in its `valueGetter` and if the col B has itself a `valueGetter`, it will call it. But I think this `valueGetter` chain adds a lot of not very usefull complexity and that we should deprecate the `getValue` for the `valueGetter` callback.


m4theushw (commented):
Wouldn't be simpler to not create new interfaces and add a warning when `getValue` is called?

> valueGetter: I don't really see the point of using this property instead of a direct row access. 

It's useful for when the row structure is nested.

flaviendelangle:
> Wouldn't be simpler to not create new interfaces and add a warning when getValue is called?

I'm only removing `getValue` for `valueGetter` so I'm not sure how we could do this, could you elaborate please ?

> It's useful for when the row structure is nested.

Would you have an example for `valueGetter` where it would as useful as possible ?

m4theushw:
> I'm only removing getValue for valueGetter so I'm not sure how we could do this, could you elaborate please ?

I was saying to not create new interfaces and only add a warning if `getValue` is called.

> Would you have an example for valueGetter where it would as useful as possible ?

Are you talking about `valueGetter` or `getValue`? I was referring to the first in https://github.com/mui-org/material-ui-x/pull/3314#pullrequestreview-821563335 to say that it's useful when the data is nested. However, `getValue` is redundant and can be removed in the major version.

flaviendelangle:
> Are you talking about valueGetter or getValue?

I am talking about the `getValue` passed as a param to `valueGetter`.
When it is passed as a param for `getRowClassName` I don't see any problem.
But for `valueGetter`, it creates chained valueGetter calls and I don't think it's a good idea.

packages/grid/_modules_/grid/models/params/gridCellParams.ts
```diff
@@ -82,9 +82,54 @@ export interface GridRenderEditCellParams extends GridEditCellProps {
...
+  /**
+   * GridApi that let you manipulate the grid.
+   */
+  api: GridApi;
+}
+
+/**
+ * Parameters passed when calling `colDef.valueGetter` in the row grouping sequence.
+ */
+export interface GridValueGetterSimpleParams<R = any> {
```
m4theushw, line 106: I don't think it's worth to create another interface to remove `getValue`. In v6 we'll need to remove it. Instead, we could only flag it as deprecated and be cautious to never call it + display warning if someone calls it.

flaviendelangle: This interface has other differences than `getValue`
Once again it's for the grouping process, so it does not have `api` either for instance.

m4theushw: > If you want to fully deprecate getValue, I can move it's deprecation from GridValueGetterFullParams to GridCellParams. That was not the goal of this PR but I'm not agaist the idea.

Yes, let's deprecate `getValue` entirely. The sooner we do, less work we have for v6.

> But for GridValueGetterSimpleParams I think we do need another interface.

For now, it's necessary because the `api` is a required property. However, in v6 this interface can be removed.

m4theushw (approved):
I'm approving to make #3277 to move faster.

Two things which could be done latter:

- Add a deprecation warning when `getValue` is called
- Leave a "TODO v6" note so we can find easily it to remove
- Add to https://github.com/mui-org/material-ui-x/issues/3287

PR #3315, opened 2021-12-01
[core] Type the api param in callback interfaces

flaviendelangle:
Extracted from #3277


packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
```diff
@@ -19,7 +19,7 @@ const useUtilityClasses = (ownerState: OwnerState) => {
   return composeClasses(slots, getDataGridUtilityClass, classes);
 };
 
-export function GridEditDateCell(props: GridRenderEditCellParams & InputBaseProps) {
+export function GridEditDateCell(props: GridRenderEditCellParams & Omit<InputBaseProps, 'id'>) {
```
alexfauquette, line 22: I do not understand why removing the `id` props from the signature. Is it because you want to force `id` to get its type from `GridRenderEditCellParams`?

packages/grid/x-data-grid-generator/src/renderer/renderEditStatus.tsx
```diff
@@ -38,14 +38,18 @@ function EditStatus(props: GridRenderEditCellParams) {
...
     api.setCellMode(id, field, 'view');
 
     if ((event as any).key) {
       // TODO v6: remove once we stop ignoring events fired from portals
       const params = api.getCellParams(id, field);
-      api.publishEvent(GridEvents.cellNavigationKeyDown, params, event);
+      api.publishEvent(
+        GridEvents.cellNavigationKeyDown,
+        params,
+        event as any as React.KeyboardEvent<HTMLElement>,
```
alexfauquette, line 51: What would you think about using a [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to avoid the `as any as`?

```
function isKeyboardEvent(event: React.KeyboardEvent<HTMLElement> | any): event is React.KeyboardEvent<HTMLElement>  {
  return (event as React.KeyboardEvent<HTMLElement> ).key !== undefined;
}
```

I do not know if it brings something or if it just for the esthetic

flaviendelangle: I need the `as any` because the `SelectProps['onChange']` `event` params is a `changeEvent` so it technically can't be a `keyboardEvent`

We talked about it with @m4theushw a while back when I asked him if there was some scenario where is was actually a `keyboardEvent` and for him the typing from the core is imprecise for this param.

Honestly I did not dif deeper.
But as `changeEvent` do not extend `keyboardEvent` in any way, I will require the `as any` to trick TypeScript event with type predicates.

Unless I'm missing something, I you find a way to avoid that ugly  `as any` I'm very interested :)

packages/grid/_modules_/grid/components/cell/GridEditDateCell.tsx
flaviendelangle: Both `GridRenderEditCellParams` and `InputBaseProps` have an `id` prop.

`GridRenderEditCellParams['id'] === GridRowId`
`InputBaseProps['id`] === string | undefined`

So now that we are typing the `api` param, TypeScript is complaining that we are trying to pass `undefined` to the `id` param on `api.setEditCellValue({ id, field, value: newParsedDate }, event);`.

And the way the component is built, the `id` prop will never be passed as `id` to the HTML element (nor should it be).
So I forced the typing to set the `id` prop as the one from `GridRenderEditCellParams` to correctly reflect the current behavior.

packages/grid/x-data-grid-generator/src/renderer/renderEditStatus.tsx
alexfauquette: Don't think I can help with such a tricky typescript issue 😅

alexfauquette (approved):

PR #3316, opened 2021-12-01
[DataGridPro] Pass a list of `fields` to the callback version of `groupingColDef`

flaviendelangle:
Extracted from #3277

The callback version of `prop.groupingColDef` has never been usable in a released version so this is not a breaking change.
This code will be particularly useful in #3277.


packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
```diff
@@ -36,18 +38,18 @@ export const useGridTreeData = (
...
-    //
-    //   colDefOverride = propGroupingColDef(params);
-    // } else {
-    //   colDefOverride = propGroupingColDef ?? {};
-    // }
+    let colDefOverride: Partial<GridColDef>;
+
+    if (typeof propGroupingColDef === 'function') {
+      const params: GridGroupingColDefOverrideParams = {
+        treeGroupingName: TREE_DATA_GROUPING_NAME,
```
alexfauquette, line 45:
```suggestion
        groupingName: TREE_DATA_GROUPING_NAME,
```

It could help to distinguish "grouping" features from "tree data"

packages/grid/_modules_/grid/models/colDef/gridColDef.ts
```diff
@@ -226,18 +226,20 @@ export interface GridColumnsMeta {
...
+   */
+  treeGroupingName: string;
 
-export interface GridColDefOverrideParams {
   /**
-   * The column we are generating before the override.
+   * The base columns this column is grouping.
    */
-  colDef: GridColDef;
+  sources: GridColDef[];
```
m4theushw, line 244: Should it be named "fields" and contain only the fields? Is there an use case where the full column definition needs to be present? In https://deploy-preview-3277--material-ui-x.netlify.app/components/data-grid/group-pivot/#custom-grouping-columns it only uses the field name. 

Passing the column reference opens a way for users to directly update the  column, which may cause unwanted behaviors.

flaviendelangle: I can see edge cases where you would want to base your grouping column on other things than the field.
But we can probably start with the fields, I agree that it would be safer and would avoid some dirty mutations on the user side.

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
flaviendelangle: Done :+1:

m4theushw (approved):
Before merging, please update the title of the PR to reflect the change between passing the `colDef` to fields.

PR #3318, opened 2021-12-02
[core] Use pre-processors for sorting and filtering

flaviendelangle:
Extracted from #3277
Part of #2994 

## Behavior

The sorting and filtering pre-processor do not work exactly as the column one.

### Behavior of the colums pre-processor

1. `useFeatureXXX` calls `registerPreProcessor` because one of the prop used in some `updateFeatureXXXColumn` has changed due to a change in `prop.featureXXXMode` (for instance).
2. `useGridColumns` receives a `preProcessorRegister` event.
3. `useGridColumns` calls `applyPreProcessors` with the unprocessed columns.
4. `useGridPreProcessing` calls all the colums pre-processor in the registration order, each pre-processor add / remove or update some columns.
5. `useGridPreProcessing` returns the pre-processed columns

Imagine that `useFeatureXXX` is called in the `DataGridPro` but only wants to add a column if `prop.featureXXX` is enabled.
Since `useFeatureXXX` is called even when `prop.featureXXX = false`, its method `updateFeatureXXXColumn` will still change its reference when `prop.featureXXXMode` changes.
So when `prop.featureXXXMode` changes, we will re-apply all the column pre-processors "uselessly". 

For the columns, it is not a major issue since the prep-processing are very light. With the new format passing the `GridColumnsRawState`, the complexity to check if a column needs to be added / removed is usually constant thanks to the lookup.

But for the sorting / filtering, we really don't want to re-apply the sorting / filtering when prop listened by a disabled feature are updated.

### Behavior of the sorting / filtering pre-processor

The logic below describes the sorting pre-processor but things are exactly the same for the filtering one.

1. `useFeatureXXX` calls `registerPreProcessor` because one of the prop used in some `updateFeatureXXXColumn` has changed due to a change in `prop.featureXXXMode` (for instance).
2. `useGridSorting` receives a `preProcessorRegister` event.
3. `useGridSorting` calls `applyPreProcessors` with an empty object.
4. `useGridPreProcessing` calls all the sorting pre-processor in the registration order, each pre-processor add a property to the empty object with its `treeGroupingName` as a key and its sorting algorithm as a value.
5. `useGridPreProcessing` returns the collection of sorting algorithms.
6. `useGridSorting` stores the collection in a ref.
6. `useGridSorting` checks if the sorting algorithm matching the current `treeGroupingName` has changed since the last sorting application, if so it re-applies the sorting, if not it don't re-apply the sorting.

This has two main behavior difference:

- If I register a new `treeData` sorting method but I am in `treeGroupingName = "groupingColumns"`, I won't re-apply the sorting
- I only call the sorting method matching the algorithm used to build the tree currently in state

## Design choices

I chose to re-use `apiRef.current.unstable_applyPreProcessors` and `apiRef.current.unstable_registerPreProcessor` to avoid increasing the API surface. But to be compatible with it I had to add a ref in the sorting and filtering hook to store the algorithm collections. 
A cleaner implementation would probably be possible by creating two types of pre-processors: 

**Chained pre-processors**:
A hook listens to `preProcessorChainRegister` and `preProcessorChainUnRegister`
These events each time a pre-processor is added / removed / changed.
Whenever one of these events is fired, the hook calls `applyPreProcessorChain` with an input, the method calls all the pre-processors and return an ouput of the same shape.

**Branch  pre-processor**:
A hook listens to `preProcessorBranchChanged`
These events each time a pre-processor is added / removed / changed.
Whenever the branch (for sorting / filtering the branch would be `state.rows.treeGroupingName` changes) AND whenever this even is fired, the hook calls `getPreProcessorBranch` with the current branch, the method returns the algorithm matching this branch.
The hook then compare this method with the method applied the last time he ran the logic (the last time `useGridSorting` ran `applySorting` for instance), if the method has changed, it re-apply the logic, if not it does nothing.

*BUT* I prefer to wait before adding a new abstraction. Pre-processor are a powerfull tools with new use-cases coming rapidly. I prefer to take a step back and wait for more use-cases to build an abstraction better fit for all of them.
For now, handling the algorithm collection in `useGridSorting` and `useGridFiltering` causes code duplication but is not big of an issue.


packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.ts
```diff
@@ -121,24 +89,114 @@ export const useGridTreeData = (
...
+      ...colDefOverride,
+    };
+  }, [apiRef, props.groupingColDef]);
+
+  const updateGroupingColumn = React.useCallback(
+    (columnsState: GridColumnsRawState) => {
+      const groupingColDefField = GRID_TREE_DATA_GROUP_COL_DEF_FORCED_PROPERTIES.field;
+
+      const shouldHaveGroupingColumn = props.treeData;
+      const haveGroupingColumn = columnsState.lookup[groupingColDefField] != null;
```
flaviendelangle, line 127: As written in my PR description, for the column pre-processing, the check to know if we need to do something has a complexity of `O(1)`.
The only linear complexity here is if we want to remove the grouping column (to find its index) but iit won't be need if the prep-processor is called after a change in a prop of another pre-processor.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -404,6 +335,37 @@ export const useGridFilter = (
     'FilterApi',
   );
 
+  /**
```
flaviendelangle, line 338: I started to structure the big hooks with section comments to make it easier to find something.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -416,23 +378,28 @@ export const useGridFilter = (
     }
   }, [apiRef, logger]);
 
-  React.useEffect(() => {
-    if (props.filterModel !== undefined) {
-      apiRef.current.setFilterModel(props.filterModel);
-    }
-  }, [apiRef, logger, props.filterModel]);
+  const handlePreProcessorRegister = React.useCallback<
```
flaviendelangle, line 381: See description of the PR for more details about this logic

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -441,4 +408,22 @@ export const useGridFilter = (
     apiRef.current.unstable_applyFilters,
   );
   useGridApiEventHandler(apiRef, GridEvents.columnsChange, handleColumnsChange);
+  useGridApiEventHandler(apiRef, GridEvents.preProcessorRegister, handlePreProcessorRegister);
+
+  /**
+   * EFFECTS
+   */
+  useFirstRender(() => {
+    filteringMethodCollectionRef.current = apiRef.current.unstable_applyPreProcessors(
```
flaviendelangle, line 417: This code is after the `useGridRegisterFilteringMethod(apiRef, 'none', flatFilteringMethod);` so in flat mode there is always a filtering algorithm defined when applying the filters for the 1st time.

packages/grid/_modules_/grid/hooks/features/filter/useGridRegisterFilteringMethod.ts
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { GridFilteringMethod, GridFilteringMethodCollection } from './gridFilterState';
+import { GridPreProcessingGroup, useGridRegisterPreProcessor } from '../../core/preProcessing';
+import { GridApiRef } from '../../../models';
+
+export const useGridRegisterFilteringMethod = (
```
flaviendelangle, line 6: Small abstraction of the collection update behavior. 
It does not bring a lot of value but without it, the code is `useGridTreeData` / `useGridGroupingColumns` was less readable.

packages/grid/_modules_/grid/hooks/features/treeData/gridTreeDataUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+  disableChildrenFiltering: boolean;
+  isRowMatchingFilters: ((rowId: GridRowId) => boolean) | null;
+}
+
+/**
+ * A node is visible if one of the following criteria is met:
+ * - One of its children is passing the filter
+ * - It is passing the filter
+ */
+export const filterRowTreeFromTreeData = (
```
flaviendelangle, line 15: Copy / paste from `useGridFilter`
The filtering behavior between `useGridTreeData` and `useGridGroupingColumns` is to different to have the same method.

packages/grid/_modules_/grid/utils/tree/sortRowTree.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface SortRowTreeParams {
+  rowIds: GridRowId[];
+  rowTree: GridRowTreeConfig;
+  disableChildrenSorting: boolean;
+  comparatorList: GridFieldComparatorList;
+  sortRowList: (rowList: GridRowTreeNodeConfig[]) => GridRowId[];
+}
+
+export const sortRowTree = (params: SortRowTreeParams) => {
```
flaviendelangle, line 16: Copy / paste from `useGridSorting`
The behavior is the same in `useGridGroupingColumns` so I set this method in an util folder to use it for both features.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -154,110 +161,34 @@ export const useGridFilter = (
...
-          }
-        }
-      } else if (props.filterMode === GridFeatureModeConstant.client && filteringMethod) {
-        for (let i = 0; i < rowIds.length; i += 1) {
-          const rowId = rowIds[i];
-          visibleRowsLookup[rowId] = filteringMethod(rowId);
-        }
-      }
+      lastFilteringMethodApplied.current = filteringMethod;
+      const filteringResult = filteringMethodCollectionRef.current[rowGroupingName]({
```
m4theushw, line 177:
```suggestion
      const filteringResult = filteringMethod({
```

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -441,4 +408,22 @@ export const useGridFilter = (
     apiRef.current.unstable_applyFilters,
   );
   useGridApiEventHandler(apiRef, GridEvents.columnsChange, handleColumnsChange);
+  useGridApiEventHandler(apiRef, GridEvents.preProcessorRegister, handlePreProcessorRegister);
+
+  /**
+   * EFFECTS
+   */
+  useFirstRender(() => {
```
m4theushw, line 416: `useFirstRender` is not an effect.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
m4theushw: This should be a comment in the code so we know to never move it out of this position.

packages/grid/_modules_/grid/hooks/core/preProcessing/gridPreProcessingApi.ts
```diff
@@ -5,6 +5,8 @@ export enum GridPreProcessingGroup {
   scrollToIndexes = 'scrollToIndexes',
   columnMenu = 'columnMenu',
   canBeReordered = 'canBeReordered',
+  registerFiltering = 'registerFiltering',
```
m4theushw, line 8: Why not? "register" confuses with the name of hook `useGridRegisterPreProcessor`.

```suggestion
  filterMethod = 'filterMethod',
```

similar for the sorting.

m4theushw (commented):
> A cleaner implementation would probably be possible by creating two types of pre-processors:

The pre-processor implements the pipe-filter pattern. What we need is another kind of abstraction implementing the strategy pattern.

packages/grid/_modules_/grid/hooks/features/sorting/useGridSorting.ts
```diff
@@ -187,9 +192,12 @@ export const useGridSorting = (
     }
 
     setGridState((state) => {
-      const rowIds = gridRowIdsSelector(state);
-      const rowTree = gridRowTreeSelector(state);
-      const shouldApplyTreeSorting = gridRowTreeDepthSelector(state) > 1;
+      const rowGroupingName = gridRowGroupingNameSelector(state);
+      const sortingMethod = sortingMethodCollectionRef.current[rowGroupingName];
+      if (!sortingMethod) {
+        throw new Error('MUI: Invalid sorting method');
```
m4theushw, line 198:
```suggestion
        throw new Error('MUI: Invalid sorting method.');
```

m4theushw (approved):

PR #3320, opened 2021-12-02
[core] Remove Typedoc

flaviendelangle:
## Goal

We used very little of the TypeDoc actual API and it was adding a major performance overhead.

I also took the opportunity to build both `x-data-grid` and `x-data-grid-pro` projects it will be very easy in the future to differentiate the two in the API (say "Pro only" in front of something only available on the pro for instance).

## Benchmark

I ran `yarn docs:api:build` twice in a row both on `master` and `remove-typedoc` (to limit somehow the impact of the cache).

|         | master   | remove-typedoc |
|---------|----------|----------------|
| Run n°1 | 17.71sec | 9.30sec        |
| Run n°2 | 15.56sec | 6.33sec        |

Knowing that in all run, we also spend a few seconds in the `tts` process, the perf gain on the rest is major.

## Output changes

The `typeToString` has some behavior difference from our custom implementation

- `(arg?: number) => void` is stringified into `type A = (arg?: number | undefined) => void` instead of `(arg?: number) => void`

- `;` is added after arguments in record types

- The argument order and type union order are more intelligent than ours (`GridRowTreeNodeConfig | null` instead of `null | GridRowTreeNodeConfig`)

- The order in `exportsSnapshots.json` has changed 

- We don't export both the type AND the value for react components in `exportsSnapshots.json`

Before:

```json
  {
    "name": "GridActionsCell",
    "kind": "Namespace"
  },
  {
    "name": "GridActionsCell",
    "kind": "Variable"
  },
```

After:

```
  {
    "name": "GridActionsCell",
    "kind": "ExportSpecifier"
  },
```

- The `kind` is not the same on `exportsSnapshot.json`


docs/package.json
```diff
@@ -55,7 +55,7 @@
     "lodash": "^4.17.21",
     "lz-string": "^1.4.4",
     "markdown-to-jsx": "^7.1.5",
-    "marked": "^4.0.6",
+    "marked": "^3.0.8",
```
flaviendelangle, line 58: The monorepo was using the version installed by Typedoc
Now that we removed it, we must keep our version sync with the one in the core instead of the 4.X (which was never used I think).

docs/pages/api-docs/data-grid/grid-api.md
```diff
@@ -10,92 +10,92 @@ import { GridApi } from '@mui/x-data-grid-pro';
...
+| <span class="prop-name">state</span>                   | <span class="prop-type">GridState</span>                                                                                                                                                                                                                                                                                                                                                | Property that contains the whole state of the grid.                                                                                                            |
+| <span class="prop-name">subscribeEvent</span>          | <span class="prop-type">&lt;E extends "resize" \| "debouncedResize" \| "viewportInnerSizeChange" \| "componentError" \| "unmount" \| "cellModeChange" \| "cellClick" \| "cellDoubleClick" \| "cellMouseDown" \| "cellMouseUp" \| ... 57 more ... \| "virtualScrollerContentSizeChange"&gt;(event: E, handler: GridEventListener&lt;...&gt;, options?: EventListenerOptions \|...</span> | Registers a handler for an event.                                                                                                                              |
```
m4theushw, line 95: "... 57 more ..."

With TypeDoc it was not very good because there was too many interfaces, but with TS not showing all options is also bad. Is it possible to not explode the keys of `GridEventsStr`? Keep only "E extends GridEventsStr"?

docs/pages/api-docs/data-grid/grid-cell-params.md
```diff
@@ -21,7 +21,7 @@ import { GridCellParams } from '@mui/x-data-grid';
...
 | <span class="prop-name">hasFocus</span>                                                      | <span class="prop-type">boolean</span>                                            | If true, the cell is the active element.                         |
 | <span class="prop-name">id</span>                                                            | <span class="prop-type">GridRowId</span>                                          | The grid row id.                                                 |
-| <span class="prop-name optional">isEditable<sup><abbr title="optional">?</abbr></sup></span> | <span class="prop-type">boolean</span>                                            | If true, the cell is editable.                                   |
+| <span class="prop-name optional">isEditable<sup><abbr title="optional">?</abbr></sup></span> | <span class="prop-type">boolean \| undefined</span>                               | If true, the cell is editable.                                   |
```
m4theushw, line 24: There's already the question mark to sign that the property is optional. Can `undefined` be removed?

image

flaviendelangle: As told in the description of the PR, it is the `typeToString` of TS which puts the `| undefined`.
Which is not wrong, but kind of heavy.
They have options on that method but I did not find one to remove it.

docs/pages/api-docs/data-grid/grid-api.md
flaviendelangle: This is bad I agree
With TypeDoc it was not possible
I'll check if with pure TS it is. And if not I can add a custom code for that property.

flaviendelangle: I got the option to remove the truncation at least

flaviendelangle:
@m4theushw I have the possibility to give more context to the `typeToStr` method.
The main diff I see is that he stop in-lining `GridRowModel` and stop adding default generics

```diff
     {
       "name": "isCellEditable",
       "description": "Controls if a cell is editable.",
-      "type": "(params: GridCellParams<any, any, any>) => boolean"
+      "type": "(params: GridCellParams) => boolean"
     },


     {
       "name": "getVisibleRowModels",
       "description": "Returns a sorted <code>Map</code> containing only the visible rows.",
-      "type": "() => Map<GridRowId, { [key: string]: any; }>"
+      "type": "() => Map<GridRowId, GridRowModel>"
     },

```

docs/pages/api-docs/data-grid/grid-cell-params.md
m4theushw: Have you tried to use `node.getText`?

```diff
diff --git a/docs/scripts/api/buildInterfacesDocumentation.ts b/docs/scripts/api/buildInterfacesDocumentation.ts
index c7257f46a..7401dd4b3 100644
--- a/docs/scripts/api/buildInterfacesDocumentation.ts
+++ b/docs/scripts/api/buildInterfacesDocumentation.ts
@@ -51,11 +51,10 @@ const parseProperty = (propertySymbol: ts.Symbol, project: Project): ParsedPrope
   tags: getSymbolJSDocTags(propertySymbol),
   symbol: propertySymbol,
   isOptional: !!propertySymbol.declarations?.find(ts.isPropertySignature)?.questionToken,
-  typeStr: project.checker.typeToString(
-    project.checker.getTypeOfSymbolAtLocation(propertySymbol, propertySymbol.valueDeclaration!),
-    propertySymbol.valueDeclaration,
-    ts.TypeFormatFlags.NoTruncation,
-  ),
+  typeStr:
+    propertySymbol.valueDeclaration && ts.isPropertySignature(propertySymbol.valueDeclaration)
+      ? propertySymbol.valueDeclaration.type?.getText() || ''
+      : '',
 });
 
 function generateMarkdownFromProperties(
```

m4theushw:
> The main diff I see is that he stop in-lining GridRowModel and stop adding default generics

I'm in favor of keeping the type exactly how it's declared in the TS file. Using `node.getText` we can get that. For the generics, they could be in the interface page, as TypeDoc does: https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IEditorOption.html. For now, the user will have to click again to see them, however, we could add later a popover to show them right when the name is hovered: https://github.com/mui-org/material-ui-x/pull/1529#discussion_r639248789

docs/pages/api-docs/data-grid/grid-cell-params.md
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/3320/commits/066f0c4daa216905786edaf9b1c7123037ab6956

Here is an example that replace `typeToString` with `getText` when available.

docs/pages/api-docs/data-grid/grid-edit-row-api.json
```diff
@@ -5,7 +5,7 @@
     {
       "name": "commitCellChange",
       "description": "Updates the field at the given id with the value stored in the edit row model.",
-      "type": "(params: GridCommitCellChangeParams, event?: MuiBaseEvent) => boolean | Promise<boolean>"
+      "type": "(\n    params: GridCommitCellChangeParams,\n    event?: MuiBaseEvent,\n  ) => boolean | Promise<boolean>"
```
m4theushw, line 8: I think we need to sanitize the output of `node.getText` to remove line breaks, spaces and the comma after the last value. A regex should do sufficient.

flaviendelangle: Yep, I finished the grouping column and did not touch this PR today.
`getText` has a good output but I need to use it everywhere and to format correctly.

I don't know how it would impact perfs, but I would be great to use `prettier.format` to have a clean output.

A Regex should be good enough at 1st though, improving the formatting is not related to this PR.

flaviendelangle: I used Prettier in the end, and the result is nice I think.

docs/scripts/api/buildInterfacesDocumentation.ts
```diff
@@ -51,224 +73,153 @@ export function generateProperties(
...
 }
 
-function generateMarkdown(
-  declaration: DeclarationContext,
-  documentedInterfaces: Map<string, boolean>,
-) {
+function generateMarkdown(object: ParsedObject, documentedInterfaces: Map<string, boolean>) {
   return [
-    `# ${declaration.name} Interface`,
+    `# ${object.name} Interface`,
```
m4theushw, line 119: The next step would be to use the name of the interface with generics, e.g. `GridCellParams<V = any, R = any, F = V>`.

m4theushw (approved):
Nothing to add here. Looks good.

docs/scripts/api/buildInterfacesDocumentation.ts
flaviendelangle: Yes I have a lot of ideas to improve the output

PR #3323, opened 2021-12-03
[test] Replace `useFakeTimers`

### LINKED ISSUE #3306: [test] Replace manually-configured fake clock with the one exposed by `createRenderer`
In https://github.com/mui-org/material-ui/pull/29684, it was added a new option to `createRenderer` to allow to set which clock implementation to use during testing. This made simpler to use a fake clock, removing the burden of importing sinon, setting up a fake clock and restoring it. The task here is to update all of our tests to relay on this new option instead of the manual configuration.

https://github.com/mui-org/material-ui-x/pull/3289/files#diff-aecf6c42ff150f1cc861e63b4b1673d61d084435d7722519b4bdab73324de851 contains an example of how it was done.

All calls, if possible, to `waitFor` should be replaced to use the mocked clock instead.

m4theushw:
Closes #3306


flaviendelangle:
Great,

I think you just have to also migrate `columnPinning.DataGridProd.test.tsx` (could you remove the `d` in the title by the way ?)

flaviendelangle (approved):

PR #3325, opened 2021-12-03
[core] Rework state update methods and deprecate `useGridApi` + `useGridState`

flaviendelangle:
Extracted from my branch to prepare a `privateApiRef`

## Goals

- Having a single way of updating the state: `apiRef.current.setState`

- Reduce the amount of deps in the `useCallback` and `useEffect` to make it clearer when they run / refresh.

Before: 

```ts
const useGridSomeFeature = (apiRef) => {
  const [, setGridState, forceUpdate] = useGridState(apiRef);

  const setFeatureState = React.useCallback((value) => {
    // We are not using the `state` returned by `useGridState` because it could be outdated
    const currentValue = apiRef.current.state.feature;
    setGridState(state => ({ ...state, feature: value }));
    forceUpdate();
  }, [apiRef, setGridState, forceUpdate);
}
```

After: 

```ts
const useGridSomeFeature = (apiRef) => {
  const setFeatureState = React.useCallback((value) => {
    const currentValue = apiRef.current.state.feature;
    apiRef.current.setState(state => ({ ...state, feature: value }));
    apiRef.current.forceUpdate();
  }, [apiRef);
}
```

- Make `apiRef.current.setState` work correctly by applying the control state constraints. 

Right now `apiRef.current.setState` exists but feels like an outdated legacy API that we should never use because it does not apply the control state constraints and therefore generate potential incoherent states.

- Prepare for private API

Right now, `useGridState` uses `apiRef.current.unstable_applyControlStateConstraint`. If we make a real private API, this method would not be available in the public one and should never be used in an exported hook.
With this PR, we could only expose `apiRef.current.setState` which would internally handle those topics.

## Deprecations

- `useGridApiRef` => we can now directly use the `apiRef.current.setState` and `apiRef.current.forceUpdate`
- `useGridApi` => I don't think it was used by users but it was exported so I am keeping it until v6

## Changelog

Should we add a `Deprecation` section or should we just keep those discrete and wait for v6 before communicating on them ?


packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -36,7 +36,6 @@ const checkFilterModelValidity = (model: GridFilterModel) => {
  * @requires useGridColumns (state, method, event)
  * @requires useGridParamsApi (method)
  * @requires useGridRows (event)
- * @requires useGridControlState (method)
```
alexfauquette, line 39: Do they require `useGridStateInitialization (method)` ?

alexfauquette (approved):
Sounds good to me, but I'm not familiar with the state management. Probably @DanailH will have a more precise review than I did

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
flaviendelangle: I remove `useGridControlState` from the `@requires` because I take for granted that any hook is after `useGridInitialization` so any hook inside `useGridInitialization` is available.

That way we only keep the list of the deps handled directly in `useDataGrid(Pro)Component` and which could be moved for another reason.

DanailH:
Sorry for the delay. I checked it and am also ok with the change - it does seem cleaner. The only question is related to the depreciation - should we have a warning massage for the deprecations in general?

packages/grid/_modules_/grid/hooks/utils/useGridApiRef.ts
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { GridApiRef } from '../../models/api/gridApiRef';
 import { GridApi } from '../../models/api/gridApi';
+import { GridState } from '../../models/gridState';
 import { EventManager } from '../../utils/EventManager';
 
 // Public developers facing overload
```
DanailH, line 7: Maybe we should think of a better name of the file that exports public API 🤔

DanailH (approved):

packages/grid/_modules_/grid/hooks/utils/useGridApiRef.ts
flaviendelangle: The `utils` folder contains stuff that can be used on several places (by opposition with the `core` or `feature` folder where you only instantiate each hook once).

But I agree that putting 100%-internal-stuff in it next to exported utility hooks is not ideal.
We could split the folder in two.

flaviendelangle:
> should we have a warning message for the deprecations in general?

I feel like the deprecation in this PR are really "just in case", those are now hooks people should have used and they should never have been exported.

But the general question remain. I can add a quick runtime warning just to make clear that it is what we have to do when we deprecate something. :+1:

PR #3278, opened 2021-11-24
[DataGridPro] Improve license file

oliviertassinari:
This change is meant to improve the DX when browsing the license files. The PR removes the `.md` extension.

Screenshot 2021-11-24 at 19 11 26

In https://mui.com/pricing/, we try to explain to the audience how to navigate the license file:

- https://unpkg.com/@mui/material@5.0.1/LICENSE 200
- https://unpkg.com/@mui/x-data-grid@5.0.1/LICENSE 200
- https://unpkg.com/@mui/x-data-grid-pro@5.0.1/LICENSE 404 ❌

I think that I was wrong to make the second link a markdown, it's never interpreted as such.

PR #3330, opened 2021-12-04
[core] Fix missing LICENSE file

oliviertassinari:
It's a regression from #3278. See how there are no license files in https://unpkg.com/browse/@mui/x-data-grid-pro@5.1.0/.


oliviertassinari:
The fix works: https://ci.codesandbox.io/status/mui-org/material-ui-x/pr/3330/builds/195281

PR #3334, opened 2021-12-04
[docs] Disable the ad when not MIT

oliviertassinari:
Take https://mui.com/components/data-grid/events/ as an example:

Screenshot 2021-12-05 at 00 09 46

We display an ad at the top and below the demos:

Screenshot 2021-12-05 at 00 10 48

This is not needed for commercial licensed software.


oliviertassinari:
cc @mui-org/x for knowledge sharing (I'm not sure this is known to most)

PR #3368, opened 2021-12-07
[DataGridPro] Add new option hideDescendantCount to Tree Data

flaviendelangle:
Extracted from #3277 

Will be properly documented and tested in the docs of the grouping columns, I don't think it's worth creating an example for the Tree Data.

### DIFF
--- a/docs/next-env.d.ts
+++ b/docs/next-env.d.ts
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
--- a/packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.tsx
+++ b/packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.tsx
@@ -3,11 +3,15 @@ import { GridApiRef } from '../../../models/api/gridApiRef';
 import { GridComponentProps } from '../../../GridComponentProps';
 import {
   GRID_TREE_DATA_GROUP_COL_DEF,
-  GRID_TREE_DATA_GROUP_COL_DEF_FORCED_FIELDS,
+  GRID_TREE_DATA_GROUP_COL_DEF_FORCED_PROPERTIES,
 } from './gridTreeDataGroupColDef';
 import { useGridApiEventHandler } from '../../utils/useGridApiEventHandler';
 import { GridEvents, GridEventListener } from '../../../models/events';
-import { GridColDef, GridGroupingColDefOverrideParams } from '../../../models';
+import {
+  GridColDef,
+  GridGroupingColDefOverride,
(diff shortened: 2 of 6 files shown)


docs/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
alexfauquette, line 2: Do you know what caused this modification?

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.tsx
```diff
@@ -51,9 +51,20 @@ export const useGridTreeData = (
       colDefOverride = propGroupingColDef;
     }
 
+    const { hideDescendantCount, ...colDefOverrideProperties } = colDefOverride ?? {};
+
+    const commonProperties: Omit<GridColDef, 'field' | 'editable'> = {
+      ...GRID_TREE_DATA_GROUP_COL_DEF,
+      renderCell: (params) => (
```
alexfauquette, line 58: Why do you not `hideDescendantCount` to the `colDef` and use this parameter in the `<GridTreeDataGroupingCell />` component like `colDef.valueOptions` for `<GridEditSingleSelectCell />`?

alexfauquette:
I agree with you on the doc. The component API should be enough

docs/next-env.d.ts
flaviendelangle: https://github.com/mui-org/material-ui-x/pull/3343

Probably this one
I checked the NextJS example and they have this line

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.tsx
flaviendelangle: On which `colDef` do you want to add it ?

flaviendelangle: I wanted to avoid adding properties on the `colDef` that are only useful for the grouping column.
But if the team feels like it's better to pass it to the `colDef` then I just have to move `hideDescendantCount` from `GridGroupingColDefOverride` to `GridColDef` and access it through the `colDef`prop in the cell component.

docs/next-env.d.ts
flaviendelangle: https://github.com/vercel/next.js/pull/28316/files

I think I found the PR

packages/grid/_modules_/grid/hooks/features/treeData/useGridTreeData.tsx
alexfauquette: Make sense, but when defining `groupingColDef` developers will probably expect their props to be in the `colDef`

alexfauquette (approved):

PR #3372, opened 2021-12-07
[test] Remove async test when possible

alexfauquette:
Seems that some of the edits are now synchronous and so do not need anymore waitFor

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
@@ -174,9 +174,8 @@ describe('<DataGridPro /> - Edit Rows', () => {
     const otherCell = getCell(2, 1);
     fireEvent.click(otherCell);
     fireEvent.focus(otherCell);
-    await waitFor(() => {
-      expect(cell).to.have.class('MuiDataGrid-cell--editing');
-    });
+
+    expect(cell).to.have.class('MuiDataGrid-cell--editing');
   });
 
   it('should allow to switch between cell mode using enter key', () => {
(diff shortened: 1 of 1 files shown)


packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
```diff
@@ -174,9 +174,8 @@ describe('<DataGridPro /> - Edit Rows', () => {
     const otherCell = getCell(2, 1);
     fireEvent.click(otherCell);
     fireEvent.focus(otherCell);
-    await waitFor(() => {
-      expect(cell).to.have.class('MuiDataGrid-cell--editing');
-    });
+
+    expect(cell).to.have.class('MuiDataGrid-cell--editing');
```
m4theushw, line 178: It's more than only removing `waitFor`. For instance, this test passes even without `onCellFocusOut`. We need to ensure that they are asserting something.

There's a promise that runs when clicking outside a cell that is in edit mode: https://github.com/mui-org/material-ui-x/blob/91c0b95f115c598901ab1025e8fd2f057c9a62a5/packages/grid/_modules_/grid/hooks/features/editRows/useGridEditRows.ts#L98

We need to wait for the promise to run. Usually, [`waitFor`](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) should be sufficient. However, it also is running with the mocked `setTimeout` so using it doesn't work. I tried to make this test to work correctly by keeping a reference to the real timeout function. It's an ugly solution but we could try until we don't have a fake-clock compatible `waitFor`.

```diff
diff --git a/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx b/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
index fc9431d00..2a3ab4b8e 100644
--- a/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
@@ -38,6 +38,8 @@ const generateDate = (
   return rawDate.getTime();
 };
 
+const nativeSetTimeout = setTimeout;
+
 describe('<DataGridPro /> - Edit Rows', () => {
   let baselineProps;
 
@@ -167,15 +169,19 @@ describe('<DataGridPro /> - Edit Rows', () => {
       />,
     );
     const cell = getCell(1, 1);
-    cell.focus();
+    fireEvent.mouseUp(cell);
+    fireEvent.click(cell);
     fireEvent.doubleClick(cell);
     expect(cell).to.have.class('MuiDataGrid-cell--editing');
 
     const otherCell = getCell(2, 1);
+    fireEvent.mouseUp(otherCell);
     fireEvent.click(otherCell);
-    fireEvent.focus(otherCell);
+    fireEvent.doubleClick(otherCell);
 
+    await new Promise((resolve) => nativeSetTimeout(resolve));
     expect(cell).to.have.class('MuiDataGrid-cell--editing');
+    expect(otherCell).to.have.class('MuiDataGrid-cell--editing');
   });
 
   it('should allow to switch between cell mode using enter key', () => {
```

PR #3387, opened 2021-12-10
[DataGridPro] Add support for master/detail

### LINKED ISSUE #211: [DataGrid] Implement Master detail/panel

m4theushw:
Preview: https://deploy-preview-3387--material-ui-x.netlify.app/components/data-grid/group-pivot/#master-detail

Closes #211

#### How does it work?

The master/detail implemented here works in two steps:

The first step is to adjust the height of the row whose we want to show the detail to include the height of the panel too. That way, the final row height is: base row height (defined via `rowHeight` prop) + detail panel height (defined via `getDetailPanelHeight` prop). Doing this, we don't need to worry about `scrollToIndexes` nor fixing the virtualization, because, in theory, it would already have #3218. The rendering of the detail panel is done later. To fill the space left for the panel, a bottom margin is applied to the row.

The second step is to render all of the detail panels. This is achieved by rendering all of them inside a div placed on top of the render zone, which can be understood as a "layer". Each panel is absolute positioned so the `top` corresponds to the `bottom` position of the row it belongs to. That way, it creates the impression that the panel is rendered immediately below the row, when it's not. I chose this approach because the alternative would be to render the panel inside the `GridRow`, which could cause some flickering when the virtualization renders new rows. The idea of adding another layer ensures that the open detail panels are rendered once and never again, no matter how much the user scrolls.


m4theushw:
I still want to add some very complex demos (e.g. combine with column pinning and use a form as the detail), but this can be done while the PR is reviewed. Don't focus too much in the changes done to support variable row height, this is not the main concern here. The changes I made are likely to change depending on #3218.

packages/grid/_modules_/grid/hooks/features/rows/useGridRowsHydration.ts
```diff
@@ -0,0 +1,82 @@
...
+    const rowHeight = gridDensityRowHeightSelector(apiRef.current.state);
+
+    setGridState((state) => {
+      const positions: number[] = [];
+      const totalHeight = rows.reduce((acc: number, row) => {
+        positions.push(acc);
+
+        const heights = apiRef.current.unstable_applyPreProcessors(
+          GridPreProcessingGroup.rowHeight,
+          { base: rowHeight }, // We use an object to make simple to check if a size was already added or not
```
m4theushw, line 38: Allowing other hooks to pre-process the row height unblocks #2144.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -112,13 +112,95 @@ You can limit the sorting to the top level rows with the `disableChildrenSorting
 
 {{"demo": "pages/components/data-grid/group-pivot/TreeDataFullExample.js", "bg": "inline", "defaultCodeOpen": false}}
 
-## 🚧 Master detail [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
+## Master detail [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
```
alexfauquette, line 115: You can also remove the 🚧 from the [feature comparison](https://deploy-preview-3387--material-ui-x.netlify.app/components/data-grid/getting-started/#feature-comparison)

For [Pricing page](https://mui.com/pricing/) I don't know if it is on X or core repo

alexfauquette (commented):
Nice documentation, I enjoyed the DataGrind inside a DataGrid. It should probably have its own page because it does not have that much relation with the tree data.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+      }
+      apiRef.current.toggleDetailPanel(params.id);
+    },
+    [apiRef, props.getDetailPanelContent],
+  );
+
+  useGridApiEventHandler(apiRef, GridEvents.cellClick, handleCellClick);
+  useGridApiEventHandler(apiRef, GridEvents.cellKeyDown, handleCellKeyDown);
+
+  const addTogleColumn = React.useCallback(
```
alexfauquette, line 120:
```suggestion
  const addToggleColumn = React.useCallback(
```

HunteRoi:
I'm eager to be able to use this feature, guys!! 

Do you have an ETA for it or should I ask Santa for a late gift? :smile:

packages/grid/_modules_/grid/components/cell/GridDetailPanelToggleCell.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+const GridDetailPanelToggleCell = (props: GridRenderCellParams) => {
+  const { id, value: isExpanded } = props;
+
+  const rootProps = useGridRootProps();
+  const apiRef = useGridApiContext();
+  const ownerState: OwnerState = { classes: rootProps.classes, isExpanded };
+  const classes = useUtilityClasses(ownerState);
+
+  const contentCache = useGridSelector(apiRef, gridExpandedRowsContentCache);
```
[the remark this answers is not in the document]
m4theushw: Do you mean to add a `hasDetailPanel(row)` method to the API? Similar to `isColumnPinned(field)`.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: To update the Pricing page it needs to be another PR in the core repo.

m4theushw:
@HunteRoi Unfortunately it won't be released this year. Probably it will be available at the beginning of January. Although, you can already use this feature in an experimental manner by pointing the dependency in `package.json` to the package generated by CodeSandbox. See https://ci.codesandbox.io/status/mui-org/material-ui-x/pr/3387/builds/203166. If you chose to try the master/detail feature, please lets us know of any bug you encounter or missing feature.

HunteRoi:
> @HunteRoi Unfortunately it won't be released this year. Probably it will be available at the beginning of January. Although, you can already use this feature in an experimental manner by pointing the dependency in `package.json` to the package generated by CodeSandbox. See https://ci.codesandbox.io/status/mui-org/material-ui-x/pr/3387/builds/203166. If you chose to try the master/detail feature, please lets us know of any bug you encounter or missing feature.

Beginning of January is good to me, but in the meantime I will try to use the CodeSandbox generated version instead. Thank you for your work! Merry Xmas and wish you already a Happy New Eve!

packages/grid/_modules_/grid/components/cell/GridDetailPanelToggleCell.tsx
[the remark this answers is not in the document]
alexfauquette: Not exactly. For now, developers have access to two parameters:
- `getDetailPanelContent`
- `getDetailPanelHeight`

And if you do not want to have a detail panel for some row, the `getDetailPanelContent` must return `null` or `undefined`. With this method, you have to evaluate all the `getDetailPanelContent(row)` to know if the detail panel is a valid component. I'm proposing to let developers provide a third prop `disableDetailPannel(row)`.

I thought about it for two reasons:
- Using `React.isValidElement` seems a bit hacky.
- I imagine for developers, knowing if the row has a detail panel will be easy. It will probably be a condition such as: "the row status is open"

alexfauquette:
@m4theushw Is it possible to make the content of the detail reachable by keyboard interaction? For example, when there is a DataGrid inside the detail, I do not find any way to access it by using the keyboard.

oliviertassinari (commented):
nitpick

packages/grid/_modules_/grid/hooks/features/rows/gridRowsState.ts
```diff
@@ -12,4 +13,9 @@ export interface GridRowsState extends GridRowGroupingResult {
    * It does not count the expanded children rows.
    */
   totalTopLevelRowCount: number;
+
```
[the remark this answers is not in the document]
m4theushw: @flaviendelangle Do you agree in keeping no blank lines between interface properties? I see you started to separate them and now we have two coding styles in the codebase.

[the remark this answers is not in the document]
flaviendelangle: Yes, I like with blank lines but let's keep consistency :+1:

m4theushw:
> Is it possible to make the content of the detail reachable by keyboard interaction? For example, when there is a DataGrid inside the detail, I do not find any way to access it by using the keyboard.

@alexfauquette Yes, you need to press <kbd>TAB</kbd> to move to the next interactive element. From there, you can use the normal arrow keys, however, if the next focused is not visible it won't scroll to it. It's not perfect since a grid inside a grid is a complex case and other things also need to be fixed but it works to some extent. Before 4a14888 it was requiring <kbd>Shift</kbd> to also be pressed to work.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+  useGridStateInit(apiRef, (state) => {
+    return {
+      ...state,
+      detailPanel: {
+        expandedRowIds: props.initialState?.detailPanel?.expandedRowIds || [],
+      },
+    };
+  });
+
+  const [, setGridState, forceUpdate] = useGridState(apiRef);
```
flaviendelangle, line 79: Do not use the deprecated `useGridState` (see #3496)

packages/grid/x-data-grid-pro/src/useDataGridProComponent.tsx
```diff
@@ -56,6 +59,7 @@ export const useDataGridProComponent = (apiRef: GridApiRef, props: DataGridProPr
   useGridPrintExport(apiRef, props);
   useGridClipboard(apiRef);
   useGridDimensions(apiRef, props);
+  useGridDetailPanel(apiRef, props);
```
flaviendelangle, line 62: We should be able to put this hook next to `useGridTreeData`, before `useGridColumns`
That way, the 1st  sync state init in `useGridColumns` already have the master detail panel pre-processor registered.
That way you have 2 pre-processing on mount instead of 3.
The more we move forward with pre-processing, the more I think `useGridColumns` and `useGridRows` should be late in the call sequence. The feature hooks should be before, they should register pre-processors and events on the rows and columns, and those should be called only once on mount, after `useGridRows` or `useGridColumns` are called.

To make it work, it seems that you just have to remove the `forceUpdate` in your new `updateCaches` method and to only run it in the `useEffect` below.
The `forceUpdate` is not needed when the `updateCaches` is called from the `rowsSet` event because `useGridRows` take care to trigger a new re-render after the event sequence.

```ts
  const updateCaches = React.useCallback(() => {
    setGridState((state) => {
      return {
        ...state,
        detailPanel: {
          ...state.detailPanel,
          ...cacheContentAndHeight(apiRef, state, getDetailPanelContent, getDetailPanelHeight),
        },
      };
    });
  }, [apiRef, forceUpdate, getDetailPanelContent, getDetailPanelHeight, setGridState]);

  useGridApiEventHandler(apiRef, GridEvents.rowsSet, updateCaches);

  React.useEffect(() => {
    updateCaches();
      forceUpdate();
  }, [updateCaches, forceUpdate]);
```

I missed it on the dedicated PR but you have the exact same thing for `useGridColumnPinning`.
All the tests pass if I move it next to `useGridTreeData` and that allows to have 1 `createColumnsState` instead of 2.
Each feature hook after `useGridColumns` causes a full pre-processing sequence which can be heavy because it causes the recomputation of the flex values.
(this can be fixed in another PR of course)

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+      if (props.getDetailPanelContent == null) {
+        return;
+      }
+
+      const content = contentCache[id];
+      if (!React.isValidElement(content)) {
+        return;
+      }
+
+      setGridState((state) => {
```
flaviendelangle, line 178: Can we call `apiRef.current.setExpandedRowIds` here ?
That way, if we add checks in `setExpandedRowIds`, all state updates will automatically have them.

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
```diff
@@ -128,6 +159,14 @@ const DataGridProVirtualScroller = React.forwardRef<
     }
   };
 
+  const getRowProps = (id: GridRowId) => {
+    if (!expandedRowIds.includes(id)) {
```
flaviendelangle, line 163: Maybe `expandedRowIds` should be an object like we discussed for the visible columns.
To avoid this linear complexity `includes` which can cause problem for very large datasets.
What would be the con of a `Record<GridRowId, boolean>` for the model ?

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
```diff
@@ -204,6 +244,47 @@ const DataGridProVirtualScroller = React.forwardRef<
...
+
+    const rowsMeta = gridRowsMetaSelector(apiRef.current.state);
+    const uniqueExpandedRowIds = [...new Set([...expandedRowIds]).values()];
+
+    for (let i = 0; i < uniqueExpandedRowIds.length; i += 1) {
+      const id = uniqueExpandedRowIds[i];
+      const content = detailPanelsContent[id];
+
+      // Check if the id exists in the current page
+      const exists = currentPage.rows.find((row) => row.id === id);
```
flaviendelangle, line 264: When pagination is disabled, this list can be very long so we should avoid looping through it twice.
We can store the `rowIndex` here instead of looking for is below, and do `if (React.isValidElement(content) && rowIndex > -1) {`

It would be even better to not have this lookup at all but I don't see an easy way of doing it.
An improvement if this seem to be a bottleneck would be to only loop the current page rows once for every panels
If the panel model was an object, it could look something like:

```tsx
const expandedPanelsRowIndexes: Record<GridRowId, number> = {}
for (let i=0; i<currentPage.rows.length; i+=1) {
  const id = currentPage.rows[i].id
  if (expandedPanelsLookup[id]) {
    expandedPanelsRowIndexes[id] = i
  }
}
```

And if the model remains an array, you can convert it to a lookup before.

That way even if you have 20 opened detail panels, you only loop on the rows once.

We could also avoid to re-run this method when the changing the rendered rows by putting the logic inside a `useMemo`

```ts
  const detailPanels = React.useMemo(() => {
    const panels: React.ReactNode[] = [];

    // Same logic

    return panels;
  }, [
    currentPage.rows,
    expandedRowIds,
    rowsMeta,
    classes.detailPanel,
    detailPanelsContent,
    detailPanelsHeights,
    getDetailPanelContent,
    rowHeight,
  ]);
```

flaviendelangle (commented):
Most of my comments are performance related because I feel like it could be a bottleneck for large dataset.

For the model structure (array vs object), we have to decide now because it's very breaking to do it after.

For the rest, I think some can be done later if you prefer.
My goal is to reduce lookup on the rows as much as possible since it can be a very long array.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+    const params = apiRef.current.getRowParams(id);
+    acc[id] = getDetailPanelHeight(params);
+    return acc;
+  }, {});
+
+  return { contentCache, heightCache };
+}
+
+export const useGridDetailPanel = (
```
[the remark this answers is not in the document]
m4theushw: I'm confused. Do you have a hook I can follow as example? I checked `useGridRowGrouping` which was the last released and it doesn't follow this order.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+  );
+
+  const getExpandedRowIds = React.useCallback<GridDetailPanelApi['getExpandedRowIds']>(
+    () => gridExpandedRowIdsSelector(apiRef.current.state),
+    [apiRef],
+  );
+
+  const setExpandedRowIds = React.useCallback<GridDetailPanelApi['setExpandedRowIds']>(
+    (ids) => {
+      setGridState((state) => {
```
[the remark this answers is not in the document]
m4theushw: Done.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
```diff
@@ -0,0 +1,246 @@
...
+    | 'onDetailPanelExpandedRowIdsChange'
+    | 'pagination'
+    | 'paginationMode'
+  >,
+): void => {
+  useGridStateInit(apiRef, (state) => {
+    return {
+      ...state,
+      detailPanel: {
+        expandedRowIds: props.initialState?.detailPanel?.expandedRowIds || [],
```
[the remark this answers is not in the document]
m4theushw: Done.

packages/grid/x-data-grid-pro/src/useDataGridProComponent.tsx
m4theushw: I reordered the hook as you asked. However it was not a simple task. Since I updated to also compute the cache synchronously, the `useGridDetailPanel` hook couldn't be before `useGridRows`. To solve that, I moved the caching logic to another hook that runs after the rows are ready.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
m4theushw: Done.

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
m4theushw: I don't have the bandwidth to update everything now, maybe in v6. I think `expandedRowIds.includes` is fine because not many rows will be in an open state at the same time.

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
m4theushw: I created a lookup and memoized it. Memoizing `detailPanels` didn't work. If lets say `expandedRowIds` changes, it will be called again but `rowsMeta.positions` might have outdated values.

packages/grid/_modules_/grid/hooks/features/detailPanel/useGridDetailPanel.ts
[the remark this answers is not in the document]
flaviendelangle: You are right... Let it as it is, I'll do a PR with a small README section to propose a fixed order and stick to it  later :+1:

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
flaviendelangle: If you think the use case where the user expand all detail panels of all pages is not something realistic, then I agree that it is negligeable.

We can wait for v6 and user feedbakcs, I'm closing this one :+1:

packages/grid/x-data-grid-pro/src/useDataGridProComponent.tsx
flaviendelangle: Yeah as soon as we starting using `getRowParams` it limit a lot what we can do...
We could probably have a workaround based on events to trigger the 1st one when the rows changes.
But I don't thinks it's worth the overhead right now.

Closing this one :+1:

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
 
-> ⚠️ This feature isn't implemented yet. It's coming.
+The master detail allows to expand a row to display additional information inside a panel.
```
mbrookes, line 329:
```suggestion
The master detail feature allows expanding a row to display additional information inside a panel.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+To start using this feature, pass a function to the `getDetailPanelContent` prop with the content to be rendered inside the panel.
```
mbrookes, line 330:
```suggestion
To use this feature, pass a function to the `getDetailPanelContent` prop with the content to be rendered inside the panel.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+Any valid React element can be used as the row detail, even another grid.
+
+The height of the detail panel content needs to be provided upfront.
+The grid assumes the value of 500px by default.
```
mbrookes, line 334:
```suggestion
The grid assumes the value of 500px by default
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+However, this can be configured by passing a function to the `getDetailPanelHeight` prop.
```
mbrookes, line 335:
```suggestion
however this can be configured by passing a function to the `getDetailPanelHeight` prop that returns the required height.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+Both props are called with a [`GridRowParams`](/api/data-grid/grid-row-params/) object, allowing to return a different value for each row.
```
mbrookes, line 336:
```suggestion
Both props are called with a [`GridRowParams`](/api/data-grid/grid-row-params/) object, allowing you to return a different value for each row.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+```tsx
+<DataGridPro
+  getDetailPanelContent={({ row }) => <div>Row ID: {row.id}</div>}
+  getDetailPanelHeight={({ row }) => 100} // Optional, default is 500px.
+/>
+```
+
+To expand a row, click in the "i" icon or press <kbd><kbd class="key">Ctrl</kbd>+<kbd class="key">Enter</kbd></kbd> inside one of the cells of the row.
```
mbrookes, line 345:
```suggestion
To expand a row click on the "i" icon, or press <kbd><kbd class="key">Ctrl</kbd>+<kbd class="key">Enter</kbd></kbd> inside one of the cells of the row.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+Returning `null` or `undefined` as the value of `getDetailPanelContent` will prevent the respective row from being expanded.
+
+{{"demo": "pages/components/data-grid/group-pivot/BasicDetailPanels.js", "bg": "inline", "defaultCodeOpen": false}}
+
+> ⚠ Always memoize the function provided to `getDetailPanelContent` and `getDetailPanelHeight`.
+> The grid bases on the referential value of these props to cache their values and optimize the rendering.
```
mbrookes, line 351:
```suggestion
> The grid depends on the referential value of these props to cache their values and optimize the rendering.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+>
+> ```tsx
+> const getDetailPanelContent = React.useCallback(() => { ... }, []);
 >
-> 👍 Upvote [issue #211](https://github.com/mui-org/material-ui-x/issues/211) if you want to see it land faster.
+> <DataGridPro getDetailPanelContent={getDetailPanelContent} />
+> ```
+>
+> ⚠ Depending on the height of the detail panel, you may see a blank space when scrolling.
+> This is caused because the grid uses a lazy approach to update the rendered rows.
```
mbrookes, line 360:
```suggestion
> This is because the grid uses a lazy approach to update the rendered rows.
```
or
```suggestion
> This is caused by the grid using a lazy approach to update the rendered rows.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+> Set `rowThreshold` to 0 to fix it:
```
mbrookes, line 361:
```suggestion
> Set `rowThreshold` to 0 to increase the rendering speed at the cost of performance:
```
(please correct my assumption...)

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+>
+> ```tsx
+> <DataGridPro rowThreshold={0} />
+> ```
+
+### Controlling the detail panels
+
+To control which rows are expanded, pass a list of row ids to the `detailPanelExpandedRowIds` prop.
```
mbrookes, line 369:
```suggestion
To control which rows are expanded, pass a list of row IDs to the `detailPanelExpandedRowIds` prop.
```
?

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+The `onDetailPanelExpandedRowIds` prop can be used to detect when a row is expanded or collapsed.
```
mbrookes, line 370: > can be used to detect when a row is expanded

"is expanded", or "is requesting to be expanded"? Is this a controlled component, or is this a callback prop? (Asking because of the preceding use of "control".)

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+On the other hand, if you only want to initialize the grid with some rows already expanded, use the `initialState` prop as follow:
```
mbrookes, line 372: Does this mean that they are expanded by default and can't be collapsed?

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+```tsx
+<DataGridPro initialState={{ detailPanel: { expandedRowIds: [1, 2, 3] } }}>
+```
+
+{{"demo": "pages/components/data-grid/group-pivot/ControlMasterDetail.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Using a form as detail panel
```
mbrookes, line 380:
```suggestion
### Using a detail panel as a form
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+As an alternative to the built-in [row editing](/components/data-grid/editing/#row-editing), a form component can be rendered inside the detail panel, allowing to edit the current row values.
```
mbrookes, line 382:
```suggestion
As an alternative to the built-in [row editing](/components/data-grid/editing/#row-editing), a form component can be rendered inside the detail panel, allowing the user to edit the current row values.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+The following demo integrates with [react-hook-form](https://react-hook-form.com/), but other form libraries can also be used instead.
```
mbrookes, line 384:
```suggestion
The following demo shows integration with [react-hook-form](https://react-hook-form.com/), but other form libraries can also be used instead.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+{{"demo": "pages/components/data-grid/group-pivot/FormDetailPanel.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Customize the detail panel toggle
```
mbrookes, line 388:
```suggestion
### Customize the detail panel toggle
```
```suggestion
### Customizing the detail panel toggle
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+To change the icon used for the toggle, you can provide a different component for the [icon slot](/components/data-grid/components/#icons) as follow:
+
+```tsx
+<DataGridPro components={{ DetailPanelToggleIcon: CustomIcon }}>
+```
+
+Although, if this is not sufficient, the entire toggle component can be overriden.
```
mbrookes, line 396:
```suggestion
If this is not sufficient, the entire toggle component can be overridden.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+To fully customize it, define a column with `field: GRID_DETAIL_PANEL_TOGGLE_FIELD`.
+The new toggle component can be provided via `renderCell`.
```
mbrookes, line 398:
```suggestion
The new toggle component can be provided via `renderCell`.
```
There's some assumed knowledge here. Without looking at the example, one might assume that `renderCell` is a prop. Can we give some contextual hints as to what it is, and/or perhaps link to a relevant section of the docs? (it doesn't come up in Algolia search)

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+It is recommended to first spread `GRID_DETAIL_PANEL_TOGGLE_COL_DEF` when defining the column, since it provides a good set of default values to start with.
```
mbrookes, line 399:
```suggestion
It is recommended to first spread `GRID_DETAIL_PANEL_TOGGLE_COL_DEF` when defining the column, since it provides a good set of default values.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+  columns={[
+    {
+      ...GRID_DETAIL_PANEL_TOGGLE_COL_DEF, // Already contains the right field
+      renderCell: (params) => <CustomDetailPanelToggle {...params}>
+    },
+  ]}
+/>
+```
+
+This approach can also be used to change the location of the toggle column, as showed below.
```
mbrookes, line 412:
```suggestion
This approach can also be used to change the location of the toggle column, as shown below.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+The grid detects the special field and does not add a second column.
```
mbrookes, line 413: Special field?

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+{{"demo": "pages/components/data-grid/group-pivot/CustomizeDetailPanelToggle.js", "bg": "inline", "defaultCodeOpen": false}}
+
+**Note**: As any ordinary cell renderer, the `value` prop is also available and it corresponds to the state of the row: `true` when expanded and `false` when collapsed.
```
mbrookes, line 417:
```suggestion
> **Note**: As any ordinary cell renderer, the `value` prop is also available and it corresponds to the state of the row: `true` when expanded and `false` when collapsed.
```
?

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+
+### Disable detail panel content scroll
+
+By default, the detail panel has as width the sum of the widths of all columns.
```
mbrookes, line 421:
```suggestion
By default, the detail panel has a width that is the sum of the widths of all columns.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+This means that, when a horizontal scrollbar is present, scrolling it will also scroll the panel content.
+To avoid this behavior, set the size of the detail panel to the outer size of the grid.
+Use `apiRef.current.getRootDimensions()` to get the latest dimension values.
+Finally, to prevent the panel from scrolling, set `position: sticky` and `left: 0`.
+
+The following demo demonstrates how this can be achieved.
```
mbrookes, line 427:
```suggestion
The following demo shows how this can be achieved.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+The grid detects the special field and does not add a second column.
+
+{{"demo": "pages/components/data-grid/group-pivot/CustomizeDetailPanelToggle.js", "bg": "inline", "defaultCodeOpen": false}}
+
+**Note**: As any ordinary cell renderer, the `value` prop is also available and it corresponds to the state of the row: `true` when expanded and `false` when collapsed.
+
+### Disable detail panel content scroll
+
+By default, the detail panel has as width the sum of the widths of all columns.
+This means that, when a horizontal scrollbar is present, scrolling it will also scroll the panel content.
```
mbrookes, line 422:
```suggestion
This means that when a horizontal scrollbar is present, scrolling it will also scroll the panel content.
```

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,114 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+To avoid this behavior, set the size of the detail panel to the outer size of the grid.
+Use `apiRef.current.getRootDimensions()` to get the latest dimension values.
+Finally, to prevent the panel from scrolling, set `position: sticky` and `left: 0`.
+
+The following demo demonstrates how this can be achieved.
+Notice that the toggle column is pinned to make sure that it will always be visible when scrolled.
```
mbrookes, line 428:
```suggestion
Notice that the toggle column is pinned to make sure that it will always be visible when the grid is scrolled.
```

mbrookes (changes requested):

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: I rephrased this entire paragraph.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: I added more context about what is `onDetailPanelExpandedRowIds`.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: "IDs" seem right.

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
m4theushw: It's not very the "rendering speed". I added more information to this warning.

flaviendelangle (approved):

joserodolfofreitas (changes requested):
I'm loving this feature!
But additionally to the inline suggestions I left on the docs, I'd like to point two UX issues that I think should be addressed:

1) When the user clicks to expand a row close to the grid footer, it's hard to see a feedback of the "expansion". You can confirm this behavior by clicking on the fifth row of the first [demo](https://deploy-preview-3387--material-ui-x.netlify.app/components/data-grid/group-pivot/#master-detail) (multiple clicks barely change the grid). There you can still see something if you pay attention, but as the row gets closer to the footer, the user may not see any feedback at all. It seems this issue will always impact the last rows. Following a short video showing this problem.

https://user-images.githubusercontent.com/550141/151578666-583dfce0-5834-468d-a39d-933dafb72e52.mp4

2) The ' i ' as default icon feels a bit out of place. First because I think the :information_source: icon doesn't really represent a master detail function (I agree it's not all bad on the examples), but second and most important, I think that by default it should represent expanded/collapsed states (**this would partially solve issue 1** - although there we may additionally need a scrolling strategy). 
I assume the usage of the :information_source: icon is meant to display a difference from the other grouping features. But I'd argue that the arrow used on the customization example works best here and should probably be the default, even when it's already used for other features.

3) Not UX issue per se, but while experimenting on issue 1, I created a demo [fork](https://codesandbox.io/s/basicdetailpanels-material-demo-forked-o3l66?file=/demo.js) that may have revealed a bug on the scrolling. 
This one is not related to the last row, but most likely to the **first expanded** row. When you expand a row for the first time out of the initial scrolling position (scroll a bit first), there's a drastic jump on the scroll that changes the view. 
The following video is recorded on latest Chrome. 
(on Firefox there's a jump but then the view ends up on the right position,
 on Safari there's no jump and it works as 'expected')

https://user-images.githubusercontent.com/550141/151578980-4a19659e-20ab-4d50-9c64-a76afac60b28.mov

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,126 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+> ```tsx
+> <DataGridPro rowThreshold={0} />
+> ```
+
+### Controlling the detail panels
+
+To control which rows are expanded, pass a list of row IDs to the `detailPanelExpandedRowIds` prop.
+Passing a callback to the `onDetailPanelExpandedRowIds` prop can be used to detect when a row was expanded or collapsed.
+
+On the other hand, if you only want to initialize the grid with some rows already expanded, use the `initialState` prop as follow:
```
mbrookes, line 372:
```suggestion
On the other hand, if you only want to initialize the grid with some rows already expanded, use the `initialState` prop as follows:
```

mbrookes (approved):

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -324,13 +324,126 @@ To do so, we need to override both the `renderCell` of the grouping column and t
...
+> ⚠ Depending on the height of the detail panel, you may see a blank space when scrolling.
+> This is caused by the grid using a lazy approach to update the rendered rows.
+> Set `rowThreshold` to 0 to force new rows to be rendered more often to fill the blank space.
+> Note that this may reduce the performance.
+>
+> ```tsx
+> <DataGridPro rowThreshold={0} />
+> ```
+
+### Controlling the detail panels
```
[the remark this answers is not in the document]
m4theushw: I went with "Controlling expanded detail panels".

m4theushw:
> 1. When the user clicks to expand a row close to the grid footer, it's hard to see a feedback of the "expansion".

@joserodolfofreitas I want to add more options to `apiRef.current.scrollToIndexes` to control where to position the row. Allowing to scroll to the detail panel could be an option, but I won't do this here. t For now, I added a small opacity when the row is expanded.

> 2. The ' i ' as default icon feels a bit out of place.

Yeah, the arrow is better but it's already taken. We need to be prepared for the worst scenario: detail panel + row grouping. In this case using the same icon for both doesn't work. We could change to "+" and "-".

> 3. Not UX issue per se, but while experimenting on issue 1, I created a demo fork that may have revealed a bug on the scrolling.

I think this can be solved with `rowThreshold=0`. There's a warning saying to use this prop but not all demos use it. I changed to use it in all demos. Updated CodeSandbox: https://codesandbox.io/s/basicdetailpanels-material-demo-forked-3ft1x?file=/demo.js

docs/src/pages/components/data-grid/group-pivot/group-pivot.md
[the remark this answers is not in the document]
m4theushw: I mentioned `initialState` only because we support it but I don't see a use case for it.

packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx
```diff
@@ -82,6 +93,21 @@ const getOverlayAlpha = (elevation: number) => {
   return alphaValue / 100;
 };
 
+const VirtualScrollerDetailPanels = styled('div', {
+  name: 'MuiDataGrid',
+  slot: 'DetailPanels',
```
DanailH, line 98: Should we document these 2 slots (ignore if it has already been done but I wasn't able to find it)?

DanailH (approved):
It looks awesome and the docs and examples are perfect, really great job! Should we aim to merge this and have it in this week's release?

m4theushw:
@flaviendelangle I changed to a "+" and "-" icon. I don't agree that we should tell users to change the icon if they want to use master detail and tree data. Kendo UI uses the same combination that we're now using: arrow for grouping and "+" for master detail.

https://www.telerik.com/kendo-angular-ui/components/grid/master-detail/
https://www.telerik.com/kendo-angular-ui/components/grid/grouping/

flaviendelangle (approved):

joserodolfofreitas (approved):
It's looking great!

corneliushka:
Any update about this feature ? Will it be available in v5.x or the upcoming major release ?

corneliushka:
Ok I'm glad it's released and available in datagrid pro, thanks you and mui team ! 
We were two searching for this and finally been able to find it by looking at released changes, then saw that a props was added to the feature recently (getDetailPanelHeight in 5.13).

For the story, 

At first, I stumbled between every documentation page (Rows, Editing, Components) having in mind words like "Edit panel" / "Detail panel" (like in material-table ) and never thought to search in Group&Pivot sub-categories, in my opinion the name 'Master detail' is not very explicit (even if this makes sense now), and maybe should be in or 'Rows' or an independent page. Maybe I'm wrong and it really belongs to Group & Pivot category because it shares some UI parts.   

Also, in mui table the equivalent feature has a different name, [collapsible table ](https://mui.com/material-ui/react-table/#collapsible-table). 

Finally, It will be great if such feature appears on [main demo ](https://mui.com/x/react-data-grid/demo/) , since we often look at it to describe a custom feature, based on what is already available on mui-x datagarid. But I understand that too much on demo can be overhelming, it's just supposed to be a demo not an exhaustive datagrid capabilities.

niralivasoya:
@flaviendelangle @m4theushw, can we close the opened row detail when we click on the other row? Is this possible with detailPanelContent in DataGridPro?

PR #3388, opened 2021-12-10
[DataGrid] Fire change event when the state changes, not when firing the prop change callback in control mode

### LINKED ISSUE #3328: [DataGrid] Can't go to next page after changing `pageSize`
### Duplicates

### Latest version

### Current behavior 😯

I have a DataGrid with 100 rows. Initially, `pageSize` is 100, so there's only one page. If I change this prop to a lower value, e.g. 50, which would display two pages, the button to go to the next page doesn't work.

### Expected behavior 🤔

It should allow to change the current page.

### Steps to reproduce 🕹

Steps:

2. Change the value in the "Page Size" combo to 25
3. Click to apply the changes
4. Click the next page button

### Context 🔦

I thought this could be something only on this demo but it happens everywhere. I tried to find the cause and I noticed that `pageSizeChange` is not being fired. In the line below, `newSubState` is already equal to the prop value so it doesn't fire the event.

https://github.com/mui-org/material-ui-x/blob/cee237389d31192aa51222555ea9a36a83ecd2dd/packages/grid/_modules_/grid/hooks/core/useGridControlState.ts#L41

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #3328 

### New behavior

When we try to update a controlled sub-state, we fire the `onXXX` prop (like today) but we do no longer fire the `GridEvents.xxxChange`

The `GridEvents.xxxChange` is now fired when the state changes.
So that internal listeners of this event can take for granted that the state did change.


m4theushw (approved):
It would be nice to add a test case covering the bug in #3328

flaviendelangle:
I'll add a test before merging :+1:

PR #3396, opened 2021-12-10
[docs] Include @mui/x-data-grid as dependency in the CodeSandbox

m4theushw:
- Go to https://mui.com/components/data-grid/pagination/#paginate-gt-100-rows
- Click to open the demo in CodeSandbox
- It doesn't run because `@mui/x-data-grid-generator` needs `@mui/x-data-grid`

### DIFF
--- a/docs/pages/_app.js
+++ b/docs/pages/_app.js
@@ -44,13 +44,13 @@ ponyfillGlobal.muiDocConfig = {
 
     if (newDeps['@mui/x-data-grid-pro'] || newDeps['@mui/x-data-grid']) {
       newDeps['@mui/material'] = versions['@mui/material'];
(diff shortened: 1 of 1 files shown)


flaviendelangle (approved):

oliviertassinari:
A side note, I believe we can kill:

https://github.com/mui-org/material-ui-x/blob/389b9e64bec7b150763c4e23b259bd8e8814b722/docs/pages/_app.js#L47

m4theushw:
@oliviertassinari Not yet, the custom cell renderers in `@mui/x-data-grid-generator` still use `@mui/styles` (we need to migrate them). However, I agree that this import should only be added if the generator is used.

PR #3424, opened 2021-12-13
[docs] Improve pagination documentation page

flaviendelangle:
[Doc preview](https://deploy-preview-3424--material-ui-x.netlify.app/components/data-grid/pagination/)

- [x] Add support for pageSize change on the service-side pagination demo

- [x] Add dynamic container height on the `autoHeight` demo to show how the grid reacts

- [x] New demo for controlled page size

- [x] Remove the `BasicPaginationGrid` and `200PaginationGrid` demo, I don't think they bring any value

- [x] Remove the `ApiRefPaginationGrid` (was not used since #3266)

- [x] Reorganize the paragraphs to have a better hierarchy

Before:
image

After:
image


docs/src/pages/components/data-grid/pagination/pagination.md
```diff
@@ -4,65 +4,75 @@ title: Data Grid - Pagination
...
+- Add a `onPageChange` callback ot load the rows when the page changes
 
-{{"demo": "pages/components/data-grid/pagination/CursorPaginationGrid.js", "bg": "inline"}}
+{{"demo": "pages/components/data-grid/pagination/ServerPaginationGrid.js", "bg": "inline"}}
 
-## Customization
+### Cursor implementation
 
-You can customize the rendering of the pagination in the footer following [the component section](/components/data-grid/components/#pagination) of the documentation.
+You can also create a cursor-based pagination.
```
alexfauquette, line 68: "create" makes me feel it is a new feature, I did not get it was the server specificity

```suggestion
You can also handle servers with cursor-based pagination.
```

docs/src/pages/components/data-grid/pagination/PageSizeAuto.js
```diff
@@ -0,0 +1,45 @@
...
+          id="main-grouping-criteria"
+          labelId="main-grouping-criteria-label"
+        >
+          <MenuItem value="300">300px</MenuItem>
+          <MenuItem value="400">400px</MenuItem>
+          <MenuItem value="500">500px</MenuItem>
+        </Select>
+      </FormControl>
+      <div style={{ height, width: '100%' }}>
+        <DataGrid autoPageSize pagination {...data} />
```
alexfauquette, line 41: `pagination` prop is in each example, even if it is useless because they all use `DataGrid` and not `DataGridPro`. Is it to remind `DataGridPro` users they need this prop?

alexfauquette (commented):
In the example "Page size options" props `pageSize` and `onPageSizeChange` are not usefull since you created a dedicated section for controlled pageSize.

I think, the order
- page size options
- automatic page size
- controlled paged size

would make more sense, because the first two are features that can be useful on their own. The last one is closely related to server-side filtering.

docs/src/pages/components/data-grid/pagination/PageSizeAuto.js
flaviendelangle: Yes, it was @oliviertassinari proposal.
I think @m4theushw agree with me that we should drop it.

flaviendelangle:
> In the example "Page size options" props pageSize and onPageSizeChange are not usefull since you created a dedicated section for controlled pageSize.

Removed

> I think, the order [...] would make more sense

Updated, it's a lot better in this order :+1:

docs/src/pages/components/data-grid/pagination/pagination.md
flaviendelangle: Done

docs/src/pages/components/data-grid/pagination/pagination.md
```diff
@@ -4,65 +4,75 @@ title: Data Grid - Pagination
 
 # Data Grid - Pagination
 
-<p class="description">Through pagination, a segment of data can be viewed from the assigned data source.</p>
+<p class="description">Easily pagination your rows and only fetch what you need.</p>
```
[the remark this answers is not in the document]
flaviendelangle: Yes :grimacing:

docs/src/pages/components/data-grid/pagination/ServerPaginationGrid.tsx
```diff
@@ -1,57 +1,78 @@
...
     };
-  }, [page, data]);
+  }, [rowsState.page, rowsState.pageSize, data.rows]);
 
   return (
     <div style={{ height: 400, width: '100%' }}>
       <DataGrid
-        rows={rows}
-        columns={data.columns}
+        columns={COLUMNS}
```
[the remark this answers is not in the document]
flaviendelangle: Oh I forgot to remove that one, I wanted to make the lazy loading cleaner but it requires to much work for this PR.

docs/src/pages/components/data-grid/pagination/PageSizeAuto.js
alexfauquette: Moreover, you insisted on the different default behavior between the two plans at the beginning. Could use a warning box to be sure devs do not miss it.

> The default pagination behavior depends on your plan.
> - On the `DataGrid`, pagination is enabled by default and can't be disabled
> - On the `DataGridPro`, pagination is disabled by default, use the `pagination` prop to enable it

alexfauquette (approved):
I let you run formatting scripts, but for me it is good. Let see if @oliviertassinari really want to keep the `pagination` prop

docs/src/pages/components/data-grid/pagination/PageSizeAuto.js
flaviendelangle: Done

docs/src/pages/components/data-grid/pagination/pagination.md
```diff
@@ -4,65 +4,75 @@ title: Data Grid - Pagination
...
+{{"demo": "pages/components/data-grid/pagination/ServerPaginationGrid.js", "bg": "inline"}}
 
-## Customization
+### Cursor implementation
 
-You can customize the rendering of the pagination in the footer following [the component section](/components/data-grid/components/#pagination) of the documentation.
+You can also handle servers with cursor-based pagination.
+To do so, you just have to keep track of the next cursor associated with each page you fetched.
 
-## Paginate > 100 rows [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
```
oliviertassinari, line 59: Why remove this demo? We broke these two anchor links

Screenshot 2021-12-31 at 19 38 26

https://mui.com/components/data-grid/getting-started/#feature-comparison

Screenshot 2021-12-31 at 19 39 06

https://mui.com/pricing/

docs/src/pages/components/data-grid/pagination/pagination.md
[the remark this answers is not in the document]
oliviertassinari: I believe it's usually a technical writing best practice to avoid using terms like "easy". I saw this mentioned in https://youtu.be/q_LlUEIJnA0?t=91.

docs/src/pages/components/data-grid/pagination/PageSizeAuto.js
oliviertassinari: > Is it to remind DataGridPro users they need this prop?

It's to make sure that developers can replace `DataGrid` with `DataGridPro` without surprises.

docs/src/pages/components/data-grid/pagination/pagination.md
flaviendelangle: ... I should have checked if there was links

Because I don't see the point of having this kind of demo.
Users do not learn anything from then compared to a line "only the pro version can paginate with more than 100 rows"
We now have long doc pages and I think we should focus on demos that demonstrate something.

oliviertassinari: > I should have looked for links

@flaviendelangle I think that it's very hard to do in practice. I bet that there are more instances of broken anchor links broken in mui.com that we are not aware of. I wish the ahrefs SEO crawler was also reporting this. I have created https://ahrefs.canny.io/site-audit/p/html-anchor-link.

> Users do not learn anything from then compared to a line "only the pro version can paginate with more than 100 rows"

I might not agree, but fair enough.

How about we only add the h2 back (without the demo)?

PR #3425, opened 2021-12-14
[DataGridPro] Add `pinnable` option

### LINKED ISSUE #3398: [DataGridProp] GridColDef missing pinnable field
### Duplicates

### Latest version

### Summary 💡

We are missing a pinnable flag in the column definition.

For other menu options, we have sortable , filterable , etc.

### Examples 🌈

_No response_

### Motivation 🔦

- for consistency
- to easily allow removing the option to pin columns from the column menu and still have a few pinned columns.

### Order ID 💳 (optional)

#19458

m4theushw:
Closes #3398


packages/grid/_modules_/grid/models/colDef/gridStringColDef.ts
```diff
@@ -10,6 +10,8 @@ export const GRID_STRING_COL_DEF: GridColTypeDef = {
   sortable: true,
   resizable: true,
   filterable: true,
+  pinnable: true,
+  editable: false,
```
m4theushw, line 14: Its default value is `false` but it was never defined.

alexfauquette (approved):
Much easier to diable pinning :)

PR #3431, opened 2021-12-14
[docs] New documentation page `State`

### LINKED ISSUE #3144: [docs] Document the selectors
Related to #1826 

The selectors are part of the public API but not properly documented.

I think it would be interesting to document them in two places:
- A `Selectors` paragraph on the feature pages (like we have an `apiRef` paragraph)
- A `Selectors` page in the new `Advanced` section (see #1826 for context).

We should be able to generate examples from the type signature of the selector. Like I am doing for the events in #3022
Each selector should also have a JSDoc description. It would be displayed on the doc and would be useful for the understanding of our codebase (even if the naming is getting better, some selector behaviors are not obvious). 

More globally, I think in the long run, any not 100% self explanatory variable / type exported from the package should have a description.

flaviendelangle:
Closes #3144 
[Doc preview](https://deploy-preview-3431--material-ui-x.netlify.app/components/data-grid/state/)

- [x] New "State" documentation page to document how to use the selectors and the `initialState` prop
- [x] New `docs:api:build` script to generate a JSON with all the selectors
- [x] Explicit warning when trying to use `useGridSelector` before state initialization instead of just letting the code crash 

For Pagination / Filter / Sorting selectors:
- [x] New auto-generated `Selectors` section on the doc page
- [x] JSDoc description for every selector and `@ignore` for selectors we don't want to expose publicly

I think we should deprecate the ill-named selectors or selectors to be removed before documenting them.
So that the selector list on the documentation is clean.
For instance selection and columns selector have not yet be renamed correctly.
If there are selectors you think we should not document, tell me so. I don't document top level selectors like `gridFilterStateSeledctor` because I would like to stop exporting them from their file. To only expose the interesting values, not the whole sub-filter.

____

I'm not a huge fan of the UI for the selector collection.
But I think the apiRef collection and events collection are not great either and that we need to rethink those 3 UIs to have something coherent.


docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+## Initialize the state
+
+Some state keys can be initialized with the `initialState` prop.
+
+> ⚠️ The `initialState` can only be used to set the initial value of the state, the grid will not react if you change the `initialState` value later on.
+>
+> If you need to fully control some models, use the control props instead (e.g. `prop.filterModel` or `prop.sortModel`).
+> More information on each feature documentation page.
+
+{{"demo": "pages/components/data-grid/state/InitialState.js", "bg": "inline", "hideToolbar": true}}
```
alexfauquette, line 18:
```suggestion
{{"demo": "pages/components/data-grid/state/InitialState.js", "bg": "inline"}}
```
Otherwise, we cannot see how you initialise the state

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+> ⚠️ This hook can only be used inside the context of the grid. Otherwise, you will have an error saying that the state is not initialized during the first render.
+
+{{"demo": "pages/components/data-grid/state/UseGridSelector.js", "bg": "inline", "hideToolbar": true}}
+
+### Catalog of selectors
+
+Some selectors are yet to be documented.
+
+{{"demo": "pages/components/data-grid/state/SelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}
```
alexfauquette, line 54:
```suggestion
{{"demo": "pages/components/data-grid/state/SelectorsNoSnap.js", "bg": "inline"}}
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you want to access sole state value in the render of your components, you can use the `useGridSelector` hook.
+
+```tsx
+const pageSize = useGridSelector(apiRef, gridPageSizeSelector);
+```
+
+> ⚠️ This hook can only be used inside the context of the grid. Otherwise, you will have an error saying that the state is not initialized during the first render.
+
+{{"demo": "pages/components/data-grid/state/UseGridSelector.js", "bg": "inline", "hideToolbar": true}}
```
alexfauquette, line 48:
```suggestion
{{"demo": "pages/components/data-grid/state/UseGridSelector.js", "bg": "inline"}}
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Direct selector access
+
+The simplest way to use a selector is to call it as a function with the state as its first argument.
+
+```tsx
+const pageSize = gridPageSizeSelector(apiRef.current.state);
+```
+
+{{"demo": "pages/components/data-grid/state/DirectSelector.js", "bg": "inline", "hideToolbar": true}}
```
alexfauquette, line 36:
```suggestion
{{"demo": "pages/components/data-grid/state/DirectSelector.js", "bg": "inline"}}
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+### With `useGridSelector`
+
+If you want to access sole state value in the render of your components, you can use the `useGridSelector` hook.
+
+```tsx
+const pageSize = useGridSelector(apiRef, gridPageSizeSelector);
+```
+
+> ⚠️ This hook can only be used inside the context of the grid. Otherwise, you will have an error saying that the state is not initialized during the first render.
```
alexfauquette, line 46: > inside the context of the grid

The only component user can design inside the context of the grid are customized components? If yes, it could be nice to add "such as custom components". Otherwise, "inside the context" seems to be abstract

docs/src/pages/components/data-grid/state/state.md
flaviendelangle: For the `ApiNoSnap` we always have the `hideToolbar: true`, it is not a React example but a component we want to inline on the page

docs/src/pages/components/data-grid/state/state.md
flaviendelangle: It can also be in render functions like the `renderCell` on `GridColDef`
But the `such as` can be interesting, I'm adding it.

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+<p class="description">Initialize and read the state of the data grid.</p>
+
+## Initialize the state
+
+Some state keys can be initialized with the `initialState` prop.
+
+> ⚠️ The `initialState` can only be used to set the initial value of the state, the grid will not react if you change the `initialState` value later on.
+>
+> If you need to fully control some models, use the control props instead (e.g. `prop.filterModel` or `prop.sortModel`).
```
DanailH, line 15:
```suggestion
> If you need to fully control specific models, use the control props instead (e.g. `prop.filterModel` or `prop.sortModel`).
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+> More information on each feature documentation page.
```
DanailH, line 16:
```suggestion
> You can find more information on the corresponding feature documentation page.
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+{{"demo": "pages/components/data-grid/state/InitialState.js", "bg": "inline"}}
+
+## Access the state [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+The state is exposed on the `apiRef` object.
+It is strongly advised not to access the state values manually because the structure of the state can change if needs be.
```
DanailH, line 23: Maybe that this sounds more descriptive?

```suggestion
It is strongly advised not to access the state values directly because the state itself is not considered a public API and its structure can change.
```

docs/src/pages/components/data-grid/state/state.md
```diff
@@ -0,0 +1,65 @@
...
+
+The `x-data-grid-pro` package exposes a set of selectors, which are functions taking the whole state as an argument and returning some value.
+You can use those to gather data from the state without worrying about the internal structure of the state.
```
DanailH, line 26:
```suggestion
The `x-data-grid-pro` package exposes a set of state selectors that take the `apiRef.current.state` as an argument and return a value.
You can use them to get data from the state without worrying about its internal structure.
```

DanailH (approved):

PR #3433, opened 2021-12-14
[DataGrid] Add `hideable` field to `GridColDef`

### LINKED ISSUE #3426: [DataGrid] Add `hideable` field to GridColDef
### Duplicates

### Latest version

### Summary 💡

We are missing a hideable flag in the column definition.

For other menu options, we have sortable, filterable, pinnable etc.

### Examples 🌈

_No response_

### Motivation 🔦

allow blocking any column from being hidden.

### Order ID 💳 (optional)

_No response_

alexfauquette:
Closes #3426 

The doc preview: https://deploy-preview-3433--material-ui-x.netlify.app/components/data-grid/columns/#hiding


flaviendelangle:
image

@alexfauquette I think "solves" is not recognized as a keywork that closes the issue

packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx
```diff
@@ -131,6 +133,7 @@ export function GridColumnsPanel() {
               <FormControlLabel
                 control={
                   <Switch
+                    disabled={column.hidable === false}
```
m4theushw, line 136: Should we show it in a disabled state or not show the option? Or even better, add a prop to `GridColumnsPanel` to allow to configure this?

image

m4theushw (commented):
@mbrookes What is more correct "hidable" or "hideable"? I only found "hidable" in the [dictionary](https://www.merriam-webster.com/dictionary/hidable), but "hideable" seems more popular among [books](https://books.google.com/ngrams/graph?content=hidable%2Chideable&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Chidable%3B%2Cc0%3B.t1%3B%2Chideable%3B%2Cc0#t1%3B%2Chidable%3B%2Cc0%3B.t1%3B%2Chideable%3B%2Cc0).

packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx
alexfauquette: I think we should show it in a disabled state because if we remove the column, it will be impossible to reorder columns by drag and drop.

mbrookes:
> What is more correct "hidable" or "hideable"? I only found "hidable" in the dictionary, but "hideable" seems more popular among books.

Both are valid, but "hideable" seems slightly more common. It might also be clearer to non-native speakers, being closer to the word "hide". I'm not sure it matters too much.

Quppa:
+1 for `hideable`, which is [a bit more common](https://www.google.com/search?q=site%3Agithub.com+"hidable") is also more common than [`hidable`](https://www.google.com/search?q=site%3Agithub.com+%22hidable%22) across GitHub, too.

m4theushw (approved):
Looks good.

oliviertassinari:
On the name of the `hideable` property: I wonder if this rule https://mui.com/guides/api/#property-naming should apply to more than the *props*. Would it bring any value?

To be clear, there are multiple different terms that are close. *prop* !== *property*, the former is specific to React JSX, the latter is for JavaScript Objects. And lastly *attributes* are for HTML elements.

PR #3441, opened 2021-12-15
[core] Increase CI efficiency

oliviertassinari:
I have applied https://github.com/mui-org/material-ui/pull/29040 here. (*a new example to illustrate the pros of mono-repo > multiple-repos*). This is in the scope of the [React infra](https://www.notion.so/mui-org/React-infra-5562c14178aa42af97bc1fa5114000cd) group.

For example, on https://github.com/mui-org/material-ui-x/pull/3437 take the restore cache step of `test_browser`:

Screenshot 2021-12-15 at 23 56 46

It uses 4 GB of storage and takes 2 minutes to happen. Now compare it to the CI build on this PR for `test_browser`:

Screenshot 2021-12-16 at 00 02 00

Better no? 😄

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,6 @@ commands:
           name: Restore yarn cache
           keys:
             - v8-yarn-{{ checksum "yarn.lock" }}
-            - v8-yarn-
       - run:
           name: Set yarn cache folder
           command: |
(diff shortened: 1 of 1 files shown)


oliviertassinari:
I think that this illustrates how when we fix something in one of the repo that might be duplicated in the other one, it's worth checking if we can't fix the other repo too.

m4theushw (approved):
Nice resource saving. Should I be added as a reviewer in future improvements made in the repo core that could be applied here too? If you hadn't opened this PR, this change would only have been done in the next time we synced the CircleCI configs.

oliviertassinari:
> Should I be added as a reviewer in future improvements made in the repo core

@m4theushw I think that we can yes. It would be wealthy to have visibility on them. If needed, we could create more groups:

Screenshot 2021-12-18 at 12 05 16

PR #3444, opened 2021-12-16
[DataGridPro] Add new prop isGroupExpandedByDefault

### LINKED ISSUE #3428: [DataGridPro] New prop `isGroupExpandedByDefault`
Part of #2994 

Users want to be able to open some groups not only based on there depth.
It is a property that AG-Grid handles and which should be simple to build.

```tsx
interface DataGridProProps {
  // .... other props
  isGroupOpenByDefault?: 
    (node: Pick<GridRowTreeNodeConfig, 'id' | 'depth' | 'groupingKey' | 'groupingField' | 'isAutoGenerated'>) => boolean
}
```

flaviendelangle:
Closes #3428

The doc will come with #3277, I don't think it's worth doing an example for the Tree Data to remove it in a few days / weeks.


packages/grid/_modules_/grid/utils/tree/buildRowTree.ts
```diff
@@ -69,6 +70,25 @@ export const buildRowTree = (params: BuildRowTreeParams): GridRowGroupingResult
 
   const groupingCriteriaToIdTree: GridGroupingCriteriaToIdTree = {};
 
+  const isGroupExpandedByDefault = (node: GridRowTreeNodeConfig) => {
+    if (!node.children || !node.children.length) {
+      return false;
+    }
+
+    if (node.depth === treeDepth - 1) {
```
alexfauquette, line 78: I do not understand why you add this condition?

If `node.depth === treeDepth - 1` it is a leaf, so `node.children` should be undefined

alexfauquette: But leafs are prevented by the previous condition `!node.children || !node.children.length`

flaviendelangle: You are right, fixed

alexfauquette (approved):

PR #3446, opened 2021-12-16
[DataGrid] Remove `line-height` from `GridCell`

### LINKED ISSUE #2859: [DataGrid] Multiple lines/children in renderCell doesn't work for MUI5
### Duplicates

### Latest version

### Current behavior 😯

Multiple lines/children in cell doesn't work for MUI5 (example of solution on [stackoverflow](https://stackoverflow.com/questions/64865959/with-material-uis-datagrid-component-how-do-i-create-multi-layered-text-in-one) works fine for `MUI4`, but not for `MUI5`);

### Expected behavior 🤔

Multiple lines/children in cell works well

### Steps to reproduce 🕹

Steps:

1. Open [codesandbox](https://codesandbox.io/s/autumn-currying-kwxef?file=/src/Demo.tsx) example and see result
2. `Devtools` disabling `lineHeight` solve this issue, but I can't find any workarounds to do it through the code (probably it assigned/calculated by scripts, mui devs knows better)

image
image

### Context 🔦

This issue blocks me. I need to render array of elements in a cell (with my own component with scroller); example of structure:
```
type Metrics = {
  key: string; 
  value: number;
};

type ServerResponse = {
  // some fields from response
  metrics: Metrics[];
}
```
And I need to show that metrics array for each entity (row) in a single cell

### Your environment 🌎

<details>

```
  Browsers: Chrome
```
image
</details>

### Order ID 💳 (optional)

_No response_

DanailH:
Fixes #2859 

A small change to the way we center content in grid cells - the result is to enable multi-lines in grid cells.

### DIFF
--- a/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
+++ b/packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
@@ -259,26 +259,14 @@ export const GridRootStyles = styled('div', {
       borderRight: `1px solid ${borderColor}`,
     },
     [`& .${gridClasses['cell--textLeft']}`]: {
-      textAlign: 'left',
+      justifyContent: 'flex-start',
     },
-    [`& .${gridClasses['cell--textLeft']}.${gridClasses['cell--withRenderer']}, & .${gridClasses['cell--textLeft']}.${gridClasses['cell--editing']}`]:
-      {
-        justifyContent: 'flex-start',
-      },
     [`& .${gridClasses['cell--textRight']}`]: {
-      textAlign: 'right',
+      justifyContent: 'flex-end',
     },
-    [`& .${gridClasses['cell--textRight']}.${gridClasses['cell--withRenderer']}, & .${gridClasses['cell--textRight']}.${gridClasses['cell--editing']}`]:
-      {
-        justifyContent: 'flex-end',
-      },
     [`& .${gridClasses['cell--textCenter']}`]: {
-      textAlign: 'center',
+      justifyContent: 'center',
     },
-    [`& .${gridClasses['cell--textCenter']}.${gridClasses['cell--withRenderer']}, & .${gridClasses['cell--textCenter']}.${gridClasses['cell--editing']}`]:
-      {
-        justifyContent: 'center',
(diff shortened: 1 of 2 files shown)


DanailH:
I marked Argos as approved - the difference was in the font.

To summarise - I removed the `line-height` because it was not possible to have multiline content in cells. You can increase the height of the cell but you can't make the text go on 2 lines for example.

The change is that now the cell is always` display: flex` and text positioning is done via `justify-content`.
Technically the `cell--withRenderer`class can be depricated. I left it a slot to avoid breaking change but to also give the ability for users to still be able to only target cells with custom components in them.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -254,34 +255,18 @@ export const GridRootStyles = styled('div', {
       alignItems: 'center',
       gridGap: theme.spacing(1),
     },
-    [`& .${gridClasses['cell--withRenderer']}`]: {
```
m4theushw, line 257: Could you update https://github.com/mui-org/material-ui-x/blob/8013d30a182139b6c1d87b23ff664f22cbf9fba3/packages/grid/_modules_/grid/gridClasses.ts#L318 and mark this class name with `// TODO v6 - remove`?

DanailH: added

m4theushw (approved):

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -213,7 +214,8 @@ export const GridRootStyles = styled('div', {
       },
     },
     [`& .${gridClasses.cell}`]: {
-      display: 'block',
+      display: 'flex',
+      alignItems: 'center',
       overflow: 'hidden',
       textOverflow: 'ellipsis',
       whiteSpace: 'nowrap',
```
m4theushw, line 221: The `textOverflow` and `whiteSpace` do nothing on flexbox. Related to #3587

PR #3447, opened 2021-12-16
[contributor] Start a markdown documentation for contributors

alexfauquette:
The idea is to have documents, where new contributors could find answers to how the DataGrid is working, such that they start looking into the code with a correct mental map.

It should answer the following questions:
- Where is the code for the bug I want to correct?
- How does this functionality is organized?


docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Accessing state
+
+The react components can access to the state with either selectors or getter methods.
+Here is an example.
+
+```js
+const apiRef = useGridApiContext();
+
+const filterModel = useGridSelector(apiRef, gridFilterModelSelector);
+const addFilterLabel = apiRef.current.getLocaleText('filterPanelAddFilter');
```
flaviendelangle, line 43: The locales are not in the state, maybe not the best example

I think we should favor selectors over methods when the selector does exactly the same thing (and even remove some methods in v6).
So maybe here we should favor an example where the method adds something compared to the selector (even if it's a oneliner)

```suggestion
const row = apiRef.current.getRow(id);
```

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `models` define the typescript interfaces of the gird
+
+## How does it work?
+
+![Workflow Scheme](./img/overviewSchemDark.svg#gh-dark-mode-only)
+![Workflow Scheme](./img/overviewSchemLight.svg#gh-light-mode-only)
+
+### Accessing state
+
+The react components can access the state with either selectors or getter methods.
```
flaviendelangle, line 36: Maybe add a link to the dev doc page ?

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+const filterModel = useGridSelector(apiRef, gridFilterModelSelector);
+const row = apiRef.current.getRow(id);
+```
+
+### Updating state
+
+To update the state, the component can either call a setter or publish an event. Here are examples with updating cell editing by setter, and handling cell key down by dispatching `GridEvents.cellKeyDown` event.
+
+```js
+apiRef.current.setEditCellValue(newParams, event);
```
flaviendelangle, line 51:
```suggestion
apiRef.current.setEditCellValue(newParams, event);
// or
```

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+![Workflow Scheme](./img/overviewSchemDark.svg#gh-dark-mode-only)
+![Workflow Scheme](./img/overviewSchemLight.svg#gh-light-mode-only)
+
+### Accessing state
+
+The react components can access the state with either selectors or getter methods.
+Here is an example.
+
+```js
+const apiRef = useGridApiContext();
```
flaviendelangle, line 40:
```suggestion
const apiRef = useGridApiContext();
// or
```

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+### Updating state
+
+To update the state, the component can either call a setter or publish an event. Here are examples with updating cell editing by setter, and handling cell key down by dispatching `GridEvents.cellKeyDown` event.
+
+```js
+apiRef.current.setEditCellValue(newParams, event);
+apiRef.current.publishEvent(GridEvents.cellKeyDown, params, event as any);
+```
+
+Publishing an event allows triggering multiple features. For example, `GridEvents.cellKeyDown` is listened by the focus, edit, and navigation feature hooks. Each of them defines an event handler responsible of modifying the state.
```
flaviendelangle, line 55:
```suggestion
An event can be consumed by multiple listeners.
For example, `GridEvents.cellKeyDown` is listened by the focus, edit, selection and navigation feature hooks. 
Each of them defines an event handler responsible for modifying there sub-state.
```

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```js
+useGridApiEventHandler(apiRef, GridEvents.cellKeyDown, handleCellKeyDown);
+```
+
+The other interest of events is that developers can listen to them by subscribing to the event for example [the documentation](https://mui.com/components/data-grid/events/) or by providing a `on<Event>` prop such as `onEditCellPropsChange`. Notice that all the events do not have an associated prop `on<Event>`
```
flaviendelangle, line 61:
```suggestion
The other interest of events is that developers can listen to them by [subscribing to the event](https://mui.com/components/data-grid/events/#subscribing-to-events) or by providing a `on<Event>` prop such as `onEditCellPropsChange`. 
Notice that all the events do not have an associated prop `on<Event>`
```

docsTech/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Specific features
+
+Here you will find more precision about how some features are designed.
+
+- [filtering](./filtering)
+- [virtualization](./virtualization)
```
flaviendelangle, line 68: If we use the word "feature" for user-facing stuff  (like you did in `defines logics for end-users: selection, filtering, editing, ...`, and I fully agree)

Then maybe we should move the `virtualization` to a `Core concepts` paragraph.
In this paragraph I will also add pages about the plugin system (execution order, state init, pre-processor, ...)

docsTech/filtering.md
```diff
@@ -0,0 +1,38 @@
+# Filtering
+
+## Structure
+
+### Data
+
+First, let's see how the filtering is saved in the state. Here are the types of the filter model. It is made of a `linkOperator`, and a list of items. Each item corresponds to a line in the filtering panel. Each item corresponds to a line in the filter panel.
+
+```ts
+interface GridFilterModel {
```
flaviendelangle, line 10: For the `Data`, I would just link to the doc api pages to avoid outdated information here.

docsTech/README.md
alexfauquette: I was looking to have same structure as hook folders in which `virtualization` is a feature, which makes sense because, that's something developers can have access to if (by disabling it of modifying buffer)

In opposite, execution order, state init, or pre-processor should be transparent to developers.

alexfauquette:
I merged it such that if you want to add/modify something you can do it easily

Issue #1076, opened 2021-02-18
[DataGrid] Allow sortable indicator on column headers without a sort applied


emily-ellevation:
When a sortable column is not sorted is there any way to surface a 3rd icon type with the existing API? As it stands, it appears the sort icon is removed from the DOM altogether when a column is unsorted.

**3rd Type: Sortable but no sort applied:**
<img width="22" alt="Screen Shot 2021-02-18 at 2 47 47 PM" src="https://user-images.githubusercontent.com/52083969/108431866-53ff4b00-71f8-11eb-966e-9daa0ee56da6.png">
Ascending:
<img width="20" alt="Screen Shot 2021-02-18 at 2 47 37 PM" src="https://user-images.githubusercontent.com/52083969/108431849-4ea20080-71f8-11eb-875f-5ae15d4e8b9c.png">
Descending:
<img width="22" alt="Screen Shot 2021-02-18 at 2 47 56 PM" src="https://user-images.githubusercontent.com/52083969/108431860-52358780-71f8-11eb-9df8-0b4311c18c77.png">

- [X] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Columns that have `sortable:true` but have a 'null' sort direction can display a 3rd neutral sort icon to indicate they are sortable.

## Examples 🌈

In the following example, the table is sorted by Student Name, School has no sort applied but displays a neutral 'unsorted/sortable' indicator.

<img width="342" alt="Screen Shot 2021-02-18 at 2 42 29 PM" src="https://user-images.githubusercontent.com/52083969/108431350-9b390c00-71f7-11eb-87e7-432fc1b99638.png">

## Motivation 🔦

We focus a lot on explicit discoverability in our UIs. Unsorted columns having no indicator that they are sortable does not allow us to give explicit feedback that columns _can_ be sorted and causes users not to use sorting. I'm aware I can apply a default sort with sort model, but this does not guarantee users can discover that _other_ columns are sortable.

dtassone:
Good point. ATM the null sort has no icon but yes we should allow a custom icon in components to override that behavior.

emily-ellevation:
Awesome, glad to hear. Do you have a sense of priority/timing? Is this something I could possibly take a shot at as a first issue?

oliviertassinari:
[...]

PR #3449, opened 2021-12-17
[DataGrid] Add `sortingOrder` to `GridColDef`

Quppa:
Adds `sortingOrder` to `GridColDef`, allowing [`sortingOrder` on `DataGrid`](https://mui.com/components/data-grid/sorting/#sort-order) to be overridden on a per-column basis. An example use case is where a name column would naturally be sorted in ascending order, but downloads in descending order.


docs/src/pages/components/data-grid/sorting/OrderSortingPerColumnGrid.tsx
```diff
@@ -0,0 +1,66 @@
...
+  {
+    id: 8,
+    platform: 'osx-x64',
+    downloads: randomQuantity(),
+  },
+];
+
+export default function OrderSortingPerColumnGrid() {
+  return (
+    <div style={{ height: 400, width: '100%' }}>
```
m4theushw, line 50: Could you use `useDemoData` here, instead of hardcoding the rows and columns? The quantity column could be used instead of the downloads.

m4theushw (commented):
Thanks for your contribution. Could you add a test case?

Why did you remove `ColumnUnsortedIcon`? It was added to satisfy #1076.

docs/src/pages/components/data-grid/sorting/OrderSortingPerColumnGrid.tsx
Quppa: Updated demo code (not sure how readable this is):
```tsx
<DataGrid
  {...data}
  columns={data.columns.map((column) => ({
    ...column,
    sortingOrder:
      column.field === 'quantity' ? ['desc', 'asc', null] : undefined,
  }))}
  sortingOrder={['asc', 'desc', null]}
/>
```

Quppa:
Thanks for the review - `GridColumnUnsortedIcon` seemed redundant since it just returns `ColumnSortedAscendingIcon` or `ColumnSortedDescendingIcon` based on `nextSortDirection`, and that logic now lives in `GridColumnHeaderSortIcon`, but I see that it's still useful for the reasons described in #1076. That said, there isn't any logic left in `GridColumnUnsortedIcon.tsx` now - is it fine to leave it in this state?

Quppa:
Added some tests and switched to passing along `nextSortDirection` using a data attribute on `Icon` (`data-nextsortdirection`). Not at all sure that this is the most elegant approach - am I missing an obvious way to pass a custom attribute (`nextSortDirection`) to the DOM element `Icon`?

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSortIcon.tsx
```diff
@@ -30,24 +31,35 @@ const useUtilityClasses = (ownerState: OwnerState) => {
...
+  let Icon;
   if (direction === 'asc') {
     Icon = icons.ColumnSortedAscendingIcon;
   } else if (direction === 'desc') {
     Icon = icons.ColumnSortedDescendingIcon;
+  } else {
+    Icon = icons.ColumnUnsortedIcon;
+    const [nextSortDirection] = sortingOrder;
+    return Icon ? (
+      <Icon fontSize="small" className={className} data-nextsortdirection={nextSortDirection} />
```
m4theushw, line 49: Do you need `data-nextsortdirection` for your application? Is not `aria-sort` already sufficient? I'm asking because the logic could be moved to `GridColumnHeaderSortIcon` if the attribute is only used to decide which icon to display.

packages/grid/x-data-grid/src/tests/sorting.DataGrid.test.tsx
```diff
@@ -91,6 +91,67 @@ describe('<DataGrid /> - Sorting', () => {
     expect(getColumnValues()).to.deep.equal(['Adidas', 'Puma', 'Nike']);
   });
 
+  it('should only allow ascending sorting using sortingOrder', () => {
```
m4theushw, line 94: Thanks for adding tests for the component-wide `sortingOrder`. 💪

Quppa: No problem - I did notice some funky (existing) behaviour when `null` is used in `sortingOrder` in somewhere other than the last position, e.g. `['asc', null, 'desc']` will always cycle between sorting by ascending and the initial order, never descending - I wonder if this should be noted in the docs or even throw an error at runtime.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSortIcon.tsx
Quppa: I think this is only used for the mouse-over hint for the next sort direction when the current state is unsorted. I couldn't see an obvious way to contain the logic to `GridColumnHeaderSortIcon` while still returning `GridColumnUnsortedIcon` for the unsorted state.

Good point about `aria-sort` - I've switched to that instead of the custom data attribute.

m4theushw: `aria-sort` can't be used twice. 

From https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort

> If a grid or table provides sort functionality, the aria-sort attribute should be set to either ascending or descending (or other) on the header cell element for the sorted column or row.

I pushed a small change to pass the `sortingOrder` prop directly to the icon, no data attribute. Are you OK with this approach?

packages/grid/x-data-grid/src/tests/sorting.DataGrid.test.tsx
m4theushw: I guess this prop should only accept `null` at the edges, never in the middle. It might be confusing to go from ascending, no sorting applied, and descending. Let's wait for someone to open an issue with an use case that requires `null` in this position.

packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderSortIcon.tsx
Quppa: Thanks - looks great and is what I'd intended to do in https://github.com/mui-org/material-ui-x/pull/3449/commits/b94f3bf977b2eb2a705bc111cb801f01999c0cab (still confused what the difference between that and your working https://github.com/mui-org/material-ui-x/pull/3449/commits/a3b11f1a793aeda8598077ec14621b4f2a2811c6 version is).

m4theushw: In b94f3bf you were using the `sortingOrder` prop and also spreading it into the DOM element, that's reason for the "React does not recognize the `sortingOrder` prop on a DOM element." warning. In a3b11f1, I don't spread it into the DOM node.

m4theushw (approved):

PR #3455, opened 2021-12-18
[DataGrid] Fix scroll jump when selection checkbox is checked

### LINKED ISSUE #3397: [DataGrid] The grid not scrollable after shift select multiple rows
### Duplicates

### Latest version

### Current behavior 😯

_No response_

### Expected behavior 🤔

_No response_

### Steps to reproduce 🕹

Steps:

1.https://mui.com/components/data-grid/demo/
2.shift select multiple rows
3.scroll the grid

### Context 🔦

v 5.2.0

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3397 

I removed the `HTMLElement.focus()` call because the cell already puts the focus on the element with `tabindex=0`:

https://github.com/mui-org/material-ui-x/blob/1ed95e17aa46cd1298957b428d2d3b3173acf673/packages/grid/_modules_/grid/components/cell/GridCell.tsx#L168-L172

### DIFF
--- a/packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
+++ b/packages/grid/_modules_/grid/components/columnSelection/GridCellCheckboxRenderer.tsx
@@ -25,7 +25,7 @@ const useUtilityClasses = (ownerState: OwnerState) => {
 
 const GridCellCheckboxForwardRef = React.forwardRef<HTMLInputElement, GridCellParams>(
   function GridCellCheckboxRenderer(props, ref) {
-    const { field, id, value, tabIndex, hasFocus } = props;
+    const { field, id, value, tabIndex } = props;
     const apiRef = useGridApiContext();
     const rootProps = useGridRootProps();
     const ownerState = { classes: rootProps.classes };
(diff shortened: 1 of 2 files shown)


alexfauquette:
Yes, removing useless useLayout is good, but I'm not sure it is the root cause. Using index as a key leads to other bugs:

### Checkbox Ripple appears on cells without focus

When scrolling, the Ripple appears again at the same position, event if the cell is different

https://user-images.githubusercontent.com/45398769/146742260-ecc207e9-ae3b-4bab-9c30-07933ea07ef6.mp4

### Action Cell focus has the same problem as Checkbox

Set the focus on an action cell, and you will not be able to scroll

https://user-images.githubusercontent.com/45398769/146742328-2c55355c-f99a-424f-837e-ab03cb9be87c.mp4


I wondering if it is not all the useEffect observing props updates that are broken. For now, we only use them to fiear focus. But developers could implement custom logic with useEffect, which will be triggered when scrolling

m4theushw:
Closed in favor of #3483

PR #3484, opened 2021-12-20
[DataGrid] Avoid rerender when pressing key inside already focused cell

### LINKED ISSUE #3480: [DataGrid] Cells are rerendered at every event which drastically decreases performance
### Duplicates

### Latest version

### Current behavior 😯

Grid cells rerender at every event (for example focus, keydown, click, etc.). 
Even in cell editing mode all cells in the grid will be rerendered with every keydown event. 
Also it's not possiibe to disable event listeners with passing null as props to event callback (for example, `onCellKeyDown={null}`
It makes impossible to use Data Grid with large tables (at least 50 row and 5 columns).

### Expected behavior 🤔

Cells should not be rendered at every event.

### Steps to reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/datagrid-performance-issue-zqn91
2. Open console
3. Enable edit mode for the first column
4. Type something
5. Take a look at the count of 'Render' outputs

### Context 🔦

I'm trying to build a large table with 50 visible rows and 8 columns with 4 custom components. Currently performance is too low to be able to use this library.

### Your environment 🌎

see code sandbox

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3480

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
+++ b/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts
@@ -12,6 +12,7 @@ import { useGridApiEventHandler } from '../../utils/useGridApiEventHandler';
 import { GridComponentProps } from '../../../GridComponentProps';
 import { isNavigationKey } from '../../../utils/keyboardUtils';
 import { useGridStateInit } from '../../utils/useGridStateInit';
+import { gridFocusCellSelector } from './gridFocusStateSelector';
 
 /**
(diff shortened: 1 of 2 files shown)


xiaoyu-tamu:
@m4theushw I just noticed that when a cell has focus ( not in editing mode ), pressing any key will re-rendering the grid.

e.g. press the CMD key on macOS re-rendering the whole grid, which has no meaning at all.

Is this PR also fix this?

DanailH (approved):

PR #3486, opened 2021-12-21
[test] Warn when focusing cells without syncing the state

m4theushw:
Previously, every time we had to `fireEvent.keyDown` in a cell, we would usually focus the cell with `cell.focus()` first. This approach has the problem that it doesn't sync the focus state, because the grid doesn't listen to the `focus` event, but only `mouseup` and `keyup`. For some simple test cases this is not a problem, however, if the focus moves from one cell to another this might cause an unexpected behavior. Take as example the test I had to modify in #3484:

```tsx
const { setProps } = render(<NavigationTestCaseNoScrollX />);

getCell(8, 1).focus();
expect(getActiveCell()).to.equal('8-1');
fireEvent.keyDown(document.activeElement!, { key: 'End', ctrlKey: true });
expect(getActiveCell()).to.equal('9-2'); // The focused cell according to the state is 9-2

getCell(8, 1).focus(); // Try to focus another cell
expect(getActiveCell()).to.equal('8-1');
setProps({ foo: 'bar' }); // Simulates a rerender
expect(getActiveCell()).to.equal('8-1'); // It fails
```

Above we're navigating with the keyboard to the cell 9-2. After the keyboard event, the state is correctly updated to store that 9-2 is the focused cell. The problem comes if we try to focus another cell. For a brief moment it works, but if a rerender occurs in the meantime, the focus gets reset to the value in the state. This PR shows a warning every time that a cell receives focus while the state is not correctly synced. To fix the test above, we can drop `cell.focus()` and simulate a click:

```diff
-cell.focus();
+fireEvent.mouseUp(cell);
+fireEvent.click(cell);
```

If you also need to ensure that a row is not selected when clicking, use `disableSelectionOnClick` or first focus on a column header and navigate with the arrow keys to the desired cell. The alternative to this PR would be to update the focus on the `focus` event, but I don't know if it's worth it to change this now. The `focus` is always preceded by a click or keyboard event and we already use these events.


packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -177,6 +180,30 @@ function GridCell(props: GridCellProps) {
     }
   });
 
+  let handleFocus: any = null;
+
+  if (process.env.NODE_ENV === 'test') {
```
flaviendelangle, line 185: We should probably also call `other.onFocus` since it can be overridden with `componentsProps`

packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
```diff
@@ -38,6 +38,11 @@ const generateDate = (
...
 };
 
+function fireClickEvent(cell: HTMLElement) {
+  fireEvent.mouseUp(cell);
+  fireEvent.click(cell);
+}
+
 const nativeSetTimeout = setTimeout;
 
 // TODO: Replace `cell.focus()` with `fireEvent.mouseUp(cell)`
```
flaviendelangle, line 48: I think we can remove this comment

flaviendelangle (approved):

packages/grid/x-data-grid-pro/src/tests/editComponents.DataGridPro.test.tsx
```diff
@@ -318,7 +324,9 @@ describe('<DataGridPro /> - Edit Components', () => {
         />,
       );
       const cell = getCell(0, 0);
-      cell.focus();
+      fireEvent.mouseUp(cell);
+      fireEvent.click(cell);
```
alexfauquette, line 328: why not `fireClickEvent(cell)`?

packages/grid/x-data-grid-pro/src/tests/editComponents.DataGridPro.test.tsx
```diff
@@ -104,6 +104,7 @@ describe('<DataGridPro /> - Edit Components', () => {
       );
 
       const cell = getCell(0, 0);
+      fireEvent.mouseUp(cell);
```
alexfauquette, line 107: Is it necessary?
In `cellEditing.DataGridPro.test.tsx` you remove all the `cell.focus()` but do not add `fireEvent.mouseUp()`

alexfauquette (approved):
A big thanks for this PR, it explains why I was struggling to write tests on cell focus yesterday 😅

packages/grid/x-data-grid-pro/src/tests/editComponents.DataGridPro.test.tsx
flaviendelangle: Maybe we could use `mousePress` utility method from the core

m4theushw: `mousePress` doesn't work. Some tests fail but I don't know why.

packages/grid/x-data-grid-pro/src/tests/editComponents.DataGridPro.test.tsx
m4theushw: Yes, even though it's a double click, it uses the cell that received `mouseUp` to update the state. Below, `cellParams` is assigned by the `mouseUp` listener. If we don't fire it, the focus is never updated.

https://github.com/mui-org/material-ui-x/blob/9285a97a5e19eca7d06cb8f3d06e0ee1a1f05823/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts#L167-L168

PR #3437, opened 2021-12-15
[docs] Improve filtering documentation page

flaviendelangle:
[Doc preview](https://deploy-preview-3437--material-ui-x.netlify.app/components/data-grid/filtering)

### Improve demos
- [x] Rework the demo to have 1 demo = 1 purpose (do not control the model in other demo that `ControlledFilters` for instance)
- [x] Add a `singleSelect` column on the `FilterOperators` demo
- [x] In the `CustomInputComponent` demo, hide the input component when the operator do not need it 
- [x] Use the same set of columns on all demo 

### Improve paragraph order

Before:

image

After:

I tried to order the section like follow:

1. How an end-user can use the filters
2. How a developer can use the built-in filters
3. How a developer can customize its own filters
4. Full Examples

image

PR #3488, opened 2021-12-21
[docs] Generate imports dynamically from the packages export list

flaviendelangle:
Extracted from #3437

- [x] When generating the markdown page of an interface, generate the imports from all the packages that do export it

- [x] Handle interface in-lining in the markdown interface page 

My use case is to document `GridFilterModel` without creating a dedicated page for `GridFilterItem` and `GridLinkOperator`.

I you have better ideas on how to handle this subject I'm interested.
Right now I am adding a list of names in the `@inline` JSDoc tag 

```ts
/**
 * Model describing the filters to apply to the grid.
 * @inline GridFilterItem
 */
export interface GridFilterModel {
```

[See the demo preview from #3437](https://deploy-preview-3437--material-ui-x.netlify.app/api/data-grid/grid-filter-model/)

- [x] Do not loose the JSDoc description and tags from the types re-exported with `export type { XXX } from './module'`


m4theushw (commented):
It's nice that we're removing the hardcoded imports.

Could you remove the part that inlines the interfaces? In theory, the PR should be only fixing the imports. I had to manually pull some changes from #3437 to able to test the new interface docs.

Regarding having docs for multiple interface in the [same page](https://deploy-preview-3437--material-ui-x.netlify.app/api/data-grid/grid-filter-model/), I don't know if we should do this. I was thinking that each interface would have its own page, no matter its size.  That way, when searching, Algolia would put the interface name as the result name. However, with inlined interfaces, `GridFilterItem` will be displayed similar to "Properties" below. 

image

Another point to debate. What should happen if an interface is shared between other very distinct interfaces. For instance, [`GridFilterOperator`](https://deploy-preview-1212--material-ui-x.netlify.app/interfaces/gridfilteroperator.html) uses `GridFilterItem` so it might make sense to document it in the `GridFilterModel` page. But `GridColDef` also needs `GridFilterOperator`, what to do in this case?

flaviendelangle:
> Could you remove the part that inlines the interfaces?

I can split the logic into 2 PRs yes

> What should happen if an interface is shared between other very distinct interfaces. 

I don't handle enum inlining but if I did I would put it in both.

If we do not inline at all, would you create a dedicated page for `GridFilterOperator`
Or would you inline enum but not interfaces ?

m4theushw (approved):

m4theushw:
> I don't handle enum inlining but if I did I would put it in both.

To be clear, `GridFilterOperator` is not an enum, but an interface.

> If we do not inline at all, would you create a dedicated page for GridFilterOperator
> Or would you inline enum but not interfaces ?

I would keep one page for each interface and enum, like TypeDoc does. Maybe only aliases should not have a dedicated page. If the problem that you're trying to solve with inline docs is to avoid the user having to navigate to another page to gets docs for a small interface, then we could show a popover with a smaller version of the docs for that interface when hovering its name.

Issue #3381, opened 2021-12-08
[DataGrid][ShadowDOM] Add GridMenu to GridSlotsComponent to override Popper.container -> Portal.container


hafley66:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

```tsx
const shadowDomChildDiv = someDiv;
const gridProps: DataGridProps = {
    componentsProps: {
        // Corresponds to UL under the popper that powers triple dot menu
        columnMenu: {},
        // These are Portal'ed components
        columnsPanel: { container: shadowDomChildDiv },
        filterPanel: { container: shadowDomChildDiv },
        panel: { container: shadowDomChildDiv },
        preferencesPanel: { container: shadowDomChildDiv },
        // I am asking for this to be added, corresponding to GridMenu in the material-ui-x codebase
        menu: { container: shadowDomChildDiv },
    }
}
```

I am opening another card to allow for defaulting the props to portal in mui/material, but this stands as a request to add GridMenu to GridSlotsComponent.

It should be possible to render elements using `Portal` inside of a shadow dom without losing styles. You can override `DataGridProps.componentsProps`+`@mui/material.theme` to default the `Portal.container` for many elements, except for portal itself, and except for the GridMenu, used in various places in DataGrid.

It is possible to override Portal props for Column Header tooltips by customizing the theme's default props for Tooltip, but since there is no ability to customize default props for Popper/Portal to propogate the shadow dom container, all of the `styled(Popper)` components will appear outside and have no styling applied. 

Some are not in the GridComponentSlots mechanism, so it is not possible to override props or component to change the `Portal.container`.

### Examples 🌈
Related Issue and context: https://github.com/mui-org/material-ui-x/issues/3319

https://codesandbox.io/s/https-github-com-mui-org-material-ui-x-issues-3319-frydd

Please click on triple dot. See how its styles are not applied since emotion is injecting into shadow dom root, and the portal container is always outside shadow dom (we also cannot pass disablePortal, but overriding container is better solution in general).

Then see how Filter and the columns selectors are styled correctly, that is because their props are customized by the props of data grid itself.

### Motivation 🔦

I currently host newer frontends in our larger monolithic frontend application. There are very old css/html resets being applied that I simply cannot afford to remove due to the nature of the problem. Micro frontend techniques have been a boon in this regard, and I needed shadow dom to avoid hostile styling overrides that are unpredictable. 

[...]

Issue #2522, opened 2021-09-02
[data grid] Add recipe using a Portal to render components from the DataGrid outside its context


flaviendelangle:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

This RFC aims at deciding whether or not we want to allow users to use our components outside of the scope of the `DataGrid`, and if so how could we make it work.

We currently have one example in our internal Storybook with a `GridPreferencesPanel` used outside of the grid scope. But it is not working anymore due to the introduction of the `useGridRootProps` hook which is not accessible outside of the grid scope.

https://github.com/mui/mui-x/blob/af50a4a9974f14afc09ce7639da4f326909c7270/packages/storybook/src/stories/playground/customize-components.stories.tsx#L411

____

If we want to allow this pattern, how could we make the `rootProps` correctly accessible ? If we just export the provider and the user gives the input props to it, our component will have access to the unprocessed props (meaning it won't have the built in classnames for instance, or the `DataGrid` free version forced props).
And as we move away from the single ref with everything in-it, this problem will probably occur a lot more.

oliviertassinari:
Looking at the source:

https://github.com/mui-org/material-ui-x/blob/34c38ca2150525e063d6ba7938a5176c7df24b23/packages/storybook/src/stories/playground/customize-components.stories.tsx#L411-L441

I would +1 for:

Removing the `useGridApiRef()` hook and `GridApiContext.Provider` from the public API. This hook currently serves two different use cases for the public usrs. First for composed data grid, to get a ref on the state. Using a useRef should be enough. Second, to initiate the state when creating a custom data grid. I would advocate that we could create a wrapper root component that takes care of the initial setup, with the context providers, hooks etc.

With such a direction, we would solve the composition problem of this story.

flaviendelangle:
> Making the root props provider public. Basically, make the APIs used in the DataGrid and DataGridPro components public

On this point, how do you solve this  ?

> If we want to allow this pattern, how could we make the rootProps correctly accessible ? If we just export the provider and the user gives the input props to it, our component will have access to the unprocessed props (meaning it won't have the built in classnames for instance, or the DataGrid free version forced props).

m4theushw:
```diff
diff --git a/packages/storybook/src/stories/playground/customize-components.stories.tsx b/packages/storybook/src/stories/playground/customize-components.stories.tsx
index bfa51b08..ceadde52 100644
[...]

PR #3490, opened 2021-12-21
[DataGrid] Add a subset of used `@mui/material` components to `GridSlotsComponent`

### LINKED ISSUE #3066: [RFC] Make base components interchangeable
### Problem

When components from MUI Core are used to compose other components in the DataGrid, in certain cases the variant or other prop is hardcoded to make them look correctly. This becomes a problem if the user wants to use a different variant (see https://github.com/mui-org/material-ui-x/issues/2944) or provide additional props by default. Currently, the default props added via theme doesn't work so there's no way to change how these components are rendered. 

https://github.com/mui-org/material-ui-x/blob/cc5e8fa6f2ccc8dc3b2660d63c8ead7a440be59e/packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx#L123

https://github.com/mui-org/material-ui-x/blob/cc5e8fa6f2ccc8dc3b2660d63c8ead7a440be59e/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx#L167

There's already the `components` and `componentsProps` which allow to override some components. However, these slots are only available for DataGrid's own components. As an example, to override the column selector in the filter panel, the user has to override the entire component.

This RFC wants to discuss how to shape the DataGrid props to allow the base components to be changed too. The natural path would be to add more slots, however, their names might collide (e.g. is `components.Pagination` the pagination component or the MUI component used to build the pagination component?).

### Proposed solution

1. Add `baseComponents` and `baseComponentsProps` props. These props would have by default the MUI Core components that are used internally:
    
    ```tsx

    <DataGrid baseComponents={{ TextField }} baseComponentsProps={{ textField: { variant: 'outlined' } }} />
    ```

1. Add the base components inside `components` and `componentsProps` props:
    
    ```tsx

    <DataGrid components={{ Base: { TextField } }} componentsProps={{ base: { textField: { variant: 'outlined' } } }} />
    ```

1. Use a prefix to distinguish the base components from the slots:
    
    ```tsx

    <DataGrid components={{ BaseTextField: TextField } }} componentsProps={{ baseTextField: { variant: 'outlined' } } }} />
    ```

Mentioning @michaldudak to help to propose other names.

DanailH:
Part of #3066 

I've made a subsite of all `@mui/material` components we use in the grid be available as a slot component. I've checked all the `@mui/material` we use and these once seemed to make the most sense for a first iteration. I'm adding the rest below. We can discuss and see if any of the others also need to be added in this first iteration (or if I need to revert any of the ones I added).

---

**List of remaining `@mui/material` components currently used in the grid:**
- CircularProgress
- TablePagination
- MenuList
- MenuItem
- Paper
- Grow
- Popper
- IconButton
- ListItemIcon
- InputBase
- InputLabel
- FormControlLabel
- Box
- Badge
- Tooltip
- NoSsr
- ClickAwayListener
- Grow
- MuiDivider


flaviendelangle:
Could we add a `Form input` section [here](https://mui.com/components/data-grid/components/#components)

DanailH:
> Could we add a `Form input` section [here](https://mui.com/components/data-grid/components/#components) ? To show how to use it to switch to another variant for instance.

Sure, but I didn't add all the form elements we use from the core. I skipped `InputBase`, `InputLabel`, `FormControlLabel`. If we are adding a `Form input` section in the docs maybe its better to also add slots for those components before that?

flaviendelangle:
Yes, if you prefer we can first release this feature in a discrete way and then document it fully when ready.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
```diff
@@ -101,9 +100,13 @@ export function GridFilterPanel() {
       </GridPanelContent>
       {!rootProps.disableMultipleColumnsFiltering && (
         <GridPanelFooter>
-          <Button onClick={addNewFilter} startIcon={<GridAddIcon />} color="primary">
+          <rootProps.components.BaseButton
+            onClick={addNewFilter}
+            startIcon={<GridAddIcon />}
+            color="primary"
```
m4theushw, line 106:
```suggestion
            color="primary"
            {...rootProps.componentsProps?.baseButton}
```

x other slots.

m4theushw (commented):
Should `Popper` also be added? Adding it seems to solve part of the problem raised in #3381.

> It is possible to override Portal props for Column Header tooltips by customizing the theme's default props for Tooltip, but since there is no ability to customize default props for Popper/Portal to propogate the shadow dom container, all of the styled(Popper) components will appear outside and have no styling applied.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
DanailH: Added

DanailH:
> Should `Popper` also be added? Adding it seems to solve part of the problem raised in #3381.
> 
> > It is possible to override Portal props for Column Header tooltips by customizing the theme's default props for Tooltip, but since there is no ability to customize default props for Popper/Portal to propogate the shadow dom container, all of the styled(Popper) components will appear outside and have no styling applied.

I added `componentsProps` for `Popper` since we never use it directly. We use it as a base for `GridMenu`. In addition, I added the `Tooltip` to slot components.

packages/grid/_modules_/grid/models/gridSlotsComponent.ts
```diff
@@ -6,15 +6,45 @@ import { GridIconSlotsComponent } from './gridIconSlotsComponent';
...
+   * @default Button
+   */
+  BaseButton: React.JSXElementConstructor<any>;
+  /**
+   * The custom Tooltip component used in the grid.
+   * @default Tooltip
+   */
+  BaseTooltip: React.JSXElementConstructor<any>;
+  /**
+   * Component rendered for each cell.
```
m4theushw, line 45:
```suggestion
   * Component rendered for each cell.
   * @default GridCell
```

packages/grid/_modules_/grid/components/menu/GridMenu.tsx
```diff
@@ -98,6 +98,7 @@ const GridMenu = (props: GridMenuProps) => {
       transition
```
m4theushw, line 98: Could you add the slot for `Popper`? To keep each `components` key with a matching `componentsProps` key. I guess you need to pass the component using the `as` prop.

m4theushw (commented):
It's missing to spread `rootProps.componentsProps?.basePopper` in the `GridPanel`.

https://github.com/mui-org/material-ui-x/blob/7bac90d6dbaef58457fda74efa224d7d97dd8f4e/packages/grid/_modules_/grid/components/panel/GridPanel.tsx#L30

packages/grid/_modules_/grid/components/menu/GridMenu.tsx
DanailH: Done

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
```diff
@@ -23,11 +23,12 @@ export interface GridPanelProps extends StandardProps<PopperProps, 'children'> {
    */
   classes?: Partial<GridPanelClasses>;
   open: boolean;
+  as: React.ElementType;
 }
 
 export const gridPanelClasses = generateUtilityClasses('MuiDataGrid', ['panel', 'paper']);
 
-const GridPanelRoot = styled(Popper, {
+const GridPanelRoot = styled('div', {
```
m4theushw, line 31: Although I suggested it, I don't know if this is the right approach. Looking at the change done in https://github.com/mui-org/material-ui-x/pull/3490/files#diff-c2fd479a822e3ad62e4338f61294f31a114e307bcbcce46c482822ef9d8fc784 maybe the default tag should be `Popper` and keep the `as` prop too. 

Looking closer I found a similar case in the core:

https://github.com/mui-org/material-ui/blob/bc016a59f8b4afeb43c98f677c9fbd1e8afb2698/packages/mui-material/src/Autocomplete/Autocomplete.js#L408

https://github.com/mui-org/material-ui/blob/bc016a59f8b4afeb43c98f677c9fbd1e8afb2698/packages/mui-material/src/Autocomplete/Autocomplete.js#L581-L582

The default value of `PopperComponent` is `Popper`.

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
```diff
@@ -23,11 +23,12 @@ export interface GridPanelProps extends StandardProps<PopperProps, 'children'> {
    */
   classes?: Partial<GridPanelClasses>;
   open: boolean;
+  as: React.ElementType;
```
m4theushw, line 26: You can extend from `MUIStyledCommonProps` as below:

```diff
diff --git a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
index 510e869a2..68f20b1a7 100644
--- a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridPanel.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
-import { styled } from '@mui/material/styles';
+import { styled, Theme } from '@mui/material/styles';
 import { generateUtilityClasses, InternalStandardProps as StandardProps } from '@mui/material';
+import { MUIStyledCommonProps } from '@mui/system';
 import ClickAwayListener from '@mui/material/ClickAwayListener';
 import Paper from '@mui/material/Paper';
 import { PopperProps } from '@mui/material/Popper';
@@ -16,14 +17,14 @@ export interface GridPanelClasses {
   paper: string;
 }
 
-export interface GridPanelProps extends StandardProps<PopperProps, 'children'> {
+export interface GridPanelProps
+  extends StandardProps<MUIStyledCommonProps<Theme> & PopperProps, 'children'> {
   children?: React.ReactNode;
   /**
    * Override or extend the styles applied to the component.
    */
   classes?: Partial<GridPanelClasses>;
   open: boolean;
-  as: React.ElementType;
 }
 
 export const gridPanelClasses = generateUtilityClasses('MuiDataGrid', ['panel', 'paper']);
```

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
DanailH: Updated

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
```diff
@@ -76,7 +78,6 @@ const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>((props, ref)
       ref={ref}
       placement="bottom-start"
       className={clsx(className, classes.panel)}
```
m4theushw, line 80: The `as` prop is missing as well as `rootProps.componentsProps.panel`.

packages/grid/_modules_/grid/components/panel/GridPreferencesPanel.tsx
```diff
@@ -22,9 +22,11 @@ export const GridPreferencesPanel = React.forwardRef<
   return (
     <rootProps.components.Panel
       ref={ref}
+      as={rootProps.components.BasePopper}
       open={columns.length > 0 && preferencePanelState.open}
       {...rootProps.componentsProps?.panel}
```
m4theushw, line 27: If the `as` prop is applied inside `GridPanel`, then it doesn't need to be applied here again.

packages/grid/_modules_/grid/components/panel/GridPanel.tsx
DanailH: I can't add them in the `GridPanel` directly because the `useGridRootProps` can be used in components outside of `DataGrid`. That's why I had to pass it down from the parent. The same applies to the other comment https://github.com/mui-org/material-ui-x/pull/3490#discussion_r779507845

```
Error: Uncaught Error: MUI: useGridRootProps should only be used inside the DataGrid/DataGridPro component.

m4theushw: We had a discussion about that in #2522.

Currently the `GridPanel` already depends on another context: `GridApiContext`. We could also use `useGridRootProps` but that would be a breaking change. So I'm OK to not do that for now.

DanailH: Ok so if we are postponing this and if this is the only thing left can we proceed with merging this PR?

m4theushw (approved):

PR #3492, opened 2021-12-22
[DataGrid] Close other actions menus when opening a new one

### LINKED ISSUE #3489: [DataGrid] Previous "menuList" hang on screen every click for "type:actions" cells
### Duplicates

### Latest version

### Current behavior 😯

Previous "menuList" hang on screen after every click for "type:actions" cells. Only if you click anywhere on screen, it would be hidden.
https://mui.com/components/data-grid/columns/#special-properties
bug
bug2

### Expected behavior 🤔

Previous "menuList" should be hidden after every click for "type:actions" cells.Only new "menuList" should show.

### Steps to reproduce 🕹

Steps:

1.Click action button in grid
2.Then,click another action button in grid
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3489

Preview: https://deploy-preview-3492--material-ui-x.netlify.app/components/data-grid/columns/#special-properties

If we continue to add more column types with interactive elements, then it makes sense to add a per-column `disableSelectionOnClick`.

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -19,13 +19,12 @@ import {
 import { gridPaginatedVisibleSortedGridRowIdsSelector } from '../pagination';
 import { gridVisibleSortedRowIdsSelector } from '../filter/gridFilterSelector';
 import { GRID_CHECKBOX_SELECTION_COL_DEF, GridColDef } from '../../../models';
-import { getDataGridUtilityClass } from '../../../gridClasses';
+import { getDataGridUtilityClass, gridClasses } from '../../../gridClasses';
 import { useGridStateInit } from '../../utils/useGridStateInit';
 import { GridPreProcessingGroup, useGridRegisterPreProcessor } from '../../core/preProcessing';
 import { GridCellModes } from '../../../models/gridEditRowModel';
 import { GridColumnsRawState } from '../columns/gridColumnsState';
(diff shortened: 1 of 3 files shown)


packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -25,6 +25,7 @@ import { GridPreProcessingGroup, useGridRegisterPreProcessor } from '../../core/
 import { GridCellModes } from '../../../models/gridEditRowModel';
 import { GridColumnsRawState } from '../columns/gridColumnsState';
 import { isKeyboardEvent } from '../../../utils/keyboardUtils';
+import { gridClasses } from '../../..';
```
flaviendelangle, line 28: We should probably not import it from the parent index file to avoid risk of circular imports

m4theushw: VSCode import suggestions never work correctly.

flaviendelangle: Neither do Webstorm one :/

flaviendelangle:
> If we continue to add more column types with interactive elements, then it makes sense to add a per-column `disableSelectionOnClick`.

I agree that putting logic for a specific type in `useGridSelection` is not great in the longer run.
But good enough for a 1st version.

flaviendelangle (approved):

PR #3494, opened 2021-12-22
[l10n] Improve French (frFR) locale

Zenoo:
The `Print` option of the `DataGrid` wasn't translated.  
This PR should fix this.


alexfauquette:
Thanks for your contribution 😊

Issue #3448, opened 2021-12-10
[DataGrid] Add other sub-states to the `initialState` prop 


cq-ubaid-khan:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Currently the datagrid only support few properties as initial state one example `pinnedColumns`, We want it to receive `columnsOrders`, `columnsSizes`, `pagination`, `columnVisibility` as an initial state of grid.

Support for this configuration will be useful where we save all of the datagrid settings in persistent location and later when visited on the same page we get the same settings applied on the grid automatically.

Currently in order to achieve this behavior we have to interact with the `apiRef`.

Thanks.

### Examples 🌈

```
const savedInitialState = {
   columnsSizes: { fieldName: { width: 30 }, otherField: { width: 50 } },
   columnsOrders: { fieldName: { orderIndex: 4 } },
   pagination: { size: 20, ... },
   columnsVisibility: { fieldName: { visible: false } },
   ...
}

<DataGrid 
   initialState={savedInitialState}
   ...
/>

```

alexfauquette:
It sounds like it is a duplicate of #820

flaviendelangle:
#820 is mainly about the new methods to imperatively store and restore the state.
But we can keep this one to discuss which state keys should be added to the `initialState` (most of the time I think those should also be added to the store / restore methods).

@cq-ubaid-khan as you saw we started small with some sub-states we knew were needed. The idea being to avoid adding to much complexity right from the start.

The `pagination` (`pageSize` and `page`) could very easily be added and for me it totally make sense.
For the columns related data it is slightly more complicated and I think we should do it in a second run. You can pre-process your column objects to provide custom initial order / sizes / visibility but we will have to think of a clean implementation on our side.

cq-ubaid-khan:
Hmm. Thanks for you response, I understand your thoughts about the complexity as I've gone through the same kind problem in the past.

However, I've implemented the solution from my own way by implementing a middle state mapping object which I can easily persist and reuse.

Pasting here so anyone in the future can get some help from this.

 ```
const layout = {
  layoutSettings: {
    field1: {
        orderIndex: 1,
        hidden: true,
        locked: false,
        ....
     }
    field2: ....,
    ....
  }
}

const layoutSettingFields = Object.keys(layout.layoutSettings);
if (layoutSettingFields.length) {
  layoutSettingFields.forEach((field) => {
[...]

PR #3495, opened 2021-12-22
[DataGrid] Add the `page` and `pageSize` to the initialState prop

flaviendelangle:
Part of #3448

I had to add a sync version of `useDemoData` to add a demo of the initial page.

## Changelog

- 🕹️ Add the `page` and `pageSize` to the `initialState` prop.
  
  You can now initialize the page and / or the page size without controlling them

  ```tsx
  <DataGrid
    initialState={{
      pagination: {
        page: 1,
        pageSize: 25,
      },
    }}
  />
  ```


alexfauquette (commented):
I'm wondering, how does it works when the initialState is set pagination, but there is no row, because for example, data are loading.  

Will it set back `page=0` because page 2 does not exist?

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
```diff
@@ -20,14 +20,22 @@ import { gridDensityRowHeightSelector } from '../density';
...
+  >,
 ) => {
   const logger = useGridLogger(apiRef, 'useGridPageSize');
   const rowHeight = useGridSelector(apiRef, gridDensityRowHeightSelector);
 
   useGridStateInit(apiRef, (state) => ({
     ...state,
-    pagination: { pageSize: props.pageSize ?? (props.autoPageSize ? 0 : 100) },
+    pagination: {
+      pageSize:
```
DanailH, line 34: Can we move complex ternary operators like this one outside into a variable? It's easier to read.

flaviendelangle:
> I'm wondering, how does it works when the initialState is set pagination, but there is no row, because for example, data are loading.
>
> Will it set back page=0 because page 2 does not exist?

Yes, initializing the page only make sense when the data loading is synchronous indeed.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
flaviendelangle: Done :+1:

packages/grid/_modules_/grid/hooks/features/pagination/useGridPageSize.ts
```diff
@@ -19,15 +19,33 @@ import { gridDensityRowHeightSelector } from '../density';
...
+    } else if (props.autoPageSize) {
+      pageSize = 0;
+    } else {
+      pageSize = 100;
+    }
+
+    return {
+      ...state,
+      pagination: {
+        pageSize,
```
m4theushw, line 45: Should `state.pagination` be spread first? It doesn't cause a bug now because `useGridPage` comes after, but if it comes before this hook for some reason, then the initial page is lost.

flaviendelangle: If the hook order is inverted, we would have other issues because `useGridPage` needs the page size to compute the `pageCount`.

docs/src/pages/components/data-grid/pagination/PageInitialState.js
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { DataGrid } from '@mui/x-data-grid';
+import { getSynchronousDemoData } from '@mui/x-data-grid-generator';
+
+const data = getSynchronousDemoData({
```
m4theushw, line 5: This function blocks the main thread. It can't be used to generate more than a couple thousands rows. It maybe only used in this very specific demo, which makes it overkill to add to `@mui/x-data-grid-generator`. You could use `useDemoData` and delay the rendering of the DataGrid once it has data: 

```tsx
const { data, loading } = useDemoData();

if (loading) {
  return null;
}
```

Or load the data from a JSON file. Check the `packages/grid/x-data-grid-generator/src/datagen-cli.ts` script (you need to fix some paths to make it  run).

m4theushw (approved):

PR #3497, opened 2021-12-22
[DataGrid] Customize FilterPanel with props

### LINKED ISSUE #3082: Customizability of DataGrid's FilterPanel
### Order ID 💳

21172

### Duplicates

### Latest version

### Summary 💡

I've implemented a custom FilterPanel, however there is some further customization that I'm not sure is possible (or if it is, how to do it) based on the documentation

### The problem in depth 🔍

Specific things I'd like to be able to customize:
- Order of columns in the columns dropdown (for usability reasons I have a certain column order, however when searching for a column to filter on it'd be easier to be able to eg set it alphabetically)
- Width of operators text field (I have customized various InputComponents for the values field but for some filters with longer text it'd be nice to be able to display it all)
- Overall just documentation of what's customizable / how would be useful

### Context 🔦

Continuing to make use of MUI DataGrid, really like it – this would help us further tune it to our needs

### Your environment 🌎

"@mui/icons-material": "^5.0.4",
"@mui/lab": "^5.0.0-alpha.51",
"@mui/material": "^5.0.4",
"@mui/styles": "^5.0.1",
"@mui/x-data-grid-generator": "^5.0.0-beta.4",
"@mui/x-data-grid-pro": "^5.0.0-beta.4",

### LINKED ISSUE #3386: [DataGridPro] Hide `OR` operator option from the `linkOperators` in Multi-column filtering
### Order ID 💳

25719

### Duplicates

### Latest version

### Summary 💡

I am using Multi-column filtering  on data grid pro but I would like to hide the `OR` option for the time being. Is there a way to do this except with styling ?

### The problem in depth 🔍

Hide GridLinkOperator.Or option from the link operators. 

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
  System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 858.14 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 15.1.0 - /usr/local/bin/node
    npm: 7.0.8 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.0 - /usr/local/bin/pod
    Gradle: 6.2.2 - /usr/local/bin/gradle
    Homebrew: 3.2.11 - /usr/local/bin/brew
    pip3: 20.3.1 - /usr/local/bin/pip3
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 4.2.1 - /usr/bin/gcc
    Git: 2.21.1 - /usr/bin/git
    Clang: 11.0.0 - /usr/bin/clang
    Subversion: 1.10.4 - /usr/bin/svn
  Servers:
    Apache: 2.4.48 - /usr/sbin/apachectl
    Nginx: 1.19.5 - /usr/local/bin/nginx
  Virtualization:
    Docker: 20.10.10 - /usr/local/bin/docker
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.3
      System Images: android-26 | Google APIs Intel x86 Atom, android-30 | Google APIs Intel x86 Atom, android-R | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.6953283
    Atom: 1.58.0
    Nano: 2.0.6 - /usr/bin/nano
    Vim: 8.1 - /usr/bin/vim
    Xcode: 11.3.1/11C504 - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Java: 1.8.0_242 - /usr/bin/javac
    Perl: 5.18.4 - /usr/bin/perl
    PHP: 7.3.29 - /usr/bin/php
    Python: 2.7.16 - /usr/bin/python
    Python3: 3.9.0 - /usr/local/bin/python3
    Ruby: 2.6.3 - /usr/bin/ruby
  Databases:
    SQLite: 3.28.0 - /Users/beslinda/Library/Android/sdk/platform-tools/sqlite3
  Browsers:
    Chrome: 96.0.4664.93
    Firefox: 74.0
    Safari: 15.1
  Monorepos:
    Lerna: 4.0.0

```
</details>

alexfauquette:
Fix #3082 and fix #3386


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -77,6 +130,8 @@ function GridFilterForm(props: GridFilterFormProps) {
   const valueRef = React.useRef<any>(null);
   const filterSelectorRef = React.useRef<HTMLInputElement>(null);
 
+  const sortedFilterableColumns = useColumnSorting(filterableColumns, columnsSort);
```
m4theushw, line 133: `React.useMemo` could be used instead of a custom hook.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -47,21 +55,66 @@ const GridFilterFormRoot = styled('div', {
...
+const getColumnLabel = (col) => col.headerName || col.field;
+
+const useColumnSorting = (filterableColumns, columnsSort) => {
+  const [sortedColumns, setSortedColumns] = React.useState(filterableColumns);
+
+  React.useEffect(() => {
+    switch (columnsSort) {
+      case 'asc':
+        setSortedColumns(
+          filterableColumns.sort((a, b) => (getColumnLabel(a) < getColumnLabel(b) ? -1 : 1)),
```
m4theushw, line 81: Comparing `'ã' < 'b'` is false. Prefer the [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) API which correctly sorts accented words.

https://github.com/mui-org/material-ui-x/blob/99d0d24cf638defc2e76933892a9f15d40fcf6e8/packages/grid/_modules_/grid/utils/sortingUtils.ts#L44

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -175,7 +230,12 @@ function GridFilterForm(props: GridFilterFormProps) {
     <GridFilterFormRoot className={classes.root}>
       <FormControl
         variant="standard"
-        sx={{ flexShrink: 0, justifyContent: 'flex-end', marginRight: 0.5, marginBottom: 0.2 }}
+        sx={[
+          { flexShrink: 0, justifyContent: 'flex-end', marginRight: 0.5, marginBottom: 0.2 },
+          ...(Array.isArray(deleteIconContainerSx)
+            ? deleteIconContainerSx
+            : [deleteIconContainerSx]),
+        ]}
```
m4theushw, line 238: As we discussed, the prop could be renamed to `deleteIconContainerProps` and allow to override other props of the `FormControl`.

```suggestion
        {...deleteIconContainerProps}
        sx={{
          flexShrink: 0, 
          justifyContent: 'flex-end',
          marginRight: 0.5,
          marginBottom: 0.2,
          ...deleteIconContainerProps.sx
        }}
```

Another alternative is to make the `FormControl` a styled component, then you don't need to care abound spreading `deleteIconContainerProps.sx`, since it's done automatically.

```tsx
const StyledFormControl = styled(FormControl)({ flexShrink: 0 });

<StyledFormControl {...deleteIconContainerProps} />
```

docs/src/pages/components/data-grid/filtering/CustomFilterPanel.js
```diff
@@ -0,0 +1,33 @@
...
+    dataSet: 'Commodity',
+    rowLength: 10,
+    maxColumns: 6,
+  });
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGridPro
+        {...data}
+        // components={{
```
DanailH, line 16: Remove commented-out code (or maybe it's still a WIP).

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -27,6 +27,14 @@ export interface GridFilterFormProps {
   applyFilterChanges: (item: GridFilterItem) => void;
   applyMultiFilterOperatorChanges: (operator: GridLinkOperator) => void;
   deleteFilter: (item: GridFilterItem) => void;
+  hasLinkOperatorColumn?: boolean;
+  linkOperators?: (GridLinkOperator.And | GridLinkOperator.Or)[];
+  columnsSort?: 'asc' | 'desc';
+  deleteIconContainerSx?: SxProps<Theme>;
```
flaviendelangle, line 33: @mnajdova in the core how to you handle when you want the user to easily customize the style of nested elements with `sx` ?

Do you pass down the `sx` function like this or do you expose the classes and let the user nest inside the root sx ?

docs/src/pages/components/data-grid/filtering/CustomFilterPanel.js
alexfauquette: In fact, I was looking for a way to show all the customization options in the same code, and comment on those which are not used in the demo. But yes, that's not explicit enough.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
mnajdova: I would recommend using `components` and `componentsProps` for this. Then developers can do:

```
<GridFilterForm
  componentsProps={{ deleteIconContainer: { sx: /* ... */ } }}
  ...
/>
```

At least this is what we do in the `@mui/base`.

alexfauquette: Do you mean using `components` props to let dev use a custom filter like that?

```js
const CustomFilterForm = (props) => <GridFilterForm
	componentsProps={{ deleteIconContainer: { sx: /* ... */ } }}
	{...props}
/>

<DataGrid
	...
	components={{
		filterForm: CustomFilterForm
	}}
```

For now, I tried to use nesting classes as follows. Considering that `<FilterPanel />` is the only component user can have access. The other components are considered as nested ones, and can only be customized by CSS selectors.

```js
<DataGrid
	...
	componentsProps={{
		filterPanel: {
			sx: {
				"& .MuiDataGrid-filterForm": { ... },
				"& .MuiDataGrid-closeIconController": { ... },
				...
			}
		}
	}}
```

docs/src/pages/components/data-grid/filtering/filtering.md
```diff
@@ -230,9 +230,26 @@ const ratingColumnType: GridColTypeDef = {
...
 You can customize the rendering of the filter panel as shown in [the component section](/components/data-grid/components/#overriding-components) of the documentation.
 
+### Customize the filter panel content
+
+The customization of the filter panel content, can be performed by passing props to the he default `<GridFilterPanel/>` component.
+These props allows you to override
+
+- The available `linkOperators`
+- The order of the column selector
+- The style of the controller for each input
```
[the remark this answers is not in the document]
alexfauquette: So the way to modify input components would be the following

```js
<DataGrid 
	componentsProps={{
		filterFormDeleteIcon: { sx: /* ... */ },
		filterFormLinkOperatorInput: { sx: /* ... */, disabled: true, hiddenLabel: true },
		filterFormOperatorInput: { sx: /* ... */ },
		filterFormColumnInput: { sx: /* ... */ },
		filterFormValueInput: { sx: /* ... */ },
	}}
/>
```

If  yes, should we keep the className?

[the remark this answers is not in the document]
m4theushw: I was expecting that we would support the following:

```tsx
<DataGrid 
  componentsProps={{
    filterPanel: {
      FilterFormProps: {
        deleteIconProps: { sx: /* ... */ },
        linkOperatorInputProps: { sx: /* ... */, disabled: true, hiddenLabel: true },
        operatorInputProps: { sx: /* ... */ },
        columnInputProps: { sx: /* ... */ },
        valueInputProps: { sx: /* ... */ },
      }
    }
  }}
/>
```

> If yes, should we keep the className?

I didn't understand. Which className?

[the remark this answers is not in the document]
alexfauquette: I'm talking about the CSS classes I introduced in this PR. The ones allowing to customize the filter panel as follow:

```js
<DataGrid
	...
	componentsProps={{
		filterPanel: {
			sx: {
				"& .MuiDataGrid-filterForm": { ... },
				"& .MuiDataGrid-closeIconController": { ... },
				...
			}
		}
	}}
/>
```

It is an alternative way to customize the filter panel so I can :
- revert the modifications, and we keep only one way to customize the filter panel inputs
- I rename classes such that they are similar to `FilterFormProps` keys. For example `.MuiDataGrid-closeIconController` would become `.MuiDataGrid-deleteIcon`.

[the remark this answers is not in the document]
m4theushw: I would keep the class names, but scoping them to avoid too generic names, e.g. `.MuiDataGrid-deleteIcon`. We might have another `.MuiDataGrid-deleteIcon` in the future. As suggestion, all inner components of the virtualization have the `virtualScroller` prefix in the class name: https://github.com/mui-org/material-ui-x/blob/24cf3fbbe7907e8264f6e2fde73a82919a5c5dd0/packages/grid/_modules_/grid/components/virtualization/GridVirtualScrollerRenderZone.tsx#L15

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+          columnField: 'rating',
+          operatorValue: '>',
+          value: '0',
+        },
+      ],
+    },
+  },
+};
+
+export default function CustomFilterPanel() {
```
m4theushw, line 34:
```suggestion
export default function CustomFilterPanelContent() {
```

docs/src/pages/components/data-grid/filtering/filtering.md
```diff
@@ -230,9 +230,32 @@ const ratingColumnType: GridColTypeDef = {
...
+### Customize the filter panel content
+
+The customization of the filter panel content can be performed by passing props to the default `<GridFilterPanel />` component.
+The props available allow to override:
+
+- The available `linkOperators` (can contains `GridLinkOperator.And` and `GridLinkOperator.Or`)
+- The order of the column selector (can be `"asc"` or `"desc"`)
+- Any prop of the input components
+
+Input components can be [customized](/customization/how-to-customize/) by using two approaches. You can pass a `sx` prop to any input container or you can use css selectors on nested components of the filter panel. More details are available in the demo.
```
m4theushw, line 242: It's easier to make suggestions if each phrase is in a different line.

```suggestion
Input components can be [customized](/customization/how-to-customize/) by using two approaches.
You can pass a `sx` prop to any input container or you can use css selectors on nested components of the filter panel.
More details are available in the demo.
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -34,7 +42,12 @@ const useUtilityClasses = (ownerState: OwnerState) => {
   const { classes } = ownerState;
 
   const slots = {
-    root: ['filterForm'],
+    root: ['filterForm', 'filterPanelFilterForm'],
```
m4theushw, line 45: Why two classes on the root slot?

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -34,7 +42,12 @@ const useUtilityClasses = (ownerState: OwnerState) => {
...
+    closeIcon: ['filterPanelDeleteIcon'],
```
m4theushw, line 46: The owner of this slot is `GridFilterForm`. Also, some classes are not matching their slot names.

```suggestion
    deleteIcon: ['filterFormDeleteIcon'],
```

x other slots.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -46,10 +59,24 @@ const GridFilterFormRoot = styled('div', {
   overridesResolver: (props, styles) => styles.filterForm,
```
m4theushw, line 59: #3716 made me remember that if we add new classes we need to update the `overridesResolver`.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -174,8 +229,16 @@ function GridFilterForm(props: GridFilterFormProps) {
...
+        {...baseFormControlProps}
+        {...deleteIconProps}
+        sx={{
+          flexShrink: 0,
+          justifyContent: 'flex-end',
+          marginRight: 0.5,
+          marginBottom: 0.2,
+          ...(deleteIconProps?.sx || {}),
+        }}
+        className={clsx(classes.closeIcon, baseFormControlProps.className)}
```
m4theushw, line 241: A custom class name will be overwritten if not applied individually.

```suggestion
        className={clsx(classes.closeIcon, baseFormControlProps.className, deleteIconProps.className)}
```

x other components.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: Because renaming `'filterForm'` would be a breaking change, so to be consistent with the `'filterPannel'` prefix, I add `'filterPanelFilterForm'` and wait v6 to remove the `'filterForm'`

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: I though it make more sense to refer to the FilterPanel as the owner, because if in the future we add other filtering interfaces, the name filterForm will not be enough to distinguish them.

But I assume its a good thing to be able to customize all the `'filterFormDeleteIcon'` at the same time, so yes let the naming explain how the codes work :)

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: Now that the prefix is "filterForm" it is not a problem anymore :)

packages/grid/_modules_/grid/components/panel/GridPanelWrapper.tsx
```diff
@@ -34,9 +35,12 @@ const GridPanelWrapperRoot = styled('div', {
 
 const isEnabled = () => true;
 
-export function GridPanelWrapper(
-  props: React.PropsWithChildren<React.HTMLAttributes<HTMLDivElement>>,
-) {
+interface GridPanelWrapperProps
+  extends React.PropsWithChildren<React.HTMLAttributes<HTMLDivElement>> {
```
m4theushw, line 39: Maybe reuse `MUIStyledCommonProps` which already includes `sx`.

https://github.com/mui-org/material-ui-x/blob/a7b2ffb7e2f0a41f865ceefc62cd95976f3e6f6a/packages/grid/_modules_/grid/components/panel/GridPanel.tsx#L20-L21

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -169,12 +225,77 @@ function GridFilterForm(props: GridFilterFormProps) {
     [currentOperator],
   );
 
+  const FilterFormDeleteIcon = React.useMemo(
```
m4theushw, line 228: Why to memoize? The styled component won't react to a change in the theme.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -169,12 +225,77 @@ function GridFilterForm(props: GridFilterFormProps) {
...
   return (
     <GridFilterFormRoot className={classes.root}>
-      <rootProps.components.BaseFormControl
+      <FilterFormDeleteIcon
         variant="standard"
-        sx={{ flexShrink: 0, justifyContent: 'flex-end', marginRight: 0.5, marginBottom: 0.2 }}
-        {...rootProps.componentsProps?.baseFormControl}
+        {...baseFormControlProps}
+        {...deleteIconProps}
+        sx={deleteIconProps?.sx || {}}
```
m4theushw, line 293: Why are you setting `sx` here? In the line above you also spread all props, including `sx`.

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+              },
+            },
+            sx: {
+              // Customize inputs using css selectors
+              '& .MuiDataGrid-filterForm': { p: 2 },
+              '& .MuiDataGrid-filterForm:nth-child(even)': {
+                backgroundColor: (theme) =>
+                  theme.palette.mode === 'dark' ? '#444' : '#ddd',
+              },
+              '& .MuiDataGrid-filterFormDeleteIcon': { display: 'none' },
```
m4theushw, line 79: There's no way to delete a filter anymore.

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+              '& .MuiDataGrid-filterFormLinkOperatorInput': { mr: 2 },
+              '& .MuiDataGrid-filterFormColumnInput': { mr: 2, width: 200 },
+              '& .MuiDataGrid-filterFormOperatorInput': { mr: 5 },
```
m4theushw, line 82: To keep the margin symmetrical.

```suggestion
              '& .MuiDataGrid-filterFormOperatorInput': { mr: 2 },
```

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+              valueInputProps: {
+                required: true,
+              },
+            },
+            sx: {
+              // Customize inputs using css selectors
+              '& .MuiDataGrid-filterForm': { p: 2 },
+              '& .MuiDataGrid-filterForm:nth-child(even)': {
+                backgroundColor: (theme) =>
+                  theme.palette.mode === 'dark' ? '#444' : '#ddd',
```
m4theushw, line 77: The color for the dark theme is good, but the color for the light theme is too dark.

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+        componentsProps={{
+          filterPanel: {
+            // Force to use "And" operator
+            linkOperators: [GridLinkOperator.And],
+            // Display columns by ascending alphabetical order
+            columnsSort: 'asc',
+            filterFormProps: {
+              // Customize inputs by passing props
+              columnInputProps: {
+                variant: 'outlined',
```
m4theushw, line 59: There's something wrong using this variant. The input border should be passing behind the label. See https://mui.com/components/text-fields/#select

image

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
mnajdova: Sorry, for some reason I have missed this comment. If the other nested components can only be customized by CSS selectors, why are you adding this new prop `deleteIconContainerSx`. It contradicts with the previous statement.

My point was that, we don't add first class props on the component for prop that should be passed on some nested elements, we prefer using `componentsProps` for this use-case.

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: Because I have to define them inside the FilterForm components to access `rootProps.components.BaseFormControl`. And I anted to avoid recomputing styled components at every render.

alexfauquette: A solution to avoid `useMemo`, and not define `styled` components inside the GrisFilterForm would be to create these components outside as follow:

```js
const FilterFormColumnInput = ({children, ...props}) => {
  const Container = styled(rootProps.components.BaseFormControl, { ... })( ... );
  
  return <Container {...props}>{children}</Container>
}
``` 

By the way, `useMemo` does not seems to break the theme update. I modified one of the styles components to make its background color change according to the theme and it works fine. (I agree, colors are ugly :p ). 

```js
const FilterFormColumnInput = React.useMemo(
    () =>
      styled(rootProps.components.BaseFormControl, {
        name: 'MuiDataGrid',
        slot: 'FilterFormColumnInput',
        overridesResolver: (_, styles) => styles.filterFormColumnInput,
      })(({ theme }) => ({ 
		width: 150,
		backgroundColor: theme.palette.mode === 'dark' ? "red" : "green" })
	),
    [rootProps.components.BaseFormControl],
  );
```

Peek 2022-01-27 11-06

m4theushw: > Because I have to define them inside the FilterForm components to access rootProps.components.BaseFormControl

You can create the styled component normally, then use the `as` prop to pass the component from the slot:

https://github.com/mui-org/material-ui-x/blob/70a372129854b609a1b352167e40ac01c45fdd6c/packages/grid/_modules_/grid/components/menu/GridMenu.tsx#L94-L95

> By the way, useMemo does not seems to break the theme update.

Yeah, it works but this means that it's not really memoizing the result. It's only memoizing the styled component, not the instance. The costly part of emotion - the serialization of the styles - is still being done in each render. In this case, I would avoid making premature optimizations. If performance was the problem a better solution would be https://github.com/emotion-js/emotion/issues/2529 or export `css` from `@mui/material/styles`.

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
alexfauquette: Ok, I will pick the one used in `<Table />` demo

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
alexfauquette: That's because of the `sx: { justifyContent: 'flex-end' }` applied to the filter form which impact the select, but not the label
I have to add the `label` prop to `<Select />` otherwise the outline is not removed under the label

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
alexfauquette: Yes, not the best demo example. I tried a new one with `order: 1` to move the button to the end

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: Yes, it's useless

I only let one `sx` for link operators input, because the `hasLinkOperatorColumn` and `showMultiFilterOperators` are easier to read in the render than in the styled function
```jsx
sx={{
	  display: hasLinkOperatorColumn ? 'block' : 'none',
	  visibility: showMultiFilterOperators ? 'visible' : 'hidden',
	  ...(baseFormControlProps.sx || {}),
	  ...(linkOperatorInputProps.sx || {}),
}}
```

tielushko:
I was wondering whether something similar could be done with the columns panel? I would personally like to hide the hide all/show all buttons and have not found a way to override the component while keeping the old toggle switches.

alexfauquette:
@tielushko I think, what you want to do is feasible, because "show/hide all" buttons are in the footer of the column panel

So you should be able to override their style with the following selectors 

Rule name | Global class | Description
-- | -- | --
columnsPanel | .MuiDataGrid-columnsPanel | Styles applied to the columns panel element.
panelFooter | .MuiDataGrid-panelFooter | Styles applied to the panel footer element.

extracted from this page of the doc
https://mui.com/api/data-grid/data-grid/#css

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -263,11 +383,17 @@ function GridFilterForm(props: GridFilterFormProps) {
...
+      </FilterFormOperatorInput>
+      <FilterFormValueInput
         variant="standard"
-        sx={{ width: 190 }}
-        {...rootProps.componentsProps?.baseFormControl}
+        as={rootProps.components.BaseFormControl}
+        {...baseFormControlProps}
+        {...valueInputProps}
+        className={clsx(
+          classes.valueInput,
```
DanailH, line 393: Maybe this has been already answered but what about when the `FilterFormValueInput` is the select or the multi-select, will it always has that `classes.valueInput` class or should we change the class based on the component?

alexfauquette: Not already answered. Modifying the class according to the input seems tricky. 
We should add a property `currentOperator.valueFormControlClass`

In my opinion, it is overkilled compared to what most of the dev will probably do with the customization:
- removing some option
- modifying input size

About the size, `classes.valueInput` is useful to remove default one. To have customer size depending on the input component, I see two cases:
- it is their own input, so they can do whatever they want
- it is one of the provided input. They can distinguish single and multiple values with CSS controller, depending on `MuiFormControl-root` and `MuiInput-root`

m4theushw: I'm not sure if I'm on the same page here but `classes.valueInput` is applied to the `FormControl` that wraps the custom filter input component. In this context, `FilterFormValueInput` will always be a `FormControl`, unless the user changes it through the slots.

alexfauquette: Yes, we are on the same page

For now, the `FormControl` is responsible for the width. So to customize the `classes.valueInput` must be used to add `width:'unset'`

docs/src/pages/components/data-grid/filtering/CustomFilterPanelContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+                sx: { mt: 'auto' },
+              },
+              operatorInputProps: {
+                variant: 'outlined',
+                size: 'small',
+                sx: { mt: 'auto' },
+              },
+              valueInputProps: {
+                required: true,
+              },
```
joserodolfofreitas, line 70: Not sure this is the best way to customize the deleteIcon color.
But I think we could add a simple example for deleteIconProps too.

```suggestion
              deleteIconProps: {
                sx: {
                  "& .MuiSvgIcon-root": { color: "red" }
                }
              },
```

alexfauquette: Yes, it's better than my reorder which was a bad accessibility practice

siriwatknp:

3. run `yarn docs:api`

DanailH (approved):

docs/data/data-grid/filtering/filtering.md
```diff
@@ -259,9 +260,34 @@ const ratingColumnType: GridColTypeDef = {
 
 You can customize the rendering of the filter panel as shown in [the component section](/components/data-grid/components/#overriding-components) of the documentation.
 
+### Customize the filter panel content
+
+The customization of the filter panel content can be performed by passing props to the default `<GridFilterPanel />` component.
+The available props allow to override:
+
+- The available `linkOperators` (can contains `GridLinkOperator.And` and `GridLinkOperator.Or`)
```
m4theushw, line 268: There's already "available" in the previous line.

```suggestion
- The `linkOperators` (can contains `GridLinkOperator.And` and `GridLinkOperator.Or`)
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -259,9 +260,34 @@ const ratingColumnType: GridColTypeDef = {
...
+- The order of the column selector (can be `"asc"` or `"desc"`)
+- Any prop of the input components
+
+Input components can be [customized](/customization/how-to-customize/) by using two approaches.
+You can pass a `sx` prop to any input container or you can use css selectors on nested components of the filter panel.
```
m4theushw, line 273:
```suggestion
You can pass a `sx` prop to any input container or you can use CSS selectors on nested components of the filter panel.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -259,9 +260,34 @@ const ratingColumnType: GridColTypeDef = {
...
+More details are available in the demo.
+
+| Props                    | Class                                     |
```
m4theushw, line 276:
```suggestion
| Props                    | CSS class                                     |
```

packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterPanel.tsx
```diff
@@ -74,14 +98,24 @@ export function GridFilterPanel() {
...
+
   React.useEffect(() => {
     if (items.length > 0) {
       lastFilterRef.current!.focus();
     }
   }, [items.length]);
 
   return (
-    <GridPanelWrapper>
+    <GridPanelWrapper sx={sx}>
```
m4theushw, line 118: We usually pass to the root all the remaining props.

https://github.com/mui/mui-x/blob/2bc8e32a4d52cdd89fb2903bce0cf8ee535267fd/packages/grid/_modules_/grid/components/menu/GridMenu.tsx#L101

m4theushw (approved):
Well done!

PR #3500, opened 2021-12-23
[core] Keep monorepo up-to-date automatically

oliviertassinari:
Try to fix https://github.com/mui-org/material-ui/pull/30122#issuecomment-1000205128. I'm not sure if this will work, based on https://docs.renovatebot.com/modules/datasource/.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -85,7 +85,7 @@
     },
     {
       "groupName": "monorepo",
-      "matchPackagePatterns": "@material-ui/monorepo",
+      "matchPackageNames": ["@material-ui/monorepo"],
       "matchDatasources": ["git-refs"]
     }
   ],
--- a/renovate.json
+++ b/renovate.json
@@ -82,6 +82,11 @@
         "@mui/system",
         "@mui/utils"
       ]
+    },
+    {
+      "groupName": "monorepo",
+      "matchPackagePatterns": "@material-ui/monorepo",
+      "matchDatasources": ["git-refs"]
     }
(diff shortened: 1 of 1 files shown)


m4theushw:
I tried this in #2924 and #2925 but it didn't work, then #2926 reverted.

The git data source is not suitable to the way that the monorepo dependency is specified. I wished that Renovate had better docs for it.

https://app.renovatebot.com/dashboard#github/mui-org/material-ui-x/536591075

image

It doesn't understand a branch as ref: https://github.com/renovatebot/renovate/blob/597da6ff59e5a3c984c0b7d0db91d13114512b26/lib/manager/npm/extract/index.ts#L326

It's in my TODO to dive into the code to see if it's viable to add support to branch-based refs on Renovate.

renovate.json
```diff
@@ -82,6 +82,11 @@
         "@mui/system",
         "@mui/utils"
       ]
+    },
+    {
+      "groupName": "monorepo",
+      "matchPackageNames": ["@material-ui/monorepo"],
+      "matchUpdateTypes": ["lockFileMaintenance"]
```
m4theushw, line 89: I tested in https://github.com/m4theushw/renovate-playground/blob/main/renovate.json and it didn't work. If `lockFileMaintenance` is not enabled, setting this option to this value doesn't have effect.

oliviertassinari: Alright you seem to have this under better hands than mine, I'm giving up 😁

PR #3501, opened 2021-12-23
[DataGrid] Fix focus on checkbox cells

### LINKED ISSUE #3382: [DataGrid] Persisted checkbox focus
### Duplicates

### Latest version

### Current behavior 😯

When we click on a checkbox and we want to click somewhere else, DataGrid keep the focus on the checkbox and scroll back to it.

### Expected behavior 🤔

When we click on a checkbox and we want to click somewhere else, DataGrid lose the focus on the checkbox and don't scroll back to it.

### Steps to reproduce 🕹

Steps:

1. Click on the checkbox
2. Scroll the page and click somewhere else

https://user-images.githubusercontent.com/12998550/145240398-49222395-b1c3-4a6d-ae03-994428f15dc2.mov

### Context 🔦

I used the DataGrid checkbox selection in my work and need to be able to scroll at the top of the page to click on a button after I choose the checkboxes without scrolling back.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3382

The stop propagation prevents having the selection event dispatched two times when clicking on the checkbox. One time by checkbox change, and the other one by row click event.

The problem is that it also prevents setting the focus on the cell. So now, interacting with the checkbox set the focus on its cell.

I did not add tests to detect the reported bug, because jsdom does not mock correctly, scrolling.

### DIFF
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -338,7 +338,7 @@ export const useGridSelection = (
         gridClasses.cell,
       );
       const field = cellClicked?.getAttribute('data-field');
-      if (field === '__check__') {
+      if (field === GRID_CHECKBOX_SELECTION_COL_DEF.field) {
         // click on checkbox should not trigger row selection
         return;
       }
--- a/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
+++ b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
@@ -338,6 +338,10 @@ export const useGridSelection = (
         gridClasses.cell,
       );
       const field = cellClicked?.getAttribute('data-field');
+      if (field === '__check__') {
+        // click on checkbox should not trigger row selection
+        return;
+      }
       if (field) {
         const column = apiRef.current.getColumn(field);
         if (column.type === 'actions') {
--- a/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
+++ b/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
@@ -198,7 +198,7 @@ describe('<DataGrid /> - Selection', () => {
       expect(getRow(0).querySelector('input')).to.have.property('checked', false);
     });
 
-    it('should set focus ont the cell when clicking the checkbox', () => {
+    it('should set focus on the cell when clicking the checkbox', () => {
       render(<TestDataGridSelection checkboxSelection />);
       expect(getActiveCell()).to.equal(null);
 
--- a/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
+++ b/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
@@ -183,6 +183,19 @@ describe('<DataGrid /> - Selection', () => {
       expect(getRow(0).querySelector('input')).to.have.property('checked', false);
     });
 
+    it('should set focus ont the cell when clicking the checkbox', () => {
+      render(<TestDataGridSelection checkboxSelection />);
+      expect(getActiveCell()).to.equal(null);
+
+      // simulate click
+      const checkboxInput = getCell(0, 0).querySelector('input');
+
+      fireEvent.mouseUp(checkboxInput);
+      fireEvent.click(checkboxInput);
+
+      expect(getActiveCell()).to.equal('0-0');
+    });
+
     it('should select all visible rows regardless of pagination', () => {
       render(<TestDataGridSelection checkboxSelection pageSize={1} rowsPerPageOptions={[1]} />);
       const selectAllCheckbox = document.querySelector('input[type="checkbox"]');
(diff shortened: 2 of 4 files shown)


flaviendelangle:
We could maybe add a test where we focus the checkbox, the click the body, then check if the checkbox is still focused

flaviendelangle (commented):
We could maybe add a test where we focus the checkbox, the click the body, then check if the checkbox is still focused

m4theushw:
> The stop propagation prevents having the selection event dispatched two times when clicking on the checkbox. One time by checkbox change, and the other one by row click event.

How about removing `event.stopPropagation` and, on `handleRowClick`, do nothing if the `event.target` is inside the checkbox column? I started to do that to avoid selecting the row when opening the actions column menu. The alternative would be to stop propagation but I didn't want this hack.

https://github.com/mui-org/material-ui-x/blob/dd4b11fb73dd33c4945550619bf517c05bc4211b/packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts#L338-L340

We can go even further in this approach and add a prop to `GridColDef` (see the description of #3492) to not select rows when an element inside the cell is clicked. I think that with more features being added this will become a common problem.

alexfauquette:
I agree on removing `event.stopPropagation` this would also fix #3382

But only blocking a column because in this case, it is the checkbox column that has similar interaction than row click seems a bit flacky.

Another idea would be to add some property to the event when a modification has been done. For example, when clicking on the checkbox, after modifying the selection, it add `event.selectionDone: true` such that when the row is listening to a click, it knows that selection has already been done by another listener before.

m4theushw:
Why do you think it's flaky? I see it differently. There're cell components that may receive clicks inside, since they contain interactive elements (e.g. the checkbox and the actions column), and the click shouldn't select the row. We already have `disableSelectionOnClick` but it disables the selection when clicking any column, not individually. The problem with keeping `event.stopPropagation()` is that eventually we will have to revisit this component to remove it.

alexfauquette:
Flaky is probably not the correct word. It is more about how to generalize the solution.

I feel like each time we will have a conflict between the event listener of nested components, we will have to find a hack in the listener of the parents to remove the duplicate. I would like to find, a solution to systematically prevent the parent from doing something already done by a child.

But I do not find a clear way of doing that, and it seems that most of the stopPropagation calls are made to deal with `ActionCell`/`row` interactions. So yes your solution should work, and we will see in the future if new features imply to redesign our strategy or not

m4theushw:
> I feel like each time we will have a conflict between the event listener of nested components, we will have to find a hack in the listener of the parents to remove the duplicate.

I don't think it's a hack. We only need to ignore all clicks coming from a specific field. The code to check which field was clicked is the same.

> I would like to find, a solution to systematically prevent the parent from doing something already done by a child.

This is complicated. We don't know what the child did. Currently, we only have problem with selection, but there could be other listeners on `rowClick`.

alexfauquette:
I followed @m4theushw suggestion. If the row click is from a checkbox cell, it does not trigger the selection

packages/grid/_modules_/grid/hooks/features/selection/useGridSelection.ts
```diff
@@ -338,6 +338,10 @@ export const useGridSelection = (
         gridClasses.cell,
       );
       const field = cellClicked?.getAttribute('data-field');
+      if (field === '__check__') {
```
m4theushw, line 341: If we change the field on day we don't need to update here.

```suggestion
      if (field === GRID_CHECKBOX_SELECTION_COL_DEF.field) {
```

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -174,15 +175,42 @@ describe('<DataGrid /> - Selection', () => {
...
+      fireEvent.click(getCell(0, 0).querySelector('input'));
+      expect(getSelectedRowIds()).to.deep.equal([0]);
+      expect(getRow(0).querySelector('input')).to.have.property('checked', true);
+
+      fireEvent.click(getCell(0, 0).querySelector('input'));
       expect(getSelectedRowIds()).to.deep.equal([]);
       expect(getRow(0).querySelector('input')).to.have.property('checked', false);
     });
 
+    it('should set focus ont the cell when clicking the checkbox', () => {
```
m4theushw, line 201:
```suggestion
    it('should set focus on the cell when clicking the checkbox', () => {
```

m4theushw (commented):
This also fixes another bug I noticed:

1. Open https://codesandbox.io/s/datagriddemo-material-demo-forked-s3u4i?file=/demo.js
2. Select one row
3. Click "Toggle checkbox selection"
4. 💥

Since there's no `click` sent by the checkbox, it doesn't clean the last clicked cell below:

https://github.com/mui-org/material-ui-x/blob/ed223c349df133419b025b08199c7e36ff82f1d4/packages/grid/_modules_/grid/hooks/features/focus/useGridFocus.ts#L134

alexfauquette:
You find really strange bugs 😅  I confirm, it does not appears on this branch

m4theushw (approved):

PR #3502, opened 2021-12-24
[DataGridPro] Fix resizing of right pinned columns

### LINKED ISSUE #3395: [DataGridPro] Resizing a right pinned column increases its width too fast
### Duplicates

### Latest version

### Current behavior 😯

When I try to resize a right pinned column, it's very hard to keep control without letting the column to grow all the way to the left.

VhNHCB7oNm

### Expected behavior 🤔

It should be easy to resize a right pinned column, as the same way it's for a left pinned column. The cursor should be moved to the left to increase the width to the left, instead of moving to the right.

### Steps to reproduce 🕹

Steps:

### Context 🔦

This was pointed by Flavien in https://github.com/mui-org/material-ui-x/pull/2946#issuecomment-970486648.

The explanation of why it happens can be found in https://github.com/mui-org/material-ui-x/pull/2946#issuecomment-977087097

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Closes #3395 

Preview: https://deploy-preview-3502--material-ui-x.netlify.app/components/data-grid/columns/#column-pinning

### DIFF
--- a/docs/pages/api-docs/data-grid/data-grid-pro.json
+++ b/docs/pages/api-docs/data-grid/data-grid-pro.json
@@ -272,6 +272,8 @@
       "columnHeadersInner--scrollable",
       "columnSeparator--resizable",
       "columnSeparator--resizing",
+      "columnSeparator--sideLeft",
+      "columnSeparator--sideRight",
       "columnSeparator",
       "columnsPanel",
       "columnsPanelRow",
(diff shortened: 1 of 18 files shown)


flaviendelangle (approved):
The behavior and logic look good to me

PR #3503, opened 2021-12-24
[docs] Create fake server hook

alexfauquette:
This hook takes in input the rows, and returns a `get` methods that work with cursor or row.

### DIFF
--- a/packages/grid/x-data-grid-generator/src/useFakeServer.ts
+++ b/packages/grid/x-data-grid-generator/src/useFakeServer.ts
@@ -1,0 +1,35 @@
+import * as React from 'react';
+import { GridRowModel } from '@mui/x-data-grid';
+
+function sleep(time: number): Promise<void> {
+  return new Promise<void>((res) => {
+    setTimeout(res, time);
+  });
+}
+
+export type FakeServerReturnType = {
+  get: any;
+};
+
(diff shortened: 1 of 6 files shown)


packages/grid/x-data-grid-generator/src/useFakeServer.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+}
+
+export type FakeServerReturnType = {
+  get: any;
+};
+
+export const useFakeServer = (rows: GridRowModel[]): FakeServerReturnType => {
+  const get = React.useCallback(
+    async ({ page = 0, pageSize = 10, cursor }) => {
```
oliviertassinari, line 16: Maybe it would be safer to not provide a default value to `page` to force the use of the API to be very explicit.

```suggestion
    async ({ page, pageSize = 10, cursor }) => {
```

oliviertassinari (commented):
It's an interesting initiative, two thoughts for the future:

1. For a large data set, having the API embedded in `useDemoData` could allow the data generation to be lazy, but it doesn't really matter for our small demos (100 rows). I wonder about the larger demos, which is my transition to 2.
2. Once we work on a fully-featured server-side demo, one that support filtering, sorting, pagination, lazy loading, we might have to introduce a SQL light engine like in https://www.ag-grid.com/javascript-data-grid/server-side-model-pagination/#server-pagination

PR #3515, opened 2021-12-27
[docs] Prepare scripts + E2E tests for migration

siriwatknp:


docs/data/data-grid/pages.ts
```diff
@@ -0,0 +1,55 @@
...
+    pathname: '/x/react-data-grid',
+    title: 'Components',
+    icon: 'ToggleOnIcon',
+    children: [
+      {
+        pathname: '/x/react-data-grid',
+        subheader: '/x/react-data-grid/overview',
+        title: 'Overview',
+      },
+      { pathname: '/x/react-data-grid/demo' },
```
flaviendelangle, line 12: Not sure about the `react-` prefix here
Could you explain why you added it instead of just `/x/data-grid/*` ?

docs/data/data-grid/pages.ts
```diff
@@ -0,0 +1,55 @@
...
+      { pathname: '/x/react-data-grid/accessibility' },
+      { pathname: '/x/react-data-grid/group-pivot', title: 'Group & Pivot' },
+    ],
+  },
+  {
+    pathname: '/x/api/mui-data-grid',
+    title: 'API',
+    icon: 'CodeIcon',
+    children: [
+      { pathname: '/x/api/mui-data-grid', title: 'API Reference' },
```
flaviendelangle, line 39: Same but `mui-`
Why not `react-data-grid` or `data-grid` ?

docs/pages/_app.js
```diff
@@ -195,36 +196,6 @@ Tip: you can access the documentation \`theme\` object directly in the console.
   );
 }
 
-// DO NOT REPLACE THIS FUNCTION WITH THE ONE FROM THE MONOREPO
```
flaviendelangle, line 198: Just in case, why was this comment here and did you check that we could replace it now ?

docs/data/data-grid/pages.ts
siriwatknp: I believe that this benefits the SEO proposed by @oliviertassinari in https://www.notion.so/mui-org/Documentation-per-product-decision-time-d9347d7d37b641129d2fea12d68555d2#67b6d390ec7c4977b5cb4d942ba2ee6b

docs/data/data-grid/pages.ts
siriwatknp: Same reason as above.

docs/pages/_app.js
siriwatknp: > did you check that we could replace it now

These are what I checked:
- the e2e-test make sure the page does not crash
- the footer of the docs works as before (can link to the next page)

siriwatknp: I suggest you try the migration steps I mentioned in the PR description.

docs/data/data-grid/pages.ts
flaviendelangle: But why `react-` on components and `mui-` on the apis ?
Just to have benefits for the  SEO for both words ?

siriwatknp: For `mui-`, I think it relates to the package or the folder in the repo:
- `/material/api/mui-material/button`
- `/material/api/mui-lab/button` (here `mui-lab` lives in material)

But this is an interesting question for `x` packages 🤔, `mui-` does not make sense here. @mui-org/x @oliviertassinari any thought?

package.json
```diff
@@ -65,11 +68,12 @@
     "@emotion/react": "^11.7.1",
     "@emotion/styled": "^11.6.0",
     "@eps1lon/enzyme-adapter-react-17": "^0.1.0",
-    "@material-ui/monorepo": "https://github.com/mui-org/material-ui.git#master",
```
siriwatknp, line 68: @flaviendelangle FYI, I think it makes more sense to update `package.json` rather than `yarn.lock`.

docs/data/data-grid/pages.ts
flaviendelangle: I'm resolving this conversation to keep a single thread

m4theushw (commented):
The script removed the page in https://mui.com/api/data-grid/. It's manually authored.

docs/data/data-grid/pages.ts
oliviertassinari: I have polished https://www.notion.so/mui-org/Documentation-per-product-decision-time-d9347d7d37b641129d2fea12d68555d2#67b6d390ec7c4977b5cb4d942ba2ee6b to be more in sync with the current direction. It covers why `react-`.

Regarding why `mui-` we have never really talked about it so far. For me, the idea was to match the code imports `import Button from '@mui/material/Button`. To be honest, there, If it was only about me, I think that I would have done a 1:1 mapping as close as possible with code, so https://mui.com/api/mui-material/button/. But skipping `mui-` or adding the product prefix makes sense too.

siriwatknp: In my opinion, the product prefix is already explicit enough and as a user, I don't really care about the package or folder (in the URL). So I am good if we keep the URL the same and just add the product prefix.

- `/material/api/button`
- `/x/api/data-grid/data-grid`
- `/x/api/data-grid/grid-api`

siriwatknp: @m4theushw @flaviendelangle is this format looks good to you (similar to existing URLs, just prepend with the prefix `x`)?
- /x/api/data-grid/data-grid
- /x/api/data-grid/grid-api

package.json
siriwatknp: Okay, I will revert the change.

docs/data/data-grid/pages.ts
flaviendelangle: The URL seems clean to me that way yes :+1:

m4theushw:
> The script removed the page in https://mui.com/api/data-grid/. It's manually authored.

Now I see a 404 when a try to access the "API Reference" page in the sidebar.

---

I also noticed that the items in the sidebar are one level up. We plan to add more components (e.g. date picker) and it will also have its dedicated set of pages.

Before:

image

After:

image

siriwatknp:
> I also noticed that the items in the sidebar are one level up

I think this is expected because now the data-grid has its own scope. I have changed the "Components" to "Data Grid".

> We plan to add more components (e.g. date picker) and it will also have its dedicated set of pages.

Yes, I believe that it will be
- `/x/react-date-picker/`
- `/x/api/date-picker/date-picker/`

flaviendelangle:
I like having one scope per set of components.
That way we will be able to split the biggest pages and have a structure closer from AG Grid.
The namings are of course temporary.

```
- Data Grid
  - Columns
    - Column Definition
    - Column Visibility
    - Column Pinning
    - Column Sizing
    - Column Groups
  - Filtering
    - Simple Column Filters
    - Custom Filter Operators
    - Server Side features
    - ...
  - Grouping & Pivot
    - Row Grouping
    - Tree Data
    - Aggregation
    - Pivoting
    - ...
```

docs/data/data-grid/pages.ts
oliviertassinari: Would it be simpler by removing the /data-grid/ prefix? So using the previous example:

- /x/api/data-grid/
- /x/api/data-grid-overview!/

It would match what we do in the other "spaces".

siriwatknp:
@m4theushw @flaviendelangle Can you review again? I think this is good to be merged.

Here is the result once we run the migration (I create a migration demo in this PR):
https://deploy-preview-3664--material-ui-x.netlify.app/x/react-data-grid/

cc @oliviertassinari

m4theushw (approved):
Looks good!

flaviendelangle (approved):
Great work !

PR #271, opened 2020-09-12
[DataGrid] Add footer border top

oliviertassinari:
I'm not entirely sure about this change. We are moving one case from KO to OK and two OK to OK-ish.

- **before** vertical scrollbar: KO
before-vertical scrollbar

- **before** no scrollbar: OK
before-no scrollbar

- **before** horizontal scrollbar: OK
before-horizontal scrollbar

- **after** vertical scrollbar: OK
after-vertical scrollbar

- **after** no scrollbar: OK-ish
after-no scrollbar

- **after** horizontal scrollbar: OK-ish
after-horizontal scrollbar

Maybe we should add a new class name that is only applied when there is a vertical scrollbar and no horizontal one?

PR #3519, opened 2021-12-28
[DataGrid] Remove `GridCell` `borderBottom` when last row

### LINKED ISSUE #2315: [DataGrid] `autoHeight` + `hideFooter` results in double bottom border

DanailH:
Fixes #2315 

I've done a potential fix for the issue. The drawback is that now that you have fewer rows than what the grid can load (a black space is visible) and you have a footer enabled then the footer also has a `borderTop`. I think this is how it should have been from the beginning but still, it is technically a breaking change so we need to review it before proceeding.

You can check what the change is here: ...

https://deploy-preview-3519--material-ui-x.netlify.app/components/data-grid/


oliviertassinari:
It reminds me of #271. IMHO the right solution is a tricker. I see a couple of KO cases in the live preview.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -269,9 +270,11 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
 
     const rows: JSX.Element[] = [];
 
+    const isEmptySpace = nextRenderContext.lastRowIndex >= currentPage.rows.length;
+
     for (let i = 0; i < renderedRows.length; i += 1) {
       const { id, model } = renderedRows[i];
-
+      const isLastRow = i === renderedRows.length - 1;
```
m4theushw, line 277: This will be `true` when on the last row of the **current render context**, which is different from the index relative to all visible rows. Is this what you want? The `index` prop gives the "real" index.

DanailH: Hmm, the `index` might also work. The thing is that I need to know if there are fewer rows than the height of the viewport so I can add the border to the last row.

DanailH:
> It reminds me of this attempt that I did in #271 at the problem but on which I failed by lack of time. IMHO the right solution is trickier. I see a couple of KO cases in the live preview.

Yes, I saw the `autoHeight` problem. Besides that  and the fact that now if there are fewer rows than what the viewport can take (you have an empty space between the footer and the last row) then both the last row and the footer have a border.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
m4theushw: I think what you need to do is something that's already being done for the columns. Instead of `columnsMeta.totalWidth`, use `currentPage.rows.length * rowHeight`. To get the height you can try using `apiRef.current.getRootDimensions`.

https://github.com/mui-org/material-ui-x/blob/6208886dbb143f7365f8387ef9d91a6261a3a9d3/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx#L299-L303

DanailH: But the current way also works. I don't need to calculate the height, I can know if there is space just by checking the indexes.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -269,8 +270,12 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
 
     const rows: JSX.Element[] = [];
 
+    const isEmptySpace = nextRenderContext.lastRowIndex >= currentPage.rows.length;
```
m4theushw, line 273: When `nextRenderContext.lastRowIndex` is greater than `currentPage.rows.length`? It's clamped to the number of rows.

https://github.com/mui-org/material-ui-x/blob/9e4c04f1b6d5a2ef8504c81c1cfe295c2f392ee4/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx#L162-L166

https://github.com/mui-org/material-ui-x/blob/9e4c04f1b6d5a2ef8504c81c1cfe295c2f392ee4/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx#L153-L156

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -288,6 +293,12 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
           selected={selectionLookup[id] !== undefined}
           index={currentPage.range.firstRowIndex + nextRenderContext.firstRowIndex! + i}
           containerWidth={availableSpace}
+          className={
+            isEmptySpace &&
+            isLastRow &&
+            !rootProps.autoHeight &&
+            gridClasses['row--lastBeforeEmpty']
```
m4theushw, line 300: Could we keep the calculation of the class name inside the `useUtilityClasses` of the `GridRow`? You can pass the input values you need as props. This allows the users to opt-out from this feature by providing a custom row through the slots.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -212,6 +212,9 @@ export const GridRootStyles = styled('div', {
           },
         },
       },
+      [`&:last-child:not(.${gridClasses['row--lastBeforeEmpty']}) .${gridClasses.cell}`]: {
```
m4theushw, line 215: This selector targets the last visible row that the virtualization decided to render. `:last-child` doesn't mean the last row of the dataset. Also, could we simplify it? Maybe target a row which has a `.MuiDataGrid-row--last` (being the true last row)?

Scrolling https://deploy-preview-3519--material-ui-x.netlify.app/components/data-grid/#commercial-version slowly I can see that sometimes some cells don't have the border bottom.

image

This is much more prominent depending on the value of `rowThreshold`: https://codesandbox.io/s/datagridprodemo-material-demo-forked-jsnbs?file=/demo.js

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
DanailH: This is just to catch if the total number of rows is less than the number of rows that can fit in the viewport (user-visible rows). If there is a better way I'm up for it.

DanailH: Updated - I used the `scrollY` in the `GridRow`.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
DanailH: Updated. I added the `row-last` class.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -282,6 +282,7 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
 
     for (let i = 0; i < renderedRows.length; i += 1) {
       const { id, model } = renderedRows[i];
+      const isLastRow = currentPage.range.firstRowIndex + i === currentPage.rows.length - 1;
```
m4theushw, line 285: I don't think this formula is giving the correct value. In https://deploy-preview-3519--material-ui-x.netlify.app/components/data-grid/rows/#row-height there're 100 rows and the last row has `data-rowindex=102` when it should end at 99.

image

In #3838 I have a situation where I also need to know if one row is the last one. I solved by using `currentPage.range.firstRowIndex + firstRowToRender + i`: https://github.com/mui-org/material-ui-x/pull/3848/files#diff-7fb7dbb1d7fe2c699d9fc5cfc5c6a56b2489af3b46dc1a2a8eaa3404575cacbdR313

The rationale is:

- `firstRowToRender` gives the first row to render in the current render context
- `i` gives which row we're rendering inside the current render context
- `currentPage.range.firstRowIndex` gives the first row to base with when pagination is on, because in the 2nd page `firstRowToRender` is 0.

siriwatknp:

3. run `yarn docs:api`

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
m4theushw: Yes, makes sense. I'm using `firstRowToRender + i` to pass ´isLastVisible` in https://github.com/mui/mui-x/pull/3848/files#diff-d686bc46bbd2d70d8fe115f2d0f87743bf452fddf658b5dbe8c79d3bd0bb52cbR151

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -222,6 +222,10 @@ export const GridRootStyles = styled('div', {
           },
         },
       },
+      [`&.${gridClasses['row--last']}:not(.${gridClasses['row--lastBeforeEmpty']}) .${gridClasses.cell}`]:
```
m4theushw, line 225: I found the purpose of `row--lastBeforeEmpty` hard to understand. How about adding a class to `MuiDataGrid-virtualScrollerContent` to signal that its content doesn't extend to the full height of the grid. You'll need to wait for #3821 to able to do it. In DataGridPro the calculation is already being done, we only need to add the class name. 

https://github.com/mui/mui-x/blob/996f727467311f20e999b8db3d1331e3526594fd/packages/grid/x-data-grid-pro/src/DataGridProVirtualScroller.tsx#L249-L250

Once you have the class name added you can create a selector like `.MuiDataGrid-root:not(.MuiDataGrid-autoHeight) .MuiDataGrid-virtualScroller:not(.MuiDataGrid-virtualScrollerContent--newClassAdded) .MuiDataGrid-row--last`.

packages/grid/_modules_/grid/gridClasses.ts
```diff
@@ -407,6 +407,8 @@ export const gridClasses = generateUtilityClasses('MuiDataGrid', [
   'row--editable',
   'row--editing',
   'row',
+  'row--lastBeforeEmpty',
+  'row--last',
```
m4theushw, line 411: You also need to add these classes to the interface in this same file.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
DanailH: That will work as well. I can put the on hold for now for the PR until the other one is merged.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -312,6 +313,7 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
           selected={isSelected}
           index={currentPage.range.firstRowIndex + nextRenderContext.firstRowIndex! + i}
           containerWidth={availableSpace}
+          isCurrentPageLastRow={currentPageLastRow}
```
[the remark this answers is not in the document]
DanailH: Yes, sorry I missed that - updated.

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
DanailH: Updated

packages/grid/x-data-grid/src/internals/components/virtualization/GridVirtualScrollerContent.tsx
```diff
@@ -30,13 +30,18 @@ const GridVirtualScrollerContent = React.forwardRef<
   HTMLDivElement,
   React.HTMLAttributes<HTMLDivElement>
 >(function GridVirtualScrollerContent(props, ref) {
-  const { className, ...other } = props;
+  const { className, style, ...other } = props;
   const rootProps = useGridRootProps();
-  const ownerState = { classes: rootProps.classes };
+  const ownerState = { classes: rootProps.classes, overflowedContent: style?.minHeight === 'auto' };
```
m4theushw, line 35: I noticed that `minHeight` is not reliable when `autoHeight` is on. Removing a row and re-adding it in https://deploy-preview-3519--material-ui-x.netlify.app/components/data-grid/layout/#auto-height, adds the overflow class. I propose to use:

```suggestion
  const ownerState = { classes: rootProps.classes, overflowedContent: !rootProps.autoHeight && style?.minHeight === 'auto' };
```

The content never overflows when `autoHeight` is on.

image

packages/grid/x-data-grid/src/internals/components/containers/GridRootStyles.ts
```diff
@@ -73,6 +73,10 @@ export const GridRootStyles = styled('div', {
     [`&.${gridClasses.autoHeight}`]: {
       height: 'auto',
     },
+    [`& .${gridClasses['virtualScrollerContent--overflowed']} .${gridClasses['row--lastVisible']} .${gridClasses.cell}`]:
```
m4theushw, line 76: The content is not supposed to overflow when `autoHeight` is on. We need to also remove the border in this condition. In https://deploy-preview-3519--material-ui-x.netlify.app/components/data-grid/layout/#auto-height sometimes I still see the double border.

image

DanailH: Ah yes thanks for the catch!

m4theushw (approved):
Good job!

PR #3521, opened 2021-12-29
[core] Add ESLint rule to prevent direct state access

m4theushw:
This is an attempt to add a custom ESLint exclusive for the DataGrid. The idea is to prevents us from doing `apiRef.current.state.foo`, instead of using a selector.

image


packages/eslint-plugin-material-ui/src/rules/no-direct-state-access.test.js
```diff
@@ -0,0 +1,74 @@
+// The polyfill of `import.meta.url` used @eslint/eslintrc tests the presence
```
m4theushw, line 1:
```suggestion
// The polyfill of `import.meta.url` used in @eslint/eslintrc tests the presence
```

flaviendelangle (approved):

PR #3527, opened 2021-12-30
[benchmark] Log the output of the script

oliviertassinari:
So we get a comparison point the next time we run the script. Regarding the measure itself, I'm not so convinced about how it correlates with UX performance, but this is another discussion.

### DIFF
--- a/benchmark/browser/README.md
+++ b/benchmark/browser/README.md
@@ -1,0 +1,60 @@
+# Browser benchmark
+
+This project is used when running the following command:
+
+```sh
+yarn browser
+```
+
+It is suppose to give developers comparable values between running different scenarios inside the browser, that can be find the `./scenarios` folder.
+
+You should use these numbers exclusively for comparing performance between different scenarios, not as absolute values.
+
+## Output
+
+```
+React Virtualized:
+  Vertical scroll:
+       Min: 44.916 fps (σ = 22.443)
+       Max: 104.586 fps (σ = 2.355)
+    Median: 91.986 fps (σ = 0.62)
+      Mean: 92.574 fps (σ = 0.608)
+  Horizontal scroll:
+       Min: 56.892 fps (σ = 14.997)
+       Max: 104.776 fps (σ = 2.154)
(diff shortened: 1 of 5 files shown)


test/utils/helperFn.ts
```diff
@@ -50,7 +50,7 @@ export function getActiveColumnHeader() {
   return `${Number(activeElement.getAttribute('aria-colindex')) - 1}`;
 }
 
-export async function sleep(duration: number) {
+export function sleep(duration: number) {
```
oliviertassinari, line 53: async is not needed here, it returns a promise, it's enough.

benchmark/browser/scripts/benchmark.js
```diff
@@ -8,7 +8,7 @@ const http = require('http');
 
 const PORT = 1122;
 
-async function delay(ms) {
+function sleep(ms) {
```
oliviertassinari, line 11: Only for naming consistency. Yes, Rx.js uses https://rxjs.dev/api/operators/delay.

benchmark/browser/README.md
```diff
@@ -0,0 +1,60 @@
+# Browser benchmark
+
+This project is used when running the following command:
+
+```sh
+yarn browser
+```
+
+It is suppose to give developers comparable values between running different scenarios inside the browser, that can be find the `./scenarios` folder.
```
m4theushw, line 9:
```suggestion
It is supposed to give developers comparable values between running different scenarios inside the browser, that can be found in the `./scenarios` folder.
```

test/utils/helperFn.ts
m4theushw: This helper function appears to never be called. Instead, in every test, where we need to wait, we create a local function. We should use `sleep` instead of duplicating code.

https://github.com/mui-org/material-ui-x/blob/6208886dbb143f7365f8387ef9d91a6261a3a9d3/test/e2e/index.test.ts#L4

m4theushw (approved):
Only a small typo. Otherwise, looks good.

PR #3547, opened 2022-01-03
[DataGridPro] Block edition for auto-generated rows

### LINKED ISSUE #3482: [DataGridPro] Tree data gaps are editable
### Duplicates

### Latest version

### Current behavior 😯

Gap rows generated by the tree data are editable, but editing them leads to an error, because these rows have no id

### Expected behavior 🤔

I assume that if there are missing rows in the tree data it is because the rows do not make sense at this hierarchical level, so the cell should no editable.

The other solution is to insert rows, but I think such a behavior (let datagrid add rows by itself) will be hard to manage for developers

### Steps to reproduce 🕹

Steps:

1. Go to https://codesandbox.io/s/treedatawithgap-material-demo-forked-1i144?file=/demo.tsx
2. Edit cells of a row

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #3482


alexfauquette (approved):
I didn't know it would be that simple

Issue #3091, opened 2021-11-04
[DataGrid] Flex values below 1 cause unwanted horizontal scroll


Haaxor1689:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

I would like to add a column that is always at the right end of the grid containing actions. To achieve this I've defined this column with `flex: 0.0001` (arbitrary small number). So in theory:

1. In tables without flex columns, its width fills the remaining empty space
2. In tables with flex columns, it won't use up any additional width

The problem is that in tables with flex columns I get unexpected horizontal scroll when using this column with flex lower than 1.

### Expected behavior 🤔

There shouldn't be any unwanted horizontal scroll if I have a column with flex value lower than 1.

### Steps to reproduce 🕹

Steps:

1. Create a DataGrid with multiple flex columns where one has flex value below 1

[Live demo](https://codesandbox.io/s/datagridpro-v5-quick-start-forked-id97f?file=/src/App.tsx)

### Context 🔦

I would like to have a column that is always at the right most end of the grid.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```  System:
    OS: Windows 10 10.0.19042
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor                
    Memory: 1.90 GB / 15.92 GB
  Binaries:
    Node: 16.9.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.9 - C:\Program Files\nodejs\npm.CMD
  Managers:
    Maven: 3.6.3 - C:\Projects\_Tools\apache-maven-3.6.3\bin\mvn.CMD
    pip3: 21.2.3 - C:\Python39\Scripts\pip3.EXE
  Utilities:
    CMake: 3.15.4 - /c/Program Files/CMake/bin/cmake
    Git: 2.33.0. - /cmd/git
  Virtualization:
    Docker: 20.10.7 - C:\Program Files\Docker\Docker\resources\bin\docker.EXE
  IDEs:
    VSCode: 1.61.2 - C:\Users\betko\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD  
    Visual Studio: 16.3.29324.140 (Visual Studio Community 2019)
  Languages:
    Java: javac 11 - /c/Projects/_Tools/jdk-11/bin/javac
[...]

PR #3550, opened 2022-01-04
[DataGrid] Introduce a `maxWidth` property in `GridColDef`

### LINKED ISSUE #2174: [DataGrid] Introduce `maxWidth` in `GridColDef`

## Summary 💡

A follow-up issue based on this comment https://github.com/mui-org/material-ui-x/pull/2101

The goal is to provide an option for developers to have full control over the width of a column by supporting `minWidth`, `maxWidth`, and `width`. `maxWidth`. This prop would be primarily targeted to work with the flex functionalities. We would also have to consider how the combination with column resize should happen.

flaviendelangle:
Closes #2174

The next step would be to improve the flex behavior to respect the CSS specification (#3091 for instance).


cherniavskii (approved):
LGTM

oliviertassinari:
@flaviendelangle I have tried the use case raised in https://github.com/mui-org/material-ui-x/issues/2174#issuecomment-996284048 (reading the git diff, I had a few doubts about the correctness). It doesn't seem to work correctly. I would expect the last column of the second demo to be full-width. Proof: https://codesandbox.io/s/columnsizinggrid-material-demo-forked-nhcv3?file=/package.json.

```jsx
      <div style={{ height: 200, width: "100%" }}>
        <DataGridPro
          columns={[
            { field: "id" },
            { field: "username", width: 100 },
            { field: "age", resizable: false, flex: 1 }
          ]}
          rows={rows}
        />
      </div>
      <div style={{ height: 200, width: "100%" }}>
        <DataGridPro
          columns={[
            { field: "id" },
            { field: "username", flex: 1, maxWidth: 100 },
            { field: "age", resizable: false, flex: 1 }
          ]}
          rows={rows}
        />
      </div>
```

Screenshot 2022-01-22 at 00 50 55

I have tried with AG Grid, it seems correct: https://plnkr.co/edit/WJCcSwIr69GwJdoc.

How about we open a new GitHub issue?

flaviendelangle:
You are right :+1:

PR #3551, opened 2022-01-05
[core] Report performance test results on each PR

m4theushw:
This PR removes the performance tests that were running as unit test. Instead, they run now in a different environment, with the React production build and closer to how the user interacts with the grid. The results will be reported as a comment in each PR, that means that the PR, initially, won't fail if a regression is introduced.

The comparison between these results and a "gold master" (assumed to be the last released version) will be done after the merge and once we have a performance snapshot for a release.

As inspiration: https://github.com/saucelabs/performance-CI-demo/blob/main/README.md


m4theushw:
@oliviertassinari Could you set `DANGER_GITHUB_API_TOKEN` on this repo? I want to use Dangerfile to report the results from the performance tests. Currently, the tests we have are unreliable since they run in a unit-test environment, with a fake clock. Also, sometimes they fail for no reason which forces to run the CI again.

Example:

| Test case            | Min   | Max   | Median | Mean  |
| -------------------- | ----- | ----- | ------ | ----- |
| Filter 100000 rows   | 179.4 | 179.4 | 179.4  | 179.4 |
| Sort 100000 rows     | 451.6 | 451.6 | 451.6  | 451.6 |
| Select 100000 rows   | 257.8 | 257.8 | 257.8  | 257.8 |
| Deselect 100000 rows | 138.1 | 138.1 | 138.1  | 138.1 |

oliviertassinari:
@m4theushw Done. The comment should come from https://github.com/mui-bot going forward.

flaviendelangle:
@m4theushw could we add the standard deviation or the variance to have a glance of the value accurateness ?
I will not treat the same way a variation à 10ms if the variance is of 1ms or 50ms.

m4theushw:
@flaviendelangle I added the standard deviation. As you can see, the values vary substantially. This means that we won't be able to see if a PR has caused a small impact on performance. However, it's going to be useful for situations where a change causes a drastic performance hit, in which a unit test, using a small dataset, wouldn't fail due timeout. In 154cc7e I intentionally caused a performance degradation (ignore the failing unit tests). The time to select all rows is 60x times higher. This is a clear signal that something is wrong.

_I'll revert 154cc7e later._

flaviendelangle:
Seems perfect :+1:

flaviendelangle (approved):

PR #3558, opened 2022-01-06
[DataGrid] Fix potential memory leak warning

### LINKED ISSUE #3276: [DataGrid] Setting `autoHeight` prop or parent element style prop `height` causes potential memory leak
### Duplicates

### Latest version

### Current behavior 😯

Observed in x-data-grid 5.0.1 with material 5.2.0 and 5.1.0, Chrome  96.0.4664.45

When setting `autoHeight={true}` on DataGrid component such as:

```
<div id="grid-parent">
    <DataGrid rows={rows} columns={columns} autoHeight={true} />
</div>
```

-OR-

Setting a dynamic height on the parent element such as:

```
<div id="grid-parent" style={{ height: 'calc(100vh - 90px)' }}>
    <DataGrid rows={rows} columns={columns} />
</div>
```

Then immediately upon resizing the browser window, React StrictMode warning appears in browser console:

"Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function."

Removing the autoHeight prop, or removing the dynamic calculated height on the parent div, the memory leak no longer occurs.

Changing the parent div height to a static value e.g. ` style={{ height: 300 }}` will cause the error, but only when the window is resized to a point where the bottom window border begins to touch the bottom grid border which enables the horizontal scroll for the entire window.

While this warning only appears in React StrictMode (and therefore only in development mode) it would still presumably be a problem in production even though the warning is not present, which is expected behavior since StrictMode warnings are meant to indicate problems to the developer.

### Expected behavior 🤔

Setting the autoHeight prop on DataGrid or setting a height in the inline style prop of the DataGrid parent div element should not cause a React state update on an unmounted component memory leak warning.

### Steps to reproduce 🕹

Steps:

1. Set autoHeight prop on DataGrid component
2. OR set inline style prop with height on DataGrid parent div element
3. Resize window by dragging bottom window border
4. If using static value for height in inline style, resize window by dragging bottom window border up past DataGrid bottom border
5. React StrictMode warning in console appears

### Context 🔦

Trying to accomplish setting dynamic height to DataGrid in order to auto resize DataGrid as user resizes their browser window. While the feature does work, underlying state and memory management seems to be affected.

### Your environment 🌎

<details>

```
Browser Used: Chrome  96.0.4664.45

    System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 14.17.6 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.14.15 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 96.0.4664.45
    Edge: Spartan (44.19041.1266.0), Chromium (96.0.1054.29)
  npmPackages:
    @mui/base:  5.0.0-alpha.56
    @mui/icons-material: ^5.2.0 => 5.2.0
    @mui/material: ^5.2.0 => 5.2.0
    @mui/private-theming:  5.2.0
    @mui/styled-engine:  5.2.0
    @mui/system:  5.2.0
    @mui/types:  7.1.0
    @mui/utils:  5.1.1
    @mui/x-data-grid: ^5.0.1 => 5.0.1
    @types/react:  17.0.35
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3276

Preview: https://deploy-preview-3558--material-ui-x.netlify.app/components/data-grid/

I don't know why we never tried that. The warning is always triggered inside `GridOverlay` as part of `GridOverlays`. It turns out that `GridOverlays` is rendered before `GridAutoSizer`. That way, subscribing inside a normal effect has no risk of losing any resize event, because React executes the effects sequentially according to the tree.

### DIFF
--- a/packages/grid/_modules_/grid/components/base/GridErrorHandler.tsx
+++ b/packages/grid/_modules_/grid/components/base/GridErrorHandler.tsx
@@ -4,6 +4,8 @@ import { useGridLogger } from '../../hooks/utils/useGridLogger';
 import { GridMainContainer } from '../containers/GridMainContainer';
 import { ErrorBoundary } from '../ErrorBoundary';
 import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
+import { GridAutoSizer, AutoSizerSize } from '../GridAutoSizer';
+import { GridEvents } from '../../models/events/gridEvents';
 
 export function GridErrorHandler(props) {
   const { children } = props;
@@ -11,6 +13,13 @@ export function GridErrorHandler(props) {
   const logger = useGridLogger(apiRef, 'GridErrorHandler');
   const rootProps = useGridRootProps();
   const error = apiRef.current.state.error;
+
+  const handleResize = React.useCallback(
+    (size: AutoSizerSize) => {
+      apiRef.current.publishEvent(GridEvents.resize, size);
+    },
+    [apiRef],
+  );
 
   return (
     <ErrorBoundary
@@ -20,10 +29,18 @@ export function GridErrorHandler(props) {
       logger={logger}
       render={(errorProps) => (
         <GridMainContainer>
-          <rootProps.components.ErrorOverlay
-            {...errorProps}
-            {...rootProps.componentsProps?.errorOverlay}
-          />
+          <GridAutoSizer
+            nonce={rootProps.nonce}
+            disableHeight={rootProps.autoHeight}
+            onResize={handleResize}
+          >
+            {() => (
+              <rootProps.components.ErrorOverlay
+                {...errorProps}
+                {...rootProps.componentsProps?.errorOverlay}
+              />
+            )}
+          </GridAutoSizer>
         </GridMainContainer>
       )}
     >
--- a/packages/grid/_modules_/grid/components/containers/GridOverlay.tsx
+++ b/packages/grid/_modules_/grid/components/containers/GridOverlay.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { unstable_useEnhancedEffect as useEnhancedEffect } from '@mui/material/utils';
 import clsx from 'clsx';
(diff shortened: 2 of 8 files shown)


packages/grid/_modules_/grid/components/base/GridErrorHandler.tsx
```diff
@@ -20,10 +29,18 @@ export function GridErrorHandler(props) {
       logger={logger}
       render={(errorProps) => (
         <GridMainContainer>
-          <rootProps.components.ErrorOverlay
-            {...errorProps}
-            {...rootProps.componentsProps?.errorOverlay}
-          />
+          <GridAutoSizer
```
m4theushw, line 32: Fixes https://codesandbox.io/s/datagriddemo-material-demo-forked-49plk?file=/demo.tsx

If the error is set by the `error` prop, `GridAutoSizer` in `GridBody` won't render and we won't have any dimension.

packages/grid/_modules_/grid/components/containers/GridOverlay.tsx
```diff
@@ -57,16 +57,26 @@ export const GridOverlay = React.forwardRef<HTMLDivElement, GridOverlayProps>(fu
...
+      handleViewportSizeChange,
+    );
+  }, [apiRef, handleViewportSizeChange]);
 
   let height: React.CSSProperties['height'] = viewportInnerSize?.height ?? 0;
   if (rootProps.autoHeight && height === 0) {
     height = 'auto';
   }
 
+  if (!viewportInnerSize) {
```
m4theushw, line 76: To prevent misalignment when there's no dimension yet.

flaviendelangle (approved):

PR #3562, opened 2022-01-07
[DataGrid] Make row editing work with preProcessEditCellProps

### LINKED ISSUE #3557: [DataGrid] Row edit commit is not working when preProcessEditCellProps is defined
### Duplicates

### Latest version

### Current behavior 😯

If in any column is defined 'preProcessEditCellProps' function, the commit of the row in table by pressing Enter encountered an error: _MUI: Row at id: x is not being edited._

Demo: https://codesandbox.io/s/roweditcontrolgrid-material-demo-forked-04sci?file=/demo.tsx

If you comment lines 50-53, commiting by enter will work fine.

### Expected behavior 🤔

Should be working as usual without defined preProcessEditCellProps function in column.

### Steps to reproduce 🕹

Steps:

1. Try edit any cell in demo
2. Press enter to commit changes

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

31956

flaviendelangle:
Fixes #3557

I think the behavior is still perfectible.
If you validation takes time and you press "Enter" before it ends, it will save the last valid value, even if it's not the one displayed on screen. Maybe it would be better to not allow committing while validating.

But I would be in favor of waiting after the row / cell file split before adding additional complexity to this feature.


packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
```diff
@@ -1520,6 +1520,34 @@ describe('<DataGridPro /> - Edit Rows', () => {
...
+              field: 'brand',
+              editable: true,
+              preProcessEditCellProps: ({ props }) => props,
+            },
+          ]}
+          rows={[{ id: 0, brand: 'Nike' }]}
+        />,
+      );
+      const cell = getCell(0, 0);
+      cell.focus();
```
m4theushw, line 1539: The internal state is not updated when only `cell.focus()` is called. See #3486.

```suggestion
      fireEvent.mouseUp(cell);
```

m4theushw (approved):

packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
flaviendelangle: Other tests should probably updated then, I copied this logic :+1:

flaviendelangle: Fixed + TODO for other tests

packages/grid/x-data-grid-pro/src/tests/editRows.DataGridPro.test.tsx
```diff
@@ -1520,6 +1521,34 @@ describe('<DataGridPro /> - Edit Rows', () => {
...
+          ]}
+          rows={[{ id: 0, brand: 'Nike' }]}
+        />,
+      );
+      const cell = getCell(0, 0);
+      fireEvent.mouseUp(cell);
+      fireEvent.doubleClick(cell);
+      const input = cell.querySelector('input')!;
+      fireEvent.change(input, { target: { value: 'Adidas' } });
+      await clock.runToLast();
```
m4theushw, line 1544: Why the `await` here? This test is not passing on Safari.

flaviendelangle: I spent quite some time to make this test work.
It did not on Chrome Headless without the `await`
But if the `await` causes problem then the solution is not good.

PR #3588, opened 2022-01-10
[core] update l10n issue with a single command line

alexfauquette:


package.json
```diff
@@ -167,6 +167,7 @@
     ]
   },
   "dependencies": {
+    "@octokit/rest": "^18.12.0",
```
m4theushw, line 170: It's a `devDependency`.

m4theushw:
Do you plan to create the GitHub workflow? It would be nice to have this automated.

alexfauquette:
I'm not planning to do it for now, because I do not know exactly how files are managed in github workflow. Which could be a problem, especially with the following part of the code.

```ts
function git(args: any) {
  return new Promise((resolve, reject) => {
    exec(`git ${args}`, (err, stdout) => {
      if (err) {
        reject(err);
      } else {
        resolve(stdout.trim());
      }
    });
  });
}
```
I agree it would be better, but for now, updating the issue at least once a week during the release seems enought

m4theushw (approved):

scripts/README.md
```diff
@@ -63,6 +63,16 @@ You can use the following script in your browser console on any GitHub page to a
 4. Release the versions on NPM: `yarn release:publish` (you need your 2FA device).
 5. Create a new tag named with the release you just did `git tag v4.0.0-alpha.30 && git push upstream --tag`
 
+### Update l10n issue
```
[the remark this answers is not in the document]
alexfauquette: So I let it, and you will remove it in the PR to add the GitHub workflow?

m4theushw:
> I'm not planning to do it for now, because I do not know exactly how files are managed in github workflow. Which could be a problem, especially with the following part of the code.

It works in the same way as in your machine. We first `git checkout` with a special GitHub action, then `yarn install` and finally run the command to update the issue.

https://github.com/mui-org/material-ui/blob/master/.github/workflows/ci.yml#L22

alexfauquette:
@m4theushw I made the script to make the update when there is a push on master. If you can verify it. I don't know how to test it except merging and hope it works

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,29 @@
+name: CI
```
m4theushw, line 1:
```suggestion
name: Update missing translations
```

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: 'yarn' # https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies
+      - run: yarn install
+        env:
+          # Don't need playwright in this job
+          PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: yarn release:tag
```
m4theushw, line 26:
```suggestion
      - name: yarn l10n --report
```

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: |
+          git remote -v
+          yarn l10n --report
```
m4theushw, line 29: The `GITHUB_TOKEN` envvar is missing.

m4theushw:
> I made the script to make the update when there is a push on master. If you can verify it. I don't know how to test it except merging and hope it works

It's a shot in the dark, we test after merge. You could configure GitHub Actions in your fork. It might not work initially but we can fix.

alexfauquette:
Bot is live 🤖🎉

image

PR #3595, opened 2022-01-11
[core] Add typing to the pre-processors methods

flaviendelangle:
Extracted from #3593 

As it do not impact the users at all, I will automerge 7 days after the review request.
I think other PRs are more worth spending time reviewing :+1:


packages/grid/_modules_/grid/hooks/core/preProcessing/gridPreProcessingApi.ts
```diff
@@ -9,6 +15,37 @@ export enum GridPreProcessingGroup {
...
+  value: GridPreProcessingGroupLookup[P]['value'],
+  context: GridPreProcessingGroupLookup[P] extends { context: any }
+    ? GridPreProcessingGroupLookup[P]['context']
+    : undefined,
+) => GridPreProcessingGroupLookup[P]['value'];
+
+type GridPreProcessorsApplierArg<
+  P extends GridPreProcessingGroup,
+  T extends { value: any },
+> = T extends { context: any } ? [P, T['value'], T['context']] : [P, T['value']];
```
m4theushw, line 43: In https://github.com/mui-org/material-ui-x/pull/3595/files#diff-b1ecb3ac2bbd884ca95d8879603f25ff6f844a259f0ceb6207fb06682cc933f6R37 you changed to accept several params. How do you plan to type if we add a pre-processor that has more than 3 arguments in the `apiRef.current.unstable_applyPreProcessors` call.


```tsx
const canBeReorderedProcessed = apiRef.current.unstable_applyPreProcessors(
  GridPreProcessingGroup.canBeReordered,
  canBeReordered,
  { targetIndex: targetColVisibleIndex },
  hasMovedLeft,
  hasMovedRight
);
```

m4theushw (approved):

packages/grid/_modules_/grid/hooks/core/preProcessing/gridPreProcessingApi.ts
flaviendelangle: This was a hack to make TypeScript accept both 1 "params" and 0 "params" without doing something like `[value, params] = args as any`.

But if it's not clear, maybe the `as any` it better.

PR #3612, opened 2022-01-13
[DataGrid] Prevent commit if `preProcessEditCellProps` resolves with an error

### LINKED ISSUE #3304: [DataGrid] - singleSelect is calling the preProcessEditCellProps twice. The second call has wrong value.
### Duplicates

### Latest version

### Current behavior 😯

singleSelect column type is calling the preProcessEditCellProps twice. The second call has invalid value, which overwrites the new selection value. Probably blur event?

### Expected behavior 🤔

All calls should have the new selected value.

### Steps to reproduce 🕹

Steps:

1. Add a singleSelect type column
2. Use preProcessEditCellProps

Example:
https://codesandbox.io/s/datagridpro-v5-quick-start-forked-pxc4e?file=/src/App.tsx

### Context 🔦

I am trying to validate the new value of singleSelect from server. Should commit the value only if the server responds with 200.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

#30063

m4theushw:
Fixes #3304

Before: https://codesandbox.io/s/datagridpro-v5-quick-start-forked-pxc4e?file=/src/App.tsx

After: https://codesandbox.io/s/datagridpro-v5-quick-start-forked-rzyrl?file=/src/App.tsx

The bug happens because we don't keep a track if the validation is running or not. With `singleSelect`, the `preProcessEditCellProps` is called when the value is changed and also when commiting, which occur at the same time in this component. To fix it, we should have a flag tracking when the validation is running or not. If it's running, we don't commit the value nor call the callback again. To properly implement this, it involves changing the state which, since the state is passed to all edit components as props, could cause a warning if users spread the props directly into the DOM node without sanitizing them. Since we can't release that breaking change, I added temporarily a new prop called `preventCommitWhileValidating` for users to opt-in. Once it's enabled, we update the `isValidating` prop in the editing state before and after calling `preProcessEditCellProps`. During commit we check if it's validating or not. In v6 I want to make that the default behavior and remove the temp prop.

The code is very ugly. There're opportunities to refactor it but it's better to do it after #3219.

_If you open the CodeSandbox above, you may notice that it only updates the Select value after `preProcessEditCellProps` has resolved. That's another PR._


flaviendelangle (commented):
If `preventCommitWhileValidating` is temporary, maybe we could put it under `experimentalFeatures` ?
Or at least explain on the doc that we only add this strange prop temporarily to avoid any breaking change and that it will be removed on v6.

m4theushw:
> If preventCommitWhileValidating is temporary, maybe we could put it under experimentalFeatures ?

@flaviendelangle I moved to `experimentalFeatures`. I had to make other changes to have access to this prop inside the hook. For the row grouping, you created an additional prop besides the one in `experimentalFeatures`.

flaviendelangle (approved):
@m4theushw yes for the row grouping, `disableRowGrouping` made sense even when the feature will be marked as stable.

PR #3618, opened 2022-01-14
[test] Split cell/row editing tests

m4theushw:
Following #3219, I think it would be better to also separate the tests. Although both edit modes share a lot of things, I find easier to see them as two complete different features. Concerning testing, the `editRows.DataGridPro.test.tsx` file has more than 1600 LOC. It's impractical to navigate between all of tests. In this PR I propose to split them into 3 new files:

1. `cellEditing.DataGridPro.test.tsx` is only to test the general functionality of the cell editing.
2. `rowEditing.DataGridPro.test.tsx` is similar to above but only for row editing tests.
3. `editComponents.DataGridPro.test.tsx` is meant to test the edit components provided by each column type. If one component behaves differently in `editMode=row` I would add its tests here.

It's not necessary to go through each file, I'm only moving the tests without changing them. I didn't automerge because I want to collect some feedback about the new structure.


flaviendelangle:
A lot better that way
Could we put the orphan `it` at the end (move `validation` and `control Editing` describe up)

flaviendelangle (approved):

PR #3275, opened 2021-11-24
[DataGrid] Refactor keyboard/click event management

alexfauquette:
Fix #3265 and fix  #3272

This is a first step to handle all the keyboard events in their feature hook https://github.com/mui-org/material-ui-x/issues/3265#issuecomment-977898327

PR #3624, opened 2022-01-14
[DataGrid] Fix input element in custom header

### LINKED ISSUE #3599: [DataGridPro] can't use TextField in header
### Duplicates

### Latest version

### Current behavior 😯

In case you customize the DataGrid header with renderHeader and render a TextField, the TextField does not work, it is impossible to change its value.

### Expected behavior 🤔

It should be possible to change TextField value.

### Steps to reproduce 🕹

Repro: [link to code sandbox](https://codesandbox.io/s/renderheadergrid-material-demo-forked-kk2jb)

Steps:

1. try to write something in the TextField in the header --> not possible

### Context 🔦

I need custom TextFields in the header to perform a custom filtering.

### Your environment 🌎

see code sandbox

### Order ID 💳 (optional)

34462

alexfauquette:
Fix #3599 

The previous test was not catching the error, because it was checking that `event.defaultPrented=false`. But the `onKeyDown` is first run by the input, and then by the header. Such that the `event.preventdefault()` is applied after we observe it.

I tried the `fireEvent.change` but this does not trigger the `keyDown` event. Then I'm adding the library `@testing-library/user-event` to use `userEvent.type` which imitates the input writing.

Now the two tests break when we set a `preventDefault` at the wrong place


packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
```diff
@@ -313,20 +314,44 @@ describe('<DataGrid /> - Keyboard', () => {
...
+      ];
+
+      render(
+        <div style={{ width: 300, height: 300 }}>
+          <DataGrid rows={rows} columns={columns} />
+        </div>,
+      );
+      const input = screen.getByTestId('custom-input');
+      input.focus();
+      userEvent.type(input, 'hello world');
```
m4theushw, line 342: Is not `fireEvent.keyDown` is sufficient? I always challenged myself to not add `@testing-library/user-event`. Although it offers some handy helpers it's rare when we need to fire a full click/keyboard event.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -161,7 +162,10 @@ export const useGridKeyboardNavigation = (
     GridEventListener<GridEvents.columnHeaderNavigationKeyDown>
   >(
     (params, event) => {
-      event.preventDefault();
+      const isNativeFocusOnHeader = event.currentTarget === event.target;
+      if (isNavigationKey(event.key) && isNativeFocusOnHeader) {
+        event.preventDefault();
```
m4theushw, line 167: If the key is not a navigation key shouldn't it do nothing? If the focus is in another element of the column header cell, lets say the sort button, it won't call `event.preventDefault()` but it still tries to navigate to the cell below.

_Click into the sort icon button then press <kbd>Arrow Down</kbd>_

msedge_3YcFi2d6YP

m4theushw (commented):
Regarding the solution, it would be simple if we could check if the event is coming from a custom rendered header. If it does, then we don't need to care about the keyboard event handling. One way is to wrap `headerComponent` below in a div. Then, we know clearly that the event came from the custom component.

https://github.com/mui-org/material-ui-x/blob/7cc6c3c68e139836fec65505254a311534f1e929/packages/grid/_modules_/grid/components/columnHeaders/GridColumnHeaderItem.tsx#L235

packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
alexfauquette: Testing with 
```js
const input = screen.getByTestId('custom-input');
input.focus();
fireEvent.keyDown(input, {
key: "h",
code: "KeyH"
});
expect(input.value).to.equal('h');
```
leads to `AssertionError: expected '' to equal 'h'`

To make the test pass, I can use
```js
const input = screen.getByTestId('custom-input');
input.focus();
fireEvent.keyDown(input, {key: "h", code: "KeyH"});
fireEvent.change(input, { target: { value: "h" } });
expect(input.value).to.equal('h');
```
But this test does not catch the bug.


The last solution I have in mind is to add a listener to the root component to see if the default have been prevented or not. But with this solution we are not really testing if the input has been modified.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: It tooks me time to understand what append

In the PR #3275 that was only refact for column header key down, I missed the following condition
```js
if (!isGridHeaderCellRoot(event.target as HTMLElement)) {
        return;
}
```
https://github.com/mui-org/material-ui-x/pull/3275/files#diff-130975bb0d3117df317939124519a996a0f43f29799b345003e063ea6c1a181aL103-L124

Before the introduction of the bug, the rule was the next one: If the focus is not on the root component, the `columnHeaderKeyDown` event has no effect. One exception is for that checkbox selection which has the focus set in the checkbox input.

I think this PR is a "regression" fix and I will open an issue to discuss how events from custom render should be managed.

m4theushw: Note that `isGridHeaderCellRoot` only checks if `event.target` is the column header. If the target is a child of the header (e.g. the sort icon) it evaluates to false. This creates a regression because if I click the sort icon and I press <kbd>Arrow Down</kbd> it will scroll the page instead of going to the cell below.

> I think this PR is a "regression" fix and I will open an issue to discuss how events from custom render should be managed.

Does this mean that you won't be fixing the column header with a custom component? If yes, could you update the title and description so #3599 doesn't get closed once this one is merged

packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
m4theushw: I took a look into how to test this. You could add an e2e test to ensure that, when an input is rendered as the column header, typing into it will work. For the unit test, you could check if someone called `event.preventDefault` or not. The return value of `fireEvent.keyDown` is `false` when the event default action was cancelled by someone. It should be `true` when dispatched by a custom rendered column header. This is how `@testing-library/user-event` implements `userEvent.type`: https://github.com/testing-library/user-event/blob/a5ca2e47f30ea71b72a443c227ea125beb4e84b7/src/keyboard/keyboardAction.ts#L88

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: Yes, `isGridHeaderCellRoot` is a hard condition, but that is what we used before the regression occurs. 

You can test the codesandbox from the issue with different versions of DataGrid:
https://codesandbox.io/s/renderheadergrid-material-demo-forked-suwsf?file=/index.tsx

- v5.2.0: 
	- TextField works perfectly
	- keyboard navigation is blocked if the focus is on the sort button
- v5.2.1: 
	- TextField broken
	- keyboard navigation is working if the focus is on the sort button

The modification between v5.2.0 and v5.2.1 is not intentional. So I propose to correct the regression I introduced in v5.2.1 to solve the issue.

With the modification added in the [commit](https://github.com/mui-org/material-ui-x/pull/3624/commits/8581bc248942b28839367f0d195735153a9ed58e) the <kbd>Arrow Down</kbd> keep same behavior as v5.2.0

m4theushw: Why not to also return here? We only handle the navigation keys.

```suggestion
        event.preventDefault();
        return;
```

---

> keyboard navigation is working if the focus is on the sort button

I'm not sure about this statement. In https://deploy-preview-3624--material-ui-x.netlify.app/components/data-grid/demo/, if I click the sort icon, then press <kbd>Page Down</kbd>, it moves the page down.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -158,10 +161,21 @@ export const useGridKeyboardNavigation = (
...
+        // There is on exception for the checkBoxHeader
+        return;
+      }
+
+      if (isNavigationKey(event.key)) {
+        event.preventDefault();
+      }
       if (!params.field) {
         return;
       }
```
m4theushw, line 181: To improve readability.

```suggestion

      if (!params.field) {
        return;
      }

```

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -158,10 +161,21 @@ export const useGridKeyboardNavigation = (
...
 
   const handleColumnHeaderKeyDown = React.useCallback<
-    GridEventListener<GridEvents.columnHeaderNavigationKeyDown>
+    GridEventListener<GridEvents.columnHeaderKeyDown>
   >(
     (params, event) => {
-      event.preventDefault();
+      if (
+        !isGridHeaderCellRoot(event.target as HTMLElement) &&
+        !(params.field === GRID_CHECKBOX_SELECTION_COL_DEF.field && isNavigationKey(event.key))
```
m4theushw, line 169: Why to also check if it's a navigation key? We only `event.preventDefault` on the navigation keys.

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: I could put an early return as follow

```js
if (isNavigationKey(event.key)) {
	event.preventDefault();
} else{
	return;
}
```
The page moves down, but not the grid. That's the behavior before the regression. Here is a comparison of key board before/after the regression and the preview of this commit.

From left to righ:
- before regression: https://deploy-preview-3302--material-ui-x.netlify.app/components/data-grid/demo/
- after regresion: https://deploy-preview-3275--material-ui-x.netlify.app/components/data-grid/demo/
- this PR: https://deploy-preview-3624--material-ui-x.netlify.app/components/data-grid/demo/

## When focus on sort
Peek 2022-01-27 12-26

## When focus on column header
Peek 2022-01-27 12-27

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: Yes, there is no reason

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: This condition is useless I added it to apply this PR comment but I miss understood it 😅 https://github.com/mui-org/material-ui-x/pull/3275#discussion_r758851911

packages/grid/_modules_/grid/hooks/features/keyboard/useGridKeyboardNavigation.ts
m4theushw: > I could put an early return as follow

The problem is `isGridHeaderCellRoot` only checking if the event was fired by the root.

Let's review what happened here:

- Before 5.2.1: clicking in any child of the column header and <kbd>ArrowDown</kbd> would scroll the page (nobody complained but I think it's not good)
- 5.2.1: Unintentionally fixed the bug above
- This PR reverts the fix

I see this as inserting a regression because 5.2.1 is already released. Could you take a look in #3692? I think it already fixed part of the problem here. The solution was to only call `event.preventDefault` for navigation keys. #3599 was kinda fixed for free by #3692. However, pressing a navigation key inside the input won't work.

alexfauquette: Yes, I had a look at #3692 It is equivalent to add
```js
if(isNavigationKey(event.key)) {
	event.preventDefault()
}
```

That was my first idea but I stopped doing it because it does not allow you to use keyboard navigation in a custom input, event more problematic, it does not allow you to use space key. The worst I imagine is putting a select input in the header. In such a case, the keyboard navigation will be completely down, because it requires keyboard arrows.

My conclusion is that only the developer knows if the focusable element added requires or not the `event.preventDefault` to prevent the page to scroll. But how to adapt the interface to let them choose, I don't know

m4theushw: What about then wrapping the custom component in a div so we know from where the event is coming from and ignore it? I proposed this in https://github.com/mui-org/material-ui-x/pull/3624#pullrequestreview-853510614. I don't see a simpler and future-proof alternative. The other way, I think, would be to add a specific attribute to each of the icons and the title to allows us to detect if the event came from them or from a custom component.

siriwatknp:

3. run `yarn docs:api`

alexfauquette:
To verify if the event comes from the custom title I added a class `classes.columnHeaderTitleContentContainer` to wrap the title without the buttons.

It's only used to make a query selector, but I let it accessible for developers

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -167,12 +170,31 @@ export const useGridKeyboardNavigation = (
...
+      if (params.colDef.renderHeader) {
+        const headerTitleNode = event.currentTarget.querySelector(
+          `.${gridClasses.columnHeaderTitleContentContainer}`,
+        );
+        isFromCustomHeader =
+          !!headerTitleNode && headerTitleNode.contains(event.target as Node | null);
+      }
+
+      if (
+        !isGridHeaderCellRoot(event.target as HTMLElement) &&
```
m4theushw, line 186: This helper is not used elsewhere. We want to remove it in v6. Could you inline it here?

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -167,12 +170,31 @@ export const useGridKeyboardNavigation = (
   );
 
   const handleColumnHeaderKeyDown = React.useCallback<
-    GridEventListener<GridEvents.columnHeaderNavigationKeyDown>
+    GridEventListener<GridEvents.columnHeaderKeyDown>
   >(
     (params, event) => {
-      if (!params.field) {
+      let isFromCustomHeader = false;
+      if (params.colDef.renderHeader) {
```
m4theushw, line 177: If `columnHeaderTitleContentContainer` is always there we can always ignore events coming from inside it. It's simpler than to also check if `colDef.renderHeader` is available.

packages/grid/x-data-grid/src/internals/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -232,13 +233,15 @@ function GridColumnHeaderItem(props: GridColumnHeaderItemProps) {
...
       >
         <div className={classes.titleContainer}>
-          {headerComponent || (
-            <GridColumnHeaderTitle
-              label={column.headerName ?? column.field}
-              description={column.description}
-              columnWidth={width}
-            />
-          )}
+          <div className={classes.titleContentContainer}>
```
m4theushw, line 236: I would name it as `titleContainerContent` to keep the hierarchy.

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: The idea was to do not search in the HTML tree if the user did not used `renderHeader`. This is an over-optimization (not even sure it optimize something)

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -166,12 +168,28 @@ export const useGridKeyboardNavigation = (
...
   const handleColumnHeaderKeyDown = React.useCallback<
-    GridEventListener<GridEvents.columnHeaderNavigationKeyDown>
+    GridEventListener<GridEvents.columnHeaderKeyDown>
   >(
     (params, event) => {
-      if (!params.field) {
+      const headerTitleNode = event.currentTarget.querySelector(
+        `.${gridClasses.columnHeaderTitleContainerContent}`,
+      );
+      const isFromCustomHeader =
```
m4theushw, line 177: To keep consistency. We're only checking if the target is inside the content, not if `renderCell` was used.

```suggestion
      const isFromInsideContent =
```

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
```diff
@@ -166,12 +168,28 @@ export const useGridKeyboardNavigation = (
...
+        !!headerTitleNode && headerTitleNode.contains(event.target as Node | null);
+      const isGridHeaderCellRoot =
+        event.target != null &&
+        (event.target as HTMLElement).classList.contains(gridClasses.columnHeader);
+
+      if (!isGridHeaderCellRoot && params.field !== GRID_CHECKBOX_SELECTION_COL_DEF.field) {
```
m4theushw, line 183: I can't remember correctly why we need to ignore when the target is not the root but would be possible to simplify this condition to?

```suggestion
      if (isFromInsideContent && params.field !== GRID_CHECKBOX_SELECTION_COL_DEF.field) {
        return;
      }
```

The discussion I found https://github.com/mui/mui-x/pull/3624#discussion_r787804724

m4theushw (commented):
I noted a bug: after clicking the sort icon and I'm stuck and the arrow keys don't work.

msedge_TpzZvOfTOP

alexfauquette:
I'm adding a tes to ensure that clicking on the sort button does not lock the focus

packages/grid/x-data-grid/src/internals/hooks/features/keyboard/useGridKeyboardNavigation.ts
alexfauquette: Yes, I though I should keep it to let buttons be triggered by <kbd>Space</kbd> But they are listenig to <kbd>Enter</kbd> 
 and <kbd>Ctrl</kbd>+<kbd>Enter</kbd> so yes I'm simplifying it

packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
```diff
@@ -365,42 +365,97 @@ describe('<DataGrid /> - Keyboard', () => {
...
+        </div>,
+      );
+      const input = screen.getByTestId('custom-input');
+      fireEvent.mouseUp(input);
+      fireEvent.click(input);
+      input.focus();
+
+      // Verify that the event is not prevented during the bubbling.
+      // fireEvent.keyDown return false if it is the case
+      // For more info, see the related discussion: https://github.com/mui-org/material-ui-x/pull/3624/files#r787767632
```
m4theushw, line 454: The previous link doesn't work.

```suggestion
      // For more info, see the related discussion: https://github.com/mui/mui-x/pull/3624#discussion_r787767632
```

m4theushw (approved):

packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
```diff
@@ -365,42 +365,97 @@ describe('<DataGrid /> - Keyboard', () => {
...
+      );
+      const input = screen.getByTestId('custom-input');
+      fireEvent.mouseUp(input);
+      fireEvent.click(input);
+      input.focus();
+
+      // This does not work with navigation keys.
+      // For now, the workaround for developers is to stop the propagation
+      // But adding input is discouraged, action column or edit mode are better options
+      expect(fireEvent.keyDown(input, { key: 'a' })).to.equal(true);
```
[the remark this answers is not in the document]
alexfauquette: OK, I will open another PR for this one 👍

PR #3647, opened 2022-01-17
[DataGrid] Fix isRowSelectable when paginationMode='server'

### LINKED ISSUE #3614: [DataGrid] isRowSelectable returns an error when using server side selection
### Duplicates

### Latest version

### Current behavior 😯

I'm using server side pagination and I want to keep the selection when I change the page so I did like in the documentation (https://mui.com/components/data-grid/selection/#usage-with-server-side-pagination).
But when I add the isRowSelectable attribute, it returns the error: No row with id _rowId_ found.
Example: https://codesandbox.io/s/controlledselectionserverpaginationgrid-material-demo-forked-o8s12

### Expected behavior 🤔

This function shouldn't return an error even if we have selected rows that are not displayed.

### Steps to reproduce 🕹

Steps:

1. Go to https://codesandbox.io/s/controlledselectionserverpaginationgrid-material-demo-forked-o8s12
2. Select rows
3. Change page

### Context 🔦

I need to keep both functionality enabled and I tryed to do it differently without success.

### Your environment 🌎

<details>

```
  System:
    OS: macOS 12.1
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: Not Found
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Safari: 15.2
  npmPackages:
    @emotion/react: ^11.6.0 => 11.7.1
    @mui/base:  5.0.0-alpha.60
    @mui/icons-material: ^5.1.1 => 5.2.4
    @mui/lab: ^5.0.0-alpha.55 => 5.0.0-alpha.60
    @mui/material: ^5.1.1 => 5.2.4
    @mui/private-theming:  5.2.3
    @mui/styled-engine:  5.2.4
    @mui/styles: ^5.1.1 => 5.2.3
    @mui/system:  5.2.4
    @mui/types:  7.1.0
    @mui/utils:  5.2.3
    @mui/x-data-grid: ^5.0.1 => 5.2.0
    @types/react:  17.0.37
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript:  4.5.4

```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #3614

### DIFF
--- a/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
+++ b/packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
@@ -42,13 +42,19 @@ const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnHeaderPa
       gridPaginatedVisibleSortedGridRowIdsSelector,
     );
 
-    const filteredSelection = React.useMemo(
-      () =>
-        typeof rootProps.isRowSelectable === 'function'
-          ? selection.filter((id) => rootProps.isRowSelectable!(apiRef.current.getRowParams(id)))
-          : selection,
-      [apiRef, rootProps.isRowSelectable, selection],
-    );
+    const filteredSelection = React.useMemo(() => {
+      if (typeof rootProps.isRowSelectable !== 'function') {
+        return selection;
+      }
+
+      return selection.filter((id) => {
+        try {
+          return rootProps.isRowSelectable!(apiRef.current.getRowParams(id));
+        } catch {
+          return false;
+        }
+      });
+    }, [apiRef, rootProps.isRowSelectable, selection]);
 
     // All the rows that could be selected / unselected by toggling this checkbox
     const selectionCandidates = React.useMemo(() => {
--- a/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
+++ b/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import { useForkRef } from '@mui/material/utils';
-import { GridRowId } from '../../../models/gridRows';
 import { useGridApiContext } from '../../utils/useGridApiContext';
 import { useGridRootProps } from '../../utils/useGridRootProps';
 import { useGridSelector } from '../../utils/useGridSelector';
@@ -16,6 +15,7 @@ import { GridEventListener, GridEvents } from '../../../models/events';
 import { useGridApiEventHandler } from '../../utils/useGridApiEventHandler';
 import { clamp } from '../../../utils/utils';
 import { GridRenderContext } from '../../../models';
+import { selectedIdsLookupSelector } from '../selection/gridSelectionSelector';
 
 // Uses binary search to avoid looping through all possible positions
 export function getIndexFromScroll(
@@ -41,7 +41,6 @@ export function getIndexFromScroll(
 
 interface UseGridVirtualScrollerProps {
   ref: React.Ref<HTMLDivElement>;
-  selectionLookup: Record<string, GridRowId>;
   disableVirtualization?: boolean;
   renderZoneMinColumnIndex?: number;
   renderZoneMaxColumnIndex?: number;
@@ -55,7 +54,6 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
 
   const {
     ref,
-    selectionLookup,
     disableVirtualization,
     onRenderZonePositioning,
     renderZoneMinColumnIndex = 0,
@@ -67,6 +65,7 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
   const cellFocus = useGridSelector(apiRef, gridFocusCellSelector);
   const cellTabIndex = useGridSelector(apiRef, gridTabIndexCellSelector);
   const editRowsState = useGridSelector(apiRef, gridEditRowsStateSelector);
+  const selectedRowsLookup = useGridSelector(apiRef, selectedIdsLookupSelector);
   const currentPage = useCurrentPageRows(apiRef, rootProps);
   const renderZoneRef = React.useRef<HTMLDivElement>(null);
   const rootRef = React.useRef<HTMLDivElement>(null);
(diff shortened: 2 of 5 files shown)


packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -272,6 +271,15 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
     for (let i = 0; i < renderedRows.length; i += 1) {
       const { id, model } = renderedRows[i];
 
+      let isSelected: boolean;
+      if (selectedRowsLookup[id] == null) {
+        isSelected = false;
+      } else if (typeof rootProps.isRowSelectable === 'function') {
+        isSelected = rootProps.isRowSelectable(apiRef.current.getRowParams(id));
```
m4theushw, line 278: I'm not sure if this change is good. In the previous way, the selection lookup was already filtered all at once and memoized. Now `isRowSelectable` might be called several times for the same row. I'm OK to call it for every `currentPage.rows` and memoize the result.

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
```diff
@@ -42,13 +42,19 @@ const GridHeaderCheckbox = React.forwardRef<HTMLInputElement, GridColumnHeaderPa
...
-          : selection,
-      [apiRef, rootProps.isRowSelectable, selection],
-    );
+    const filteredSelection = React.useMemo(() => {
+      if (typeof rootProps.isRowSelectable !== 'function') {
+        return selection;
+      }
+
+      return selection.filter((id) => {
+        try {
```
m4theushw, line 51: What do you think about filtering `selection` to keep only the row ids that are visible in the current page. This would remove the need to use a try/catch.

In https://mui.com/components/data-grid/demo/ I can't accumulate ids to the previous selection. Gmail doesn't keep the checkbox state between pages, it always begins unchecked.

msedge_S3Bp2ipmFJ

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -411,6 +411,19 @@ describe('<DataGrid /> - Selection', () => {
       );
       expect(getSelectedRowIds()).to.deep.equal([]);
     });
+
+    it('should not crash when paginationMode="server" and some selected rows are not provided to the grid', () => {
+      render(
```
m4theushw, line 416: `.not.to.throw()`?

https://github.com/mui-org/material-ui-x/blob/f581ac12a3dfc9097f3877b3334a935b5fdbfdac/packages/grid/x-data-grid-pro/src/tests/rows.DataGridPro.test.tsx#L773

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
flaviendelangle: We would loose the logic behind `checkboxSelectionVisibleOnly`
Right now users can decide if they want the select all to work across all pages or only on the current page
(`VisibleOnly` is wrongly named since "visible" doesn't mean "current page" right now though)

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
flaviendelangle: On `master` this code throw
I guess we can just render and it would fail if it throw anyway.
My goal was to be extra-clear here.

m4theushw: I was referring to the error message. By not using `.not.to.throw()`, the test fails because an error was written to the console, not because the test itself failed.

flaviendelangle: Isn't the console error catch by our custom `toErrorDev` instead of `to.throw` ?

m4theushw: Yeah, true. Either `.toErrorDev` or `to.throw` works. My point is more that this test has no assertion and relies on the default behavior of failing if `console.error` is called.

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
flaviendelangle: My thought behind this is that `renderedRows` will never be hundreds of rows long whereas `currentPage.rows` might have 100 000 items.
But I could memoize `renderedRows` and create a memoized selection lookup based on `renderedRows`.
That way we don't loop on the potentially very long `currentPage.rows` and we only re-run when the 1st and/or last  rendered rows changes.

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
flaviendelangle: Yes indeed
If you have a better idea for this kind of test where we just want to avoid a regression with a crash.

m4theushw: Use `.toErrorDev()`? There're other examples in the "console error" describe block in this same file.

```tsx
expect(() => {
  render(
    <TestDataGridSelection
      paginationMode="server"
      selectionModel={[1, 4]}
      isRowSelectable={(params) => params.id > 0}
      checkboxSelection
    />,
  );
  // @ts-expect-error need to migrate helpers to TypeScript
}).not.toErrorDev();
```

flaviendelangle:
```
      expect(() => {
        render(<TestCase initialState={{ pinnedColumns: { left: ['currency'] } }} />);
        render(<TestCase initialState={{ pinnedColumns: { right: ['currency'] } }} />);
      }).not.to.throw();
```

For column pinning you used `not.to.throw`
But OK to prefer `not.toErrorDev`

flaviendelangle: And I understand what you were saying, I thought I did `not.to.throw` but I did neither of them so the test was confusion.
I just added `not.toErrorDev`

packages/grid/_modules_/grid/components/columnSelection/GridHeaderCheckbox.tsx
m4theushw: Ok we can ignore this idea to not break `checkboxSelectionVisibleOnly`. However, I'm not a big fan of using a try/catch. It might hide things. If `isRowSelectable` or `apiRef.current.getRowParams` crashes for some reason the error will be swallowed and an error boundary won't catch it.

flaviendelangle: Could we say in v6 that `getRowParams` return a nullable value instead of crashing ?
That way we can avoid `try / catch`
As this method is called a lot of times, I'm not a fan of adding a check before to know if the row do exist. But that would also by an alternative solution.

m4theushw (approved):

PR #3648, opened 2022-01-17
[DataGrid] Differenciate the pro and community versions of `GridState`, `GridApi` and `GridApiRef`

flaviendelangle:
Part of #924 

## Goal

To reduce the amount of imports from file that should be moved to the `x-data-grid` package to files that should be moved to the `x-data-grid-pro` package.

### Interface based on the package 

If you import `GridApi`, `GridState` or `GridInitialState` from `@mui/x-data-grid`, you will have the community version.
If you import `GridApi`, `GridState` or `GridInitialState` from `@mui/x-data-grid-pro`, you will have the pro version.

Internally we use `GridApiPro` and `GridApiCommunity` while waiting for the bundle split.
Afterwards, we should be able to simplify things.

For all the interfaces depending on the api (for instance `GridRenderCellParams` to type its `api` param or  `GridColDef` to type its `renderCell` params through `GridRenderCellParams`), I added an `Api` generic which extends `GridApiCommon`.
When exported from `@mui/x-data-grid`, the default generic value is the community api, when exported from `@mui/x-data-grid-pro`, the default generic value is the pro api.

### Untyped interfaces

Some use case were to complicated to cover for now.
For instance, the `columns` prop of `DataGrid` has the type `GridColumns`. Otherwise, you could have errors when passing pro columns to the community version because TypeScript consider that the pro and community api are not compatible.

### `createSelector`

The method now ask as a parameter the largest state used in one of its sub-selectors.

For instance the following selector will ask for `GridApiPro` because `gridRowGroupingModelSelector` is a pro selector.

```ts
const mySelector = createSelector(gridPageSizeSelector, gridRowGroupingModelSelector, () => null)
```

**Note:** Everything still works with the `apiRef`, I just simplified the typing to accept a ref which extends `{ instanceId: number, state: S }`

## What's next ?

In v6 we should drop all the `api` parameter to reduce as much as possible the scope of the plan-based typing.
This requires to open the `apiRef` to the community plan (which should be fairly easy after this PR).


scripts/x-data-grid-pro.exports.json
```diff
@@ -330,8 +335,9 @@
   { "name": "gridSortModelSelector", "kind": "Variable" },
   { "name": "GridState", "kind": "Interface" },
   { "name": "GridStateApi", "kind": "Interface" },
-  { "name": "GridStateChangeParams", "kind": "Interface" },
```
flaviendelangle, line 333: Was never used

packages/grid/_modules_/grid/models/api/gridApiRef.ts
```diff
@@ -1,7 +1,19 @@
...
+ * TODO v6: Make the `GridApi` generic mandatory. This interface should only be used when accessing dynamically api methods or state based on the plan.
  */
-export type GridApiRef = React.MutableRefObject<GridApi>;
+export type GridApiRef<GridApi extends GridApiCommon = GridApiCommunity> =
+  React.MutableRefObject<GridApi>;
+
+/**
+ * The reference storing the api of the Community-plan grid.
+ */
+export type GridApiRefCommunity = React.MutableRefObject<GridApiCommunity>;
```
cherniavskii, line 14: This would be an equivalent, right?

```ts
export type GridApiRefCommunity = GridApiRef<GridApiCommunity>;
```

Is there any difference, or is it just matter of preference?

cherniavskii (commented):
Looks great!
I have few questions though

packages/grid/_modules_/grid/hooks/utils/useGridApiContext.ts
```diff
@@ -14,5 +18,5 @@ export function useGridApiContext() {
     );
   }
 
-  return apiRef;
+  return apiRef as any;
```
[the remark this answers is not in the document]
flaviendelangle: Without as `as`, Typescript complains that the state value can have a different type.

But the following works and is cleaner:

```suggestion
  return apiRef as GridApiRef<GridApi>;
```

packages/grid/_modules_/grid/hooks/core/useGridStateInitialization.ts
```diff
@@ -1,33 +1,38 @@
...
+      stateId,
+    };
+  }, []);
+
+  const setState = React.useCallback<GridStateApi<GridApi['state']>['setState']>(
     (state) => {
-      let newState: GridState;
-      if (typeof state === 'function') {
+      let newState: GridApi['state'];
+      if (isFunction(state)) {
```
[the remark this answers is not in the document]
flaviendelangle: `isFunction` tells TypeScript that the returned value is a function

```ts
export function isFunction(value: any): value is Function {
  return typeof value === 'function';
}
```

When using `typeof xxx === 'sometype'`, I don't know why but sometimes TypeScript narrows the typing after the check and sometime not.

And on this particular code, if I keep the `typeof state === 'function'`, TypeScript complains because it still thinkgs that `state` can be of type `GridState` in line 36.

packages/grid/_modules_/grid/hooks/core/useGridStateInitialization.ts
```diff
@@ -105,11 +110,11 @@ export const useGridStateInitialization = (apiRef: GridApiRef, props: DataGridPr
 
   const forceUpdate = React.useCallback(() => rawForceUpdate(() => apiRef.current.state), [apiRef]);
 
-  const stateApi: Omit<GridStateApi, 'state'> = {
+  const stateApi: any = {
```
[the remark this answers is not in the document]
flaviendelangle: TypeScript complains in line 119 when calling `useGridApiMethod`.
I did my best to avoid this `any` and have the Typing exactly work with the various plans whenever possible, but for this line I did not succeed :sob:

packages/grid/_modules_/grid/models/api/gridApiRef.ts
flaviendelangle: When I did this line, `GridApiRef` did not have its generic yet
But you are right, it's better that way :+1:

packages/grid/_modules_/grid/hooks/utils/useGridApiContext.ts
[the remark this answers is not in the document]
cherniavskii: Interesting!

I think this is the reason:
Screenshot 2022-01-26 at 12 46 14
 
 `GridApiContext` uses `GridApiRef` which uses `GridApiCommunity` by default.
 
I've found this solution:

```tsx
export function useGridApiContext<GridApi extends GridApiCommon = GridApiCommunity>() {
  // override GridApiContext type here
  const apiRef = React.useContext(GridApiContext as React.Context<GridApiRef<GridApi> | undefined>);

  if (apiRef === undefined) {
    throw new Error(
      [
        'MUI: Could not find the data grid context.',
        'It looks like you rendered your component outside of a DataGrid or DataGridPro parent component.',
        'This can also happen if you are bundling multiple versions of the data grid.',
      ].join('\n'),
    );
  }

  return apiRef;
}
```

Which returns proper type for returned apiRef:
Screenshot 2022-01-26 at 12 50 43

Not sure if it's any better than your solution.
What do you think?

packages/grid/_modules_/grid/hooks/core/useGridStateInitialization.ts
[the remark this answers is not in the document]
cherniavskii: Well, good to know!

packages/grid/_modules_/grid/hooks/core/useGridStateInitialization.ts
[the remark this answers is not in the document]
cherniavskii: Couldn't figure out why it's like that as well :/

cherniavskii (approved):

packages/grid/_modules_/grid/hooks/utils/useGridApiContext.ts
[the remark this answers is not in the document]
flaviendelangle: It moves the `as` somewhere else, I don't think it's very different in the end :+1:

docs/src/pages/components/data-grid/editing/FullFeaturedCrudGrid.tsx
```diff
@@ -10,14 +10,14 @@ import {
...
   DataGridPro,
-  GridApiRef,
   GridColumns,
   GridRowParams,
   MuiEvent,
   GridToolbarContainer,
   GridActionsCellItem,
   GridEventListener,
   GridEvents,
+  GridApiRefPro,
```
m4theushw, line 20: Ideally it would be nice to keep it named as `GridApiRef` (same for `GridState`) and each package exposes a different interface. However, we can't yet because of the `_modules_` folder. Should we still use `GridApiRef` in the demos and mark `GridApiRefPro` as internal until we remove the modules folder? I guess we don't need the "Pro" suffix once each interface is moved to its correct folder.

flaviendelangle: Right now we have `GridApiRef` which takes the api as a generic and return it's ref version. And `GridApiRefPro` / `GridApiRefCommunity` which are the plan-based apis.
This allow to reduce the typing breaking change.

But if we are looking forward to what would be ideal in v6.  ​Differentiating the two makes it easier to create a different doc page for the free and the pro interface.
If we have the same name, we could still do it (or a single page with a flag "pro only" when the property is only available on pro.

m4theushw: If we add the suffix I assume we will remove `GridApiRef` in the next major, right? If we do, we have the [selectors](https://mui.com/components/data-grid/state/#catalog-of-selectors) page which uses `GridApiRef`. If we want to type them correctly, this page and all the remaining mentions should be updated in future.

> ​Differentiating the two makes it easier to create a different doc page for the free and the pro interface.

I guess each package should have its own docs, so each `GridApi` would have its own URL.

flaviendelangle: Yes my goal was to remove `GridApiRef` in v6 as it is just an alias for `React.MutableRefObject` with a constraint on the generic.

cherniavskii:
I wonder how hard would that be to upgrade from `DataGrid` to `DataGridPro` after this split?
Ideally, upgrade should be as easy as possible.
Maybe we should consider a codemod for this upgrade?

flaviendelangle:
We will have the same problem for all the other interfaces.
With my current approach, you have:

- `GridRenderCellParams` (equivalent to `GridRenderCellParams<GridApiCommon>`), where the params that receive the `api` have an untyped `apiRef.current.state`, untyped `apiRef.current.setState` and no pro-only methods 
- `GridRenderCellParams<GridApiCommunity>`
- `GridRenderCellParams<GridApiPro>` 

A codemode could probably easily replace all the `GridApiCommunity` by `GridApiPro`

If we want the pro / community to be 100% transparent for users, the pro package should overload dozens of interfaces.

m4theushw:
> You totally loose the possibility of using DataGrid + DataGridPro on the same project with a correct typing.

Maybe with module augmentation, instead of global augmentation, it's possible to only target `@mui/x-data-grid-pro`.

> We have 3 possible paths:

I would go with option 3. For example, `@mui/x-data-grid` exports a basic `GridState` then `@mui/x-data-grid-pro` extends it and adds the Pro-only keys. It will be painful but once we do we never have to change again. I wouldn't mind if to do it we had to duplicate a bunch of files. The module augmentation is kind of a temporary solution.

flaviendelangle:
> Maybe with module augmentation, instead of global augmentation, it's possible to only target @mui/x-data-grid-pro.

I don't think so unfortunately.
I can't augment the interface from `@mui/x-data-grid-pro` from `@mui/x-data-grid-pro`, or I least I did not succeed at doing it.

> The module augmentation is kind of a temporary solution, although it works I wouldn't use it for everything. The GridApiRef should be correctly typed without module augmentation.

Not following you here
If I augment `GridState` for instance, `GridApiRef['current']['state']` will automatically have the pro keys.
I can't cherry pick some interfaces.

____

One downside of the model augmentation is that is totally blocks the hook-only api typing. Users won't be able to set a typing with less features than there module (for instance a `DataGrid` without any filtering).
But as hook-only api won't be a thing before at least v7, I don't think it's a problem.

m4theushw:
> Not following you here

I'm considering that we would remove `api` from the params interface in v6. If we don't do this, I don't feel right asking Pro-users to use module augmentation to get the correct types. I would use it only to type the `api` in `GridRenderCellParams`.

> For option 3, would you go with the following ?

Yes, exactly like that. Once we have the internal folder we could even improve it by keeping the generic `GridRenderCellParams` in that folder and only exporting a `GridRenderCellParams` without the generic API argument.

flaviendelangle:
> Yes, exactly like that. Once we have the internal folder we could even improve it by keeping the generic GridRenderCellParams in that folder and only exporting a GridRenderCellParams without the generic API argument.

Yes :+1:

flaviendelangle:
@m4theushw I have an implementation passing all the CI steps.
It is very verbose and I would be in favor of removing the `api` param whenever possible in v6 to remove the amount of interfaces and method impacted by the plan.

There are some cleaning to do on the doc generation, but I want to do the bundle split before because half of my work is going to be dropped after the split.

docs/scripts/api/buildInterfacesDocumentation.ts
```diff
@@ -33,9 +33,7 @@ interface ParsedProperty {
   typeStr: string;
 }
 
-const INTERFACES_WITH_DEDICATED_PAGES = [
-  // apiRef
-  'GridApi',
```
m4theushw, line 38: Could it point to `GridApiPro` to prevent the 404 in https://deploy-preview-3648--material-ui-x.netlify.app/api/data-grid/grid-api/?

flaviendelangle: I can add a redirection
And in the future we will indeed have to improve the doc to handle those renamed interfaces across plans.

flaviendelangle:
I am merging this one to be able to move on without blocking the releases for to long.

PR #3504, opened 2021-12-25
[docs] Lint markdown in the CI

oliviertassinari:
Same as https://github.com/mui-org/material-ui/pull/30395

PR #3651, opened 2022-01-17
[core] Add language to markdown code block

m4theushw:
#3527 was created before #3504 so it was merged without running `yarn markdownlint` which broke the CI.


oliviertassinari:
Arf, thanks

Issue #3377, opened 2021-12-08
[data grid] Implement server-side data source in tree data


flaviendelangle:
Part of #2994 

### Add new props to the grid

Add two properties to lazy load the children of a row.

```
hasChildrenOnServer?: (row: GridRowModel) => boolean;
fetchRowChildren?: (row: GridRowModel) => Promise<GridRowModel[]>;
```

A simplified implementation would look something like:

```
const setRowChildrenExpansion = async (id, isExpanded) => {
  const row = apiRef.current.getRow(id);
  if (isExpanded && props.fetchRowChildren) {
    const shouldFetch = prop.hasChildrenOnServer ? prop.hasChildrenOnServer(row) : true;
    if (shouldFetch) {
      const newRows = await prop.fetchRowChildren(row);
      apiRef.current.updateRows(newRows);
    }
  }
  // ... rest of the current behavior
}
```

We could add a loading animation to improve the UX.

### Let user implement it

Let the user handle it with `GridEvents.rowExpansionChange` and just add an example to the doc.
It wont scale with future optimizations but could be enough for the time being.

https://codesandbox.io/s/datagridpro-v5-quick-start-forked-ncklk?file=/src/App.tsx

### Do not reset the whole state when adding new rows with `updateRows`

Whatever the solution we pick, we are limited by the fact the when we add a row, we re-generate the whole tree and therefore loose the expansion statuses.

A first important step would be to handle correctly partial tree updates.

mbiggs-gresham:
I would love to see this feature implemented. We currently are doing an infinite scroll grid in the pro version and have just seen the addition of the tree data and was wondering how we could make the two work together. 

Would you envision it would be possible to make it so that not ALL children have to be fetched on the expansion. We are likely to see some very large numbers of child rows, so i'm just wondering if we could somehow make it so that we can retrieve the child rows as part of the infinite scroll.

flaviendelangle:
In #4851 we are talking about modifications on the internal state structure to be able to do lazy loading properly.

Bindeep:
Is there any estimated time when this will be added?

flaviendelangle:
We will need to make breaking changes to implement this one correctly
So we need to wait for the next major (v6)
I can't give an exact time, but Q4 2022 or Q1 2023 is likely

bclements:
@joserodolfofreitas  Premium customer here. We need the lazy loading functionality asap. Developing a new application so happy to utilize something that has a breaking change in it.

[...]

PR #3657, opened 2022-01-18
[DataGridPro] Add doc example for tree data children lazy loading

flaviendelangle:
Part of  #3377 
[Doc preview](https://deploy-preview-3657--material-ui-x.netlify.app/components/data-grid/group-pivot/#children-lazy-loading)

### DIFF
--- a/docs/src/pages/components/data-grid/group-pivot/TreeDataLazyLoading.js
+++ b/docs/src/pages/components/data-grid/group-pivot/TreeDataLazyLoading.js
@@ -1,0 +1,355 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import {
+  DataGridPro,
+  getDataGridUtilityClass,
+  GridEvents,
+  useGridApiContext,
+  useGridApiRef,
+  useGridRootProps,
+} from '@mui/x-data-grid-pro';
+import { unstable_composeClasses as composeClasses } from '@mui/base';
+import Box from '@mui/material/Box';
+import IconButton from '@mui/material/IconButton';
+
+export const isNavigationKey = (key) =>
+  key === 'Home' ||
+  key === 'End' ||
+  key.indexOf('Arrow') === 0 ||
+  key.indexOf('Page') === 0 ||
+  key === ' ';
+
+const ALL_ROWS = [
+  {
+    hierarchy: ['Sarah'],
+    jobTitle: 'Head of Human Resources',
+    recruitmentDate: new Date(2020, 8, 12),
+    id: 0,
+  },
+  {
+    hierarchy: ['Thomas'],
+    jobTitle: 'Head of Sales',
+    recruitmentDate: new Date(2017, 3, 4),
+    id: 1,
+  },
+  {
+    hierarchy: ['Thomas', 'Robert'],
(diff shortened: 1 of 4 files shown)


docs/src/pages/components/data-grid/group-pivot/group-pivot.md
```diff
@@ -124,6 +124,17 @@ You can limit the sorting to the top level rows with the `disableChildrenSorting
 > const invalidRows = [{ path: ['A'] }, { path: ['B'] }, { path: ['A', 'A'] }];
 > ```
 
+### Children lazy-loading
+
+> ⚠️ This feature isn't implemented yet. It's coming.
+>
+> 👍 Upvote [issue #3377](https://github.com/mui-org/material-ui-x/issues/3377) if you want to see it land faster.
+
+A temporary version can be built outside the grid as shown bellow.
```
joserodolfofreitas, line 133:
```suggestion
Alternatively, you can achieve a similar behavior by implementing this feature outside the component as shown bellow.
```

joserodolfofreitas (approved):
Looking good :+1:

docs/src/pages/components/data-grid/group-pivot/TreeDataLazyLoading.js
```diff
@@ -0,0 +1,359 @@
...
+
+  const slots = {
+    root: ['treeDataGroupingCell'],
+    toggle: ['treeDataGroupingCellToggle'],
+  };
+
+  return composeClasses(slots, getDataGridUtilityClass, classes);
+};
+
+const GroupingCellWithLazyLoading = (props) => {
```
alexfauquette, line 161: This is already a complex component. Could add a comment to say that it is not the main one, and explain briefly what it does. For example

This component override the Grouping cell to show the expand/collapse icon according to the `descendantCount` property instead of the `descendants` property

docs/src/pages/components/data-grid/group-pivot/TreeDataLazyLoading.js
```diff
@@ -0,0 +1,359 @@
...
+};
+
+export default function TreeDataLazyLoading() {
+  const apiRef = useGridApiRef();
+  const [rows, setRows] = React.useState([]);
+
+  React.useEffect(() => {
+    fakeDataFetcher().then(setRows);
+
+    const handleRowExpansionChange = async (node) => {
```
alexfauquette, line 292: By curiosity, why putting it in a `useEffect` instead of a `useCallback` ?

alexfauquette (approved):
Looks good.

You could add one or two sentences before the demo to say: 
- we add a property `descendantCount` to rows, to know which group cell can be extended
- we override both `<GouppingCell />` and `handleRowExpansionChange` to take this new property into account

I should simplify the reading, but I'm not sure it is needed

docs/src/pages/components/data-grid/group-pivot/TreeDataLazyLoading.js
flaviendelangle: This `useEffect` is only called once to initialize the event listeners and init the rows.

PR #3666, opened 2022-01-19
[DataGridPro] Remove function overloading for `useGridApiRef`

flaviendelangle:
Part of #3208 
Part of #3648 

The goal here is to have a hook with a single interface. That way I can override the interface based on the package that is exporting it.

- [x] Remove the format `useGridApiRef(apiRef)`; the hook now only create the ref but don't initialize it

- [x] Stop using the logger for the api initialization. I would like to drop this logger that nobody seems to use. But at 1st I removed it from `useGridApiInitialization` and its dependencies to be able to call `useGridApiInitialization` before `useGridLoggerFactory` and init the `apiRef` in it.

- [x] Init 100% of the `apiRef` inside the `useGridApiInitialization`


alexfauquette (approved):

xiaoyu-tamu:
@flaviendelangle Could you guild me how to migrate from `const apiRef = useGridApiRef(customApiRef);`? 

Which hook should I use if I want to pass apiRef from parent component?

flaviendelangle:
[Here is a working example](https://codesandbox.io/s/datagridpro-v5-quick-start-forked-468ct?file=/src/App.tsx)

This part was meant to be private, sorry if it was unclear and we removed it during a minor.

PR #3671, opened 2022-01-20
[DataGrid] Improve the invalid sortModel and filterModel warnings

flaviendelangle:


packages/grid/_modules_/grid/utils/warning.ts
```diff
@@ -0,0 +1,26 @@
...
+export const wrapWithWarningOnCall = <Args extends any[], R extends any>(
+  method: (...args: Args) => R,
+  message: string | string[],
+) => {
+  if (process.env.NODE_ENV === 'production') {
+    return method;
+  }
+
+  return (...args: Args) => {
+    buildWarning(message);
```
alexfauquette, line 23: Is it dispatching only one warning? I thought the build warning should be outside of the returned method, as follow
 
```js
const warning = buildWarning(message);
return (...args: Args) => {
    warning();
    ...
}
```

flaviendelangle: Oh you are right...

m4theushw (commented):
If you want to throw an error, you can catch it in the test by wrapping the component in an error boundary. You also need to mock `console.error` to prevent React from logging it.

The prop-types could still be used but it would be better to migrate them to a different file to able to test them with `PropTypes.checkPropTypes`.

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -118,3 +119,40 @@ export const buildAggregatedFilterApplier = (
...
+      };
+    }
+
+    const hasItemsWithoutIds = model.items.find((item) => item.id == null);
+    if (hasItemsWithoutIds) {
+      filterModelMissingItemIdWarning();
+
+      return {
+        ...model,
+        items: [model.items[0]],
```
alexfauquette, line 152: You could either let items as they are (without id) or generate ids such as `item.id = "youForgetToSetThisId-1"` for example.
The solution before this commit was to only send a warning

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
```diff
@@ -57,17 +49,13 @@ export const useGridFilter = (
...
+      props.filterModel ?? props.initialState?.filter?.filterModel ?? getDefaultGridFilterModel();
 
     return {
       ...state,
       filter: {
-        filterModel:
-          props.filterModel ??
-          props.initialState?.filter?.filterModel ??
-          getDefaultGridFilterModel(),
+        filterModel: checkFilterModelValidity(filterModel, props.disableMultipleColumnsFiltering),
```
alexfauquette, line 58: It could be renamed `validatedFilterModel` because now this function dispatch the warnings and returned a cleaned model

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
alexfauquette: I just realized missing id are filled in `cleanFilterItem`

https://github.com/alexfauquette/material-ui-x/blob/master/packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts#L25:L27

flaviendelangle: Yes when using `upsertFilterItem` or `showFilterPanel` we allow the input item to lack the id.
But when the model is created by the user and passed as a prop or directly to `setFilterModel`, the id is required.

alexfauquette: Why not using the same behavior in `setFilterModel`? 

Like for the `key` prop in react to render an array. As a developer, if I don't set `key` I expose myself to bugs and I see warnings in the console. But in most cases, from a user perspective, my code runs as expected. Here the code says set multiple filters, and the user will only see one filter

flaviendelangle: Because when controlled we want the state to reflect what the user gives, no matter how invalid it is.
We could revert the logic and do the cleaning in a selector like we do for the column grouping.

alexfauquette: If the state must reflect what the developer provides, it would make sense to have
```diff
-return {
- 	...model,
-     items: [model.items[0]],
- };
+return model;
```
About adding the ids in the selector, I don't see how you could do it in such a way that it would be consistent between two calls

siriwatknp:

3. run `yarn docs:api`

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: You are right, I mixed several things...
It would be nice to have what I described but it clearly requires some deeper changes.
We could add the id here indeed.

packages/grid/_modules_/grid/hooks/features/filter/useGridFilter.ts
flaviendelangle: I went for `sanitizeSortModel` and `sanitizeFilterModel`

alexfauquette (approved):
LGTM
Nice to have all the consistency checks at the same place :)

flaviendelangle:
> The prop-types could still be used but it would be better to migrate them to a different file to able to test them with PropTypes.checkPropTypes. I had to remove PropTypes.checkPropTypes in https://github.com/mui-org/material-ui-x/pull/2395#discussion_r702289523.

@m4theushw To avoid handling to many subjects on this project, I can revert the proptype change.
It is very incomplete because it's not testing the `initialState` prop at all.
But we could probably improve it rather than dropping it.

packages/grid/_modules_/grid/hooks/features/sorting/gridSortingUtils.ts
```diff
@@ -22,13 +23,27 @@ interface GridParsedSortItem {
   getSortCellParams: (id: GridRowId) => GridSortCellParams;
 }
 
+const sortModelDisableMultiColumnsSortingWarning = buildWarning([
+  'MUI: The `sortModel` can only contain a single item when `prop.disableMultipleColumnsSorting` is set to `true`.',
+  'If you are using the `DataGrid` community version, this property is always `true`.',
```
m4theushw, line 28:
```suggestion
  'MUI: The `sortModel` can only contain a single item when the `disableMultipleColumnsSorting` prop is set to `true`.',
  'If you are using the community version of the `DataGrid`, this prop is always `true`.',
```

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -54,6 +37,70 @@ export const cleanFilterItem = (item: GridFilterItem, apiRef: GridApiRef) => {
   return cleanItem;
 };
 
+const filterModelDisableMultiColumnsFilteringWarning = buildWarning([
+  'MUI: The `filterModel` can only contain a single item when `prop.disableMultipleColumnsFiltering` is set to `true`.',
+  'If you are using the `DataGrid` community version, this property is always `true`.',
```
m4theushw, line 42:
```suggestion
  'MUI: The `filterModel` can only contain a single item when the `disableMultipleColumnsFiltering` prop is set to `true`.',
  'If you are using the community version of the `DataGrid`, this prop is always `true`.',
```

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -54,6 +37,70 @@ export const cleanFilterItem = (item: GridFilterItem, apiRef: GridApiRef) => {
...
+      if (item.operatorValue == null) {
+        hasItemWithoutOperator = true;
+      }
+    }
+
+    if (hasItemsWithoutIds) {
+      filterModelMissingItemIdWarning();
+    }
+
+    if (hasItemWithoutOperator || hasItemsWithoutIds) {
```
m4theushw, line 83: Should we make `operatorValue` required in v6? If yes, then makes sense to also add a warning for it.

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -54,6 +37,70 @@ export const cleanFilterItem = (item: GridFilterItem, apiRef: GridApiRef) => {
...
+      return {
+        ...model,
+        items: [model.items[0]],
+      };
+    }
+
+    let hasItemsWithoutIds = false;
+    let hasItemWithoutOperator = false;
+
+    for (let i = 0; i < model.items.length; i += 1) {
```
m4theushw, line 67: If performance is one concern we could stop iterating once `hasItemsWithoutIds` or `hasItemWithoutOperator` is true. Otherwise I wouldn't  mind if we used `Array.some` to compute theses variables. In practice there won't be more than a few filters applied at the same time.

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -54,6 +37,70 @@ export const cleanFilterItem = (item: GridFilterItem, apiRef: GridApiRef) => {
...
+const filterModelMissingItemIdWarning = buildWarning(
+  "MUI: The 'id' field is required on `filterModel.items` when you use multiple filters.",
+);
+
+export const sanitizeFilterModel = (
+  model: GridFilterModel,
+  disableMultipleColumnsFiltering: boolean,
+  apiRef: GridApiRef,
+) => {
+  if (model.items.length > 1) {
```
m4theushw, line 54: We're not setting the default `operatorValue` when `filterModel` contains only 1 item. Actually this causes a bug which could be solved here: https://codesandbox.io/s/controlledfilters-material-demo-forked-u1utu?file=/demo.tsx

image

packages/grid/x-data-grid/src/tests/sorting.DataGrid.test.tsx
```diff
@@ -348,7 +348,10 @@ describe('<DataGrid /> - Sorting', () => {
   });
 
   describe('prop: initialState.sorting', () => {
-    const Test = (props: Omit<DataGridProps, 'rows' | 'columns'>) => (
+    const Test = (
+      props: Omit<DataGridProps, 'rows' | 'columns'> &
+        Partial<Pick<DataGridProps, 'rows' | 'columns'>>,
```
m4theushw, line 353: How about?

```suggestion
      props: Partial<DataGridProps>,
```

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: I would say yes, I don't like when we are doing magic like adding a default operator.

@alexfauquette if you have a thought on this

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: Yes performance is not a concern at all here, I don't imagine a grid with like 10 000 filter items.

packages/grid/x-data-grid/src/tests/sorting.DataGrid.test.tsx
```diff
@@ -450,5 +453,51 @@ describe('<DataGrid /> - Sorting', () => {
...
+                    sort: 'desc',
+                  },
+                ],
+              },
+            }}
+          />,
+        );
+
+        expect(getColumnValues(0)).to.deep.equal(['2', '0', '1', '3']);
+      }).toWarnDev(
```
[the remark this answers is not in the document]
flaviendelangle: Why do we wanter `console.error` + `.toErrorDev` instead of `console.warn` + `.toWarnDev` ?

I'm making the filtering test pass by removing the proptypes check.

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
alexfauquette: Yes, we could make it required.

If it does not cause a problem, I would be in favor of keeping the default operator, just to save the ass of developers that miss use the grid in production

flaviendelangle: As I see it, it would be like the ID

- The typing says it's required
- If none is provided, we fire a warning and provide a default behavior

packages/grid/x-data-grid/src/tests/sorting.DataGrid.test.tsx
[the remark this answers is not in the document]
m4theushw: `console.error` is more noticeable than a warning. Also, users could have disabled warnings to reduce noise in the DevTools console but kept errors on to get React errors.

image

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
alexfauquette: I added a line in Preparing v6 such that we don't forget it

flaviendelangle: I added a warning + doc line + TODO

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: Indeed, fixed

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -53,6 +36,82 @@ export const cleanFilterItem = (
...
+  'error',
+);
+
+const filterModelMissingItemIdWarning = buildWarning(
+  "MUI: The 'id' field is required on `filterModel.items` when you use multiple filters.",
+  'error',
+);
+
+const filterModelMissingItemOperatorWarning = buildWarning([
+  'MUI: One of your filtering item have no `operatorValue` provided.',
```
m4theushw, line 53: Why it uses `console.warn` while `filterModelMissingItemIdWarning` above outputs an error?

```suggestion
  'MUI: One of the filtering items has no `operatorValue` provided.',
```

m4theushw (approved):

packages/grid/_modules_/grid/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: https://github.com/mui/mui-x/pull/3671#discussion_r806837263

I kept a warning here out of consistency with the other deprecation messages.
But I can replace it with an error.

It's a param of `buildWarning` to choose the severity of the message

PR #3674, opened 2022-01-20
[core] Add eslint rule to force default export equals to filename in doc

alexfauquette:
This should never append again after https://github.com/mui-org/material-ui-x/pull/3564#discussion_r788753058

Eslint-rule


flaviendelangle (approved):

.eslintrc.js
```diff
@@ -59,5 +59,11 @@ module.exports = {
       },
       parserOptions: { tsconfigRootDir: __dirname, project: ['./tsconfig.json'] },
     },
+    {
+      files: ['docs/src/pages/**/*.js', 'docs/src/pages/**/*.tsx'],
+      rules: {
+        'filenames/match-exported': ['error', [null, 'kebab', 'snake']],
```
m4theushw, line 65: Keeping `kebab` and `snake` makes it to still suggest other names, although they can't be used because of another rule:

image

Maybe?

```suggestion
      files: ['docs/src/pages/components/**/*.js', 'docs/src/pages/components/**/*.tsx'],
      rules: {
        'filenames/match-exported': ['error', [null]],
```

PR #3678, opened 2022-01-20
[DataGrid] csCZ and skSK locale update

Haaxor1689:
Updated csCZ and skSK locales with new keys

### DIFF
--- a/packages/grid/_modules_/grid/locales/csCZ.ts
+++ b/packages/grid/_modules_/grid/locales/csCZ.ts
@@ -116,10 +116,13 @@ const csCZKGrid: Partial<GridLocaleText> = {
   footerTotalVisibleRows: (visibleCount, totalCount) => {
     const str = totalCount.toLocaleString();
     const firstDigit = str[0];
-    const op = ['4','6','7'].indexOf(firstDigit) >= 0 || (firstDigit === '1' && str.length % 3 === 0) ? 'ze' : 'z';
-    return `${visibleCount.toLocaleString()} ${op} ${str}`
+    const op =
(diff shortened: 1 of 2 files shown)


alexfauquette:
Thanks for the locale update. Don't care about the prettier, I will do it

packages/grid/_modules_/grid/locales/csCZ.ts
```diff
@@ -112,8 +113,15 @@ const csCZKGrid: Partial<GridLocaleText> = {
   footerTotalRows: 'Celkem řádků:',
 
   // Total visible rows footer text
-  // footerTotalVisibleRows: (visibleCount, totalCount) =>
-  //   `${visibleCount.toLocaleString()} of ${totalCount.toLocaleString()}`,
+  footerTotalVisibleRows: (visibleCount, totalCount) => {
+    const str = totalCount.toLocaleString();
+    const firstDigit = str[0];
+    const op =
+      ['4', '6', '7'].includes(firstDigit) || (firstDigit === '1' && str.length % 3 === 0)
```
alexfauquette, line 120: I replce the `indexOf(firstDigit) >= 0` by `.includes(firstDigit)` to simplify the reading
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

@Haaxor1689 I have a doubt on `.toLocaleString()` I think, you should use `.toString()` because the first one can add some spaces, commas, or dots to format the number.

```js
1000000.num.toLocaleString("cz-CZ");
// returns 1,000,000

1000000.num.toString();
// returns 1000000
```

Haaxor1689: Yeah that's a good point, I'll change it to use `toString`.

alexfauquette (approved):
Thanks again for your contribution :)

PR #3680, opened 2022-01-20
[core] Improve GitHub label workflows

DanailH:
This change is meant to improve the support workflow. We have:

1. Renamed `support` to `support: StackOverflow` to make it clear about what this is about. It's about fair value sharing. We give MIT components away for free. In exchange, MUI expects developers to give back by contributing. The expectation is not for maintainers to spend even more time to do others' work (when we apply this label), but for maintainers to process new contributions.
2. Renamed `question` to `support: question`. This is similar to the above, but we find the question so *valuable* to answer that we take it in.
3. Add a new `support: commercial` label. It's unlikely that the community will provide a lot of help to each other on the MUI X Pro plan, and will clearly not happen on the MUI X Premium plan. We have to take this in. However, bug reports and features requests should go to the usual channels, so we are simplifying the issue template.

### DIFF
--- a/.github/workflows/support-stackoverflow.yml
+++ b/.github/workflows/support-stackoverflow.yml
@@ -1,5 +1,5 @@
 # Configuration for support-requests - https://github.com/dessant/support-requests
-name: 'Support Requests'
+name: 'Support StackOverflow'
 
 on:
   issues:
@@ -22,10 +22,10 @@ jobs:
           issue-comment: |
             👋 Thanks for using MUI X!
 
-            We use GitHub issues exclusively as a bug and feature requests tracker, however,
+            We use GitHub issues as a bug and feature requests tracker, however,
             this issue appears to be a support request.
 
-            For support, please check out https://mui.com/getting-started/support/. Thanks!
+            For support, please check out https://mui.com/components/data-grid/getting-started/#support. Thanks!
 
             If you have a question on StackOverflow, you are welcome to link to it here, it might help others.
             If your issue is subsequently confirmed as a bug,  and the report follows the issue template, it can be reopened.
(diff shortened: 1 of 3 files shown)


flaviendelangle (approved):

oliviertassinari (commented):
I would propose we keep this bot, and restructure the labels at the same time

1. `support - stackoverflow`. We rename `support` to this in the X and Core repositories. We don't provide support on MIT content. We delegate it to the community, the bot is important to save us time. With the date picker components, we will use it more often. The same way the support questions on DataGrid should use it.
2. `support: question`. We rename `question` to this in the X and Core repositories. This are so users can ask **valuable** question. Valuable is key on this one. Usually, such question turns into a bug, a docs, a discussion, an enhancement, or simply stay here close to build a knowledge base on GitHub. No bots
3. `support - commercial`. This will be used by all the issues open with https://github.com/mui-org/material-ui-x/blob/master/.github/ISSUE_TEMPLATE/3.pro-support.yml

DanailH:
I've updated this PR to handle the change of the labels, so `support` -> `support: StackOverflow` and we keep the bot. For the other `support` label we can think of the automation later.

oliviertassinari (approved):
I have updated the description and changed a few more things

flaviendelangle (approved):

.github/workflows/support-stackoverflow.yml
```diff
@@ -16,16 +16,16 @@ jobs:
...
           # to a support page, or set to `false` to disable
           issue-comment: |
             👋 Thanks for using MUI X!
 
-            We use GitHub issues exclusively as a bug and feature requests tracker, however,
+            We use GitHub issues as a bug and feature requests tracker, however,
             this issue appears to be a support request.
 
-            For support, please check out https://mui.com/getting-started/support/. Thanks!
+            For support, please check out https://mui.com/components/data-grid/getting-started/#support. Thanks!
```
oliviertassinari, line 28: Later on, to move to: https://mui.com/x/getting-started/support/

Issue #3608, opened 2022-01-12
Provide bundle format that is usable via CDN


Tiberriver256:
### Summary 💡

I would love to have a bundle format of this library we could use via CDN and declare as a webpack external module similar to what `@material-ui/core` and `@mui/material` has here:

https://github.com/mui-org/material-ui/blob/b9117f2ff7fa99d85dd66b5c240dcb32e833a1c9/packages/mui-material/scripts/rollup.config.js#L175-L214

### Motivation 🔦

We are building a component library using React Mui and X Pro that will be used in single spa. In a micro-frontend framework it's going to be important for the performance of the site to be able to have the component library and all it's dependencies available as an external module.

### Order ID 💳 (optional)

#34712

flaviendelangle:
I plan to migrate to the core bundling infrastructure in the following months.
But unfortunately it requires a big internal change which takes a lot of time.

Tiberriver256:
Great to hear it's on the roadmap. Thank you!

oliviertassinari:
IMHO we should fix this issue as "won't fix". The use case described is typically what developers should build and optimize on their side. It's footgun otherwise. 

I have updated https://github.com/mui-org/material-ui/issues/30660 with this opportunity. It links another instance of issue asking for UMD that we closed as won't fix.

Tiberriver256:
Why not match what you're doing with @mui/material? Even the issue you tagged has mention of prepping it for consumption by a CDN.

It's not making sense for me that you would say 'All modules are usable via cdn, except for this one. This one you have to re-bundle yourself.' or am I missing something?

oliviertassinari:
See https://github.com/mui-org/material-ui/issues/23098#issuecomment-710484703. What @mui/material is a 7 year old decision, that is no longer relevant for what we should build for the next 3 years, AFAIK.

Tiberriver256:
Ah I see, perhaps I should edit this issue to be a more generic request to be able to use this module via CDN? Whether it's UMD or ES Modules doesn't really matter.

flaviendelangle:
Putting aside the discussion about the UMD specific format
I started working on the bundling structure this week, you should have the same formats as the packages hosted on https://github.com/mui-org/material-ui by the end of the month :crossed_fingers:

Zenoo:
Is there any easy way to use `@mui/x-data-grid` in a CDN deployed MUI website in the meantime?  
[...]

PR #3451, opened 2021-12-17
[DatePicker] Initial commit

flaviendelangle:
Closes #2398

[Doc preview](https://deploy-preview-3451--material-ui-x.netlify.app/x/react-date-pickers/getting-started/)
[Codesandbox pro example](https://codesandbox.io/s/datagrid-v5-quick-start-forked-5kwtg7?file=/src/App.tsx)

## Changes

For the view types, I removed the duplicates and only kept the type with the name of the view component, not the ones with the name of the pickers (`clock` not `time`, `calendar` not `date`, `calendarOrClock` not `dateTime`)

- ~~`ClockView`~~, ~~`TimePickerView`~~, `ClockPickerView` 
- ~~`DatePickerView`~~, `CalendarPickerView`
- ~~`DateTimePickerView`~~, `CalendarOrClockPickerView`

Removing `ClockPickerStandalone` and instead have `ClockPicker` handle the controlled / uncontrolled pattern just like in `CalendarPicker`

- [x] Add `root` slot for `ClockPicker`

- [x] Rename `PickersCalendar` into `DayPicker`

## PRs ready to be picked:
- https://github.com/mui-org/material-ui/pull/30514
- https://github.com/mui-org/material-ui/pull/30749 => DONE
- https://github.com/mui-org/material-ui/pull/30773 => DONE
- https://github.com/mui-org/material-ui/pull/30768 => DONE
- https://github.com/mui-org/material-ui/pull/30738 => DONE
- https://github.com/mui-org/material-ui/pull/29071
- https://github.com/mui-org/material-ui/pull/30786 => DONE
- https://github.com/mui/material-ui/pull/30561 => DONE

Issue #2398, opened 2021-08-21
[RFC] Migrate date pickers from the lab to MUI X


oliviertassinari:
This issue is meant to give a space to discuss moving the date picker components from the lab to MUI X.
I propose that once we assign a full-time product & engineering owner on them, we do the move, and expose the components as:

- `@mui/x-pickers`
- `@mui/x-pickers-pro`

The component is currently a shared responsibility of the core team, where Sebastian has lately been the one spending the most time on them. 

Advantages:
- If we continue with the date range picker inside the Pro plan, it seems simpler to isolate the whole codebase here, to keep the core fully MIT.
- The core has been successful up until now without these date pickers components. Moving them here would allow the core team to focus on the DX and the design dimensions on the core. From what I understand, it's what is ultimately better aligned with the success of the product this team focuses on
- It would allow giving more development bandwidth to the date pickers. It feels like there is a significant backlog of problems that we could focus on, but it's currently hard to balance time between the date picker and the other core components.

Disadvantages:

- One could argue that the last time we moved the components from *mui-org/material-ui-pickers* to *mui-org/material-ui" it was a mess. We ended up having to update a lot of the code of the component. I personally that this was great. It forced us to make the picker components consistent with the other ones. While the same applies here? Yes, likely. I do expect that the migration won't be very smooth. However, what I hope is the opposite outcome as in the previous migration. Meaning, we won't touch the component but it will force us to make sure MUI X has the same development environment as the main repository.
- One could argue that the core should own any component that has a native counterpart. Why? I didn't follow the reasoning so I would leave this to who wants to defend the argument.
Instead, I can argue why the opposite might work better: Bootstrap is successful with few components. Focusing on too many components in the core feels risky, it could prevent the team to focus on what has more impact: the DX for the most common use cases, the design appeal, and the ease of customizations.

oliviertassinari:
We are moving forward with this RFC. We are in the process of moving the pickers to `@mui/x-pickers` and the date range picker to `@mui/x-pickers-pro`. @flaviendelangle is leading this initiative.

flaviendelangle:
In `Option 1`, I made a few changes (see PR description).

I think we can wait 2-3 weeks before releasing an alpha.
During this time we decide how deep we want to change the components.
If we want to keep some big parts (option 1), we release an alpha with almost no changes compared to the lab and iterate on it.
If we want to rewrite almost everything (option 2), we don't release to avoid making the community migrate for something that won't remain accessible in the future releases.

[...]

PR #3714, opened 2022-01-24
[core] Prepare infra for pickers migration

flaviendelangle:
Extracted from #3451 
Part of #3608
Part of #2398 

[Codesandbox example](https://codesandbox.io/s/datagrid-v5-quick-start-forked-t0s0z?file=/package.json) to show that the babel config modifications don't break the grid.

- [x] Sync `.browserlist` with the core

- [x] Allow to use the core bundling strategy by re-importing `scripts/build.js`, `scripts/buildTypes.js` and copying `scripts/copy-files.js` (the current core file copy the license for each module, I will improve it later to only copy it if the package does not already have one)

- [x] Add eslint rule to avoid import from a package inside itself

- [x] Copy eslint rule from the core for the custom `material-ui/mui-name-matches-component-name`

- [x] Rename `yarn eslint` into `yarn lint` to follow the code naming and enable the caching

- [x] Edit the babel config to add `@babel/plugin-proposal-private-methods` and `@babel/plugin-proposal-private-property-in-object` like in the core


.circleci/config.yml
```diff
@@ -142,7 +142,7 @@ jobs:
       - install_js
       - run:
           name: Eslint
-          command: yarn eslint:ci
+          command: yarn lint:ci
```
oliviertassinari, line 145: It might be better to fix this in the core repo. I mean eslint seems clearer 😁

m4theushw: But we have different linters: `markdownlint`, `jsonlint`, `stylelint` (this could be [removed](https://github.com/mui-org/material-ui-x/pull/3259#issuecomment-975509724) I think).

scripts/copy-files.js
```diff
@@ -0,0 +1,179 @@
+// Used by @mui/x-pickers and @mui/x-pickers-pro
```
m4theushw, line 1: Could you rename this file to `copyFiles.js`.

.circleci/config.yml
flaviendelangle: Done

scripts/copy-files.js
flaviendelangle: It is `copy-files` in the core but sure

.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
         '*.test.js',
         '*.test.ts',
         '*.test.tsx',
+        'test/**',
```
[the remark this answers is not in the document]
flaviendelangle: Indeed, reverted

m4theushw (approved):

PR #3718, opened 2022-01-24
[DataGrid] Fix focus when blur event rerender the grid

### LINKED ISSUE #3712: [DataGrid] Focus reverts to last focused DataGrid cell on re-rendering component
### Duplicates

### Latest version

### Current behavior 😯

If you click into a cell inside the DataGrid, then tab to focus on a new element, a state update on the parent component will cause focus to revert to the last focused cell in the DataGrid.

### Expected behavior 🤔

Focus should remain on the element you tabbed to.

### Steps to reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/datagriddemo-material-demo-forked-gwzvw?file=/demo.tsx with Chrome.
2. Click into any cell within the table body.
3. Tab to the "+" button below the table.
4. Hit Enter to activate the button. This will cause the counter state to update, which has the side effect of reverting focus to the last focused cell in the table

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
  System:
    OS: Linux 5.10 Debian GNU/Linux 9 (stretch) 9 (stretch)
  Binaries:
    Node: 12.14.1 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.14.13 - /usr/src/app/node_modules/.bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @mui/base:  5.0.0-alpha.55
    @mui/icons-material: ^5.2.1 => 5.2.1
    @mui/lab: ^5.0.0-alpha.55 => 5.0.0-alpha.55
    @mui/material: ^5.1.1 => 5.1.1
    @mui/private-theming:  5.1.1
    @mui/styled-engine:  5.1.1
    @mui/system:  5.1.1
    @mui/types:  7.1.0
    @mui/utils:  5.1.1
    @mui/x-data-grid: ^5.2.1 => 5.2.1
    @types/react:  17.0.11
    react: ^16.13.1 => 16.14.0
    react-dom: ^16.13.1 => 16.14.0
    typescript: ^4.3.5 => 4.3.5```
</details>


### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3712 

### The bug in brief

When a click outside of the grid implies a rerender of the grig, the focus goes back to the previous cell.

### What appends

The problem is that rows rerendering occur before the `handleDocumentClick` which is responsible for removing the focus in the internal state. So when the click causes a rendering we have:
- The browser focus is moved to the clicked element
- The rows rerender and the internal state is such that one cell will have prop `hasFocus=true`
- The `handleDocumentClick` is run which clean the `state.focus`

I do not understand why the `handleDocumentClick` is executed after.

### DIFF
--- a/packages/grid/_modules_/grid/components/cell/GridCell.tsx
+++ b/packages/grid/_modules_/grid/components/cell/GridCell.tsx
@@ -175,7 +175,7 @@ function GridCell(props: GridCellProps) {
         apiRef.current.scroll(scrollPosition);
       }
     }
-  });
+  }, [hasFocus, cellMode, apiRef]);
 
   return (
     <div


m4theushw:
> I do not understand why the handleDocumentClick is executed after.

This listener is the last to be executed because it's attached to `document`. Any other listener observing `click`s inside a `document` child will be executed first. If this same listener causes a rerender, it runs before the listener responsible for updating the focus.

alexfauquette:
Thanks for the explanation :) I thought that react was waiting for all the events to finish before making a new render

m4theushw (commented):
Could you add a test case? 

With the fix it seems to work correctly. 👍 However, when navigating with the keyboard the focus state is still inconsistent. #3712 uses a `button` as example, which by default also triggers a click and syncs the focus state. If another component is used instead, the key press won't trigger `handleDocumentClick` so the focus state will not be updated. Try reproducing the steps of #3712 in this [CodeSandbox](https://codesandbox.io/s/datagriddemo-material-demo-forked-xnfbk?file=/demo.tsx). There's no focus state update after pressing a key inside the fake button.

The fix proposed here only works because the effect sees that the cell already has focus and doesn't focus it again. But we should also sync the state later. I guess a solution similar to `handleDocumentClick` could be applied: check if the `keyDown` was inside the cell, then on `keyPress` update the state, etc.

alexfauquette:
I tried your codesanbox with the fix, and I don't see any problem:
https://codesandbox.io/s/create-react-app-forked-lsmbh?file=/src/App.js

m4theushw:
> I tried your codesanbox with the fix, and I don't see any problem:

I wasn't very clear. When navigating away with the keyboard it's not syncing the state (note the missing `null`). That's something we need to fix later.

msedge_9tLu7qVLbX

m4theushw (approved):

PR #3733, opened 2022-01-25
[DataGridPro] Fix missing `styleOverrides` on pinned columns

### LINKED ISSUE #3716: Class pinnedColumns can't be overwritten
### Duplicates

### Latest version

### Current behavior 😯

I try to style the pinnedColumns class with a customer style via createTheme and ThemeProvider.
Other styles can be overwritten but the pinnedColumns class doesn't seem to be accessible.

```javascript
 MuiDataGrid: {
          styleOverrides: {
            // This works
            root: {
              backgroundColor: '#fff',
            },
            // This works too
            pinnedColumnHeaders: {
              backgroundColor: '#fff',
              boxShadow: 'none',
            },
            // This doesn't work
            pinnedColumns: {
              backgroundColor: '#fff',
              boxShadow: 'none',
            },
          },
        },
```

### Expected behavior 🤔

We should be able to theme pinnedColumns too.

### Order ID 💳 (optional)

35878

alexfauquette:
Fix #3716 

CodeSandbox with the fix: https://codesandbox.io/s/disablecolumnpinningbuttons-material-demo-forked-se9cx?file=/demo.tsx


packages/grid/_modules_/grid/gridClasses.ts
```diff
@@ -389,6 +389,9 @@ export const gridClasses = generateUtilityClasses('MuiDataGrid', [
   'pinnedColumns',
   'pinnedColumns--left',
   'pinnedColumns--right',
+  'pinnedColumnHeaders',
```
m4theushw, line 392: You also need to add these classes to the interface in this same file. There're should be docs changes in this PR.

alexfauquette: Is there a reason why we don't reuse the interface keys to generate the list of classes.

I found 3 elements in `gridClasses` which are not parts of the interface, and I don't know if it is expected or not :
- 'actionsCell'
- 'groupingCriteriaCell'
- 'groupingCriteriaCellToggle'

m4theushw: I copied the same approach from MUI Core. I don't know how they check to avoid missing classes.

m4theushw (approved):
It's working.

PR #3754, opened 2022-01-27
[test] Include module augmentation for Chai custom matchers

m4theushw:
In the core repo, `test/utils/initMatchers.ts` is within the `baseUrl` of the root `tsconfig.json`, so the module augmentation also gets included. In the X repo, this file is not part of the root directory. It's only included when running the tests, so the TypeScript compiler can't see the new types. My solution was to explicitly include the file in each `tsconfig.json`. It had to be done in each config because of the way that TypeScript [handles](https://github.com/mui-org/material-ui-x/pull/3754#discussion_r793553999) config inheritance.

The monorepo was updated to get https://github.com/mui-org/material-ui/pull/30798.


packages/grid/x-data-grid/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "noEmit": true,
     "types": ["mocha", "node", "chai", "chai-dom"]
   },
-  "include": ["src"]
+  "include": ["src", "../../../node_modules/@mui/monorepo/test/utils/initMatchers.ts"]
```
flaviendelangle, line 7: Is there something preventing us from adding it to the root `tsconfig.json` file of the repo ?

m4theushw: Yes, `include` in the root `tsconfig.json` gets overwritten by the package `tsconfig.json` when extending. https://www.typescriptlang.org/tsconfig#extends

cherniavskii (approved):
Nice!

PR #3766, opened 2022-01-28
[docs] Unify multi-filtering introduction with the multi-sorting introduction

flaviendelangle:


m4theushw (commented):
Doesn't look vague this "Single and multi-filtering" section? It only contains a warning. Maybe the message should be below the first demo in the page.

image

flaviendelangle:
@m4theushw that's pretty much what I've done
But for the sorting, @joserodolfofreitas prefers to have a direct anchor to the "multi sorting" and for me both pages should be identical on that regard.

joserodolfofreitas:
> Doesn't look vague this "Single and multi-filtering" section? It only contains a warning. Maybe we kill it and move the message to below the first demo in the page. I understand we want to say that there're two types of filtering, but we already used the single filtering for the first demo.
> 
> image

We're changing it to add the blue cube to the pro feature, and I think the warning box, although arguably redundant, highlights a big difference on the community vs pro in a very clear way, so I thought that we should leave it.

joserodolfofreitas (approved):

PR #3773, opened 2022-01-29
[docs] Fix flag layout shift

oliviertassinari:
Open https://master--material-ui-x.netlify.app/components/data-grid/demo/. The width of the flag change between no loaded and loaded img:

Screenshot 2022-01-29 at 01 31 01
 
This creates a layout shift which is distracting. To be clear, this is a low value fix, but I noticed it, knew the solution as I faced it with the Autocomplete's demo already so went for it: small improvements like this compound.

Preview: https://deploy-preview-3773--material-ui-x.netlify.app/components/data-grid/demo/

### DIFF
--- a/packages/grid/x-data-grid-generator/src/renderer/renderCountry.tsx
+++ b/packages/grid/x-data-grid-generator/src/renderer/renderCountry.tsx
@@ -19,13 +19,15 @@ const Country = React.memo(function Country(props: CountryProps) {
         display: 'flex',
         alignItems: 'center',
         '&  > img': {
-          mr: '4px',
+          mr: 0.5,
+          flexShrink: 0,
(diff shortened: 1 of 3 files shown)


packages/grid/x-data-grid-generator/src/renderer/renderCountry.tsx
```diff
@@ -19,13 +19,15 @@ const Country = React.memo(function Country(props: CountryProps) {
         display: 'flex',
         alignItems: 'center',
         '&  > img': {
-          mr: '4px',
+          mr: 0.5,
+          flexShrink: 0,
```
oliviertassinari, line 23: The fix

flaviendelangle (approved):

PR #3798, opened 2022-01-31
[core] Stop using an enum for `GridPreProcessingGroup`

flaviendelangle:
Extracted from #3688

100% `core`, no impact for users.

Enums can't be extended between the community and the pro package.
For the pre-processing I reverted the logic by generating the type as the keys of the lookup.
Maybe the `GridEvents` could also be a type instead of an enum in v6.


alexfauquette (approved):
I'm not that much familiar with typescript. That's what I understand:

The `GridPreProcessingGroup` enum can not be extended which is a problem to add for example row grouping.

So instead, you use `keyof GridPreProcessingGroupLookup` which is equivalent to 
`
'hydrateColumns' | 'scrollToIndexes' | ...
`

This type is used for the `unstable_applyPreProcessors` to indicate to which group the preprocessor must be applied

If yes, its good for me

flaviendelangle:
@alexfauquette Exactly, and it is also used to type directly the callbacks

```tsx
  const updateSelectionColumn = React.useCallback<
    GridPreProcessor<GridPreProcessingGroup.hydrateColumns>
  >(
```

PR #3814, opened 2022-02-01
[core] Save performance snapshot to AWS S3

m4theushw:
One more step to improve the performance results reported in each PR. This PR is saving the snapshot to S3 so we can later compare the current PR's results with the results from the last merged commit. I copied the same config, including bucket name, used in MUI Core. Maybe we should save under a different prefix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs: 
+  aws-s3: circleci/aws-s3@3.0.0
 
 parameters:
   browserstack-force:
@@ -34,6 +36,7 @@ defaults: &defaults
     # expose it globally otherwise we have to thread it from each job to the install command
     BROWSERSTACK_FORCE: << pipeline.parameters.browserstack-force >>
     REACT_DIST_TAG: << parameters.react-dist-tag >>
+    AWS_REGION_ARTIFACTS: eu-central-1
   working_directory: /tmp/mui
   docker:
     - image: circleci/node:12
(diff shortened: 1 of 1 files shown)


m4theushw:
@oliviertassinari Could you configure the environment variables in CircleCI and check if it has permissions to write into S3?

.circleci/config.yml
```diff
@@ -267,6 +270,25 @@ jobs:
...
+                value: << pipeline.git.branch >>
+          steps:
+            # persist performance snapshot on S3
+            - aws-s3/copy:
+                arguments: --content-type application/json
+                aws-access-key-id: AWS_ACCESS_KEY_ID_ARTIFACTS
+                aws-region: AWS_REGION_ARTIFACTS
+                aws-secret-access-key: AWS_SECRET_ACCESS_KEY_ARTIFACTS
+                from: performance-snapshot.json
+                to: s3://mui-org-ci/artifacts/$CIRCLE_BRANCH/$CIRCLE_SHA1/
```
oliviertassinari, line 291: I would suggest we isolate the different repositories. This is what I see on https://s3.console.aws.amazon.com/s3/buckets/mui-org-ci?region=eu-central-1&prefix=artifacts/pull/29876/9f74e56b5e340db2004ea566c128da687ab0a509/&showversions=false

Screenshot 2022-02-07 at 16 39 53

Either maybe with a file prefix

```suggestion
                from: mui-x-performance-snapshot.json
                to: s3://mui-org-ci/artifacts/$CIRCLE_BRANCH/$CIRCLE_SHA1/
```

or with a repository prefix:

```suggestion
                from: performance-snapshot.json
                to: s3://mui-org-ci/artifacts/mui-x/$CIRCLE_BRANCH/$CIRCLE_SHA1/
```

oliviertassinari (commented):
I have added the AWS keys in the config

.circleci/config.yml
Janpot: IMO, hierarchically, the repo prefix makes most sense.
```
from: size-snapshot.json
to: s3://mui-org-ci/artifacts/$CIRCLE_PROJECT_REPONAME/$CIRCLE_BRANCH/$CIRCLE_SHA1/
```
I would suggest to do the same in the material-ui repo, and maintain the same file name (`size-snapshot`) so that we don't have to special case the contributor-dashboard table

flaviendelangle (commented):
Anything new on this PR ? Should we continue working on it ? Closing it ?

PR #3840, opened 2022-02-04
[DataGridPro] Add clear error when the tree data has duplicated paths

### LINKED ISSUE #3829: [DataGridPro] Tree view error message for duplicate path is not helping
### Duplicates

### Latest version

### Current behavior 😯

I was using a tree view when I get the following error message
> TypeError
> Cannot read properties of undefined (reading 'children')

https://codesandbox.io/s/rendercellgrid-material-demo-forked-1udwf?file=/demo.tsx

The error occurs [here](https://github.com/alexfauquette/material-ui-x/blob/excel-export/packages/grid/_modules_/grid/utils/tree/buildRowTree.ts#L162:L173)
because `tempNode` is undefined and so `tempNode.children` makes js return an error, but I cant find a way to put the warning

### Expected behavior 🤔

If two paths are the same, the error message should explicitly says "TreeDataPath should be unique. Yours rows with ids x and y have the same one".

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Closes #3829

### DIFF
--- a/packages/grid/_modules_/grid/utils/tree/buildRowTree.ts
+++ b/packages/grid/_modules_/grid/utils/tree/buildRowTree.ts
@@ -25,6 +25,11 @@ interface BuildRowTreeParams extends GridRowGroupParams {
   defaultGroupingExpansionDepth: number;
   isGroupExpandedByDefault?: (node: GridRowTreeNodeConfig) => boolean;
   groupingName: string;
+  onDuplicatePath: (
+    firstId: GridRowId,
+    secondId: GridRowId,
+    path: BuildRowTreeGroupingCriteria[],
+  ) => void;
 }
 
 interface TempRowTreeNode extends Omit<GridRowTreeNodeConfig, 'children' | 'childrenExpanded'> {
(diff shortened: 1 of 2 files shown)


packages/grid/_modules_/grid/utils/tree/buildRowTree.ts
```diff
@@ -108,19 +113,24 @@ export const buildRowTree = (params: BuildRowTreeParams): GridRowGroupingResult
...
-          depth === row.path.length - 1
-            ? row.id
-            : `auto-generated-row-${row.path
-                .map((groupingCriteria) => `${groupingCriteria.field}/${groupingCriteria.key}`)
-                .slice(0, depth + 1)
-                .join('-')}`;
+
+      if (keyConfig) {
+        if (depth === row.path.length - 1) {
+          params.onDuplicatePath?.(keyConfig.id, row.id, row.path);
```
flaviendelangle, line 119: For the row grouping, this scenario should never occur unless the user has duplicated ids, which would throw another error before.

alexfauquette:
I assume the introduction of `onDuplicatePath` argument is here to customize the error, but if another error occurs before for row grouping, why not hard coding it.

I also thought about adding `if(process.env.NODE_ENV !== 'production')` for [bundle size reduction](https://github.com/mui-org/material-ui-x/pull/3757/files/24e63964901b69244ed2ef3962e14c73d226b12b#r795062300, but since its an error and not a warning, I don't know if we should display it in production to help error correction or not

flaviendelangle:
> why not hard coding it.

Because for the Row Grouping there is no concept of "path" as far as the user is concerned.
The technical error is "two rows have the same grouping criterias"
For the Tree Data, it means that two rows have the same path
For the Row Grouping, it means that two rows have the same ID

alexfauquette (approved):

PR #3845, opened 2022-02-04
[DataGrid] Do not restore the focus to the trigger element of the `GridPanel` when unmounting it

### LINKED ISSUE #3830: [data grid] Closing a preference panel causes a scroll to its DataGrid
### Current behavior 😯

A page contains several tables (DataGrids), each of them contains ability of columns toggling
Click on 'columns selector' button displays pop-up with column list
If, instead, ones scroll down to next (last) DataGrid at the bottom of page and then click on 'columns selector' button - 
it scrolls up to the previously open ones

### Expected behavior 🤔

Assumed that new pop-up opens near DataGrid which it controls

### Steps to reproduce 🕹

Steps:

1. Find a page with two or more DataGrids with 'columns selector' controls
2. Open 'columns selector' pop-up at upper DataGrid
3. scroll down to bottom DataGrid and open 'columns selector' header button
4.look which DataGrid is displayed

### Context 🔦

Sorry if some details missed, I'm not a frontend dev

### Your environment 🌎
Chrome 97.0.4692.99 on Mac OS 11.6

flaviendelangle:
FIx #3830

Preview: https://codesandbox.io/s/datagridpro-v5-quick-start-forked-l3y8h

As the panel does not block the scroll, it is unsafe to restore the previous focus when unmounting it since it can be totally outside the viewport.

### DIFF
--- a/packages/grid/_modules_/grid/components/panel/GridPanelWrapper.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridPanelWrapper.tsx
@@ -43,7 +43,7 @@ export function GridPanelWrapper(
   const classes = useUtilityClasses(ownerState);
 
   return (
-    <TrapFocus open disableEnforceFocus isEnabled={isEnabled}>
+    <TrapFocus open disableEnforceFocus disableRestoreFocus isEnabled={isEnabled}>
       <GridPanelWrapperRoot tabIndex={-1} className={clsx(className, classes.root)} {...other} />
     </TrapFocus>
   );


m4theushw (commented):
Using `disableRestoreFocus` breaks the accessibility of the grid since now the focus will not return to the filter button and someone using a screen reader will have to start again for the beginning of the page after closing the panel. The scroll issue could have been solved if TrapFocus allowed to `element.focus({ preventScroll: true })`. Since it doesn't I would be in favor of adding a slot for `GridPanelWrapper` and a prop to configure its TrapFocus or a DataGrid prop. For accessibility reasons, the action of moving the focus back to the element that opened the panel is expected.

siriwatknp:

3. run `yarn docs:api`

flaviendelangle:
Will be fixed by https://github.com/mui/material-ui/pull/30991

DanailH (approved):

DanailH:
Do we want to keep this one open or can we merge it? It's been open since February.

flaviendelangle:
I did not find any viable solution that do not break the focus :/
If someone wants to have a look

PR #3848, opened 2022-02-05
[DataGrid] Add support for margin between rows

### LINKED ISSUE #2144: [DataGrid] Allow to add margin between rows

m4theushw:
Closes #2144 

Preview: https://deploy-preview-3848--material-ui-x.netlify.app/components/data-grid/rows/#row-spacing

In #3387, we added a pre-processor to allow to customize the final row height. This way, we got almost for free the support for custom margins between rows. This PR proposes to add a new `getRowSpacing` prop where developers can pass the top and bottom margin values for each row. Since `border` also behaves like `margin` the prop doesn't really add the margins, but takes these values into account when calculating the content size and which rows to render during scroll.


packages/grid/_modules_/grid/hooks/utils/useCurrentPageRows.ts
```diff
@@ -41,11 +42,19 @@ export const useCurrentPageRows = (
 ) => {
   const response = getCurrentPageRows(apiRef, props);
 
+  const lookup = React.useMemo(() => {
```
flaviendelangle, line 45: If you put non-trivial computation here they will be computed for each hook that calls `useCurrentPageRows` and will not take advantage of the selector memoization.
For very large list of rows I think this is to be avoided.

If you are only using it in 1 or 2 `useCurrentPageRows`, I would move the `lookup` computation to the hook / components which calls `useCurrentPageRows`.

If you are using it in more, I would move the `lookup` computation in `getCurrentPageRows` in selectors (one for the paginated version and one for the non-paginated version).

packages/grid/_modules_/grid/models/gridRows.ts
```diff
@@ -80,7 +80,16 @@ export type GridRowsLookup = Record<GridRowId, GridRowModel>;
  */
 export type GridRowId = string | number;
 
-export type GridRowEntry = { id: GridRowId; model: GridRowModel };
+export interface GridRowEntry {
+  /**
+   * The row model of the row that the current cell belongs to.
```
flaviendelangle, line 85: We are using this interface in a lot of selectors that are not related to cells.
Could we rephrase these descriptions to be more generic ?

packages/grid/_modules_/grid/models/params/gridRowParams.ts
```diff
@@ -29,7 +29,21 @@ export interface GridRowParams<R extends GridRowModel = GridRowModel> {
...
 
 /**
- * Object passed as parameter in the row getRowHeight callback.
+ * Object passed as parameter in the row `getRowClassName` callback prop.
+ */
+export interface GridRowClassNameParams extends GridRowParams {
+  /**
+   * Whether this row is the first visible or not.
+   */
+  isFirstVisible: boolean;
```
flaviendelangle, line 38: Right now we mostly use the term "visible" to refer to the rows after sorting + filtering + children colapsing processed.

From what I understand, the `isFirstVisible` and `isLastVisible` is used here to talk about the paginated rows (and when pagination is off, all the sorted + filtered + collapsed rows).

I don't think we should use `visible` here.
Not in v5 because it conflicts with other usages of the term `visible` and makes it incoherent.
Not in v6 because I think we should stop using the term `visible` which is not clear (as we saw that each dev in the team has a different definition based on the day and the use case).

______

I would be in favor of using the term `page` even if the pagination is off since it's like it we had `pageSize=Infinity` and pagination on.

So `isFirstInPage` for instance.

But if you have a better idea I'm open to suggestion.
Just let's avoid term that can be understood in various ways.

joserodolfofreitas (approved):
Nice feature. 
I'm adding a suggestion to make the example more granular.

siriwatknp:

3. run `yarn docs:api`

packages/grid/_modules_/grid/models/params/gridRowParams.ts
m4theushw: By using "visible" I wanted to adopt the same concept from [AG Grid](https://ag-grid.com/react-data-grid/grid-api/#reference-displayedRows). The visible (or displayed) rows are those you can see by only scrolling the grid. In the past it was heavily misused but I would ignore and start using "visible" in this way from now.

flaviendelangle: If we are acting that `visible` should mean "accessible by scroll"
Then could we 

1. Create an issue for renaming everything to this new nomenclature or at least put a line in #3287 
2. Be more precise in the description of this prop, at least until v6 `Wheter this row is the first row reachable by scrolling or not.` or something similar. So that we are sure people understand `visible` in the correct way

flaviendelangle (commented):
What is the reason behind the public api you made for this feature ?
Why the user have tu set the CSS margin itself ? Couldn't we do it automatically, so he just have to set a row spacing and everything works ?

I feel like your approach is very flexible but not very intuitive. 
For the variable row height for instance, we are not asking to the user to set the CSS height himself.

m4theushw:
> Why the user have tu set the CSS margin itself ? Couldn't we do it automatically, so he just have to set a row spacing and everything works ?

It's because border also behaves like a margin but it can have color and style. If we set everything automatically `getRowSpacing` would have to return a object of CSS properties, and for border we need to parse the value to extract the width.

docs/src/pages/components/data-grid/rows/rows.md
```diff
@@ -100,6 +100,15 @@ If you need some rows to have different row heights this can be achieved using t
 > <DataGridPro getRowHeight={handleGetRowHeight} />
 > ```
 
+## Row spacing
+
+By using the `getRowSpacing` prop you can specify the vertical space between rows.
+This prop is called with a [`GridRowSpacingParams`](/api/data-grid/grid-row-spacing-params/) object.
+Observe that the value specified will only be used internally to calculate which rows to render during scroll and for space allocation.
```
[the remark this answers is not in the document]
m4theushw: I added a better explanation of how to use the new prop and that `getRowClassName` should be also used.

flaviendelangle:
Could we have something like `getRowSpacing: (params: XXX) => ({ type: 'margin' | 'border', value: number })`
And set `borderWidth` if `type === 'border'` or `margin` if `type === 'margin'` ?

I would like to have the insight of the rest of the team.
Because the current api feels weird to me.

packages/grid/x-data-grid/src/internals/hooks/features/rows/useGridRowsMeta.ts
```diff
@@ -44,16 +48,11 @@ export const useGridRowsMeta = (
...
-      paginationMode,
-    });
-
     apiRef.current.setState((state) => {
       const positions: number[] = [];
       const densityFactor = gridDensityFactorSelector(state, apiRef.current.instanceId);
       const currentRowHeight = gridDensityRowHeightSelector(state, apiRef.current.instanceId);
-      const currentPageTotalHeight = rows.reduce((acc: number, row) => {
+      const currentPageTotalHeight = currentPage.rows.reduce((acc: number, row) => {
         positions.push(acc);
```
alexfauquette, line 56: Should the position start with `initialHeights.spacingTop`?

alexfauquette (commented):
It seems there is a problem with the detail panel. The gap between the row and the detail increase along rows
image

https://codesandbox.io/s/create-react-app-with-typescript-forked-olcpk5?file=/src/App.tsx

One thing I do not understand is why specifying `spacingTop` and `spacingBottom` if at the end we just sum those two values

packages/grid/x-data-grid/src/internals/models/params/gridRowParams.ts
```diff
@@ -29,7 +29,21 @@ export interface GridRowParams<R extends GridRowModel = GridRowModel> {
 }
 
 /**
- * Object passed as parameter in the row getRowHeight callback.
+ * Object passed as parameter in the row `getRowClassName` callback prop.
```
DanailH, line 32: Can we do something like:
```
interface GridRowPossitionVisibility {
  /**
   * Whether this row is the first visible or not.
   */
  isFirstVisible: boolean;
  /**
   * Whether this row is the last visible or not.
   */
  isLastVisible: boolean;
}
```
and then use that to build the 2 new interfaces?

packages/grid/x-data-grid/src/internals/hooks/features/rows/useGridRowsMeta.ts
m4theushw: I didn't understand. This `positions.push(acc);` is adding the position of the previous row. Note that `positions[0]` is always 0.

packages/grid/x-data-grid/src/internals/models/params/gridRowParams.ts
m4theushw: Done.

m4theushw:
> It seems there is a problem with the detail panel. The gap between the row and the detail increase along rows

@alexfauquette True, I was not adding the top space when positioning the detail panel. Now it's fixed. However, I noted that adding bottom margin with `getRowSpacing` to a row that also has a detail panel doesn't work because both features relay on `marginBottom`. Changing to border also doesn't work. I added a paragraph explaining this limitation to the docs. If someone complains about it there's a possible solution which is to create a "fake margin" on each row by using a div.

---

> Could we have something like getRowSpacing: (params: XXX) => ({ type: 'margin' | 'border', value: number })

@flaviendelangle I added a new `rowSpacingType` prop to configure to use border or margin. That way users can only use one type.

packages/grid/x-data-grid/src/internals/hooks/features/rows/useGridRowsMeta.ts
alexfauquette: My idea was the following:
If the first row has a `margin-top: 5px` which in your system correspond to `initialHeights.spacingTop=5` then the firs row position should reflect this margin with `positions[0] = '5px'`

m4theushw: With `initialHeights.spacingTop > 0` the `positions` array is more like how many pixels should I scroll to reach the beginning of the top margin of a row. Although visually the margin is outside the row, here it's part of the row so I think makes sense to keep 0 as the initial value. I don't want to change because I also don't know the consequences of this yet. But one problem of not discounting the margin is that when navigating with the keyboard it scroll up to the top margin. Maybe we could have a option in `scrollToIndexes` to control de behavior. #4013 asks for more parameters.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -142,12 +143,24 @@ export const useGridRows = (
...
+  const currentPage = useCurrentPageRows(apiRef, props);
 
   const getRow = React.useCallback<GridRowApi['getRow']>(
     (id) => gridRowsLookupSelector(apiRef)[id] ?? null,
     [apiRef],
   );
 
+  const lookup = React.useMemo(
+    () =>
+      currentPage.rows.reduce((acc, { id }, index) => {
```
flaviendelangle, line 155:
```suggestion
      currentPage.rows.reduce<Record<GridRowId, number>>((acc, { id }, index) => {
```

This should also work by the way

packages/grid/x-data-grid/src/models/api/gridSortApi.ts
```diff
@@ -53,6 +53,7 @@ export interface GridSortApi {
    * The index is based on the sorted but unfiltered row list.
    * @param {GridRowId} id The `GridRowId` of the row.
    * @returns {number} The index of the row.
+   * @deprecated Use `unstable_getRowIndexRelativeToCurrentPage` instead.
```
flaviendelangle, line 56: We advise users to use a private API instead of a public one ?
Does this mean that `unstable_getRowIndexRelativeToCurrentPage` will become public at some point ?

m4theushw: It would be public in v6 and replace `getRowIndex`. I made it public now.

docs/data/data-grid/rows/RowMarginGrid.tsx
```diff
@@ -0,0 +1,32 @@
...
+    };
+  }, []);
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGrid
+        {...data}
+        getRowSpacing={getRowSpacing}
+        sx={{
+          '& .MuiDataGrid-row': {
```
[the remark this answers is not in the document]
m4theushw: I changed to `gridClasses`. Note that no other demo is using it. If we use it here we should start using it in the future demos to keep consistency.

flaviendelangle (approved):

docs/data/data-grid/rows/RowMarginGrid.js
```diff
@@ -0,0 +1,32 @@
...
+  }, []);
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGrid
+        {...data}
+        getRowSpacing={getRowSpacing}
+        sx={{
+          [`& .${gridClasses.row}`]: {
+            bgcolor: (theme) => (theme.palette.mode === 'dark' ? '#000' : '#efefef'),
```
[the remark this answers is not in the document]
m4theushw: Either `#000` or `theme.palette.background.default` mix with the page background and I can't see the difference. I changed to other colors.

docs/data/data-grid/rows/rows.md
```diff
@@ -100,6 +100,36 @@ If you need some rows to have different row heights this can be achieved using t
...
+```tsx
+<DataGrid
+  getRowSpacing={...}
+  rowSpacingType="border"
+  sx={{ '& .MuiDataGrid-row': { borderTopColor: 'yellow', borderTopStyle: 'solid' } }}
+/>
+```
+
+> ⚠ Adding a bottom margin or border to rows that also have a [detail panel](/components/data-grid/group-pivot/#master-detail) is not recommended.
+> As alternative, increase the top spacing of the next row to compensate.
```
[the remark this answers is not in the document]
m4theushw: It's simpler to always compensate. I improved the explanation. I think that if the user wants to apply a margin when the panel is closed and once it opens change the margin he is doing something very specific to explain in a warning message.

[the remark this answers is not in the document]
alexfauquette: It is the word "compensate" that miss leads me. I thought it means users have to compensate for the size of the expended detail panel. An explanation like that could be enough.

> Adding a bottom margin or bottom border to rows that also have a [detail panel](/components/data-grid/group-pivot/#master-detail) is not recommended because the detail panel relays on the bottom margin to work.
> 
> As an alternative, only use the top spacing to define the space between rows.

alexfauquette (approved):

docs/data/data-grid/rows/rows.md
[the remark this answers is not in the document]
m4theushw: Updated.

PR #3881, opened 2022-02-07
[docs] Use Grammarly to fix some typo

alexfauquette:
I tried a Grammarly extension to help catch English errors in the doc. @samuelsycamore could you check that I'm not introducing  new ones

### DIFF
--- a/docs/data/data-grid/columns/columns.md
+++ b/docs/data/data-grid/columns/columns.md
@@ -125,7 +125,7 @@ The value generated is used for filtering, sorting, rendering, etc. unless overr
 ### Value setter
 
 The value setter is to be used when editing rows and it is the counterpart of the value getter.
-It allows customizing how the entered value is stored in the row.
+This enables you to customize how the entered value is stored in the row.
 A common use case for it is when the data is a nested structure.
 Refer to the [cell editing](/components/data-grid/editing/#saving-nested-structures) documentation to see an example using it.
 
@@ -157,7 +157,7 @@ In the following demo, a formatter is used to display the tax rate's decimal val
 {{"demo": "ValueFormatterGrid.js", "bg": "inline"}}
 
 The value generated is only used for rendering purposes.
-Filtering and sorting will not rely on the formatted value.
+Filtering and sorting do not rely on the formatted value.
 Use the [`valueParser`](/components/data-grid/columns/#value-parser) to support filtering.
 
 ### Value parser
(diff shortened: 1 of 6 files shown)


siriwatknp:

3. run `yarn docs:api`

samuelsycamore:
Looks good overall. I see lots of improvements here. A few suggestions _(in italics)_:

128. _This enables you to customize_ how the entered value is stored in the row.

160. Filtering and sorting _do_ not rely on the formatted value.

444. Pinned (or frozen, locked, or sticky) columns are columns that are visible _at all times_ while the user scrolls the grid horizontally.

478. While the `initialState` prop only works for setting pinned columns during the initialization, the `pinnedColumns` prop allows _you to modify which columns are pinned at any time_.

527. _You can modify this default behavior with column spanning._

270. _The example below demonstrates how to achieve server-side filtering._

70. You can override the **headerName** or any property of the `GridColDef` interface, except the `field`, the `type`, and the properties related to inline **edition.** I'm not familiar with the context here, but should this be **addition**?

96. - Add _an_ `onPageChange` callback to load the rows when the page changes

docs/data/data-grid/columns/columns.md
```diff
@@ -524,7 +524,7 @@ Grouping columns allows you to have multiple levels of columns in your header an
 > 👍 Upvote [issue #192](https://github.com/mui/mui-x/issues/192) if you want to see it land faster.
 
 Each cell takes up the width of one column.
-Column spanning allows to change this default behavior.
+Column spanning allows modifying this default behavior.
 It allows cells to span multiple columns.
```
alexfauquette, line 528:
```suggestion
It allows cells to span multiple rows.
```

alexfauquette:
@samuelsycamore Thanks for the feedback. For making suggestions, github has an amazing tool

In "Files changed", on the left of each line of code, you have a button to make a comment. And in the comment editor, you have this button
image
It adds the line of code you clicked. You then modified it, and other will see your suggestion as a git diff in its context


Peek 2022-02-10 09-31

alexfauquette:
About your point 134. the context is the following:

Users are able to group rows, and this part of the doc is about how to customize the columns that allow to extend or collapse a group of rows. Values of this column can not be edited by the end-user, so we say to developers that they can not override "properties related to inline edition"

samuelsycamore:
Thank you so much for that walkthrough @alexfauquette! I'll be sure to use that tool going forward. 👍

PR #3882, opened 2022-02-07
[DataGrid] Add indexes relative to the filtered rows and the current page to the `getRowClassName` and `getRowSpacing` props

### LINKED ISSUE #2334: [DataGrid] Add API to style even and odd rows differently

## Summary 💡

Some might want to style even and odd rows slightly differently to help end-users navigate a data table. It's usually when there is no hover style on the rows. When there is a hover style, it's not recommended.

## Examples 🌈

An example with TablePlus, a SQL client

Capture d’écran 2021-08-13 à 19 00 59

## Details

It's a follow-up #2327 where we break this feature. A possible solution is shared in https://github.com/mui-org/material-ui-x/pull/2327#pullrequestreview-728769169. If too many developer complain, we can revert too.

flaviendelangle:
Fixes #2334

Add two params to `GridRowVisibilityParams`, which is passed to `props.getRowClassName` and `props.getRowSpacing`:

- `indexRelativeToExpandedRows`
- `indexRelativeToCurrentPage`


docs/src/pages/components/data-grid/style/StripedGrid.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { styled } from '@mui/material/styles';
+import { DataGrid, GridRowProps, GridRow, gridClasses } from '@mui/x-data-grid';
+import { useDemoData } from '@mui/x-data-grid-generator';
+import clsx from 'clsx';
+
+const StripedDataGrid = styled(DataGrid)({
+  [`& .${gridClasses.row}.odd`]: {
+    backgroundColor: '#EEEEEE',
+    '&:hover, &.Mui-hovered': {
```
alexfauquette, line 10: It needs to handle the case `.Mui-selected`. Overwise odd lines do not provide feedback when selected

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -296,6 +297,12 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
         isSelected = true;
       }
 
+      const indexes: GridRowIndexes = {
+        visibleRows: currentPage.range.firstRowIndex + nextRenderContext.firstRowIndex! + i,
+        pageRows: nextRenderContext.firstColumnIndex + i,
```
m4theushw, line 302: Column?

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -296,6 +297,12 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
         isSelected = true;
       }
 
+      const indexes: GridRowIndexes = {
+        visibleRows: currentPage.range.firstRowIndex + nextRenderContext.firstRowIndex! + i,
```
m4theushw, line 301: This value is wrong. I suggested a new formula in https://github.com/mui/mui-x/pull/3519/files/6cfcd930f646a7866b46456fade5d35c0dc31e26#diff-7fb7dbb1d7fe2c699d9fc5cfc5c6a56b2489af3b46dc1a2a8eaa3404575cacbd

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -15,11 +15,37 @@ import { GridStateColDef } from '../models/colDef/gridColDef';
...
 import { gridColumnsMetaSelector } from '../hooks/features/columns/gridColumnsSelector';
 import { useGridSelector } from '../hooks/utils/useGridSelector';
+import { GridRenderEditCellParams } from '../models';
+
+export interface GridRowIndexes {
+  /**
+   * Index of the row in the whole sorted and filtered dataset.
+   */
+  visibleRows: number;
+
```
m4theushw, line 25:
```suggestion
```

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -15,11 +15,37 @@ import { GridStateColDef } from '../models/colDef/gridColDef';
...
+}
 
-export interface GridRowProps {
+export interface GridRowProps extends React.HTMLAttributes<HTMLDivElement> {
   rowId: GridRowId;
   selected: boolean;
+  /**
+   * Index of the row in the whole sorted and filtered dataset.
+   * If some rows above have expanded children, this index also take those children into account.
+   * @deprecated Use `props.indexes.visibleRows` instead.
```
m4theushw, line 45: I'm not sure about deprecating it. I don't see a use case for the other values passed in `indexes`. I would first fix `index` to pass the correct value.

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -15,11 +15,37 @@ import { GridStateColDef } from '../models/colDef/gridColDef';
...
+
+  /**
+   * Index of the row in the current page.
+   * If the pagination is disabled, this value will be equal to the `dataset` value.
+   */
+  pageRows: number;
+
+  /**
+   * Index of the row in the list of rows currently rendered by the virtualization engine.
+   * If the pagination is disabled, this value will be equal to the `page` value.
```
alexfauquette, line 34:
```suggestion
   * If the virtualization is disabled, this value will be equal to the `pageRows` value.
```

packages/grid/_modules_/grid/components/GridRow.tsx
```diff
@@ -15,11 +15,37 @@ import { GridStateColDef } from '../models/colDef/gridColDef';
...
+
+export interface GridRowIndexes {
+  /**
+   * Index of the row in the whole sorted and filtered dataset.
+   */
+  visibleRows: number;
+
+  /**
+   * Index of the row in the current page.
+   * If the pagination is disabled, this value will be equal to the `dataset` value.
```
alexfauquette, line 28:
```suggestion
   * If the pagination is disabled, this value will be equal to the `visibleRows` value.
```

alexfauquette (commented):
About naming, the `xxxRows` let me expect an array, whereas it is a number.

Maybe `indexes.visibleIndex` or `indexes.fromVisibleRows`

For the demo position, it looks good to me

packages/grid/_modules_/grid/components/GridRow.tsx
flaviendelangle: What is the correct value in your mind ?

m4theushw: https://github.com/mui/mui-x/pull/3882#discussion_r800605561

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
m4theushw: The comment says "Index of the row in the current page." but it uses `nextRenderContext.firstColumnIndex`.

packages/grid/_modules_/grid/components/GridRow.tsx
flaviendelangle: Formula fixed

packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualScroller.tsx
flaviendelangle: That's a typo, fixed

flaviendelangle:
@alexfauquette I went for `indexes.fromFilteredRows` and `index.fromPageRows`

docs/src/pages/components/data-grid/style/StripedGrid.tsx
flaviendelangle: I improved the striped behavior to use the theme and handle `.Mui-selected`

docs/src/pages/components/data-grid/style/StripedGrid.js
```diff
@@ -0,0 +1,79 @@
...
+import { useDemoData } from '@mui/x-data-grid-generator';
+import clsx from 'clsx';
+
+const ODD_OPACITY = 0.2;
+
+const StripedDataGrid = styled(DataGrid)(({ theme }) => ({
+  [`& .${gridClasses.row}.even`]: {
+    backgroundColor: '#EEEEEE',
+    '&:hover, &.Mui-hovered': {
+      backgroundColor: alpha(theme.palette.primary.main, ODD_OPACITY),
```
alexfauquette, line 14:
```suggestion
      backgroundColor: '#E0E0E0',
```
The hover should move from light to darker gray. Here I propose the hover moves from grey-200 to grey-300 according to https://material.io/resources/color/#!/?view.left=0&view.right=0&primary.color=bcbcbc

siriwatknp:

3. run `yarn docs:api`

docs/src/pages/components/data-grid/style/StripedGrid.js
flaviendelangle: How do you make the hover and / or selected consistent with this color ?

packages/grid/_modules_/grid/components/GridRow.tsx
m4theushw: How about passing the index relative to only the rows reacheable by scroll (being `nextRenderContext.firstRowIndex! + i`) to `getRowClassName`? This way, developers can assign a custom class name without having to provide a custom row component. I don't like passing the `indexes` prop because if want to remove it later it will require a breaking change.

flaviendelangle: I agree that passing the indexes to `getRowClassName` makes the example easier. I will improve the PR that way :+1: 

We could rename `useCurrentPageRows` => `useGridCurrentPageRows` to match the other exporter hooks and in that case I agree that user can handle there more specific usecases without the `indexes` prop.

m4theushw: > We could rename useCurrentPageRows => useGridCurrentPageRows to match the other exporter hooks and in that case I agree that user can handle there more specific usecases without the indexes prop.

Sure, lets rename `useCurrentPageRows` and export it, in another PR maybe. I want to avoid deprecating props whenever is possible because in v6 it will be one more thing for developers to care about.

flaviendelangle: If we do the export in another PR, what remains for this one ?

m4theushw: What remains is to pass the indexes to `getRowClassName` and not deprecate the `index` prop. With that, developers can already create stripped rows without overriding the row component.

flaviendelangle: I will finish this PR after https://github.com/mui/mui-x/pull/3848 is merged to avoid having two PRs adding the `GridRowClassNameParams`

flaviendelangle: Yes, this PR is not urgent so I prefer to finish #3896 before :+1:

docs/data/data-grid/style/style.md
```diff
@@ -117,6 +117,12 @@ Choose between one of the following values: 'left' | 'right' | 'center'.
 
 **Note**: You must use `headerAlign` to align the content of the header.
 
+## Striped rows
+
+The following demo illustrates how the rows of the grid can be stripped.
```
m4theushw, line 122:
```suggestion
You can use the `indexRelativeToCurrentPage` param passed to `getRowClassName` to apply alternating styles to the rows.

code example...

The following demo illustrates how the rows of the grid can be striped.
```

docs/data/data-grid/style/style.md
```diff
@@ -117,6 +117,12 @@ Choose between one of the following values: 'left' | 'right' | 'center'.
...
+
+{{"demo": "StripedGrid.js", "bg": "inline", "defaultCodeOpen": false}}
```
m4theushw, line 124: I think the code example above can be replaced if we show the preview.

```suggestion
{{"demo": "StripedGrid.js", "bg": "inline"}}
```

m4theushw (commented):
Which problem `indexRelativeToExpandedRows` is solving? I can't remember it.

flaviendelangle:
> Which problem indexRelativeToExpandedRows is solving? I can't remember it.

Initially I did a prop so I kept the `indexRelativeToExpandedRows` which was the equivalent of the current `props.index` in `GridRow`.
But for the params of `getRowClassName` or `getRowSpacing` I don't have a real use-case in mind.
Some users might want to style based on the all-page-index, but that's probably not a frequent use-case.

We can leave it, as it does not bring a lot of additional complexity.
Or remove it and add if somebody asks for it later.

flaviendelangle:
I removed it
Let's wait for a concrete use case, adding it will be very easy

PR #3900, opened 2022-02-09
[DataGrid] Add `valueExportFormatter` to `GridColDef`

### LINKED ISSUE #3806: [DataGrid] CSV export does not work for the `Commodity` and `Employee` data sets `country` column
### Duplicates

### Latest version

### Current behavior 😯

The `valueFormatter` of a `singleSelect` column can return an object. It is encourage to return one with two properties: value and label such that it works well with `Select` component.

In such case, the csv export display `[object Object]` see [the demo with single select field](https://codesandbox.io/s/exportdefaulttoolbar-material-demo-forked-3hsfw?file=/demo.js)
```csv
Name,Rating,Country,Created on,Is admin?
George Edwards,4,[object Object],12/3/2021,true
Ruth Simmons,2,[object Object],12/20/2021,true
Barbara Greer,5,[object Object],11/22/2021,false
Daisy Morgan,2,[object Object],1/24/2022,true
```

### Expected behavior 🤔

According to the typescript, `valueFormatter` return type is 
```ts
GridCellValue = string | number | boolean | Date | null | undefined | object; 
```

I propose to add a `valueExportFormatter` which return string. The reason is that `valueFormatter` is made for providing data to a React component, which is more flexible than providing data to a csv file with can only handle string

### Steps to reproduce 🕹

Steps:

1. go to the codesandbox https://codesandbox.io/s/exportdefaulttoolbar-material-demo-forked-3hsfw?file=/demo.js
2. export csv
3. country column contains `[object Object]`

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

DanailH:
Related to #3806 

The `valueExportFormatter` callback is used to specify an export value when doing a CSV export.


alexfauquette:
I realized that the problem also occurs in clipboard feature, which could be solved by the modification

docs/src/pages/components/data-grid/export/ExportDefaultToolbar.tsx
```diff
@@ -2,16 +2,23 @@ import * as React from 'react';
 import { useDemoData } from '@mui/x-data-grid-generator';
 import { DataGrid, GridToolbar } from '@mui/x-data-grid';
 
+const VISIBLE_FIELDS = ['name', 'rating', 'country', 'dateCreated', 'isAdmin'];
+
 export default function ExportDefaultToolbar() {
   const { data, loading } = useDemoData({
-    dataSet: 'Commodity',
+    dataSet: 'Employee',
```
m4theushw, line 9: Could we keep the commodity dataset? All demos in this page use it. In this dataset the `counterPartyCountry` uses the `singleSelect` with a array of objects as `valueOptions`.

packages/grid/_modules_/grid/hooks/features/rows/useGridParamsApi.ts
```diff
@@ -156,6 +156,14 @@ export function useGridParamsApi(apiRef: GridApiRef) {
           api: apiRef.current,
         });
       }
+      if (colDef.valueExportFormatter) {
+        params.exportedValue = colDef.valueExportFormatter({
+          id,
+          field: params.field,
+          value: params.value,
+          api: apiRef.current,
```
m4theushw, line 164: I would avoid adding the `api` param in new callbacks.

m4theushw (commented):
I'm not sure if we need this new prop. Maybe we could first fix the generated columns that miss a `valueForamtter`. https://github.com/mui/mui-x/issues/3806#issuecomment-1034431625

PR #3902, opened 2022-02-09
[DataGrid] Send warning when the `rowCount` is provided with server

### LINKED ISSUE #3602: [DatagridPro] Using GraphQl and Apollo Graphql using pagination does not work
### Duplicates

### Latest version

### Current behavior 😯

I set up a paginated server with data and  want to display the data in the grid

When i press the > next page arrow, the page jumps to 0 (even if controlled).

It happens every-time i go to a new page.

### Expected behavior 🤔

I should be able to control pages via page prop.
When clicking next, it should move to the next page

### Steps to reproduce 🕹

Steps:

1.  Create a datagridPro where data can be undefined during searches
2. Set 
  pagination
  paginationMode="server"
  set incoming data with pagination
  OnPageChange callback

  and most importantly
  rowCount={data?.pageInfo.totalCount}

3. Observe the pagination jumping back to page one.

The reason is, because data is undefined between calls, row count is undefined and jumps the page back to 0. This is default Apollo behavior.

Could you add a fix to not change the rowCount until loading is done (if provided), or keep the last known good value.

Another alternative would be to just worn users. I lost a good few hours thinking it was a problem with my pagination call or page setting.
Cheers!

### Context 🔦

Doing a graphql with datagrid pro

### Your environment 🌎

<details>

```
using Brave

  System:
    OS: macOS 12.1
  Binaries:
    Node: 16.13.1 - /usr/local/bin/node
    Yarn: 3.1.1 - /usr/local/bin/yarn
    npm: 8.1.2 - /usr/local/bin/npm
  Browsers:
    Chrome: Not Found
    Edge: Not Found
    Firefox: Not Found
    Safari: 15.2
  npmPackages:
    @mui/base:  5.0.0-alpha.64 
    @mui/icons-material: ^5.2.5 => 5.2.5 
    @mui/lab: ^5.0.0-alpha.64 => 5.0.0-alpha.64 
    @mui/material: ^5.2.8 => 5.2.8 
    @mui/private-theming:  5.2.3 
    @mui/styled-engine:  5.2.6 
    @mui/system:  5.2.8 
    @mui/types:  7.1.0 
    @mui/utils:  5.2.3 
    @mui/x-data-grid-pro: ^5.2.2 => 5.2.2 
    @mui/x-license-pro:  5.2.2 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    typescript: ^4.5.4 => 4.5.4 
```
</details>

### Order ID 💳 (optional)

#35493

alexfauquette:
Fix #3602

The first commit is made to send a warning if the `paginationMode="server"` and the `rowCount` is `undefined`. This should help developers to fix the bug.

About the default behavior of Appolo (having data undefined during loading), the only solution I found is to keep the previous `rowCount` when it moves from a number to `undefined`. I coded it for discussion, but I think it's better to ask for developers to provide clean props than desynchronize the state and the props of the DataGrid


packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -147,12 +151,29 @@ export const useGridPage = (
 
   useGridApiEventHandler(apiRef, GridEvents.pageSizeChange, handlePageSizeChange);
 
+  const noRowCountInServerMode = buildWarning([
```
flaviendelangle, line 154: Should probably be outside of the render

packages/grid/_modules_/grid/utils/warning.ts
```diff
@@ -0,0 +1,12 @@
+export const buildWarning = (message: string | string[]) => {
```
flaviendelangle, line 1: In `wrapWithDeprecationWarning` I am doing the env check inside the method, which I think makes the product-code more readable.

And I propose we move the `wrapWithDeprecationWarning` in this file as well and rename it `wrapWithWarning` as there goal is very similar.

alexfauquette: You can do so because you don't have a if condition to verify before sending the warning.

Since it is a deprecated function, as soon as the function is called you log the message saying "hey it's deprecated"

Here it s not the case, I have a custome `if` between the env check and the warning. So the wrapper will need to get a function computing the condition in which warning must be logged a bit like that, but not sure it makes the code clearer

```js
export const buildConsistencyWarning = (message: string | string[], warningCondition: () => boolean) => {
  if (process.env.NODE_ENV === 'production') {
    return () => null
  }
  let alreadyWarned = false;

  const cleanMessage = Array.isArray(message) ? message.join('\n') : message;

  return (...params) => {
    if (!alreadyWarned && warningCondition(...params)) {
      alreadyWarned = true;
      console.warn(cleanMessage);
    }
  };
};

```

flaviendelangle: It has the benefits of hiding the `env` topic in a single file and be sure to never forget it.
But as you like, I agree that the gain is less clear that for the deprecation for the reason you explaied.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -150,9 +158,23 @@ export const useGridPage = (
...
+        noRowCountInServerMode();
+      }
+    }
+  }, [props.rowCount, props.paginationMode]);
+
   React.useEffect(() => {
     apiRef.current.setState((state) => {
-      const rowCount = props.rowCount !== undefined ? props.rowCount : visibleTopLevelRowCount;
+      let rowCount;
+      if (props.paginationMode === 'server') {
```
flaviendelangle, line 172: Would the following be clearer ?

```suggestion
      if (props.rowCount !== undefined) {
        rowCount = props.rowCount
      } else if (props.paginationMode === 'server') {
        rowCount = state.pagination.rowCount
      } else {
        rowCount = visibleTopLevelRowCount
      }
```

flaviendelangle:
I don't know what the best behavior is here.
Could we consider that when `loading={true}` we don't reset some props when they switch to undefined ?

alexfauquette:
I'm not in favor of using `loading` prop to adapt the internal behavior.

For now, `loading` is only about displaying the loading overlay. Not adding custom logic in it allows developers to use it with no side effects.

packages/grid/_modules_/grid/hooks/features/pagination/useGridPage.ts
```diff
@@ -151,9 +159,25 @@ export const useGridPage = (
   /**
    * EFFECTS
    */
+  React.useEffect(() => {
+    if (process.env.NODE_ENV !== 'production') {
+      if (props.paginationMode === 'server' && props.rowCount == null) {
+        noRowCountInServerMode();
```
DanailH, line 165: Should we only warn once? We seem to be doing it in the other places.

flaviendelangle: The check to only warn once is hidden inside `buildWarning`

flaviendelangle:
>  I think it's better to ask for developers to provide clean props than desynchronize the state and the props of the DataGrid

Agreed

packages/grid/x-data-grid/src/internals/hooks/features/pagination/useGridPage.ts
```diff
@@ -44,13 +45,20 @@ const mergeStateWithPage =
     }),
   });
 
+const noRowCountInServerMode = buildWarning([
+  "MUI: the 'rowCount' prop is undefined will using paginationMode='server'",
```
flaviendelangle, line 49:
```suggestion
  "MUI: the 'rowCount' prop is undefined while using paginationMode='server'",
```

packages/grid/x-data-grid/src/internals/hooks/features/pagination/useGridPage.ts
```diff
@@ -160,9 +168,25 @@ export const useGridPage = (
...
+  React.useEffect(() => {
+    if (process.env.NODE_ENV !== 'production') {
+      if (props.paginationMode === 'server' && props.rowCount == null) {
+        noRowCountInServerMode();
+      }
+    }
+  }, [props.rowCount, props.paginationMode]);
+
   React.useEffect(() => {
     apiRef.current.setState((state) => {
```
flaviendelangle, line 180: I suggest to: 

- Remove this logic
- Add a doc note on the pagination server saying that if the `rowCount` switch to undefined while loading, users must to the trick you implemented here
- Implement the same trick but in our server pagination demo (and change the fake server behavior to return "undefined" while loading
- Change the warning to link to the doc note

alexfauquette: 🤯 That's a mind-blowing suggestion: 
creating a horrible demo to let users know how to do 😄

flaviendelangle: "Horrible", it's the Apollo Client behavior :laughing: 
But alternatively, we can add a new "Apollo Client" section. It would make sense to start giving examples with various famous data management libraries.

alexfauquette:
I moved the message to the level of an `'error'` because it is more important now that we do not correct it by magic.

docs/data/data-grid/pagination/pagination.md
```diff
@@ -95,6 +95,20 @@ If your dataset is too big, and you only want to fetch the current page, you can
 - Provide a `rowCount` prop to let the grid know how many pages there are
 - Add an `onPageChange` callback to load the rows when the page changes
 
+Since `rowCount` prop is used to compute the number of available pages, switching it to `undefined` during loading reset page to zero.
+To avoid this problem, we recommend to keep the previous value of `rowCount` during the loading as follow:
```
flaviendelangle, line 99:
```suggestion
To avoid this problem, we recommend to keep the previous value of `rowCount` while loading as follow:
```

PR #3908, opened 2022-02-10
[DataGridPro] Always export the pageSize and page when it has been initialized or is being controlled

### LINKED ISSUE #3905: [DataGridPro] Should take into account custom initialState when exporting pageSize
### Duplicates

### Latest version

### Current behavior 😯

If the user selects a differente `pageSize` from the default (100) it won't be exported when using the `apiRef.current.exportState()` method.

If we set a default pageSize different from 100 using the `initialState` prop (take for example 50) the export function will be exporting this

```
pagination: {
    pageSize: 50
}
```

while if we change the pageSize manually to 100 it won't export anything.

See https://github.com/mui/mui-x/pull/3593/files#diff-0284f2cbf3405ea8b1f0ff9c6c16cda41e0e330d1f459b9336e1ab0f3a4c7098R98

### Expected behavior 🤔

`exportState` should export the pageSize prop when it's different from the default but this default shouldn't be hardcoded. It should take into account the `initialState.pagination.pageSize` prop if it's passed.

### Steps to reproduce 🕹

_No response_

### Context 🔦

I'm trying to export and import table state to localStorage. Right now if I default my table pageSize to 50 via initialState it will be unnecessarily exported (since it's the default) and it won't be exported if the user change the pageSize to 100.
If the user changed the pageSize to 100 it will be restored to 50 when reading the exported state from localStorage (since the configuration wasn't exported).

### Your environment 🌎

<details>

```
 System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.18.1 - ~/.nvm/versions/node/v14.18.1/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v14.18.1/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.1/bin/npm
  Browsers:
    Chrome: 98.0.4758.80
    Edge: Not Found
    Firefox: 96.0.3
    Safari: 14.1.2
  npmPackages:
    @mui/base:  5.0.0-alpha.61
    @mui/icons-material: ^5.2.5 => 5.2.5
    @mui/lab: ^5.0.0-alpha.62 => 5.0.0-alpha.62
    @mui/material: ^5.2.5 => 5.2.5
    @mui/private-theming:  5.2.3
    @mui/styled-engine:  5.2.5
    @mui/styles: ^5.2.3 => 5.2.3
    @mui/system:  5.2.5
    @mui/types:  7.1.0
    @mui/utils:  5.2.3
    @mui/x-data-grid-pro: ^5.5.0 => 5.5.0
    @mui/x-license-pro:  5.5.0
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.4.4 => 4.5.2
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fix #3905


alexfauquette (approved):
Sounds good to correct the bug

I'm wondering why the other states do not need the same modification. For example the column pinning could be: 
- start with an initial state in which there is a pinned column on the left
- user don't like pinned column, she removes it 
- the state export returns nothing because `pinnedColumnsToExport.left.length === 0 && pinnedColumnsToExport.right === true`
- she reopen the page, the pinned column is back 🙀

Why not exporting all the restorable state without caring if values are the default one or not. You provide all the data to the user. If it is for performances reasons, it could be done in the `restore`

flaviendelangle:
It was a proposal of @m4theushw
The goal was to avoid breaking changes if we change our default value (see https://github.com/mui/mui-x/pull/3593#pullrequestreview-861032062)

Maybe we could apply the logic of this PR to the other exportable states and only strip the default value if it's not being neither controlled nor initialized by the user.

alexfauquette:
Yes, that would be a solution

I'm not sure to understand the breaking change you want to avoid, here is what I get:

If we change the default value of the DataGrid on a property that respects all the following properties
- not controlled by dev
- not defined in `initialState` by dev
- not modified by end-user

Then the end-user should see the grid with the new default prop.

Is that the reason for not exporting all the substates by default?

flaviendelangle:
Yes exactly

alexfauquette:
Ok, keeping size small make sense, but the rules to know what must be exported will probably be complicated to explain to users

I propose to merge this PR and see when we will have users feedbacks

PR #3910, opened 2022-02-10
[DataGrid] Remove last filter item when no value to clean and close the filter panel

### LINKED ISSUE #3841: Cannot clear filters in MUI Datagrid
### Duplicates

### Latest version

### Current behavior 😯

When a filter is set without a value (only column and operators) it is not possible to clear the filters. This is also broken in the [documentation of datagrid](https://mui.com/components/data-grid/filtering/#custom-filter-panel) (click filters -> set operators to "is empty" -> try to clear the filters).

### Expected behavior 🤔

You should be able to clear the filter when clicking on the clear filter button.

### Steps to reproduce 🕹

Steps:

1. Go to [code sandbox](https://codesandbox.io/s/datagrid-v5-quick-start-forked-seycm?file=/src/App.tsx).
2. Try to clear the predefined filter.

### Context 🔦

In our project we want a predefined filter on rows that have no value for a certain column. The people working with this table are now unable to clear the predefined filter. A workaround is to select another operator, this will cause the filter to clear.

Possibly related to https://github.com/mui-org/material-ui-x/issues/3670 and https://github.com/mui-org/material-ui-x/issues/2940.

### Your environment 🌎

<details>

```
  System:
    OS: Linux 5.16 Arch Linux
  Binaries:
    Node: 14.17.3 - ~/.nvm/versions/node/v14.17.3/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.17.3/bin/yarn
    npm: 6.14.13 - ~/.nvm/versions/node/v14.17.3/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: 96.0.3
  npmPackages:
    @emotion/react: 11.4.1 => 11.4.1 
    @emotion/styled: 11.3.0 => 11.3.0 
    @mui/base:  5.0.0-alpha.67 
    @mui/material: ^5.4.0 => 5.4.0 
    @mui/private-theming:  5.3.0 
    @mui/styled-engine:  5.3.0 
    @mui/system:  5.4.0 
    @mui/types:  7.1.0 
    @mui/utils:  5.3.0 
    @mui/x-data-grid: ^5.5.0 => 5.5.0 
    @types/react: ^17.0.0 => 17.0.19 
    react: ^17.0.2 => 17.0.2 
    react-dom: ^17.0.2 => 17.0.2 
    typescript: ^4.1.2 => 4.3.5 

```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix  #3841

It modifies the behavior of the filter pannel by closing it when deleting the last item. But I see that as an improvement instead of a breaking change


packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -149,7 +149,12 @@ function GridFilterForm(props: GridFilterFormProps) {
 
   const handleDeleteFilter = () => {
     if (rootProps.disableMultipleColumnsFiltering) {
-      applyFilterChanges({ ...item, value: undefined });
+      if (item.value === undefined) {
+        deleteFilter(item);
+      } else {
+        // TODO v6: simplify the behavior by always removing the filter form
```
flaviendelangle, line 155:
```suggestion
        // TODO v6: simplify the behavior by always remove the filter form
```

flaviendelangle (approved):

DanailH (approved):

packages/grid/x-data-grid/src/internals/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -243,7 +243,12 @@ function GridFilterForm(props: GridFilterFormProps) {
 
   const handleDeleteFilter = () => {
     if (rootProps.disableMultipleColumnsFiltering) {
-      applyFilterChanges({ ...item, value: undefined });
+      if (item.value === undefined) {
```
m4theushw, line 246: Note that it only deletes the last item when clicking twice. In a string column, the empty input means `item.value=""` which falls into the else branch.

msedge_dvGxFA5EbL

packages/grid/x-data-grid/src/internals/components/panel/filterPanel/GridFilterPanel.tsx
```diff
@@ -93,9 +93,13 @@ function GridFilterPanel(props: GridFilterPanelProps) {
 
   const deleteFilter = React.useCallback(
     (item: GridFilterItem) => {
+      const shouldCloseFilterPanel = items.length === 1;
       apiRef.current.deleteFilterItem(item);
+      if (shouldCloseFilterPanel) {
+        apiRef.current.hideFilterPanel();
```
m4theushw, line 99: If the developer is controlling the filter items through the `filterModel` prop we don't know if he accept the new value passed via `onFilterModelChange`. In this case we can't assume that when `hideFilterPanel` is called it will have 0 filter items. I had the same problem with the `editRowsModel` prop triggering synchronous actions after changing the prop value.

packages/grid/x-data-grid/src/internals/components/panel/filterPanel/GridFilterForm.tsx
alexfauquette: And I assume the isAnyOf after being cleaned is an empty array, so we have two more cases to check, but we can not use `!item.value` due to boolean filters

packages/grid/x-data-grid/src/internals/components/panel/filterPanel/GridFilterPanel.tsx
alexfauquette: I forget it was controllable. We could track `items.length` and close the pannel when reaching 0

PR #3911, opened 2022-02-10
[docs] Add example of custom filtering operator based on built-in ones

flaviendelangle:
Part of #3895

I took the example of #3895 but if you have a better idea of how we can document how to re-use built in operators to add custom behavior.


joserodolfofreitas (approved):
Nice example!

PR #3912, opened 2022-02-10
[DataGrid] Fix `Commodity` and `Employee` CSV export of the `country` column

### LINKED ISSUE #3806: [DataGrid] CSV export does not work for the `Commodity` and `Employee` data sets `country` column

DanailH:
Fixes #3806 

I also added a `console.warn` to the `serialiseRow` function to ensure that the developer is notified about this requirement.


alexfauquette (approved):
I'm wondering if we could not warn developers as soon as they define a `renderCell` without a `formattedValue`. For those who do not use the export feature or the clipboard copy, it would be on useless warning. But for others, it teaches good practice, highlight the problem before a user do it.

But for now, this is already a good improvement :)

m4theushw (approved):
Since you're improving the CSV export of the generated data, there're two small things which could be fixed:

- Boolean value are exported as `true` or `false`, which is not good
- The avatar column in the Employee could be removed from the CSV, it's exported as a hex value because of the color 👎

Otherwise #3806 is fixed.

DanailH:
I looked into fixing the 2 issues - after checking I think its best to fix them on the demos we are using rather than in the data hook itself because it may seem weird when someone is playing with the demos and there isn't a clear way how those 2 are done.
For the demos on the export page, these 2 are not visible. I can fix it on the main demo?

m4theushw:
> The avatar column in the Employee could be removed from the CSV, it's exported as a hex value because of the color

This one can be solved by only adding `disableExport` in this column I guess.

> Boolean value are exported as true or false, which is not good

I think we have three paths:
- Update all translations to label the boolean values as yes/no instead of true/false. IHMO true/false is too technical, e.g. is admin: true
- Add custom translations in all demos
- Add a custom `valueFormatter` in all generated `boolean` columns

DanailH:
Ok, I've added `disableExport` to the avatar column in the Employee dataset columns.

Also I changed the default translation for the `booleanCellTrueLabel`, `booleanCellFalseLabel` to `yes` and `no`.

PR #1722, opened 2021-05-21
[DataGrid] Fix date operators not working with date-time values

m4theushw:
Closes #1710 
To be merged after #1708

The fix for this bug involves resetting the time components of the date value before comparing it. The components we need to reset depend on the column type:
- If `date`, then resets the hours, minutes, seconds, and milliseconds
- If `dateTime`, then resets only seconds and milliseconds.

I also took the opportunity to remove a lot of duplicated code and move it to an abstraction.

PR #3915, opened 2022-02-10
[DataGrid] Add specific label for linkOperator

### LINKED ISSUE #4061: [DataGrid] `GridFilterPanel` UI Issue
### Duplicates

### Latest version

### Current behavior 😯

Outlined variant doesn't work correctly. 

While setting the variant of linkOperatorInputProps in FilterPanel, as outlined, the label cuts the outlined border. 

### Expected behavior 🤔

The variant outlined should not cut the border for the grid Filter Panel. 

### Steps to reproduce 🕹

Steps:

1. Override componentProps for filter Panel
2. Set variant as outlined for linkOperatorInputProps
3. The label will cut the border

```
componentsProps={{
filterPanel: {
              filterFormProps: {
                linkOperatorInputProps: {
                  variant: 'outlined',
                  size: 'small',
                  sx: {mt: 'auto'},
                },
              },
           }
     }}
```

https://codesandbox.io/s/customfilterpanelcontent-material-demo-forked-948jwd?file=/demo.js:1865-1887

### Context 🔦

I am trying to keep my theme consistent and use outlined textfields all over my app.

### Your environment 🌎

<details>

```
System:
    OS: macOS 12.0.1
  Binaries:
    Node: 14.17.2 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.13 - /usr/local/bin/npm
  Browsers:
    Chrome: 98.0.4758.109
    Edge: Not Found
    Firefox: Not Found
    Safari: 15.1
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #4061

Preview: https://deploy-preview-3915--material-ui-x.netlify.app/components/data-grid/

The demo has two problem:

Screenshot 2022-02-10 at 14 42 25

### Pb 1

The link operator has the label "Operators". Thius bug is here from a long time ago. I create a key for linkOperator, and let it empty in all languages, such that dev can customize it.

### Pb 2

passing required to the input value does not show the star, so I simply removed it from the demo


packages/grid/_modules_/grid/constants/localeTextConstants.ts
```diff
@@ -41,6 +41,7 @@ export const GRID_DEFAULT_LOCALE_TEXT: GridLocaleText = {
   // Filter panel text
   filterPanelAddFilter: 'Add filter',
   filterPanelDeleteIconLabel: 'Delete',
+  filterPanelLinkOperator: '',
```
m4theushw, line 44: We need to add a label. Keeping it empty violates https://www.w3.org/TR/using-aria/#fifthrule

The correct name would be "logic operator" but it's too long. A benchmark might be necessary. For me, "operator" is perfect and "operators" (referring to those like "contains") should be changed to  "rule" or "condition". But if we change then we lose sync with the translation constant. Anyway, the API uses some confusing names for the filter model: https://github.com/mui/mui-x/pull/1722#discussion_r640583455

packages/grid/_modules_/grid/constants/localeTextConstants.ts
```diff
@@ -41,6 +41,7 @@ export const GRID_DEFAULT_LOCALE_TEXT: GridLocaleText = {
...
   filterPanelOperators: 'Operators',
```
m4theushw, line 45: I don't know why it's in plural but it refers to the current value. The same for "Columns".

```suggestion
  filterPanelOperators: 'Operator',
```

packages/grid/_modules_/grid/constants/localeTextConstants.ts
alexfauquette: "logic operator" seems good. From what I see in other tables, they nether display the label (And/Or) is clear enough, except notion which adds an helper: "All filters much match"

image

We can add it to the aria-label of the input, and not show it on top of the input

docs/data/data-grid/filtering/CustomFilterPanelContent.js
```diff
@@ -65,9 +65,6 @@ export default function CustomFilterPanelContent() {
                 size: 'small',
                 sx: { mt: 'auto' },
               },
-              valueInputProps: {
```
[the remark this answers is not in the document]
alexfauquette: Not easily, because the value input component can be any thing developer wants.

By default, we set `variant="standard"` in GridFilterInputValue
https://github.com/alexfauquette/material-ui-x/blob/refacto-export-button/packages/grid/_modules_/grid/components/panel/filterPanel/GridFilterInputValue.tsx#L128:L128

To customize them, developers can [edit operators](https://mui.com/components/data-grid/filtering/#create-a-custom-operator). To make them outlined thy should add to the default operators `InputComponentProps: {variant: "outlined"}`

packages/grid/_modules_/grid/constants/localeTextConstants.ts
m4theushw: Removing the label may not be ideal, there's no demo without it. We could experiment with the [ToggleButton](https://mui.com/components/toggle-button/). It doesn't require a visible label. There's also a demo using a [Menu](https://mui.com/components/menus/#selected-menu).

alexfauquette: I don't think ToggleButton is a good option. We already have a lot of information on the page. Showing both "And" and "Or" options will add confusion.
About the `Menu` demo, I don't see the difference. 

The demo is using a combination of `aria-label` and `aria-labelledby` to compensate the missing label, plus some `role: 'listbox'`, `aria-haspopup="listbox"`, ... to respect a11y.

The current solution add `aria-label` to compensate the missing label, and we use meaning full HTML (`select` and `option`)  to let the screen reader understand the structure.

If your point is that [selects accessibility](https://mui.com/components/selects/#accessibility) does not contain demo without label, I think the topic of this section is mostly how to add a label on Select component

m4theushw: It may look a bit weird if the user changes the variant to `filled` but the current approach fixes the accessibility.

image

m4theushw (commented):
The "margin" between the logic operators and the columns select is missing. Actually it's not a margin but only the select not taking the full width. Now we need to add a margin to compensate.

In https://deploy-preview-3915--material-ui-x.netlify.app/components/data-grid/

image

In https://mui.com/components/data-grid/demo/

image

packages/grid/_modules_/grid/constants/localeTextConstants.ts
alexfauquette: In such a case it would be nice to have a way to add a label, but I do not see how to make that easy to find for developers. 

What do you think about wanting to see if some users require to display the label?

docs/data/data-grid/filtering/CustomFilterPanelContent.js
```diff
@@ -65,9 +69,6 @@ export default function CustomFilterPanelContent() {
...
-                required: true,
-              },
               deleteIconProps: {
                 sx: {
                   '& .MuiSvgIcon-root': { color: '#d32f2f' },
```
Hameezr, line 74: Is there a way to change the Icon of your own? Maybe adding a prop would for the delete icon would be helpful

alexfauquette: That's out of the focus of this PR. I opened issue #4063 to add this icon to the slot of icons that can be override

m4theushw (approved):

packages/grid/x-data-grid/src/constants/localeTextConstants.ts
```diff
@@ -41,7 +41,8 @@ export const GRID_DEFAULT_LOCALE_TEXT: GridLocaleText = {
   // Filter panel text
   filterPanelAddFilter: 'Add filter',
   filterPanelDeleteIconLabel: 'Delete',
-  filterPanelOperators: 'Operators',
+  filterPanelLinkOperator: 'Logic operator',
+  filterPanelOperators: 'Operator',
```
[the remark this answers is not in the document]
m4theushw: `yarn l10n` messed with the TODO comment in other locales. Maybe add it before:

```suggestion
  // TODO v6: rename to filterPanelOperator
  filterPanelOperators: 'Operator',
```

Issue #3770, opened 2022-01-28
[data grid][test] Allow to take more screenshots of the Print Export


m4theushw:
### Summary 💡

Due to the fact that we relay on the print preview window to test the Print Export and that we reuse the same Playwright `Page` instance across all tests, there's no way to close the print preview to take another screenshot. We can't have more than one test checking the print export.

https://github.com/mui-org/material-ui-x/blob/6985fa5c6c7d1e02039fa06dfec139888672f2cb/test/regressions/index.test.js#L157

We should improve the test suite by re-organizing it in a way where each print export test uses a dedicate `Page` instance. While the remaining tests could share the same instance.

### Motivation 🔦

We should have more visual tests for the Print Export because currently we only have one and this one is very simple, not using any complex feature (row grouping or master/detail).

To not have to use the CI to test, you can run the regressions tests locally with the command below:

```sh
docker run --rm -w="/tmp/material-ui-x" -e NODE_ENV=development -v=$(pwd):/tmp/material-ui-x mcr.microsoft.com/playwright@sha256:acd388b314cfc406b227af7888264c29588e3629eafe467ece0f989fa65c38ba sh -c "xvfb-run yarn test:regressions"
```

The screenshots will be saved in `test/regressions/screenshots/chrome` as usual.

michelengelen:
This was an internal note about the 2022 regression suite, which reused one Playwright page across all tests and therefore capped print export at a single screenshot. That suite has since been rewritten, and the task was never picked up in over three years. More print export coverage remains a fair idea, but if we invest there we will scope it against the current infrastructure rather than this description.

Closing as not planned.

PR #3917, opened 2022-02-10
[DataGrid] Fix column selection for print export

### LINKED ISSUE #3904: [RFC] Print and CSV have same options but different behavior
## The current API 

To select exported columns, developers can use 
- `printOptions.fields` / `csvOptions.fields`
- `printOptions.allColumns` / `csvOptions.allColumns`
- `colDef.disableExport`

And the doc for [csv](https://mui.com/components/data-grid/export/#custom-exported-content) and [print](https://mui.com/components/data-grid/export/#custom-exported-content-2) are the same

## The problem

The way these options are applied is different for print and CSV

### CSV rules

if `csvOptions.fields` is defined, this is the authority. 
if `allColumns=true` we consider all the columns. Otherwise, only visible columns are considered. And from these considered columns, we remove those with `disableExport=true`

### Print rules

if `allColumns=true` all the columns are printed
if `printOptions.fields` is defined we print columns in `printOptions.field` excepted if they have `disableExport=true`
if `printOptions.fields` is not defined, we print visible columns, even if `disableExport=true`

## The questions

- Should we have the same behavior? I think yes, but since print and csv are for different purposes, it's open to debate. 
- If yes which behavior should we use?

@mui/x if you have opinion or related issues

### LINKED ISSUE #3906: [DataGrid] Columns not correctly restored after printing when `columnVisibilityModel` and `printOptions.fields` is used
### Duplicates

### Latest version

### Current behavior 😯

After printing a page for the 2nd time I see wrong visible columns in the grid. This seems to only occur if the `columnVisibilityModel` prop is used and custom fields are also specified via `printOptions.fields`.

### Expected behavior 🤔

It should keep the same columns that were visible before printing.

### Steps to reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/exportdefaulttoolbar-material-demo-forked-39bx2?file=/demo.tsx:740-752
2. Click Export > Print
3. Cancel the print
4. Click again Export > Print
5. Cancel the print
6. It only shows the "id" column, which is hidden according to `columnVisibilityModel`.

image

### Context 🔦

I was playing with the options in #3904.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3906
Fix #3904

Preview: https://deploy-preview-3917--material-ui-x.netlify.app/components/data-grid/export/#default-toolbar

Todo

- [x] clarify the doc on exported columns
- [x] verify it works with both visibility model and `hiden` property
- [x] Add tests
	- [x] column visibility restoration (form `visibilityModel` and `colDef.hide`)
	- [x] column selection paramteres


alexfauquette:
I struggle to test, because print feature is for now only tested by a screenshot.

- For the aspect "print and csv columns selection are the same" not sure it's necessary to test, because they now use the same function.
- For the issue #3906 I'm wondering if I should not simply add a screenshot of the grid after the print, to verify restoring column swork

m4theushw:
> I'm wondering if I should not simply add a screenshot of the grid after the print, to verify restoring column swork

You can't. Once the print window is open we can't run any other test nor close the window. I opened #3770 to allow to take other screenshots but it won't help here because we need the grid after printing and through Playwright we can't click the "Print" button (it's in a different window). One option is to have an experimental flag for a "silent print mode" where we don't call `window.print()` and just restore the columns as they were before.

alexfauquette:
I made a try on the testing solution.

I tried with the "silent print mode", but tests with Playwright use the production build, so options I pass to it will also impact users.

So I preferred to use `process.env.NODE_ENV` to modify the behavior of the print feature in tests. I used it two times

First to remove the print window
```js
if (process.env.NODE_ENV !== 'test') {
	printWindow.contentWindow!.print();
}
```

Second to remove the need of events `onload` and `onafterprint`
```js
if (process.env.NODE_ENV === 'test') {
        // In test env, run the all pipeline without waiting for loading
        handlePrintWindowLoad(printWindow, options);
        handlePrintWindowAfterPrint(printWindow);
      } else {
        printWindow.onload = () => handlePrintWindowLoad(printWindow, options);
        printWindow.contentWindow!.onafterprint = () => handlePrintWindowAfterPrint(printWindow);
      }
```

Thanks to these modifications, it is possible to observe the DataGrid printed and its restoration, by listening to events:
- first modification is to print the grid
- second modification is for the restoration of the initial state

@m4theushw do you see a problem to this way of unlocking print testing?

DanailH:
For the testing part - why don't we add one more example to snap with a different configuration?

alexfauquette:
@DanailH To verify the exported columns are correct, we could add more snap.

But issue #3906 is about restoring the columns. Since we do not have access to the print window, we can not close it. And so the code in charge of restoring the columns is never tested.

packages/grid/_modules_/grid/hooks/features/export/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface GridGetColumnsToExportParams {
+  /**
+   * The API of the grid.
+   */
+  apiRef: React.MutableRefObject<GridApiCommunity>;
+  options: GridExportOptions;
+}
+
+export const getColumns = ({
```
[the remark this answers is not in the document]
DanailH: Either one will work but my personal preference is `getColumnsToExport`

DanailH:
Ah then yeah - not much you can do about that.

DanailH (approved):

packages/grid/x-data-grid/src/internals/hooks/core/useGridStateInitialization.ts
```diff
@@ -73,11 +73,11 @@ export const useGridStateInitialization = <Api extends GridApiCommon>(
         // Events are here to forward to other hooks and apply changes.
         // You are trying to update several states in a no isolated way.
         throw new Error(
-          `You're not allowed to update several sub-state in one transaction. You already updated ${
-            updatedControlStateIds[0]
-          }, therefore, you're not allowed to update ${updatedControlStateIds.join(
-            ', ',
-          )} in the same transaction.`,
+          `You're not allowed to update several sub-state in one transaction. You already updated ${JSON.stringify(
```
flaviendelangle, line 76: Why do we need this ?

packages/grid/x-data-grid/src/internals/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -227,28 +230,25 @@ export const useGridPrintExport = (
       doc.current!.body.removeChild(printWindow);
 
       // Revert grid to previous state
-      apiRef.current.setState((state) => ({
-        ...state,
-        ...previousGridState.current,
-      }));
 
-      apiRef.current.unstable_enableVirtualization();
+      // We can not simply use setState(...previousGridState.current) because of this verification
```
flaviendelangle, line 234: Could we use `apiRef.current.restoreState` once #3816 is merged ?
It has a small wrapper that avoids to run several `forceUpdate`.

packages/grid/x-data-grid/src/internals/hooks/features/export/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const getColumnsToExport = ({
+  apiRef,
+  options,
+}: GridGetColumnsToExportParams): GridStateColDef[] => {
+  const columns = allGridColumnsSelector(apiRef);
+
+  let exportedColumns: GridStateColDef[];
+  if (options.fields) {
+    exportedColumns = options.fields
```
flaviendelangle, line 22: We can remove the `let` and just return inside the `if`

flaviendelangle (commented):
For the `process.env.NODE_ENV === 'test'` I don't feel confident since I did not work on the print testing.

@m4theushw and @DanailH should be able to answer better than me.

packages/grid/x-data-grid/src/internals/hooks/features/export/useGridPrintExport.tsx
flaviendelangle: In fact I think you can already use it

packages/grid/x-data-grid/src/internals/hooks/core/useGridStateInitialization.ts
alexfauquette: Because I get in my console 

> You're not allowed to update several sub-state in one transaction. You already updated [object Object]

flaviendelangle: Oh right I turned it into an object some time ago.
But the `hasPropChanged` has no use here

```suggestion
        throw new Error(
          `You're not allowed to update several sub-state in one transaction. You already updated ${
            updatedControlStateIds[0].stateId
          }, therefore, you're not allowed to update ${updatedControlStateIds.join(
            ', ',
          ).map(el => el.id)} in the same transaction.`,
        );
```

alexfauquette: It's out of the PR topic, but why is there a `updatedControlStateIds.forEach` wherase you check juste before the length is at most 1? Would it make sense to use `console.error` or to remove the forEach?

```js
if (updatedControlStateIds.length > 1){
	throw Error(
		// ...
	)
}
// ...
updatedControlStateIds.forEach(({ stateId, hasPropChanged }) => { 
	// ...
})
```

packages/grid/x-data-grid/src/internals/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -43,7 +44,7 @@ export const useGridPrintExport = (
   const columns = useGridSelector(apiRef, allGridColumnsSelector);
   const doc = React.useRef<Document | null>(null);
   const previousGridState = React.useRef<any>();
```
flaviendelangle, line 46:
```suggestion
  const previousGridState = React.useRef<GridInitialStateCommunity>();
```

flaviendelangle (approved):

packages/grid/x-data-grid/src/internals/hooks/core/useGridStateInitialization.ts
flaviendelangle: You are right, we could probably just run on the 1st item instead of looping

alexfauquette:
@flaviendelangle I thought failing test was flaky tests, but it was the restore state  method that do no work with `colDef.hide`

The `exportState()` does export columnVisibilityModel inly if it is specified as a DataGrid `prop` or in `initialState`. So I have to do the restoration manually with `setColumnVisibilityModel` and `gridColumnVisibilityModelSelector`

My assumption is that if I update and restore column visibility by using `setColumnVisibilityModel` it will not create an inconsistency with the `colDef.hide` because
- If `setColumnVisibilityModel` has an impact on the `hide` property, it will be restored by the same logic
- If `setColumnVisibilityModel` has no impact on the `hide` property, then no problem

The correction is in the last commit

By the way, that's an argument in favor of exporting `columnVisibility` even if not controlled.

flaviendelangle:
Indeed...
Well, it was too good to be true

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
+import * as React from 'react';
+import { DataGrid, GridToolbarExport, GridRowsProp, GridColumns } from '@mui/x-data-grid';
```
m4theushw, line 2:
```suggestion
import { DataGridPro, GridToolbarExport, GridRowsProp, GridColumns } from '@mui/x-data-grid-pro';
```

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+              }}
+            />
+          </div>,
+        );
+        expect(onColumnVisibilityModelChange.callCount).to.equal(0);
+
+        // Trigger print
+        fireEvent.click(screen.queryByRole('button', { name: 'Export' }));
+        fireEvent.click(screen.queryByRole('menuitem', { name: 'Print' }));
+        await clock.runToLast();
```
m4theushw, line 84: Two things:

- `clock.runToLast` doesn't return a promise. You can use a hack to wait for the promise:
https://github.com/mui/mui-x/blob/acb00d2005b3855de60b82e8b39d3325f8b21b50/packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.test.tsx#L817
- Do we need to mock the time here? Replace `nativeSetTimeout` with `setTimeout` if the time is real in the above hack.

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+        // verify column visibility has been restored
+        expect(onColumnVisibilityModelChange.secondCall.firstArg).to.deep.equal({
+          col1: gridVisible,
+          col2: false,
+        });
+      });
+    });
+  });
+  describe('column selection', () => {
```
m4theushw, line 146: "selection" remembers me of checkbox selection.

```suggestion

  describe('columns to print', () => {
```

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+                Toolbar: () => (
+                  <GridToolbarExport
+                    printOptions={{ fields: printVisible ? ['col1', 'col2'] : ['col2'] }}
+                  />
+                ),
+              }}
+              initialState={{
+                columns: {
+                  columnVisibilityModel: {
+                    // Hide columns status and traderName, the other columns will remain visible
```
m4theushw, line 70: These columns don't exist.

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+      // Need layouting
+      this.skip();
+    }
+  });
+
+  describe('restore visibility', () => {
+    allBooleanConfigurations.forEach(({ printVisible, gridVisible }) => {
+      it(`should set and restore 'col1' from ${printVisible ? "'visible'" : "'hidden'"} to ${
+        gridVisible ? "'visible'" : "'hidden'"
+      } (visibilityModel)`, async function test() {
```
m4theushw, line 52: Ok there's too much information in the name of this test case. I would first create two `describe` blocks: one to test the `columnVisibilityModel` prop and another for the `hide` attribute. Then, I would detail better what are the replacements. At first time I had to look at the variable name to know what the first "visible" / "hidden" is referring to. The final test name is not too meaningful.

```
should set and restore 'col1' from 'visible' to 'hidden'
```

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { DataGrid, GridToolbarExport, GridRowsProp, GridColumns } from '@mui/x-data-grid';
+import { createRenderer, screen, fireEvent } from '@mui/monorepo/test/utils';
+import { expect } from 'chai';
+import { spy } from 'sinon';
+
+describe('<DataGridPro /> - Export - Print', () => {
+  const { render, clock } = createRenderer({ clock: 'fake' });
+
+  const rows: GridRowsProp = [
+    { id: 1, col1: '1-1', col2: '1-2' },
```
m4theushw, line 11: Could you use `getData` or `useData` instead? I think we were avoiding hardcoded data in tests.

https://github.com/mui/mui-x/blob/acb00d2005b3855de60b82e8b39d3325f8b21b50/packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx#L37

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
alexfauquette: Ok, then I don't know why adding the wait makes tests pass

As a solution, I propose to use the `apiRef.current.exportDataAsPrint()` instead of clicking such that I can put an `await` to be sure the method exit

I add another test, to verify that the method receive the correct parameters when Print button is clicked

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
alexfauquette: Now, the message is the following one

```
should have 'currencyPair' 'visible' in print and 'hidden' in screen
```

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
alexfauquette: I do not manage to listen at the `apiRef.current.exportDataAsPrint` using spy, so I can not verify it receive the correct arguments

flaviendelangle:
Can we merge this one before releasing v5.6 ?

alexfauquette:
@flaviendelangle Yes, feedback from @m4theushw was about tests, which can be improved in another PR

packages/grid/x-data-grid-pro/src/tests/exportPrint.DataGrid.test.tsx
```diff
@@ -0,0 +1,262 @@
+import * as React from 'react';
+import { DataGrid, GridToolbarExport, GridRowsProp, GridColumns } from '@mui/x-data-grid';
+import { createRenderer, screen, fireEvent } from '@mui/monorepo/test/utils';
+import { expect } from 'chai';
+import { spy } from 'sinon';
+
+describe('<DataGridPro /> - Export - Print', () => {
```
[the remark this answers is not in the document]
m4theushw: I've unresolved because the file was not renamed.

m4theushw (approved):

packages/grid/x-data-grid/src/hooks/features/export/useGridPrintExport.tsx
```diff
@@ -53,27 +55,20 @@ export const useGridPrintExport = (
           return;
         }
 
-        const columnVisibilityModel = gridColumnVisibilityModelSelector(apiRef);
-        const columns = gridColumnDefinitionsSelector(apiRef);
+        const exportedColumnFields = getColumnsToExport({
```
mjomble, line 58: @alexfauquette I ran into an issue here.

I set `disableExport: true` on a column, but it still showed up in the print document.
I discovered that `getColumnsToExport` was never called because I didn't specify `fields` or `allColumns`, so this function returned a few rows above (`if (!fields && !allColumns) { ... return;`)

It seems to me like `disableExport` should still be checked even if `fields` and `allColumns` are not used.

PR #3930, opened 2022-02-12
[I10n] Improve Hebrew (he-IL) locale

ColdAtNight:
Add missing translations & fixes
I am not sure why the check fails. I ran all required scripts before pushing the changes.

### DIFF
--- a/packages/grid/_modules_/grid/locales/heIL.ts
+++ b/packages/grid/_modules_/grid/locales/heIL.ts
@@ -90,7 +90,7 @@ const heILGrid: Partial<GridLocaleText> = {
     count !== 1 ? `${count.toLocaleString()} שורות נבחרו` : `שורה אחת נבחרה`,
 
   // Total rows footer text
-  footerTotalRows: 'סך הכל שורות: ',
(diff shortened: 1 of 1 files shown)


packages/grid/_modules_/grid/locales/heIL.ts
```diff
@@ -90,18 +90,18 @@ const heILGrid: Partial<GridLocaleText> = {
     count !== 1 ? `${count.toLocaleString()} שורות נבחרו` : `שורה אחת נבחרה`,
 
   // Total rows footer text
-  footerTotalRows: 'סך הכל שורות: ',
```
alexfauquette, line 93: Are you sure you want to remove the space after `":"`?

alexfauquette (approved):
Seems that some tests flaky tests are back. Don't worry, we will deal with them

packages/grid/_modules_/grid/locales/heIL.ts
ColdAtNight: Yes, I looked at other translations and it is the case in all of them

PR #3953, opened 2022-02-15
[core] Drop `_modules_` folder

flaviendelangle:
Follow up #3688
Part of #924
____

- [x] Adapt the prop types generation to handle several folders. I unified the TypeScript program / checker creation with `buildApi` to keep the logic in a single place.

- [x] Move all the community code from `packages/grid/_modules_/grid` to `packages/grid/x-data-grid/src/internals`

This pull request does not remove the `unstable_` exports from `@mui/x-data-grid` because we need the new bundling strategy which will come in the next Pull Request.

____

I suggest to do the deep review in #3965 in which you will have a clearer view of the releasable version


docs/scripts/getTypeScriptProjects.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Project, ProjectNames, Projects } from './api/utils';
+
+const workspaceRoot = path.resolve(__dirname, '../../');
+
+interface CreateProgramOptions {
+  name: ProjectNames;
+  rootPath: string;
+
+  /**
+   * Config to us to build this package
```
m4theushw, line 12:
```suggestion
   * Config to use to build this package.
```

All other descriptions are missing the period too.

docs/scripts/getTypeScriptProjects.ts
```diff
@@ -0,0 +1,106 @@
...
+   * The path must be relative to the root path.
+   */
+  otherComponentFiles?: string[];
+}
+
+const createProject = (options: CreateProgramOptions): Project => {
+  const { name, tsConfigPath, rootPath, entryPointPath, componentsFolder, otherComponentFiles } =
+    options;
+
+  const compilerOptions = ts.parseJsonConfigFileContent(
```
m4theushw, line 40: The current generator uses this helper: https://github.com/mui/mui-x/blob/5f6b29d6d48753ffcb555c03df07cd18e4a90cc8/docs/scripts/generateProptypes.ts#L8

It parses and returns the compiler options, also checking if there's an error.

docs/scripts/getTypeScriptProjects.ts
```diff
@@ -0,0 +1,106 @@
...
+    exports,
+    program,
+    checker,
+    workspaceRoot,
+    componentsFolder:
+      componentsFolder === undefined ? undefined : path.join(rootPath, componentsFolder),
+    otherComponentFiles:
+      otherComponentFiles === undefined
+        ? undefined
+        : otherComponentFiles.map((file) => path.join(rootPath, file)),
```
m4theushw, line 73: To avoid line breaks.

```suggestion
    componentsFolder: componentsFolder ? path.join(rootPath, componentsFolder) : undefined,
    otherComponentFiles: otherComponentFiles?.map((file) => path.join(rootPath, file)),
```

m4theushw (commented):
Looks good. Only left a few small suggestions.

docs/scripts/getTypeScriptProjects.ts
flaviendelangle: I try not to use TTP in a code that is also called for `docs:api:build`
But the logic is very simple and can be inlined.

alexfauquette (approved):

alexfauquette:
Are the conflicts easy to solve, or should we wait for this PR before merging on master to avoid the accumulation of conflicts?

flaviendelangle:
> Are the conflicts easy to solve, or should we wait for this PR before merging on master to avoid the accumulation of conflicts?

As long as you are not doing major refacto it's fine
I'm struggling with the bundling migration so I don't want to block you for days.

m4theushw (approved):

PR #3955, opened 2022-02-15
[DataGrid] Allow for truncated and multiline content in grid cells

### LINKED ISSUE #3587: [DataGrid] Text inside the cell is not truncated
### Duplicates

### Latest version

### Current behavior 😯

Text is not truncated with the ellipsis inside Data Grid Cell. As I see that happened after updating block model for the Cell (it was switched from `display: block` to `display: flex` in the latest v5.2.2 release.

Screenshot 2022-01-10 at 12 39 36

### Expected behavior 🤔

Text should be truncated with the ellipsis like it was in the previous releases

Screenshot 2022-01-10 at 12 44 07

### Steps to reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/datagrid-v5-2-2-truncated-text-7mc4p?file=/src/App.tsx
2. See that the text is not truncated at the first row

### Context 🔦

_No response_

### Your environment 🌎

see code sandbox

### Order ID 💳 (optional)

**35653**

DanailH:
Fixes #3587 

I added 2 snaps to make sure that we don't break it again.
To fix this I had to wrap the content of the cell in a `div` element. If a `renderCell` is provided its return value won't be wrapped in that `div` element.

Storybook: https://deploy-preview-3955--material-ui-x.netlify.app/storybook/?path=/story/datagridpro-test-styling--truncated-cell-text-snap


packages/grid/_modules_/grid/components/cell/GridCell.tsx
```diff
@@ -106,6 +107,7 @@ function GridCell(props: GridCellProps) {
     onKeyDown,
     onDragEnter,
     onDragOver,
+    touchRippleRef,
```
cherniavskii, line 110: It was passed to a `div` and caused warning, right?

DanailH: My bad, that is not needed - reverted. There was an issue with this but it was for the checkbox

packages/grid/_modules_/grid/components/containers/GridRootStyles.ts
```diff
@@ -226,10 +227,13 @@ export const GridRootStyles = styled('div', {
     [`& .${gridClasses.cell}`]: {
       display: 'flex',
       alignItems: 'center',
+      overflow: 'hidden',
+      borderBottom: `1px solid ${borderColor}`,
```
m4theushw, line 231: > In that case, the text won't overflow but break and fit into the cell, which actually should be the expected result but it opened for debates.

Removing `whiteSpace: nowrap` broke our demos. If it was an alpha version we could say that it's expected, but doing it now is risky.

image

DanailH: Yeah ... ok the fix the easy part is that it is fixable by just keeping the `textOverflow: 'ellipsis'` and `whiteSpace: 'nowrap'` on the cell class.

m4theushw: `textOverflow` in the cell does nothing, but `whiteSpace` should be kept.

DanailH:
@cherniavskii I was thinking of adding another option to the `GridColDef` to distinguish the 2 but without it looks cleaner. Also with it, additional configuration will probably be needed for the bug to be actually fixed. With the current solution, there are no regressions and both cases (truncated and multiline) will work.

cherniavskii:
> Also with it, additional configuration will probably be needed for the bug to be actually fixed

If you mean https://github.com/mui/mui-x/issues/3587, then it will be fixed by default with no configuration.

> With the current solution, there are no regressions

What about the one you mentioned in PR description?
> It can only impact the case where a renderCell is provided with long text content. In that case, the text won't overflow but break and fit into the cell

cherniavskii:
Adding `noWrap` option has one advantage - no need to use `renderCell` to enable text wrapping

DanailH:
> What about the one you mentioned in PR description?
> 
> > It can only impact the case where a renderCell is provided with long text content. In that case, the text won't overflow but break and fit into the cell

that is already fixed by keeping the `whiteSpace: 'nowrap'`

m4theushw:
@cherniavskii Whether we add `noWrap` or not, it could be a prop in `GridCell`. This avoids adding more options to `GridColDef` that are only used in a small part of the grid.

cherniavskii:
> that is already fixed by keeping the whiteSpace: 'nowrap'

Ok, I didn't notice that change. It's great then!

@m4theushw 
> Whether we add noWrap or not, it could be a prop in GridCell

I like this!

m4theushw (approved):
Looks good.

PR #3959, opened 2022-02-15
[DataGrid] Don't use `field` to match columns in the columns panel when `headerName` is available

### LINKED ISSUE #3927: [DataGrid] Don't use `field` to match columns in the columns panel when `headerName` is available
### Order ID 💳

25274

### Duplicates

### Latest version

### The problem in depth 🔍

We need the ability to limit the search filter in the GridColumnsPanel, that allows for us to search for available columns to hide/show. Currently the filter is performing an OR match on the column's field or the column's header test. In our use case, many of the "field" values for columns are internal names that are not meaningful to an end user. Also, our field names are static regardless of a user's locale, but the heading text will change per locale. It's a bit misleading that the "field" is automatically included in the search.

We'd like the ability to provide a custom filter to the find columns feature, or to simply turn off the "field" name checking logic.

| Field     | Header| Search Text |Current Match | Expected Match |
| ----------- | ----------- |---------- |---------- |---------- |
| internalName1      | Display 1| internal|Yes |No|
| internalName1      | Display 1| display| Yes|Yes|

I know that we can utilize the slots functionality in order to override the ColumnsPanel that is used and we can supply our own, but that's a bit heavy handed for our use case. The default columns panel is great with the exception of including the "field" in the search filter.

### Your environment 🌎

<details>

```
```
</details>

pkratz:
Fixes #3927

### DIFF
--- a/packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx
+++ b/packages/grid/_modules_/grid/components/panel/GridColumnsPanel.tsx
@@ -93,18 +93,16 @@ export function GridColumnsPanel() {
     setSearchValue(event.target.value);
   }, []);
 
-  const currentColumns = React.useMemo(
-    () =>
-      !searchValue
-        ? columns
-        : columns.filter(
-            (column) =>
-              column.field.toLowerCase().indexOf(searchValue.toLowerCase()) > -1 ||
-              (column.headerName &&
-                column.headerName.toLowerCase().indexOf(searchValue.toLowerCase()) > -1),
-          ),
-    [columns, searchValue],
-  );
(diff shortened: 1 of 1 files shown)


alexfauquette (approved):
Yes, make sense to only use the displayed title to filter columns

flaviendelangle:
Maybe we could even do it in `createColumnsState` and set `colDef.field` as the default value of `column.headerName`.
That way in the search method we only care about `headerName`

But it's a change that would require more checks to be sure not to break anything, so I'm ok with starting small.

flaviendelangle (approved):

Issue #3763, opened 2022-01-28
DataGridPro - preProcessEditCellProps called for every cell within edited row


itskemo:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Tested on DataGridPro with editMode: "row"

In Columns definition when preProcessEditCellProps is defined for multiple cells it is called for all cells right after editing a cell or exiting the edit mode, however not all the cell values has been updated. Is this an expected behaviour since it is not documented in docs - at least not in a way I could find it.

### Expected behavior 🤔

Assuming this should be called on the actual cell props change,  I'm expecting this to be called only for the cell that the props has been changed.

Additionally on row edit mode exit - I believe the callback should not be called for all the cells.

### Steps to reproduce 🕹

Steps:

1. Run the sample here: https://codesandbox.io/s/datagridpro-v5-quick-start-forked-7w171?file=/src/App.tsx
2. Open the console
3. Double click the row to enter the edit mode
4. Edit any value within the row by typing into it a single letter
5. two console logs should be outputted

### Context 🔦

Trying to implement a server-side validation on individual cell values, however it simply invokes the requests for ALL cells within the row when editing a cell value.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

alexfauquette:
That is intended for client-side validation requiring to compare different cells.

In the following demo, the `preProcessEditCellProps` is defined on the "payment method" column. And this column can display an error is you toggle the "is paid?" one without entering the payment method.
https://mui.com/components/data-grid/editing/#conditional-validation
 
![Peek 2022-01-28 10-47](https://user-images.githubusercontent.com/45398769/151524871-e2137d9e-e02a-447e-ad4f-427e1aec6be4.gif)

Notice that you have access to the row in `preProcessEditCellProps` parms. So you can compare the `props.value` to the row value
```ts
export interface GridPreProcessEditCellProps {
  /**
   * The grid row id.
   */
  id: GridRowId;
  /**
   * The row that is being editted.
   */
  row: GridRowModel;
  /**
   * The edit cell props.
   */
  props: GridEditCellProps;
}
```

Here is an adaptation of you rcodesandbox
https://codesandbox.io/s/datagridpro-v5-quick-start-forked-peyod?file=/src/App.tsx

[...]

PR #3963, opened 2022-02-16
[DataGrid] New editing API

m4theushw:
This PR proposes a new API for cell and row editing. To avoid breaking changes, it's only available if `experimentalFeatures.enableNewEditingAPI` is `true`. Since we switch hooks, developers can't change which API to use in runtime.

After working a lot with the current API it seems to me that during development the listeners were created before the API methods. This caused weird things like publishing events and also calling other API methods.

https://github.com/mui/mui-x/blob/3101d8d25e47be477f104e22af7b867919875a6f/packages/grid/x-data-grid/src/internals/hooks/features/editRows/useGridCellEditing.ts#L228-L230

But these published events also call API methods. Their only purpose is to allow developers to override the default behavior.

https://github.com/mui/mui-x/blob/3101d8d25e47be477f104e22af7b867919875a6f/packages/grid/x-data-grid/src/internals/hooks/features/editRows/useGridCellEditing.ts#L299-L301

This was making the editing hooks a mess to maintain. To fix these problems I took a different direction here. First I created the API methods, then binded them to events. Generic listeners, like `cellKeyDown`, are dumb now and they will only be used to publish a specific editing event, like `cellEditStart`. The specific event listeners will also be dumb, only calling a method from the API with the correct params. All the complex logic will be inside the API method.

### Methods for the cell editing API

- `isCellEditable` determines if a cell is editable
- `getCellMode` returns if a cell is in edit or view mode
- `startCellEditMode` puts a cell into  edit mode
- `stopCellEditMode` puts a cell into view mode and updates the original row with the value stored
- `setEditCellValue` updates the value of a cell

### Methods for the row editing API

- `getRowMode` returns if a cell is in edit or view mode
- `startRowEditMode` puts row into  edit mode
- `stopRowEditMode` puts a row into view mode and updates the original row with the values stored
- `setEditCellValue` updates the value of a cell

### New props

- `processRowUpdate` (name open for discussion) is called when putting a cell/row back to view mode. It's meant to be used to send the new row to the server. Since it's called BEFORE `updateRows` it allows developers to also pre-process the row that will be saved. If the promise is rejected, then we don't update the row and the row doesn't goes to view mode. Previously, to send the row to the server we taught developers to use the `rowEditStop` event but this had a problem that it updated the row even when the server threw an error, it was optimistic. Now we use a pessimistic approach, which IHMO is what users want for a CRUD application. See https://marmelab.com/react-admin/CreateEdit.html#undoable

### Methods to be removed in v6

- `setRowMode`
  - Why? Replaced by `startRowEditMode` and `stopRowEditMode`
- `setCellMode`
  - Why? Replaced by `startCellEditMode` and `stopCellEditMode`
- `commitRowChange`
  - Why? Functionality merged into `stopRowEditMode`
- `commitCellChange`
  - Why? Functionality merged into `stopCellEditMode`

---

Demo with server-side validation: https://codesandbox.io/s/validateservernamegrid-material-demo-forked-5iflei?file=/demo.tsx

Demo with server-side persistence: https://codesandbox.io/s/celleditserversidepersistence-material-demo-forked-7xcpgg?file=/demo.tsx (it doesn't save if the name is empty)

Demo with row editing, validation and with the default start/stop editing behaviors disabled https://codesandbox.io/s/starteditbuttongrid-material-demo-forked-y84q5e?file=/demo.tsx

The docs will be done in a follow-up PR.


packages/grid/x-data-grid-pro/src/tests/detailPanel.DataGridPro.test.tsx
```diff
@@ -158,6 +158,20 @@ describe('<DataGridPro /> - Detail panel', () => {
     expect(screen.queryByText('Detail')).to.equal(null);
   });
 
+  [...new Array(100).keys()].forEach((i) => {
```
m4theushw, line 161: Ignore. It came from #4031 when I pulled the changes.

packages/grid/x-data-grid/src/internals/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -74,6 +75,10 @@ function GridEditSingleSelectCell(props: GridRenderEditCellParams & Omit<SelectP
     setOpen(false);
     const isValid = await api.setEditCellValue({ id, field, value: event.target.value }, event);
 
+    if (rootProps.experimentalFeatures?.enableNewEditingAPI) {
+      return;
```
m4theushw, line 79: The edit cell in the current `singleSelect` column automatically commits the value after selecting an option. This doesn't work well with `preProcessEditCellProps` because if doing a AJAX validation it freeze the UI while waiting for the server to respond. Now we don't automatically commit the value. The user has to click outside as in any other column.

packages/grid/x-data-grid/src/internals/models/params/gridCellParams.ts
```diff
@@ -129,6 +129,10 @@ export interface GridValueSetterParams {
    * The row that is being edited.
    */
   row: GridRowModel;
+  /**
+   * The original row with the values before entering the edit mode.
+   */
+  originalRow?: GridRowModel;
```
flaviendelangle, line 135: Maybe document when this key is defined or not.

flaviendelangle (commented):
I did not test the behavior yet
Overall, the new API is a lot easier to read.

packages/grid/x-data-grid/src/internals/models/api/gridEditingApi.ts
```diff
@@ -126,7 +126,128 @@ export interface GridCellEditingApi extends GridEditingSharedApi {
...
+   */
+  startCellEditMode(params: GridStartCellEditModeParams): void;
+  /**
+   * Puts the cell at the given row id and field into view mode and updates the original row with the new value stored.
+   * If `params.ignoreModifications` is `false` it will discard the modifications made.
+   * @param {GridStopCellEditModeParams} params The row id and field of the cell to stop editing.
+   * @returns {Promise<boolean>} A promise which resolves with `true` if it succeeds of `false` if props are being processed or `processRowUpdate` fails.
+   * @ignore - do not document.
+   */
+  stopCellEditMode(params: GridStopCellEditModeParams): Promise<boolean>;
```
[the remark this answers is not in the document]
m4theushw: It will be documented in #4060.

packages/grid/x-data-grid/src/internals/models/api/gridEditingApi.ts
```diff
@@ -126,7 +126,128 @@ export interface GridCellEditingApi extends GridEditingSharedApi {
...
+  /**
+   * @ignore - do not document.
+   */
+  unstable_setCellEditingEditCellValue: (params: GridEditCellValueParams) => Promise<boolean>;
```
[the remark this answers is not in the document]
m4theushw: Done. This method is not meant to be called directly. `setEditCellValue` calls it internally or the row-editing equivalent one.

packages/grid/x-data-grid/src/internals/models/api/gridEditingApi.ts
```diff
@@ -126,7 +126,128 @@ export interface GridCellEditingApi extends GridEditingSharedApi {
...
+  extends GridEditingSharedApi,
+    Pick<GridCellEditingApi, 'getCellMode'> {
+  /**
+   * Puts the cell at the given row id and field into edit mode.
+   * @param {GridStartCellEditModeParams} params The row id and field of the cell to edit.
+   * @ignore - do not document.
+   */
+  startCellEditMode(params: GridStartCellEditModeParams): void;
+  /**
+   * Puts the cell at the given row id and field into view mode and updates the original row with the new value stored.
```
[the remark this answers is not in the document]
m4theushw: I only updated the description of the new methods. The old ones won't be touched anymore.

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -74,7 +74,7 @@ export interface GridRowEditingApi extends GridEditingSharedApi {
    */
   getRowMode: (id: GridRowId) => GridRowMode;
   /**
-   * Updates the row at the given id with the values stored in the edit row model.
+   * Updates the row correspoding the given id with the values stored in the edit row model.
```
flaviendelangle, line 77:
```suggestion
   * Updates the row corresponding to the given id with the values stored in the edit row model.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -94,7 +94,7 @@ export interface GridRowEditingApi extends GridEditingSharedApi {
  */
 export interface GridCellEditingApi extends GridEditingSharedApi {
   /**
-   * Updates the field at the given id with the value stored in the edit row model.
+   * Updates the field correspoding the given id with the value stored in the edit row model.
```
flaviendelangle, line 97:
```suggestion
   * Updates the field corresponding to the given id with the value stored in the edit row model.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -201,20 +201,24 @@ export interface GridNewCellEditingApi
   extends GridEditingSharedApi,
     Pick<GridCellEditingApi, 'getCellMode'> {
   /**
-   * Puts the cell at the given row id and field into edit mode.
+   * Puts the cell correspoding to the given row id and field into edit mode.
```
flaviendelangle, line 204:
```suggestion
   * Puts the cell corresponding to the given row id and field into edit mode.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -201,20 +201,24 @@ export interface GridNewCellEditingApi
...
    * @param {GridStartCellEditModeParams} params The row id and field of the cell to edit.
    * @ignore - do not document.
    */
   startCellEditMode(params: GridStartCellEditModeParams): void;
   /**
-   * Puts the cell at the given row id and field into view mode and updates the original row with the new value stored.
+   * Puts the cell correspoding to the given row id and field into view mode and updates the original row with the new value stored.
```
flaviendelangle, line 210:
```suggestion
   * Puts the cell corresponding to the given row id and field into view mode and updates the original row with the new value stored.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -224,20 +228,24 @@ export interface GridNewRowEditingApi
   extends GridEditingSharedApi,
     Pick<GridRowEditingApi, 'getRowMode'> {
   /**
-   * Puts the row at the given id into edit mode.
+   * Puts the row correspoding to the given id into edit mode.
```
flaviendelangle, line 231:
```suggestion
   * Puts the row corresponding to the given id into edit mode.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -224,20 +228,24 @@ export interface GridNewRowEditingApi
...
    * @param {GridStartCellEditModeParams} params The row id edit.
    * @ignore - do not document.
    */
   startRowEditMode(params: GridStartRowEditModeParams): void;
   /**
-   * Puts the row at the given id and into view mode and updates the original row with the new values stored.
+   * Puts the row correspoding to the given id and into view mode and updates the original row with the new values stored.
```
flaviendelangle, line 237:
```suggestion
   * Puts the row corresponding to the given id and into view mode and updates the original row with the new values stored.
```

alexfauquette (approved):
Nothing to add

packages/storybook/src/stories/grid-rows.stories.tsx
```diff
@@ -513,7 +515,7 @@ export function EditRowsControl() {
         apiRef.current.updateRows([cellUpdate]);
         apiRef.current.publishEvent(
           GridEvents.cellEditStop,
-          apiRef.current.getCellParams(id, field),
+          apiRef.current.getCellParams(id, field) as GridCellEditStopParams,
```
flaviendelangle, line 518: Is `reasons` really defined here ? 
Otherwise, can't it cause problem to let typescript think we are passing it to `cellEditStop` when we aren't ?

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -670,4 +674,11 @@ export interface DataGridPropsWithoutDefaultValue extends CommonProps {
    * For each feature, if the flag is not explicitly set to `true`, the feature will be fully disabled and any property / method call will not have any effect.
    */
   experimentalFeatures?: Partial<GridExperimentalFeatures>;
+  /**
+   * Callback called before updating a row with new values in the row and cell editing.
+   * @param {GridRowModel} newRow Row object with the new values.
+   * @param {GridRowModel} oldRow Row object with the old values.
+   * @returns {Promise<GridRowModel>} A promise which resolves with the final values to update the row.
+   */
+  processRowUpdate?: (newRow: GridRowModel, oldRow: GridRowModel) => Promise<GridRowModel>;
```
flaviendelangle, line 683:
```suggestion
  processRowUpdate?: (newRow: GridRowModel, oldRow: GridRowModel) => GridRowModel | Promise<GridRowModel>;
```

If you remove the `async` on https://codesandbox.io/s/test-new-editing-h9xbfx?file=/demo.tsx, you have a TS error.
But it does work no ?

packages/storybook/src/stories/grid-rows.stories.tsx
m4theushw: I made `reason` optional. It's only available if the new API is used.

packages/grid/x-data-grid/src/models/params/gridCellParams.ts
```diff
@@ -112,6 +114,11 @@ export interface GridValueSetterParams {
    * The row that is being edited.
    */
   row: GridRowModel;
+  /**
+   * The original row with the values before entering the edit mode.
```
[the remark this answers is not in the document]
m4theushw: I removed this attribute because I don't remember why I added it.

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -126,7 +126,136 @@ export interface GridCellEditingApi extends GridEditingSharedApi {
...
+  /**
+   * Whether or not to ignore the modifications made on this cell.
+   * @default false
+   */
+  ignoreModifications?: boolean;
+  /**
+   * To which cell to move focus after finishing editing.
+   * @default "none"
+   */
+  cellToMoveFocus?: 'none' | 'below' | 'right' | 'left';
```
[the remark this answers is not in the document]
m4theushw: I went with `cellToFocusAfter`.

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
m4theushw: It works because the call is wrapped by `Promise.resolve`. Should we change the definition to include the non-promise return? If we do we have to also explain in the docs. Currently, I say "when it resolves" because it only accepts a promise, but if we change to also allow to return the object I'll have to say in places "when it resolves, or returns" instead. `processRowUpdate` is meant to save the row in a database. I suppose only a small portion of users will use it to save to some local state.

packages/grid/x-data-grid/src/models/params/gridCellParams.ts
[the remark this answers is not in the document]
flaviendelangle: Could it be related to https://github.com/mui/mui-x/issues/3763 ?

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
flaviendelangle: I think it is easy enough to support the sync update.

packages/grid/x-data-grid/src/models/params/gridCellParams.ts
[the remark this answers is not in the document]
m4theushw: No, I'm passing `originalRow` to the edit cell. FOr #3763 there's `hasChanged`.

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
m4theushw: So should I keep as it is or allow `GridRowModel` without a promise?

flaviendelangle: Allow `GridRowModel` without a promise

m4theushw: Done

flaviendelangle (approved):

m4theushw:
I merged to fix the conflicts in #4060. You can still suggest changes to the API in #4060

packages/grid/x-data-grid/src/internals/components/cell/GridEditSingleSelectCell.tsx
jamrsim: If `preProcessEditCellProps` is not being used and/or the desired UX is that once a user has made a `MenuItem` selection the `singleSelect` will automatically commit, is there a prescribed method for doing so?  Or, is there a workaround to make that happen? Thank you.

jamrsim: Based on the auto-stop example, my understanding is this cannot be done with the `singleSelect` column type but rather must be accomplished with a custom column type. While the custom column type may be a `Select`, it would not be a singleSelect. Thank you.

PR #3965, opened 2022-02-16
[DataGrid] Use the same bundling scripts as the core repository

flaviendelangle:
Part of #18
Fixes #50
Fixes #51
Fixes #924 
Fixes #3967
Fixes #4024

Follow up on #3953

[Codesandbox for DataGridPro](https://codesandbox.io/s/datagridpro-v5-quick-start-forked-l4w31y?file=/package.json)

[Doc preview](https://deploy-preview-3965--material-ui-x.netlify.app/api/data-grid/grid-api/), you can see the "pro" icon on the methods only accessible on the pro version of the interface.

_____

- [x] Tested on Create React App / Webpack (dev and prod builds) by @flaviendelangle 
- [x] Tested on Next JS (dev and prod builds) by @cherniavskii 
- [x] Tested on Vite JS by @cherniavskii 
- [x] Tested on Codesandbox by @flaviendelangle 

- [x] Documentation accessible 
- [x] Storybook accessible

PR #3981, opened 2022-02-18
[DataGridPremium] Add Excel export

### LINKED ISSUE #198: [data grid] Implement Excel export
### Benchmark

- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/export/
- https://www.telerik.com/kendo-react-ui/components/grid/excel-export/
- https://demos.telerik.com/kendo-ui/grid/excel-export
- https://ag-grid.com/javascript-grid-excel/
- http://tabulator.info/docs/4.7/download#xlsx
- https://ej2.syncfusion.com/react/demos/#/material/grid/default-exporting
- https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--common-features

alexfauquette:
Close #198 

Doc preview: 
https://deploy-preview-3981--material-ui-x.netlify.app/x/react-data-grid/export/#excel-export

Todo:

- [x] Improve the documentation
- [x] Add tests
	- [x] Verify when UX is displayed
	- [x] Test the exported file
- [x] Correct `any` types added
- [x] Update the technical doc about  `'columnMenu'`
- [x] fix exclejs bug with dates
- [x] document how to set colun type

## Implemented features

The excel export is only visible from the `x-data-grid-pro` package

`excelOptions` are similar to `csvOptions`

For now, exported features are the column type (number, boolean, date, and single-select) and tree structure
The sheet can be edited before and after the rows containing data

## Changelog

- 👔 Add Excel export

It is possible to export your Data Grid content in Excel format. 
This feature is planned to be a premium one. To test it, enable its experimental flag as follow:
```jsx
<DataGridPro experimentalFeatures={{ excelExport: true }} />
```

That's all, now your users can export to excel format 🎉 
For more detail on how to customize the export, please visit our documentation.


package.json
```diff
@@ -189,6 +189,7 @@
     ]
   },
   "dependencies": {
+    "exceljs": "^4.3.0",
```
flaviendelangle, line 192: Should this deps be on `@mui/x-data-grid-pro` as an optional peer deps ?

packages/grid/x-data-grid-pro/src/index.ts
```diff
@@ -38,3 +38,8 @@ export type {
   GridColumns,
   GridComparatorFn,
 } from './typeOverload';
+export type {
```
flaviendelangle, line 41: To improve after #3965 to remove the explicit export here

packages/grid/x-data-grid-pro/src/internals/components/GridExcelExportMenuItem.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+// import PropTypes from 'prop-types';
+import MenuItem from '@mui/material/MenuItem';
+import { GridExportMenuItemProps } from '@mui/x-data-grid';
+import { useGridApiContext } from '../hooks/utils/useGridApiContext';
+import { GridExcelExportOptions } from '../hooks/features/export';
+// import { DataGridProProps } from '../../models/dataGridProProps';
```
flaviendelangle, line 7: Those two commented lines should probably be removed

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/gridExcelExportInterface.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface GridExcelExportApi {
+  /**
+   * Returns the grid data as an exceljs workbook.
+   * This method is used internally by `exportDataAsExcel`.
+   * @param {GridExcelExportOptions} options The options to apply on the export.
+   * @returns {any} The data in a exceljs workbook object.
+   */
+  getDataAsExcel: (options?: GridExcelExportOptions) => Promise<Workbook> | null;
+  /**
+   * Downloads and exports a excel of the grid's data.
```
flaviendelangle, line 41:
```suggestion
   * Downloads and exports an Excel file of the grid's data.
```

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+  });
+  return {
+    row,
+    dataValidation,
+    outlineLevel,
+  };
+};
+
+const serialiseColumn = (column: GridColDef) => {
```
flaviendelangle, line 117:
```suggestion
const serializeColumn = (column: GridStateColDef) => {
```

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  if (exceljsPreprocess) {
+    await exceljsPreprocess({
+      workbook,
+      worksheet,
+    });
+  }
+
+  if (includeHeaders) {
+    worksheet.addRow(columns.map((column) => column.headerName || column.field));
```
flaviendelangle, line 156: Not for this PR
But it make me more confident that we should do https://github.com/mui/mui-x/pull/3959#issuecomment-1042699805

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/useGridExcelExport.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const logger = useGridLogger(apiRef, 'useGridExcelExport');
+  const inactivatedError = buildWarning(
+    [
+      'MUI: export .excel is experimental and must be activated by setting',
+      '<DataGridPro experimentalFeatures={{excelExport: true}} />',
+    ],
+    'error',
+  );
+
+  const getDataAsExcel = React.useCallback(
```
flaviendelangle, line 78:
```suggestion
  const getDataAsExcel = React.useCallback<GridExcelExportApi['getDataAsExcel']>(
```

Same for `exportDataAsExcel`
In the past we had slight interface differences between the `GridApi` signature and the column definition due to those type explicit writting.

flaviendelangle (commented):
Looks great
I did not test the behavior yet, just low level code suggestions.

One limit of the bundling is that even projects not using this feature will spend time bundling exceljs (even if on a separate chunk). I don't know if it is 100% negligible. 
And if not, how to improve it.

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+  exceljsPostprocess?: (processInput: GridExceljsProcessInput) => Promise<void>;
+}
+
+export async function buildExcel(options: BuildExcelOptions, api): Promise<Excel.Workbook> {
+  const { columns, rowIds, getCellParams, includeHeaders, exceljsPreprocess, exceljsPostprocess } =
+    options;
+
+  const excelJS = await getExcelJs();
+  const workbook: Excel.Workbook = new excelJS.Workbook();
+  const worksheet = workbook.addWorksheet('Sheet1');
```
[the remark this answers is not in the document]
alexfauquette: @DanailH had the same idea. The problem is that there is so many things to customize, passing an option for each of them will probably be a pain.

Devs can use `exceljsPreprocess` to get access to the exceljs workSheet and then edits all its properties, such as `worksheet.name = 'My pretty sheet'` (https://github.com/exceljs/exceljs/issues/381)

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+        };
+        const formattedValueOptions = getFormattedValueOptions(
+          cellParams.colDef,
+          valueOptions,
+          api,
+        );
+        dataValidation[column.field] = {
+          type: 'list',
+          allowBlank: true,
+          formulae: [formattedValueOptions.map((x) => `"${x}"`)],
```
m4theushw, line 80: When I try to open in Excel the file generated in https://deploy-preview-3981--material-ui-x.netlify.app/components/data-grid/export/#excel-export it claims that the validation formula is wrong and it asks if I want to to try to recover the file. In the ExcelJS GitHub page it says that the values should be separated by comma. In my Excel (Portuguese) I need to use semicolon for it to work:

image

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+  includeHeaders: boolean;
+  exceljsPreprocess?: (processInput: GridExceljsProcessInput) => Promise<void>;
+  exceljsPostprocess?: (processInput: GridExceljsProcessInput) => Promise<void>;
+}
+
+export async function buildExcel(options: BuildExcelOptions, api): Promise<Excel.Workbook> {
+  const { columns, rowIds, getCellParams, includeHeaders, exceljsPreprocess, exceljsPostprocess } =
+    options;
+
+  const excelJS = await getExcelJs();
```
m4theushw, line 142: We could show a message asking users to install exceljs if not installed.

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/useGridExcelExport.tsx
```diff
@@ -0,0 +1,148 @@
...
+ * @requires useGridFilter (state)
+ * @requires useGridSorting (state)
+ * @requires useGridSelection (state)
+ * @requires useGridParamsApi (method)
+ */
+export const useGridExcelExport = (
+  apiRef: React.MutableRefObject<GridApiPro>,
+  props: Partial<Pick<DataGridProProps, 'experimentalFeatures'>>,
+): void => {
+  const featureIsActivated = props?.experimentalFeatures?.excelExport === true;
```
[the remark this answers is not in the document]
alexfauquette: To be precise it is `boolean | undefined` but yes in that case `undefined` or `false` are the same

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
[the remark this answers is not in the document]
flaviendelangle: I guess starting with 0 customization options on our side and just a callback to let user process manually the Worksheet make sense.
And if some use case can be simplified on our side and are often used by users, wel can add options later.

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+  columns: GridStateColDef[];
+  rowIds: GridRowId[];
+  getCellParams: (id: GridRowId, field: string) => GridCellParams;
+  includeHeaders: boolean;
+  exceljsPreprocess?: (processInput: GridExceljsProcessInput) => Promise<void>;
+  exceljsPostprocess?: (processInput: GridExceljsProcessInput) => Promise<void>;
+}
+
+export async function buildExcel(options: BuildExcelOptions, api): Promise<Excel.Workbook> {
+  const { columns, rowIds, getCellParams, includeHeaders, exceljsPreprocess, exceljsPostprocess } =
```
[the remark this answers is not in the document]
alexfauquette: It was to have the same signature than csvExport, but I had to add the api in order to format the value options

packages/grid/x-data-grid/src/internals/components/toolbar/GridToolbarExport.tsx
```diff
@@ -70,14 +71,27 @@ const GridToolbarExport = React.forwardRef<HTMLButtonElement, GridToolbarExportP
...
     const { csvOptions = {}, printOptions = {}, ...other } = props;
 
+    const apiRef = useGridApiContext();
+
+    const defaultButtons = [
+      <GridCsvExportMenuItem options={csvOptions} />,
+      <GridPrintExportMenuItem options={printOptions} />,
+    ];
+
     if (csvOptions?.disableToolbarButton && printOptions?.disableToolbarButton) {
```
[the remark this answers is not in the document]
alexfauquette: This should be fixed now

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+  });
+  return {
+    row,
+    dataValidation,
+    outlineLevel,
+  };
+};
+
+const serialiseColumn = (column: GridStateColDef) => {
```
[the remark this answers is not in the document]
alexfauquette: also applied the correction to csvSeriali(s|z)er

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
alexfauquette: Do you mean ad a try/catch to customize the error message? Because it already display a message explaining that exceljs is not imported.

I added a sentence in the doc to explain dev should add the package

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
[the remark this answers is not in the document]
m4theushw: Yes, good. I think this was last usage of British English in the codebase.

packages/grid/x-data-grid-pro/src/internals/hooks/features/export/serializer/excelSerializer.ts
alexfauquette: I miss understood the doc.
Instead of using
```ts
formulae: ['"option1,oprion2,option3"']
```
I used 
```ts
formulae: ['"option1","oprion2","option3"']
```

That's corrected. It works better on Google Sheet, on Microsoft, I don't know

docs/data/data-grid/export/ExcelCustomExcelExport.tsx.preview
```diff
@@ -0,0 +1,11 @@
+<DataGridPro
```
DanailH, line 1: Can we rename the file to `CustomExelExport.tsx`?
X the other files.

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (exceljsPreprocess) {
+    await exceljsPreprocess({
+      workbook,
+      worksheet,
+    });
+  }
+
+  if (includeHeaders) {
+    worksheet.addRow(columns.map((column) => column.headerName || column.field));
+  }
```
DanailH, line 152:
```suggestion
  }
  
```

packages/grid/x-data-grid/src/constants/localeTextConstants.ts
```diff
@@ -30,6 +30,7 @@ export const GRID_DEFAULT_LOCALE_TEXT: GridLocaleText = {
   toolbarExportLabel: 'Export',
   toolbarExportCSV: 'Download as CSV',
   toolbarExportPrint: 'Print',
+  toolbarExportExcel: 'Export Excel',
```
DanailH, line 33:
```suggestion
  toolbarExportExcel: 'Download as Excel',
```

docs/data/data-grid/export/ExcelCustomExcelExport.tsx.preview
flaviendelangle: I propose `ExcelCustomExport`

I likfe prefixing by the "demo category", for the row grouping / tree data for instance, it helps visually splitting them into two groups.

Or we could create a sub-folder.

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,177 @@
...
+          formulae: [`"${formattedValueOptions.map((x) => x.toString()).join(',')}"`],
+        };
+        const formattedValue = api.getCellParams(id, column.field).formattedValue;
+        if (process.env.NODE_ENV !== 'production') {
+          if (String(cellParams.formattedValue) === '[object Object]') {
+            warnInvalidFormattedValue();
+          }
+        }
+        row[column.field] =
+          typeof formattedValue === 'object' ? formattedValue.label : formattedValue;
```
m4theushw, line 83: Same thing?

```suggestion
        row[column.field] = formattedValue.label ?? formattedValue;
```

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,177 @@
...
+        break;
+      }
+      case 'boolean':
+      case 'number':
+      case 'date':
+      case 'dateTime':
+        row[column.field] = api.getCellParams(id, column.field).value;
```
m4theushw, line 90: The `boolean` column, for example, has "yes" and "no" as labels but it exports using the [`TRUE`](https://support.microsoft.com/en-us/office/true-function-7652c6e3-8987-48d0-97cd-ef223246b3fb) function from Excel. I don't think users are used to use this function to denote boolean values.

```suggestion
        row[column.field] = api.getCellParams(id, column.field).formattedValue;
```

packages/grid/x-data-grid/src/utils/exportAs.ts
```diff
@@ -10,9 +10,10 @@ import { GridExportFormat } from '../models/gridExport';
  * https://unpkg.com/browse/@progress/kendo-file-saver@1.0.7/dist/es/save-as.js
  * https://github.com/ag-grid/ag-grid/blob/9565c219b6210aa85fa833c929d0728f9d163a91/community-modules/csv-export/src/csvExport/downloader.ts
  */
-export function exportAs(
+
+export function exportAs<ExtraExtensions extends string>(
   blob: Blob,
-  extension: GridExportFormat = 'csv',
+  extension: GridExportExtension | ExtraExtensions = 'csv',
   filename: string = document.title,
```
m4theushw, line 17: A small detail but the page may not have title, for instance, the playground doesn't have. In this case it download as "extension.extension".

```suggestion
  filename: string = document.title || 'untitled',
```

packages/grid/x-data-grid/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
+    "jsx": "react",
```
m4theushw, line 4: Why this change?

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,38 @@ The demo below shows how to add a JSON export.
...
+> ```
+>
+> The feature is stable in its current form, and we encourage users willing to migrate to the Premium plan once available to start using it.
+
+The excel export allows translating columns' type and tree structure of a DataGrid to an Excel file.
+
+Columns with type `'boolean'`, `'number'`, `'singleSelect'`, `'date'`, and `'dateTime'` will be exported in their corresponding type in Excel. To make this work, please ensure that `rows` values have the correct type, or [convert types to correct ones](/components/data-grid/columns/#converting-types).
+
+This feature relies on the package [exceljs](https://github.com/exceljs/exceljs).
+To export in Excel format, you have to install this package with `yarn add exceljs` or `npm install exceljs`.
```
m4theushw, line 205: Please use the same format from

https://github.com/mui/mui-x/blob/d7cb410367eddf19e421c8d2d81a24f05e00eb64/docs/data/data-grid/getting-started/getting-started.md?plain=1#L13-L19

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,38 @@ The demo below shows how to add a JSON export.
...
+
+{{"demo": "ExcelExport.js", "bg": "inline"}}
+
+### Template Excel
+
+To manage Excel file creation, the DataGrid uses [exceljs](https://github.com/exceljs/exceljs) internally.
```
m4theushw, line 211: We already introduced this requirement in the introduction.

```suggestion
```

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,38 @@ The demo below shows how to add a JSON export.
...
+
+You can customize the document before and after the creation of rows containing data, by using functions `exceljsPreprocess` and `exceljsPostprocess`.
```
m4theushw, line 213: Could you describe better when `exceljsPreprocess` is ideal instead of `exceljsPostprocess`.

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,38 @@ The demo below shows how to add a JSON export.
...
+These functions input is an object `{workbook, worksheet}`, which are `exceljs` objects allowing you to manipulate the excel file.
```
m4theushw, line 214:
```suggestion
Both functions receive as input an object `{ workbook, worksheet }`, which are `exceljs` objects allowing you to manipulate the Excel file.
```

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,38 @@ The demo below shows how to add a JSON export.
...
+
+Thanks to these two methods, you can add metadata to your documents, or add custom content on top or bottom of the worksheet.
+In the demo below, we add some rows at the top of the Excel file to add a custom header.
```
m4theushw, line 217: Could you add an example inline of how to use any of the methods? It can show how to change the worksheet title or something else.

packages/grid/x-data-grid-pro/src/components/GridExcelExportMenuItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useGridApiContext } from '../hooks/utils/useGridApiContext';
+import { GridExcelExportOptions } from '../hooks/features/export';
+
+export type GridExcelExportMenuItemProps = GridExportMenuItemProps<GridExcelExportOptions>;
+
+const GridExcelExportMenuItem = (props: GridExcelExportMenuItemProps) => {
+  const apiRef = useGridApiContext();
+  const { hideMenu, options } = props;
+
+  if (options?.disableToolbarButton) {
```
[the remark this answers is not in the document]
alexfauquette: The structure is the following.
```jsx
<GridToolbarExport>
	<GridToolbarExportContainer>
		<GridCsvExportMenuItem />
	<GridToolbarExportContainer>
<GridToolbarExport>
```

 We do the check when rendering `<GridToolbarExport />` but if dev create their own menu with `<GridToolbarExportContainer />` to add another export format. The check will not be done 

So for consistency, the buttons should bee removed if they have the prop `options={{ disableToolbarButton: true }}`

packages/grid/x-data-grid/tsconfig.json
alexfauquette: A mistake from when I was struggling with the migration

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
alexfauquette: Not if `formattedValue` is `undefined`. Another solution is:
```
row[column.field] = formattedValue && (formattedValue.label ?? formattedValue)
```

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
alexfauquette: That's the way ExcelJS handle boolean
https://github.com/exceljs/exceljs#boolean-value

I tried `row[column.field] = "TRUE"` and `row[column.field] = "=True"` but it is always considered as a string and not the boolean value

m4theushw: We can wait for feedback from the community. The way you did is also correct. I only think that users will always convert from `TRUE` to another value that makes more sense. If I open the Excel export in Google Spreadshets or Excel it displays as follow:

image

If I create this same spreadsheet by hand I would've used "Y"/"N" or "Yes"/"No" instead. Anyway, I think someone will ask for an option to convert boolean values to a custom string value.

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
m4theushw: I'm not following. If `formattedValue=undefined` then `row[column.field]` will be `undefined` too, like the previous code.

packages/grid/x-data-grid-pro/src/components/GridExcelExportMenuItem.tsx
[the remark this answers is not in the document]
m4theushw: I meant to remove the following check: https://github.com/mui/mui-x/pull/3981/files#diff-bfbe77acfbaf2424d3fd7f75b5bd63ebfcd9cfd6f068ec7d3af4295ec3cac906R105-R107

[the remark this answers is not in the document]
alexfauquette: This check is here to such that if all buttons are disabled, the export button is not visible
```jsx
if (preProcessedButtons.length === 0) {
	return
}
```

packages/grid/x-data-grid-pro/src/hooks/features/export/serializer/excelSerializer.ts
m4theushw: You can use `?` to tell that `formattedValue` may be `undefined`.

```tsx
row[column.field] = formattedValue?.label ?? formattedValue;
```

packages/grid/x-data-grid-pro/src/components/GridExcelExportMenuItem.tsx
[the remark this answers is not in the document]
m4theushw: There're two places checking the exact same thing. In `GridExcelExportMenuItem` (here) and in the image below:

image

One of them can be removed, I propose to remove from `addExportMenuButtons` to keep similar to how other buttons work. I don't see a problem in returning a component that will render `null` at the end.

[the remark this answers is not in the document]
alexfauquette: You don't see a problem displaying a button Export which will open an empty menu?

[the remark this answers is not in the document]
alexfauquette: Other default buttons have the same check here
https://github.com/alexfauquette/material-ui-x/blob/excel-export-2/packages/grid/x-data-grid/src/components/toolbar/GridToolbarExport.tsx#L76:L82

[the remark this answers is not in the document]
m4theushw: In this case, you can remove the check from `GridExcelExportMenuItem` and leave the one in the pre-processor.

```suggestion
```

In the other buttons we're also checking twice now. Once in each button and also when adding to the menu items array.

https://github.com/mui/mui-x/blob/0be075d82657599b9ed9c9bf2c13af3020bff039/packages/grid/x-data-grid/src/components/toolbar/GridToolbarExport.tsx#L35-L37

https://github.com/mui/mui-x/blob/0be075d82657599b9ed9c9bf2c13af3020bff039/packages/grid/x-data-grid/src/components/toolbar/GridToolbarExport.tsx#L54-L56

packages/grid/x-data-grid/src/components/toolbar/GridToolbarExport.tsx
```diff
@@ -68,16 +69,32 @@ export const GridPrintExportMenuItem = (props: GridPrintExportMenuItemProps) =>
...
+      ...(csvOptions.disableToolbarButton ? [] : [<GridCsvExportMenuItem options={csvOptions} />]),
+      ...(printOptions.disableToolbarButton
+        ? []
+        : [<GridPrintExportMenuItem options={printOptions} />]),
+    ];
+
+    const preProcessedButtons = apiRef.current.unstable_applyPreProcessors(
+      'exportMenu',
+      defaultButtons,
+      excelOptions,
```
m4theushw, line 86: If we pass `excelOptions` directly, we can't support other buttons in the future.

```suggestion
      { excelOptions },
```

packages/grid/x-data-grid-pro/src/hooks/features/export/useGridExcelExport.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useGridRegisterPreProcessor, buildWarning, exportAs } from '@mui/x-data-grid/internals';
+import { DataGridProProps } from '../../../models/dataGridProProps';
+import { GridApiPro } from '../../../models/gridApiPro';
+import {
+  GridExcelExportApi,
+  GridExportExtension,
+  GridExcelExportOptions,
+} from './gridExcelExportInterface';
+import { buildExcel } from './serializer/excelSerializer';
+import { defaultGetRowsToExport, getColumns } from './serializer/utils';
```
DanailH, line 12: Since you created helpers does it make sence to also use them for the CSV export?

packages/grid/x-data-grid/src/models/gridExport.ts
```diff
@@ -98,3 +124,8 @@ export interface GridPrintExportOptions extends GridExportOptions {
  * Available export formats.
  */
 export type GridExportFormat = 'csv' | 'print';
+
+/**
+ * Available export extensions.
+ */
+export type GridExportExtension = 'csv';
```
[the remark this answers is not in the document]
alexfauquette: Only in pro package
https://github.com/alexfauquette/material-ui-x/blob/excel-export-2/packages/grid/x-data-grid-pro/src/hooks/features/export/gridExcelExportInterface.ts#L10:L11

packages/grid/x-data-grid-pro/src/hooks/features/export/useGridExcelExport.tsx
alexfauquette: Yes, I forget the PR was merged 👍

package.json
```diff
@@ -177,5 +177,8 @@
   },
   "dependencies": {
     "html-webpack-plugin": "^5.5.0"
+  },
+  "optionalDependencies": {
```
flaviendelangle, line 181: On the core they use `peerDependenciesMeta`

```json
  "peerDependencies": {
    "@emotion/react": "^11.5.0",
    "@emotion/styled": "^11.3.0",
    "@types/react": "^16.8.6 || ^17.0.0",
    "react": "^17.0.0",
    "react-dom": "^17.0.0"
  },
  "peerDependenciesMeta": {
    "@types/react": {
      "optional": true
    },
    "@emotion/react": {
      "optional": true
    },
    "@emotion/styled": {
      "optional": true
    }
  },
```

docs/package.json
```diff
@@ -49,6 +49,7 @@
     "cross-env": "^7.0.3",
     "docsearch.js": "^2.6.3",
     "doctrine": "^3.0.0",
+    "exceljs": "^4.3.0",
```
joserodolfofreitas, line 52: It's working on our documentation but I couldn't use the feature on codesandbox because it's missing this dependency.

joserodolfofreitas (commented):
I added a few suggestions and considerations to the doc page.

And I checked the exported files on `Google sheets`, `Excel (Microsoft 365)`, and `Apple Numbers`.

`Numbers` doesn't seem to support rowGrouping, and on the second example, `Google sheets` didn't display field values (selected values based on types - neither singleSelect nor boolean) - This case with Google sheets is weird because on the first example it works well with `contract types`. 

But `Excel` worked relatively well on both examples, so I believe we're good to start. (I only spotted a minor usability problem on groups inside groups within the first example, but that's probably not a problem we can solve).

On the future, we should strive for giving the same look (or as close as viably possible) to the exported file. I understand this might be a huge amount of work, but we can start with the quick wins that can help us display a better "integration" of our features with the excel format. 
Like on the following image, where we are able to better display the "path" within the input dataset.

image

docs/package.json
alexfauquette: I don't know how to add packages to the codesandbox examples. But I'm not sure it is a good idea to add it for every user if it's needed for only one feature.
Adding a package on codesandbox is fast. Just write the name of the package in the dependencies field

image

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,69 @@ The demo below shows how to add a JSON export.
...
+
+function exceljsPostProcess({ worksheet }) {
+  // add a text after the data
+  worksheet.addRow(); // Add empty row
+
+  const newRow = worksheet.addRow();
+  newRow.getCell(1).value = 'Those data are for internal use only';
+}
+```
+
```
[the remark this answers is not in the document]
alexfauquette: The goal was to say 

> Look at what you can do, but for that go checking the exceljs doc

Like that there is no possible confusion between what comes from to the data grid, and what comes from exceljs

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,69 @@ The demo below shows how to add a JSON export.
...
+ // with npm
+ npm install exceljs
+
+ // with yarn
+ yarn add exceljs
+```
+
+### Template Excel
+
+You can customize the document before and after the creation of rows containing data, by using functions `exceljsPreProcess` and `exceljsPostProcess`.
```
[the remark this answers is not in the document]
alexfauquette: I do not think we can change the lib without making breaking change.

If we change the library, it will have different methods, and so developers will have to rewrite there `excelPreProcessor` and  `excelPostProcessor`

To add a new library, we can add an option `library` set by default to `"exceljs"` and if it developpers set it to `"xlsx"` we use that library and `xlsxPreProcessor` instead of `exceljsPreProcessor`

docs/package.json
flaviendelangle: > I don't know how to add packages to the codesandbox examples

You are blocked by the freeze ?

alexfauquette: Easier than that, I don't know what is "the freeze", and I do not know how codesandbox are generated by the doc 😅

flaviendelangle: Oh I thought you were talking about regular codesandboxes, not the one of the doc.

https://github.com/mui/mui-x/pull/3396/files

You have an example here about how to modify the deps injected in the codesandbox.
We just have to add `exceljs` for `@mui/x-data-grid-pro` I think.

alexfauquette: We can not add dependencies for specific examples. It works thanks to a regExp finding all the `import ... from ...` to know which are the dependencies.

As suggested, I added it as a pro dependency, does not seems to decrease perf of opening codesandbox 👍

docs/data/data-grid/export/export.md
```diff
@@ -185,12 +185,69 @@ The demo below shows how to add a JSON export.
 
 {{"demo": "CustomExport.js", "bg": "inline", "defaultCodeOpen": false}}
 
-## 🚧 Excel export [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
+## Excel export [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
 
-> ⚠️ This feature isn't implemented yet. It's coming.
+> ⚠️ This feature is temporarily available on the Pro plan until the release of the Premium plan.
```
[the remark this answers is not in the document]
alexfauquette: I keep this comment open, I will do the modification when we effectively have the Premium package

packages/grid/x-data-grid-pro/src/index.ts
alexfauquette: Do you still see space for improvement?

docs/data/data-grid/export/ExcelCustomExport.js
```diff
@@ -0,0 +1,478 @@
...
+});
+
+const isNavigationKey = (key) =>
+  key === 'Home' ||
+  key === 'End' ||
+  key.indexOf('Arrow') === 0 ||
+  key.indexOf('Page') === 0 ||
+  key === ' ';
+
+const CustomGridTreeDataGroupingCell = (props) => {
```
flaviendelangle, line 28: Does the custom grouping cell bring any value here ?
If not could we remove it to keep the demo focused on the Excel export ?

docs/data/data-grid/export/ExcelExport.js
```diff
@@ -0,0 +1,158 @@
...
+
+function CustomToolbar() {
+  return (
+    <GridToolbarContainer>
+      <GridToolbarExport />
+    </GridToolbarContainer>
+  );
+}
+
+export default function ExcelExport() {
```
flaviendelangle, line 143: Could we do a very basic Excel export demo without any tree related feature
And then an example with the tree data or row grouping (don't know which one makes the most sense).

That way someone that only wants to know how the Excel export work don't have any overhead related to the tree feature.

docs/data/data-grid/export/export.md
```diff
@@ -36,7 +36,7 @@ By default, the export menu displays all the available export formats, according
 
 - [Print](#print-export)
 - [CSV](#csv-export)
-- [Excel](#excel-export) [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/) (🚧 Not delivered yet)
+- [Excel](#excel-export) [<span class="plan-premium"></span>](https://mui.com/store/items/material-ui-pro/)
```
flaviendelangle, line 39: We will need to update [this line](https://github.com/mui/material-ui/blob/424a31de18680baab7442c8cca8ff8c2530f8eb3/docs/src/components/pricing/PricingTable.tsx#L582)

docs/pages/_app.js
```diff
@@ -50,6 +50,7 @@ ponyfillGlobal.muiDocConfig = {
     if (newDeps['@mui/x-data-grid-pro']) {
       newDeps['@mui/material'] = versions['@mui/material'];
       newDeps['@mui/x-data-grid'] = versions['@mui/x-data-grid'];
+      newDeps.exceljs = 'latest';
```
flaviendelangle, line 53: `versions['exceljs`]` did not work ? Like we do for `@mui/material`

packages/grid/x-data-grid-pro/src/DataGridPro/useDataGridProComponent.tsx
```diff
@@ -137,6 +138,7 @@ export const useDataGridProComponent = (
   useGridDimensions(apiRef, props);
   useGridEvents(apiRef, props);
   useGridStatePersistence(apiRef);
+  useGridExcelExport(apiRef, props);
```
flaviendelangle, line 141: Move next to CSV export

packages/grid/x-data-grid/src/components/toolbar/GridToolbar.tsx
```diff
@@ -34,18 +35,13 @@ const GridToolbar = React.forwardRef<HTMLDivElement, GridToolbarProps>(function
       <GridToolbarColumnsButton />
       <GridToolbarFilterButton />
       <GridToolbarDensitySelector />
-      <GridToolbarExport csvOptions={csvOptions} printOptions={printOptions} />
+      <GridToolbarExport
+        csvOptions={csvOptions}
+        printOptions={printOptions}
+        excelOptions={excelOptions}
```
flaviendelangle, line 41: It is not great to have a reference to Excel here in a community component
But to avoid it we would need a do a premium specific `GridToolbar`

Could you add a comment `Stop mentioning Excel here` so that we remember that it is not great ?

packages/grid/x-data-grid/src/components/toolbar/GridToolbarExport.tsx
```diff
@@ -68,16 +63,32 @@ export const GridPrintExportMenuItem = (props: GridPrintExportMenuItemProps) =>
...
 const GridToolbarExport = React.forwardRef<HTMLButtonElement, GridToolbarExportProps>(
   function GridToolbarExport(props, ref) {
-    const { csvOptions = {}, printOptions = {}, ...other } = props;
+    const { csvOptions = {}, printOptions = {}, excelOptions, ...other } = props;
 
-    if (csvOptions?.disableToolbarButton && printOptions?.disableToolbarButton) {
+    const apiRef = useGridApiContext();
+
+    const defaultButtons = [
+      ...(csvOptions.disableToolbarButton ? [] : [<GridCsvExportMenuItem options={csvOptions} />]),
```
flaviendelangle, line 71: Do we intend to move the csv and print options as pre-processor just like for the Excel in the future ?

docs/data/data-grid/export/ExcelCustomExport.js
alexfauquette: Yes, it does not bring that much value

docs/pages/_app.js
alexfauquette: `versions.exceljs` is not defined at this stage. 

To do so, I added it to `csbGetVersions` the `exceljs: 'latest'` property

m4theushw (commented):
I tried to export to Excel the demo in https://deploy-preview-3981--material-ui-x.netlify.app/components/data-grid/demo/ and Excel warned that some column had invalid data.

https://codesandbox.io/s/fullfeatureddemo-material-demo-forked-xweore

alexfauquette:
@m4theushw When you say "some column had invalid data" is it for countries which contain a `,`?

docs/pages/_app.js
```diff
@@ -66,6 +67,7 @@ ponyfillGlobal.muiDocConfig = {
       '@mui/x-data-grid-pro': getMuiPackageVersion('x-data-grid-pro', muiCommitRef),
       '@mui/x-data-grid-generator': getMuiPackageVersion('x-data-grid-generator', muiCommitRef),
       '@mui/x-data-grid': getMuiPackageVersion('x-data-grid', muiCommitRef),
+      exceljs: 'latest',
```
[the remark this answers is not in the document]
m4theushw: Yes. I think it's better than to force all demos to download this dependency. Once `@mui/x-data-grid-premium` is available we can change.

m4theushw:
> @m4theushw When you say "some column had invalid data" is it for countries which contain a ,?

I can't say this is the problem. I exported the same demo with only 1 row and the warning also appeared.

alexfauquette:
Since the coma is used to split cell value options, the value option `Congo, Democratic Republic of the` is interpreted as `["Congo", "Democratic Republic of the"]`  which does not contain the cell value `"Congo, Democratic Republic of the"`. 

Then G-Sheet returns an error: "Invalid input, The cell value must be part of the specified list"

I modified the behavior of the `valueOptions` as follows to handle coma, which appears to also solve bugs in LibreOffice:

Options are now placed in another sheet. Each `singleSelect` column has its own column containing the value options, and the data validation refer to this sheet

The limitation is when developers use a function to return different `valueOptions` depending on the row. For this specific case, I kept the previous solution for data validation. So it will not be able to contain coma in the options. It is not completely impossible to put all the options in another sheet, but the excel formulae will probably be impossible to read for a human

docs/scripts/api/buildInterfacesDocumentation.ts
```diff
@@ -64,6 +65,7 @@ const OTHER_GRID_INTERFACES_WITH_DEDICATED_PAGES = [
   'GridColDef',
   'GridCsvExportOptions',
   'GridPrintExportOptions',
+  'GridExcelExportOptions',
```
m4theushw, line 68: You also need to add to `docs/data/pages.ts` to appear in the sidebar.

packages/grid/x-data-grid-pro/src/hooks/features/export/gridExcelExportInterface.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Name given to the worksheet containing the columns valueOptions.
+   * valueOptions are added to this worksheet if they are provided as an array.
+   */
+  valueOptionsSheetName?: string;
+  /**
+   * Method applied before adding the grid content to the workbook
+   * @param {GridExceljsProcessInput} processInput object containing the workbook and the worksheet
+   * @returns {Promise<void>} Modifications are done by mutating the workbook or the worksheet
+   * */
```
m4theushw, line 31:
```suggestion
   * Method called before adding the rows o the workbook.
   * @param {GridExceljsProcessInput} processInput object containing the workbook and the worksheet.
   * @returns {Promise<void>} A promise which resolves after processing the input.
   * */
```

Same for `exceljsPostProcess`.

packages/grid/x-data-grid-pro/src/hooks/features/export/gridExcelExportInterface.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Returns the grid data as an exceljs workbook.
+   * This method is used internally by `exportDataAsExcel`.
+   * @param {GridExcelExportOptions} options The options to apply on the export.
+   * @returns {Promise<Excel.Workbook>} The data in a exceljs workbook object.
+   */
+  getDataAsExcel: (options?: GridExcelExportOptions) => Promise<Excel.Workbook> | null;
+  /**
+   * Downloads and exports an Excel file of the grid's data.
+   * @param {GridExcelExportOptions} options The options to apply on the export.
```
m4theushw, line 58: It also returns a promise.

```suggestion
   * @param {GridExcelExportOptions} options The options to apply on the export.
   * @returns {Promise<void>} A promise which resolves after exporting to Excel.
```

packages/grid/x-data-grid-pro/src/tests/exportExcel.DataGridPro.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    it('should not display export option if not enabled', () => {
+      render(<TestCaseExcelExport components={{ Toolbar: GridToolbar }} />);
+
+      fireEvent.click(screen.queryByRole('button', { name: 'Export' }));
+      expect(screen.queryByRole('menuitem', { name: 'Print' })).not.to.equal(null);
+      expect(screen.queryByRole('menuitem', { name: 'Download as Excel' })).to.equal(null);
+    });
+
+    it('should generate a file if enabled', async () => {
```
m4theushw, line 75:
```suggestion
    it('should generate a file if enabled', () => {
```

packages/grid/x-data-grid-pro/src/tests/exportExcel.DataGridPro.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+      render(<TestCaseExcelExport experimentalFeatures={{ excelExport: true }} />);
+      expect(apiRef.current.getDataAsExcel()).not.to.equal(null);
+    });
+
+    it('should display export option if not enabled', () => {
```
m4theushw, line 80:
```suggestion
    it('should display export option if enabled', () => {
```

packages/grid/x-data-grid-pro/src/tests/exportExcel.DataGridPro.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+              apiRef={apiRef}
+              experimentalFeatures={{ excelExport: true }}
+            />
+          </div>
+        );
+      };
+      render(<Test />);
+
+      const workbook = (await apiRef.current.getDataAsExcel()) as Excel.Workbook;
+      const worksheet = workbook.worksheets[0];
```
m4theushw, line 129:
```suggestion
      const workbook = await apiRef.current.getDataAsExcel();
      const worksheet = workbook!.worksheets[0];
```

Same for "should not export actions columns".

m4theushw (approved):
Looks good. I don't see the warning anymore when opening the full featured demo in Excel. I left some minor suggestions.

flaviendelangle (approved):

packages/grid/x-data-grid-premium/src/components/GridExcelExportMenuItem.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import MenuItem from '@mui/material/MenuItem';
+import { GridExportMenuItemProps } from '@mui/x-data-grid';
```
flaviendelangle, line 4: I tend to import every from `@mui/x-data-grid-pro` in `@mui/x-data-grid-premium` (same for `@mui/x-data-grid-pro/internals`

It has two benefits

1. Double check that we do export everything of the Community Plan inside the Pro Plan
2. If you have to import elements from both Community and Pro Plan, you can do it in a single line

packages/grid/x-data-grid-pro/src/index.ts
flaviendelangle: What is the goal of this export ?

Shouldn't it be exported b one of the other `export *` from this file ?

packages/grid/x-data-grid-premium/src/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export * from '@mui/x-data-grid/locales';
+export * from '@mui/x-data-grid/models';
+export * from '@mui/x-data-grid-pro/models';
+export * from '@mui/x-data-grid/context';
+export * from '@mui/x-data-grid/colDef';
+export * from '@mui/x-data-grid/utils';
+
+export * from './DataGridPremium';
+export * from './hooks';
+export * from './models';
```
alexfauquette, line 19: @flaviendelangle This lines are why I do not export `GridExportFormat`, `GridExportExtension` from `'@mui/x-data-grid/models'`. 

If I do so, it will conflict with the overridden types exported by `export * from './hook` because in `./hook/feature/export` I override types to add excel to the exported formats

flaviendelangle: Make sense :+1:

packages/storybook/package.json
```diff
@@ -21,6 +21,7 @@
     "@mui/x-data-grid": "5.7.0",
     "@mui/x-data-grid-generator": "5.7.0",
     "@mui/x-data-grid-pro": "5.7.0",
+    "@mui/x-data-grid-premium": "5.7.0",
```
oliviertassinari, line 24: Maybe it's ambitious to create a new package in the same PR that adds a big feature 😁

packages/grid/x-data-grid-pro/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mui/x-data-grid-pro",
   "version": "5.7.0",
-  "description": "The commercial edition of the data grid component (MUI X).",
+  "description": "The Pro plan edition of the data grid component (MUI X).",
   "author": "MUI Team",
   "main": "src/index.ts",
   "license": "SEE LICENSE IN LICENSE.txt",
```
oliviertassinari, line 7: **Off-topic** Well, this one is funny.

1. first the file extension is wrong, it's without `.txt`

```suggestion
  "license": "SEE LICENSE IN LICENSE",
```

Proof: https://github.com/mui/mui-x/blob/3d2ff099f4407a3cf88b49f63910c80a88e5882e/packages/grid/x-data-grid-pro/LICENSE

Oops my bad.

2. Second, the license is gone. Compare https://unpkg.com/browse/@mui/x-data-grid-pro@5.5.0/ (OK) and https://unpkg.com/browse/@mui/x-data-grid-pro@5.6.0/ (KO) cc @flaviendelangle

docs/data/data-grid/export/CsvGetRowsToExportRowGrouping.js
```diff
@@ -71,7 +71,6 @@ export default function CsvGetRowsToExportRowGrouping() {
             model: ['commodity'],
           },
         }}
-        experimentalFeatures={{ rowGrouping: true }}
```
oliviertassinari, line 74: Are we sure it's soon enough to remove the experimental flag on this feature? Did we have enough product feedback to be confident it won't require a BC? At a minimum, I would encourage we don't bundle in the same PR a new feature with a change of another feature going from experimental to stable.

packages/grid/x-data-grid-premium/src/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,235 @@
...
+  GridRowId,
+  GridColDef,
+  GridValueFormatterParams,
+  GridApi,
+  ValueOptions,
+} from '@mui/x-data-grid-pro';
+import { buildWarning } from '@mui/x-data-grid/internals';
+import { GridExceljsProcessInput } from '../gridExcelExportInterface';
+
+const getExcelJs = () => import('exceljs');
```
oliviertassinari, line 13: I will be curious to see how this pattern behaves in production. First time I see an attempt at it. I will be positively surprised if it works.

docs/pages/_app.js
```diff
@@ -50,12 +50,20 @@ ponyfillGlobal.muiDocConfig = {
     if (newDeps['@mui/x-data-grid-pro']) {
       newDeps['@mui/material'] = versions['@mui/material'];
       newDeps['@mui/x-data-grid'] = versions['@mui/x-data-grid'];
+      newDeps.exceljs = versions.exceljs;
```
oliviertassinari, line 53: To move to the premium plan?

```suggestion
```

docs/data/data-grid/export/ExcelCustomExport.tsx
```diff
@@ -0,0 +1,326 @@
+import * as React from 'react';
+import 'exceljs';
```
oliviertassinari, line 2: I would assume that fixing the code export dependencies would be enough to make this one work

```suggestion
```

docs/data/data-grid/export/export.md
```diff
@@ -220,16 +224,65 @@ The demo below shows how to add a JSON export.
...
+
+```sh
+ // with npm
+ npm install exceljs
+
+ // with yarn
+ yarn add exceljs
+```
+
+### Customizing the exported file
```
oliviertassinari, line 246: Maybe to keep it one line?

```suggestion
### Customizing the export
```
Screenshot 2022-04-04 at 23 58 38

oliviertassinari (commented):
The demo works on my mac :)

packages/storybook/package.json
flaviendelangle: As explained yesterday, the PR has been approved, then rebased upon #4223 
You can find the clean diff on my repo (https://github.com/flaviendelangle/material-ui-x/pull/20) if you want to have a look at the Excel Export PR itself.

For the review concerning the Premium Plan, please review directly #4223

docs/data/data-grid/export/CsvGetRowsToExportRowGrouping.js
flaviendelangle: For me we only added the experiment feature due to the plan.
The feature itself is marked as "stable" on the doc

image

If the Row Grouping was intended for pro, I would never have been under an experimental flag in the first place.
So removing the experimental flag seems logical to me since we are removing the only reason why it exists.

docs/data/data-grid/export/ExcelCustomExport.tsx
alexfauquette: Not really, if there is no import of the library, `versions.exceljs` will be `undefined` and so it will not be added to libraries
The idea of this import is to add `exceljs` only in the demonstrations that require it

```js
newDeps.exceljs = versions.exceljs;
```

alexfauquette:
@flaviendelangle I modified the way export buttons are added. Now each button is managed by the associated hook.

Instead of adding the React element to the array, we add on object `{component: <TheButton />, componentName: 'button name' }` The `componentName` property is used to sort buttons

packages/grid/x-data-grid-premium/src/hooks/features/export/serializer/excelSerializer.ts
```diff
@@ -0,0 +1,235 @@
...
+          }
+        }
+        row[column.field] = formattedValue?.label ?? formattedValue;
+        break;
+      }
+      case 'boolean':
+      case 'number':
+      case 'date':
+      case 'dateTime':
+        row[column.field] = api.getCellParams(id, column.field).value;
```
cherniavskii, line 101: Not sure why, but this is how dates look like in Apple Numbers:

Screenshot 2022-04-05 at 18 14 29

This is how it looks like in Google Drive (notice 1-day difference comparing to dates in data grid)
Screenshot 2022-04-05 at 18 15 37

docs/data/data-grid/export/CsvGetRowsToExportRowGrouping.js
oliviertassinari: Thanks for the clarification. Ok, so it's only something we talk as an option in https://www.notion.so/mui-org/Version-Sync-Strategy-between-X-and-Core-f877ece7d0d04c339a2b929b96164211#449ab8b4db6d4c228d055a236932e93c.

alexfauquette:
@cherniavskii I think the problems with dates are fixed

- About the 1 day mistake it's due to timezone management. Seems that using UTC fixes it
- About the Apple Numbers date formatting, I specified a number format for columns containing date/dateTime. Hope it will fix the problem

cherniavskii:
@alexfauquette 
Yes, both issues are fixed now!
Here's Apple Numbers:

docs/data/data-grid/export/export.md
```diff
@@ -220,16 +224,85 @@ The demo below shows how to add a JSON export.
...
+This can be used to specify value formatting or to add some colors.
+
+```jsx
+<GridToolbarExport
+  excelOptions={{
+    // replace the dd.mm.yyyy default date format
+    recruitmentDay: { numFmt: 'dd/mm/yyyy' },
+    // set this column in green
+    incomes: { font: { argb: 'FF00FF00' } },
+  }}
```
cherniavskii, line 262: `columnsStyles` is missing

```suggestion
  excelOptions={{
    columnsStyles: {
      // replace the dd.mm.yyyy default date format
      recruitmentDay: { numFmt: 'dd/mm/yyyy' },
      // set this column in green
      incomes: { font: { argb: 'FF00FF00' } },
    }
  }}
```

docs/data/data-grid/export/export.md
```diff
@@ -220,16 +224,85 @@ The demo below shows how to add a JSON export.
...
+You can customize the document using two callback functions:
+
+- `exceljsPreProcess` called **before** adding the rows' dataset.
+- `exceljsPostProcess` called **after** the dataset has been exported to the document.
+
+Both functions receive `{ workbook, worksheet }` as input.
+They are [exceljs](https://github.com/exceljs/exceljs#interface) objects and allow you to directly manipulate the Excel file.
+
+Thanks to these two methods, you can modify the metadata of the exported spreadsheet.
+You can also use it to add add custom content on top or bottom of the worksheet, as follow:
```
cherniavskii, line 277:
```suggestion
You can also use it to add custom content on top or bottom of the worksheet, as follow:
```

docs/data/data-grid/export/export.md
```diff
@@ -220,16 +224,85 @@ The demo below shows how to add a JSON export.
...
+  worksheet.getCell('A2').value = new Date();
+}
+
+function exceljsPostProcess({ worksheet }) {
+  // Add a text after the data
+  worksheet.addRow(); // Add empty row
+
+  const newRow = worksheet.addRow();
+  newRow.getCell(1).value = 'Those data are for internal use only';
+}
```
cherniavskii, line 295: It's not clear from the docs where should those callbacks go.
Maybe add this?

```suggestion
}

// ...

<GridToolbarExport
  excelOptions={{
    exceljsPreProcess,
    exceljsPostProcess,
  }}
/>
```

docs/data/data-grid/export/ExcelCustomExport.tsx
cherniavskii: I've noticed that `Edit in Codesandbox` throws an error:
> Uncaught TypeError: Cannot read properties of undefined (reading 'match')

Which refers to: https://github.com/mui/material-ui/blob/32192fbb6d511b79986731b3cf9dc09a27e0d3aa/docs/src/modules/utils/helpers.ts#L202

docs/data/data-grid/export/ExcelCustomExport.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+  worksheet.getCell('A1').border = {
+    bottom: { style: 'medium', color: { argb: 'FF007FFF' } },
+  };
+
+  worksheet.getCell('A1').font = {
+    name: 'Arial Black',
+    size: 14,
+  };
+  worksheet.getCell('A1').alignment = { vertical: 'top', horizontal: 'center' };
```
cherniavskii, line 298:
```suggestion
  worksheet.getCell('A1').alignment = { vertical: 'top', horizontal: 'center', wrapText: true };
```

Before:
Screenshot 2022-04-26 at 18 45 12
After:
Screenshot 2022-04-26 at 18 45 38

docs/data/pages.ts
```diff
@@ -47,10 +47,15 @@ const pages = [
           },
           { pathname: '/x/api/data-grid/grid-row-params', title: 'GridRowParams' },
           { pathname: '/x/api/data-grid/grid-row-spacing-params', title: 'GridRowSpacingParams' },
+          { pathname: '/x/api/data-grid/grid-csv-export-options', title: 'GridCsvExportOptions' },
           {
             pathname: '/x/api/data-grid/grid-print-export-options',
             title: 'GridPrintExportOptions',
           },
+          {
+            pathname: '/x/api/data-grid/grid-excel-export-options',
```
cherniavskii, line 56: Maybe link to this page in Excel export docs?
Also, it seems to be missing in left nav sidebar

flaviendelangle:
https://github.com/mui/mui-x/pull/3981#pullrequestreview-954819271

For me its the goal yes
@alexfauquette will rebase on master tomorrow and we should merge it just after

docs/data/data-grid/export/export.md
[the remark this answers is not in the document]
flaviendelangle: I think our Excel doc content will depend a lot on the user feedback
It's a feature where the user could do everything without us, but with a high value anyway. So if users struggle to find info in the ExcelJS doc that is often useful for them, I would be in favor of adding examples on our doc.

But only if there is a strong need of course

docs/data/data-grid/export/export.md
```diff
@@ -220,16 +224,96 @@ The demo below shows how to add a JSON export.
...
-> 👍 Upvote [issue #198](https://github.com/mui/mui-x/issues/198) if you want to see it land faster.
-> You will be able to export the displayed data to Excel with an API call, or using the grid UI.
+Columns with types `'boolean'`, `'number'`, `'singleSelect'`, `'date'`, and `'dateTime'` are exported in their corresponding type in Excel. Please ensure the `rows` values have the correct type, you can always [convert them](/components/data-grid/columns/#converting-types) as needed.
+
+{{"demo": "ExcelExport.js", "bg": "inline", "defaultCodeOpen": false}}
+
+This feature relies on [exceljs](https://github.com/exceljs/exceljs).
+To install it:
+
+```sh
```
cherniavskii, line 238: Maybe move installation part to the very top of the section, so that it's harder to miss dependency requirement?

alexfauquette:
@cherniavskii As you noticed, the `import 'exceljs';` was causing internal errors. As a quick fix, I used, such that the regular expression used by the script find the library correctly
```js
import {} from 'exceljs'
```

cherniavskii (approved):

alexfauquette:
It appears transpiling `.ts` to `.js` transform the `import {} from 'exeljs'` to `import 'exeljs'` so the tests can not pass, or the bug will not be fixed for `.js` demo
I opened a PR to fix the bug on core #32492

PR #4010, opened 2022-02-22
[DataGrid] Clean and document the columns selector

flaviendelangle:
- [x] Apply our naming convention to every column selector (and deprecate the old ones)
- [x] Document the correctly named selectors
- [x] Change some internals usages to avoid linear complexity when possible

_____

Here is the list of the new documented selectors.
I tried to be consistent with the sorting / filtering selectors.

I am interest in your opinion on the following points:

- Should we avoid documenting some of those selectors (and maybe stop exporting them at some point) ? This file contains a lot of selectors that were clearly built for internal usages.
- Is the documentation clear enough to understand when a selector value contains the hidden columns or not ?
- Should the filterable columns doc be on the filtering category or on the columns category ?

And if you think some should be named differently.

image


packages/grid/x-data-grid/src/internals/hooks/features/sorting/useGridSorting.ts
```diff
@@ -301,12 +301,10 @@ export const useGridSorting = (
...
     // When the columns change we check that the sorted columns are still part of the dataset
     const sortModel = gridSortModelSelector(apiRef);
-    const latestColumns = allGridColumnsSelector(apiRef);
+    const latestColumns = gridColumnLookupSelector(apiRef);
 
     if (sortModel.length > 0) {
-      const newModel = sortModel.filter((sortItem) =>
-        latestColumns.find((col) => col.field === sortItem.field),
-      );
+      const newModel = sortModel.filter((sortItem) => latestColumns[sortItem.field]);
```
flaviendelangle, line 307: Using the lookup instead of the array avoids the linear complexity here

packages/grid/x-data-grid/src/internals/hooks/features/columns/gridColumnsSelector.ts
```diff
@@ -1,38 +1,82 @@
...
+export const gridColumnDefinitionsSelector = createSelector(
+  gridColumnFieldsSelector,
   gridColumnLookupSelector,
   (allFields, lookup) => allFields.map((field) => lookup[field]),
 );
 
+/**
+ * Get the column visibility model, containing the visibility status of each column.
+ * If a column is not registered in the model, it is visible.
+ * @¢category Columns
```
alexfauquette, line 47:
```suggestion
 * @category Columns
```

alexfauquette (commented):
For `gridColumnsMetaSelector` I don't understand why it is named meta. Maybe "placement". And in the description, we have `positions: number[]` I don't know if it is about the column order, or if it is the x coordinate of the left border of the column, or something else.

The selector `gridColumnsTotalWidthSelector` is redundant with the previous one. The `gridColumnsMetaSelector` could be renamed `gridColumnsPositionSelector` and do not export the `totalWidth`

>Should we avoid documenting some of those selectors (and maybe stop exporting them at some point)? This file contains a lot of selectors that were clearly built for internal usages.

I don't see use cases for `gridVisibleColumnLengthSelector`. For the other ones, they seem to be really specific, but I can imagine people needing them for high-level of customization

> Is the documentation clear enough to understand when a selector value contains the hidden columns or not?

Yes, that's clear, but I'm not sure dev will think about the implication, especially because `gridFilterableColumnDefinitionsSelector` arrives before `gridVisibleColumnDefinitionsSelector`

What about splitting them in two categories `columnsVisible` and `columnsDefined`:
- visible columns: Those selectors does not take into account hidden columns
- defined columns: Those selectors consider all the defined columns, including hidden ones

Such that you keep the alphabetic order on put selectors about visible columns first.


> Should the filterable columns doc be on the filtering category or on the columns category?

Why not put them on both pages?

flaviendelangle:
> The selector gridColumnsTotalWidthSelector is redundant with the previous one. The gridColumnsMetaSelector could be renamed gridColumnsPositionSelector and do not export the totalWidth

I agree that `meta` is weird.
If I understand correctly, you would propose something like this ?

```tsx
/**
 * Get the left position in pixel of each visible columns relative to the left of the first column.
 * PR note: I would replace the reduce with a for loop here becaue a non-pure reduce feels wrong
 */
const gridColumnPositionsSelector = createSelector(
  gridVisibleColumnDefinitionsSelector,
  (visibleColumns) => {
    const positions: number[] = [];

    const totalWidth = visibleColumns.reduce((acc, curCol) => {
      positions.push(acc);
      return acc + curCol.computedWidth;
    }, 0);

    return positions
  },
);

/**
 * Get the summed width of all the visible columns.
 * @category Columns
 */
export const gridColumnsTotalWidthSelector = createSelector(
  gridVisibleColumnDefinitionsSelector,
  gridColumnPositionsSelector,
  (visibleColumns, positions) => {
    return positions[positions.length - 1] + visibleColumns[visibleColumns.length - 1].computedWidth
  },
);
```

> I don't see use cases for gridVisibleColumnLengthSelector. For the other ones, they seem to be really specific, but I can imagine people needing them for high-level of customization

`gridVisibleColumnLengthSelector` is used internally in several places. We often want to have the amount of visible columns for the virtualization or the keyboard navigation for instance.
But of course, you can replace it with a one line alternative. We could not document it an perhaps stop exporting it in v6.

> Yes, that's clear, but I'm not sure dev will think about the implication, especially because gridFilterableColumnDefinitionsSelector arrives before gridVisibleColumnDefinitionsSelector

The selectors are ordered alphabetically, but we could probably improve it.
If we split in 2 categories, I just have to improve the `SelectorsDocs` component to accept a list of categories.

> Why not put them on both pages?

Same as above, if I accept a list of categories on `SelectorsDocs` we can do it.

alexfauquette:
@flaviendelangle Yes, you get what I'm proposing

About splitting `SelectorsDocs` I was thinking about creating two distinct categories:
- ` * @category ColumnsVisible`
- ` * @category Columns`

Such that in the doc you could have 

```md
## Selectors

### Visible columns
Those selectors does not take into account hidden columns

{{"demo": "ColumnsVisibleSelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}

### Defined columns
Those selectors consider all the defined columns, including hidden ones

{{"demo": "ColumnsSelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}
```

That does not allows to put selectors on two pages, but it' a way to split them

flaviendelangle:
image

We could probably improve the order of the selectors in the doc in a future PR

flaviendelangle:
@alexfauquette Should we rename `gridColumnPositionsSelector` => `gridVisibleColumnPositionsSelector` and `gridColumnsTotalWidthSelector` => `gridVisibleColumnsTotalWidthSelector` ?

flaviendelangle:
Yes I was to be 100% explicit
But I agree that it's probably comprehensible without it

docs/data/data-grid/columns/columns.md
```diff
@@ -528,6 +528,22 @@ You can modify this default behavior with column spanning.
...
+
+### Visible columns
+
+Those selectors do not take into account hidden columns
+
+{{"demo": "VisibleColumnsSelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}
+
+### Defined columns
+
+Those selectors consider all the defined columns, including hidden ones
```
alexfauquette, line 541:
```suggestion
Those selectors consider all the defined columns, including hidden ones.
```

docs/data/data-grid/columns/columns.md
```diff
@@ -528,6 +528,22 @@ You can modify this default behavior with column spanning.
...
+
+{{"demo": "ColumnsSelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}
+
+More information about the selectors and how to use them on the [dedicated page](/components/data-grid/state#access-the-state)
```
alexfauquette, line 545:
```suggestion
More information about the selectors and how to use them on the [dedicated page](/components/data-grid/state#access-the-state).
```

alexfauquette (approved):

PR #4028, opened 2022-02-24
[DataGrid] Enable the documentation of the portable state

### LINKED ISSUE #820: [DataGrid] Add methods to save and restore the state

flaviendelangle:
Preview: https://deploy-preview-4028--material-ui-x.netlify.app/components/data-grid/state/#save-and-restore-the-state

Closes #820


flaviendelangle:
@m4theushw @joserodolfofreitas if you have proposals to improve the UI

docs/data/data-grid/state/state.md
```diff
@@ -66,40 +66,36 @@ Some selectors are yet to be documented.
 
 ## Save and restore the state
 
-> ⚠️ This feature isn't implemented yet. It's coming.
+The current state of the grid can be exported using `apiRef.current.exportState()`.
+It can then be restored by either passing it to the `initialState` prop or to the `apiRef.current.restoreState()` method.
+
+Watch out for controlled models and their callbacks (`onFilterModelChange` if you use `filterModel` for instance), the grid will call those callbacks when restoring the state.
```
m4theushw, line 72:
```suggestion
Watch out for controlled models and their callbacks (`onFilterModelChange` if you use `filterModel`, for instance), as the grid will call those callbacks when restoring the state.
```

docs/data/data-grid/state/state.md
```diff
@@ -66,40 +66,36 @@ Some selectors are yet to be documented.
...
+But if the callback is not defined or if calling it does not update the prop value, then the restored value will not be applied.
+
+### Restore the state with `initialState`
+
+> ⚠️ If you restore the page using `initialState` before the data is fetched, the grid will automatically move to the 1st page.
+
+{{"demo": "RestoreStateInitialState.js", "bg": "inline", "defaultCodeOpen": false}}
```
m4theushw, line 79: When I click "Remount the grid" the alert is taking too much space, exceeding the page width. Instead of unmounting and mounting again, we could have a second DataGrid instance and when the button is clicked it transfers the state to the other grid. Unmounting and mounting again works but the user may forget the previous values.

Could you open another PR and delete this demo so it appears as a new file here and we can make suggestions to its code? I guess only renaming won't work.

docs/data/data-grid/state/state.md
```diff
@@ -66,40 +66,36 @@ Some selectors are yet to be documented.
...
-> ⚠️ This feature isn't implemented yet. It's coming.
+The current state of the grid can be exported using `apiRef.current.exportState()`.
+It can then be restored by either passing it to the `initialState` prop or to the `apiRef.current.restoreState()` method.
+
+Watch out for controlled models and their callbacks (`onFilterModelChange` if you use `filterModel` for instance), the grid will call those callbacks when restoring the state.
+But if the callback is not defined or if calling it does not update the prop value, then the restored value will not be applied.
+
+### Restore the state with `initialState`
+
+> ⚠️ If you restore the page using `initialState` before the data is fetched, the grid will automatically move to the 1st page.
```
m4theushw, line 77: We miss a small paragraph introducing what the following demo does.

docs/data/data-grid/state/state.md
flaviendelangle: I do a automerge PR where I delete `RestoreStateInitialState` files so that the PR we are one gets the ful file ?

flaviendelangle: Done

flaviendelangle: > we could have a second DataGrid instance and when the button is clicked it transfers the state to the other grid. Unmounting and mounting again works but the user may forget the previous values.

The demo works with an `initialState` so we have to remount the grid in which we inject.
That can be a 2nd grid instead of unmounting the main one.

docs/data/data-grid/state/state.md
flaviendelangle: I added a small introduction

docs/data/data-grid/state/state.md
```diff
@@ -66,40 +66,39 @@ Some selectors are yet to be documented.
...
+You can pass the state returned by `apiRef.current.exportState()` to the `initialState` prop.
+In the demo bellow, clicking on _Recreate the 2nd grid_ will re-mount the 2nd grid with the current state of the 1st grid.
+
+> ⚠️ If you restore the page using `initialState` before the data is fetched, the grid will automatically move to the 1st page.
+
+{{"demo": "RestoreStateInitialState.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Restore the state with `apiRef` [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+{{"demo": "RestoreStateApiRef.js", "bg": "inline", "defaultCodeOpen": false}}
```
m4theushw, line 86: We miss a paragraph explaining how to interact with the demo.

docs/data/data-grid/state/RestoreStateApiRef.tsx
```diff
@@ -0,0 +1,324 @@
...
+  ) => void;
+  onSubmit: () => void;
+  isValid: boolean;
+}) => {
+  const { label, onLabelChange, onSubmit, isValid } = props;
+  const [isAddingView, setIsAddingView] = React.useState(false);
+
+  if (isAddingView) {
+    return (
+      <React.Fragment>
```
m4theushw, line 180: It doesn't look good. The text field is too tall. I would use `prompt` or put it inside a `Dialog`.

image

docs/data/data-grid/state/RestoreStateInitialState.tsx
```diff
@@ -0,0 +1,80 @@
...
+          components={{ Toolbar: GridCustomToolbar }}
+          componentsProps={{ toolbar: { syncState } }}
+        />
+      </Box>
+      <Box sx={{ width: '100%', height: 300, bgcolor: 'background.paper' }}>
+        <DataGridPro
+          {...data}
+          loading={loading}
+          initialState={savedState.initialState}
+          key={savedState.count}
```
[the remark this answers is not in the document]
flaviendelangle: It is the `initialState` so we need to remount the grid whenever the state changes.
Just causing a re-render won't re-apply the initial state.

That's why I was unmounting / remounting the main grid before.

docs/data/data-grid/state/RestoreStateApiRef.tsx
```diff
@@ -0,0 +1,330 @@
...
+  onSubmit: () => void;
+  isValid: boolean;
+}) => {
+  const { label, onLabelChange, onSubmit, isValid } = props;
+  const [isAddingView, setIsAddingView] = React.useState(false);
+
+  return (
+    <React.Fragment>
+      <Button endIcon={<AddIcon />} onClick={() => setIsAddingView(true)}>
+        Add a custom view
```
m4theushw, line 184: Rename to "Save current view"? The user first changes the params (sorting, filter, column size), then he saves these params.

docs/data/data-grid/state/state.md
flaviendelangle: Done

docs/data/data-grid/state/state.md
```diff
@@ -66,40 +66,42 @@ Some selectors are yet to be documented.
...
+In the demo below, clicking on **Recreate the 2nd grid** will re-mount the 2nd grid with the current state of the 1st grid.
+
+> ⚠️ If you restore the page using `initialState` before the data is fetched, the grid will automatically move to the 1st page.
+
+{{"demo": "RestoreStateInitialState.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Restore the state with `apiRef` [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+You can pass the state returned by `apiRef.current.exportState()` to the `apiRef.current.restoreState` method.
+In the demo below, clicking on **Save current view** will create a snapshot of the state that you can apply on the grid later through the **Custom view** menu.
```
m4theushw, line 87: To avoid questions from the users. We only save what has changed. If I click on "Save current view" without making any change, then make a change, save again and restore the first saved snapshot it won't change anything. That's only how it works.

```suggestion
In the demo below, clicking on **Save current view** will create a snapshot of the changes made in the state, considering the initial state.
You can apply these changes on the grid later selecting a saved view in the **Custom view** menu.
```

docs/data/data-grid/state/RestoreStateApiRef.js
```diff
@@ -0,0 +1,358 @@
...
+    return Object.values(state.views).every(
+      (view) => view.label !== state.newViewLabel,
+    );
+  }, [state.views, state.newViewLabel]);
+
+  const canBeMenuOpened = state.isMenuOpened && Boolean(state.menuAnchorEl);
+  const popperId = canBeMenuOpened ? 'transition-popper' : undefined;
+
+  return (
+    <GridToolbarContainer style={{ height: 48 }}>
```
m4theushw, line 293: The toolbar has an additional space on the top.

image


```suggestion
    <GridToolbarContainer>
```

docs/data/data-grid/state/RestoreStateApiRef.tsx
```diff
@@ -0,0 +1,339 @@
...
+          id={popperId}
+          open={state.isMenuOpened}
+          anchorEl={state.menuAnchorEl}
+          transition
+          placement="bottom-start"
+        >
+          {({ TransitionProps }) => (
+            <Fade {...TransitionProps} timeout={350}>
+              <Paper>
+                <List>
```
m4theushw, line 294: `List` is not accessible. When I open it I can't navigate through the items via keyboard. [`Menu`](https://mui.com/components/menus/#selected-menu) handles that. [`MenuList`](https://mui.com/components/menus/#menulist-composition) can also be used. We use the last to be able to still use a `Popper`

flaviendelangle: I migrated to a `MenuList`

docs/data/data-grid/state/RestoreStateApiRef.js
flaviendelangle: I was for the input, I forgot to remove it

PR #4030, opened 2022-02-24
[core] Implement strategy pattern for pre-processors

flaviendelangle:
We have two quite different types of processors:

1. Pipe processors (currently handled by `useGridPreProcessing`)

If several processors are registered, we want to apply all of them.
For instance, if we have 3 hooks registering an `exportState` processor, we must call all the processor to get the full portable state.

The processors of this kind are:
- `hydrateColumns`
- `scrollToIndexes`
- `columnMenu`
- `exportState`
- `restoreState`
- `rowHeight`

2. Strategy processors (newly handled by `useGridStrategyProcessing`)

If several processors are registered, we want to apply the one matching our current strategy.
For instance, if we have both tree data and row grouping registering a `filtering` processor, we want to only apply the filtering process of the current tree strategy.

The processors of this kind are:
- `rowTreeCreation`
- `filtering`
- `sorting`

Before this PR, `rowTreeCreation` was handled by a custom processing hook and `filtering` / `sorting` had small hacky wrappers to be able to use the pipe processor system.

I wrote a full description of this new kind of processors in the JSDoc of `useGridStrategyProcessing`, to clarify what this hook can and can't do at the moment.

____

Design pattern side note: From what I understand, usually the "strategy pattern" is not responsible for managing the active strategy. It should be the hook asking for a processing that should say which strategy they want to apply.
But in our case, I felt like centralizing this behavior made everything much more simple and avoided risk of desynchronization between the various parts using the strategy.


packages/grid/x-data-grid-pro/src/DataGridPro/useDataGridProComponent.tsx
```diff
@@ -81,10 +85,10 @@ export const useDataGridProComponent = (
   useGridInitializeState(columnPinningStateInitializer, apiRef, props);
   useGridInitializeState(rowGroupingStateInitializer, apiRef, props);
   useGridInitializeState(columnsStateInitializer, apiRef, props);
-  useGridRowGrouping(apiRef, props); // FIXME Needs to be called before the rows state initialization because it registers a rows group builder
```
flaviendelangle, line 84: I moved the strategy processors into `useGridTreeDataPreProcessors` and `useGridRowGroupingPreProccessors` so we can now respect the clean order

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/gridRowGroupingInterfaces.ts
```diff
@@ -2,6 +2,13 @@ export type GridRowGroupingModel = string[];
 
 export interface GridRowGroupingState {
   model: GridRowGroupingModel;
+
+  /**
+   * Tracks the model on the last pre-processing
+   * Allows to check if we need to re-build the grouping columns when the grid upserts a column.
+   * TODO: Move outside of the state
```
flaviendelangle, line 9: I prefixed it to be able to move it in a minor.
@m4theushw maybe doing the same for the cache of the rows would be nice. Even if I agree that in practice probably nobody will try to access it.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/gridRowGroupingUtils.ts
```diff
@@ -155,3 +158,20 @@ export const mergeStateWithRowGroupingModel =
...
+
+export const setStrategyAvailability = (
+  apiRef: React.MutableRefObject<GridApiPro>,
+  disableRowGrouping: boolean,
+) => {
+  let isAvailable: () => boolean;
+  if (disableRowGrouping) {
+    isAvailable = () => false;
+  } else {
+    isAvailable = () => {
```
flaviendelangle, line 170: `isAvailable` is a callback because when initializing the pre-processor, the state is not ready yet so we want just do something like

```ts
apiRef.current.unstable_setStrategyAvailability(
  ROW_GROUPING_STRATEGY,
  gridRowGroupingSanitizedModelSelector(apiRef).length > 0, 
);
```

And we don't want to find the value directly via the props because `gridRowGroupingSanitizedModelSelector` relies on the columns lookup. Computing the same value with the props would be heavy and in theory wrong because it would lack `hydrateColumns` processings.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -351,6 +151,30 @@ export const useGridRowGrouping = (
   /**
    * PRE-PROCESSING
    */
+  const addColumnMenuButtons = React.useCallback<GridPreProcessor<'columnMenu'>>(
```
flaviendelangle, line 154: Just a copy paste, thie processing of this files were in two parts of the hook.

Maybe we could move all of them to the `useGridRowGrouping` hook.
For those recent feature hooks; not a lot would remain (see `useGridTreeData` that has a single function now).

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -420,16 +245,31 @@ export const useGridRowGrouping = (
...
+      apiRef.current.setState((state) => ({
+        ...state,
+        rowGrouping: {
+          ...state.rowGrouping,
+          unstable_sanitizedModelOnLastRowTreeCreation: rowGroupingModel,
+        },
+      }));
 
       // Refresh the column pre-processing
+      // TODO: Add a clean way to re-run a pipe processing without faking a change
```
flaviendelangle, line 262: Here, both for the pipe and straegy processing, we ae doing a hacky thing
I have several very imperfect ideas on how to improve this later.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGroupingPreProcessors.ts
```diff
@@ -107,5 +137,157 @@ export const useGridRowGroupingPreProcessors = (
     [getGroupingColDefs],
   );
 
+  const createRowTree = React.useCallback<GridStrategyProcessor<'rowTreeCreation'>>(
```
flaviendelangle, line 140: This whole method is a copy paste

packages/grid/x-data-grid-pro/src/hooks/features/treeData/useGridTreeDataPreProcessors.tsx
```diff
@@ -82,5 +106,90 @@ export const useGridTreeDataPreProcessors = (
     [props.treeData, getGroupingColDef],
   );
 
+  const createRowTree = React.useCallback<GridStrategyProcessor<'rowTreeCreation'>>(
```
flaviendelangle, line 109: This whole method is a copy paste

packages/grid/x-data-grid/src/hooks/core/preProcessing/gridPreProcessingApi.ts
```diff
@@ -54,14 +49,14 @@ export interface GridPreProcessingApi {
    * Register a pre-processor and emit an event to notify the agents to re-apply the pre-processors.
    * @param {GridPreProcessingGroup} group The name of the group to bind this pre-processor to.
    * @param {number} id An unique and static identifier of the pre-processor.
-   * @param {PreProcessorCallback} callback The pre-processor to register.
+   * @param {GridPreProcessor} callback The pre-processor to register.
    * @returns {() => void} A function to unregister the pre-processor.
    * @ignore - do not document.
    */
-  unstable_registerPreProcessor: (
+  unstable_registerPreProcessor: <G extends GridPreProcessingGroup>(
```
flaviendelangle, line 56: I improved the typing of the pipe processors to match the strategy processors format

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -317,6 +321,77 @@ export const useGridRows = (
...
+    setRows,
+    updateRows,
+    setRowChildrenExpansion,
+    getRowNode,
+  };
+
+  /**
+   * EVENTS
+   */
+  const groupRows = React.useCallback(() => {
```
flaviendelangle, line 338: Copy paste of `handleGroupRows`

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
```diff
@@ -0,0 +1,133 @@
...
+
+      const processor = groupCache[currentStrategy] as GridStrategyProcessor<any>;
+      return processor(params);
+    },
+    [apiRef],
+  );
+
+  const getActiveStrategy = React.useCallback<
+    GridStrategyProcessingApi['unstable_getActiveStrategy']
+  >(() => {
```
m4theushw, line 106: Should it receive the group as argument and return which strategy is active for that specific group?

Currently this hook can only be used to group rows, but it may be useful for other things too.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
```diff
@@ -280,11 +80,11 @@ export const useGridRowGrouping = (
       const currentModel = gridRowGroupingModelSelector(apiRef);
       if (currentModel !== model) {
         apiRef.current.setState(mergeStateWithRowGroupingModel(model));
-        updateRowGrouping();
+        setStrategyAvailability(apiRef, props.disableRowGrouping);
```
m4theushw, line 83: Could we put this in an effect to avoid having to call `setStrategyAvailability` in all situations where the model may change? In practice what I want to say is to memoize the callback that determines if a strategy is available, then the effect would have this callback as dependency. When the reference changes, `unstable_setStrategyAvailability` is called.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
m4theushw: A bit related to https://github.com/mui/mui-x/pull/4030#discussion_r817027799

Maybe the strategy pre-processor that should change and trigger again the pre-processing. This already happens with the pipe pre-processors where in the next render, if one processor has changed, they are called again.

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
flaviendelangle: Done
I added a concept of `strategyGroup`, with only one existing right now (`rowTree` which contains the 3 pre-processors `rowTreeCreation`, `sorting`, `filtering`)
That way, if we want to create pre-proccessors with an independent list of strategies, we will just have to put it in another strategy group).

Everything is transparent, except for `getActiveStrategy` which takes the strategy group as a param, as you proposed.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
flaviendelangle: The problem is that `setStrategyAvailability` must be updated when the columns are updated, but it does not have the column state as an argument, because the first time we are calling it, the state is not ready yet.

I don't see how we could both initialize it in `useGridRowGroupingPreProcessors` and put it in an `useEffect` in `useGridRowGrouping`.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
flaviendelangle: It is also the case here
But in some scenarios, you don't have any deps of the `useCallback` that changes because the change is hidden deep inside the state.

For instance, the `updateColumns` a few lines above is to trigger the `hydrateColumns` pre-process.

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
```diff
@@ -0,0 +1,152 @@
...
+    (strategyName, processorName, processor: GridStrategyProcessor<any>) => {
+      if (!strategiesCache.current[processorName]) {
+        strategiesCache.current[processorName] = {};
+      }
+
+      const groupPreProcessors = strategiesCache.current[processorName] as {
+        [strategyName: string]: GridStrategyProcessor<any>;
+      };
+      const previousProcessor = groupPreProcessors[strategyName];
+      if (previousProcessor !== processor) {
```
m4theushw, line 72: You could early return the unmount function here if `previousProcessor === processor`. The same if `previousProcessor` is not defined.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
m4theushw: I might be missing something but the effect will only be called when the state is already initialized and the pre-processors are registered. `gridRowGroupingSanitizedModelSelector` is used to get the number of items, so the length could a dependency.

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/gridStrategyProcessingApi.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export type GridStrategyProcessor<P extends GridStrategyProcessorName> = (
+  params: GridStrategyProcessingLookup[P]['params'],
+) => GridStrategyProcessingLookup[P]['value'];
+
+export interface GridStrategyProcessingApi {
+  /**
+   * Register a strategy processor.
+   * If the strategy is active, emit and event to notify the agents to re-apply the processor.
```
m4theushw, line 44:
```suggestion
   * Registers a strategy processor.
   * If the strategy is active, it emits an event to notify the agents to re-apply the processor.
```

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
flaviendelangle: Not sure it is a lot cleaner

We still have to return the same cleanup function when the processing did not change.

packages/grid/x-data-grid-pro/src/hooks/features/rowGrouping/useGridRowGrouping.tsx
flaviendelangle: On the effect of `useGridRowGrouping` that calls `setRowGroupingModel` yes.
But for the `useFirstRender` of `useRowGroupingPreProcessing` that inits the strategy before the initialization of the rows state, the columns state is not ready so we can't pass `gridRowGroupingSanitizedModelSelector` as a dep.

We could separate the two behaviors but I'm not sure it's better.

m4theushw (approved):

PR #4031, opened 2022-02-24
[test] Reduce memory usage to run unit tests

m4theushw:
In #3963 I rewrote all of the tests while keeping the old tests still working. As consequence, `yarn test:unit` started to run out of memory in CircleCI or timing-out in some non-related test case. Initially I though this could be a memory leak in the code of the new editing API but by adding more tests in HEAD ([81282b6](https://github.com/mui/mui-x/pull/4031/commits/81282b65c12f33f9986830632a5cd1c97293b6ab)) the same happened. So I took a heap snapshot while running the unit tests to see what was eating so much memory. I noticed that the new approach to remove listeners added in #3310 was leaving a reference alive in memory that was causing many objects from not getting removed by the garbage collector.

#### Some concepts to have in mind:

_From https://www.yourkit.com/docs/java/help/sizes.jsp_ (it's Java but applies to JavaScript too)

- "Shallow size" is the amount of memory allocated for a specific object.
- "Retained size" is the amount of memory allocated for a specific object plus the shallow sizes of the objects that are accessible from it. Once a object is garbage collected, all this amount will also be freed from the memory.

image

I believe that this reference was kept in memory because when we unmock the clock after each test  (only those that use the fake timer), we don't run the pending timers. To solve this problem I used a simple solution of cleaning the `timeouts` object after each test. This was accomplished by adding an `unstable_resetCleanupTracking` function. After this `yarn test:unit` passed: [efb8737](https://github.com/mui/mui-x/pull/4031/commits/efb873707809fb88f51ef88ed4a7beffa10e2b8b).

To see how much memory is being allocated while running the tests use the following command:

```bash
cross-env NODE_ENV=test node --trace_gc --expose_gc ./node_modules/.bin/mocha 'packages/**/*.test.{js,ts,tsx}' --exclude '**/node_modules/**' --timeout 3000
```

Add `--inspect-brk` to able to debug in Chrome and use the Memory tab.

After this fix I also noticed that CircleCI is using less memory to run the pipeline:

_The last [build](https://app.circleci.com/pipelines/github/mui/mui-x/13940/workflows/d073b9dd-37d3-4a8b-acad-80198b78a740/jobs/81709/resources) in master._

image

_After running [dfd4a68](https://github.com/mui/mui-x/commit/dfd4a68c1a3441f414ba73eb7e52ab65e0e0fc4e)_

image

I might be biased but the curve seems flatter, while the previous one shows an uptrend.


flaviendelangle:
I have the same memory issue in #3451

m4theushw:
@flaviendelangle Could you merge this PR into #3451 and see if it runs? #3963 became green after it.

packages/grid/x-data-grid-pro/src/tests/detailPanel.DataGridPro.test.tsx
```diff
@@ -158,6 +158,20 @@ describe('<DataGridPro /> - Detail panel', () => {
     expect(screen.queryByText('Detail')).to.equal(null);
   });
 
+  [...new Array(100).keys()].forEach((i) => {
```
m4theushw, line 161: Only for demonstration purposes. They will be removed before merge.

test/utils/setup.js
```diff
@@ -71,6 +73,12 @@ function throwOnUnexpectedConsoleMessages(methodName, expectedMatcher) {
   });
 }
 
+mochaHooks.afterEach.push(function restoreDefaultSandbox() {
+  // Restore Sinon default sandbox to avoid memory leak
+  // See https://github.com/sinonjs/sinon/issues/1866
+  sinon.restore();
```
m4theushw, line 79: It's a recommendation from https://sinonjs.org/releases/latest/general-setup/

packages/grid/x-data-grid/src/internals/hooks/utils/useGridApiEventHandler.ts
```diff
@@ -92,17 +92,13 @@ export function createUseGridApiEventHandler(registry: CleanupTracking) {
...
-      ? new FinalizationRegistryBasedCleanupTracking()
-      : new TimerBasedCleanupTracking();
-}
+const registry =
+  typeof FinalizationRegistry !== 'undefined'
+    ? new FinalizationRegistryBasedCleanupTracking()
+    : new TimerBasedCleanupTracking();
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const unstable_resetCleanupTracking = () => registry.reset();
```
m4theushw, line 101: It's weird but `this` inside `registry.reset` becomes `undefined` if I use:

```ts
export const unstable_resetCleanupTracking = registry.reset;
```

flaviendelangle:
@m4theushw it seems to fix our problem even with several hundreds of new tests.

Before: https://app.circleci.com/pipelines/github/mui/mui-x/13403/workflows/a1009314-15c9-4209-ad45-943afc3d41cd/jobs/78459
After: https://app.circleci.com/pipelines/github/mui/mui-x/13974/workflows/dbb648cf-5ba0-4d07-90ea-2a45de0b5e42/jobs/81909

You can push code on this PR is you want to try things :+1:

packages/grid/x-data-grid/src/internals/utils/cleanupTracking/TimerBasedCleanupTracking.ts
```diff
@@ -4,24 +4,37 @@ import { CleanupTracking, UnregisterToken, UnsubscribeFn } from './CleanupTracki
 const CLEANUP_TIMER_LOOP_MILLIS = 1000;
 
 export class TimerBasedCleanupTracking implements CleanupTracking {
-  timeouts = new Map<number, NodeJS.Timeout>();
+  timeouts? = new Map<number, NodeJS.Timeout>();
```
alexfauquette, line 7: What does this `?` means? Never encountered such a character before an `=`

alexfauquette (approved):
Thanks for the explanation, I learned a lot 👍

packages/grid/x-data-grid/src/internals/utils/cleanupTracking/TimerBasedCleanupTracking.ts
m4theushw: `?` is used to tell that the attribute can be `undefined`. Using it is the same as:

```tsx
export class TimerBasedCleanupTracking implements CleanupTracking {
  timeouts: Map<number, NodeJS.Timeout> | undefined = new Map<number, NodeJS.Timeout>();
}
```

Only classes and interfaces allow it. More information: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#optional-class-properties

PR #4034, opened 2022-02-25
[DataGridPro] Row reorder

### LINKED ISSUE #206: [data grid] Implement Row reorder
### Benchmark

- https://ag-grid.com/javascript-grid-row-dragging/
- https://www.telerik.com/kendo-react-ui/components/grid/advanced-features/row-reordering/
- https://ant.design/components/table/#components-table-demo-drag-sorting
- https://ej2.syncfusion.com/react/demos/#/material/grid/drag-drop-within-grid
- https://handsontable.com/docs/8.0.0/demo-moving.html
- https://adazzle.github.io/react-data-grid/docs/examples/row-reordering
- https://adazzle.github.io/react-data-grid/canary/?path=/story/demos--rows-reordering
- https://github.com/mbrn/material-table/issues/1219

DanailH:
Fixes #206 

Preview: https://deploy-preview-4034--material-ui-x.netlify.app/x/react-data-grid/rows/#row-reorder

## Row  reorder

This is a `DataGridPro` feature.

The approach I took is the same as the column reorder. I added a new entry in the state to track the dagging row ID. I added one callback - `onRowOrderChange` to match the column reordering in terms of API. The difference between the two callbacks is that I'm not calling the row to reorder one when there is sorting on filtering (thought about it but it has to be called for reach row and that is just too much, but I'm open to suggestions).
Last but not least - there is a prop on the grid itself to disable the row reorder as it is enabled by default (same as the column reorder) - `disableRowReorder`.

Follow up issue:
- https://github.com/mui/mui-x/issues/4380

---

### TODO:

- [x] Tests
- [x] Docs
- [x] Disable row reorder when Tree Data / Row Grouping
- [x] Disable row reorder when sorting
- [ ] Merge https://github.com/mui/material-ui/pull/31875 (to be done after this PR is merged)
- [x] Column pinning UI (**not a blocker**)


packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -317,6 +318,11 @@ export interface DataGridPropsWithDefaultValues {
    * @default false
    */
   disableColumnResize: boolean;
+  /**
+   * If `true`, reordering rows is disabled.
+   * @default false
+   */
+  disableRowReorder: boolean;
```
flaviendelangle, line 325: You should be able to define this one in `DataGridProPros` and don't put a default value on `DataGrid` because the hook `useGridRowReorder` is not called at all on the community plan.

See `disableColumnPinning` for equivalent

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -317,6 +318,11 @@ export interface DataGridPropsWithDefaultValues {
    * @default false
    */
   disableColumnResize: boolean;
+  /**
+   * If `true`, reordering rows is disabled.
```
flaviendelangle, line 322:
```suggestion
   * If `true`, the reordering of rows is disabled.
```

packages/grid/x-data-grid/src/models/events/gridEventLookup.ts
```diff
@@ -27,6 +27,22 @@ export interface GridRowEventLookup {
   rowDoubleClick: { params: GridRowParams; event: React.MouseEvent<HTMLElement> };
   rowMouseEnter: { params: GridRowParams; event: React.MouseEvent<HTMLElement> };
   rowMouseLeave: { params: GridRowParams; event: React.MouseEvent<HTMLElement> };
+  rowDragStart: {
```
flaviendelangle, line 30: Those new events are pro-only and can be defined on the module augmentation of `x-data-grid-pro` (search for `GridEventLookupPro`)

packages/grid/x-data-grid/src/hooks/features/sorting/useGridSorting.ts
```diff
@@ -204,6 +204,31 @@ export const useGridSorting = (
     [apiRef],
   );
 
+  const setRowIndex = React.useCallback<GridSortApi['setRowIndex']>(
+    (rowId, targetIndexPosition) => {
+      const allRows = gridSortedRowIdsSelector(apiRef);
```
flaviendelangle, line 209: You are still updating the sorting list.
Was updating `state.rows.ids` with the same value as `updatedRows` if this function + cleaning to sorting model to complicated or is it incompatible with something ?

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+      originRowIndex.current = apiRef.current.getRowIndex(params.id);
+    },
+    [isRowReorderDisabled, classes.rowDragging, logger, apiRef],
+  );
+
+  const handleDragEnter = React.useCallback<
+    GridEventListener<GridEvents.cellDragEnter | GridEvents.rowDragEnter>
+  >((params, event) => {
+    event.preventDefault();
```
m4theushw, line 121: Why do we need this listener? For the column reorder I remember it was added to prevent a conflict with React DnD. But the listener available in the column reorder prevents the event even when there's no column being dragged so this listener here could be removed, I think.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,220 @@
...
+        const hasMovedTop =
+          cursorMoveDirectionY === CURSOR_MOVE_DIRECTION_TOP && dragRowIndex > targetRowIndex;
+        const hasMovedBottom =
+          cursorMoveDirectionY === CURSOR_MOVE_DIRECTION_BOTTOM && dragRowIndex < targetRowIndex;
+
+        if (hasMovedTop || hasMovedBottom) {
+          apiRef.current.setRowIndex(dragRowId, targetRowIndex);
+
+          const rowOrderChangeParams: GridRowOrderChangeParams = {
+            row: apiRef.current.getRow(dragRowId),
```
m4theushw, line 160: Should the other row be included too?

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,220 @@
...
+        const targetRowIndex = apiRef.current.getRowIndex(params.id);
+        const dragRowIndex = apiRef.current.getRowIndex(dragRowId);
+
+        const cursorMoveDirectionY = getCursorMoveDirectionY(cursorPosition.current, coordinates);
+        const hasMovedTop =
+          cursorMoveDirectionY === CURSOR_MOVE_DIRECTION_TOP && dragRowIndex > targetRowIndex;
+        const hasMovedBottom =
+          cursorMoveDirectionY === CURSOR_MOVE_DIRECTION_BOTTOM && dragRowIndex < targetRowIndex;
+
+        if (hasMovedTop || hasMovedBottom) {
```
m4theushw, line 156: I know this was copied from the column reorder but is there a need for these checks here? We already know `params.id !== dragRowId`.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  React.useEffect(() => {
+    return () => {
+      clearTimeout(removeDnDStylesTimeout.current);
+    };
+  }, []);
+
+  // TODO: remove sortModel check once row reorder is sorting compatible
+  // remove treeDepth once row reorder is tree compatible
+  const isRowReorderDisabled = React.useCallback((): boolean => {
```
m4theushw, line 85: Convert to `React.useMemo`?

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,220 @@
...
+      // Prevent drag events propagation.
+      // For more information check here https://github.com/mui/mui-x/issues/2680.
+      event.stopPropagation();
+
+      dragRowNode.current = event.currentTarget;
+      dragRowNode.current.classList.add(classes.rowDragging);
+
+      apiRef.current.setState((state) => ({
+        ...state,
+        rowReorder: { ...state.rowReorder, dragRow: params.id },
```
m4theushw, line 105: In https://github.com/mui/mui-x/issues/820#issuecomment-897906608 I proposed to clean a bit the state and replace some keys with internal state. This would include `columnReorder`. Could we use a simple `React.useState` here? We only need to track which row is being dragged.

m4theushw (commented):
Great work. I did a first review looking more into the API. I think this feature deserves some more work in the docs, specially with server-side examples.

> Last but not least - there is a prop on the grid itself to disable the row reorder as it is enabled by default (same as the column reorder) - disableRowReorder.

Should it be disabled by default? Users can't select the text of the cells anymore. Even with `user-select: all` is not the same experience. I think that the row reorder feature depends a lot in the backend being able to save the data and some applications don't care about the ordering, while others care. If everybody can reorder rows and the backend is not properly set up to save the order it may cause problems for developers.

To still support text selection we could add a column with a handle, exclusively for row dragging. That way, dragging other cells does nothing. It will be tricky as we need to switch the `draggable` attribute of the row only when this toggle is clicked, but it's feasible.

docs/data/data-grid/getting-started/getting-started.md
```diff
@@ -183,7 +183,7 @@ The enterprise components come in two plans: Pro and Premium.
 | **Row**                                                                                   |           |                                    |                                            |
 | [Row height](/components/data-grid/rows/#row-height)                                      |    ✅     |                 ✅                 |                     ✅                     |
 | [Row spanning](/components/data-grid/rows/#row-spanning)                                  |    🚧     |                 🚧                 |                     🚧                     |
-| [Row reordering](/components/data-grid/rows/#row-reorder)                                 |    ❌     |                 🚧                 |                     🚧                     |
+| [Row reordering](/components/data-grid/rows/#row-reorder)                                 |    ❌     |                 ✅                 |                     ✅                     |
```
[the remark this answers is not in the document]
DanailH: PR is opened -> https://github.com/mui/material-ui/pull/31875

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
DanailH: Yes, I tried that but the problem is that I need to access `rootProps.disableRowReorder` inside the `GridRow.tsx` because otherwise even tho the feature hook doesn't run the drag events and the `draggable` are still added to the demo and you are able to visually drag a row.

This feature has more in common with the `disableColumnReorder` in terms of the setup.

packages/grid/x-data-grid/src/hooks/features/sorting/useGridSorting.ts
DanailH: I didn't understand the suggestion. From what I saw the order of the rows is taken from the `state.sorting.sorterRows`. the other `state.rows.ids` isn't used to determine the actual order, no? Maybe I'm missing something?

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
flaviendelangle: True if it is used in a community component we are stuck :/

packages/grid/x-data-grid/src/hooks/features/sorting/useGridSorting.ts
flaviendelangle: My proposal was when you reorder the column:

1. Update `state.rows.ids` with the actual order you now have (the same value you are currently storing in `state.sorting.sortedRows`)
2. Remove all the sort model items

That way, the sort engine will just take the order of `state.rows.ids` because it has no sorting rule.

Advantages:

- If the user reorder a row, the add a sorting rule, then remove it, it will not loose the reorder he did
- We are not updating a derived state that could be changed (for instance if a user edit a cell, it will re-apply the sorting and thus loose the reorder)

Disadvantages:

- If the sort model is being controlled, it can be harder to manage

For me the subject is the same as #4160
If we remove the automatic sorting / filtering when the data is partially update
Then updating `state.sorting.sortedRows` can be viable (and for the 1st advantage, the user can manually re-inject its rows inside the `props.rows` if he wants when the order changes).
If not, the editing part will probably be problematic.

packages/grid/x-data-grid/src/models/events/gridEventLookup.ts
DanailH: It same reason as https://github.com/mui/mui-x/pull/4034#discussion_r829908169, unfortunately.

packages/grid/x-data-grid/src/hooks/features/sorting/useGridSorting.ts
DanailH: Updated. I also moved this API to the `useGridRows`.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
DanailH: Not sure? Why would you need the other row, the two Ids are provided?

m4theushw: It's mostly for convenience when the developer needs to update the order of the rows in the database. Imagine that someone is listening to `rowOrderChange` and only receives the row that was dragged, not also the row where it was dragged to. Maybe we wait for someone to ask for it.

janderssen:
Hi guys,

The work done on this is terrific, and I know it is a feature us PRO customers are so looking forward to using, so I decided to give a big encouragement message :-)

Great work, and looking forward to using it going forward!

DanailH:
> To still support text selection we could add a column with a handle, exclusively for row dragging. That way, dragging other cells does nothing. It will be tricky as we need to switch the `draggable` attribute of the row only when this toggle is clicked, but it's feasible.

@m4theushw I thought about the problem and as a solution, we can have a first cell (like with the checkbox) that indicates that you can drag it and by dragging it you will drag the entire row. The column with those cells will only be visible when `enableRowReorder` is used (change from `disableRowReorder`).

DanailH:
Just a quick update - I've reworked the reordering - how a special column is added and you need to drag an icon in order to reorder a row (that solves the text selection problem). I also made the row reorder disabled my default.

Does is it still to be updated. I'll work on that but if we can review the implementation that would be great.

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -708,4 +709,9 @@ export interface DataGridPropsWithoutDefaultValue extends CommonProps {
     newRow: GridRowModel,
     oldRow: GridRowModel,
   ) => Promise<GridRowModel> | GridRowModel;
+  /**
+   * If `true`, the reordering of rows is enabled.
+   * @default false
+   */
+  enableRowReorder?: boolean;
```
flaviendelangle, line 716: This in a pro-only prop with a default value

1. Move this line to `DataGridProPropsWithDefaultValues` without the `?`
2. Add a default value to `DATA_GRID_PRO_PROPS_DEFAULT_VALUES`
3. Remove the forced prop in `enableRowReorder` and `DATA_GRID_FORCED_PROPS`


That way you have `?: boolean` for `DataGridProProps` and `: boolean` for `DataGridProProcessedProps` which is used internally.
And on the community plan, the props is never used so if the user passes `true` it won't matter (like for the tree data for instance).

____

Concerning the name
We use `props.pagination` not `props.enablePagination`, same for `props.treeData`
Should we use `props.rowReordering` for instance ?

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
```diff
@@ -205,4 +205,11 @@ export interface DataGridProPropsWithoutDefaultValue
    * @returns {JSX.Element} The row detail element.
    */
   getDetailPanelContent?: (params: GridRowParams) => React.ReactNode;
+  /**
+   * Callback fired when a row is being reordered.
+   * @param {GridRowOrderChangeParams} params With all properties from [[GridRowOrderChangeParams]].
+   * @param {MuiEvent<{}>} event The event object.
+   * @param {GridCallbackDetails} details Additional details for this callback.
+   */
+  onRowOrderChange?: GridEventListener<GridEvents.rowsScrollEnd>;
```
flaviendelangle, line 214:
```suggestion
  onRowOrderChange?: GridEventListener<GridEvents.rowOrderChange>;
```

packages/grid/x-data-grid-pro/src/models/gridRowOrderChangeParams.ts
```diff
@@ -0,0 +1,19 @@
+import { GridRowModel } from '@mui/x-data-grid';
+
+/**
+ * Object passed as parameter of the row order change event.
+ */
+export interface GridRowOrderChangeParams {
+  /**
+   * The row data.
+   */
+  row: GridRowModel | null;
```
flaviendelangle, line 10: Can it really be null ?

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -336,12 +336,39 @@ export const useGridRows = (
...
+
+      apiRef.current.setState((state) => ({
+        ...state,
+        rows: {
+          ...state.rows,
+          ids: updatedRows,
+        },
+      }));
+      apiRef.current.forceUpdate();
+      apiRef.current.applySorting();
```
flaviendelangle, line 360: Why do we apply sorting after the `forceUpdate` ?

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
```diff
@@ -24,6 +24,12 @@ export interface GridRowApi {
    * @param {GridRowModel[]} rows The new rows.
    */
   setRows: (rows: GridRowModel[]) => void;
+  /**
+   * Moves a row from its original position to the position given by `targetIndexPosition`.
+   * @param {GridRowId} rowId The row id
+   * @param {number} targetIndexPosition The new position (0-based).
+   */
+  setRowIndex: (rowId: GridRowId, targetIndexPosition: number) => void;
```
flaviendelangle, line 32:
```suggestion
  setRowIndex: (rowId: GridRowId, targetPosition: number) => void;
```

or

```suggestion
  setRowIndex: (rowId: GridRowId, targetIndex: number) => void;
```

Same for `oldIndexPosition` on the method body.

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  const draggableEventHandlers = {
+    onDragStart: publish(GridEvents.rowDragStart),
+    onDragOver: publish(GridEvents.rowDragOver),
+    onDragEnd: publish(GridEvents.rowDragEnd),
+  };
+
+  return (
+    <div className={gridClasses.actionsCell} draggable={isDraggable} {...draggableEventHandlers}>
```
m4theushw, line 60: `gridClasses.actionsCell`?

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <rootProps.components.RowReorderIcon />
+
+      <div className={gridClasses.rowDraggableContainer}>{params.row.id}</div>
```
m4theushw, line 63: `row.id` is the `id` attribute of the row. If the user has `getRowId` this won't work. You can use the `id` prop.

How could we allow users to customize the content shown during dragging? The ID is not useful.

AG Grid requires one column to have the `rowDrag` flag so they know which content to render: https://ag-grid.com/javascript-data-grid/row-dragging/#enabling-row-dragging

image

One way is to render the `id` as default but encourage users to customize it by manually adding a column with the `__reorder__` field. We use a similar approach in the master/detail to allow to change the position of the toggle column: https://mui.com/components/data-grid/group-pivot/#customizing-the-detail-panel-toggle

packages/grid/x-data-grid/src/components/GridRow.tsx
```diff
@@ -126,7 +126,7 @@ function GridRow(props: React.HTMLAttributes<HTMLDivElement> & GridRowProps) {
   const publish = React.useCallback(
     (
         eventName: keyof GridRowEventLookup,
-        propHandler: React.MouseEventHandler<HTMLDivElement> | undefined,
+        propHandler?: React.MouseEventHandler<HTMLDivElement> | undefined,
```
m4theushw, line 129: A click in the reorder handle shouldn't fire `rowClick`. You need to add a check for it in this file. The idea is:

`cellClick` is triggered for clicks everywhere
`rowClick` is not triggered on clicks in built-in columns with interactive elements

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
DanailH: I think that would be ideal. To let the user provide a value for that column but not render it. Also, the `valueFormatter`will work in that case so it should be customizable enough. I'll update the implementation.

DanailH: I've updated it so that if you pass a value to the `__reorder__` filed or a row it will use that. If no value is passed it will default to using the `id`. I think this + the additional classes on both the cell and inner container will make it customizable enough to start using it as you can also provide a React element to be a field value.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+        const rowOrderChangeParams: GridRowOrderChangeParams = {
+          row: apiRef.current.getRow(dragRowId),
+          targetIndex: originRowIndex.current!,
+          oldIndex: originRowIndex.current!,
+        };
+        apiRef.current.publishEvent(GridEvents.rowOrderChange, rowOrderChangeParams);
+
+        originRowIndex.current = null;
+      }
```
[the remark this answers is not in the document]
DanailH: I updated it so that the rowOrderChange is only emitted when the dragging stops. I think that would be optimal as you don't want to get spammed with multiple reorders, we should only care about the last position.

docs/data/data-grid/rows/rows.md
```diff
@@ -146,14 +146,26 @@ Row spanning allows to change this default behavior.
...
+
+In addition, row reordering emits the following events that can be imported:
+
+- `rowDragStart`: emitted when dragging of a row starts.
+- `rowDragOver`: emitted when dragging a row over another row.
+- `rowDragEnd`: emitted when dragging of a row stops.
+
+> ⚠️ For now, row reordering is disabled if working is applied to the grid.
+>
+> In addition, if row grouping or Tree Data is used the row reordering is also disabled.
```
cherniavskii, line 167: Maybe move it to the warning above so that all cases when reordering is disabled are in one place?

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,145 @@
...
+      // For more information check here https://github.com/mui/mui-x/issues/2680.
+      event.stopPropagation();
+
+      dragRowNode.current = event.currentTarget;
+      dragRowNode.current.classList.add(classes.rowDragging);
+
+      setDragRowId(params.id);
+
+      removeDnDStylesTimeout.current = setTimeout(() => {
+        dragRowNode.current!.classList.remove(classes.rowDragging);
```
cherniavskii, line 78: Why is it called on drag start?

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -0,0 +1,145 @@
+import * as React from 'react';
```
cherniavskii, line 1: I've noticed that editing + row reordering combination acts weird:
- start cell editing
- reorder row
- end cell editing
- row is back at original position

https://user-images.githubusercontent.com/13808724/162024287-d7cf77f8-a58d-42a1-b858-194f9dc50fa6.mov

Codesandbox: https://codesandbox.io/s/roworderinggrid-material-demo-forked-ldorwe

Should editing be stopped once reordering starts? What do you think?

docs/data/data-grid/rows/rows.md
flaviendelangle:
```suggestion
> In addition, if row grouping or tree data is being used, the row reordering is also disabled.
```

Is "A is used" a valid formulation ?

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+  gridSortModelSelector,
+  useGridApiContext,
+  useGridRootProps,
+  useGridSelector,
+} from '@mui/x-data-grid';
+import { DataGridProProcessedProps } from '../models/dataGridProProps';
+
+const GridRowReorderCell = (params: GridRenderCellParams) => {
+  const apiRef = useGridApiContext();
+  const rootProps = useGridRootProps();
```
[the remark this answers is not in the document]
flaviendelangle: From `../hooks/utils/useGridRootProps` would be even better.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
DanailH: It's a "hack", same is done for the column reorder. Basically, we want to add a class to the DOM element that is about to be dragged before the native dragging API takes over and the element is out of the DOM flow, but because there is only 1 element we want to preserve the original styling while dragging.

docs/data/data-grid/rows/rows.md
```diff
@@ -146,14 +146,26 @@ Row spanning allows to change this default behavior.
...
 
 Row reorder is used to rearrange rows by dragging the row with the mouse.
 
+By default, `DataGridPro` does not allow row reordering. To enable reordering on all rows by dragging the special row reordering cells and moving it up or down, use the `rowReordering` prop.
+
+{{"demo": "RowOrderingGrid.js", "disableAd": true, "bg": "inline"}}
+
+To capture changes in the order of the dragged row use the `onRowOrderChange` callback.
+
+In addition, row reordering emits the following events that can be imported:
```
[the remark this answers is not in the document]
DanailH: I reworked the sentence to make more sense.

packages/grid/x-data-grid/src/components/containers/GridRootStyles.ts
```diff
@@ -282,11 +285,14 @@ export const GridRootStyles = styled('div', {
     [`& .${gridClasses.booleanCell}[data-value="false"]`]: {
       color: theme.palette.text.disabled,
     },
-    [`& .${gridClasses.actionsCell}`]: {
+    [`& .${gridClasses.actionsCell}, & .${gridClasses.rowReorderCell}`]: {
       display: 'inline-flex',
```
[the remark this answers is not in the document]
DanailH: Updated.

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
DanailH: I've disabled it, nice catch.

fourteenmeister:
React 18 support needing with Reordering rows.

DanailH:
> React 18 support needing with Reordering rows.

I'm sorry I didn't understand what you meant. Can you rephrase it?

fourteenmeister:
> I'm sorry I didn't understand what you meant. Can you rephrase it?

Merge https://github.com/mui/mui-x/pull/4332 into current branch please

DanailH:
@m4theushw I've updated the demo so if it's ok I think we can merge this and have the feature as part of this week's release.

docs/data/data-grid/rows/RowOrderingGrid.tsx
```diff
@@ -1,17 +1,61 @@
 import * as React from 'react';
-import { DataGridPro } from '@mui/x-data-grid-pro';
+import {
+  DataGridPro,
+  GridRowModel,
+  GridRowOrderChangeParams,
+} from '@mui/x-data-grid-pro';
 import { useDemoData } from '@mui/x-data-grid-generator';
 
+function updateRowPossition(
```
flaviendelangle, line 9:
```suggestion
function updateRowPosition(
```

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const GridRowReorderCell = (params: GridRenderCellParams) => {
+  const apiRef = useGridApiContext();
+  const rootProps = useGridRootProps();
+  const sortModel = useGridSelector(apiRef, gridSortModelSelector);
+  const treeDepth = useGridSelector(apiRef, gridRowTreeDepthSelector);
+  const editRowsState = useGridSelector(apiRef, gridEditRowsStateSelector);
+  // eslint-disable-next-line no-underscore-dangle
+  const cellValue = params.row.__reorder__ || params.row.id;
```
m4theushw, line 41: `params.row.id` won't work if another field is used as ID. You can replace it with `params.id`.

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,101 @@
...
+  classes?: DataGridProProcessedProps['classes'];
+  isDraggable: boolean;
+};
+
+const useUtilityClasses = (ownerState: OwnerState) => {
+  const { isDraggable, classes } = ownerState;
+
+  const slots = {
+    root: ['rowReorderCell', isDraggable && 'rowReorderCell--draggable'],
+    container: ['rowDraggableContainer'],
```
m4theushw, line 28: It's not very a "container". Maybe something with "placeholder", `rowReorderCellPlaceholder`?

packages/grid/x-data-grid-pro/src/colDef/gridRowReorderColDef.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const GRID_REORDER_COL_DEF: GridColDef = {
+  ...GRID_STRING_COL_DEF,
+  field: '__reorder__',
+  type: 'reorder',
+  sortable: false,
+  filterable: false,
+  width: 50,
+  align: 'center',
+  headerAlign: 'center',
+  headerName: ' ',
```
m4theushw, line 13: We need a name.

image

packages/grid/x-data-grid-pro/src/tests/reorder.DataGridPro.test.tsx
```diff
@@ -79,265 +80,363 @@ describe('<DataGridPro /> - Reorder', () => {
       await raf();
       expect(getColumnHeadersTextContent()).to.deep.equal(['brand', 'id']);
     });
```
m4theushw, line 82: No need to do now but this file could be split into one for row reordering and another for column reordering.

packages/grid/x-data-grid/src/components/containers/GridRootStyles.ts
```diff
@@ -41,6 +41,7 @@ export const GridRootStyles = styled('div', {
     { [`& .${gridClasses.menuList}`]: styles.menuList },
     { [`& .${gridClasses['row--editable']}`]: styles['row--editable'] },
     { [`& .${gridClasses['row--editing']}`]: styles['row--editing'] },
+    { [`& .${gridClasses['row--dragging']}`]: styles['row--dragging'] },
     { [`& .${gridClasses.row}`]: styles.row },
     { [`& .${gridClasses.sortIcon}`]: styles.sortIcon },
     { [`& .${gridClasses.withBorder}`]: styles.withBorder },
```
m4theushw, line 47: We need to add the overrides for `rowReorderCell` and the other classes added.

packages/grid/x-data-grid/src/utils/utils.ts
```diff
@@ -10,6 +10,10 @@ export function isObject(value: any): value is Record<string, any> {
   return typeof value === 'object';
 }
 
+export function isObjectEmpty(value: any): boolean {
+  return Object.keys(value).length === 0;
```
m4theushw, line 14: Not important, but this function is only used inside `x-data-grid-pro`. It could be moved to a utils file inside this package, or even be inlined since the check is simple.

docs/data/data-grid/rows/rows.md
```diff
@@ -135,6 +135,26 @@ To add a border instead, set `rowSpacingType` to `"border"` and customize the co
 
 You can check the [styling rows](/x/react-data-grid/style/#styling-rows) section for more information.
 
+## Row reorder [<span class="plan-pro"></span>](https://mui.com/store/items/material-ui-pro/)
+
+Row reorder is used to rearrange rows by dragging the row with the mouse.
```
m4theushw, line 140: Not only with the mouse. :)

```suggestion
Row reordering allows to rearrange rows by dragging the special reordering cell.
```

docs/data/data-grid/rows/rows.md
```diff
@@ -135,6 +135,26 @@ To add a border instead, set `rowSpacingType` to `"border"` and customize the co
...
+
+By default, `DataGridPro` does not allow row reordering. To enable reordering on all rows by dragging the special row reordering cells and moving it up or down, use the `rowReordering` prop.
```
m4theushw, line 142: We need to consider `DataGridPremium` now, so better to avoid using the component name. For code examples, there's no escape.

```suggestion
By default, row reordering is disabled.
To enable it, you need to add the `rowReordering` prop.

<DataGridPro rowReordering />
```

docs/data/data-grid/rows/rows.md
```diff
@@ -135,6 +135,26 @@ To add a border instead, set `rowSpacingType` to `"border"` and customize the co
...
+
+{{"demo": "RowOrderingGrid.js", "disableAd": true, "bg": "inline"}}
+
+To capture changes in the order of the dragged row use the `onRowOrderChange` callback.
```
m4theushw, line 146:
```suggestion
To capture changes in the order of the dragged row, you can pass a callback to the `onRowOrderChange` prop.
This callback is called with a `GridRowOrderChangeParams` object.
```

m4theushw (commented):
> I've updated the demo so if it's ok I think we can merge this and have the feature as part of this week's release.

First, do we have enough items to release this week?

Apart from that, the feature is working correctly. 🎉 I only think the docs needs some final touches. We didn't explain how to change the icon and how to set the special `__reorder__` field. Most of my suggestions for the core part are simple. We could merge this PR with the docs commented out and open another PR to refine it.

packages/grid/x-data-grid-pro/src/tests/reorder.DataGridPro.test.tsx
DanailH: I spit the 2 files ;)

packages/grid/x-data-grid/src/utils/utils.ts
DanailH: Removed the helper.

docs/data/data-grid/rows/RowOrderingGrid.tsx
```diff
@@ -0,0 +1,61 @@
...
+    dataSet: 'Commodity',
+    rowLength: 20,
+    maxColumns: 20,
+  });
+
+  const [rows, setRows] = React.useState(data.rows);
+  const [loading, setLoading] = React.useState(false);
+
+  React.useEffect(() => {
+    setRows(data.rows);
```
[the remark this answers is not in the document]
DanailH: > It would be nice to also set __reorder__. The ID is a hidden field in this dataset.

I didn't get that why do we need to set the special column?
For the other part of the feedback - done.

DanailH:
> We didn't explain how to change the icon and how to set the special `__reorder__` field. Most of my suggestions for the core part are simple. We could merge this PR with the docs commented out and open another PR to refine it.

I added a section explaining how to change the row reordering icon but the special `__reorder__` I haven't set it up in a way that you can provide your own column like this in a way that the details panel toggle column is setup. I can do it but I took the approach of the checkbox selection column. I'm not sure how much of a use case it will be for people to provide their own reorder column?
To speed up the process I can always add that functionality later on if there is demand and there aren't going to be any breaking changes.

docs/data/data-grid/rows/RowOrderingGrid.tsx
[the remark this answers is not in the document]
m4theushw: I was meaning this

msedge_h8kpx80D49

It shows the ID as placeholder, but the ID is not visible.

packages/grid/x-data-grid/src/locales/arSD.ts
```diff
@@ -131,6 +131,9 @@ const arSDGrid: Partial<GridLocaleText> = {
   // Master/detail
   expandDetailPanel: 'توسيع',
   collapseDetailPanel: 'طوي',
+
+  // Row reordering text
+  // rowReorderingHeaderName: 'Row reordering',
```
cherniavskii, line 136: Is it used somewhere?

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -0,0 +1,100 @@
...
+  );
+
+  const draggableEventHandlers = {
+    onDragStart: publish(GridEvents.rowDragStart),
+    onDragOver: publish(GridEvents.rowDragOver),
+    onDragEnd: publish(GridEvents.rowDragEnd),
+  };
+
+  return (
+    <div className={classes.root} draggable={isDraggable} {...draggableEventHandlers}>
```
cherniavskii, line 89: Side note: we should consider keyboard accessibility for this feature.
2 options come to my mind:
- minimal: `Ctrl/Cmd + arrow` keyboard shortcut to move the row up/down
- advanced: Start reorder with `Space`, move row with arrows and stop reorder with `Space`

cherniavskii (approved):

joserodolfofreitas (approved):
I think we can add the follow-up issue to the Row ordering doc page to see if it gets upvote traction.

packages/grid/x-data-grid/src/locales/arSD.ts
DanailH: yes, for in columns panel.

cherniavskii: Oh, right

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
cherniavskii: @DanailH fair enough!

docs/data/data-grid/rows/RowOrderingGrid.tsx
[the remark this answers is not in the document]
DanailH: done

fourteenmeister:
> added a section in the docs regarding further customizations https://deploy-preview-4034--material-ui-x.netlify.app/x/react-data-grid/rows/#customizing-the-row-reordering-icon (it will be visible once the build is green)

Thanks!

cherniavskii (approved):
Nice update on docs!

docs/data/data-grid/rows/rows.md
```diff
@@ -135,6 +135,77 @@ To add a border instead, set `rowSpacingType` to `"border"` and customize the co
...
+
+In addition, you can import the following events to customize the row reordering experience:
+
+- `rowDragStart`: emitted when dragging of a row starts.
+- `rowDragOver`: emitted when dragging a row over another row.
+- `rowDragEnd`: emitted when dragging of a row stops.
+
+### Customizing the reorder value
+
+By default when you start dragging a row that rows' `id` is displayed in the draggable box. To change this you can give a value to the `__reorder__` field for each row.
```
m4theushw, line 161:
```suggestion
By default, when you start dragging a row, the `id` is displayed in the draggable box.
To change this, you can give a value to the `__reorder__` field for each row.
```

docs/data/data-grid/rows/rows.md
```diff
@@ -135,6 +135,77 @@ To add a border instead, set `rowSpacingType` to `"border"` and customize the co
...
+
+```tsx
+<DataGridPro
+  components={{
+    RowReorderIcon: CustomMoveIcon,
+  }}
+/>
+```
+
+Additional customization way is to add a column with `field: __reorder__` to your set of columns. That way you can overwrite any of the properties from the `GRID_REORDER_COL_DEF` column.
```
m4theushw, line 187:
```suggestion
Another way to customize is to add a column with `field: __reorder__` to your set of columns.
That way, you can overwrite any of the properties from the `GRID_REORDER_COL_DEF` column.
```

m4theushw (approved):

docs/data/data-grid/rows/RowOrderingGrid.tsx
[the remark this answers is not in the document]
m4theushw: I still see the ID.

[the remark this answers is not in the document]
DanailH: in this demo https://deploy-preview-4034--material-ui-x.netlify.app/x/react-data-grid/rows/#row-reorder - yes. I didn't update it but I've added a section explaining how to customize it.

DanailH:
rowReorder

DanailH:
rowReorder

DanailH:
ezgif com-gif-maker

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
mbrookes: (TBH, when I tried it I expected it to drag the entire row, and make a gap for the dragged row to be dropped into, according to the mouse position, rather than the current approach of having a proxy for the row, and the row rearranging before it's been released.)

DanailH: It works similarly to the column reorder but I agree the UX can be improved. Regarding the value displayed while dragging - it can be the first column, which won't be a problem to be changed but it creates a problem when you reorder columns (or pin a column for that matter).

For tagging the entire row - I explored this as well but if there are a lot of columns it just looks ugly (virtualized columns are not rendered).

oliviertassinari:
Happy to see this feature land. A few questions on the UI/UX of the integration:

1. Why is the dragging cell not displaying the content of the first cell? I don't understand the logic behind showing the id bey default over the first visible row:

Screenshot 2022-04-29 at 23 40 42

2. From what I understand, we have two different ways to approach this problem: either move the row as dragging (e.g. AG  Grid) or only as an indicator of where the cell will be once releasing the mouse (e.g. Airtable, Notion)

Screenshot 2022-04-29 at 23 44 33

Since we went with the former, I guess to match the UX of the column reorders, would it make sense to gray it a bit the row, to help with spatial orientation?

3. When dragging beyond the viewport, the scroll is moving very fast, too fast for me to move the row to where I target.
As a side note, with column reordering, this is no longer working.
4. Are we sure about the look and feel of the drag button? I would have personally expected a small icon, maybe 18px, and to use the `theme.palette.action.active` design token like the IconButton: https://mui.com/material-ui/react-button/#icon-button.

**Before**
Screenshot 2022-04-29 at 23 55 25

**After**
Screenshot 2022-04-29 at 23 55 19

DanailH:
> Happy to see this feature land! A few questions on the UI/UX of the current implementation, we might have opportunities to push it further:
> 
> 1. Why is the dragging cell not displaying the content of the first cell? I don't understand the logic behind showing the id by default over the first visible row:
> 
> Screenshot 2022-04-29 at 23 40 42
> 
> 2. From what I understand, we have two different ways to approach this problem: either move the row as dragging (e.g. AG  Grid) or only as an indicator of where the cell will be once releasing the mouse (e.g. Airtable, Notion, Google Sheets, Excel)
> 
> Screenshot 2022-05-02 at 00 00 16
> 
> jQuery changed their implement a few months ago https://www.telerik.com/support/whats-new/kendo-jquery-ui:
> 
> Screenshot 2022-05-02 at 00 32 53
> 
> Since we went with the former, I guess to save time as it matches the logic of the column reorders (but with a worse UX, I would advise opening a new issue about it), would it make sense to gray it a bit the row, to help with spatial orientation?
> 
> 3. When dragging beyond the viewport, the scroll is moving very fast, too fast for me to move the row to where I target.
>    As a side note, with column reordering, this is no longer working.
> 4. Are we sure about the look and feel of the drag button? I would have personally expected a small icon, maybe 18px, and to use the `theme.palette.action.active` design token like the IconButton: https://mui.com/material-ui/react-button/#icon-button.
> 
> **Current** Screenshot 2022-04-29 at 23 55 25
> 
> **Proposal** Screenshot 2022-04-29 at 23 55 19

Thanks for the feedback @oliviertassinari.
Regarding your first question - @mbrookes had a similar one raised here https://github.com/mui/mui-x/pull/4034#discussion_r861147102. Basically, I chose the ID because if you reorder the first column then the content will change. That's why I've also added an example explaining how to change the value that is being displayed. Another option of the API was to mark which cell should be displayed there but I feel that the way it is currently built allows for more customizations.
Regarding showing the entire row when dragging - it will create problems when virtualized columns.

For the second question - I agree - there is room for improvement. I did it to match the same behavior we have for the columns. We can create a ticket to update both to match.
PS: I also noticed that dragging a column doesn't move the scroll left or right if it is done over the columns container

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
mbrookes: > if there are a lot of columns it just looks ugly

I noticed from the benchmark examples that only one attempts to do this, and even then is just showing the data, not the row, so the approach of using a single cell makes sense.

For the default, whichever is logically first whether pinned or dragged there would be fine. And to override, allowing the developer to specify a column by name rather than having to provide a value for each row would be far preferable, although the latter option could still be retained.

Regarding the UX, the other issue I noticed is with the row highlight. If it's a selected row that's being dragged, the blue highlight follows the row being moved, but if a row is grey (mouse-over?), it doesn't, but rather moves to seemingly random rows if you scroll past the currently displayed rows. You can see it in the recording used in the Twitter post. If needed we can jump on a screens-share so I can demonstrate.

For the visual appearance of the table when dragging, when I looked at the benchmarks, I noticed that AG Grid and one other implementation use a transition that makes it very obvious what's happening, but a common pattern is to show a line at the insertion point. Not a huge fan of that, but it's clearer than the current implementation. Having the moved row clearly indicated (as it is when selected before being moved) works too. Perhaps highlight the row being moved in blue, even if it isn't selected, then remove the highlight when the moved row is "dropped"?

I wasn't too sure about the green plus icon on the cursor, since nothing is being added (appreciate this was copied from column reordering). Perhaps "grab" and grabbing" would be better?  (Not 100% sure about that either...)

And finally, the Material DragIndicator icon is quite "heavy". Perhaps it could be rendered in a lighter shade, and/or swapped with DragHandle?

oliviertassinari:
1. Would it be possible to display the content of the first column, or to concatenate the content of the first few columns?
2. Yeah ok, makes sense it's what I assumed. Could you or @joserodolfofreitas open a new issue? I do think that we should change this behavior once we find the time (meaning get important enough, unlikely anytime soon)
3. Sounds great, a new issue would help.
4. Would this work as a quick follow-up?

mbrookes:
I had added [some further feedback](https://github.com/mui/mui-x/pull/4034#discussion_r863184696) there this evening, but hadn't seen @oliviertassinari's comments. It seems we share some of the same concerns, but I also raised some additional ones. Copied below, but see linked discussion for context.

>> if there are a lot of columns it just looks ugly
>
>I noticed from the benchmark examples that only one attempts to do this, and even then is just showing the data, not the row, so the approach of using a single cell makes sense.
>
> For the default, whichever is logically first whether pinned or dragged there would be fine. And to override, allowing the developer to specify a column by name rather than having to provide a value for each row would be far preferable, although the latter option could still be retained.
>
> Regarding the UX, the other issue I noticed is with the row highlight. If it's a selected row that's being dragged, the blue highlight follows the row being moved, but if a row is grey (mouse-over?), it doesn't, but rather moves to seemingly random rows if you scroll past the currently displayed rows. You can see it in the recording used in the Twitter post. If needed we can jump on a screens-share so I can demonstrate.
>
> For the visual appearance of the table when dragging, when I looked at the benchmarks, I noticed that AG Grid and one other implementation use a transition that makes it very obvious what's happening, but a common pattern is to show a line at the insertion point. Not a huge fan of that, but it's clearer than the current implementation. Having the moved row clearly indicated (as it is when selected before being moved) works too. Perhaps highlight the row being moved in blue, even if it isn't selected, then remove the highlight when the moved row is "dropped"?
>
> I wasn't too sure about the green plus icon on the cursor, since nothing is being added (appreciate this was copied from column reordering). Perhaps "grab" and grabbing" would be better? (Not 100% sure about that either...)
>
> And finally, the Material DragIndicator icon is quite "heavy". Perhaps it could be rendered in a lighter shade, and/or swapped with DragHandle?

DanailH:
Thanks for the feedback. I'll organise it into a separate issue today

DanailH:
Here is the follow up issue https://github.com/mui/mui-x/issues/4754

oussamachah2020:
hello, m having an issue here.
image

after adding a new record, the drag button become disabled even if the table is on mode view. i need to reload the page to get it to work

any solution please i need asap 

here is my code 

```js
import { DataGridPro, GridRowOrderChangeParams } from "@mui/x-data-grid-pro";
import {
  createNewEtape,
  fetchEtapes,
  getGreatestRank,
  updateEtapeByDragAndDrop,
} from "../Loaders/etape";
import { client } from "../lib/appwrite";
import {
  GridColDef,
  GridEventListener,
  GridRowEditStopReasons,
  GridRowId,
  GridRowModes,
  GridRowModesModel,
  GridRowsProp,
} from "@mui/x-data-grid";
import * as moment from "moment-timezone";
import { Etape } from "../types/etapeType";
// import SaveIcon from "@mui/icons-material/Save";
// import CancelIcon from "@mui/icons-material/Close";
import { Autocomplete, Button, TextField } from "@mui/material";
import { randomId } from "@mui/x-data-grid-generator";
import AddIcon from "@mui/icons-material/Add";
import ChoiceInput from "./ChoiceInput";
import { db } from "../../firebase";
import { collection, onSnapshot, query } from "firebase/firestore";

function updateRowPosition(
  initialIndex: number,
  newIndex: number,
  rows: Array<Etape>
) {
  console.log(initialIndex, newIndex);

  const rowsClone = [...rows];
  const row = rowsClone.splice(initialIndex, 1)[0];
  rowsClone.splice(newIndex, 0, row);

  return rowsClone;
}

interface EditToolbarProps {
  onClick: () => void;
}

function EditToolbar(props: EditToolbarProps) {
  return (
    <Button color="primary" startIcon={<AddIcon />} onClick={props.onClick}>
      Add record
    </Button>
  );
}

export default function RowOrderingGrid() {
  const [rows, setRows] = React.useState<Etape[]>([]);
  const [loading, setLoading] = React.useState(false);
  const [rowModesModel, setRowModesModel] = React.useState<GridRowModesModel>(
    {}
  );

  const handleRowModesModelChange = (newRowModesModel: GridRowModesModel) => {
    setRowModesModel(newRowModesModel);
  };

  const handleRowEditStop: GridEventListener<"rowEditStop"> = (
    params,
    event
  ) => {
    if (params.reason === GridRowEditStopReasons.rowFocusOut) {
      event.defaultMuiPrevented = true;
    }
  };

  const handleSaveClick = (id: string) => () => {
    setRowModesModel((prev) => ({
      ...prev,
      [id]: { mode: GridRowModes.View },
    }));
  };

  const handleRankChange = async () => {
    let new_rank: number;

    const data = await getGreatestRank();

    if (data.length == 0) {
      new_rank = 1;
    } else {
      new_rank = data[0].rank + 1;
    }

    return new_rank;
  };

  const AddNewRecord = () => {
    const id = randomId();
    const now = moment().tz("America/New_York");

    setRows((oldRows) => [
      ...oldRows,
      {
        id,
        titre: "",
        id_code: "",
        date_creation: now.format(),
        derniere_modification: now.format(),
        choix: [],
        rank: 0,
        isNew: true,
      },
    ]);
    setRowModesModel((oldModel) => ({
      ...oldModel,
      [id]: { mode: GridRowModes.Edit, fieldToFocus: "titre" },
    }));
  };

  const createEtape = async (updatedRow: Etape) => {
    setLoading(true);
    const new_rank_value = await handleRankChange();

    createNewEtape(updatedRow.titre, updatedRow.id_code, new_rank_value)
      .then(() => {
        console.log("ok");
        console.log(rowModesModel[updatedRow.id]?.mode);
      })
      .catch((err) => console.error(err))
      .finally(() => {
        setLoading(false);
      });
  };

  function loadFromBackend() {
    fetchEtapes()
      .then((res: Etape[]) => {
        setRows(res.sort((a, b) => a.rank - b.rank));
      })
      .catch((err) => {
        console.error(err);
      });
  }

  React.useEffect(() => {
    loadFromBackend();
  }, []);

  React.useEffect(() => {
    const q = query(collection(db, "etape"));
    const unsubscribe = onSnapshot(q, (querySnapshot) => {
      querySnapshot.forEach((doc) => {
        if (doc.exists()) {
          loadFromBackend();
        }
      });
    });

    return () => {
      unsubscribe();
    };
  }, []);

  const handleRowOrderChange = async (params: GridRowOrderChangeParams) => {
    setLoading(true);
    const { oldIndex, targetIndex } = params;

    const newRows = updateRowPosition(oldIndex, targetIndex, rows);

    for (let i = 0; i <= newRows.length - 1; i++) {
      try {
        await updateEtapeByDragAndDrop({ ...newRows[i], rank: i + 1 });
      } catch (error) {
        console.error(error);
      }
    }

    setLoading(false);
  };

  const columns: GridColDef<any>[] = [
    { field: "titre", headerName: "Titre", flex: 1, editable: true },
    {
      field: "date_creation",
      headerName: "Date de création",
      renderCell(params) {
        return (
          <p>{`${moment(params.value).format("YYYY-MM-DD")} - ${moment(
            params.value
          ).format("HH:mm:ss")}`}</p>
        );
      },
    },
    {
      field: "derniere_modification",
      headerName: "Dérniere modification",
      flex: 1,
      renderCell(params) {
        return (
          <p>{`${moment(params.value).format("YYYY-MM-DD")} - ${moment(
            params.value
          ).format("HH:mm:ss")}`}</p>
        );
      },
    },
    {
      field: "id_code",
      headerName: "ID Code",
      flex: 1,
      editable: true,
    },
    // {
    //   field: "choix",
    //   headerName: "Choix",
    //   flex: 1,
    //   renderCell(params) {
    //     return <ChoiceInput docId={params.row.id} />;
    //   },
    // },
  ];

  return (
    <div style={{ height: 500, width: "100%" }}>
      <DataGridPro
        loading={loading}
        rows={rows}
        columns={columns}
        rowReordering
        editMode="row"
        rowModesModel={rowModesModel}
        onRowModesModelChange={handleRowModesModelChange}
        onRowEditStop={handleRowEditStop}
        getRowId={(row) => row.id}
        pagination
        pageSizeOptions={[5, 10]}
        processRowUpdate={(updatedRow, originalRow) => {
          createEtape(updatedRow);
          // handleSaveClick(originalRow.id);
        }}
        onRowOrderChange={handleRowOrderChange}
        slots={{
          toolbar: EditToolbar,
        }}
        slotProps={{
          toolbar: { onClick: () => AddNewRecord() },
        }}
      />
    </div>
  );
}
```

PR #4036, opened 2022-02-26
[core] Initialize remaining states before feature hooks

m4theushw:
In #3896 I left some states being initialized inside their feature hook. This PR changes to have all states initialized before the feature hooks. Also removes `useGridStateInit`.


packages/grid/x-data-grid-pro/src/useDataGridProComponent.tsx
```diff
@@ -74,6 +90,16 @@ export const useDataGridProComponent = (
   /**
    * Register all state initializers here.
    */
+  useGridInitializeState(columnMenuStateInitializer, apiRef, props);
```
flaviendelangle, line 93: Could we keep the same order as the feature hook themselves ?
In 90% of the case, the order don't matter here.
But if one day we want to launch plugins and therefore loop on a list of objects to call those `useGridInitializeState` then the main hook, it will be easier if the order is always the same.

packages/grid/x-data-grid/src/internals/hooks/features/selection/useGridSelection.ts
```diff
@@ -45,19 +66,10 @@ export const useGridSelection = (
...
-
-    if (Array.isArray(props.selectionModel)) {
-      return props.selectionModel;
-    }
-
-    return [props.selectionModel];
-  }, [props.selectionModel]);
-
-  useGridStateInit(apiRef, (state) => ({ ...state, selection: propSelectionModel ?? [] }));
+  const propSelectionModel = React.useMemo(
```
flaviendelangle, line 69: Small behavior change here that I think we can accept
If a user control and gives a non-array value, before I was not doing a new `setState` in the `useEffect` tracking the model.
Now it does because the array created in `getSelectionModelPropValue` called to initialize the state won't have the same reference and the array called to create `propSelectionModel`.

flaviendelangle (approved):

PR #4055, opened 2022-02-28
Fix license generating script

Janpot:
- The CommonJs build output seems to be at the root of the build folder now.
- Updated CLI params aliases to comply with the error message `"MUI: You forgot to pass an order number. $ > licensegen -o order_123."` https://github.com/mui/mui-x/blob/36a9e9ed388438b1df43ccfa26637700f9835e0d/packages/x-license-pro/src/license-cli.ts#L32


flaviendelangle:
@Janpot #3965 changes the path so we will have to test after merging them.

Janpot:
@flaviendelangle https://github.com/mui/mui-x/pull/3965 fixed the import path, so all that remains now is the CLI alias

flaviendelangle (approved):

PR #4057, opened 2022-02-28
[DataGridPro] Disable export for detail panel column

gustavhagland:
Currently, export a table with detailed panels exports a column with name: 
`__detail_panel_toggle__`and only values: ` true` or `false` indicating the state.

These are not usefull values and should be disabled by default


m4theushw (approved):
Thanks!

PR #4066, opened 2022-03-01
[docs] Fix typo in Data Grid/Editing/Client-side validation

krallj:
In the documentation for [Data Grid/Editing/Client-side validation](https://mui.com/components/data-grid/editing/#client-side-validation), `params` is incorrectly typed. This PR makes the example consistent with the `preProcessEditCellProps` property of [GridColDef](https://mui.com/api/data-grid/grid-col-def/#properties).


alexfauquette (approved):
Thanks for the fix :)

It let m realise that the demo also need some cleaning

m4theushw (approved):
Thanks.

PR #4069, opened 2022-03-02
[DataGrid] Add Filter Panel delete icon slot

### LINKED ISSUE #4063: [DataGrid] Add a slot to modify delete icon of filter panel
### Duplicates

### Latest version

### Summary 💡

The [GridIconSlotsComponent](https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/models/gridIconSlotsComponent.ts#L24:L24) does not provide a slot for the close icon used in the `<FilterForm />` component.

Should allows to override this icon as we do for others

### Examples 🌈

_No response_

### Motivation 🔦

Make all the icons are customizable

### Order ID 💳 (optional)

_No response_

Hameezr:
Fix #4063

### DIFF
--- a/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
@@ -12,7 +12,6 @@ import { gridFilterableColumnDefinitionsSelector } from '../../../hooks/features
 import { useGridSelector } from '../../../hooks/utils/useGridSelector';
 import { GridFilterItem, GridLinkOperator } from '../../../models/gridFilterItem';
 import { useGridApiContext } from '../../../hooks/utils/useGridApiContext';
-import { GridCloseIcon } from '../../icons';
 import { GridTranslationKeys } from '../../../models/api/gridLocaleTextApi';
 import { useGridRootProps } from '../../../hooks/utils/useGridRootProps';
 import { DataGridProcessedProps } from '../../../models/props/DataGridProps';
@@ -287,9 +286,7 @@ function GridFilterForm(props: GridFilterFormProps) {
           onClick={handleDeleteFilter}
           size="small"
         >
-          {rootProps.components.FilterPanelDeleteIcon?.name ? 
-          <rootProps.components.FilterPanelDeleteIcon />
-          : <GridCloseIcon fontSize="small" />}
+          <rootProps.components.FilterPanelDeleteIcon fontSize="small" />
         </IconButton>
       </FilterFormDeleteIcon>
       <FilterFormLinkOperatorInput
--- a/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
@@ -287,7 +287,7 @@ function GridFilterForm(props: GridFilterFormProps) {
           onClick={handleDeleteFilter}
           size="small"
         >
-          <GridCloseIcon fontSize="small" />
+          <rootProps.components.FilterPanelDeleteIcon/>
         </IconButton>
       </FilterFormDeleteIcon>
       <FilterFormLinkOperatorInput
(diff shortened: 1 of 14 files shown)


packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -287,7 +286,7 @@ function GridFilterForm(props: GridFilterFormProps) {
           onClick={handleDeleteFilter}
           size="small"
         >
-          <GridCloseIcon fontSize="small" />
```
alexfauquette, line 290: Removing the `fontSize="small"` prop modify the default appearance of the icon

## New icon
image
## Previouse icon
image

packages/grid/x-data-grid/src/models/gridIconSlotsComponent.ts
```diff
@@ -110,4 +110,9 @@ export interface GridIconSlotsComponent {
    * @default GridRemoveIcon
    */
   DetailPanelCollapseIcon: React.JSXElementConstructor<any>;
+  /**
+   * Icon displayed for deleting the filter from filter Panel.
+   * @default GridCloseIcon
+   */
+  FilterPanelDeleteIcon: React.JSXElementConstructor<any>;
```
DanailH, line 117: We should also add a new entry in the components Prop

alexfauquette: I'm not sure to understand. For now, icons does not appears in `componentsProps`, only in `components`

And the type of `components` extends the icons one 
https://github.com/alexfauquette/material-ui-x/blob/master/packages/grid/x-data-grid/src/internals/models/gridSlotsComponent.ts#L8:L8

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
Hameezr: I've added a conditional logic to fix this issue but I believe that's not the best approach, since icons are used directly from svg's and there could be a component to render the icons, used globally with props such as size and etc.

Hameezr: I have a suggested approach for using the icons more efficiently in my mind, but that'll require some time since they're being used in a lot of places. I'll create a separate PR for that once I'm able to figure out the best approact

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -286,7 +287,9 @@ function GridFilterForm(props: GridFilterFormProps) {
           onClick={handleDeleteFilter}
           size="small"
         >
+          {rootProps.components.FilterPanelDeleteIcon?.name ? 
           <rootProps.components.FilterPanelDeleteIcon />
+          : <GridCloseIcon fontSize="small" />}
```
alexfauquette: Why using `rootProps.components.FilterPanelDeleteIcon?.name` as a criterion? Any icon created by `createSvgIcon()` will have a name.

A simpler solution would be to set `<rootProps.components.FilterPanelDeleteIcon fontSize="small" />`

The drawback is that all the icons will receive the prop `fontSize="small"` with no way to change it.

If developers want to use big icons, they can ignore the prop as follow:
```jsx
const MyBigIcon = () => <CloseIcon fontSize="large" />

<DataGrid components={{ FilterPanelDeleteIcon: MyBigIcon }} />
```

@Hameezr  Does that sound good to you?

alexfauquette (commented):
I merged `master` such that we can check the icon is still the same (I added a test to catch this regression)

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
Hameezr: Yes, this will solve the problem and add some customizability as well! Thanks for the suggestion, will make this change

alexfauquette (approved):
Thanks for your contribution, this feature will be available in next release :)

PR #4073, opened 2022-03-02
[docs] Clean ValidateRowModelControlGrid demo

alexfauquette:
- Add type to `preProcessEditCellProps`
- define `columns`/`rows` before using them

### DIFF
--- a/docs/data/data-grid/editing/ValidateRowModelControlGrid.tsx
+++ b/docs/data/data-grid/editing/ValidateRowModelControlGrid.tsx
@@ -1,12 +1,13 @@
 import * as React from 'react';
 import Box from '@mui/material/Box';
-import { DataGrid, GridColumns, GridRowsProp } from '@mui/x-data-grid';
 import {
-  randomCreatedDate,
-  randomEmail,
-  randomTraderName,
-  randomUpdatedDate,
-} from '@mui/x-data-grid-generator';
+  DataGrid,
+  GridColumns,
+  GridPreProcessEditCellProps,
+} from '@mui/x-data-grid';
+import { useDemoData } from '@mui/x-data-grid-generator';
+
+const VISIBLE_FIELDS = ['name', 'email', 'dateCreated', 'lastLogin'];
 
 function validateEmail(email) {
   const re =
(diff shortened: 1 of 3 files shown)


docs/data/data-grid/editing/ValidateRowModelControlGrid.tsx
```diff
@@ -15,6 +16,39 @@ function validateEmail(email) {
...
+    },
+    {
+      field: 'dateCreated',
+      headerName: 'Date Created',
+      type: 'date',
+      width: 180,
+      editable: true,
+    },
+    {
+      field: 'lastLogin',
```
m4theushw, line 44: The employee dataset doesn't this column. You can use `lastUpdated` to fake it.

```suggestion
      field: 'lastUpdated',
```

alexfauquette:
Do you have an idea of why numbers appear in argos whereas they are not visible in the documentation example?

docs/data/data-grid/editing/ValidateRowModelControlGrid.tsx
```diff
@@ -1,12 +1,13 @@
...
-  randomTraderName,
-  randomUpdatedDate,
-} from '@mui/x-data-grid-generator';
+  DataGrid,
+  GridColumns,
+  GridPreProcessEditCellProps,
+} from '@mui/x-data-grid';
+import { useDemoData } from '@mui/x-data-grid-generator';
+
+const VISIBLE_FIELDS = ['name', 'email', 'dateCreated', 'lastLogin'];
```
m4theushw, line 10: The last column is not visible.

```suggestion
const VISIBLE_FIELDS = ['name', 'email', 'dateCreated', 'lastUpdated'];
```

m4theushw (approved):
> Do you have an idea of why numbers appear in argos whereas they are not visible in the documentation example?

The name column adds a suffix if the name is already being used. You don't see it in the docs example because there we use random names.

https://github.com/mui/mui-x/blob/9e2c4b43436b970d269db2b623dbe2763182f601/packages/grid/x-data-grid-generator/src/columns/employees.columns.tsx#L50-L51

docs/data/data-grid/editing/ValidateRowModelControlGrid.tsx
alexfauquette: I realized I was looking at the wrong example in the doc 🙈

PR #4075, opened 2022-03-02
[test] Only test custom input keyboard event in edit mode

alexfauquette:
Follow up of https://github.com/mui/mui-x/pull/3624#discussion_r817668116


packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.test.tsx
```diff
@@ -591,6 +591,34 @@ describe('<DataGridPro /> - Cell Editing', () => {
     expect(cell).to.have.class('MuiDataGrid-cell--editing');
   });
 
+  it('should be able to type in a custom input', () => {
```
m4theushw, line 594: To be more precise about what the test is doing.

```suggestion
  it('should not call .preventDefault on key events inside a cell in edit mode', () => {
```

m4theushw (approved):

PR #4082, opened 2022-03-03
[core] Make is clearer this is only for questions

oliviertassinari:
After reading https://groups.google.com/a/mui.com/g/custom-work/c/5RRqH0YkZ6o/m/aWBVaxkrBAAJ?utm_medium=email&utm_source=footer, I recalled the point that @cherniavskii did during our last retrospective that this issue template is only meant for cases where the problem is not a bug nor a feature request. But the current description is not very clear.

To be clear, this PR is modifying https://github.com/mui/mui-x/issues/new/choose


joserodolfofreitas (approved):
At first, I thought about not releasing the premium template yet. But then, why not?
It's looking good like this. And clearly framing the templates as questions is a great idea.

PR #4086, opened 2022-03-04
[DataGridPro] Allows to control the theme

### LINKED ISSUE #3378: [DataGrid] Print style is varying between light and dark mode
### Duplicates

### Latest version

### Current behavior 😯

The colors contrast is less important if the print is done from dark mode than light mode

image

image

### Expected behavior 🤔

The print should be the same in dark and light

### Steps to reproduce 🕹

Steps:

1. go to https://mui.com/components/data-grid/demo/
2. print the DataGrid
3. change light/dark mode
4. print the DataGrid
5. Notice the two preview are different

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
To fix #3378 I propose to take control on the theme, Such that the DataGrid could switch between light and dark theme

Not sure it's useful or overkill compared to the problem. But it would allow switching to the light theme when printing, and come back to the default one after

### DIFF
--- a/docs/data/data-grid/style/ToggleColorMode.js
+++ b/docs/data/data-grid/style/ToggleColorMode.js
@@ -27,7 +27,7 @@ export default function ToogleColorMode() {
         exclusive
         onChange={(event, newValue) => {
           setPaletteMode(newValue);
-          apiRef.current.setThemePalette(newValue);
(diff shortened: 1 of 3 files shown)


flaviendelangle:
Maybe the core team can give us some advise here

PR #4089, opened 2022-03-04
[DataGrid] Deprecate and stop typing the api params of `GridCellParams` / `GridValueGetterParams` and affiliated

flaviendelangle:


packages/grid/x-data-grid/src/models/params/gridCellParams.ts
```diff
@@ -71,51 +64,41 @@ export interface GridCellParams<
...
-export interface GridRenderCellParams<
-  V = any,
-  R = any,
-  F = V,
-  Api extends GridApiCommon = GridApiCommunity,
-> extends GridCellParams<V, R, F, Api> {
+export interface GridRenderCellParams<V = any, R = any, F = V> extends GridCellParams<V, R, F> {
   /**
    * GridApi that let you manipulate the grid.
+   * @deprecated Use the `apiRef` returned by `useGridApiContext` or `useGridApiRef`
```
m4theushw, line 70: Community users don't have access to `useGridApiRef`.

```suggestion
   * @deprecated Use the `apiRef` returned by `useGridApiContext` or `useGridApiRef` (only available in DataGridPro).
```

x other params.

m4theushw (approved):
I would initially only set the `Api` argument to have `any` as default value, but no problem not making it generic anymore and using `any` directly.

packages/grid/x-data-grid/src/models/params/gridCellParams.ts
flaviendelangle: I think that if we set the default value as `any`, 99% of our users will never use it.
And in that case, we better get rid of the whole complexity.

Fixed for the comment.

We have a few use-case where removing the `api` param breaks internal code
For instance in `GRID_CHECKBOX_SELECTION_COL_DEF`:

```ts
  valueGetter: (params) => {
    const selectionLookup = selectedIdsLookupSelector(
      (params as GridValueGetterFullParams).api.state,
      (params as GridValueGetterFullParams).api.instanceId,
    );
    return selectionLookup[params.id] !== undefined;
  },
```

Maybe we will have to pass the `state` and `instanceId` as a param but it would re-introduce generics to type correctly.
But let's take the time to discuss it before v6, the deprecation is for the long term.

PR #4111, opened 2022-03-07
[core] Prepare the api build scripts for multi packages support

flaviendelangle:
- [x] Allow to write doc files in other folder than `/data-grid` (for instance `/date-pickers`)
- [x] Improve build perfs by using the `tsconfig.build.json` of the package instead of the root `tsconfig.json` (allow to take advantage of the `references` file and do not compile test files)
- [x] Move the logic to decide the components we want to generate the doc and the proptypes from into `getTypeScriptProject`, the logic differ between the grid and the pickers and letting that in there current file was hard to read
- [x] Rename grid-only build processes `buildGrid...`  for clarity
- [x] Generate export json for `x-license-pro` and `x-data-grid-generator`


alexfauquette (commented):
I still have to review more carefully the `getTypeScriptProjects.ts` which seems to contain most of the logic

docs/scripts/getTypeScriptProjects.ts
```diff
@@ -69,29 +106,58 @@ const createProject = (options: CreateProgramOptions): Project => {
...
-    componentsFolder: componentsFolder ? path.join(rootPath, componentsFolder) : undefined,
-    otherComponentFiles: otherComponentFiles?.map((file) => path.join(rootPath, file)),
     prettierConfigPath: path.join(workspaceRoot, 'prettier.config.js'),
   };
 };
 
+/**
+ * Transforms a list of folders and files into a list of file paths containing components.
+ * The file must have the name of the component.
+ * @param {string} folders The folders from which we want to extract components
```
alexfauquette, line 122: same bellow
```suggestion
 * @param {string[]} folders The folders from which we want to extract components
```

docs/scripts/getTypeScriptProjects.ts
```diff
@@ -100,12 +166,71 @@ export const getTypeScriptProjects = () => {
...
+  );
+
+  projects.set(
+    'x-data-grid-generator',
+    createProject({
+      name: 'x-data-grid-generator',
+      rootPath: path.join(workspaceRoot, 'packages/grid/x-data-grid-generator'),
+      documentationFolderName: 'data-grid',
+    }),
+  );
```
alexfauquette, line 203: Why adding the data grid generator to the documented projects?

docs/scripts/api/buildInterfacesDocumentation.ts
```diff
@@ -236,7 +234,8 @@ function generateImportStatement(objects: ParsedObject[], projects: Projects) {
         project.name
       }'`;
     })
-    .filter((el): el is string => !!el);
+    .filter((el): el is string => !!el)
+    .sort((a, b) => b.length - a.length);
```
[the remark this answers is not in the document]
alexfauquette: I think you should add a comment to explain it, otherwise nobody will be able to understand why this line has been added

docs/scripts/getTypeScriptProjects.ts
flaviendelangle: To generate the export JSON
It is not related to the pickers at all.
But after a few months with those `package.exports.json`, I use them a lot to now what I start / stop exporting.

alexfauquette: I was thinking that we don't want to document this package. But yes, it does not hurt to know what is modified in the export

flaviendelangle: Adding it to the project has no added any new content except the export file.
In the future, we could add keys like `documentSelectors: false` if the package exported selectors and we did not want to document them for instance.

alexfauquette (approved):

Issue #3969, opened 2022-02-16
[license] Side effect issue: The license key is not installed in production


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Considering the following configuration:

```jsx
import {} from "./mui"

ReactDOM.render(
    <PreviewComponent>
        <ModalProvider>
            <App />
        </ModalProvider>
    </PreviewComponent>,
    document.getElementById("root")
)
```

**mui.js**

```jsx
import { LicenseInfo } from "@mui/x-data-grid-pro"

LicenseInfo.setLicenseKey("fooo")
```

https://mui.zendesk.com/agent/tickets/2356

When building for production the license key is not installed. 

### Expected behavior 🤔

When building for production, the license key is installed. 

### Context 🔦

We set `sideEffect: false` in https://unpkg.com/browse/@mui/x-data-grid-pro@5.5.1/package.json so in our example, the bundler assumes that that import is dead code, it removes it all when building for production. It's a nasty bug as it works in dev mode.

Now, this side effect false is not true, there is a side effect. I'm not sure how to fix this problem. 

Maybe it would make sense to have `@mui/x-license-pro` as a user facing dependency. @flaviendelangle how do you plan to solving the license key problem with the date picker?

### Your environment 🌎

v5.5.1

flaviendelangle:
Nasty one indeed...
Do you have a reproduction case ? therwise I can try to create one.

I did not work on the licensing of the `DateRangePicker` yet, I'm fully on the bundling rework right now.
But I feel like `@mui/x-license-pro` should be a user facing depency. It may be strange for user to just import `LicenseInfo` from one of our product packages when using several one of them.

oliviertassinari:
> Do you have a reproduction case?

I don't. I have opened this issue to bring visibility to the three shaking issue. I think that we could snooze this side effect issue for some time, wait to see more developers struggling to install the license.

cherniavskii:
As of now, we recommend using the `@mui/x-license-pro` package directly: https://mui.com/x/introduction/licensing/#how-to-install-the-key
Would it make sense to set `sideEffects: true` for the `@mui/x-license-pro` package?

oliviertassinari:
> Would it make sense to set sideEffects: true for the @mui/x-license-pro package?

@cherniavskii I come to this conclusion by reading https://webpack.js.org/guides/tree-shaking/. However, maybe we can be more specific:

[...]

Issue #1364, opened 2021-04-05
[License] Improve the error message when using the key from the docs


oliviertassinari:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Developers might not realize that they fail to provide the correct license key

## Examples 🌈

1. Open https://material-ui.com/components/data-grid/getting-started/#license-key-installation
2. Copy the line
3. Forget to replace with the correct key
4. The error message could be very specific

## Motivation 🔦

- Improve the DX
- Reduce the number of inbound support requests

## Order id 💳

23247

oliviertassinari:
I have a couple of more ideas to improve the DX with the license:

1. In the console error, when we log `showInvalidLicenseError` we could plot the license key used. It could help developers that are using env variable realize they used the wrong key.
2. In the console error, when we log `showExpiredLicenseError` we could plot the expiration date of their license, as well as the package+version+date they tried to use that is no longer valid. Without such information, it can be a which hunt for developers to figure out how to solve the problem 🧙‍♀️🙃.

oliviertassinari:
#4315 introduced the notion of scope for the license key, that #4651 put into use. Now, these efforts increase the surface area to this GitHub issue. Now, we can also better warn about wrong scope usage. We currently say "invalid" when using Premium features with a valid Pro plan license.

https://github.com/flaviendelangle/material-ui-x/blob/336bd216f6c6d960025e91e94f237039cdd232cc/packages/x-license-pro/src/verifyLicense/verifyLicense.test.ts#L115-L125

which can be hard to debug for developers.

PR #4123, opened 2022-03-08
[core] Prepare `@mui/x-license-pro` for date pickers

flaviendelangle:
- [x] Move the release info from `@mui/x-data-grid-pro` to `@mui/x-license-pro`. It can be problematic if user keep an outdated version of `@mui/x-license-pro` and update `@mui/x-data-grid-pro` or `@mui/x-date-pickers-pro` to try fooling the check. 

The alternative would be to store one release info per pro package.
Something like:

```ts
// DataGridPro.tsx
let RELEASE_INFO = '__RELEASE_INFO__';

// eslint-disable-next-line no-useless-concat
if (process.env.NODE_ENV !== 'production' && RELEASE_INFO === '__RELEASE' + '_INFO__') {
  // eslint-disable-next-line no-underscore-dangle
  RELEASE_INFO = ponyfillGlobal.__MUI_RELEASE_INFO__;
}

LicenseInfo.setReleaseInfo('x-data-grid-pro', RELEASE_INFO);
```

- [x] Move the `Watermark` component in `@mui/x-license-pro` to use it in `@mui/x-date-pickers`
- [x] Do the license check in the render method of `DataGridPro` (or `DateRangePicker` and affiliated) instead of in `Watermark`
- [x] Only check once the license even when calling several time `useLicenseVerifier` (on `DateRangePicker`, we show the watermak deep below the top component, so it would be a pain to drill the prop down to it.


packages/x-license-pro/src/utils/licenseInfo.ts
```diff
@@ -1,29 +1,41 @@
...
+  // eslint-disable-next-line no-underscore-dangle
+  RELEASE_INFO = ponyfillGlobal.__MUI_RELEASE_INFO__;
+}
+
+// eslint-disable-next-line no-underscore-dangle
+ponyfillGlobal.__MUI_LICENSE_INFO__.releaseInfo = RELEASE_INFO;
+
 export class LicenseInfo {
   public static getKey(): string {
+    // eslint-disable-next-line no-underscore-dangle
```
flaviendelangle, line 28: I moved the lint disable on each line otherwise I don't now why, eslint said that the directive was not used, but removing it was causing errors...

packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
```diff
@@ -8,16 +8,26 @@ import {
 } from '../utils/licenseErrorMessageUtils';
 import { LicenseStatus } from '../utils/licenseStatus';
 
+let sharedLicenseStatus: LicenseStatus | undefined;
+
 export function useLicenseVerifier(): LicenseStatus {
   return React.useMemo(() => {
-    const newLicenseStatus = verifyLicense(LicenseInfo.getReleaseInfo(), LicenseInfo.getKey());
-    if (newLicenseStatus === LicenseStatus.Invalid) {
+    if (sharedLicenseStatus !== undefined) {
```
m4theushw, line 15: Beyond checking if there's a status saved you could also check if the current `LicenseInfo.getKey()` matches the one used when `sharedLicenseStatus` was set. This prevents the user from changing the key from a valid to an invalid and don't see the watermark.

flaviendelangle: Done

packages/grid/x-data-grid-pro/src/DataGridPro/DataGridPro.tsx
```diff
@@ -11,30 +11,18 @@ import {
...
-
-LicenseInfo.setReleaseInfo(RELEASE_INFO);
-
 const DataGridProRaw = React.forwardRef<HTMLDivElement, DataGridProProps>(function DataGridPro(
   inProps,
   ref,
 ) {
   const props = useDataGridProProps(inProps);
   const apiRef = useDataGridProComponent(props.apiRef, props);
+  useLicenseVerifier();
```
oliviertassinari, line 25: What's the motivation for not having Watermark do it?

A related matter, how do we display a watermark with the date range picker? maybe on the background of the page? 🤔

oliviertassinari (commented):
Thoughts:

1. I think that the "-pro" in `@mui/x-license-pro` doesn't really make sense anymore since we are not moving in the direction where "Pro" is the term that encompasses paid plans. We double down on Pro being **one** plan. I would propose we rename it: `@mui/x-license-commercial`. In the past, we were calling it:
https://github.com/mui/mui-x/blob/5b7689187789ccc5ed539f969da6ce33c3e25222/packages/x-license/package.json#L2
1. I think that the "release info" should be owned by each package, not by the license package manager. First, the primary intent of this is to guarantee that developers never use IP they are not licensed to use (@flaviendelangle shown a case where this would fail). Second, we are supposed to rarely release a new version of `@mui/x-license-pro`, it's going to be true with #4202.
In practice, I think that when verifying the license, we could provide the "release info" as an argument for the call.
1. I would suggest that we make the `@mui/x-license-pro` package public. We could list the need to install it in https://mui.com/components/data-grid/getting-started/#installation. In v5, we would need to do this gracefully and in v6 we would make this mandatory. The main value would be about having a single infrastructure to manage the license. Not 10 different ways to install the license, not one API on each components. It would be an opportunity to remove https://github.com/mui/mui-x/blob/140f39480052c1ca41561eea6fcd28c3a07e1da9/packages/x-license-pro/package.json#L12 which is not true and fix #3969.

oliviertassinari:
Strange, I see no backlink in #4202 and #3969, adding this comment to force it.

packages/x-license-pro/src/utils/licenseInfo.ts
```diff
@@ -1,29 +1,41 @@
...
 // decide to duplicate to improve the size of the chunks.
+// eslint-disable-next-line no-underscore-dangle
 ponyfillGlobal.__MUI_LICENSE_INFO__ = ponyfillGlobal.__MUI_LICENSE_INFO__ || {
   key: undefined as undefined | string,
   releaseInfo: undefined as undefined | string,
 };
 
+// This is the package release date. Each package version should update this const
+// automatically when a new version is published on npm.
+let RELEASE_INFO = '__RELEASE_INFO__';
```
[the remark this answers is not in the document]
flaviendelangle: Yes

packages/grid/x-data-grid-pro/src/DataGridPro/DataGridPro.tsx
flaviendelangle: When doing the headless version, we need to check the license even if no watermark is called

flaviendelangle:
@oliviertassinari 

1. I agree, but I would keep the topic around the date pickers and the premium plan distinct. The Date Pickers do not need this rename yet.

flaviendelangle:
> Not 10 different APIs to install the license, one per component. It would be an opportunity to remove

But right now it is not the case right ?
Users can import `setReleaseInfo` from `@mui/x-data-grid-pro` or from `@mui/x-date-pickers-pro` or from `@mui/x-license-pro`, they are the same and activate the license for every packages.

packages/grid/x-data-grid-pro/src/DataGridPro/DataGridPro.tsx
flaviendelangle: image

packages/grid/x-data-grid-pro/src/utils/releaseInfo.ts
```diff
@@ -0,0 +1,15 @@
+import { ponyfillGlobal } from '@mui/utils';
+
+export const getReleaseInfo = () => {
```
flaviendelangle, line 3: This utility function is not very usefull for the grid because there is currently only one call
But for the pickers, we have several commercial components so it's better to have a centralized place.
And keeping things similar in all packages is cleaner I think.

This code can't be in `x-license-pro` because the `string-replace-loader` build step must be done with the date of the build of each package.

packages/x-license-pro/src/utils/licenseInfo.ts
```diff
@@ -1,29 +1,30 @@
...
-    return ponyfillGlobal.__MUI_LICENSE_INFO__.releaseInfo;
+  public static getLicenseKey(): MuiLicenseInfo['key'] {
+    return LicenseInfo.getLicenseInfo().key;
   }
 
   public static setLicenseKey(key: string) {
-    ponyfillGlobal.__MUI_LICENSE_INFO__.key = key;
-  }
-
-  public static setReleaseInfo(encodedReleaseInfo: string) {
```
flaviendelangle, line 26: The release info are not handled on each package instead of centralized here

packages/grid/x-data-grid-pro/src/DataGridPro/DataGridPro.tsx
flaviendelangle: For the check, I memoize it so calling `useLicenseVerifier` a lot of time is not a big problem I think.
For the date pickers I call it in `DateRangePicker`, in `MobileDateRangePicker`, in `DesktopDateRangePicker`, in `StaticDateRangePicker` etc... basically in each pro component exported. 
That removes the struggle to think whether or not the check is done in every scenario. The hook is called several time, the 1st time it warns, then it just returns the memoized status and does nothing more.

joserodolfofreitas:
> 1. I think that the "-pro" in `@mui/x-license-pro` doesn't really make sense anymore since we are not moving in the direction where "Pro" is the term that encompasses paid plans. We double down on Pro being **one** plan. I would propose we rename it: `@mui/x-license-commercial` cc @joserodolfofreitas for thoughts. In the past, we were calling it:

I agree we should rename it, but perhaps focus the name on the role of the package.
What do you think about `@mui/x-license-validator` or `@mui/x-licensor` for a short version?
https://dictionary.cambridge.org/dictionary/english/licensor

Also:
> I think that the "release info" should be owned by each package

+1 to that.

joserodolfofreitas:
> 3. I would suggest that we make the `@mui/x-license-pro` package public. We could list the need to install it in https://mui.com/components/data-grid/getting-started/#installation. In v5, we would need to do this gracefully and in v6 we would make this mandatory. The main value would be about having a single infrastructure to manage the license. Not 10 different APIs to install the license, one per component. It would be an opportunity to remove https://github.com/mui/mui-x/blob/140f39480052c1ca41561eea6fcd28c3a07e1da9/packages/x-license-pro/package.json#L12

Regarding this point, I see a lot of value in having a single infrastructure, but can't it be indirectly installed as a dependency of x-packages? I think the risk

packages/grid/x-data-grid-pro/src/DataGridPro/DataGridPro.tsx
oliviertassinari: Ok, in the future, I think that we could consider rendering a watermark portal under the `<body>`, position fixed, similar to an Intercom widget or a https://www.netlify.com/products/deploy-previews widget.

oliviertassinari (commented):
I think that it's good enough to move forward 👍. We might be able to add a test case, but maybe it's overkill.

On top of #4250 and #4251, I have a couple of more ideas to improve the DX with the license:

1. In the console error, when we log `showInvalidLicenseError` we could plot the license key used. It could help developers that are using env variable realize they used the wrong key.
2. In the console error, when we log `showExpiredLicenseError` we could plot the expiration date of their license, as well as the package+version+date they tried to use that is no longer valid. Without such information, it can be a which hunt for developers to figure out how to solve the problem 🧙‍♀️🙃.

Added to #1364

PR #4126, opened 2022-03-08
[core] Add CLI to decode license key

flaviendelangle:
@mbrookes, with this you would be able to do:

```shell
npm i -g @mui/x-license-pro
npx licensedecode -k xxx
```

I took the same naming strategy as `licensegen` but in a 2nd version I would like to correctly namespace the commands and do:

```shell
npx mui-license gen -o xxx
npx mui-license decode -k xxx
```

### DIFF
--- a/packages/x-license-pro/bin/license-decode-script.js
+++ b/packages/x-license-pro/bin/license-decode-script.js
@@ -1,0 +1,5 @@
+#!/usr/bin/env node
+/* eslint-disable */
+require = require('esm')(module);
+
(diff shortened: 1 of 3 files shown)


mbrookes (commented):
Great, thanks!

oliviertassinari:
In the long run, I wonder if we shouldn't progressively move from the CLI to only custom UI builds with our low-code product cc @mui/toolpad. CLIs are great but they are as useful as they can be discovered. If we have a page that lists all the CLI, we might want to make it interactive, directly, skipping the need to install and run the scripts.

PR #4154, opened 2022-03-10
[core] Keep same reference to the column visibility model if no column has changed

m4theushw:
The test below is failing when running with React 18 in #4155. 

image

The wrong assertion is because the StrictMode in React 18 has [StrictEffects](https://github.com/reactwg/react-18/discussions/19), causing all effects to be called twice, no matter if the dependencies have changed or not. The effect that causes `onColumnVisibilityModelChange` to be called a 2nd time is the following one that is supposed to recreate the columns state:

https://github.com/mui/mui-x/blob/1922f48f0f230cb48bb5915bde0eaec1d0ae8388/packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.ts#L398

Since we pass `reset=true`, the column visibility model for `hide` is always reinitialized.

https://github.com/mui/mui-x/blob/1922f48f0f230cb48bb5915bde0eaec1d0ae8388/packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts#L312-L317

This PR changes to only reinitialize it if a column has changed its visibility. In other cases it will keep the reference. Despite that, I also renamed some variables to improve readability. I can't add a test here because there's no way to call again a effect without changing its dependencies. To test it's better to download #4155 and apply the changes here there.


packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
```diff
@@ -311,39 +313,64 @@ export const createColumnsState = ({
...
+      // edge case where the column is not in the model but it also doesn't exist in the `columns`
+      // prop, meaning that the column is being added. In that case, we assume that the column was
+      // not visible before for it be added to the model.
+      let isVisibleBefore = currentColumnVisibilityModel[field];
+      if (isVisibleBefore === undefined) {
+        if (isInsideStateInitializer) {
+          isVisibleBefore = true;
+        } else {
+          const currentState = gridColumnsSelector(apiRef.current.state);
+          isVisibleBefore = currentState.all.includes(field);
```
m4theushw, line 357: This change was because https://github.com/mui/mui-x/blob/1922f48f0f230cb48bb5915bde0eaec1d0ae8388/packages/grid/x-data-grid/src/tests/layout.DataGrid.test.tsx#L602 was failing.

flaviendelangle: Can we avoid the `O(n²)` complexity here by using the lookup ?

packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
```diff
@@ -259,17 +259,19 @@ export const createColumnsState = ({
   columnsTypes,
   currentColumnVisibilityModel = gridColumnVisibilityModelSelector(apiRef),
   shouldRegenColumnVisibilityModelFromColumns,
-  reset,
+  keepOnlyColumnsToUpsert = false,
```
flaviendelangle, line 262: Nothing against the two variable renamed

packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
```diff
@@ -311,39 +313,64 @@ export const createColumnsState = ({
...
-        if (
-          !columnsToUpsertLookup[field] &&
-          columnsLookupBeforePreProcessing[field] === columnsStateWithPreProcessing.lookup[field]
-        ) {
-          return;
+        newColumnVisibilityModel[field] =
+          !columnsStateWithoutColumnVisibilityModel.lookup[field].hide;
+      });
+    } else if (keepOnlyColumnsToUpsert) {
+      // At this point, `keepOnlyColumnsToUpsert` changes its meaning to keep the columns
```
flaviendelangle, line 325:
```suggestion
      // At this point, `keepOnlyColumnsToUpsert` have a new meaning: keep the columns
```

The sentence is valid but I had to read it a few times to understand

flaviendelangle (commented):
v6 will simplify this so much...

packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
m4theushw: Feel free to suggest improvements.

flaviendelangle (approved):

Issue #701, opened 2020-12-06
[DataGrid] Keyboard events trigger interactions twice


oliviertassinari:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

When focusing an input, the grid handle <kbd>Space</kbd>, <kbd>ArrowLeft</kbd>, etc.

## Expected Behavior 🤔

When focusing an input, the grid should ignore most of the keyboard events.

## Steps to Reproduce 🕹

Steps:

1. https://codesandbox.io/s/material-demo-forked-4k9vo?file=/package.json
2. Try typing text containing one space, or using the arrow to make an edit.

![Dec-06-2020 20-39-42](https://user-images.githubusercontent.com/3165635/101290533-3890c080-3803-11eb-83e1-4c0407a01289.gif)

## Context 🔦

Noticeable since #673. The issue come to my attention looking at #700.

## Your Environment 🌎

| Tech          | Version |
| ------------- | ------- |
| Material-UI X | not released yet  |

Issue #4151, opened 2022-03-10
[data-grid] Terminology for various row listings


flaviendelangle:
Related to #3717 

The grid contains more than half a dozen row listings depending on what the user (or our team) is trying to achieve.
In v5, some of those listings were not consistently named, one of the goal of v6 is to clarify as much as possible that part.

# Proposals

- Decide one terminology for each listing above and use this terminology in the name of every selector / api method / params property that uses this listing

- Remove `apiRef.current.getRowIndex`, keep `apiRef.current.getRowIndexRelativeToVisibleRows` and maybe create a `apiRef.current.getRowIndexRelativeToExpandedRows` (see below to discuss the naming of the "expanded" row listing)

- Remove all the api methods that are just calling a selector (`apiRef.current.getAllRowIds` for instance)

# Existing listings

## Rows in the order given by the user

Terminology in v5: `rows`
Terminology in v6: `rows`

### Usages

- `gridRowIdsSelector` 

- `apiRef.current.getRowsCount`
- `apiRef.current.getAllRowIds` (delete in v6, can use `gridRowIdsSelector`)

## Rows after the sorting

Terminology in v5: `sorted rows`
Terminology in v6: `sorted rows`

### Usages

- `gridSortedRowIdsSelector` 
- `gridSortedRowEntriesSelector`

- `apiRef.current.getSortedRows` (delete in v6, can use `gridSortedRowEntriesSelector`)
- `apiRef.current.getSortedRowIds` (delete in v6, can use `gridSortedRowIdsSelector`)
- `apiRef.current.getRowIdFromRowIndex` (very unclear, delete in v6, can any row id selector to get the index from the wanted row listing)

## Rows after the sorting and filtering

Terminology in v5: `filtered sorted rows`
Terminology in v6: `filtered sorted rows` (I keep the `sorted` because we could have a filtered non-sorted row listing)

- `gridFilteredSortedRowIdsSelector`
- `gridFilteredSortedRowEntriesSelector`

## Rows after the sorting and filtering + without the collapsed children

Terminology in v5: `visible rows`
Terminology in v6: `expanded rows` ? `sorted expanded rows` ? `non collapsed rows` ? `all page rows` ?

If the tree is flat, this listing is equal to "Rows after the sorting and filtering".
But they must be clearly distinguished because when we have row children, depending on the use case we must use one or another.
For instance the keyboard navigation uses this listing whereas some exports use the filtered rows to include the collapsed children.

### Usages

[...]

PR #4157, opened 2022-03-11
[DataGrid] Improve selection with keyboard

### LINKED ISSUE #4156: [DataGrid] Two rows are deselected when using the shift key
### Duplicates

### Latest version

### Current behavior 😯

Select multiple rows.
Deselect the last selected row using the Shift key.

Two rows are deselected.

https://user-images.githubusercontent.com/36615680/157634519-97b1c535-c709-4c4d-ac17-769f9008954d.mov

### Expected behavior 🤔

Only one row is deselected.

### Steps to reproduce 🕹

Steps: https://codesandbox.io/s/233v6d?file=/demo.tsx MUI official example

1. Select multiple rows
2. Deselect the last selected row using `shift key + click`.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
System:
    OS: macOS 12.1
  Binaries:
    Node: 14.17.0 - /usr/local/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.14.13 - /usr/local/bin/npm
  Browsers:
    Chrome: 99.0.4844.51
    Edge: Not Found
    Firefox: 78.0.2
    Safari: 15.2
  npmPackages:
    @types/react: ^17.0.3 => 16.14.1 
    react: ^17.0.2 => 16.14.0 
    react-dom: ^16.13.1 => 16.14.0 
    styled-components: ^5.2.0 => 5.2.1 
    typescript: ^4.2.3 => 3.8.3 
```
</details>

### Order ID 💳 (optional)

_No response_

### LINKED ISSUE #4185: [data-grid] Pressing Shift + ArrowUp unselect two rows when the row above the focus is selected
### Duplicates

### Latest version

### Current behavior 😯

https://user-images.githubusercontent.com/3309670/158176732-ba216b0c-9ff8-4dcd-b8c4-7ce627bf0df4.mp4

### Expected behavior 🤔

https://user-images.githubusercontent.com/3309670/158176743-4becd124-e784-4d63-a36e-94f0134f06b8.mp4

### Steps to reproduce 🕹

Steps:

1. Go to https://mui.com/components/data-grid/selection/#multiple-row-selection
2. Select multiple rows
3. Press Shift + Arrow up while focusing the last row of the selected range

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #4185 

- [x] When reducing a selection range, keep the focus key selected
- [x] Drop `useGridKeyboard` (movie the publication of `cellNavigationKeydown` in `useGridKeyboardNavigation` and the content of `expandCellSelection` in `useGridSelection`)


packages/grid/x-data-grid/src/DataGrid/useDataGridComponent.tsx
```diff
@@ -64,6 +63,7 @@ export const useDataGridComponent = (props: DataGridProcessedProps) => {
   useGridInitializeState(rowsMetaStateInitializer, apiRef, props);
   useGridInitializeState(columnMenuStateInitializer, apiRef, props);
 
+  useGridKeyboardNavigation(apiRef, props);
```
flaviendelangle, line 66: `useGridSelection` must consume the `cellKeydown` after the navigation have been applied.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -388,25 +392,67 @@ export const useGridSelection = (
...
+            }
+          } else if (isPreviousRowSelected) {
+            // We are navigating to the top of the page and removing selected rows
+            start = newRowIndex + 1;
+            end = previousRowIndex;
+          } else {
+            // We are navigating to the top of the page and adding selected rows
+            start = newRowIndex;
+            end = previousRowIndex - 1;
+          }
```
alexfauquette, line 444: Sounds easier to read width a 
```js
if (newRowIndex > previousRowIndex){
	if (isPreviousRowSelected) { ... }
	else { ... }
}
else {
	if (isPreviousRowSelected) { ... }
	else { ... }
}
````

such that we directly understand the first conditional level is about "moving top/bottom" second level is about adding/removing

**Review by alexfauquette — COMMENTED** [review]
From the preview, it seems you did not fix the issue #4156

In `expandRowRangeSelection`, there is 
```js
if (startIndex > endIndex) {
endId = visibleRowIds[endIndex + 1];
} else {
endId = visibleRowIds[endIndex - 1];
}
```

Which forget the case when user click on the same cell 
```js
else if (startIndex === endIndex) {
	endId = visibleRowIds[endIndex];
}
```

flaviendelangle:
> From the [preview](https://deploy-preview-4157--material-ui-x.netlify.app/components/data-grid/selection/), it seems you did not fix the issue https://github.com/mui/mui-x/issues/4156

https://deploy-preview-4157--material-ui-x.netlify.app/components/data-grid/selection/#multiple-row-selection

https://user-images.githubusercontent.com/3309670/157883552-ba7aa821-b6c5-4571-a7f4-db0d240c1627.mp4

For me the behavior is good now

> Which forget the case when user click on the same cell

It is handled by an early return above:

```
        if (!focusCell || focusCell.id === params.id) {
          return;
        }
```

alexfauquette (approved):
The bug could be tested, but I'm not sure it's interesting to add it

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -388,25 +395,70 @@ export const useGridSelection = (
...
+        handleSingleRowSelection(params.id, event);
         return;
       }
 
       if (event.key.toLowerCase() === 'a' && (event.ctrlKey || event.metaKey)) {
         event.preventDefault();
         selectRows(apiRef.current.getAllRowIds(), true);
       }
+
+      if (isNavigationKey(event.key) && event.shiftKey) {
```
m4theushw, line 414: Could you create an issue or explain in the PR description what you're fixing here? https://github.com/mui/mui-x/pull/4157/#issuecomment-1065146849

m4theushw (commented):
I would be very careful when changing how the selection works. I would prefer to have a PR only fixing #4156. With #4157 I can't expand the selection holding <kbd>Shift</kbd> while pressing <kbd>ArrowDown</kbd>.

https://user-images.githubusercontent.com/42154031/157904399-2da71cab-43ad-4913-8e00-4b0f69ba85ba.mp4

This happens because we do nothing if the target is not the cell root. There's a TODO to improve this check but I don't know the side effects doing it now.

https://github.com/mui/mui-x/blob/50f084dd73777b18dc86fc7e13a90da40f5741ec/packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts#L387

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -50,8 +51,10 @@ export const selectionStateInitializer: GridStateInitializer<
 });
 
 /**
- * @requires useGridRows (state, method)
- * @requires useGridParamsApi (method)
+ * @requires useGridRows (state, method) - can be after
+ * @requires useGridParamsApi (method) - can be after
+ * @requires useGridFocus (state) - can be after
```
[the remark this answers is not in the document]
flaviendelangle: Yes I need to clean those

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -108,6 +112,9 @@ export const useGridSelection = (
         const visibleRowIds = gridVisibleSortedRowIdsSelector(apiRef);
         const startIndex = visibleRowIds.findIndex((rowId) => rowId === startId);
         const endIndex = visibleRowIds.findIndex((rowId) => rowId === endId);
+        if (startIndex === endIndex) {
+          return;
```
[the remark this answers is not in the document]
flaviendelangle: And Google Drive doe nothing
Which one feel the bet to the team here ?

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -108,6 +112,9 @@ export const useGridSelection = (
         const visibleRowIds = gridVisibleSortedRowIdsSelector(apiRef);
         const startIndex = visibleRowIds.findIndex((rowId) => rowId === startId);
         const endIndex = visibleRowIds.findIndex((rowId) => rowId === endId);
```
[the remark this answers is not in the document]
flaviendelangle: I don't think so because when expanding with the mouse, we support cross-page expansion.

You click on a row of page 1
You go to page 2
You shift click a row of page 2

=> It selects all the row between your two clicks.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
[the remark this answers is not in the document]
alexfauquette: For me both are ok. If a user whats to unselect the last row, she can shift+click on the previouse row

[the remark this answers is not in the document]
flaviendelangle: Yes if we choose the "do nothing" I will note it in #4156 :+1:

flaviendelangle:
> I would be very careful when changing how the selection works. I would prefer to have a PR only fixing https://github.com/mui/mui-x/issues/4156.

I will extract the fix for #4156 in a standalone PR :+1: 

> This happens because we do nothing if the target is not the cell root. There's a TODO to improve this check but I don't know the side effects doing it now.

@m4theushw ok so before this PR, the `isGridCellRoot` check was done to select to row by pressing Enter + Shift or to select all rows by pressing Ctrl + A
But it was not applied to expand the row selection by pressing Shift + Arrow Up / Down.

Do we need the check at all for selection ? Was this check for some scenarios but not others the right behavior or just an error ?

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
flaviendelangle: https://github.com/mui/mui-x/issues/4185

Here it is

m4theushw: Yeah, it's a bug. Can you move this code to another PR?

m4theushw:
> Do we need the check at all for selection ? Was this check for some scenarios but not others the right behavior or just an error ?

The check was added to fix #701. It might be useless now. But if keep it, then it's better to extend to check if the `keydown` came from inside the cell, not only the root.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
m4theushw: Selection with <kbd>Shift</kbd> + arrow keys is still broken. I would keep this PR as small as possible, only fixing #4156, which is simple.

flaviendelangle: What I said is that I am doing the opposite
Instead of moving 90% of this PR to another one, I am moving the 10% related to #4156 and keeping the rest here.

It does not change anything, it's just easier to do.

flaviendelangle: https://github.com/mui/mui-x/pull/4196

packages/grid/x-data-grid/src/hooks/features/keyboardNavigation/useGridKeyboardNavigation.ts
```diff
@@ -262,6 +270,24 @@ export const useGridKeyboardNavigation = (
...
+      if (!event.currentTarget.contains(event.target as Element)) {
+        return;
+      }
+
+      // Get the most recent params because the cell mode may have changed by another listener
+      const cellParams = apiRef.current.getCellParams(params.id, params.field);
+
+      if ((cellParams.cellMode === GridCellModes.Edit) !== isNavigationKey(event.key)) {
+        apiRef.current.publishEvent(GridEvents.cellNavigationKeyDown, cellParams, event);
+      }
```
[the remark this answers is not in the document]
flaviendelangle: :facepalm:  yes

packages/grid/x-data-grid/src/hooks/features/keyboardNavigation/useGridKeyboardNavigation.ts
```diff
@@ -262,6 +270,24 @@ export const useGridKeyboardNavigation = (
...
+    (params, event) => {
+      // Ignore portal
+      if (!event.currentTarget.contains(event.target as Element)) {
+        return;
+      }
+
+      // Get the most recent params because the cell mode may have changed by another listener
+      const cellParams = apiRef.current.getCellParams(params.id, params.field);
+
+      if (cellParams.cellMode !== GridCellModes.Edit && isNavigationKey(event.key)) {
```
m4theushw, line 283: <kbd>Backspace</kbd> is not a navigation key if it is pressed while the focus is on the checkbox column. It jumps if a click only on the border of the checkbox. By doing an additional check here, you can remove the `event.preventDefault`. Also, the modification in the order that the hooks are called can be reverted.

msedge_5cXogi4eMh

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -384,23 +388,69 @@ export const useGridSelection = (
...
+      if (apiRef.current.getCellMode(params.id, params.field) === GridCellModes.Edit) {
+        return;
+      }
+
+      if (isNavigationKey(event.key) && event.shiftKey) {
+        const focusCell = gridFocusCellSelector(apiRef);
+        if (focusCell && focusCell.id !== params.id) {
+          event.preventDefault();
+
+          const isPreviousRowSelected = apiRef.current.isRowSelected(focusCell.id);
```
m4theushw, line 401: We're relying on the fact that the focus was not updated yet. If the order of the listeners is changed, or we change to listen for the event in the capture phase, it won't work. Would it work if `lastRowToggled` is used here instead?

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -384,23 +388,69 @@ export const useGridSelection = (
...
+                end = previousRowIndex - 1;
+              }
+            }
+
+            const rowsBetweenStartAndEnd = visibleRows.rows
+              .slice(start, end + 1)
+              .map((row) => row.id);
+            apiRef.current.selectRows(rowsBetweenStartAndEnd, !isPreviousRowSelected);
+          } else {
+            apiRef.current.selectRow(focusCell.id, !isPreviousRowSelected, true);
```
m4theushw, line 437: Could you invert `canHaveMultipleSelection` -> `!canHaveMultipleSelection` and put this logic at the top, with a immediate return? The logic above is a bit long so we scroll a few lines to see only 1 in the else block. The sooner we return, the better to read.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -384,23 +388,69 @@ export const useGridSelection = (
...
+            } else {
+              // eslint-disable-next-line no-lonely-if
+              if (isPreviousRowSelected) {
+                // We are navigating to the top of the page and removing selected rows
+                start = newRowIndex + 1;
+                end = previousRowIndex;
+              } else {
+                // We are navigating to the top of the page and adding selected rows
+                start = newRowIndex;
+                end = previousRowIndex - 1;
```
m4theushw, line 428: There's a bug here. If I press <kbd>ArrowUp</kbd> already with <kbd>Shift</kbd> pressed, it ignores the first row.

msedge_ubqGBx7SP5

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -346,48 +346,48 @@ describe('<DataGrid /> - Selection', () => {
     });
   });
 
+  /* eslint-disable material-ui/disallow-active-element-as-key-event-target */
```
m4theushw, line 349: Could you update the tests so we don't need to ignore this ESLint rule? To keep consistent with the core.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
flaviendelangle: Fixed

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
flaviendelangle: > We're relying on the fact that the focus was not updated yet.

The opposite, I missed-named this variable, it should be `isNewRowSelected`
I access the previous row through `params.id` and the new one through `focusCell.id`

I don't think we could easily cover all use cases
If you navigate with the keyboard without selecting rows, your "previousRow" still needs to be the cell the focus is currently in, not the last row toggled.
I agree that this is a constraint on the call order and that is why I am putting `useGridKeyboardNavigation` before `useGridSelection`.

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -165,13 +165,13 @@ describe('<DataGrid /> - Selection', () => {
...
       expect(getSelectedRowIds()).to.deep.equal([0]);
-
-      fireEvent.keyDown(cell00, {
-        key: 'ArrowDown',
-        shiftKey: true,
-      });
-
-      expect(getSelectedRowIds()).to.deep.equal([0]);
+      //
+      // fireEvent.keyDown(cell00, {
```
m4theushw, line 169: Commented code?

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -384,23 +388,69 @@ export const useGridSelection = (
...
+              end = newRowIndex;
+            }
+          } else {
+            // eslint-disable-next-line no-lonely-if
+            if (isNextRowSelected) {
+              // We are navigating to the top of the page and removing selected rows
+              start = newRowIndex + 1;
+              end = previousRowIndex;
+            } else {
+              console.log('HEY');
```
m4theushw, line 436: ⚠

```suggestion
```

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
m4theushw: Sounds fair. I only don't like the constraint on the call order, but I don't have any other simple solution. We could listen to `cellFocusIn` and try to track if the new focus came from the keyboard navigation. It's a way to not depend on the call order but much more complex. We can leave it as it is.

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
```diff
@@ -384,23 +388,69 @@ export const useGridSelection = (
...
         return;
       }
 
-      // Get the most recent params because the cell mode may have changed by another listener
-      const cellParams = apiRef.current.getCellParams(params.id, params.field);
-      const isEditMode = cellParams.cellMode === GridCellModes.Edit;
-      if (isEditMode) {
-        return;
+      if (isNavigationKey(event.key) && event.shiftKey) {
+        const focusCell = gridFocusCellSelector(apiRef);
```
m4theushw, line 403:
```suggestion
        // The cell that has focus after the keyboard navigation
        const focusedCell = gridFocusCellSelector(apiRef);
```

packages/grid/x-data-grid/src/hooks/features/selection/useGridSelection.ts
flaviendelangle: Alternatively, the keyboard navigation could publish an event `keyboardNavigation` with the cell before and after.

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
flaviendelangle: Done

m4theushw (commented):
I had a quick look. We definitely need this demo but I'm not sure if we need it now. We don't know yet why someone would want a selector that returns only filtered rows, but not sorted. I guess the general use case is to return the currently visible rows or, at the most, the top level rows / expanded rows. Maybe wait for #4151?

packages/grid/x-data-grid/src/tests/selection.DataGrid.test.tsx
```diff
@@ -420,7 +419,8 @@ describe('<DataGrid /> - Selection', () => {
       const data = getData(20, 1);
       render(<TestDataGridSelection {...data} rowHeight={50} checkboxSelection hideFooter />);
       const checkboxes = screen.queryAllByRole('checkbox', { name: /select row/i });
-      checkboxes[0].focus();
+      fireClickEvent(checkboxes[0]);
+      expect(document.activeElement).to.equal(checkboxes[0]);
```
m4theushw, line 423:
```suggestion
      expect(checkboxes[0]).toHaveFocus();
```

m4theushw (approved):

PR #4186, opened 2022-03-14
[docs] Simplify server examples

alexfauquette:
## Context

For now, we fake server interaction in examples, and to do so we need to use the hook `useDemoData` which returns all the data. Then the code useful for server interaction is mixed with the code used to simulate server delay.

## Proposal

This PR aims to simplify fake server simulation and make it closer to what developers experiment in there codebase. To do so, I add to the package `x-data-grid-generator` a `serverConfiguration` which returns statics data such as `columns` definition, `initialState`, and `useQuery` hook inspired from apollo-client hook.

- To be close to developers' issues, the returned object of `useQuery` is the same as the apollo-client hook.
- To be easy to read, `useQuery` parameters are close to DataGrid state, such that we don't have to map them to a graphQL or a REST request

At the end, server demos should have their server interactions reduced to the following code snippet:
```jsx
const queryOptions = React.useMemo(() => ({page, pageSize, filterModel, sortModel}), [...])
const { isLoading, data, rowCount } = useQuery(queryOptions);
```

## Updated Demo preview

- [server side filtering](https://deploy-preview-4186--material-ui-x.netlify.app/components/data-grid/filtering/#server-side-filter)
- [server side sorting](https://deploy-preview-4186--material-ui-x.netlify.app/components/data-grid/sorting/#server-side-sorting)
- [server pagination](https://deploy-preview-4186--material-ui-x.netlify.app/components/data-grid/pagination/#basic-implementation)
- [server cursor pagination](https://deploy-preview-4186--material-ui-x.netlify.app/components/data-grid/pagination/#cursor-implementation)

## TODO 

- [x] Add sorting
- [x] Add filtering
- [ ] Add `useMutation`? (=> for another PR)


docs/data/data-grid/filtering/ServerFilterGrid.tsx
```diff
@@ -1,83 +1,39 @@
...
 
-function loadServerRows(commodityFilterValue?: string): Promise<any> {
-  const serverRows = [
-    { id: '1', commodity: 'rice' },
-    { id: '2', commodity: 'soybeans' },
-    { id: '3', commodity: 'milk' },
-    { id: '4', commodity: 'wheat' },
-    { id: '5', commodity: 'oats' },
-  ];
+const DATASET_OPTION: UseDemoDataOptions = {
```
m4theushw, line 5: To be consistent with `SERVER_OPTIONS`. Same for other demos.

```suggestion
const DATASET_OPTIONS: UseDemoDataOptions = {
```

docs/data/data-grid/pagination/ServerPaginationGrid.tsx
```diff
@@ -1,74 +1,38 @@
...
+const DATASET_OPTION: UseDemoDataOptions = {
+  dataSet: 'Commodity',
+  rowLength: 100,
+  maxColumns: 6,
+};
 
-  return { isLoading, data, rowCount };
+const SERVER_OPTIONS = {
+  minDelay: 100,
+  maxDelay: 300,
```
m4theushw, line 13: Why not make these the default values? We're repeating them quite often.

docs/data/data-grid/pagination/CursorPaginationGrid.tsx
```diff
@@ -1,92 +1,77 @@
...
+      // We add nextCursor when available
+      mapPageToNextCursor.current[page] = nextCursor;
+    }
+  }, [page, nextCursor, isLoading]);
 
-    return () => {
-      active = false;
-    };
-  }, [page, data]);
+  // Some api client return undefine while loading
```
m4theushw, line 52: Some typos.

```suggestion
  // Some API clients return undefined while loading
```

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,246 @@
...
+export interface QueryOptions {
+  cursor?: GridRowId;
+  page?: number;
+  pageSize?: number;
+  // TODO: implement the behavior liked to following models
+  filterModel?: GridFilterModel;
+  sortModel?: GridSortModel;
+}
+
+export const serverConfiguration = (
```
m4theushw, line 180: It needs to be an imperative name since it's a function. AG Grid calls it `createFakeServer`.

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,246 @@
...
+    }, [dataGenerationIsLoading, queryOptions, rows]);
+
+    // We use queryOptions pointer to be sure that isLoading===true as soon as the options change
+    const effectShouldStart = queryOptionsRef.current !== queryOptions;
+
+    return {
+      isLoading: dataGenerationIsLoading || isLoading || effectShouldStart,
+      data,
+      rowCount,
+      nextCursor,
```
m4theushw, line 241: With the exception of `isLoading`, a request to an API usually brings back the row count and the next cursor in the same object. If we save this object in the state it means only a single rerender. Instead, here it requires at least 2 more rerenders to update these other values.

image

m4theushw (commented):
For inspiration:

- https://github.com/marmelab/react-admin/tree/master/packages/ra-data-fakerest
- https://github.com/marmelab/FakeRest

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
alexfauquette: If I understand correctly the setState hooks are not batched because they are in a promise

alexfauquette:
@m4theushw I looked at the react-admin way to handle server demo. I think adding a library to fake requests will require much more time because there is all the filtering model to translate in REST. I have really poor knowledge about REST, and our filtering/sorting model is more flexible than react-admin models

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,271 @@
...
+        // dataset is not ready
+        return () => {};
+      }
+
+      queryOptionsRef.current = queryOptions;
+      let active = true;
+
+      setIsLoading(true);
+      setResponse({ pageInfo: {}, data: [] });
+      loadServerRows(rows, queryOptions, serverOptionsWithDefault, columnsWithDefaultColDef).then(
```
DanailH, line 239: Can we convert it to `async / await`?

alexfauquette: Would it be easier to understand? I personally prefer to read all the code in its running order, instead of defining an async function of two lines that will be called at the end

alexfauquette:
@DanailH You have the second argument of `createFakeServer` which allows modifying the response delay by providing a `minDelay` and `maxDelay`. The response delay will be chosen uniformly between this two bounds provided in `ms`.

All the demos use the default values
```js
const DEFAULT_SERVER_OPTIONS: ServerOptions = {
  minDelay: 100,
  maxDelay: 300,
  useCursorPagination: true,
};
```

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
DanailH: I guess it is more of personal preferences. I think In the codebase we opt-in for using async / await. Also, you can make a self-executing function in the `useEfect` and not have it being defined and called immediately after.

alexfauquette: done

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
m4theushw: Yeah, each call to `setState` will cause a rerender. In https://github.com/mui/mui-x/pull/4186/files/7d68109152a6cc509b3b7305f0c41f9652133117#diff-94d76f5b777c114259dd01edd0339fd1541887e6b5646ad97b14fd1d2fcfec5dR222-R225 we first update the data, then the row count. This may cause that we first get the data without the correct row count. One way to solve this is to save the entire object a single `setState`.

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,275 @@
...
+      if (dataGenerationIsLoading) {
+        // dataset is not ready
+        return () => {};
+      }
+
+      queryOptionsRef.current = queryOptions;
+      let active = true;
+
+      setIsLoading(true);
+      setResponse({ pageInfo: {}, data: [] });
```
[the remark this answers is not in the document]
alexfauquette: Resetting data was a bad idea. It leads to really strange behavior. I now just reset `pageInfo` to imitate the problem we faced with apollo Client which set `rowNumber` to undefined

alexfauquette:
When the column width change, it's because I replaced the grid content. For example, `[ServerFilterGrid](https://github.com/mui/mui-x/pull/4186/files#diff-f70cbc4d35bca441b3f93502d9babdbd0ac9657720302b457b39f6b49a840df6)` demo set `serverRows` to simulate the server, now it uses the default demo data

packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,280 @@
...
+
+const DEFAULT_DATASET_OPTIONS: UseDemoDataOptions = {
+  dataSet: 'Commodity',
+  rowLength: 100,
+  maxColumns: 6,
+};
+
+const DEFAULT_SERVER_OPTIONS: ServerOptions = {
+  minDelay: 100,
+  maxDelay: 300,
```
oliviertassinari, line 196: We might want to add a logic as we did for `DISABLE_CHANCE_RANDOM` to disable this for the visual regression tests, it's unstable: https://www.argos-ci.com/mui/mui-x/builds/1826

Screenshot 2022-06-05 at 15 37 53

Issue #4144, opened 2022-03-10
[docs] Incorrect import


jamesdbruner:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Sorry to make a bug ticket for the docs having typos but this is essentially it: Throughout the data grid docs you'll find imports from `@mui/x-data-grid-generator` but that package doesn't exist. `@material-ui/x-data-grid-generator` is the correct import.

To reiterate...
```js
import { useDemoData } from '@mui/x-data-grid-generator';
```

Should be
```js
import { useDemoData } from '@material-ui/x-data-grid-generator';
```

flaviendelangle:
https://www.npmjs.com/package/@mui/x-data-grid-generator

`@mui/x-data-grid-generator` does exist.

Our packages have been renamed from `@material-ui/...` to `@mui/...` last Fall.

jamesdbruner:
Hmm well, I'm having issues installing that package in that case. My first guess would have been that you guys are using something like verdaccio or nexus internally which is why it resolves for you but I'm really not sure. It could just be my machine but other packages download and install just fine. This is the command I ran and the output: `npm i @mui/x-data-grid-generator`

```node
npm ERR! code ENOENT
npm ERR! syscall chmod
npm ERR! path /Desktop/Projects/demo/node_modules/@mui/x-data-grid-generator/bin/data-gen-script.js
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, chmod '/Desktop/Projects/demo/node_modules/@mui/x-data-grid-generator/bin/data-gen-script.js'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

This is what my package.json deps looks like 
```json
  "dependencies": {
    "@emotion/react": "^11.8.1",
    "@emotion/styled": "^11.8.1",
    "@fontsource/roboto": "^4.5.3",
    "@mui/icons-material": "^5.5.0",
    "@mui/material": "5.5.0",
    "@mui/system": "^5.2.8",
    "@mui/x-data-grid-pro": "^5.6.0",
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
  },
  "devDependencies": {
    "@vitejs/plugin-react": "^1.0.7",
    "vite": "^2.8.0"
  },
  "peerDependencies": {
    "@mui/material": "^5.0.0",
    "react": "^17.0.0",
[...]

PR #4193, opened 2022-03-14
[core] Check if `process` is available

m4theushw:
Fixes https://github.com/mui/mui-x/issues/4144#issuecomment-1066853139

Vite doesn't make available environment variables other than `process.env.NODE_ENV`.

### DIFF
--- a/packages/grid/x-data-grid-generator/src/services/random-generator.ts
+++ b/packages/grid/x-data-grid-generator/src/services/random-generator.ts
@@ -15,7 +15,7 @@ import { GridDataGeneratorContext } from './gridColDefGenerator';
 const chanceId = globalChance();
 let chance: Chance.Chance;
 
-if (process.env.DISABLE_CHANCE_RANDOM) {
+if (typeof process !== 'undefined' && process.env.DISABLE_CHANCE_RANDOM) {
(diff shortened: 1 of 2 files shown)


oliviertassinari (changes requested):
Argos reports that this change is not working. We do no longer enter the first `if` branch in the visual test suite.

Why is it not working? Because we do a global search and replace for `process.env`:

https://github.com/mui/mui-x/blob/ffcc4cf69c0f3cd2df760cb1969d00200f1be65c/test/regressions/webpack.config.js#L24-L28

The docs https://webpack.js.org/plugins/define-plugin/ recommend doing the search and replace for `process.env. DISABLE_CHANCE_RANDOM` instead.

Maybe we can solve this by replacing the whole `process` variable?

m4theushw:
> Maybe we can solve this by replacing the whole process variable?

@oliviertassinari It doesn't appear to be recommended. From the docs:

> Using the latter [`process: { env: { NODE_ENV: JSON.stringify('production') } }`] will overwrite the process object which can break compatibility with some modules that expect other values on the process object to be defined.

I changed to use a global variable, instead of an attribute of `process.env`.

cherniavskii (approved):

PR #4195, opened 2022-03-15
[docs] Sync the headers with core

oliviertassinari:
A continuation of https://github.com/mui/material-ui/pull/31807


oliviertassinari:
Ideally, we would have some sort of automation to copy the files but I assume that this PR is not harming. We can work on automation if the current setup ever create bugs.

PR #3831, opened 2022-02-03
v5.5.0

flaviendelangle:

PR #4202, opened 2022-03-15
[core] Fix Lerna package change detection

oliviertassinari:
See https://github.com/mui/mui-x/pull/3831#discussion_r827443134 for why. A TL:DR of the problem:

1. `git checkout v5.6.1`
2. `yarn lerna changed`
3. LOL WUT?

```
$ yarn lerna changed
yarn run v1.22.17
$ /Users/oliviertassinari/mui-x/node_modules/.bin/lerna changed
lerna notice cli v4.0.0
lerna info Looking for changed packages since v4.0.0-alpha.25
@mui/x-license-pro
@mui/x-data-grid-generator
@mui/x-data-grid-pro
@mui/x-data-grid
lerna success found 4 packages ready to
```

I have discovered this problem trying to make my point on that `@mui/x-license-pro` should almost never be published in the current state of things. We need to use an annotated tag. It's done correctly in https://github.com/mui/material-ui. Sebastian fixed this problem in the core a while ago.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -35,8 +35,8 @@ The following steps must be proposed as a pull request.
 - [ ] Checkout the last version of the working branch
 - [ ] Make sure you have the latest dependencies installed: `yarn`.
 - [ ] Build the packages: `yarn release:build`.
-- [ ] Release the versions on NPM: `yarn release:publish` (you need your 2FA device).
-- [ ] Create a new tag named with the release you just did `git tag v4.0.0-alpha.30 && git push upstream --tag`
+- [ ] Release the versions on npm: `yarn release:publish` (you need your 2FA device).
+- [ ] Create a new tag named with the release you just did `git tag -a v4.0.0-alpha.30 -m "Version 4.0.0-alpha.30" && git push upstream --tag`
 
 ### Publish the documentation
 


scripts/README.md
```diff
@@ -35,8 +35,8 @@ The following steps must be proposed as a pull request.
 - [ ] Checkout the last version of the working branch
 - [ ] Make sure you have the latest dependencies installed: `yarn`.
 - [ ] Build the packages: `yarn release:build`.
-- [ ] Release the versions on NPM: `yarn release:publish` (you need your 2FA device).
-- [ ] Create a new tag named with the release you just did `git tag v4.0.0-alpha.30 && git push upstream --tag`
+- [ ] Release the versions on npm: `yarn release:publish` (you need your 2FA device).
+- [ ] Create a new tag named with the release you just did `git tag -a v4.0.0-alpha.30 -m "Version 4.0.0-alpha.30" && git push upstream --tag`
```
michaldudak, line 39: Couldn't Core's releaseTag script be used instead?

oliviertassinari: We could uniformize, but I won't have the time, I was already going above and beyond with this PR 😁. From my perspective, I don't really get what `releaseTag` adds to the table (feels overkill). 

As a side note, we could backport #3920

michaldudak: > I don't really get what releaseTag adds to the table

Prevents typos, at the very least :)

> we could backport https://github.com/mui/mui-x/pull/3920 to the core.

In this case, I don't see what problem does having a checklist in the PR description solve.

oliviertassinari: > Prevents typos, at the very least :)

Yeah ok 😁

> In this case, I don't see what problem does having a checklist in the PR description solve.

Personally, the main value for me was that it better explains what each step is about, I didn't notice the checkbox.

oliviertassinari:
In the event that we change something in the build structure, something that would not be detected by `yarn release:version`, we can update the version of the packages manually. AFAIK, lerna compare each package's version againt what's in the npm registry and decide to release or not during the `yarn release:publish` step.

PR #4216, opened 2022-03-17
[core] Use a pipe processor for `GridPreferencePanel` children

flaviendelangle:
Extracted from #4208 

If we want to add a preference panel from a pro / premium feature, we are currently stuck.
Maybe we need to change the whole approach of the preference panel at some point, but this PRs should not introduce any breaking change.


packages/grid/x-data-grid/src/components/panel/GridPreferencesPanel.tsx
```diff
@@ -28,12 +24,10 @@ export const GridPreferencesPanel = React.forwardRef<
...
-      )}
-
-      {!rootProps.disableColumnFilter && isFiltersTabOpen && (
-        <rootProps.components.FilterPanel {...rootProps.componentsProps?.filterPanel} />
+      {apiRef.current.unstable_applyPreProcessors(
+        'preferencePanel',
+        null,
+        preferencePanelState.openedPanelValue ?? GridPreferencePanelsValue.filters,
       )}
     </rootProps.components.Panel>
```
alexfauquette, line 32: That's a bit weird to see `apiRef.current.unstable_applyPreProcessors` called inside the rendering

Could be more readable as follow, but not 100% sure it's the same
```jsx
const panelContent = apiRef.current.unstable_applyPreProcessors(...);
...
<rootProps.components.Panel ...>
	{panelContent}
</rootProps.components.Panel>
```

flaviendelangle: It is the same yes
Done

alexfauquette (approved):
LGTM

I was a bit confused with `applyPreprocessors`. From the name I thought it was kind of a reducer which applies all preprocess methods on state data. I did not realise it could also return a value

flaviendelangle:
> I was a bit confused with applyPreprocessors. From the name I thought it was kind of a reducer which applies all preprocess methods on state data. I did not realise it could also return a value

I plan to rename all this pre-processing code to `pipe processing` 
The pattern is indeed basically a reducer.

m4theushw (approved):

PR #4222, opened 2022-03-18
[docs] Document Print export `X-Frame-Options` limitation

DanailH:
Follow up from https://github.com/mui/material-ui/pull/31807

Preview: https://deploy-preview-4222--material-ui-x.netlify.app/components/data-grid/export/#print-export

### DIFF
--- a/docs/data/data-grid/export/export.md
+++ b/docs/data/data-grid/export/export.md
@@ -185,6 +185,10 @@ The demo below shows how to add a JSON export.
 
 {{"demo": "CustomExport.js", "bg": "inline", "defaultCodeOpen": false}}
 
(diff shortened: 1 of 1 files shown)


docs/data/data-grid/export/export.md
```diff
@@ -185,6 +185,10 @@ The demo below shows how to add a JSON export.
 
 {{"demo": "CustomExport.js", "bg": "inline", "defaultCodeOpen": false}}
 
+> ⚠️ Due to the fact that the Print export relies on the usage of an `iframe` to prepare and only print the data grid and not the entire page, there is a limitation around the usage of `X-Frame-Options`.
```
alexfauquette, line 188: The reason why we use iframe is not very interesting for readers. All they want to know is how to make it work

```suggestion
> ⚠️ Due to the fact that the Print export relies on the usage of an `iframe`, there is a limitation around the usage of `X-Frame-Options`.
```

alexfauquette (approved):

PR #4227, opened 2022-03-18
[l10n] Improve Persian (fa-IR) locale

SaeedZhiany:
Farsi (Persian) is my native language, and I'm glad to help to improve this package :)

### DIFF
--- a/packages/grid/x-data-grid/src/locales/faIR.ts
+++ b/packages/grid/x-data-grid/src/locales/faIR.ts
@@ -30,8 +30,8 @@ const faIRGrid: Partial<GridLocaleText> = {
   // Export selector toolbar button text
   toolbarExport: 'خروجی',
   toolbarExportLabel: 'خروجی',
-  toolbarExportCSV: 'دانلود به‌صورت CSV',
-  // toolbarExportPrint: 'Print',
+  toolbarExportCSV: 'دانلود به صورت CSV',
+  toolbarExportPrint: 'چاپ',
 
   // Columns panel text
   columnsPanelTextFieldLabel: 'پیداکردن ستون',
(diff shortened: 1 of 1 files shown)


packages/grid/x-data-grid/src/locales/faIR.ts
```diff
@@ -43,7 +43,7 @@ const faIRGrid: Partial<GridLocaleText> = {
   // Filter panel text
   filterPanelAddFilter: 'افزودن فیلتر',
   filterPanelDeleteIconLabel: 'حذف',
-  // filterPanelLinkOperator: 'Logic operator',
+  filterPanelLinkOperator: 'عملگر منطقی',
```
SaeedZhiany, line 46: I translated this according to the value `Logic operator`, but I see that the field name ends with `LinkOperator` that may different in real usage.

So I translated both here and you pick them as you need.

```ts
LinkOperator: 'عملگر ارتباط'
logicOperator: 'عملگر منطقی'
```

m4theushw: We want to rename this key in v6 to `filterPanelLogicOperator`. This value is used for screen readers in https://github.com/mui/mui-x/blob/1a86b4aee3481106bb14dfd3c0a339dd5943d6ce/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx#L312

You can keep the translation for "logic operator" since it means the 
AND/OR operators.

SaeedZhiany: So the current translation is ok and there is no need to change.

alexfauquette:
Thanks for your contribution 👍

PR #4229, opened 2022-03-19
[core] Reuse previous state when updating the columns prop

m4theushw:
Another test is breaking with React 18: https://app.circleci.com/pipelines/github/mui/mui-x/15295/workflows/19c27e89-53a6-4fc9-9911-d67cf320db16/jobs/89500

The reason is that StrictEffects calls intentionally the following effect twice during mount. In the first call, we skip because the state initializer is handling the state. However, in the second call, we reset the entire state as if the `columns` had changed. This causes a problem when the state is exported because `hasBeenResized` is lost when resetting the state the second time. This PR is changing to never reset `columns.lookup`.

https://github.com/mui/mui-x/blob/75e7fb627969485d2c5628b0e692cadd2e043b6c/packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.ts#L425-L433


packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.ts
```diff
@@ -430,24 +433,26 @@ export function useGridColumns(
 
     logger.info(`GridColumns have changed, new length ${props.columns.length}`);
 
+    if (
+      previousColumnsProp.current === props.columns &&
+      previousColumnTypesProp.current === columnTypes
+    ) {
+      return;
```
m4theushw, line 440: The following test fails in React 18 if we don't check again if the props changed since the last time the state was recreated:

image

The rationale is that the `columns` sets one order for the columns, then `initialState` sets another. We first initialize with `initialState` but the second effect call changes to the `columns`'s order. According to the test, the initial state should win.

docs/data/data-grid/filtering/CustomMultiValueOperator.tsx
```diff
@@ -124,15 +123,21 @@ export default function CustomMultiValueOperator() {
...
+  const columns = React.useMemo(() => {
+    const newColumns = [...data.columns];
+
+    if (newColumns.length > 0) {
+      const index = newColumns.findIndex((col) => col.field === 'quantity');
+      const quantityColumn = newColumns[index];
+
+      newColumns[index] = {
+        ...quantityColumn,
+        filterOperators: quantityOnlyOperators,
```
m4theushw, line 135: This demo had two `quantity` columns. It was keeping the same `field` for both. There's a "jump" when scrolling horizontally if two columns have the same field.

packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.ts
```diff
@@ -384,15 +387,15 @@ export function useGridColumns(
 
       const columnsState = createColumnsState({
         apiRef,
-        columnsTypes,
+        columnsTypes: columnTypes,
         columnsToUpsert: [],
         initialState: undefined,
-        shouldRegenColumnVisibilityModelFromColumns: !isUsingColumnVisibilityModel,
+        shouldRegenColumnVisibilityModelFromColumns: !isUsingColumnVisibilityModel.current,
```
m4theushw, line 393: Access `ref.current` to reduce the dependencies.

packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
```diff
@@ -339,46 +340,67 @@ export const createColumnsState = ({
...
-      if (
-        !mergedColumn.hasBeenResized &&
-        COLUMNS_DIMENSION_PROPERTIES.some((propertyName) => newColumn[propertyName] !== undefined)
-      ) {
-        mergedColumn.hasBeenResized = true;
-      }
+    const validDimensions = COLUMNS_DIMENSION_PROPERTIES.reduce<string[]>((acc, key) => {
+      const value = newColumn[key];
+      return typeof value === 'number' && value !== Infinity ? [...acc, key] : acc;
+    }, []);
```
alexfauquette, line 386: why not keep the `COLUMNS_DIMENSION_PROPERTIES.some(...)` since you do not reuse `validDimensions` except to test its length?

packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.ts
alexfauquette: This seems to be new version of the check just before with `isFirstRender.current` which could be removed

The check has been added in #2782

m4theushw: I didn't remove it because it was not breaking anything but yeah it can be removed. All of these `isFirstRender.current` checks won't work with StrictEffects. We can ensure a effect only runs once but we can't ensure it won't be called in the first render. The effects should be resilient to be called more than once: https://github.com/reactwg/react-18/discussions/18

packages/grid/x-data-grid/src/hooks/features/columns/gridColumnsUtils.ts
alexfauquette: You can do the same with `some` such that we understand directly why there is a loop on `COLUMNS_DIMENSION_PROPERTIES`
```js
const hasValidDimension = COLUMNS_DIMENSION_PROPERTIES.some((key) => {
      const value = newColumn[key];
      return typeof value === 'number' && value !== Infinity;
    }); 
```

m4theushw: Updated.

packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.tsx
```diff
@@ -179,22 +182,22 @@ export function useGridColumns(
...
-    [apiRef, columnsTypes],
+    [apiRef, columnTypes],
   );
 
   const updateColumns = React.useCallback<GridColumnApi['updateColumns']>(
     (columns) => {
       const columnsState = createColumnsState({
         apiRef,
-        columnsTypes,
+        columnsTypes: columnTypes,
```
flaviendelangle, line 192:
```suggestion
        columnTypes: columnTypes,
```

Not blocking of course, but we can rename it on `createColumnState` as well

packages/grid/x-data-grid/src/tests/columns.DataGrid.test.tsx
```diff
@@ -26,7 +26,7 @@ describe('<DataGridPro /> - Columns', () => {
     </div>
   );
 
-  describe('prop: initialState.columns.columns', () => {
+  describe('prop: initialState.columns', () => {
```
flaviendelangle, line 29:
```suggestion
  describe('prop: initialState.columns.orderedFields / initialState.columns.dimensions', () => {
```

`columns.columns` was the old name during my column dimension and order PR
But I think we still need to be explicit, because `initialState.columns.columnVisibilityModel` is in another `describe`

flaviendelangle (approved):

PR #4245, opened 2022-03-20
[l10n] Improve Korean (ko-KR) locale

kyeongsoosoo:
I found that the translation for `filterOperatorIsAnyOf` is missing, so I added☺️


flaviendelangle (approved):

flaviendelangle:

PR #4247, opened 2022-03-21
[DataGrid] Adjust type of GridColumnHeaderTitle description prop

baahrens:
As this gets passed to `Tooltip`s `title` prop, it should match the type I believe. 

ref: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Tooltip/Tooltip.d.ts#L165


flaviendelangle (approved):

flaviendelangle:

m4theushw:
This one can be included in the next release.

PR #4252, opened 2022-03-21
[DataGrid] Fix focus after stopping row edit mode

m4theushw:
Fixes https://github.com/mui/mui-x/issues/4201#issuecomment-1070922426

- Renames the `fieldFromRowBelowToFocus` param in `apiRef.current.stopRowEditMode` to `cellToFocusAfter` (same as cell editing)
- Pressing <kbd>Tab</kbd> at the last field of a row will move the focus to the first column of the next row
- Similarly, <kbd>Shift</kbd>+<kbd>Tab</kbd> will move focus to the last column of the previous row
- Adds `apiRef.current.unstable_moveFocusToRelativeCell`


packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts
```diff
@@ -29,7 +29,12 @@ export const useGridCellEditing = (
   apiRef: React.MutableRefObject<GridApiCommunity>,
   props: Pick<
     DataGridProcessedProps,
-    'editMode' | 'processRowUpdate' | 'onCellEditStart' | 'onCellEditStop'
+    | 'editMode'
+    | 'processRowUpdate'
+    | 'onCellEditStart'
+    | 'onCellEditStop'
+    | 'pagination'
```
flaviendelangle, line 36: `pagination` and `paginationMode` aren't used

packages/grid/x-data-grid/src/hooks/features/editRows/useGridEditing.new.ts
```diff
@@ -16,7 +16,10 @@ export const editingStateInitializer: GridStateInitializer = (state) => ({
 
 export const useGridEditing = (
   apiRef: React.MutableRefObject<GridApiCommunity>,
-  props: Pick<DataGridProcessedProps, 'isCellEditable' | 'editMode' | 'processRowUpdate'>,
+  props: Pick<
+    DataGridProcessedProps,
+    'isCellEditable' | 'editMode' | 'processRowUpdate' | 'pagination' | 'paginationMode'
```
flaviendelangle, line 21: Same

packages/grid/x-data-grid/src/hooks/features/editRows/useGridRowEditing.new.ts
```diff
@@ -208,8 +212,14 @@ export const useGridRowEditing = (
 
       apiRef.current.unstable_runPendingEditCellValueMutation(id);
 
-      const fieldFromRowBelowToFocus =
-        reason === GridRowEditStopReasons.enterKeyDown ? field : undefined;
+      let cellToFocusAfter: 'below' | 'right' | 'left' | 'none' | undefined;
```
flaviendelangle, line 215:
```suggestion
      let cellToFocusAfter: GridStopRowEditModeParams['cellToFocusAfter']
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -192,9 +192,16 @@ interface GridStopRowEditModeParams {
    */
   ignoreModifications?: boolean;
   /**
-   * The field from the row below to move the focus after updating the row.
+   * The field that has focus when the editing is stopped.
+   * Used to calculate which cell to move the focus after finishing editing.
```
flaviendelangle, line 196:
```suggestion
   * Used to calculate which cell to move the focus to after finishing editing.
```

Or

```suggestion
   * Used to calculate which cell should be focused after finishing editing.
```

packages/grid/x-data-grid/src/models/api/gridEditingApi.ts
```diff
@@ -192,9 +192,16 @@ interface GridStopRowEditModeParams {
...
    */
-  fieldFromRowBelowToFocus?: string;
+  field?: string;
+  /**
+   * To which cell to move focus after finishing editing.
+   * Only works if the field is also specified.
```
flaviendelangle, line 201: Maybe describe what is the behavior for the focus when the field is not specified ?

flaviendelangle (approved):

mackanfkp:
Hi, I seem to have a few issues with this, after upgrading to datagrid-pro 5.7.0.

1. It doesn't seem to consider non editable columns. As an example, I am in edit mode. I click TAB to go to the next cell, but if the that cell is not editable, shouldn't it continue to the next one? (and if that is not editable either, try the next and so on...).

2. If I set editable on all columns it works okay on page 1, when using pagination mode (client). But on the second page it doesn't work properly. It can't find the next row, because the calculated row-index is, after clamp call, larger than the indexes in the `currentPage.rows`. I think it has to do with the clamp call or how we are receiving the visible rows. 

I located the issue to `moveFocusToRelativeCell` in node_modules/@mui/x-data-grid/hooks/features/focus/useGridFocus.js

---

A few of my settings/props on the datagrid:
```
apiRef={apiRef}
editMode="row"
rows={rows}
page={page}
pageSize={pageSize}
rowsPerPageOptions={[5, 10, 25, 50, 100]}
sortModel={sortModel}
sortingOrder={['asc', 'desc']}
pagination={true}
paginationMode={'client'}
isCellEditable={(params) => { return true | false // logic for a certain cell }}
```
Debbuging:
Screenshot 2022-03-31 at 16 25 39

Actual error in console:
Screenshot 2022-03-31 at 16 36 05

m4theushw:
> It doesn't seem to consider non editable columns. As an example, I am in edit mode. I click TAB to go to the next cell, but if the that cell is not editable, shouldn't it continue to the next one? (and if that is not editable either, try the next and so on...).

I'm not sure if I understood correctly. By "consider" do you mean that it should always move the focus to the cell at right, no matter if editable or not? Currently, we jump to the next editable column. Internally we don't handle <kbd>Tab</kbd>. It's the browser that moves the focus to the next focusable element. I prepared this [CodeSandbox](https://codesandbox.io/s/basicroweditinggrid-material-demo-forked-xioktl?file=/demo.tsx) where the Age column is not editable and pressing <kbd>Tab</kbd> skips it. Could you provide more information on how to reproduce this unexpected behavior? Maybe open a new issue.

> If I set editable on all columns it works okay on page 1, when using pagination mode (client). But on the second page it doesn't work properly. It can't find the next row, because the calculated row-index is, after clamp call, larger than the indexes in the currentPage.rows. I think it has to do with the clamp call or how we are receiving the visible rows.

This will be fixed by #4326. I forgot to clamp the values. 🙄

mackanfkp:
Thanks, for very fast response and solution @m4theushw !!

Regarding the tab issue I mentioned. I see now that I explained it a bit wrong. It was when I tabbed from the last editable column, and the `processRowUpdate` was triggered, which I guess triggers the `moveFocusToRelativeCell` internally. 

But, nevermind this. Seems this probably will solve itself now that the clamp/index issue is solved. My initial thought was that it was going to focus next editable row/column, and make that editable, but I see now, in your code sandbox example, that it goes to correct row/column, and focuses the column, but not in edit mode - which is all okay!

Issue #4219, opened 2022-03-17
[DataGridPro]Cannot tab to the detail panel in correct order when there are other clickable items in columns


sshty2018:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When we add a detail panel and there are clickable items in other columns, user cannot tab to the detail panel in the correct order.

### Expected behavior 🤔

Detail Panel should get focus in the correct order

### Steps to reproduce 🕹

https://codesandbox.io/s/keen-rumple-3sqwdg?file=/src/App.tsx

1. Tab to a row in the above example and hit CTRL+Enter to open the detail panel. 
3. Hit Tab. The focus is now on column with "Show All" link
4. Hit Tab again. The focus is now on the "Show All" link of the next row.
5. Detail Panel gets focus only after tabbing through all the links.

Is there any workaround for this?

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

alexfauquette:
It sounds like the focus order is broken because of the rendering strategy. components are rendered in the following order:
- left pinned columns
- right pinned columns
- unpinned columns
- detail panel

(see https://github.com/alexfauquette/material-ui-x/blob/0c473f4d24d74860e39e2c9c5e936a5fe25487a2/packages/grid/x-data-grid-pro/src/components/DataGridProVirtualScroller.tsx)

So if we put focusable elements in left/right pinned columns, we get the same behavior, focus move along columns instead of rows
https://codesandbox.io/s/winter-rgb-wlxfko?file=/src/App.tsx

To solve this we should refactor the rendering such that each row are added one after the other. About workaround, I only know one way to override focus order: `tab-index` but using this for all the rows would be a nightmare.

I'm wondering if moving with <kbd>Tab</kbd> is the main goal or is it to access the detail panel with keyboard.

For the second option we could imagine that Arrow Down move focus to the detail panel if it is open

alexfauquette:
I get an idea of a dirty fix if really needed. You can override the tab order programmatically as follow:

- listen to `onKeyDown`
- look if the detail panel with same row id exists.
- if yes use `event.preventDefault()` to block focus modification, and set it manually to the detail panel

For this to work, you should also override default focus logic of the first and last focusable element of detail panel

```jsx
function RenderExtraIcon({ row }) {
[...]

PR #4254, opened 2022-03-22
[docs] Explain how to manage focus with `renderCell`

alexfauquette:
Part of #4219 

The goal of this PR is to explain in the doc how to remove focusable elements from the tab sequence if the cell does not have the focus

### DIFF
--- a/docs/data/data-grid/columns/columns.md
+++ b/docs/data/data-grid/columns/columns.md
@@ -196,6 +196,7 @@ const columns: GridColDef[] = [
           color="primary"
           size="small"
           style={{ marginLeft: 16 }}
+          tabIndex={params.hasFocus ? 0 : -1}
         >
           Open
         </Button>
@@ -205,13 +206,17 @@ const columns: GridColDef[] = [
 ];
 ```
 
-{{"demo": "RenderCellGrid.js", "bg": "inline"}}
+{{"demo": "RenderCellGrid.js", "bg": "inline", "defaultCodeOpen": false }}
 
-**Note**: It is recommended to also set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/#exported-cells) the data.
-
-> ⚠️ When using `renderCell` with object cell values
-> remember to handle [sorting](/components/data-grid/sorting/#custom-comparator).
-> Otherwise, sorting won't work.
+> ⚠️ Using `renderCell`, requires paying attention to the following points.
+>
+> - If the value returned `valueGetter` does not correspond to the column `type`, you should:
+>
+>   - handle [sorting](/components/data-grid/sorting/#custom-comparator) by providing `sortComparator` to the column.
+>   - set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/#exported-cells) the data.
+>
+> - According to [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#grid), only one of the focusable elements contained by the grid should be included in the page tab sequence.
+>   To render focusable elements, use `params.hasFocus` to remove them from the tab sequence when the cell does not have the focus.
 
 #### Render edit cell
 
--- a/docs/data/data-grid/columns/RenderCellGrid.js
+++ b/docs/data/data-grid/columns/RenderCellGrid.js
@@ -15,6 +15,14 @@ const columns = [
           color="primary"
           size="small"
           style={{ marginLeft: 16 }}
+          // Remove button from tab sequence when cell does not have focus
+          tabIndex={params.hasFocus ? 0 : -1}
+          onKeyDown={(event) => {
+            if (event.key === ' ') {
+              // Prevent key navigation when focus is on button
+              event.stopPropagation();
+            }
+          }}
         >
           Open
         </Button>
--- a/docs/data/data-grid/columns/RenderCellGrid.tsx
+++ b/docs/data/data-grid/columns/RenderCellGrid.tsx
@@ -15,6 +15,14 @@ const columns: GridColDef[] = [
           color="primary"
           size="small"
           style={{ marginLeft: 16 }}
+          // Remove button from tab sequence when cell does not have focus
+          tabIndex={params.hasFocus ? 0 : -1}
+          onKeyDown={(event) => {
+            if (event.key === ' ') {
+              // Prevent key navigation when focus is on button
+              event.stopPropagation();
+            }
+          }}
         >
           Open
         </Button>


docs/data/data-grid/columns/columns.md
```diff
@@ -205,13 +206,17 @@ const columns: GridColDef[] = [
...
-> remember to handle [sorting](/components/data-grid/sorting/#custom-comparator).
-> Otherwise, sorting won't work.
+> ⚠️ Using `renderCell`, requires paying attention to the following points.
+>
+> - If the value returned `valueGetter` does not correspond to the column `type`, you should:
+>
+>   - handle [sorting](/components/data-grid/sorting/#custom-comparator) by providing `sortComparator` to the column.
+>   - set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/#exported-cells) the data.
+>
+> - According to [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#grid), only one of the focusable elements contained by the grid should be included in the page tab sequence.
```
m4theushw, line 218: Move to the accessibility page? I think we'll have other sections there about #4219

docs/data/data-grid/columns/columns.md
```diff
@@ -205,13 +206,17 @@ const columns: GridColDef[] = [
...
+{{"demo": "RenderCellGrid.js", "bg": "inline", "defaultCodeOpen": false }}
 
-**Note**: It is recommended to also set a `valueFormatter` providing a representation for the value to be used when [exporting](/components/data-grid/export/#exported-cells) the data.
-
-> ⚠️ When using `renderCell` with object cell values
-> remember to handle [sorting](/components/data-grid/sorting/#custom-comparator).
-> Otherwise, sorting won't work.
+> ⚠️ Using `renderCell`, requires paying attention to the following points.
+>
+> - If the value returned `valueGetter` does not correspond to the column `type`, you should:
```
m4theushw, line 213:
```suggestion
> - If the type of the value returned by `valueGetter` does not correspond to the column's `type`, you should:
```

m4theushw (commented):
Could you also stop the ripple when the cell loses focus?

image

https://github.com/mui/mui-x/blob/249969c611a2d89ff895bdb639551b6dae0d5e3d/packages/grid/x-data-grid/src/components/columnSelection/GridCellCheckboxRenderer.tsx#L68-L76

docs/data/data-grid/columns/columns.md
alexfauquette: Done

docs/data/data-grid/accessibility/FocusManagement.tsx
```diff
@@ -0,0 +1,51 @@
...
+];
+const wrongColumns: GridColumns = [
+  { field: 'link', renderCell: WrongRenderLink, width: 200 },
+];
+
+const rows: GridRowsProp = [{ id: 1 }, { id: 2 }, { id: 3 }, { id: 4 }, { id: 5 }];
+
+export default function FocusManagement() {
+  return (
+    <div style={{ width: '100%', display: 'flex', justifyContent: 'space-between' }}>
```
m4theushw, line 36: Selecting one cell breaks the demo.

image

Maybe split into two demos so the focus reset button can be used:

image

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
 
 The grid responds to keyboard interactions from the user and emits events when key presses happen on the grid cells.
 
+### Tab sequence
+
+According to [WCAG](https://www.w3.org/TR/wai-aria-practices-1.1/#grid), only one of the focusable elements contained by the grid should be included in the page tab sequence.
+For an element to be included in the tab sequence it needs to have a `tab-index` value of zero or greater.
```
m4theushw, line 51: We can treat it as a prop.

```suggestion
For an element to be included in the tab sequence, it needs to have a `tabIndex` value of zero or greater.
```

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
...
+When a cell of the data grid is focused, the first element with the attribute `tab-index=0` will receive the focus.
+If there is no element with `tab-index=0` the focus is set on the cell itself.
+
+In the example below, the second grid does not remove links from the tab sequence, which implies having to pass through all the links before accessing the pagination controls.
+This behavior makes it complicated to navigate between elements when using large datasets.
+
+{{"demo": "FocusManagement.js", "bg": "inline", "defaultCodeOpen": false}}
+
+If you are customizing cell rendering with [`renderCell`](/x/react-data-grid/columns/#render-cell) method, you become responsible for removing focusable elements from the page tab sequence.
+To do so, use the `tabIndex` prop passed to the `renderCell` params, to know if the rendered cell has the focus, and so if the elements should be removed from the tab sequence.
```
m4theushw, line 62:
```suggestion
To do so, use the `tabIndex` prop passed to the `renderCell` params to know if the rendered cell has focus and if the element should be removed from the tab sequence.
```

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
 
 The grid responds to keyboard interactions from the user and emits events when key presses happen on the grid cells.
 
+### Tab sequence
+
+According to [WCAG](https://www.w3.org/TR/wai-aria-practices-1.1/#grid), only one of the focusable elements contained by the grid should be included in the page tab sequence.
+For an element to be included in the tab sequence it needs to have a `tab-index` value of zero or greater.
+
+When a cell of the data grid is focused, the first element with the attribute `tab-index=0` will receive the focus.
```
m4theushw, line 53: Not sure if "inner element" makes sense but I want to say that is the first element INSIDE the focused cell.

```suggestion
When a cell of the grid is focused, the first inner element with `tabIndex=0` will receive the focus.
```

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
...
+If there is no element with `tab-index=0` the focus is set on the cell itself.
```
m4theushw, line 54:
```suggestion
If there is no element with `tabIndex=0`, the focus is set on the cell itself.
```

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
...
+
+In the example below, the second grid does not remove links from the tab sequence, which implies having to pass through all the links before accessing the pagination controls.
```
m4theushw, line 56: Make it the first? So we first show the wrong and them the correct approach.

```suggestion
In the example below, the first grid does not remove links from the tab sequence, which implies having to pass through all the links before accessing the pagination controls.
```

docs/data/data-grid/accessibility/accessibility.md
```diff
@@ -45,6 +45,32 @@ The user can override this setting with the toolbar density selector if provided
...
+This behavior makes it complicated to navigate between elements when using large datasets.
+
+{{"demo": "FocusManagement.js", "bg": "inline", "defaultCodeOpen": false}}
+
+If you are customizing cell rendering with [`renderCell`](/x/react-data-grid/columns/#render-cell) method, you become responsible for removing focusable elements from the page tab sequence.
```
m4theushw, line 61:
```suggestion
If you are customizing cell rendering with the [`renderCell`](/x/react-data-grid/columns/#render-cell) method, you become responsible for removing focusable elements from the page tab sequence.
```

docs/data/data-grid/accessibility/FocusManagement.tsx
m4theushw: I'm not very sure if we should keep two grids inside the same demo. The "reset focus" button is very useful to test the accessibility here. If I click it the focus goes to the first grid.

m4theushw (commented):
Once #4600 is merged we can add a link to the newly added section: https://github.com/mui/mui-x/pull/4600#discussion_r856245602

docs/data/data-grid/column-definition/RenderCellGrid.tsx
```diff
@@ -1,25 +1,55 @@
 import * as React from 'react';
 import Button from '@mui/material/Button';
+import { TouchRippleActions } from '@mui/material/ButtonBase/TouchRipple';
```
oliviertassinari, line 3: This is a private import, it's not allowed 🚫, we will need to change this. More details in https://mui.com/material-ui/guides/minimizing-bundle-size/

PR #4257, opened 2022-03-22
[docs] Add interactie example for row selectors

### LINKED ISSUE #3717: Create a doc example to play with the various row selectors
### Duplicates

### Latest version

### Summary 💡

Add a playground in the doc where users can plan with the filtering / sorting / pagination etc... and see the impact of the various row-based selectors.

https://codesandbox.io/s/datagrid-v5-quick-start-forked-dxm4h

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #3717

Add an example that allows testing various row selectors

## To verify

- Is the interaction understandable
- Is there too much information
- Is the demo in the correct page

## Todo

- [ ] improve code snipet styling
- [ ] use new naming
- [ ] add missing configurations

### DIFF
--- a/docs/data/data-grid/state/RowSelectorPlayground.tsx
+++ b/docs/data/data-grid/state/RowSelectorPlayground.tsx
@@ -1,0 +1,550 @@
+import * as React from 'react';
+
+import { debounce } from '@mui/material/utils';
+
+import Box from '@mui/material/Box';
+import Paper from '@mui/material/Paper';
+import FormLabel from '@mui/material/FormLabel';
+import FormControl from '@mui/material/FormControl';
+import FormGroup from '@mui/material/FormGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import Checkbox from '@mui/material/Checkbox';
+import Tabs from '@mui/material/Tabs';
+import Tab from '@mui/material/Tab';
+import List from '@mui/material/List';
+import ListItem from '@mui/material/ListItem';
+import ListItemText from '@mui/material/ListItemText';
+import Slider from '@mui/material/Slider';
+import Typography from '@mui/material/Typography';
+
+import {
+    DataGridPro,
+    GridColumns,
+    GridRowsProp,
+    gridRowIdsSelector,
+    useGridApiRef,
+    GridEvents,
+    GridFilterModel,
+    gridSortedRowIdsSelector,
+    gridFilteredSortedRowIdsSelector,
+    gridVisibleSortedRowIdsSelector,
+    gridVisibleSortedTopLevelRowEntriesSelector,
+    gridPaginatedVisibleSortedGridRowEntriesSelector,
+} from '@mui/x-data-grid-pro';
+
+const getTreeDataPath = (row) => [row.bicycleType, row.color];
+
+const allColumns: GridColumns = [
+    {
+        field: 'bicycleType',
+        headerName: 'Bicycle Type',
+        type: 'singleSelect',
+        valueOptions: ['Road Bike', 'Mountain Bike', 'Touring Bike', 'Folding Bike'],
+        width: 150,
+    },
+    {
+        field: 'color',
+        headerName: 'Color',
+        type: 'singleSelect',
+        valueOptions: ['blue', 'red', 'white', 'black'],
+        width: 100,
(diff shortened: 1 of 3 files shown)


docs/data/data-grid/state/RowSelectorPlayground.tsx
```diff
@@ -0,0 +1,550 @@
...
+        <TabPanel value={tabPanemIndex} index={2}>
+          <Box
+            sx={{
+              pl: 3,
+              backgroundColor: 'black',
+              color: 'white',
+              py: 1,
+            }}
+          >
+            <pre>
```
flaviendelangle, line 540: You have an `HighlightedCode` component in `docs/src/modules/components/HighlightedCode` that could help you

flaviendelangle (commented):
I am not sure it is very comprehensive in its current form (the one I did was clearly no better).

I have some crashes.
Here is one in video, I got another one where the `visibleRowsLookup` is not defined but I can't reproduce it.

https://user-images.githubusercontent.com/3309670/159919908-ed45e2c7-e083-45c7-8ccd-8c1ddc0e2285.mp4

alexfauquette:
@flaviendelangle thanks for the `<HighlightedCode />` advice :)

There is probably too much information in a single demo.  removed the tabular representation, such that the reader can see the selector be updated when clicking on the checkboxes

I'm also hiding the grid behind an accordion to be sure the grid will not take too much attention.

I hope it will encourage readers to click on the checkboxes, to understand how they impact the selector, and only after that play with the grid to verify they understood how the selectors react under each configuration

docs/data/data-grid/state/RowSelectorPlaygroundNoSnap.tsx
```diff
@@ -0,0 +1,511 @@
...
+  {
+    field: 'id',
+    headerName: 'row id',
+    flex: 1,
+  },
+];
+
+const rows: GridRowsProp = [
+  {
+    id: 'Touring-blue-945',
```
m4theushw, line 70: I would have used integer IDs because they fit completely in the column. It's easier to match what will be returned with that is rendered. 

image

m4theushw (commented):
I had a quick look. We definitely need this demo but I'm not sure if we need it now. We don't know yet why someone would want a selector that returns only filtered rows, but not sorted. I guess the general use case is to return the currently visible rows or, at the most, the top level rows / expanded rows. Maybe wait for #4151?

docs/data/data-grid/state/state.md
```diff
@@ -64,6 +64,14 @@ Some selectors are yet to be documented.
 
 {{"demo": "SelectorsNoSnap.js", "bg": "inline", "hideToolbar": true}}
 
+### Row selectors
+
+Rows can be impacted by many features such as sorting, filtering, grouping, or pagination.
+Depending on your use case, you could want to apply the effect of only some features.
+To help you to find the right selector, here is a tool to experiment with row selectors.
```
joserodolfofreitas, line 71:
```suggestion
To help you find the right selector, here is a tool to experiment with different features' combinations.
```

joserodolfofreitas (commented):
Refreshing kind of documentation. I think the interactivity of the example is an excellent way to understand how the features can be combined. Great work!

**Is the interaction understandable?**
Yes. I think it's generally intuitive to play with. Although leaving a selector enabled just to meet with a "no dedicated selector" is not an ideal experience. 
I don't get why the "Apply Filtering", is not only enabled when "Apply Sorting" is enabled for instance. 

**Is there too much information?**
No, I think the amount of information is okay. Btw, good idea to start with a hidden grid.

**Is the demo on the correct page?**
Perhaps we can move the section above the [With useGridSelector](https://deploy-preview-4257--material-ui-x.netlify.app/components/data-grid/state/#with-usegridselector) and its full catalog.

Other comments
* Because the grid on the example starts collapsed, I'd suggest that we add an element that tells the grid is updated. Perhaps making the title "Interactive example" dynamic to show the number of rows displayed (or something else that calls attention to the updates on the grid). 

* It'd be interesting to show the source code for the example or to add a simpler codesandbox with the row selectors in action.

PR #4261, opened 2022-03-23
[core] Rename the "pre-processing" concept "pipe-processing"

flaviendelangle:
Rename every occurrence of `pre-processing` into `pipe-proccessing`


m4theushw (commented):
What this PR is doing? Could you update the description?

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
```diff
@@ -27,38 +27,46 @@ type UntypedStrategyProcessors = {
...
  * Each processor name is part of a strategy group which can only have one active strategy at the time.
+ * For now, there is only one strategy group which is `rowTree`.
  */
 export const useGridStrategyProcessing = (apiRef: React.MutableRefObject<GridApiCommunity>) => {
   const availableStrategies = React.useRef(
     new Map<string, { group: GridStrategyGroup; isAvailable: () => boolean }>(),
   );
   const strategiesCache = React.useRef<{
-    [P in GridStrategyProcessorName]?: { [strategyName: string]: GridStrategyProcessor<P> };
+    [P in GridStrategyProcessorName]?: { [strategyName: string]: GridStrategyProcessor<any> };
```
alexfauquette, line 69: Why change from `GridStrategyProcessor<P>` to `GridStrategyProcessor<any>`?

flaviendelangle:
Isn't the title self-explaining ? I'm fine with putting a description on my PR, but this one would basically be a copy paste of the title.
This PR is renaming the concept of "pre-processing" into "pipe-processing" to be explicit. So any variable with `preprocessor` is renamed.

`useGridPreProcessing` => `useGridPipeProcessing`
`apiRef.current.unstable_applyPreProcessors` => `apiRef.current.unstable_applyPipeProcessors`
`GridEvents.preProcessorChange` => `GridEvents.pipeProcessorChange`
... etc

alexfauquette (approved):
Nice explanation

packages/grid/x-data-grid/src/hooks/core/strategyProcessing/useGridStrategyProcessing.ts
flaviendelangle: To remove the `as` here:

https://github.com/mui/mui-x/pull/4261/files/6d57025804ca832e046ee35ea2221743adc75ceb#diff-181808c9d80007231666dce8bff669fdeb8662aae1030b8315bc30f5ccbf33d3R79

This typing is never really used, I would have preferred it to work but I did not succeed

m4theushw (approved):
> Isn't the title self-explaining ?

No, it contains hidden information. For someone that doesn't have the same context as you, it raises questions. Which are the "pre-processing hooks and methods" renamed? How they were renamed to "pipe-processing"? Reading the explanation we can see that is not only hooks and methods that were renamed, but some interfaces and events. It's good to always describe the PR, even if we think we're saying the obvious.

PR #4265, opened 2022-03-23
[DataGrid] Add onPanelClose / onPanelOpen event

### LINKED ISSUE #3568: Filter - onClose
### Duplicates

### Latest version

### Summary 💡

How can we listen to the onClose of the Grid filter ?

### Examples 🌈

_No response_

### Motivation 🔦

When the filtering is done in the server and to reduce the number of server calls (indeed we can use unbounce). We could plug the server call to the close of the filter

### Order ID 💳 (optional)

#19458

kyeongsoosoo:
Closes #3568 

Added `onPanelClose` and `onPanelOpen` event.
Event is triggered when the preferencePanel is closed/opened.
Parameter passes `openedPanelValue`, so the user can choose which panel to trigger an event(`filter | column | both`).

I tested it by adding event stories.
It's my first PR handling `DataGrid` logic. If there is something that can be improved, please let me know.

https://user-images.githubusercontent.com/59413880/159771252-1228f226-33a3-41b9-b7ca-df21e19729bf.mov


packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
```diff
@@ -25,6 +25,7 @@ export interface GridPanelProps
    */
   classes?: Partial<GridPanelClasses>;
   open: boolean;
+  onPanelClose?: () => void;
```
m4theushw, line 28: "panel" is not necessary because all props refer to `GridPanel`.

```suggestion
  onClose?: () => void;
```

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -589,6 +589,21 @@ export interface DataGridPropsWithoutDefaultValue extends CommonProps {
    * @param {GridCallbackDetails} details Additional details for this callback.
    */
   onPageSizeChange?: (pageSize: number, details: GridCallbackDetails) => void;
+
```
m4theushw, line 592:
```suggestion
```

packages/grid/x-data-grid/src/components/panel/GridPreferencesPanel.tsx
```diff
@@ -21,11 +22,31 @@ export const GridPreferencesPanel = React.forwardRef<
     preferencePanelState.openedPanelValue ?? GridPreferencePanelsValue.filters,
   );
 
+  const handleClose = React.useCallback(() => {
+    apiRef.current.hidePreferences();
+    apiRef.current.publishEvent(GridEvents.panelClose, {
```
m4theushw, line 27: My initial idea was to call an `onClick` here. This prop could be passed doing

```tsx
<DataGrid componentsProps={{ panel: { onClose: ... } }} />
```

But we can add the events if the other reviewers also agree. I would only name them as `preferencePanelOpen` and `preferencePanelClose` (rename the props too) to match the component and fire them inside 

https://github.com/mui/mui-x/blob/e66404d4b15c5bfc524cfb6ab1b39e013cd91167/packages/grid/x-data-grid/src/hooks/features/preferencesPanel/useGridPreferencesPanel.ts#L30

and

https://github.com/mui/mui-x/blob/e66404d4b15c5bfc524cfb6ab1b39e013cd91167/packages/grid/x-data-grid/src/hooks/features/preferencesPanel/useGridPreferencesPanel.ts#L50

m4theushw (commented):
Thanks for the contribution! You can fix the CircleCI task by running `yarn docs:api` and committing the changes.

packages/grid/x-data-grid/src/components/panel/GridPreferencesPanel.tsx
kyeongsoosoo: Thanks for the review. I'll update this after other reviewers comment!

kyeongsoosoo: I updated the PR following the review. Thanks for the suggestions.

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -589,6 +589,21 @@ export interface DataGridPropsWithoutDefaultValue extends CommonProps {
...
+  /**
+   * Callback fired when a preferencePanel is closed.
```
alexfauquette, line 594:
```suggestion
  onPageSizeChange?: (pageSize: number, details: GridCallbackDetails) => void;
  /**
   * Callback fired when a preferencePanel is closed.
```

alexfauquette (approved):
Looks good 👍

--- 
Small comment not directly related to this PR code. For now, we can not have two preference panels open at the same time.

> Parameter passes `openedPanelValue`, so the user can choose which panel to trigger an event(`filter | column | both`).

The panel content logic is the [following](https://github.com/alexfauquette/material-ui-x/blob/row-selector-playground/packages/grid/x-data-grid/src/components/panel/GridPreferencesPanel.tsx#L17:L21) 

```js
const isColumnsTabOpen = preferencePanelState.openedPanelValue === GridPreferencePanelsValue.columns;
const isFiltersTabOpen = !preferencePanelState.openedPanelValue || !isColumnsTabOpen;
```

So if the parameter is `"column"` it renders the column pannel. Otherwise, it will render the filter panel.

packages/grid/x-data-grid/src/models/props/DataGridProps.ts
```diff
@@ -590,6 +590,20 @@ export interface DataGridPropsWithoutDefaultValue<R extends GridValidRowModel =
...
   onPageSizeChange?: (pageSize: number, details: GridCallbackDetails) => void;
+  /**
+   * Callback fired when a preferencePanel is closed.
+   * @param {GridPreferencePanelParams} params With all properties from [[GridPreferencePanelParams]].
+   * @param {MuiEvent<{}>} event The event object.
+   * @param {GridCallbackDetails} details Additional details for this callback.
+   */
+  onPreferencePanelClose?: GridEventListener<GridEvents.preferencePanelClose>;
+  /**
+   * Callback fired when a preferencePanel is opened.
```
m4theushw, line 601:
```suggestion
   * Callback fired when the preferences panel is opened.
```

kyeongsoosoo:
@m4theushw I didn't touch any logic of `SecondsTimePicker`, but I'm failing the argos test. Should I fix something??

flaviendelangle:
That one is interesting, the snapshot seems to be unstable with the time...
I will have a look, I approved the Argos change to avoid blocking this PR

m4theushw (approved):
Thanks!

PR #4267, opened 2022-03-23
[DataGrid] Add `onProcessRowUpdateError` prop

m4theushw:
Add prop to not swallow errors thrown by `processRowUpdate`.

1. Go to https://mui.com/components/data-grid/editing/#persistence
2. Change the name to ""
3. Press <kbd>Enter</kbd>

image


packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts
```diff
@@ -290,18 +294,30 @@ export const useGridCellEditing = (
...
+          Promise.resolve(processRowUpdate(rowUpdate, row))
+            .then((finalRowUpdate) => {
+              apiRef.current.updateRows([finalRowUpdate]);
+              updateFocusedCellIfNeeded();
+              updateOrDeleteFieldState(id, field, null);
+            })
+            .catch(handleError);
+        } catch (errorThrown) {
+          handleError(errorThrown);
+        }
```
alexfauquette, line 313: In terms of DX I'm wondering if the error should propagate when `onProcessRowUpdateError` is not set.

I imagine a dev trying to add serverinteraction which does not work due to an error in `processRowUpdate`, and they do not received it because it's catch by a function doing nothing

m4theushw: I added a `console.error` when `onProcessRowUpdateError` is missing. If we propagate the error and it's thrown inside the promise, then the error boundary won't catch it. The developer has to look at the console.

alexfauquette (approved):

PR #4271, opened 2022-03-24
[l10n] Improve Danish (da-DK) locale

simplenotezy:
- Added translations for missing strings
- Swapped `filterOperatorIsEmpty` and `filterOperatorIsNotEmpty` as they were misplaced to the opposite value
- Changed "1 count" to "One count" as it is a general rule of thumb in Danish that numbers below 10 should be written with letters


m4theushw (approved):
Thanks!

PR #4273, opened 2022-03-24
[DataGrid] Add missing classes on `gridClasses` and `gridPanelClasses`

flaviendelangle:
- [x] Add missing classes on `gridClasses` that were listed in the array given to `generateUtilityClasses` but not on the `GridClasses` interface
- [x] Rename `GridPanelClasses['root']` into `GridPanelClasses['panel']` (the typing was wrong, I kept the actual classes used in the code to avoid breaking change)
- [x] Add a generic to `generateUtilityClasses` for both `gridClasses` and `gridPanelClasses` to have TS complains when the interface and the array are not equivalents

### DIFF
--- a/packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
@@ -12,7 +12,7 @@ import { isEscapeKey } from '../../utils/keyboardUtils';
 
 export interface GridPanelClasses {
   /** Styles applied to the root element. */
-  root: string;
+  panel: string;
   /** Styles applied to the paper element. */
   paper: string;
 }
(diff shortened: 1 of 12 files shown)


packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
```diff
@@ -27,7 +27,10 @@ export interface GridPanelProps
   open: boolean;
 }
 
-export const gridPanelClasses = generateUtilityClasses('MuiDataGrid', ['panel', 'paper']);
+export const gridPanelClasses = generateUtilityClasses<keyof GridPanelClasses>('MuiDataGrid', [
```
flaviendelangle, line 30: Adding the generic to make sure that TS do not accept adding a class here but not on the interface

packages/grid/x-data-grid/src/constants/gridClasses.ts
```diff
@@ -382,7 +394,7 @@ export function getDataGridUtilityClass(slot: string): string {
   return generateUtilityClass('MuiDataGrid', slot);
 }
 
-export const gridClasses = generateUtilityClasses('MuiDataGrid', [
+export const gridClasses = generateUtilityClasses<GridClassKey>('MuiDataGrid', [
```
flaviendelangle, line 397: Adding the generic to make sure that TS do not accept adding a class here but not on the interface

flaviendelangle:
> Do you know why this element does not use gridClasse like the other elements? Should it be changed in v6?

No idea no :grimacing:

PR #4276, opened 2022-03-24
[DataGrid] Add slot for filter icon button

### LINKED ISSUE #4255: [question] How to customise DataGrid Filter Panel to only show filters applied to current column
### Order ID 💳

#27566

### Duplicates

### Latest version

### The problem in depth 🔍

When a user clicks the Filter icon on a column header, I would like to display a list of filters that are applied to that column only. Currently it shows all filters applied to the grid.

I can override the panel using the `components` prop, but please can you advise how I can determine which column they've opened it against? Can I get access to the field name somehow?

### Your environment 🌎

<details>

```
  System:
    OS: Windows 10 10.0.17134
  Binaries:
    Node: 14.19.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
  Browsers:
    Chrome: 99.0.4844.51
    Edge: Spartan (42.17134.2208.0)
  npmPackages:
    @mui/base:  5.0.0-alpha.66
    @mui/icons-material: ^5.3.1 => 5.3.1
    @mui/lab: ^5.0.0-alpha.66 => 5.0.0-alpha.66
    @mui/material: ^5.3.1 => 5.3.1
    @mui/private-theming:  5.3.0
    @mui/styled-engine:  5.3.0
    @mui/styles: ^5.3.0 => 5.3.0
    @mui/system: ^5.2.8 => 5.3.0
    @mui/types:  7.1.0
    @mui/utils:  5.3.0
    @mui/x-data-grid-pro: ^5.5.1 => 5.5.1
    @mui/x-license-pro:  5.5.1
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.5.5 => 4.5.5
```
</details>

m4theushw:
Fixes #4255


DanailH:
I thought we don't do slots for icons? If we do one here should we also add slots for all the other icons?

alexfauquette (commented):
About the naming what do you think about

```diff
-ColumnHeaderFilterIcon
+ColumnHeaderFilterIconButton
```

My concern is about differentiating the new slot from `ColumnFilteredIcon`. From naming, it's, complicated to understand which one is about the `Icon` or the `IconButton`

m4theushw:
> I thought we don't do slots for icons? If we do one here should we also add slots for all the other icons?

@DanailH We do slots for icons. This PR is adding a slot for the button that uses the icon. Only using the slot icon won't work for the use case in #4255 because clicking on the borders won't call `onClick`.

> About the naming what do you think about

@alexfauquette Done! I also renamed `GridColumnHeaderFilterIcon` to match the slot name.

> My concern is about differentiating the new slot from ColumnFilteredIcon

In theory, the slot for the icon doesn't mention where the icon is used, only what is the icon. `GridCheckIcon` -> `BooleanCellTrueIcon`.

alexfauquette (approved):

Issue #4275, opened 2022-03-24
[data-grid] Clicking header and scrolling causes headers to skew


aleppek:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

In a DataGrid with a somewhat large number of columns, if you scroll all the way to the right, click on a column header (not the menu or the sort icon, but the label), then scroll back left, all the column headers will be skewed left relative to their columns by a significant amount. 

### Expected behavior 🤔

The column headers should always stay aligned with their columns. 

### Steps to reproduce 🕹

Please see this CodeSandbox: https://codesandbox.io/s/nice-frost-gt209k?file=/src/App.js

Steps:

1. Scroll all the way right.
2. Click the label of the final column.
3. Scroll back left all the way. 
4. Observe that the column headers have skewed left. 
5. If you can't see it at first, try adding a few more columns. It seems to be related to how much scrolling needs to be done, so it's related to both the number of columns and the width of the view.

### Context 🔦

This is being used in a commercial product, where the users won't always see a large number of columns, but sometimes will. It is primarily a user experience issue. 

### Your environment 🌎

  System:
    OS: Linux 5.13 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i9-10980HK CPU @ 2.40GHz
    Memory: 1.50 GB / 15.63 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Managers:
    Apt: 2.0.6 - /usr/bin/apt
    Gradle: 6.8 - ~/.sdkman/candidates/gradle/current/bin/gradle
    pip3: 20.3.3 - /usr/local/bin/pip3
  Utilities:
    Make: 4.2.1
    GCC: 9.4.0
    Git: 2.25.1
  Servers:
    Nginx: 1.18.0
  Virtualization:
    Docker: 20.10.13 - /usr/bin/docker
  IDEs:
    Nano: 4.8
    Vim: 8.1
  Languages:
    Bash: 5.0.17 - /usr/bin/bash
    Java: 11.0.14
    Perl: 5.30.0 - /usr/bin/perl
    Python3: 3.8.10
  Databases:
    MySQL: 10.3.34 (MariaDB) - /usr/bin/mysql
    SQLite: 3.31.1 - /usr/bin/sqlite3
  Browsers:
    Chrome: 99.0.4844.82 -> This is the browser used.
[...]

PR #4280, opened 2022-03-25
[DataGrid] Prevent column header scroll

m4theushw:
Fixes 1st bug from https://github.com/mui/mui-x/issues/4275#issuecomment-1079451619

Same solution from #3718 and #2558

Before: https://codesandbox.io/s/intelligent-pasteur-5etmhd?file=/src/App.js
After: https://codesandbox.io/s/immutable-star-qb78zz?file=/src/App.js


packages/grid/x-data-grid/src/components/columnHeaders/GridColumnHeaderItem.tsx
```diff
@@ -201,13 +202,15 @@ function GridColumnHeaderItem(props: GridColumnHeaderItemProps) {
     const columnMenuState = apiRef.current.state.columnMenu;
     if (hasFocus && !columnMenuState.open) {
       const focusableElement = headerCellRef.current!.querySelector<HTMLElement>('[tabindex="0"]');
-      if (focusableElement) {
-        focusableElement!.focus();
-      } else {
-        headerCellRef.current!.focus();
+      const elementToFocus = focusableElement || headerCellRef.current;
+      elementToFocus?.focus({ preventScroll: true });
```
oliviertassinari, line 206: AFAIK this `preventScroll` is not viable because it doesn't work on Safari. What I would propose instead is to look into why the container is scrollable in the first place. We use translate 3d, so why is it possible for `scrollLeft` to not always be 0 in the first place?

m4theushw: Even adding `overflow: hidden` I still can scroll a div. It's expected: https://stackoverflow.com/questions/33703204/is-scrolltop-and-scrollleft-for-overflow-hidden-elements-reliable

msedge_0wKPMSGImu

In the past,  we even did a small hack by listening to `scroll` and calling `scrollLeft=0` to prevent scrolling the container - which had `overflow: hidden` - when using a touchpad device.

https://github.com/mui/mui-x/blob/b81daef3a604c87620d3cfcb354cf5071d3836a9/packages/grid/_modules_/grid/hooks/features/virtualization/useGridVirtualRows.ts#L487-L492

> AFAIK this preventScroll is not viable because it doesn't work on Safari.

That's way I call `scrolLeft=0` down below if `preventScroll` is not supported.

oliviertassinari: Right ok, great points.

oliviertassinari: Actually, I really like this part:

apiRef.current.columnHeadersContainerElementRef!.current!.scrollLeft = 0;

alexfauquette: The "no row" bug can be fixed. It's an early return to modify in scroll hook

https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/hooks/features/scroll/useGridScroll.ts#L56:L58

---

Is it necessary to prevent scroll? I thought adding the `hasFocus` un the `useLayoutEffect` dependencies was enough to prevent scroll on rendering

The only bug I can think of with the scroll on rendering is when the element is unmounted because it goes out of the virtualization engine and comes back. But that's a larger problem.

For example, I set focus on a cell, I can use keyboard navigation as long as the cell is in the virtualisation buffer. As soon as it exits the buffer, the focus moves to `<body>` and the gird does not respond anymore to keyboard interactions
https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/hooks/features/scroll/useGridScroll.ts#L56:L58

m4theushw: I removed `preventScroll` and kept only `scrollLeft=0`. I didn't realize the side effect of using it.

> I thought adding the hasFocus un the useLayoutEffect dependencies was enough to prevent scroll on rendering

It's not sufficient. If I remove `scrollLeft=0` and scroll very fast to left and right the bug appears at some point. This happens because the column header component may be mounted during scroll.

> The "no row" bug can be fixed. It's an early return to modify in scroll hook

Interesting, I won't fix it here. Feel free to submit a PR.

> Now, maybe we don't need to care. Who would try to scroll/use the arrow key? There are no data loaded 😁

If what caused "no results" was applying a filter in a column that's outside the visible area, then the user may use the arrow keys to navigate to them. We had a issue about that when we blocked the scrollbar if `rows=[]`: https://github.com/mui/mui-x/issues/3795#issuecomment-1025628771

packages/grid/x-data-grid/src/hooks/features/columnHeaders/useGridColumnHeaders.tsx
```diff
@@ -171,7 +171,7 @@ export const useGridColumnHeaders = (props: UseGridColumnHeadersProps) => {
 
       columns.push(
         <GridColumnHeaderItem
-          key={i}
+          key={column.field}
```
m4theushw, line 174: Prevents the effect triggering twice.

PR #4283, opened 2022-03-26
[core] Remove dead code

oliviertassinari:
I have found this by looking into #4280 in more detail. This code path is never reached, it seemed to have been here for when we implemented the virtualization without using a native scroll container, so we had to reset the scrollLeft and scrollRight. This is no longer the case.


m4theushw (approved):
Yeah, it's legacy. We can also remove the following ref now: https://github.com/mui/mui-x/blob/ed594d76258c68fb8691c38f1bb5d6d93b3e53ca/packages/grid/x-data-grid/src/components/base/GridBody.tsx#L64

PR #4302, opened 2022-03-29
[DataGrid] Allows to use keyboard navigation even with no rows

alexfauquette:
Fix the following bug: https://github.com/mui/mui-x/pull/4280#discussion_r835798116

- Go in codesandbox.io/s/datagridprodemo-material-demo-forked-8e4l8w
- Click on a column header
- Press <kbd>Arrow Right</kbd> to navigate to the last column
- You are not able to do it because there is no row in the grid

### DIFF
--- a/packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
+++ b/packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
@@ -319,6 +319,25 @@ describe('<DataGrid /> - Keyboard', () => {
       expect(virtualScroller.scrollLeft).not.to.equal(0);
     });
 
+    it('should scroll horizontally when navigating between column headers with arrows even if rows are empty', function test() {
+      if (isJSDOM) {
+        // Need layouting for column virtualization
+        this.skip();
+      }
+      render(
+        <div style={{ width: 60, height: 300 }}>
+          <DataGrid autoHeight={isJSDOM} {...getData(10, 10)} rows={[]} />
+        </div>,
+      );
+      getColumnHeaderCell(0).focus();
+      const virtualScroller = document.querySelector(
+        '.MuiDataGrid-virtualScroller',
+      )! as HTMLElement;
+      expect(virtualScroller.scrollLeft).to.equal(0);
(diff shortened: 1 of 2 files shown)


packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
```diff
@@ -319,6 +319,25 @@ describe('<DataGrid /> - Keyboard', () => {
...
+        <div style={{ width: 60, height: 300 }}>
+          <DataGrid autoHeight={isJSDOM} {...getData(10, 10)} rows={[]} />
+        </div>,
+      );
+      getColumnHeaderCell(0).focus();
+      const virtualScroller = document.querySelector(
+        '.MuiDataGrid-virtualScroller',
+      )! as HTMLElement;
+      expect(virtualScroller.scrollLeft).to.equal(0);
+      fireEvent.keyDown(document.activeElement!, { key: 'ArrowRight' });
```
m4theushw, line 337: To able to re-enable the ESLint rule in the future.

```suggestion
      fireEvent.keyDown(getColumnHeaderCell(0), { key: 'ArrowRight' });
```

m4theushw (approved):

packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx
alexfauquette: @m4theushw are you talking about the following line?
```ts
// @ts-ignore Remove once the test utils are typed
```

The file is full of those `document.activeElement!` should I remove them?

alexfauquette: I merged because it's not related to the bug. If moving from `document.activeElement` to getter methods, I will do a distinct PR cleaning the all file

m4theushw: I was talking about https://github.com/mui/mui-x/blob/df6be87ee4e0438d6ec8465977a19299953d17ec/packages/grid/x-data-grid/src/tests/keyboard.DataGrid.test.tsx#L64

It's good to not use `document.activeElement` in new tests. In the future there will be less tests to change.

PR #4306, opened 2022-03-29
[core] Remove unused events.json on old location

flaviendelangle:


alexfauquette:
They are unused for documentation generation, but are they used in PR review to check no event is removed or modified like with `x-data-grid.exports.json`?

flaviendelangle:
There is the exact same file [here](https://github.com/mui/mui-x/blob/master/docs/data/data-grid/events/events.json)
This PR is only removing the old version.

alexfauquette:
Do you know why tests are failing?

PR #4311, opened 2022-03-30
[data grid] Fix column dimensions import/export with flex and resizing

### LINKED ISSUE #4310: [data grid] Saving and retrieving state does not work with fluid resizable columns
### Order ID 💳

36792

### Duplicates

### Latest version

### The problem in depth 🔍

Hi! Here is a summary:
- I have a fluid width DataGrid, where almost all columns have flex property in order to fill the full width of the parent.
- I'd like to use `apiRef.current.exportState()` to save the user's layout changes into storage. This includes: changing column width, reordering columns, and hiding columns.
- I'd like to use the `initialState` prop to restore the previously saved state on a component mount. The idea is to render DataGrid that remembers your customizations since the last time.
----
For some reason, it does not work as expected with my setup:
 1.  Hidden columns are not respected when restoring the state. 
hide

 2. Resized columns width appears broken after state restoration.
resize

The order of the columns is respected though.

Please help me determine am I hitting limitations of the `apiRef.current.exportState()`/`initialState` in this case, or it is something else.

Here is a demo sandbox: https://codesandbox.io/s/mui-datagrid-retrieve-state-issue-z6sokq?file=/demo.js

### Your environment 🌎

_No response_

flaviendelangle:
Fixes #4310

This PR fixes two small bugs raised in #4310

- When a user resizes a column, we set `flex: undefined` on this column. This is to make sure that the column is not automatically resized when we re-run the sizing process of all columns (when a new column is created for instance). But when saving the state, this `flex: undefined` is not stored in the JSON, so when restoring, the columns has flex again.

Solution: Set `flex: 0` instead of `flex: undefined`

- The default `maxWidth` property equals `Infinity`. But `JSON.stringify(Infinity) === 'null'`. So when restoring it, we have `maxWidth: null`, which is handled like `maxWidth: 0`.

It was raised by @m4theushw in https://github.com/mui/mui-x/pull/3816#discussion_r806158304, I thought `maxWidth: null` would work but I was clearly wrong.

Solution: Set `maxWidth: 'Infinity'` in the export and parse in correctly on restore.


packages/grid/x-data-grid-pro/src/hooks/features/columnResize/useGridColumnResize.tsx
```diff
@@ -136,7 +136,7 @@ export const useGridColumnResize = (
 
     colDefRef.current!.computedWidth = newWidth;
     colDefRef.current!.width = newWidth;
-    colDefRef.current!.flex = undefined;
+    colDefRef.current!.flex = 0;
```
m4theushw, line 139: This change doesn't seem necessary to fix #4310 but makes sense to set `0`.

m4theushw (commented):
Wouldn't be easier to not export values with `Infinity` or `null`? We can use the default values when restoring.

flaviendelangle:
@m4theushw 

> Wouldn't be easier to not export values with Infinity or null? We can use the default values when restoring.

It was my 1st idea
But it does not support one edge case:

- Create a custom column type with a `maxWidth: 200`
- Create a column with this column type and `maxWidth: Infinity`
- Export the column
- Re-import the column => you have `maxWidth: 200` instead of `maxWidth: Infinity`

packages/grid/x-data-grid-pro/src/hooks/features/columnResize/useGridColumnResize.tsx
flaviendelangle: When I did not do this one, I had the flex coming back on re-import.
Maybe it only occurs in some scenario.

As `0` make sense, I propose to keep it.

m4theushw:
> But it does not support one edge case:

We could export `Infinity` as `-1` instead. This would avoid to have to update `GridColumnDimensions`.

flaviendelangle:
> We could export Infinity as -1 instead

When a column definition have `maxWidth: -1`, would you set `maxWidth: Infinity` before storing it in the state
Or would you change the of the functions using `maxWidth` (`useGridColumnResize` for instance).

m4theushw:
> Or would you change the of the functions using maxWidth (useGridColumnResize for instance).

I didn't understand this part. But I would always convert `-1` to `Infinity`. Currently, we don't check if the value is negative.

m4theushw:
It's simpler to convert all dimensions that have `-1` to `Infinity` than to check their values to not clamp. So for the question below my answer is yes.

> When a column definition have maxWidth: -1, would you set maxWidth: Infinity before storing it in the state

flaviendelangle:
Done

PR #4315, opened 2022-03-30
[core] Update `x-license-pro` license to handle premium package

DanailH:
I added a new parameter in the license string called `TYPE`. By default, if a type is not provided when generating a license the type is set as `pro` so it can be backward compatible.
As far as I can see with the new license structure existing licenses will still work. Once we have the `DataGridPremium` component then we can check if the `TYPE` exists and if it is `premium`.
I added a couple of tests to handle the new parameter and updated the cli.

**Note**: there was a misspelling of the exported `generateLicence` function so I fixed that as well (not sure if it is a breaking change it should only be used by us internally).


flaviendelangle:
> Note: there was a misspelling of the exported generateLicence function so I fixed that as well (not sure if it is a breaking change it should only be used by us internally).

For me we can consider this lib as internal only

oliviertassinari (commented):
I have tested the following: 1. generate a new license with the code in this PR 2. apply it to the latest npm version 3. it works https://codesandbox.io/s/datagridprodemo-material-demo-forked-ppitu3?file=/demo.tsx it's nicely backward compatible :)

packages/x-license-pro/src/cli/license-cli.ts
```diff
@@ -70,14 +76,17 @@ export function licenseGenCli() {
...
+          type: argv.type as LicenseType,
         };
 
         console.log(
           `Generating new license number for order ${
             licenseDetails.orderNumber
-          } with expiry date ${licenseDetails.expiryDate.toLocaleDateString()}`,
+          } with expiry date ${licenseDetails.expiryDate.toLocaleDateString()} and type "${
+            licenseDetails.type
+          }`,
```
m4theushw, line 87: The quotes were not closed.

```suggestion
          }"`,
```

m4theushw (commented):
Should we increment `KEYVERSION` too? We can derive the version by checking if `TYPE` is present, but in essence the key format was changed.

packages/x-license-pro/src/cli/license-cli.ts
DanailH: Updated to v2.

packages/x-license-pro/src/generateLicense/generateLicense.ts
```diff
@@ -2,18 +2,29 @@ import { md5 } from '../encoding/md5';
...
 
 const licenseVersion = '1';
+
+export type LicenseType = 'pro' | 'premium';
+
 export interface LicenseDetails {
   orderNumber: string;
   expiryDate: Date;
+  // TODO: to be made required once the store is updated
+  type?: LicenseType;
```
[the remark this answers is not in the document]
DanailH: The scope does sound better. I went with `type` because I was thinking in terms of bundling all components but if we want to have per component license then `scope` is much better.

packages/x-license-pro/src/cli/license-cli.ts
```diff
@@ -70,14 +76,17 @@ export function licenseGenCli() {
         const licenseDetails = {
           expiryDate: new Date(new Date().getTime() + parseInt(argv.expiry, 10) * oneDayInMs),
           orderNumber: argv.order,
+          type: argv.type as LicenseScope,
```
oliviertassinari, line 79: I'm not following, why mix the type and scope terminology?

DanailH: My bad, I must have missed renaming it here. I've updated it

packages/x-license-pro/src/generateLicense/generateLicense.ts
```diff
@@ -1,19 +1,30 @@
 import { md5 } from '../encoding/md5';
 import { base64Encode } from '../encoding/base64';
 
-const licenseVersion = '1';
+const licenseVersion = '2';
```
flaviendelangle, line 4: Is there any interest to keep the old algorithm available to be able to generate licenses of version 1 ? 

Something like `generateLicense/generateLicenseV1.ts` with a copy paste of the `generateLicense/generateLicense.ts` as it is before this PR

flaviendelangle (approved):

packages/x-license-pro/src/generateLicense/generateLicense.ts
DanailH: I think the reason was so that we can check the version of the license if we need to do some specific checks but I don't think there is a value in keeping the old algorithm.

PR #4317, opened 2022-03-30
[DataGrid] Add quick filtering engine

### LINKED ISSUE #2842: [data grid] Implement Quick Filter
### Duplicates

### Latest version

### Summary 💡

Upvote this issue if you want it to be implemented natively.

It's worth to be noted that we have a demo showing how to add this functionality to the grid: https://mui.com/components/data-grid/filtering/#quick-filter.

### Examples 🌈

Screenshot 2022-03-28 at 20 17 13

- https://ag-grid.com/javascript-grid-filter-quick/
- https://material-table.com/#/docs/features/search
- https://demos.telerik.com/kendo-ui/grid/search-panel
- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/searching/
- https://ej2.syncfusion.com/react/demos/#/material/grid/searching
- https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/filtering/

### Motivation 🔦

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #2842

The quick filter model I took is the following:

It uses a list of values to filter and it can returns 
- rows that contain all the values (default behavior)
- rows that contain at least one of the values

It's applied after the basic filtering because it makes more comparisons than the basic one.

For now, they both work together (filtered rows must pass both the quick and the basic filtering) This decision is to avoid strange behaviors such as: "When a user set a quick filter value, it reset the filter panel" or "When the quick filter is activated, the filter panel does not have effect"

For now, I added properties in the `filterModel`. I'm not sure if it is the correct way to do that or if filtering should add another model `quickFilterModel`

To support row grouping, the signature of the internal method `shouldApplyFilter` is modified. instead of taking a filter item as an input, it takes the columnField

Demo: https://deploy-preview-4317--material-ui-x.netlify.app/x/react-data-grid/filtering/#quick-filter

## TODO

- [x] Tests
- [x] support `singleSelect` columns
- [x] Evaluate the feasibility of `date` columns 

## Changelog

You can now add a quick filtering search bar to your grid. To so, either pass `showQuickFilter` prop to the `<GridToolbar />` or use the `<GridToolbarQuickFilter />` component in your custom toolbar.

More information about how to customize the filtering logic in the [documentation](https://mui.com/x/react-data-grid/filtering/#quick-filter)

image


packages/grid/x-data-grid/src/models/gridFilterItem.ts
```diff
@@ -28,4 +28,9 @@ enum GridLinkOperator {
   Or = 'or',
 }
 
-export { GridLinkOperator };
+enum GridQuickFilterLogic {
```
flaviendelangle, line 31: Do we need two distinct enums ?

How about naming `quickFilterLogic` => `quickFilterLinkOperator` and keep a single enum ?

packages/grid/x-data-grid/src/hooks/features/filter/useGridFilter.tsx
```diff
@@ -177,6 +192,20 @@ export const useGridFilter = (
     [apiRef],
   );
 
+  const setQuickFilterValues = React.useCallback<GridFilterApi['setQuickFilterValues']>(
+    (values) => {
+      const filterModel = gridFilterModelSelector(apiRef);
+      if (JSON.stringify(filterModel.quickFilterValues) === JSON.stringify(values)) {
```
flaviendelangle, line 198: You can use the `isDeepEqual` function

packages/grid/x-data-grid/src/hooks/features/filter/useGridFilter.tsx
```diff
@@ -78,10 +79,24 @@ export const useGridFilter = (
   const applyFilters = React.useCallback<GridFilterApi['unstable_applyFilters']>(() => {
     apiRef.current.setState((state) => {
       const filterModel = gridFilterModelSelector(state, apiRef.current.instanceId);
-      const isRowMatchingFilters =
+      const isRowMatchingStandardFilters =
```
flaviendelangle, line 82: I propose to rename `buildAggregatedFilterApplier` => `buildAggregatedFilterItemsApplier` and to create a new `buildAggregatedFilterApplier`:

```ts
const buildAggregatedFilterApplier = (filterModel, apiRef) => {
  const isRowMatchingFilterItems = buildAggregatedFilterItemsApplier(filterMode, apiRef);
  const isRowMatchingQuickFilter = buildAggregationQuickFilterApplier(filterModel, apiRef);

  if (isRowMatchingFilterItems == null && isRowMatchingQuickFilter == null) {
    return null
  }

  if (isRowMatchingFilterItems == null) {
    return isRowMatchingQuickFilter
  }

  if (isRowMatchingQuickFilter == null) {
    return isRowMatchingFilterItems
  }
  
  return (rowId, shouldApplyFilter) => isRowMatchingFilterItems(rowId, shouldApplyFilter) && isRowMatchingQuickFilter(rowId, shouldApplyFilter)
}
```

(with the correct params typing of course)

That way, in `useGridFilter` you just have the same code as before.
I find it more readable.

docs/data/data-grid/filtering/QuickFilteringCustomizedGrid.tsx
```diff
@@ -0,0 +1,83 @@
...
+const VISIBLE_FIELDS = ['name', 'rating', 'country', 'dateCreated', 'isAdmin'];
+
+export default function QuickFilteringCustomizedGrid() {
+  const { data } = useDemoData({
+    dataSet: 'Employee',
+    visibleFields: VISIBLE_FIELDS,
+    rowLength: 100,
+  });
+
+  // Otherwise filter will be applied on fields such as the hidden column id
```
flaviendelangle, line 38: I don't understand this comment

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,68 @@ The example below demonstrates how to achieve server-side filtering.
...
+
+```js
+initialState={{
+  filterModel: {
+    quickFilterLogic: GridQuickFilterLogic.Or
+  }
+}}
+```
+
+The quick filter will only consider columns with types `string` or `number`
```
flaviendelangle, line 330: Could it support `singleSelect` ?

And below you talk about a `date` column.
Are `string` and `number` the only supported by default ? If so, maybe this sentence should be reworded.

docs/data/data-grid/filtering/QuickFilteringCustomizedGrid.tsx
alexfauquette: The useDemoData returns by default a column `{field: 'id', hidden: true}`

This column is used in the filtering but not visible. So if you quick filter with value `"45"` it will return some rows because their id contains "45" somewhere in its random characters

I think I will remove this and use the opportunity to show how to deactivate the quick filter on the column `"id"`

docs/data/data-grid/filtering/filtering.md
alexfauquette: Yes, single-select could be supported. I wanted first to validate the global solution before doing this particular case. About the date, there is quite a lot of way to specify a date in a text area

packages/grid/x-data-grid/src/models/gridFilterItem.ts
alexfauquette: I though two enums would be better, because it allows to add logic mode distinct from the logic operator in the future if requested, but not sure we will have such request

packages/grid/x-data-grid/src/models/colDef/gridColDef.ts
```diff
@@ -215,6 +216,19 @@ export interface GridColDef<R extends GridValidRowModel = any, V = any, F = V> {
...
+   * This function can return `null` to skip filtering for this value and column.
+   * @param {any} value The value with which we want to filter the column.
+   * @param {GridStateColDef} colDef The column from which we want to filter the rows.
+   * @param {React.MutableRefObject<GridApiCommunity>} apiRef The API of the grid.
+   * @returns {null | ((params: GridCellParams) => boolean)} The function to call to check if a row pass this filter value or not.
+   */
+  getApplyQuickFilterFn?: (
+    value: any,
+    colDef: GridStateColDef,
+    apiRef: React.MutableRefObject<GridApiCommunity>,
```
m4theushw, line 230: Is there a use case where `colDef` and `apiRef` need to be accessed?

`colDef` could be redundant since the function is defined inside the column definition.

packages/grid/x-data-grid/src/models/gridFilterItem.ts
m4theushw: We can start small and reuse `GridLinkOperator` until we need to support a logic operator exclusively in the quick filter.

packages/grid/x-data-grid/src/models/gridFilterModel.ts
```diff
@@ -15,4 +15,17 @@ export interface GridFilterModel {
    * @default `GridLinkOperator.Or`
    */
   linkOperator?: GridLinkOperator;
+
```
m4theushw, line 18: No blank lines separating the interface members. :)

```suggestion
```

packages/grid/x-data-grid/src/tests/quickFiltering.DataGrid.test.tsx
```diff
@@ -0,0 +1,1095 @@
...
+  GridFilterItem,
+  GridPreferencePanelsValue,
+  GridQuickFilterLogic,
+  GridToolbarQuickFilter,
+} from '@mui/x-data-grid';
+import { getColumnValues } from 'test/utils/helperFn';
+
+const isJSDOM = /jsdom/.test(window.navigator.userAgent);
+
+describe('<DataGrid /> - Filter', () => {
```
m4theushw, line 19:
```suggestion
describe('<DataGrid /> - Quick Filter', () => {
```

packages/grid/x-data-grid/src/tests/quickFiltering.DataGrid.test.tsx
```diff
@@ -0,0 +1,1095 @@
...
+  };
+
+  const TestCase = (props: Partial<DataGridProps>) => {
+    return (
+      <div style={{ width: 300, height: 300 }}>
+        <DataGrid {...baselineProps} components={{ Toolbar: GridToolbarQuickFilter }} {...props} />
+      </div>
+    );
+  };
+  describe('component', () => {
```
m4theushw, line 49:
```suggestion

  describe('component', () => {
```

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const apiRef = useGridApiContext();
+  const rootProps = useGridRootProps();
+
+  const [searchValue, setSearchValue] = React.useState('');
+
+  const handleSearchValueChange = React.useCallback(
+    (event) => {
+      const newSearchValue = event.target.value;
+      setSearchValue(newSearchValue);
+      apiRef.current.setQuickFilterValues(
```
m4theushw, line 46: It needs to be debounced. Otherwise, the grid rerenders on every keystroke.

packages/grid/x-data-grid/src/models/colDef/gridColDef.ts
alexfauquette: The code is taken form how the [`GridFilterInputSingleSelect`](https://github.com/alexfauquette/material-ui-x/blob/quick-filter/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputSingleSelect.tsx#L10:L32) renders options

m4theushw: > The code is taken form how the [GridFilterInputSingleSelect](https://github.com/alexfauquette/material-ui-x/blob/quick-filter/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputSingleSelect.tsx#L10:L32) renders options

This makes me think that maybe `singleSelect` should have a default value formatter.

alexfauquette: @m4theushw I added this to #3287

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
alexfauquette: Done

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }, [apiRef]);
+
+  return (
+    <GridToolbarQuickFilterRoot
+      as={rootProps.components.BaseTextField}
+      variant="standard"
+      value={searchValue}
+      onChange={handleSearchValueChange}
+      placeholder={apiRef.current.getLocaleText('toolbarQuickFilterPlaceholder')}
+      label={apiRef.current.getLocaleText('toolbarQuickFilterLabel')}
```
m4theushw, line 77: I don't think that the repeated label looks good when combining with other buttons.

image
Maybe keep it only for screen readers.

```suggestion
      aria-label={apiRef.current.getLocaleText('toolbarQuickFilterLabel')}
```

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
...
+      role="search"
+      InputProps={{
+        startAdornment: <SearchIcon fontSize="small" />,
+        endAdornment: (
+          <IconButton
+            aria-label={apiRef.current.getLocaleText('toolbarQuickFilterDeleteIconLabel')}
```
m4theushw, line 83: The `title` is sufficient.

```suggestion
```

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
...
+    borderColor: 'divider',
+  },
+}));
+
+const defaultSearchValueParser = (searchText: string) =>
+  searchText.split(' ').filter((word) => word !== '');
+
+export type GridToolbarQuickFilterProps = TextFieldProps & {
+  quickFilterParser?: (input: string) => any[];
+  debounceMs?: number;
```
m4theushw, line 33: Could you fill the descriptions?

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <GridToolbarQuickFilterRoot
+      as={rootProps.components.BaseTextField}
+      variant="standard"
+      value={searchValue}
+      onChange={handleSearchValueChange}
+      placeholder={apiRef.current.getLocaleText('toolbarQuickFilterPlaceholder')}
+      label={apiRef.current.getLocaleText('toolbarQuickFilterLabel')}
+      role="search"
```
m4theushw, line 78: `role="search"` goes to the search container but we don't have one. https://www.w3.org/TR/wai-aria-practices-1.2/examples/landmarks/search.html

```suggestion
      type="search"
```

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
...
+            title={apiRef.current.getLocaleText('toolbarQuickFilterDeleteIconLabel')}
+            size="small"
+            style={{ visibility: searchValue ? 'visible' : 'hidden' }}
+            onClick={clearSearchValue}
+          >
+            <ClearIcon fontSize="small" />
+          </IconButton>
+        ),
+      }}
+      {...other}
```
m4theushw, line 93:
```suggestion
      {...other}
      {...rootProps.componentsProps?.baseTextField}
```

packages/grid/x-data-grid/src/models/gridFilterModel.ts
```diff
@@ -8,11 +8,21 @@ export interface GridFilterModel {
...
+   * values used to quick filter rows
+   * @default `[]`
+   */
+  quickFilterValues?: any[];
+  /**
+   * - `GridLinkOperator.And`: the row must pass all the values.
+   * - `GridLinkOperator.Or`: the row must pass at least one value.
+   * @default `GridLinkOperator.Or`
+   */
+  quickFilterLinkOperator?: GridLinkOperator;
```
m4theushw, line 27: The "link operator" name is confusing. I remember we discussed about renaming to "logic operator". Wouldn't make sense to start using this new name here?

```suggestion
  quickFilterLogicOperator?: GridLinkOperator;
```

I found https://github.com/mui/mui-x/pull/3915#discussion_r806575519

packages/grid/x-data-grid/src/colDef/gridSingleSelectOperators.ts
```diff
@@ -11,6 +13,51 @@ const parseObjectValue = (value: GridFilterItem) => {
...
+      if (optionLabel.slice(0, value.length).toLowerCase() === value.toLowerCase()) {
+        if (!potentialValues.includes(optionValue)) {
+          potentialValues.push(optionValue.toString());
+        }
+      }
+    });
+  }
+
+  return ({ value: columnValue }: GridCellParams): boolean => {
+    return columnValue != null ? potentialValues.includes(columnValue.toString()) : false;
```
m4theushw, line 57: Should the `valueFormatter` be applied to `columnValue`?

In https://deploy-preview-4317--material-ui-x.netlify.app/components/data-grid/filtering/#quick-filter I can't search for country names because the value tested is an object.

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
 
 ## Quick filter
 
-The grid does not natively include quick filtering.
-However, it can be implemented as in the demo below.
+The package also provides a component `<GridToolbarQuickFilter/>` which filters rows without specifying the concerned columns.
```
m4theushw, line 312: I think we could first introduce what quick filter is.

```suggestion
Quick filter allows to filter rows using free text.
To enable it, you can add the `<GridToolbarQuickFilter />` component to the toolbar.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
...
+By default, the quick filter considers the input as a list of values separated by space and keeps only rows that contain all the values.
 
 {{"demo": "QuickFilteringGrid.js", "bg": "inline", "defaultCodeOpen": false}}
 
-> ⚠️ This feature isn't natively implemented in the grid package. It's coming.
->
-> 👍 Upvote [issue #2842](https://github.com/mui/mui-x/issues/2842) if you want to see it land faster.
+### Custom filtering logic
+
+The logic quick filter logic can be switched to filter rows that contain _at least_ one value instead for each of them.
```
m4theushw, line 319:
```suggestion
The logic used for quick filter can be switched to filter rows that contain _at least_ one of the values specified, instead of testing if it contains all of them.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
...
+To do so, set `quickFilterLinkOperator` equals to `GridLinkOperator.Or` as follow:
```
m4theushw, line 320:
```suggestion
To do so, set `quickFilterLinkOperator` to `GridLinkOperator.Or` as follow:
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
...
+
+```js
+initialState={{
+  filterModel: {
+    quickFilterLinkOperator: GridLinkOperator.Or
+  }
+}}
+```
+
+With default settings, the quick filter will only consider columns with types `string`,`number` and `singleSelect`
```
m4theushw, line 330:
```suggestion
With the default settings, quick filter will only consider columns with types `string`, `number` and `singleSelect`.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
...
+
+- For columns with type `string` the cell must contain the value to pass the filter
+- For columns with type `number` the cell must equal the value to pass the filter
+- For columns with type `singleSelect` the cell must start like the value to pass the filter
+
+To modify or add the quick filter operators, add the property `getApplyQuickFilterFn` to the column definition.
+This function is quite similar to `getApplyFilterFn`.
+This function takes as an input a value of the quick filter and returns another function that takes the cell value as an input and returns `true` if it satisfies the operator condition.
+
+In the example bellow, we filter a `date` column by checking if it contains the correct year.
```
m4theushw, line 340: No "we".

```suggestion
In the example below, a custom filter is created for the `date` column to check if it contains the correct year.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,70 @@ The example below demonstrates how to achieve server-side filtering.
...
+    quickFilterLinkOperator: GridLinkOperator.Or
+  }
+}}
+```
+
+With default settings, the quick filter will only consider columns with types `string`,`number` and `singleSelect`
+
+- For columns with type `string` the cell must contain the value to pass the filter
+- For columns with type `number` the cell must equal the value to pass the filter
+- For columns with type `singleSelect` the cell must start like the value to pass the filter
```
m4theushw, line 334: I didn't understand what "start like the value" meant.

```suggestion
- For columns with type `string`, the cell must contain the value to pass the filter
- For columns with type `number`, the cell must be equal to the value to pass the filter
- For columns with type `singleSelect`, the cell must start like the value to pass the filter
```

m4theushw (commented):
What about adding a prop to `GridToolbar` to include  `GridToolbarQuickFilter` as the last item? I think it's easier for developers to reuse the default toolbar than to create a custom one to have the default buttons + quick filter.

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
joserodolfofreitas: +1 to consider not repeating the label.

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
alexfauquette: Regarding accessibility, I observed `aria-label` is applied on the div wrapping `<input/>` and `<button/>` is it a problem?

packages/grid/x-data-grid/src/colDef/gridSingleSelectOperators.ts
alexfauquette: My strategy is to first loop on `valueOptions` to find which option values are accepted. Such that for each row I only check if the value is in "valid" options

I forget to use `parseObjectValue` in the second step to extract `value` property when the cell value is an object. Should be fixed now

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
alexfauquette: I replaced the `label` in the outdated code by `aria-label`

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,72 @@ The example below demonstrates how to achieve server-side filtering.
 
 ## Quick filter
 
-The grid does not natively include quick filtering.
-However, it can be implemented as in the demo below.
+Quick filter allows filtering rows using free text.
```
joserodolfofreitas, line 312:
```suggestion
Quick filter allows filtering rows by multiple columns with a single text input.
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,72 @@ The example below demonstrates how to achieve server-side filtering.
...
+  filterModel: {
+    quickFilterLogicOperator: GridLinkOperator.Or
+  }
+}}
+```
+
+With the default settings, quick filter will only consider columns with types `string`,`number`, and `singleSelect`.
+
+- For columns with type `string`, the cell must contain the value to pass the filter
+- For columns with type `number`, the cell must be equal to the value to pass the filter
```
joserodolfofreitas, line 335:
```suggestion
- For `number` columns, the cell must be **equal** any of the values
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,72 @@ The example below demonstrates how to achieve server-side filtering.
...
+- For columns with type `singleSelect`, the cell must start like the value to pass the filter
```
joserodolfofreitas, line 336:
```suggestion
- For `singleSelect`columns , the cell must **start with** any of the values
```

joserodolfofreitas (commented):
I'm wondering if by default we display the search filter on the left (and move the toolbar buttons to the the right), assuming that generally speaking, the search input is probably going to be an interaction used more often than the toolbar buttons. 
What do you think?

Also, perhaps we can add a codesandbox with an example for getApplyQuickFilterFn, so the users can see it in action without searching for the getApplyFilterFn (which might eventually end up in a different page if we split the filtering doc page as well).




p.s.: example for getApplyFilterFn?
getApplyFilterFn

packages/grid/x-data-grid/src/components/toolbar/GridToolbar.tsx
```diff
@@ -9,32 +10,62 @@ import { GridToolbarDensitySelector } from './GridToolbarDensitySelector';
...
-      <GridToolbarFilterButton />
-      <GridToolbarDensitySelector />
-      <GridToolbarExport csvOptions={csvOptions} printOptions={printOptions} />
+      <GridToolbarButtonsWrapper>
+        <GridToolbarColumnsButton />
+        <GridToolbarFilterButton />
+        <GridToolbarDensitySelector />
+        <GridToolbarExport csvOptions={csvOptions} printOptions={printOptions} />
+      </GridToolbarButtonsWrapper>
+      {showQuickFilter && <GridToolbarQuickFilter {...quickFilterOptions} />}
```
m4theushw, line 68: An alternative is to remove `GridToolbarButtonsWrapper`  and use a div with `flex: 1` to fill the empty space. Both works.

```suggestion
      <Box sx={{ flex: 1 }} />
      {showQuickFilter && <GridToolbarQuickFilter {...quickFilterOptions} />}
```

One thing to note. It's trying to keep the search bar inline with the buttons on mobile. I think it could show the buttons on the first line, then the search bar in the next line taking all width.

image

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export type GridToolbarQuickFilterProps = TextFieldProps & {
+  /**
+   * Function responsible for parsing text input in an array of independent values for quick filtering.
+   * @param {string} input The value entered by the user
+   * @returns {any[]} The array of value on which quick filter is applied
+   */
+  quickFilterParser?: (input: string) => any[];
+  /**
+   * The debounce time in milliseconds.
```
m4theushw, line 36:
```suggestion
   * The debounce time in milliseconds.
   * @default 500
```

packages/grid/x-data-grid/src/models/gridFilterModel.ts
```diff
@@ -8,11 +8,21 @@ export interface GridFilterModel {
...
   linkOperator?: GridLinkOperator;
+  /**
+   * values used to quick filter rows
+   * @default `[]`
+   */
+  quickFilterValues?: any[];
+  /**
+   * - `GridLinkOperator.And`: the row must pass all the values.
+   * - `GridLinkOperator.Or`: the row must pass at least one value.
+   * @default `GridLinkOperator.Or`
```
m4theushw, line 25: The value in `getDefaultGridFilterModel` is "AND". One of the places is wrong.

```suggestion
   * @default `GridLinkOperator.And`
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -309,14 +309,76 @@ The example below demonstrates how to achieve server-side filtering.
...
+
+The logic used for quick filter can be switched to filter rows that contain _at least_ one of the values specified instead of testing if it contains all of them.
+To do so, set `quickFilterLogicOperator` to `GridLinkOperator.Or` as follow:
+
+```js
+initialState={{
+  filterModel: {
+    quickFilterLogicOperator: GridLinkOperator.Or
+  }
+}}
```
m4theushw, line 329: Without `items` it crashes.

```suggestion
initialState={{
  filter: {
    filterModel: {
      items: [],
      quickFilterLogicOperator: GridLinkOperator.Or,
    },
  },
}}
```

m4theushw (commented):
> I'm wondering if by default we display the search filter on the left (and move the toolbar buttons to the the right), assuming that generally speaking, the search input is probably going to be an interaction used more often than the toolbar buttons.
What do you think?

My opinion, keep it always on the right. I'm assuming that the idea of displaying the toolbar buttons on the right is only if `showQuickFilter` is enabled, otherwise it's a breaking change. If we invert, users will have to learn that to reach the toolbar buttons they need to move the mouse a bit more, different from grids that don't have the quick filter enabled. I'm imagining a page with several grids, one below the other. In this case, the position of the buttons would jump between left and right as the user scroll. I found two examples that have the actions + search on the right:

https://datatables.net/

image

Airtable doesn't always show the input, but the icon is always on the right.

image

⚠ This doesn't block us from adding a demo or explaining in the docs how to display the search bar on the left.

packages/grid/x-data-grid/src/components/toolbar/GridToolbar.tsx
alexfauquette: Yes, that's better with a native 'wrap'
image

packages/grid/x-data-grid/src/components/containers/GridToolbarContainer.tsx
```diff
@@ -27,6 +27,7 @@ const GridToolbarContainerRoot = styled('div', {
 })(({ theme }) => ({
   display: 'flex',
   alignItems: 'center',
+  flexWrap: 'wrap',
   padding: theme.spacing(0.5, 0.5, 0),
```
m4theushw, line 31: The input bottom border touches the column headers if we leave as `0px`.

image

```suggestion
  padding: theme.spacing(0.5),
```

m4theushw (approved):
I left some minor visual suggestions. Otherwise, it looks good to me to be released. Good job! How about enabling it in the demo in https://mui.com/x/react-data-grid/demo/?

packages/grid/x-data-grid/src/components/containers/GridToolbarContainer.tsx
alexfauquette: If I do that, it will also modify the Toolbar even when there is no quick filter

What about adding `paddingBottom:theme.spacing(0.5)` to the quick filter component?

m4theushw: IMO the bottom padding should be there since the beginning. The already existent buttons also touch the border.

image

packages/grid/x-data-grid/src/components/toolbar/GridToolbarQuickFilter.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+import SearchIcon from '@mui/icons-material/Search';
```
duttakapil21, line 4: Our app started failing to compile this morning because it couldn't resolve `@mui/icons-material`, seems like it is not added in the dependencies of this module. We had to separately install `@mui/icons-material` to get our app working again. We are not using Material Icons in our project at all either, is this an absolute necessity to fetch and use the SearchIcon from Material Icons in this package?

alexfauquette: This has been reported in #4906
The fix will be available in next release

PR #4319, opened 2022-03-31
[docs] Move `useKeepGroupingColumnsHidden` on `@mui/x-data-grid-premium`

flaviendelangle:
Until now, the `useKeepGroupingColumnsHidden` was copy pasted on every demo related to row grouping.
Its goal is to keep the columns used as a group hidden, which is a very common use case.

This PR adds `useKeepGroupingColumnsHidden` to `@mui/x-data-grid-premium`.

## Goals

- Make the demo code more readable and focused on the topic they are trying to cover
- Allow users to re-use the hook without copy / pasting

## Work

- [x] Create a `useKeepGroupingColumnsHidden` hook in `@mui/x-data-grid-premium`
- [x] Support both the initial state and the control model
- [x] Add a doc example


docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -41,8 +41,8 @@ You can use the `onRowGroupingModelChange` prop to listen to changes to the page
 
 ### Single grouping column
 
-By default, the grid will display a single column holding all grouped columns.
-If you have multiple grouped columns, this column name will be set to "Group".
+By default, the grid will display a single column holding all grouping columns.
```
flaviendelangle, line 44: I cleaned the naming here

- grouping column = column created to toggle the groups
- grouped column = column from which we are grouping (the one we want to hide here)

flaviendelangle:
Adding @samuelsycamore for the doc section

alexfauquette (commented):
It seems to be buggy with controlled column visibility model
https://codesandbox.io/s/rowgroupingbasicexample-material-demo-forked-czjilh?file=/demo.tsx

If you ungroup and group by company, heders and columns will be unsyc

flaviendelangle:
> It seems to be buggy with controlled column visibility model

@alexfauquette you are using both the controlled model and the initial state here

Here is the correct version: https://codesandbox.io/s/rowgroupingbasicexample-material-demo-forked-5g0piv

alexfauquette:
@flaviendelangle My example is not controlling the grouping state but the column visibility.

Probably it does not make sense to use this hook when controlling visibility because the developer is already responsible for which column is visible

Maybe a code example of how to listen to `rowGroupingModelChange` and update the controlled `columnVisibilityModel`

I tried something, but that clearly not a success 🙈

https://codesandbox.io/s/rowgroupingbasicexample-material-demo-forked-ioml8o?file=/demo.tsx

flaviendelangle:
Oh you are right...
I can probably support it

flaviendelangle:
@alexfauquette would it be ok for you if we just add a warning saying that it's not compatible with the controlled column visibility ?

By the way, if a column is hidden by the user on init, if it is being grouped then ungrouped, should be show it or keep it hidden ?

alexfauquette:
Yes, a warning is good enough for the hook

> By the way, if a column is hidden by the user on init, if it is being grouped then ungrouped, should be show it or keep it hidden?

I may miss a special case, but if the end-user group by a column, it means the column was visible. Overwise she would not have access to the column menu. So the column should be set back to visible when ungroupped

They might be special cases, where grouping is controlled, but that's too special to be handled natively. We can keep the native behavior simple show/hide when ungroup/group. And let developers do their custom logic if they need to support strange behavior

flaviendelangle:
> but if the end-user group by a column, it means the column was visible.

Unless he does it through the apiRef based on some external buttons (like on the screen below)

image


I don't think it is a negligible edge-case, users could have a column always hidden serving as a grouping criteria and a button to group / ungroup.

But I guess we can no support it for now, and wait for feedback. It's not a very common use case either.

PR #4322, opened 2022-03-31
[core] Add technical doc for pipe processing and family processing

flaviendelangle:


docsTech/processing.md
```diff
@@ -0,0 +1,263 @@
...
+#### `columnMenu`
+
+**Publisher**: `GridColumnMenu` component on render.
+
+**Why register to this processing**: Add one or multiple menu items to `GridColumnMenu`.
+
+**Example**:
+
+```tsx
+// Column pinning plugin
```
flaviendelangle, line 57: @alexfauquette what do you think is the best to demonstrate how each processor work ?

- A copy paste of a real use case, potentially simplified (I removed logic that were not interesting)
- A link to the file 
- A fake use case

docsTech/processing.md
```diff
@@ -0,0 +1,272 @@
...
+  },
+  [apiRef, classes, getSelectionColumn],
+);
+
+useGridRegisterPipeProcessor(apiRef, 'hydrateColumns', updateSelectionColumn);
+```
+
+> ⚠ The `columnsState` passed to the processors can contain the columns returned by the previous processing.
+> If the plugin is not enabled during the current processing, it must check if its columns are present, and if so remove them.
+
```
alexfauquette, line 50: If this is related to the example. It would be easier to understand as a comment in the code

docsTech/processing.md
```diff
@@ -0,0 +1,272 @@
...
+#### `rowHeight`
+
+**Publisher**: `useGridRowsMeta` plugin before updating `state.rowsMeta` (it is called for each row).
+
+**Why register to this processing**: Modify the base height of a row of add the height of some custom elements (eg: processor of the Detail Panel plugin).
```
alexfauquette, line 55:
```suggestion
**Why register to this processing**: Modify the base height of a row of add the height of some custom elements (eg: processor of the Detail Panel plugin increase the row height when the detail panel is open).
```

docsTech/processing.md
```diff
@@ -0,0 +1,272 @@
+# Processing
+
+## Introduction
```
alexfauquette, line 3: A suggestion from what I understood

```suggestion
## Introduction

Each feature is contained in a single hook, but features are not independent.
For example, the detail panel has an impact on the row height.

To allows hooks to interact and produce a coherent state, various patterns are presented on this page.

For each pattern, you will find a list of where such pattern is used, why it is necessary, and an overview of its behavior.

## Summary 

- Pipe-processing
  - Plugin state enrichment
  - Add custom behavior to an api method
  - Feature limitation
  - Component children processing

- Strategy-processing
```

alexfauquette (commented):
Sorry for the delay, I missed the mention. It's an amazing explanation.

I propose to add a summary to explicit the structure of the document. 

I learned stuff, and next time I need to use processors, I will have a look at this file before reading the code

flaviendelangle:
@alexfauquette thanks for your feedbacks

You will probably have to edit the `'columnMenu'` example on #3981

alexfauquette (approved):
Maybe add a section and a link to this page in the `README.md`

```md
## Internal state management

- [processing](./processing)
```

PR #4325, opened 2022-03-31
[DataGrid] Improve accessibility of the `actions` column

### LINKED ISSUE #3948: [DataGrid] Make actions menu accessible
### Duplicates

### Latest version

### Summary 💡

The menu added by the `actions` column type is not accessible. Once open, keyboard arrows don't work. Pressing <kbd>Esc</kbd> should close it.

image

Related to https://github.com/mui/mui-x/pull/3375#issuecomment-1038820718

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #3948


packages/grid/x-data-grid/src/components/toolbar/GridToolbarDensitySelector.tsx
```diff
@@ -101,7 +101,7 @@ export const GridToolbarDensitySelector = React.forwardRef<HTMLButtonElement, Bu
           aria-label={apiRef.current.getLocaleText('toolbarDensityLabel')}
           aria-expanded={open ? 'true' : undefined}
           aria-haspopup="menu"
-          aria-labelledby={densityMenuId}
+          aria-controls={densityMenuId}
```
m4theushw, line 104: Not related to #3948 but this button already has a label.

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
```diff
@@ -16,76 +17,198 @@ interface TouchRippleActions {
...
+      setFocusedButtonIndex(index);
+      ignoreCallToFocus.current = true;
+
+      if (onClick) {
+        onClick(event);
+      }
+    };
+
+  const handleRootKeyDown = (event: React.KeyboardEvent) => {
+    if (iconButtons.length + menuButtons.length === 1) {
```
alexfauquette, line 129: Should it be
```suggestion
    if (numberOfButtons <= 1) {
```

Because even if `menuButtons` contains lot of options it's still one button so right/left arrow should do nothing

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
```diff
@@ -16,76 +17,198 @@ interface TouchRippleActions {
...
       });
     }
   }, [hasFocus]);
 
+  React.useEffect(() => {
+    if (focusedButtonIndex >= 0) {
+      const child = rootRef.current?.children[focusedButtonIndex] as HTMLElement;
+      child.focus();
+    }
+  }, [focusedButtonIndex]);
```
alexfauquette, line 71: Is it necessary or the following line does the job?
```jsx
tabIndex={focusedButtonIndex === iconButtons.length ? tabIndex : -1}
```

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
```diff
@@ -16,76 +17,198 @@ interface TouchRippleActions {
...
+
+    let newIndex: number = focusedButtonIndex;
+    if (event.key === 'ArrowRight') {
+      newIndex += 1;
+    } else if (event.key === 'ArrowLeft') {
+      newIndex -= 1;
+    }
+
+    if (newIndex < 0 || newIndex >= numberOfButtons) {
+      return; // We're already in the first or last item = do nothing
```
alexfauquette, line 141:
```suggestion
      return; // We're already in the first or last item = do nothing and let the grid listen the event
```

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
```diff
@@ -104,10 +227,39 @@ GridActionsCell.propTypes = {
...
   /**
    * The column of the row that the current cell belongs to.
    */
   colDef: PropTypes.object.isRequired,
+  /**
+   * The column field of the cell that triggered the event.
+   */
+  field: PropTypes.string.isRequired,
+  /**
+   * A ref pointing to the element that should receive focus.
```
alexfauquette, line 243: I do not see the ref passed to an element

```suggestion
   * A ref allowing to set imperative focus.
```

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
```diff
@@ -16,76 +17,198 @@ interface TouchRippleActions {
...
+    isEditable,
+    field,
+    value,
+    formattedValue,
+    row,
+    rowNode,
+    cellMode,
+    getValue,
+    tabIndex,
+    position = 'bottom-end',
```
alexfauquette, line 41:
```suggestion
    field,
    row,
    rowNode,
    tabIndex,
    position = 'bottom-end',
```

Does it make sense to add props linked to cell values for the action cell?

packages/grid/x-data-grid/src/tests/rows.DataGrid.test.tsx
```diff
@@ -244,46 +247,116 @@ describe('<DataGrid /> - Rows', () => {
...
+          ]}
+        />,
+      );
+      const firstCell = getCell(0, 0);
+      firstCell.focus();
       expect(getActiveCell()).to.equal('0-0');
+
+      fireEvent.keyDown(firstCell, { key: 'ArrowRight' });
+      const printButton = screen.getByRole('menuitem', { name: 'print' });
+      expect(printButton).toHaveFocus();
```
alexfauquette, line 309: Why do you have to press `arrowRight` to set the focus on the first button?

m4theushw: The test component has 2 columns. First I focus on the first cell, then <kbd>ArrowRigth</kbd> to go to the next column.

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
m4theushw: `tabIndex` only adds or removes the element from the tab sequence. It's `element.focus` that will call focus.

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
m4theushw: They come automatically from `GridCellParams`.

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
m4theushw: Yeah, it's not passed to an element only on this type of column. But on others, it could be. `actions` has internal navigation so we need to handle it differently. Ideally `focusElementRef` would go to an element.

How about?

```suggestion
   * A ref allowing to set imperative focus.
   * It can be passed to the element that should receive focus.
```

packages/grid/x-data-grid/src/components/cell/GridActionsCell.tsx
alexfauquette: I think I get your point. You are removing those properties to `other` to do the TODO comment

packages/grid/x-data-grid/src/tests/rows.DataGrid.test.tsx
alexfauquette: Ok, I missed that you added `id` column

alexfauquette (approved):
With your proposal for the documentation modification, it's good for me

PR #4332, opened 2022-04-02
[core] Add React 18 to peer dependencies

m4theushw:
CodeSandbox: https://codesandbox.io/s/datagridprodemo-material-demo-forked-qvr2t0?file=/demo.tsx

In the comments, I explain the changes I had to make. I updated the peer deps of the date picker components but I don't know if they are compatible, are they? We didn't really test it in React 18 so one way to review this PR could be to open the example above and check if all features still work as expected. The tests are passing in React 18. #4155


packages/grid/x-data-grid/src/hooks/features/columnHeaders/useGridColumnHeaders.tsx
```diff
@@ -87,13 +98,30 @@ export const useGridColumnHeaders = (props: UseGridColumnHeadersProps) => {
...
+      let canUpdateInnerPosition = false;
+
       if (nextRenderContext !== prevRenderContext.current || !prevRenderContext.current) {
-        setRenderContext(nextRenderContext);
+        // ReactDOM.flushSync cannot be called on `scroll` events fired inside effects
+        if (isUIEvent(event)) {
+          // To prevent flickering, the inner position can only be updated after the new context has
+          // been rendered. ReactDOM.flushSync ensures that the state changes will happen before
+          // updating the position.
+          ReactDOM.flushSync(() => {
```
m4theushw, line 111: Fixes

msedge_c95qQZ4C70

cherniavskii:
Should we start using React 18 in our docs? This may help to discover issues sooner. What do you think?

m4theushw:
> Should we start using React 18 in our docs? This may help to discover issues sooner. What do you think?

@cherniavskii I'm waiting for the core to update first. Keeping two versions of React may not work correctly. Once the core updates we can merge #4155.

jeveloper:
Thank you and i'm glad to see these amazing changes.
However, wouldn't the main lib mui/material be in conflict as it stricty supports only up to react 17?
I don't assume the main lib is getting overhauled this quickly to match depdencies.

Thanks

m4theushw:
> However, wouldn't the main lib mui/material be in conflict as it stricty supports only up to react 17?

@jeveloper No, the latest version of `@mui/material` is already updated to support React 18.

Issue #1106, opened 2021-02-23
[DataGrid] Provide an API to access the client-side filtered rows


giq-re:
No easy method to get the current state of visible rows as the filter component updates. the onFilterModelChange function is one step behind the current state of the table - which may be the intended behavior - because it may be a private method.

- [x] The issue is present in the latest release.
- [x ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

onFilterModelChange one step behind actual state of the table - think this is because its a private method and is actually updating the state of the table. 

## Expected Behavior 🤔
a method to get the actual visible rows as the filter updates.

## Steps to Reproduce 🕹

below is a link to a codesanbox
https://codesandbox.io/s/material-demo-forked-nzong?file=/demo.js:867-907

Steps:

1. simply filter on quantity
2. watch the console as you type: >= 500000 and see how the visibleRowsLookup is one step behind actual state
3.
4.

## Context 🔦
notice how it is always one behind actual state...when you look at 
the console of the params object it is actual correct. this must be an 
asynchronous action. I would like to be able to access visible rows
in real time to update state in another part
of the application....imagine if you were updating state in a map (such as google maps and mapbox - my actual scenario - and you may want to style or zoom or visualize other data related to the displayed table in another part of the screen) 
visualization platform -  update state based on 
any applied filters. you may not want to handle or you may not be able to handle all these updates with getSelectedRows - you may not care about whats selected....just filtered....
I don't believe this is an edge case...this is something that I imagine is going to be needed.  if I am missing something...please let me know....I can def be a bonehead....

thanks for the help. 

## Your Environment 🌎

<details>
  <summary>`npx @material-ui/envinfo`</summary>
  
```
  Don't forget to mention which browser you used.
  Output from `npx @material-ui/envinfo` goes here.
```
</details>

checkmatez:
I've encountered the same problem: how to get rows after filters were applied?
If I try to use `onFilterModelChange` and access `visibleRows` inside the state, they lag behind one update on filters.
Also, being able to access internal state via React ref would be nice:
```
const ref = React.useRef();
console.log(ref?.state)

<DataGrid ref={ref} />
```
unfortunately, ref gets passed to the outer div, and not the DataGrid component as a whole.

dtassone:
@checkmatez to access the internal state you could use the `onStateChange` prop.

```tsx
[...]

PR #4359, opened 2022-04-04
[data grid] Apply filtering before sorting

flaviendelangle:
Extracted from #4208 

This PR invert the order of the `useGridFilter` and `useGridSorting` plugins.
The change touches important parts of the grid and should not be done lightly.
I waited for something really blocking before doing it, but for the Aggregation I don't think we can avoid it. Being able to aggregate on the filtered values is the most common use case (default behavior on AG Grid).

## Why ?

1. **Aggregation**: We want to be able run the aggregation only on the filtered rows of each group **AND** to be able to sort the groups according to their aggregated value. The computation of the aggregated values must occur after the filtering and before the sorting.

2. **Row Grouping**: We want to be able to sort the groups according to the number of children they contain (see #2994).

## What's next ?

- Apply the same change on #4223  (or master if merged after it)


packages/grid/x-data-grid/src/models/events/gridEvents.ts
```diff
@@ -259,7 +259,16 @@ enum GridEvents {
...
+  /**
+   * Fired when the sorted rows are updated
+   * @ignore - do not document
+   */
+  sortedRowsSet = 'sortedRowsSet',
+  /**
+   * Fired when the filtered rows are updated
+   * @ignore - do not document.
+   */
+  filteredRowsSet = 'filteredRowsSet',
```
flaviendelangle, line 271: This event is never used internally
I created it to handle #1106 but I can remove it from this PR

packages/grid/x-data-grid/src/hooks/features/sorting/useGridSorting.ts
```diff
@@ -119,19 +119,18 @@ export const useGridSorting = (
...
-        sorting: {
-          ...state.sorting,
-          sortedRows: gridRowIdsSelector(state, apiRef.current.instanceId),
-        },
-      }));
-      return;
-    }
-
     apiRef.current.setState((state) => {
+      if (props.sortingMode === GridFeatureModeConstant.server) {
```
flaviendelangle, line 123: I moved everything into a single `setState` to make sure we always publish the `sortedRowsSet` event after

cherniavskii (approved):
LGTM

m4theushw (approved):

PR #4361, opened 2022-04-04
[DataGrid] Enable using non-native Select in filter panel

### LINKED ISSUE #4127: [DataGrid] Add MUI Select component to data grid filter panel
### Duplicates

### Latest version

### Summary 💡

So what I'm saying is that the select box for choosing columns and operators of the filter panel is not consistent with the rest of the design and uses that native look
I would appreciate it if you add support of [Select](https://mui.com/components/selects/) component to provide a consistent UI

### Examples 🌈

Screenshot 2022-03-08 192929

### Motivation 🔦

_No response_

### Order ID 💳 (optional)

_No response_

kyeongsoosoo:
Preview: https://codesandbox.io/s/columntypesgrid-material-demo-forked-ll8bmp?file=/demo.tsx

Closes #4127 

This PR resolves the issue that `Select` component in `Filter Panel` only uses `native UI`.

Following the [suggestion](https://github.com/mui/mui-x/issues/4127#issuecomment-1061962245), I made 3 changes.

1. `ClickAwayListener` in `Filter Panel` uses `onMouseUp` mouseEvent.
```ts
<ClickAwayListener mouseEvent="onMouseUp" onClickAway={handleClickAway}>
```

2. When `rootProps.componentsProps?.baseSelect.native` is set to false, `baseSelect` component uses MUI component and `MenuItem` component is used instead of `option` tag.

3. `rootProps.componentsProps?.baseSelect` is passed to `InputComponent` when `Select` Component is used.
```ts
      SelectProps={{
        native: isSelectNative,
        ...rootProps.componentsProps?.baseSelect,
      }}
```

NOTE. To avoid breaking change, I set  default `native` to true.

## Before

https://user-images.githubusercontent.com/59413880/161584552-93ef11d9-5bb7-47e6-b418-6d00ef3a6860.mov

## After

https://user-images.githubusercontent.com/59413880/161584659-dfd3bd8b-43d8-4abe-94f2-afa653d2fb0b.mov


alexfauquette (commented):
Supporting both native and non-native select seems to be a pain due to the switch between `<option />` and `<MenuItem />`

What would you think about adding a helper component to switch between them more easily

```jsx
const SelectOption = ({ isNative, ...props }) => isNative ? <option {...props} /> : <MenuItem {...props} />
```

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -314,14 +317,20 @@ function GridFilterForm(props: GridFilterFormProps) {
...
           {...rootProps.componentsProps?.baseSelect}
         >
-          {linkOperators.map((linkOperator) => (
-            <option key={linkOperator.toString()} value={linkOperator.toString()}>
-              {apiRef.current.getLocaleText(getLinkOperatorLocaleKey(linkOperator))}
-            </option>
-          ))}
+          {linkOperators.map((linkOperator) =>
+            isBaseSelectNative ? (
+              <option key={linkOperator.toString()} value={linkOperator.toString()}>
```
m4theushw, line 325: The props are the same for `MenuItem` and `option`. You could store in an `OptionComponent` variable which option component to use, then do:

```tsx
const OptionComponent = ...

linkOperators.map((linkOperator) => (
  <OptionComponent key={...} value={...} />
))
```

Same for the other components. This reduce code duplication.

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputBoolean.tsx
```diff
@@ -21,6 +25,32 @@ export function GridFilterInputBoolean(props: GridFilterInputValueProps & TextFi
...
+      <rootProps.components.BaseTextField
+        label={apiRef.current.getLocaleText('filterPanelInputLabel')}
+        value={filterValueState}
+        onChange={onFilterChange}
+        variant="standard"
+        select
+        SelectProps={{
+          native: true,
+          ...rootProps.componentsProps?.baseSelect,
+        }}
```
m4theushw, line 39: We could keep the same `rootProps.components.BaseTextField` but only have a `SelectProps` object which attributes vary depending on `isSelectNative`. Also, to avoid code duplication.

packages/grid/x-data-grid/src/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -137,6 +137,7 @@ function GridEditSingleSelectCell(props: GridRenderEditCellParams & Omit<SelectP
       fullWidth
       {...other}
       {...rootProps.componentsProps?.baseSelect}
+      native={false}
```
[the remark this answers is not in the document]
m4theushw: > Another solution is to support native select by adding an argument isNative to renderSingleSelectOptions such that it can switch between <MenuItem/> and <option />

+1 for this option. It will be odd to have all selects working with `native=false` and only this one that doesn't.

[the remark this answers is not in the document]
kyeongsoosoo: I chose the latter solution(support native select).
I tested the change using Storybook, and I found that `NativeSelect` casts the value to a string.
To fix that issue, I copied the util from `filterPanelUtils` and formatted the value.
```ts
  const handleChange = async (event: React.KeyboardEvent<HTMLInputElement>) => {
    setOpen(false);
    const target = event.target as HTMLInputElement;
    // NativeSelect casts the value to a string.
    const formattedTargetValue = getValueFromValueOptions(target.value, valueOptionsFormatted);
    const isValid = await api.setEditCellValue({ id, field, value: formattedTargetValue }, event);
    ...
```

kyeongsoosoo:
@alexfauquette  @m4theushw 
Thanks for the review. I refactored some code following the suggestions.

packages/grid/x-data-grid/src/components/cell/GridEditSingleSelectCell.tsx
```diff
@@ -12,18 +12,39 @@ import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
...
-const renderSingleSelectOptions = (option: ValueOptions) =>
-  typeof option === 'object' ? (
-    <MenuItem key={option.value} value={option.value}>
-      {option.label}
-    </MenuItem>
+interface SelectOptionProps {
+  isNative: boolean;
+  children: React.ReactNode;
+  value: any;
+}
```
alexfauquette, line 21: This informs typescript that `...props` can contain any porps of the `MenuItem`

```suggestion
interface SelectOptionProps extends MenuItemProps {
  isNative: boolean;
  children: React.ReactNode;
  value: any;
}
```

The type `MenuItemProps` can be imported from `'@mui/material/MenuItem'`

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -102,6 +103,23 @@ const FilterFormValueInput = styled(FormControl, {
...
+const SelectOption = ({ isNative, children, value, ...props }: SelectOptionProps) =>
+  isNative ? (
+    <option value={value} {...props}>
+      {children}
+    </option>
+  ) : (
+    <MenuItem value={value} {...props}>
+      {children}
+    </MenuItem>
+  );
```
alexfauquette, line 121: Since this code is duplicated, maybe a dedicated file `components/SelectOption.tsx` could be more appropriate

alexfauquette (commented):
Changes look nice. Maybe putting the `SelectOption` in a dedicated file to avoid duplicated code, and for me it will be good.

packages/grid/x-data-grid/src/components/cell/GridEditSingleSelectCell.tsx
kyeongsoosoo: I considered extending `MenuItemProps`, but `props` from `MenuItemProps` cannot be passed to `option` tag.
What would you think about seperating `optionProps` and `menuItemProps` to pass accurate `props`
It will be like...
```ts
type SelectOptionProps = {
  isNative: boolean;
  children: React.ReactNode;
  value: any;
  optionProps?: React.DetailedHTMLProps<
    React.OptionHTMLAttributes<HTMLOptionElement>,
    HTMLOptionElement
  >;
  menuItemProps?: MenuItemProps;
};

function SelectOption({
  isNative,
  children,
  value,
  optionProps,
  menuItemProps,
  ...props
}: SelectOptionProps) {
  return isNative ? (
    <option value={value} {...optionProps} {...props}>
      {children}
    </option>
  ) : (
    <MenuItem value={value} {...menuItemProps} {...props}>
      {children}
    </MenuItem>
  );
}
```

m4theushw (commented):
I pushed [a4110eb](https://github.com/mui/mui-x/pull/4361/commits/a4110ebbd37b54d002edea2319722232084a984b) simplifying the way we choose which component to use for the options. I think we were duplicating too much code. What do you think now?

kyeongsoosoo:
It looks simpler than my first version.
Thanks for a better solution.

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputValue.tsx
```diff
@@ -63,16 +70,21 @@ function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldPr
...
+  const baseSelectProps = rootProps.componentsProps?.baseSelect || {};
+  const isSelectNative = baseSelectProps.native ?? true;
+
   const singleSelectProps: TextFieldProps =
     type === 'singleSelect'
       ? {
           select: true,
           SelectProps: {
-            native: true,
+            ...rootProps.componentsProps?.baseSelect,
```
alexfauquette, line 82:
```suggestion
            native: true,
            ...rootProps.componentsProps?.baseSelect,
```
Or
```suggestion
            native: isSelectNative,
            ...rootProps.componentsProps?.baseSelect,
```

alexfauquette (approved):
I added a commit to change import helpers from the single select

Just spot a typo, otherwise it looks good

m4theushw (approved):

PR #4368, opened 2022-04-05
[core] Unify tests

flaviendelangle:
- [x] Name all the `describe` focusing on a prop `props: propName` (we had both `props: propName` and `prop: propName`)
- [x] Remove default value to `getColumnValues` (I find it very unclear that `getColumnValues()` gets the 1st column values, better be explicit here)
- [x] Use `document.querySelector` generic instead of casting the return value

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/rowGrouping.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/rowGrouping.DataGridPro.test.tsx
@@ -104,7 +104,7 @@ describe('<DataGridPro /> - Group Rows By Column', () => {
       });
     });
 
-    describe('prop: rowGroupingModel', () => {
+    describe('props: rowGroupingModel', () => {
       it('should not call onRowGroupingModelChange on initialisation or on rowGroupingModel prop change', () => {
         const onRowGroupingModelChange = spy();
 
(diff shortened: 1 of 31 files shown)


packages/grid/x-data-grid-pro/src/tests/rowGrouping.DataGridPro.test.tsx
```diff
@@ -104,7 +104,7 @@ describe('<DataGridPro /> - Group Rows By Column', () => {
       });
     });
 
-    describe('prop: rowGroupingModel', () => {
+    describe('props: rowGroupingModel', () => {
```
m4theushw, line 107: The core repo uses `prop: propName`: https://github.com/mui/material-ui/blob/32fd4a3249fb1de139b17ed54bd54802b3d50f44/packages/mui-material/src/Autocomplete/Autocomplete.test.js#L112

flaviendelangle: I can put `prop: propName` to be consistent with the core then

flaviendelangle: Done

m4theushw (approved):

PR #4378, opened 2022-04-06
[core] Move away for the event system to trigger pipe processings

flaviendelangle:
Extracted from #4208

**This PR only impacts the pipe processing which are triggered when `GridEvents.columnsChange` is fired. Pipe processing lilke `columnMenu` or `scrollToIndexes` are not impacted.**

## Current behavior for processors

1. Plugin A registers a processor
2. Plugin B run the processors on mount
3. Plugin B listens to `GridEvents.pipeProcessorRegister` and re-apply the processors whenever this event is fired on the right group
4. Plugin A has an update on one of the deps of its processor, it triggers a re-registration and therefore a re-application. 

A problem appears when we want to re-apply the processors but none of the deps of the processor have changed.

For instance in the row grouping, we want to re-apply `hydrateColumns` when the columns in a way which changes the sanitized `rowGroupingModel`.
If we just use `gridRowGroupingSanitizedModelSelector` in the render of `useGridRowGroupingProcessors`  to update the deps of the processor, then we would have two problems:

1. Any change to the columns would re-apply `hydrateColumns`, creating an infinite loop 
2. The processor re-registration being in an effect, the grid would re-render with the new columns but the old grouping column and then a 2nd time with the new columns and the new grouping column. 

The current hack to listen to `GridEvents.columnsChange` and to fire manually `hydrateColumns` by calling `apiRef.current.updateColumns([])`.
This hacky solutions works and re-apply `hydrateColumns` synchronously when the columns are updated in a way that changes the sanitized `rowGroupingModel`.

## Why do we need to fix it ?

For the aggregation, I need to do something similar.
When the columns changes, I need to re-apply:
-  `hydrateColumns` to wrap / unwrap the aggregated columns (the same hack can be used)
- `hydrateRows` (a new pipe processing that adds rows and is used by the aggregation to create the footers)

The same hack can't be used for `hydrateRows`.
If I call `apiRef.current.updateRows([])`, the update of the state is throttled which is problematic here.
I could create a new `apiRef.current.unstable_syncUpdateRows`, but I think it is worth cleaning the pattern rather than creating more dirt.

### For the processors

I introduced a new method `unstable_requestPipeProcessorsApplication` which aims at firing the appliers of a group imperatively.

```diff
-apiRef.current.updateColumns([]);
+apiRef.current.unstable_requestPipeProcessorsApplication('hydrateColumns');
```

### For the appliers

Instead of listening to `GridEvents.pipeProcessorRegister` to re-apply their logic, plugins can now use `useGridRegisterPipeApplier` to give a callback which will be automatically be called

- When a processor is registered (equivalent of `GridEvents.pipeProcessorRegister`)
- When `apiRef.current.unstable_requestPipeProcessorsApplication` is called

```diff
-const handlePipeProcessorRegister = React.useCallback<
-  GridEventListener<GridEvents.pipeProcessorRegister>
->(
-  (name) => {
-    if (name !== 'hydrateColumns') {
-      return;
-    }
-    const columnsState = createColumnsState({ /* ... */ });
-    setGridColumnsState(columnsState);
-  },
-  [apiRef, logger, setGridColumnsState, columnTypes],
-);
-
-useGridApiEventHandler(apiRef, GridEvents.pipeProcessorRegister, handlePipeProcessorRegister);

+const hydrateColumns = React.useCallback(() => {
+  const columnsState = createColumnsState({ /* ... */ });
+  setGridColumnsState(columnsState);
+}, [apiRef, logger, setGridColumnsState, columnTypes]);
+
+useGridRegisterPipeApplier(apiRef, 'hydrateColumns', hydrateColumns);
```

______

#4322 will document in detail this behavior


m4theushw (approved):
No problem pushing this forward. I think we should still avoid calling `unstable_requestPipeProcessorsApplication` as much as possible, to not cause an additional rerender. The features should be designed to be resilient when the model contains invalid data, e.g. missing columns.

flaviendelangle:
> I think we should still avoid calling unstable_requestPipeProcessorsApplication as much as possible

Yes
And if you have a concrete way of not using it in specific scenario, I am always interested

PR #4383, opened 2022-04-06
[DatePicker] Fix previous input value issues

yyq321:
Re-open https://github.com/mui/material-ui/pull/31929 here

fixes mui/mui-x#4478, fixes mui/mui-x#4469

not sure whether mui/material-ui#4358 is really a bug. I think people need a way to cancel their selections on the desktop.

mui/mui-x#4478 and mui/mui-x#4469 are caused by calling `setInitialDate` whenever we close the picker. This design could not handle the input change after closing. Instead, we should call `setInitialDate` just before the picker opens.

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
+++ b/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
@@ -72,6 +72,7 @@ export const usePickerState = <TInput, TDateValue>(
   }
 
   const [initialDate, setInitialDate] = React.useState<TDateValue>(draftState.committed);
+  const [wrapper, setWrapper] = React.useState<WrapperVariant>(null);
 
   // Mobile keyboard view is a special case.
   // When it's open picker should work like closed, cause we are just showing text field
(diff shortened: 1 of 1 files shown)


yyq321:
@flaviendelangle thanks for reminding me. I modified another line to handle `disableCloseOnSelect`. If `disableCloseOnSelect` is `true`, users accept the value selected when closing the picker. Else, users cancel their selections when closing the picker.

flaviendelangle:
@yyq321 thanks for your detailed feedback

I am to clarify this behavior in #4408 
I agree that it is a key behavior and that any change should be done carefully

BGehrels:
> * On `desktop` and `disableCloseOnSelect` is `false`: clicking away (`onDismiss`) should reset the date to `initialDate` and close pickers

I'm not sure about this: This feels nice when you *first* select a date, but when you open a picker for the second time, you probably want to correct an error you did the first time. "Open-> fix day -> click away -> picker closes" ist a very convinient way and feels "natural". "Open -> fix the day -> reclick the selected hour -> reclick the selected minute -> picker closes" is annoying.

I would also argue that there is no use case for cancelling an accidential selection on desktop, as you can always change it. Cancelling feels more like a 90ies Wizard UX than a modern immediate-visual-feedback-change-as-you-go approach

Last but not least: usually, the value will be shown in the connected TextArea and be updated immediatly while you change the pickers, as shown in the second video [here](https://github.com/mui/material-ui/pull/31093#issue-1138700161). So the UX metaphor is "selecting is equivalent to typing", as the field updates the same way as if i type in there. But: No TextArea ever clears/resets when i click away. So neither should the pickers selection.

flaviendelangle:
@BGehrels can we continue this discusson on #4408 ?
I will document correctly the `usePickersState` hook. These reset represent like a third of the recent issues around the pickers so let's clean it correctly

alexfauquette:
@flaviendelangle do you mean PR #4408 instead of the issue https://github.com/mui/material-ui/issues/4408 ?

flaviendelangle:
Closing this one if favor of #4408

Issue #4374, opened 2022-04-06
[data grid] help creating an edit cell where dropdown options depend on selection made in a different edit cell.


nickraphael:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Lets suppose the first edit cell in my table lets the user choose car manufacturer.  Then, depending on that choice, the available options in the next edit cell are filtered car models.  Also, the whole row is in edit mode (editMode="row").

I've tried coding this manually with renderEditCell and outputting mui-select but the interplay is a nightmare.  Validation becoming super complex.

Has anyone got any advice or code samples?

Many thanks

alexfauquette:
You have a similar example in the demo for single select columns which uses a function to define `valueOptions`, such as depending on the row values, the proposed discount is adapted

https://mui.com/components/data-grid/columns/#special-properties

```jsx
valueOptions: ({ row }) => { 
 if (row === undefined) {
    return ['EU-resident', 'junior'];
  }
  const options = [];
  if (!['United Kingdom', 'Brazil'].includes(row.country)) {
    options.push('EU-resident');
  }
  if (row.age < 27) {
    options.push('junior');
  }
  return options;
},
```

nickraphael:
I’ve tried that approach but the incoming row seems to be the initial model - not the updated row once the first field is updated.

alexfauquette:
Yes, the row is the row before editing. If you use row edit mode, there is no simple way to do that for now

With cell edit mode, you should not have the problem since after editing the first cell the row will be updated

If you really need row editing, a workaround is to use the `apiRef` such that in `valueOptions` you can get access to the editing state
```js
const valueOptions = React.useCallback(({id: rowId, row}) => {
	const editState = gridEditRowsStateSelector(apiRef);
	const constructor = editState[rowId]?.constructor?.value || row.constructor

	return carModels[constructor]
}, [apiRef])
```

nickraphael:
Interesting thank you.  I’ll try that.  I’m still left with the issue of the value of field2 needing to be reset when field1 changes.  Otherwise I get that console warning about the field not being present in the list of options.  
in our example, that might be a row with constructor=Ford and model=Ranger.  User changes constructor to Mazda which changes the model list so that it no longer contains ranger.  If I don’t reset the value of carModel, I get that warning in the console.
Is there any event that fires when an item is selected in a singleSelect cell when in editMode=row?

m4theushw:
[...]

PR #4392, opened 2022-04-07
[DataGrid] Pass updated row to edit components

m4theushw:
Part of #4374 

When editing a row, the `row` prop passed to the edit components is the row with the values before entering the edit mode. This prevents the use of this prop to, for example, change the options returned by `valueOptions` depending on the value of another field. This PR changes this behavior to always pass the row with the updated values. On cell editing, only the field that is being edited is updated, the others remain the same, even if they are being edited too.

CodeSandbox showing how this PR helps #4374: https://codesandbox.io/s/conditionalvalidationgrid-material-demo-forked-tnsu0e?file=/demo.tsx


packages/grid/x-data-grid/src/components/GridRow.tsx
```diff
@@ -257,7 +257,13 @@ function GridRow(props: React.HTMLAttributes<HTMLDivElement> & GridRowProps) {
     }
 
     if (editCellState != null && column.renderEditCell) {
-      const params = { ...cellParams, ...editCellState, api: apiRef.current };
+      let updatedRow = row;
+      if (apiRef.current.unstable_getRowWithUpdatedValues) {
+        // Only the new editing API has this method
+        updatedRow = apiRef.current.unstable_getRowWithUpdatedValues(rowId, column.field);
+      }
```
alexfauquette, line 264: Is something blocking in the previous API to create the same method?

m4theushw: Nothing blocks. I was considering that the legacy API wouldn't receive more fixes or improvements. Should it be added there too?

cherniavskii (approved):

PR #4398, opened 2022-04-08
[time picker] Only update time when updating input in TimePicker

### LINKED ISSUE #4397: [TimePicker] When changing only the time of a TimePicker using the keyboard, the date also changes to today's date.
### Duplicates

### Latest version

### Current behavior 😯

This issue was reported [here](https://github.com/mui/material-ui/issues/26799) and marked as fixed, but it was not fixed and can still be seen on the documentation.

When you use the keyboard to change the time of a TimePicker in the input component, rather than press the clock icon and use the picker, the date of the Date object it emits is always today's date. This is not the case when you use the picker Popover section.

### Expected behavior 🤔

The behavior should be consistent between using the keyboard and the Popover. Moreover, since it is a TimePicker, it is unexpected for it to ever change the date.

### Steps to reproduce 🕹

Steps:

1. Go to 'React Components' section of [the Pickers documentation](https://mui.com/components/pickers/). It is important to note that all four Pickers point to the same Date object.
2. In any of the Date pickers, change the date to a day other than today's date.
3. In the TimePicker, type in a new time. Only use the keyboard input into the TextField, not the Popover time selection component.
4. Notice that the date has changed to today's date.

### Context 🔦

Like the documentation, our application has separate Date and Time Pickers that point to the same Date object. Because of this, changing the time manually changes the date, making scheduling very difficult.

### Your environment 🌎

Tested on Chrome and Firefox, on both our app (where it was discovered) and the documentation (where it can be reproduced)

<details>

System:
    OS: macOS 11.6.2
  Binaries:
    Node: 14.18.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 8.5.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 99.0.4844.84
    Edge: Not Found
    Firefox: 98.0.2
    Safari: 15.2
  npmPackages:
    @emotion/react: ^11.8.2 => 11.8.2
    @mui/base:  5.0.0-alpha.72
    @mui/icons-material: ^5.5.0 => 5.5.1
    @mui/lab: ^5.0.0-alpha.72 => 5.0.0-alpha.73
    @mui/material: ^5.5.0 => 5.5.1
    @mui/private-theming:  5.4.4
    @mui/styled-engine:  5.4.4
    @mui/styles: ^5.5.0 => 5.5.1
    @mui/system:  5.5.1
    @mui/types:  7.1.3
    @mui/utils:  5.4.4
    @mui/x-data-grid:  5.6.1
    @mui/x-data-grid-pro: ^5.6.1 => 5.6.1
    @mui/x-license-pro:  5.6.1
    @types/react:  17.0.40
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.6.2 => 4.6.2
</details>

flaviendelangle:
Fixes #4397

@alexfauquette I would like your opinion on this one.

My logic is to get the time (hour, minute, seconds, milliseconds) from the new date parsed from the input. But to get the date (day, month, year) from the previous date.

It works great, except if you don't have the `ignoreInvalidInputs` set to `true` (it blocks the onChange firing when the date is invalid).
Because as soon as you get an invalid date, your previous date becomes null on the next `onChange` and I can't merge it.

What do you prefer ?

- We only fix for `ignoreInvalidInputs=true`
- We keep track of the last valid date in a ref 
- Something else ?


packages/x-date-pickers/src/DesktopTimePicker/DesktopTimePicker.tsx
```diff
@@ -18,6 +18,13 @@ const valueManager: PickerStateValueManager<unknown, unknown> = {
   parseInput: parsePickerInputValue,
   areValuesEqual: (utils: MuiPickersAdapter<unknown>, a: unknown, b: unknown) =>
     utils.isEqual(a, b),
+  updateValue: (utils, prevValue, newValue) => {
```
flaviendelangle, line 21: Optional method (to be named in a better way) that creates the new value from the previous valud and the new one.

Allows us in TimePicker to only take the time from the new value but keep the date of the old value.

TODO: Do the same for MobileTimePicker (maybe refacto to  avoid duplication on the value managers at some point)

_____

date-fns handles this behavior out of the box, you can pass a reference date and when updating a mask it only applies the change of the elements in the mask, the rest is taken from the reference date.
It would be perfect here, but the others date libs do not support it from what I see.

alexfauquette:
For me, it makes sense to track the last valid date. Since we are talking about a time picker, I assume the date can only be modified by the `value` prop

Why using a `ref` instead of a state?

flaviendelangle:
The ref was to avoid a re-render since we are only querying its value in a callback.

I did a first version with a state and a test

alexfauquette:
It sounds good. Do you want to do the `MobileTimePicker` in this PR?

flaviendelangle:
Yes, I applied the same logic on the mobile

packages/x-date-pickers/src/internals/hooks/usePickerState.ts
```diff
@@ -141,14 +160,24 @@ export const usePickerState = <TInput, TDateValue>(
     [acceptDate, disableCloseOnSelect, isMobileKeyboardViewOpen, draftState.draft],
   );
 
+  const handleInputChange = React.useCallback(
+    (date: TDateValue, keyboardInputValue?: string) => {
+      const cleanDate = valueManager.mergeOldAndNewValues
+        ? valueManager.mergeOldAndNewValues(utils, lastValidDateValue, date)
+        : date;
+      onChange(cleanDate, keyboardInputValue);
+    },
```
alexfauquette, line 169: If `valueManager` can be provided by developers, maybe renaming `prevValue` by `prevValidValue` to be more accurate on what is done internally

flaviendelangle: It is fully internal

alexfauquette (approved):

packages/x-date-pickers/src/DesktopTimePicker/DesktopTimePicker.tsx
```diff
@@ -18,6 +18,13 @@ const valueManager: PickerStateValueManager<unknown, unknown> = {
   parseInput: parsePickerInputValue,
   areValuesEqual: (utils: MuiPickersAdapter<unknown>, a: unknown, b: unknown) =>
     utils.isEqual(a, b),
+  mergeOldAndNewValues: (utils, prevValue, newValue) => {
```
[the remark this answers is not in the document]
flaviendelangle: Renamed
I keep `utils` 1st like on the other methods.
But feel free to change it later :+1:

PR #4436, opened 2022-04-10
[docs] Fix small external links issue

oliviertassinari:
Same spirit as https://github.com/mui/material-ui/pull/32212

There are more issues like this, I don't have the time to take care of them all.


siriwatknp (approved):
👍 Thanks for the fixes!

PR #4458, opened 2022-04-11
[l10n] Add Hungarian (hu-HU) locale

x22tri:
Hi,
Here's the Hungarian translation for the Data Grid. MUI Core already has Hungarian locale support but this repo didn't have any.
This is my first PR on any of the MUI projects--hope I did everything correctly. If not, please tell me.


alexfauquette (commented):
Thanks for your contribution

It sounds good. There are two lines for which you commented a translation. To pass tests, the file should either have
- commented lines with English text
- uncommented line with translation

Could you indicate if the translation is good and must be uncommented, or if we should keep the comment

x22tri:
The lines are translated, I just thought there was a special reason they were commented out in the version I copied from.

alexfauquette:
Nice, I tested it and it works nicely 👍. Should be available in the next release

Thanks again for the translation

PR #4465, opened 2022-04-12
[pickers] Add TDate generic to CalendarOrClockPicker component

flaviendelangle:
- [x] Add a generic `TDate` to `CalendarOrClockPickerProps`
- [x] Avoid complicated type casting for the `onViewChange` and `setOpenView`, the type itself being very simple I prefer to re-define it

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/date-helpers-hooks.tsx
+++ b/packages/x-date-pickers/src/internals/hooks/date-helpers-hooks.tsx
@@ -40,7 +40,7 @@ export function usePreviousMonthDisabled<TDate>(
 }
 
 export function useMeridiemMode<TDate>(
-  date: TDate,
+  date: TDate | null,
   ampm: boolean | undefined,
   onChange: PickerOnChangeFn<TDate>,
 ) {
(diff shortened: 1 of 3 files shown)


packages/x-date-pickers/src/internals/hooks/date-helpers-hooks.tsx
```diff
@@ -49,7 +49,8 @@ export function useMeridiemMode<TDate>(
 
   const handleMeridiemChange = React.useCallback(
     (mode: 'am' | 'pm') => {
-      const timeWithMeridiem = convertToMeridiem<TDate>(date, mode, Boolean(ampm), utils);
+      const timeWithMeridiem =
```
flaviendelangle, line 52: I think the methods in `date-io` actually handles the nullish value correctly most of the time
But there typing says otherwise, so I prefer to play safe here.
The date can be null, the `getHours` method used inside `convertToMeridiem` do not accept `null` so I don't call it when the date is null.

packages/x-date-pickers/src/internals/models/props/baseToolbarProps.ts
```diff
@@ -8,7 +8,7 @@ export interface BaseToolbarProps<TDate>
   extends ExportedCalendarPickerProps<TDate>,
     ExportedClockPickerProps<TDate> {
   ampmInClock?: boolean;
-  date: TDate;
+  date: TDate | null;
```
flaviendelangle, line 11: There is nothing preventing the date from being null, it is even the default value in most of our demos

PR #4472, opened 2022-04-12
[pickers] Remove aria-hidden attribute from week day label component

### LINKED ISSUE #4470: [DatePicker] Days of week are marked aria-hidden making them inaccessible to screen readers
### Duplicates

### Latest version

### Current behavior 😯

Location: Second row of Date Picker where days of the week are listed "S M T W T F S" are all marked aria-hidden:

<span class="MuiTypography-root MuiTypography-caption css-10qkmc3" aria-hidden="true">S</span>

### Expected behavior 🤔

Columns for day of week are relevant information so should not be marked aria-hidden.  This was flagged as a "critical" issue (one below "Blocker") by our a11y auditors.

Could you pls remove the aria-hidden attribute?

Here's where it is in the code:
Screen Shot 2022-02-15 at 8 54 03 AM
 

### Steps to reproduce 🕹

Steps:

1. Go to Date picker demo page "Basic Usage":  https://mui.com/components/date-picker/#basic-usage
2. Open up devtools and select any one of "S M T W T F S" 
3.  You'll see the aria-hidden attribute
Screen Shot 2022-02-15 at 8 57 30 AM

### Context 🔦

In middle of a11y audit by 3rd party reputable a11y firm. 

### Your environment 🌎

<details>

```
```
</details>

flaviendelangle:
Fixes #4470

@m4theushw you know the accessibility guidelines more than me
Do we agree that the request make sense ?


m4theushw (commented):
Only removing `aria-hidden` won't solve the problem. According to [this example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html), the calendar is represented by a `grid`. In this case, the days of the week are the column headers, so they should have `role=columnheader`. Additionally, each column header has `aria-label` with the complete day of the week, not the abbreviation.

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
```diff
@@ -335,7 +335,7 @@ const PickersDayRaw = React.forwardRef(function PickersDay<TDate>(
       centerRipple
       data-mui-test="day"
       disabled={disabled}
-      aria-label={!children ? utils.format(day, 'fullDate') : undefined}
+      aria-label={!children ? utils.format(day, 'fullDateWithWeekday') : undefined}
```
LukasTy, line 338:
```suggestion
```
This `aria-label` shouldn't even be needed once the roles are configured as suggested by ARIA example.

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -161,7 +161,12 @@ export function DayPicker<TDate>(props: PickersCalendarProps<TDate>) {
     <React.Fragment>
       <PickersCalendarDayHeader>
```
LukasTy, line 162:
```suggestion
    <div role="grid">
      <PickersCalendarDayHeader role="row">
```

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -161,7 +161,12 @@ export function DayPicker<TDate>(props: PickersCalendarProps<TDate>) {
...
         {utils.getWeekdays().map((day, i) => (
-          <PickersCalendarWeekDayLabel aria-hidden key={day + i.toString()} variant="caption">
+          <PickersCalendarWeekDayLabel
+            key={day + i.toString()}
+            variant="caption"
+            role="columnheader"
+            aria-label={day}
```
LukasTy, line 168:
```suggestion
            aria-label={utils.format(utils.addDays(startOfCurrentWeek, i), 'weekday')}
```

LukasTy:
I've gone through the change and it should correctly solve the issue mentioned.
But I've investigated the example provided by ARIA and noticed that our approach is quite a bit different and also a bit hacky (although it works better on MacOS VoiceOver).

But going with the implementation more closely resembling the example given, allows `NVDA` on Windows to work more correctly.

I can't provide all the needed changes via suggestion comments, hence I'll post them as snippets here.

Add the following on L159
```tsx
const startOfCurrentWeek = utils.startOfWeek(now);
```

```tsx
<PickersCalendarWeekContainer
  data-mui-test="pickers-calendar"
  ref={slideNodeRef}
  role="rowgroup" // instead of `grid`
>
```

Replace wrapping div with role of `cell` on L216 with only the day component:
```tsx
<PickersDay role="gridcell" key={pickersDayProps.key} {...pickersDayProps} />
```

And, of course, add closing `</div>` instead of `</React.Fragment>` on L226.

After these changes VoiceOver would read the days only as numbers, but more correct screen reader implementation like NVDA plays beautifully reading day of the month (column header).

What do you think about this change @flaviendelangle @m4theushw ?

PR #4480, opened 2022-04-13
[core] Remove `rowsCache` from state

m4theushw:


packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -105,26 +100,27 @@ const getRowsStateFromCache = (
...
-    state: convertGridRowsPropToState({
-      rows: props.rows,
-      getRowId: props.getRowId,
-      prevState: {
-        value: {
-          idRowsLookup: {},
-          ids: [],
-        },
-        rowsBeforePartialUpdates: [],
+  apiRef.current.rowsCache = convertRowsPropToState({
```
flaviendelangle, line 103: If we keep it in the apiRef, what do you think about doing `apiRef.current.caches.rows` ?
Or `apiRef.current.unstable_caches.rows` if we want to keep it private for now.

I would like to move `unstable_sanitizedModelOnLastRowTreeCreation` out of the state (same for the equivalent of the Aggregation)
Maybe it's worth scoping it instead of doing it on the root (same for the DOM refs btw)

For the alternative you gave in the PR description, it don't know how we could adapt it in the future to allow custom plugins with there own cache without touching the `useDataGridXXXComponent` hook.
But as long as the caches system is private, we can rewrite it to be compatible when we know what the plugins will actually look like

flaviendelangle:
I think we can go on with this one, #4208 will need a very similar implementation (I did a custom one for now to have something that works)

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
m4theushw: I went with `unstable_caches`. I hope that theme augmentation works to enrich the cache interface to allow custom plugins to have their own cache.

flaviendelangle (approved):

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
flaviendelangle: It should work perfectly

PR #4486, opened 2022-04-13
[pickers] Manage input value without using the focus

### LINKED ISSUE #4381: [DatePicker] using onBlur leads to inconsistency in the internal state
### Duplicates

### Latest version

### Current behavior 😯

Keep the `DatePicker` focused while using an `endAdornment` to clear its value. While clicking the clear button, the input fields value stays. The `onChange` event is triggered correctly, though. 

Problem appears since I added the followring to the clear buttons `onMouseDown` event:
 ```
const dontGrabFocusOnMouseDown = (event) => {
    event.preventDefault();
    event.stopPropagation();
  };
```

### Expected behavior 🤔

The input field value should be cleared as well. 
(hint: I use the same behavior with various versions `TextField` and `Autocomplete` where it works properly)

### Steps to reproduce 🕹

https://codesandbox.io/s/materialuipickers-material-demo-forked-s7umik?file=/demo.js

Steps to reproduce:
1. Focus the text field
2. Press the clear button

### Context 🔦

I have a customized `DatePicker` where I would like to add a clear button as `endAdornment` if focused (defined by our UX guideline). The input field of the `DatePicker` should stay focused such that I stopped propagating the mouse down event upwards clicking the clear button. This caused the issue.

### Your environment 🌎

```
  System:
    OS: macOS 12.3
  Binaries:
    Node: 16.14.2 - ~/work/customer/son/fe-repo/.direnv/node/bin/node
    Yarn: Not Found
    npm: 8.5.0 - ~/work/customer/son/fe-repo/.direnv/node/bin/npm
  Browsers:
    Chrome: 99.0.4844.84
    Edge: Not Found
    Firefox: 99.0
    Safari: 15.4
  npmPackages:
    @mui/base:  5.0.0-alpha.75
    @mui/icons-material: ^5.1.1 => 5.1.1
    @mui/lab: ^5.0.0-alpha.76 => 5.0.0-alpha.76
    @mui/material: ^5.1.1 => 5.6.0
    @mui/private-theming:  5.6.0
    @mui/styled-engine:  5.6.0
    @mui/styles: ^5.1.1 => 5.1.1
    @mui/system:  5.6.0
    @mui/types:  7.1.3
    @mui/utils:  5.6.0
    @types/react: ^17.0.34 => 17.0.34
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.4.4 => 4.4.4
```

### Order ID 💳 (optional)

_No response_

### LINKED ISSUE #4479: DatePicker (and presumably other pickers) not updated on value change if focused

### LINKED ISSUE #4485: [DatePicker] input cannot be cleared while focused
### Duplicates

### Latest version

### Current behavior 😯

If the input is focused, when programatically setting the value for the `DatePicker` component to `null`, the old value remains visible in the input

### Expected behavior 🤔

The input should be cleared when setting the value to `null`, even while focused

### Steps to reproduce 🕹

Steps:

1. Click in the 'name' input and type a valid string
2. Click in the 'Date of birth' input and type a valid date (e.g. type '01021990')
3. While still focused in the 'Date of birth' input, press 'Enter' to submit the form, which clears the values used by both inputs
4. Observe that the old 'Date of birth' remains visible, but the 'name' clears correctly

https://codesandbox.io/s/agitated-snow-p9xb2?file=/src/Demo.tsx

### Context 🔦

I am trying to clear form input after 'submitting' the form. If the user has their focus in the 'date of birth' field while submitting, the input value does not clear correctly after submission. If the user is focused in a normal `TextField`, or otherwise clicks the 'Submit' button directly, all fields clear as expected.

Note: mui/mui-x#4870 is a similar issue, but not the same issue. mui/mui-x#4870 is regarding the input not clearing for 'invalid' dates, while this issue occurs for both 'invalid' and 'valid' dates, as long as the field is focused.

### Your environment 🌎

<details>

```
  System:
    OS: Linux 5.11 Ubuntu 20.04.2 LTS (Focal Fossa)
  Binaries:
    Node: 12.18.3 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.11 - /usr/local/bin/npm
  Browsers:
    Chrome: 89.0.4389.90
    Firefox: 94.0
```
</details>

### LINKED ISSUE #4870: [pickers] Impossible to clear an invalid date
When a date picker contains an invalid date, and the value is set to null (outside of the date picker), the DOM input element value is not changed (and thus still displaying the old value).

## Current Behavior 😯

Input keeps the old value

## Expected Behavior 🤔

Input should be cleared

## Steps to Reproduce 🕹

https://codesandbox.io/s/datepicker-with-clear-button-nz6kh?file=/demo.js

Working scenario: valid date
1. Type a valid date
2. Click the clear button
3. The input is cleared

Bug scenario: invalid date
1. Type an invalid date (e.g. 35/33/2020)
2. Click the clear button
3. The input is not cleared, it keeps the old value

### PR DESCRIPTION
**Pull request by alexfauquette**
Related issues (are they fixed?)

- Fix #4485 -> to test manually: [codesandbox](https://codesandbox.io/s/basicdatepicker-material-demo-forked-4y4scj?file=/demo.tsx)
- Fix #4479 -> to test manually: [codesandbox](https://codesandbox.io/s/basicdatepicker-material-demo-forked-i3f43q?file=/demo.tsx)
- Fix #4381 -> to test manually: [codesandbox](https://codesandbox.io/s/materialuipickers-material-demo-forked-9fkz8e?file=/demo.js)
- Fix #4870 -> to test manually: [codesandbox](https://codesandbox.io/s/datepicker-with-clear-button-forked-vu93x1?file=/demo.js)

### The original problem

The `useMaskedInput` track the `input` focus state for the following reason (a comment can be found here https://github.com/mui/material-ui/pull/25972/files )

If a user enters `'01/01/1'` date adapters consider it as the 1st of January 0001. Then `onChange` is called with `new Date(1,1,1)`. Since we encourage developers to do the following, the TextField `value` will be set to `'01/01/0001'`, leading to really difficult editing behavior (impossible to delete a single character in the year).

```jsx
<DatePicker
	value={value}
	onChange={(newValue) => setValue(newValue)}
/>
```

### Why it is not a good solution

The assumption is

> If the input has focus, we should let what is typed as it is

But this behavior lets the components have two different renderings for the same props, leading to bug #4479 in which the dev wants to implement the following feature: "When I press <kbd>Del</kbd> it erases the date". 

The straightforward solution is to use `onKeyDow` and set the value to `null` if <kbd>Del</kbd> is pressed. But if the input has the focus the modification will be ignored by the component, and so the previous value will still be visible.

### Proposed solution

In this PR I remove the focus management since it leads to inconsistencies between the controlled value and the displayed one
 
Add `acceptInconsistentFormat` prop with the default set to `false`. an input is considered as an `inconsistentFormat` if parsing and formatting is not the identity function. For example with inputFormat `dd/MM/YYYY`, it makes the following transformations
```js
'10/12/1' -> Date(1, 11, 10) -> '10/12/0001' // inconsistent
'10/12/1994' -> Date(1994, 11, 10) -> '10/12/1994' // consistent
```
So with `acceptInconsistentFormat=false` the first input will not call `onChange`

### Why it does not work

We need to be able to desync the displayed text from the date input. Because we need to be able to type partial date which corresponds to invalid date

When filling the date we can wait to reach a complete one with the `acceptInconsistentFormat=true`
But if the user entered a date (for example `'01/01/1998'`) and deletes a character, it will now be `'01/01/199'` which should call `onChange` to say that the date is now invalid

### New solution

I observed that currently the `onChange` is called even with date such as `01/01/19` which is not formatted because the component has the focus. Now it will still call `onChange` with date `01/01/19` but not format the new date, because the current text already corresponds to the new date


packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
```diff
@@ -68,18 +68,29 @@ describe('<DesktopDatePicker />', () => {
...
-        label="Masked input"
-        inputFormat="dd/MM/yyyy"
-        value={adapterToUse.date('2018-01-01T00:00:00.000Z')}
-        onChange={onChangeMock}
-        InputAdornmentProps={{
-          disableTypography: true,
-        }}
-      />,
-    );
+    const Test = () => {
```
flaviendelangle, line 71: You should be able to use `WrappedDesktopDatePicker` which does exactly what you reimplement here (handled an internal state and expose an `onChange` if you want to track calls)

packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.test.tsx
```diff
@@ -48,16 +48,22 @@ describe('<DesktopDateTimePicker />', () => {
...
-      <DesktopDateTimePicker
-        renderInput={(params) => <TextField autoFocus {...params} />}
-        ampm={false}
-        inputFormat="mm.dd.yyyy hh:mm"
-        mask="__.__.____ __:__"
-        onChange={() => {}}
-        value={null}
-      />,
-    );
+    const Test = () => {
```
flaviendelangle, line 51: Same, you can use the wrapped version

packages/x-date-pickers/src/internals/hooks/useMaskedInput.tsx
```diff
@@ -59,44 +57,46 @@ export const useMaskedInput = <TInputDate, TDate>({
...
-  const valueChanged = previousInputValueRef.current !== currentInputValue;
-  // Update the input value only if the value changed outside of typing
-  if (notTyping && valueChanged && (rawValue === null || utils.isValid(rawValue))) {
-    if (currentInputValue !== innerInputValue) {
-      setInnerInputValue(currentInputValue);
-    }
+  const parsedValue = rawValue === null ? null : utils.date(rawValue);
+
+  // Track the value of the input
+  const [innerInputValue, setInnerInputValue] = React.useState(parsedValue);
```
flaviendelangle, line 63: Can you set the explicit generic on `useState` for those two states ?
It would be easier to read since it's a tricky part of the codebase

alexfauquette:
I added the "Fix" before each issue. If someone else also wants to experiment the bugs with this PR fix, I linked a codesandbox  for each issue.

The codesandbox is the initial minimal reproduction of the issue, with the package generated by this PR

Lukortech:
DatePicker in `@mui/x-date-pickers` does not update value onChange, still.

alexfauquette:
Hi @Lukortech, could you open an issue with a code sandbox demo of what you want to achieve and what is the unexpected behavior? Otherwise we will not be able to keep track of your bug

Issue #3571, opened 2022-01-08
[data grid] Excel column filters


satra2806:
### Summary 💡

Hi team, is it possible to create filter like Handsontable:

https://handsontable.com/docs/column-filter/#basic-configuration

<img width="504" alt="Screenshot 2022-01-24 at 15 51 56" src="https://user-images.githubusercontent.com/3165635/150805776-de460eac-6cf1-4ac7-be30-0ee2d4399e7d.png">

Google Spreadsheet

<img width="338" alt="Screenshot 2022-01-24 at 15 54 14" src="https://user-images.githubusercontent.com/3165635/150806147-4238e176-8ce5-4c92-9025-639b9dc32379.png">

https://www.ag-grid.com/javascript-data-grid/filter-set/

<img width="427" alt="Screenshot 2022-01-24 at 15 57 42" src="https://user-images.githubusercontent.com/3165635/150806720-1b39f01d-c19b-4b2e-8dfc-e00cf3a8583a.png">

m4theushw:
Not natively, #670 should implement it. Although, I was playing with a custom column menu and it's indeed possible to embedded a filter form inside the column menu. Here's the CodeSandbox: https://codesandbox.io/s/datagrid-v5-quick-start-forked-de2m6?file=/src/App.tsx

The only problem with this example is that when the filter runs the menu is closed. This happens because the `rowsScroll` event is fired every time that the number of visible rows changes. If the CodeSandbox satisfy what you're looking for, we could add a prop to disable the default behavior of closing the menu during scroll.

oliviertassinari:
Interesting, I have added more benchmark in the issues's description: It sounds like https://github.com/mui-org/material-ui-x/issues/201#issuecomment-732761412

michelengelen:
Closing this as a duplicate of #670, which tracks the same request: an Excel-style set filter with a checkbox list of the column's distinct values. Your benchmarks (Handsontable, Google Sheets, AG Grid) were already added to that issue's description back in 2022, so nothing from this thread is lost.

The feature is not implemented yet. Two building blocks exist in the meantime:

- [Header filters](https://mui.com/x/react-data-grid/filtering/header-filters/) (Pro, since v7) put a filter input on each column, and their inputs are customizable per column.
- The `isAnyOf` operator covers multi-value filtering on string and `singleSelect` columns.

[...]

PR #4498, opened 2022-04-14
[DataGrid] Don't close column menu when updating rows

### LINKED ISSUE #4139: [data grid] Column menu closes when rows are updated
### Duplicates

### Latest version

### Current behavior 😯

click on the 3dots on column header and your menu is open, only until data is refreshed, then it is automatically closed

### Expected behavior 🤔

menu should stay open

### Steps to reproduce 🕹

Steps:

1. go to https://codesandbox.io/s/35f6j?file=/demo.tsx:10053-10076
2. add this code to FullFeaturedDemo component: `React.useEffect(() => setInterval(loadNewData, 3000), [])`
3. try open any column menu: it won't stay open for long

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #4139 

Preview: https://deploy-preview-4498--material-ui-x.netlify.app/x/react-data-grid/

CodeSandbox: https://codesandbox.io/s/fullfeatureddemo-material-demo-forked-98j7cp?file=/demo.tsx

I implemented the solution from https://github.com/mui/mui-x/issues/4139#issuecomment-1092047815. I chose this one because it also unlocks the possibility to have filters inside the column menu. Previously, we couldn't because the column menu closes when filtering: https://github.com/mui/mui-x/issues/3571#issuecomment-1009524740

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/columnHeaders.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/columnHeaders.DataGridPro.test.tsx
@@ -65,8 +65,7 @@ describe('<DataGridPro /> - Column Headers', () => {
       clock.runToLast();
       expect(screen.queryByRole('menu')).not.to.equal(null);
       const virtualScroller = document.querySelector('.MuiDataGrid-virtualScroller')!;
-      virtualScroller.scrollTop = 50;
-      virtualScroller.dispatchEvent(new Event('scroll'));
+      fireEvent.wheel(virtualScroller);
       clock.runToLast();
       expect(screen.queryByRole('menu')).to.equal(null);
     });
--- a/packages/grid/x-data-grid-pro/src/tests/columnHeaders.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/columnHeaders.DataGridPro.test.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 // @ts-ignore Remove once the test utils are typed
 import { createRenderer, fireEvent, screen } from '@mui/monorepo/test/utils';
 import { expect } from 'chai';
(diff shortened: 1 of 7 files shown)


flaviendelangle (approved):

alexfauquette:
I'm not sure this solution fix the following scenario when we will try to implement column menu filtering https://github.com/mui/mui-x/issues/3571#issuecomment-1009524740

- scroll along the grid
- open the menu and start filtering

If there are fewer rows passing the filter than where you currently are, the grid will scroll up and the close the column menu

I do not see other solution for this scenario than adding a value to scroll event indicating its origin: 'native', 'content modification', ...

m4theushw:
> If there are fewer rows passing the filter than where you currently are, the grid will scroll up and the close the column menu

@alexfauquette I was considering this an edge case. We could check how much was scrolled since the last event. Anyway, I changed to listen to the `wheel` event.

alexfauquette (approved):
Looks good. Notice that `onWheel` will not be triggered when using the scroll bar


https://user-images.githubusercontent.com/45398769/164170695-3c1a024d-31fc-4f28-9e59-6b8f3f988019.mp4

flaviendelangle (commented):
`onWheel` is not called when scrolling on a tactile device.

m4theushw:
> onWheel is not called when scrolling on a tactile device.

@flaviendelangle Which tactile device do you refer to? Scrolling in a touchpad is like using the mouse wheel, so it emits `wheel`. https://stackoverflow.com/a/60242837

flaviendelangle:
I am not talking about a Touchpad, but directly scrolling by swiping on a touch screen.
I tested the demo link on my phone, scrolling the rows by swiping does not close the column menu.
We would need to listen to `onTouchMove` for instance.

flaviendelangle (approved):
I now work on my phone :+1:

Issue #4487, opened 2022-01-04
[DateTimePicker] Mask with non numeric data are not applied


cattre:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Using the DateTimePicker component, a warning is displayed in the console about the mask not matching the input value:

`The mask "__/__/____ __:__" you passed is not valid for the format used dd MMMM yyyy HH:mm. Falling down to uncontrolled not-masked input. `

No mask has been provided, so I assume this is using the default one.

The spaces instead of slashes are one issue, but adding a custom mask of '__ __ ____ __:__' resolves this.

It appears there is an issue with the full month name (MMMM) as a numeric month (MM) works fine (and '__ ____ ____ __:__' doesn't help).

### Expected behavior 🤔

Custom date formats do not throw warnings in the console

### Steps to reproduce 🕹

Live example here:

https://codesandbox.io/s/elegant-nobel-lqyol?file=/src/App.js:518-536

### Context 🔦

Similar to issue posted here - https://github.com/mui-org/material-ui-pickers/issues/1776 - but this repo has been archived so I'm unsure if the answers are still relevant with MUI 5.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

nimaes80:
In my case first I set `LocalizationProvider` in root of peroject
```
import { AdapterMoment } from "@mui/x-date-pickers/AdapterMoment";
import AdapterJalali from "@date-io/date-fns-jalali";
import { LocalizationProvider } from "@mui/x-date-pickers/LocalizationProvider";
// ...

<LocalizationProvider dateAdapter={i18n.language === "fa" ? AdapterJalali : AdapterMoment}>
// ...
</LocalizationProvider>
```
then I use `DatePicker` like blow
```
<DatePicker
	mask="____/__/__"
	value={date}
	onChange={(newValue) => setDate(newValue)}
	renderInput={(params) => <TextField {...params} />}
/>
```
note that I use `@date-io/date-fns-jalali` for Jalali Time before that I see that yellow box also time dosnt show correctly with other library's

woorykim:
https://github.com/mui/mui-x/issues/4487#issuecomment-1097731043
Thanks to you, I solved it👍 Thank you!
```js
mask=""
[...]

PR #4501, opened 2022-04-14
[pickers] Add details in invalid mask error

alexfauquette:
Part of #4487

I customize the error message depending on the main problems encountered in the issue

I'm wondering if we should write a doc page to explain how format and mask can be customized

### DIFF
--- a/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
+++ b/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
@@ -68,7 +68,7 @@ describe('text-field-helper', () => {
         expect(runMaskValidation).toWarnDev(
           [
             `The mask "${mask}" you passed is not valid for the format used ${format}.`,
-            `Falling down to uncontrolled not-masked input.`,
+            `Falling down to uncontrolled no-mask input.`,
           ].join('\n'),
         );
       }
--- a/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
+++ b/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
@@ -66,7 +66,10 @@ describe('text-field-helper', () => {
         expect(runMaskValidation()).to.be.equal(true);
       } else {
         expect(runMaskValidation).toWarnDev(
-          `The mask "${mask}" you passed is not valid for the format used ${format}. Falling down to uncontrolled not-masked input.`,
+          [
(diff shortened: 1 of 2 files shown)


packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
```diff
@@ -66,7 +66,10 @@ describe('text-field-helper', () => {
         expect(runMaskValidation()).to.be.equal(true);
       } else {
         expect(runMaskValidation).toWarnDev(
-          `The mask "${mask}" you passed is not valid for the format used ${format}. Falling down to uncontrolled not-masked input.`,
+          [
+            `The mask "${mask}" you passed is not valid for the format used ${format}.`,
+            `Falling down to uncontrolled not-masked input.`,
```
flaviendelangle, line 71:
```suggestion
            `Falling down to uncontrolled un-masked input.`,
```

or

```suggestion
            `Falling down to uncontrolled no-mask input.`,
```

packages/x-date-pickers/src/internals/utils/text-field-helper.ts
```diff
@@ -79,9 +79,33 @@ export function checkMaskIsValidForCurrentFormat(
...
   if (!isMaskValid && utils.lib !== 'luxon' && process.env.NODE_ENV !== 'production') {
-    console.warn(
-      `The mask "${mask}" you passed is not valid for the format used ${format}. Falling down to uncontrolled not-masked input.`,
-    );
+    const defaultWarning = [
+      `The mask "${mask}" you passed is not valid for the format used ${format}.`,
+      `Falling down to uncontrolled no-mask input.`,
+    ];
+
+    if (format.includes('MMM')) {
```
flaviendelangle, line 87: Not sure to understand why we do this `includes('MMM')` check

flaviendelangle: OK, is this incompatibility impossible to change (in the future) or is it just a quick fix waiting for a rework ?

We are in the `if(utils.lib == 'luxon')` is it intended ?

alexfauquette: It's not impossible. For later we could imagine that typing `'1 j'` proposes to complete with `'1 jun'` `'1 jul'` or `'1 jan'` but not for now.

Seems that luxon library was causing problems. I don't know why they decided to do hide the warning message for this specific library

flaviendelangle (approved):

PR #4502, opened 2022-04-14
[DataGrid] Remove `aria-label` when `title` is defined

alexfauquette:
From https://github.com/mui/mui-x/pull/4317#discussion_r849598128

### DIFF
--- a/docs/data/data-grid/filtering/QuickFilteringGrid.tsx
+++ b/docs/data/data-grid/filtering/QuickFilteringGrid.tsx
@@ -35,7 +35,6 @@ function QuickSearchToolbar(props: QuickSearchToolbarProps) {
           endAdornment: (
             <IconButton
               title="Clear"
-              aria-label="Clear"
               size="small"
               style={{ visibility: props.value ? 'visible' : 'hidden' }}
               onClick={props.clearSearch}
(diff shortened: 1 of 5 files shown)


docs/data/data-grid/filtering/QuickFilteringGrid.tsx
```diff
@@ -35,7 +35,6 @@ function QuickSearchToolbar(props: QuickSearchToolbarProps) {
           endAdornment: (
             <IconButton
               title="Clear"
-              aria-label="Clear"
```
m4theushw, line 38: I would invert here. Keep `aria-label` but remove `title`. We lose the tooltip but the function of a clear button next to an input is very obvious. The demos for `IconButton` only have `aria-label`: https://mui.com/material-ui/react-button/#icon-button

m4theushw (approved):
I propose to keep `aria-label` + `title` in all components, only removing `title` from the quick filtering demo. 

`title` is not reliable: https://www.w3.org/WAI/tutorials/forms/labels/#using-the-title-attribute

By doing the opposite, keeping only `aria-label`, we lose the tooltip. This is okay-ish for buttons whose function is obvious but for others the title is necessary.

Issue #4503, opened 2022-04-14
DateRangePicker clearable does not show clear button


zigang93:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

no clear button at DateRangePicker
but it work fine with DatePicker

https://codesandbox.io/s/basicdatepicker-material-demo-forked-r6zgui?file=/demo.tsx

### Expected behavior 🤔

have clear button

### Steps to reproduce 🕹

sandbox link: https://codesandbox.io/s/basicdatepicker-material-demo-forked-r6zgui?file=/demo.tsx

flaviendelangle:
Same problem on `DesktopTimePicker`, fixing in #4408

PR #4510, opened 2022-04-14
[DateRangePicker] fix missing clearable and clearText props

zigang93:
fix https://github.com/mui/mui-x/issues/4503

currently I am using free version of DateRangePicker from @mui/lab ^5.0.0-alpha.69
can I push the fix to old mui lab ?


zigang93:
wrong branch

PR #4517, opened 2022-04-15
[pickers] Add l10n support

### LINKED ISSUE #4500: [pickers] Allows community to add new translations
### Duplicates

### Latest version

### Current behavior 😯

On the DatePicker component, there are labels like "Previous month", used to provide a label for the low-vision and blind users using screen reader (accessibility). They are always display in English even if the local is for example French.

MUI provides natively no translation for those label and it seems there is no native way to contribute to translate them.

We can translate them on our side but it'd be better to provide them on the repo to benefit to all.

### Expected behavior 🤔

MUI would provide native translation for accessibility labels or provide a way for contributor to help translating them.

### Steps to reproduce 🕹

Steps:
1. Open a non-English page with a DatePicker in it.
2. Open the date picker calendar, check it is in your locale.
3. Inspect the "previous month" button, check if the string, attribute aria-label on the button, is in your locale.

Result:
It stays on English locale.

Expected result:
It should be in the user's locale.

### Context 🔦

When you're a low-vision or a blind person using screen reader or a motor disabled person using speech recognition, you rely on label. For screen reader users, it is to understand the meaning of the button, and for speech recognition users, it's to indicate to your tool where to click.

### Your environment 🌎

* Microsoft Windows 10.0.19041.1645
* Microsoft Edge 100.0.1185.39

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #4500

new doc page: https://deploy-preview-4517--material-ui-x.netlify.app/x/react-date-pickers/localization/

## Initial solution

To support localization we could define an object with `defaultProps` to provide needed translations at all components.

I wanted to have something similar to the DataGrid: one file per location. The problem is that we have much more than one component here and it's hard to list for each component which translation key do they need.

In this draft PR What I propose to do is to add a `localText` prop to every component we export and set the default value with the `ThemeProvider`

Another option could be to Create a fake component `'MuiPickers'` of `'MuiPickersLocale'` such that components reads default props of both as follow

```jsx
const inPropsWithLocale = useThemeProps({ props: inProps, name: 'MuiPickers' });
const props = useThemeProps({ props: inPropsWithLocale, name });
```

## Current Solution

The `LocalizationProvider` accepts a new prop: `localeText` that contains all the translations

Components that were setting translation by default have now their props deprecated. If no prop is provided, the local text is obtained from `useLocaleText`. All keys are available in `enUS` and contributors can add other translations

About where the defaults are set: 
- the left/right arrows text depends on the component. For calendars it is "prev/next month" for clocks it is "prev/next view" so the default must be set in the parent component
- the cancel, ok, start,  ... were sometimes set in the exported component, sometime in the internal. SInce they do not carry, I put them into the internal component to check deprecated props are not used only once

### Works for a follow-up PR

- [ ] Support functions such as the `defaultGetClockLabelText`
- [ ] Adapt `yarn l10n` script for the new files


m4theushw:
The DataGrid has a single entry, the `DataGrid`, which allows users to change all translations in one place. This doesn't happen here, you can use `DatePicker` but also only `MobileDatePicker`. What is common to all is the `LocalizationProvider` wrapping them. What about adding `localeText` to it?

> Another point : should we drop the prop (rightArrowButtonText for instance) ?

Yes, in the DataGrid we don't allow to override texts with props.

alexfauquette:
Following @m4theushw  recommendation, I moved the `localeText` prop to the `LocalizationProvider` component.

This allows to update the text as follow

```jsx
const theme = createTheme(frFR)
<ThemeProvider theme={theme}>
```
or
```jsx
<LocalizationProvider dateAdapter={AdapterDateFns}
      localeText={frFR.components.MuiLocalizationProvider.defaultProps.localeText}
>
      ```

File: .eslintrc.js [hunk 1]
```diff
@@ -61,7 +61,7 @@ module.exports = {
       parserOptions: { tsconfigRootDir: __dirname, project: ['./tsconfig.json'] },
     },
     {
-      files: ['docs/src/pages/components/**/*.js', 'docs/src/pages/components/**/*.tsx'],
+      files: ['docs/data/**/*.js', 'docs/data/**/*.tsx'],
```
alexfauquette, line 64: Just realized this rule was not applied

flaviendelangle:
@alexfauquette the title still contain `Draft` but the PR is ready for review, can I review it ?

flaviendelangle:
Truth is, I would delete the current props instead of deprecating them
I think we are changing to much stuff to handle a real deprecation period.

alexfauquette:
When removing props, I realized I was breaking the action buttons, because they rely on the definition of `cancelText` and `okText` to know if they should appear 

Need #4655 to be done in order to distinguish the button content and if it should be rendered

docs/data/date-pickers/localization/LocalizedDatePicker.tsx
```diff
@@ -0,0 +1,58 @@
...
+  en: '__/__/____',
+  ru: '__.__.____',
+  de: '__.__.____',
+};
+
+export default function LocalizedDatePicker() {
+  const [locale, setLocale] = React.useState<keyof typeof maskMap>('ru');
+  const [value, setValue] = React.useState<Date | null>(new Date());
+
+  const selectLocale = (newLocale: any) => {
```
flaviendelangle, line 31: We can probably type it correctly

Create a `LocalKey` type `type LocalKey = keyof typeof maskMap` and use it both for the state init and here

docs/data/date-pickers/localization/LocalizedDatePicker.tsx
```diff
@@ -0,0 +1,58 @@
...
+    setLocale(newLocale);
+  };
+
+  return (
+    <LocalizationProvider dateAdapter={AdapterDateFns} locale={localeMap[locale]}>
```
flaviendelangle, line 36: Should we name it more explicitly to avoid any confusion with the `localeText` ? `dateLocale` for instance ?

docs/data/date-pickers/localization/localization.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Pickers - Localization
+---
+
+# Pickers - Localization
+
+<p class="description">Pickers allow to support users from different locales, with formatting, RTL, and localized strings.</p>
```
flaviendelangle, line 7:
```suggestion
<p class="description">Date and time pickers allow to support users from different locales, with formatting, RTL, and localized strings.</p>
```

The official naming of the project is "Date and time pickers"
I think we can shortcut with "pickers" when inside a page that already clearly scoped the topic.
But for a title / description we should probably be explicit

____

Same for the title I think

On the Getting Start page we have "Date / Time pickers" which is not great I think (https://mui.com/x/react-date-pickers/getting-started/)

@samuelsycamore I you have a preference

docs/data/date-pickers/localization/localization.md
```diff
@@ -0,0 +1,105 @@
...
+
+The default locale of MUI is English (United States). If you want to use other locales, follow the instructions below.
+
+## Localization
```
flaviendelangle, line 11:
```
# Pickers - Localization

## Localization
```

The h2 title can probably be removed and the h3 converted to h2.

docs/data/date-pickers/localization/localization.md
```diff
@@ -0,0 +1,105 @@
...
+
+Localization can impact pickers components rendering in two distincts ways: The date format, and the components attributes such as `aria-label`.
+
+### Date-engine locale
```
flaviendelangle, line 15: Not a fan a `date-engine` but I don't have a better naming

docs/data/date-pickers/localization/localization.md
```diff
@@ -0,0 +1,105 @@
...
+
+Use `LocalizationProvider` to change the date-engine locale that is used to render pickers. Here is an example of changing the locale for the `date-fns` adapter:
+
+{{"demo": "LocalizedDatePicker.js"}}
+
+### Translation keys
+
+As the rest of MUI components, you can modify text and translations.
```
flaviendelangle, line 23: Would be great to create a new issue with auto update like for the grid (or re-use the same one)

In another PR of course

docs/data/date-pickers/localization/localization.md
```diff
@@ -0,0 +1,105 @@
...
+
+const theme = createTheme(
+  {
+    palette: {
+      primary: { main: '#1976d2' },
+    },
+  },
+  bgBG, // x-date-pickers translations
+  dataGridBgBG, // x-data-grid translations
+  coreBgBG, // core translations
```
flaviendelangle, line 70: Mayber add the pickers translations to the sibling example in the data grid doc
https://mui.com/x/react-data-grid/localization/#locale-text

packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
```diff
@@ -116,4 +142,19 @@ LocalizationProvider.propTypes = {
    * Locale for the date library you are using
    */
   locale: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
+  /**
+   * Locale for components texts
+   */
+  localeText: PropTypes.shape({
```
flaviendelangle, line 148: On the datagrid we don't generate the shape (I don't know why because `localeText` is not in the blacklist of `generateProptypes` (cc @m4theushw)
We should probably have `PropTypes.object` here as well

packages/x-date-pickers/src/internals/utils/warning.ts
```diff
@@ -0,0 +1,20 @@
+export const buildDeprecatedPropsWarning = (message: string | string[]) => {
+  let alreadyWarned = false;
+
+  const cleanMessage = Array.isArray(message) ? message.join('\n') : message;
```
flaviendelangle, line 4: Move this line after the env check

packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.tsx
```diff
@@ -116,6 +116,7 @@ DesktopDatePicker.propTypes = {
   /**
    * Clear text message.
    * @default 'Clear'
+   * @deprecated
```
flaviendelangle, line 119: Maybe add a message on the deprecation for each prop

```suggestion
   * @deprecated Use the `localeText` prop of `LocalizationProvider` instead, see [link to the doc]
```

packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
```diff
@@ -10,6 +13,8 @@ export interface MuiPickersAdapterContextValue<TDate> {
   };
 
   utils: MuiPickersAdapter<TDate>;
+
```
flaviendelangle, line 16:
```suggestion
```

packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
```diff
@@ -33,13 +38,27 @@ export interface LocalizationProviderProps {
...
+export function LocalizationProvider(inProps: LocalizationProviderProps) {
+  const props = useThemeProps({ props: inProps, name: 'MuiLocalizationProvider' });
+
+  const {
+    children,
+    dateAdapter: Utils,
+    dateFormats,
+    dateLibInstance,
+    locale,
+    localeText = {},
```
flaviendelangle, line 59: If you define the default value here, then `contextValue` memoization does not work when there is no locale.

```ts
        ...enUS.components.MuiLocalizationProvider.defaultProps.localeText,
        ...localeText ?? {},
```

That would be better I think

packages/x-date-pickers/src/locales/utils/pickersLocaleTextApi.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export interface PickersLocaleText {
+  previousMonth: string;
+  nextMonth: string;
+  openPreviousView: string;
+  openNextView: string;
+  cancel: string;
+  clear: string;
+  ok: string;
+  today: string;
```
flaviendelangle, line 12: Should we be more precise with the labels ?

```suggestion
  todayButtonLabel: string;
```

flaviendelangle (commented):
Great work !
That will be so much cleaner with it

docs/data/date-pickers/localization/LocalizedDatePicker.tsx
alexfauquette: What about `adapterLocale` because it is the argument passed to the adapter

flaviendelangle: Works for me :+1:

docs/data/date-pickers/localization/localization.md
alexfauquette: Good idea

packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
flaviendelangle: Maybe add `localeText` to the black list then

docs/data/date-pickers/localization/localization.md
alexfauquette: Issue created #4914

PR #4570, opened 2022-04-19
[docs] Fix CodeSandbox links for demo with pickers

alexfauquette:
Preview: https://deploy-preview-4570--material-ui-x.netlify.app/x/react-date-pickers/date-range-picker/

### DIFF
--- a/docs/pages/_app.js
+++ b/docs/pages/_app.js
@@ -76,6 +76,8 @@ ponyfillGlobal.muiDocConfig = {
       '@mui/x-data-grid-pro': getMuiPackageVersion('x-data-grid-pro', muiCommitRef),
       '@mui/x-data-grid-generator': getMuiPackageVersion('x-data-grid-generator', muiCommitRef),
       '@mui/x-data-grid': getMuiPackageVersion('x-data-grid', muiCommitRef),
+      '@mui/x-date-pickers': getMuiPackageVersion('x-date-pickers', muiCommitRef),
+      '@mui/x-date-pickers-pro': getMuiPackageVersion('x-date-pickers-pro', muiCommitRef),
       'date-fns': 'latest',
     };
     return output;


docs/pages/_app.js
```diff
@@ -60,6 +60,7 @@ ponyfillGlobal.muiDocConfig = {
 
     if (newDeps['@mui/x-date-pickers']) {
       newDeps['@mui/material'] = versions['@mui/material'];
+      newDeps['date-fns'] = 'latest';
```
alexfauquette, line 63: Demos use `data-fns` but it is not installed in the dependencies, so I added `date-fns: 'latest'`

docs/pages/_app.js
```diff
@@ -75,6 +76,7 @@ ponyfillGlobal.muiDocConfig = {
       '@mui/x-data-grid-pro': getMuiPackageVersion('x-data-grid-pro', muiCommitRef),
       '@mui/x-data-grid-generator': getMuiPackageVersion('x-data-grid-generator', muiCommitRef),
       '@mui/x-data-grid': getMuiPackageVersion('x-data-grid', muiCommitRef),
+      '@mui/x-date-pickers': getMuiPackageVersion('x-date-pickers', muiCommitRef),
```
alexfauquette, line 79: `versions['@mui/x-date-pickers']` was not defined, so it was reseting the value

docs/pages/_app.js
m4theushw: I also fixed this in #4508

alexfauquette: @m4theushw I let you add the second line such that pro demo works too, and I close this PR?

docs/pages/_app.js
m4theushw: You also need to set the version for `@mui/x-date-pickers-pro`. Look https://codesandbox.io/s/kd23vd?file=/package.json It's always `latest`.

docs/pages/_app.js
m4theushw: I only added to `@mui/x-date-pickers` because even Pro demos need it to import `AdapterDateFns`. This PR is still useful to fix

m4theushw (approved):

docs/pages/_app.js
simon-attila-fr: Hello, the issue still persists, I see an error message when I open the demo in codesandbox:
https://codesandbox.io/s/3ruukc?file=/demo.tsx

I added date-fns dependency to my fork, it works now.

PR #4575, opened 2022-04-19
[docs] Remove legacy pages for old URLs

m4theushw:
Preview: https://deploy-preview-4575--material-ui-x.netlify.app/x/react-data-grid/

The new docs structure is live. We don't need to keep the pages for the old structure. I ran `yarn docs:post-migration` and removed the feature toggles.


siriwatknp (approved):
👍 Thanks for the cleanup

Issue #4466, opened 2022-04-12
[pickers] Use components slots on every component


flaviendelangle:
## Intended behavior

- Replace all the `props.XXXComponent` and `props.XXXProps` by slots
- Always have a `componentsProps` slot matching the `components` slot
- Use `useSlotProps` and `SlotComponentProps` for every slot

## Related issues 

- #4557

## Components

I grouped the slots, we can do one PR per group.
You can take example on #5890 which clean the slots for the `PickerArrowSwitcher` component

### Slots of `PickerArrowSwitcher` (#5890)

Slots `LeftArrowButton`, `RightArrowButton`, `LeftArrowIcon`, `RightArrowIcon`

- [x] Replace all the `props.XXXComponent` and `props.XXXProps` by slots
- [x] Always have a `componentsProps` slot matching the `components` slot
- [x] Use `useSlotProps` and `SlotComponentProps` for every slot

### Slots of `PickersCalendarHeaderSlotsComponent` (#6943)

Slots: `SwitchViewButton`, `SwitchViewIcon`

- [x] Replace all the `props.XXXComponent` and `props.XXXProps` by slots
- [x] Always have a `componentsProps` slot matching the `components` slot (missing `switchViewIcon`)
- [x] Use `useSlotProps` and `SlotComponentProps` for every slot

### Slots of `PureDateInput` / `KeyboardDateInput`

Slots: `OpenPickerIcon`, `InputAdornment` (current `props.InputAdornmentProps`), `OpenPickerButton` (current `props.OpenPickerButtonProps`), `Input` (current `props.InputProps`)

**REMOVED ON V6**

### Slots of `PickersViewLayout` (#6381)

Slots: `Paper` (current `props.PaperProps`), `Popper` (current `props.PopperProps`), `Transition` (current `props.TransitionComponent`), `PaperContent`, `ActionBar`, `TrapFocus` (current `props.TrapFocusProps`), `Dialog` (current `props.DialogProps`), `Tabs` and `Toolbar` (current `props.ToolbarComponent`)

- [x] Replace all the `props.XXXComponent` and `props.XXXProps` by slots
- [x] Always have a `componentsProps` slot matching the `components` slot
- [ ] Use `useSlotProps` and `SlotComponentProps` for every slot

PR #4589, opened 2022-04-21
[pickers] Add explicit interfaces for components slots and components slots props

flaviendelangle:
Part of #4466

## Goal

Clean the typing of the `components` and `componentsProps` props before expanding the usage of the pattern in future PRs.

- [x] The `XXXSlotsComponent` and `XXXSlotsComponentsProps` are not nullable, it is the `XXXProps` interface which has a `Partial` (like on the DataGrid)

- [x] If the `components` or `componentsProps` is defined in one of the interface extended by the current component AND this interface is not tightly linked to the new component, then redefine explicitly `components` and `componentsProps` (ie: do not redefine on `DesktopDatePickerProps` if defined on `BaseDatePickerProps`, but redefine on `CalendarPickerProps` if defined on `PickersCalendarHeader`)

## What's next ?

- Add JSDoc to the existing `components` and there `componentsProps`
- Add a `componentsProps` for each existing `components` like on the DataGrid
- Move the `props.XXXXProps` and `props.XXXXComponent` (`PopperProps`, `TransitionComponent`, ...) to the slot pattern
- Improve doc generation for the slots of the pickers


packages/x-date-pickers-pro/src/StaticDateRangePicker/StaticDateRangePicker.tsx
```diff
@@ -7,72 +7,17 @@ import {
...
-import { ExportedDateRangePickerInputProps } from '../DateRangePicker/DateRangePickerInput';
+import { useDateRangeValidation } from '../internal/hooks/validation/useDateRangeValidation';
+import { DateRangePickerView } from '../DateRangePicker/DateRangePickerView';
 import { parseRangeInputValue } from '../internal/utils/date-utils';
 import { getReleaseInfo } from '../internal/utils/releaseInfo';
+import { BaseDateRangePickerProps } from '../DateRangePicker/shared';
 
 const releaseInfo = getReleaseInfo();
 
-interface BaseDateRangePickerProps<TDate>
```
flaviendelangle, line 29: This interface was a copy paste of the one in `DateRangePicker/shared`

packages/x-date-pickers/src/internals/models/props/basePickerProps.tsx
```diff
@@ -7,7 +7,7 @@ export interface BasePickerProps<TInputValue, TDateValue> {
    */
   className?: string;
   /**
-   * If `true` the popup or dialog will immediately close after submitting full date.
+   * If `true` the popup or dialog will not immediately close after submitting full date.
```
flaviendelangle, line 10: Not related but this description was wrong as well

PR #4591, opened 2022-04-21
[docs] Add plan in the nav bar for pro-only and premium-only pages

flaviendelangle:
Closes #4581

image

Preview: https://deploy-preview-4591--material-ui-x.netlify.app/x/react-data-grid/

PR #4600, opened 2022-04-21
[docs] Split the 'Columns' page

flaviendelangle:
Split the huge [Columns](https://mui.com/x/react-data-grid/columns)

[Doc preview](https://deploy-preview-4600--material-ui-x.netlify.app/x/react-data-grid/)

⚠ Some pages are not very detailed, but the goal of this PR is to do the split, not improve all the content. We should only focus on improving the title structure / page description for now and see later to add examples.

### New pages

Under parenthesis I added all the sections of the current *Columns* page that are being moved to the new page

- [x] Column definition (Column types, Custom column types)

  - [ ] Clean content (I do it in this PR because it is very outdated and does not make sense now that it is a standalone page)

- [x] Column dimensions (Width, Minimum width, Fluid width and Resizing)

- [x] Column visibility (Column visibility)

- [x] Column header (Headers, Render header, Styling header, Column menu) 

- [x] Column ordering (Column reorder)

I did not name it *Column reordering* because I think we should add a section about `GridInitialState.columns.orderedFields` 

- [x] Column pinning (Column pinning)

  - [ ]  Add pro icon after #4591 is merged

  - **Later**: Should have the same structure as other controllable model (one section "Pass pinned columns to the grid" with two children "Initialize the pinned columns" and "Controlled pinned columns" like we do for [other models](https://mui.com/x/react-data-grid/sorting/#pass-sorting-rules-to-the-grid). Not sure a "Structure of the model" is usefull here

- [x] Column spanning (Column spanning)

  - [ ]  Add pro icon after #4591 is merged

- [x] Column groups (Column groups)

  In AG Grid it is under *Accessories*.
  Not sure where it should leave in our documentation.
  Those sections should probably be reworked in depth
  
- [x] Cells (Value getter Value setter, Value formatter, Value parser, Render cell, Styling cell)

### Points to review

- Description of the new pages
- Name of the new pages (should it be *Column Pinning* or *Column pinning* ?)
- Name of the new pages in the nav bar (should it be *Column Pinning* or *Column pinning* ?)


docs/data/data-grid/column-visibility/column-visibility.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Data Grid - Column Visibility
+---
+
+# Data Grid - Column Visibility
+
+<p class="description">Decide which column should be visible.</p>
```
joserodolfofreitas, line 7:
```suggestion
<p class="description">Define which columns are visible.</p>
```

docs/data/data-grid/column-spanning/column-spanning.md
```diff
@@ -0,0 +1,67 @@
+---
+title: Data Grid - Column Spanning
+---
+
+# Data Grid - Column Spanning
+
+<p class="description">Span cells across several columns with column spanning.</p>
```
joserodolfofreitas, line 7:
```suggestion
<p class="description">Span cells across several columns.</p>
```

joserodolfofreitas (commented):
This feels like a great improvement as is. 
But perhaps we can consider making "Columns" and "Group & Pivot" (and future sections) collapsible. Otherwise as we split more sections, the navBar experience will probably worsen. 
What do you think?

Btw, I love that now we can see the future features with the icon on the nav bar.
Would it be too crowded if we add Pro and Premium icons to the features as well? The idea is to show the availability of the feature as earliest as possible, before the user clicks on the item only to see he cannot use it.

> Description of the new pages
I left some suggestions in line. 
Column groups is hitting a 404.

> Name of the new pages (should it be Column Pinning or Column pinning ?)
+1 for "Column pinning"

> Name of the new pages in the nav bar (should it be Column Pinning or Column pinning ?)
+1 for "Column pinning"

docs/data/pages.ts
```diff
@@ -15,8 +15,21 @@ const pages = [
...
+        pathname: '/x/react-data-grid/columns',
+        children: [
+          { pathname: '/x/react-data-grid/column-definition' },
+          { pathname: '/x/react-data-grid/column-dimensions' },
+          { pathname: '/x/react-data-grid/column-visibility' },
+          { pathname: '/x/react-data-grid/column-header' },
+          { pathname: '/x/react-data-grid/column-ordering' },
+          { pathname: '/x/react-data-grid/column-pinning' },
+          { pathname: '/x/react-data-grid/column-spanning' },
+          { pathname: '/x/react-data-grid/column-groups 🚧' },
```
alexfauquette, line 28:
```suggestion
          { pathname: '/x/react-data-grid/column-groups' },
```

flaviendelangle:
> Would it be too crowded if we add Pro and Premium icons to the features as well? The idea is to show the availability of the feature as earliest as possible, before the user clicks on the item only to see he cannot use it.

It's in another PR: https://github.com/mui/mui-x/pull/4591

flaviendelangle:
@joserodolfofreitas descriptions and title updated :+1: 

Where would you put the Column Selector and Column Menu sections ? (@m4theushw)
Do we need the Column Selector section in the 1st place ? It is described in the feature pages it impacts, not sure this global section serves a lot.

docs/data/data-grid/column-definition/column-definition.md
```diff
@@ -0,0 +1,148 @@
...
+
+{{"demo": "CustomColumnTypesGrid.js", "bg": "inline"}}
+
+> ⚠ If an unsupported column type is used the `string` column type will be used instead.
+
+## Column menu
+
+By default, each column header displays a column menu. The column menu allows actions to be performed in the context of the target column, e.g. filtering. To disable the column menu, set the prop `disableColumnMenu={true}`.
+
+{{"demo": "ColumnMenuGrid.js", "bg": "inline"}}
```
alexfauquette, line 115: This section could move to the "Column header" page

flaviendelangle: Done

docs/data/data-grid/column-definition/column-definition.md
```diff
@@ -0,0 +1,148 @@
...
+
+To enable the toolbar you need to add `Toolbar: GridToolbar` to the grid `components` prop.
+
+In addition, the column selector can be shown by using the "Show columns" menu item in the column menu.
+
+The user can choose which columns are visible using the column selector from the toolbar.
+
+To disable the column selector, set the prop `disableColumnSelector={true}`.
+
+{{"demo": "ColumnSelectorGrid.js", "bg": "inline"}}
```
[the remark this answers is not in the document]
flaviendelangle: I have the same idea
Remove it + add a column visibility focused example in `column-visibility`

docs/data/data-grid/cells/cells.md
```diff
@@ -0,0 +1,151 @@
+---
+title: Data Grid - Cells
+---
+
+# Data Grid - Cells
```
m4theushw, line 5: I don't think we need a page for cells. Although it discusses cells, in reality it refers to the cells of a specific column. In this case, I would rename it something like "Column content" or move some sections from here to https://deploy-preview-4600--material-ui-x.netlify.app/x/react-data-grid/column-definition/. To make the page shorter I think we can remove some sections. If we move it back to Column definition I propose the following structure:

- Column definition
  - Providing content
    1. Value getter
    2. Value setter (remove, only used for editing)
    3. Value formatter
    4. Value parser (maybe remove, is used for both filtering and editing, but editing already explains extensively, filtering could duplicate the content)
  - Rendering content
    - Custom cells
    - Making cells accessible (could link to https://github.com/mui/mui-x/pull/4254)
    - Render edit cell (remove)

m4theushw:
> But perhaps we can consider making "Columns" and "Group & Pivot" (and future sections) collapsible. Otherwise as we split more sections, the navBar experience will probably worsen.

I agree, it seems to me that the "Columns" text is too hidden. All of its children are at the same level of the items outside the group. Because of that, we need to add the "Column" prefix to keep them grouped. Consequentially, all pages are formed by two words, while the top level pages are only one. Ideally, I think "Columns" could be displayed like "Rows", then below it we have all its children, left-padded to keep clear they are children. Of course, this needs changes in `AppNavDrawer` so it can't be done here. cc @danilo-leal 

In the core docs makes sense to keep at the same level. As example, the Backdrop has not relation with, e.g. Skeleton.

image

danilo-leal:
@m4theushw you mean doing something like this?

Screen Shot 2022-04-26 at 03 28 10

docs/data/data-grid/cells/cells.md
flaviendelangle: I applied a structure similar to what you proposed

samuelsycamore:
I think that would be a big improvement @danilo-leal !

m4theushw:
@danilo-leal Yes, and I would go even far and remove the "Columns" prefix from the children items. Only leave them in the title of the page.

flaviendelangle:
@m4theushw are you good with the split in it's current form or do we need modifications ?

docs/data/pages.ts
```diff
@@ -15,7 +15,19 @@ const pages = [
       { pathname: '/x/react-data-grid/getting-started' },
       { pathname: '/x/react-data-grid/migration-v4', title: 'Migration from v4' },
       { pathname: '/x/react-data-grid/layout' },
-      { pathname: '/x/react-data-grid/columns' },
+      {
+        pathname: '/x/react-data-grid/columns',
+        children: [
+          { pathname: '/x/react-data-grid/column-definition' },
```
m4theushw, line 21: Did you merge your PR after the monorepo has been updated? It's showing the menu items as "Column Definition", not "Column definition".

docs/data/data-grid/column-definition/column-definition.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+{{"demo": "BasicColumnsGrid.js", "bg": "inline"}}
+
+> ⚠️ The `columns` prop should keep the same reference between two renders.
+> The columns are designed to be definitions, to never change once the component is mounted.
+> Otherwise, you take the risk of losing elements like column width or order.
+> You can create the array outside the render function or memoize it.
+
+## Providing content
```
m4theushw, line 30: Add a paragraph introducing this section? Same for "Rendering content".

docs/data/pages.ts
flaviendelangle: The monorepo is update to date yes

docs/data/data-grid/column-definition/column-definition.md
flaviendelangle: For "Rendering content" I just removed the `### Custom cell` title.
We now have

```
## Rendering content

// explains renderCell

### Styling cells

### Expand cell renderer
```

That way it goes from the most generic to the most advanced use case

docs/data/pages.ts
```diff
@@ -15,7 +15,19 @@ const pages = [
       { pathname: '/x/react-data-grid/getting-started' },
       { pathname: '/x/react-data-grid/migration-v4', title: 'Migration from v4' },
       { pathname: '/x/react-data-grid/layout' },
-      { pathname: '/x/react-data-grid/columns' },
+      {
+        pathname: '/x/react-data-grid/columns',
```
m4theushw, line 19: `scopePathnames` is missing. Otherwise, the sidebar menu is always closed when viewing a child page.

m4theushw (approved):

docs/data/pages.ts
flaviendelangle: Same for `group-pivot`
We did not nest routes in the URL, so our current sidebar menu expansion approach does not work correctly.

AG-Grid does not nest either https://www.ag-grid.com/javascript-data-grid/filter-date/
And I don't think we should.

But in that case we need to update the behavior of the side bar so open when the current page is a children

m4theushw: Why `scopePathnames` was not added here? Although the URLs are not nested, the menu items are. The "Group & Pivot" group has `scopePathnames`.

flaviendelangle: I'm not saying that the parent menu should close, I did not understand the goal of `scopePathnames` correctly

m4theushw: Defining `scopePathnames` is simpler than to build a logic to iterate through all parents and children to find the current page. If we're going to have more groups in the sidebar might make sense to implement the automatic approach.

PR #4615, opened 2022-04-22
[pickers] Add early Error when dateAdapters and format props does not match

### LINKED ISSUE #4598: [pickers] Incorrect typing for `LocalizationProvider::locale` prop
When  using `date-fns` adapter, the constructor for the adapter expects a `Locale` object from `date-fns`, however the type suggests we can pass a string, which leads to runtime errors.

## Current Behavior 😯

Currently, the type for the `locale` prop to `LocalizationProvider` is a union consisting of `Locale` and `string`, but it should only accept `Locale`. As you can see here:

https://github.com/dmtrKovalenko/date-io/blob/b919b995f7697468b096712a46be666492ca71b8/packages/date-fns/src/date-fns-utils.ts#L90-L96

## Expected Behavior 🤔

The prop should not allow strings
## Steps to Reproduce 🕹
Use `string` locale prop

Sandbox: https://codesandbox.io/s/epic-gagarin-3oznz

alexfauquette:
Fix #4598

### DIFF
--- a/packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
+++ b/packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
@@ -42,16 +42,14 @@ export function LocalizationProvider(props: LocalizationProviderProps) {
   const { children, dateAdapter: Utils, dateFormats, dateLibInstance, locale } = props;
 
   if (process.env.NODE_ENV !== 'production') {
-    const lib = new Utils({}).lib
+    const lib = new Utils({}).lib;
     if (lib === 'date-fns' && typeof locale === 'string') {
       throw Error(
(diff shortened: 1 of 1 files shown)


alexfauquette:
Do you mean adding the test [here](https://github.com/dmtrKovalenko/date-io/blob/e669819498f2059f9a8441518d9f6c902f077806/packages/date-fns/src/date-fns-utils.ts#L90-L96) in each adapter of date-io?

```ts
  constructor({
    locale,
    formats,
  }: { formats?: Partial<DateIOFormats>; locale?: Locale } = {}) {
	// Test if local type is appropriate and throw an error if not
    this.locale = locale;
    this.formats = Object.assign({}, defaultFormats, formats);
  }
  ```

flaviendelangle:
Ideally if TypeScript could throw that would be great (through a generic that would type the local key of `IUtils`
I did not succeed during my quick try today, but I think it's doable.

But otherwise, yes.

PR #4617, opened 2022-04-22
[pickers] Rework TDate, TInputDate, TValue and TInputValue generics

flaviendelangle:


ozgurbaserdem:
Hi @flaviendelangle, I've just upgraded form alpha.0 to alpha.7 and discovered that this change breaks our builds in all DatePickers where we've extended the MUI Props (mobile, desktop, static). 

Is there any example docs on how to handle the extending properly after this change? I just realised that this might be more of an TypeScript question, hmm.. 

Anyway this is where it breaks for us, which I believe should be related to this PR:
Screenshot 2022-06-27 at 21 49 17

flaviendelangle:
Hi,

Could you link me a code snippt where to typescript error is in your code ?

ozgurbaserdem:
> Hi,
> 
> Could you link me a code snippt where to typescript error is in your code ?

Does this work for you?
https://codesandbox.io/s/elastic-mcclintock-oxfm5t

flaviendelangle:
Your `index.ts` file is empty right ?

ozgurbaserdem:
> Your `index.ts` file is empty right ?

My apologies, does this work better: https://codesandbox.io/s/actionbarcomponentprops-demo-mui-x-forked-5kn9ck?file=/demo.tsx

Ignore the link and comments (I basically just changed the demo.tsx but dependencies etc are copies from one of the MUI doc sandboxes). The TypeError when extending still persists though.

flaviendelangle:
`    "@mui/x-date-pickers": "https://pkg.csb.dev/mui/mui-x/commit/d4c0daac/@mui/x-date-pickers",`

This is an install from a specific commit
Do you know which commit it is ?
You should try with the latest version, this comment might be very outdated.

flaviendelangle:
With that being said

On this PR I added generics to most of the components
So if you extend the props you have to pass them.

If you know what the type is (for example for Date FNS is `Date` which is very convinient) you just pass it to our interface
https://codesandbox.io/s/actionbarcomponentprops-demo-mui-x-forked-57hd3o?file=/demo.tsx:0-1333

PR #4626, opened 2022-04-23
[docs] Quick sidenav fix

oliviertassinari:
Handle https://github.com/mui/mui-x/pull/4591#issuecomment-1107535769. It's a dirty fix, but it works 😁. From my perspective, the notion of `scopePathnames` should not exist in the first place.

https://deploy-preview-4626--material-ui-x.netlify.app/x/react-data-grid/tree-data/

### DIFF
--- a/docs/data/pages.ts
+++ b/docs/data/pages.ts
@@ -34,6 +34,13 @@ const pages = [
       {
         pathname: '/x/react-data-grid/row-grouping',
         title: 'Group & Pivot',
+        scopePathnames: [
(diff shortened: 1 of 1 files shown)


flaviendelangle (approved):

oliviertassinari:
cc @mui/docs-infra I think that in the future, we could consider changing the tree view expand logic. We could have it so it find the leaf, and then expand all the parents. Right now, it uses URL pathname prefixes.

PR #4645, opened 2022-04-25
[test] `disabled` prop disable calendar days

### LINKED ISSUE #4400: Cannot disable ClockPicker, StaticTimePicker, StaticDateTimePicker

### LINKED ISSUE #4594: [DatePicker] StaticDatePicker doesn't honor disabled

### LINKED ISSUE #4650: [static pickers] Handle `disable` and `readOnly` on all static pickers
This is a follow up of #4594

What is tested with both `readOnly` and `disabled` on [pickers](https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx#L88:L107) is that clicking on the calendar does not open the views.

For now, only calendars apply the `readOnly` with `if (readOnly) { return; }` at the beginning of `handleDaySelect`, `onMonthSelect` and `handleYearSelection`.

You can experiment with them in the following demo

https://codesandbox.io/s/material-ui-issue-forked-0v1z9d?file=/src/demo.js

Noticed inconsistencies

- `StaticTimePicker` are still editable with `disabled` or `readOnly`
- `StaticDateRangePicker` 
	- with `disabled` allows to see next month, but not the previous one
	- start/end dates in the headers are focusable but do nothing

alexfauquette:
Fix #4594 by testing the corrected bug
Fix #4400
Fix #4650

In `readOnly` mode
- edit buttons appear normal, but do nothing
- navigation is enabled between months and views

In `disabled` mode
- edit buttons are disabled
- navigation is enabled between views but not months
- Special case of range picker month navigation between the first and last month of the selected range


packages/x-date-pickers/src/StaticDatePicker/StaticDatePicker.test.tsx
```diff
@@ -67,4 +67,22 @@ describe('<StaticDatePicker />', () => {
     expect(getYearButton(2030)).to.have.attribute('disabled');
     expect(getYearButton(2031)).not.to.have.attribute('disabled');
   });
+
+  it('prop `disabled` – disables all days', () => {
```
flaviendelangle, line 71: In other tests, they do the same check for `readOnly`, I did not look into those for now but maybe it also make sense here.

flaviendelangle, line 71: In other tests, they do the same check for `readOnly`, I did not look into those for now but maybe it also make sense here.

packages/x-date-pickers/src/CalendarPicker/PickersCalendarHeader.tsx
```diff
@@ -148,11 +148,11 @@ export function PickersCalendarHeader<TDate>(props: PickersCalendarHeaderProps<T
   const selectPreviousMonth = () => onMonthChange(utils.getPreviousMonth(month), 'right');
 
   const isNextMonthDisabled = useNextMonthDisabled(month, {
-    disableFuture: disableFuture || disabled,
```
alexfauquette, line 151: `disableFuture` is disabling compare to the current day, not the selected one, so it does not make sense, especially for the range picker. I moved the responsibility to adapt `minDate`/`maxDate` when picker is disabled to the parent

flaviendelangle (commented):
I am not very familiar with the `readOnly` pattern
Should the hover background be disabled on `readOnly` when hovering a day in the `DayPicker` ?

alexfauquette:
I never used `readOnly` in an application. I looked at other components, and it seems the only difference is that you can not modify the content.

I don't thin `readOnly` impacts the visual feedback. For example, the `Select` has the same visual feedback on hover, and when you click on it, it looks like a focused field but does not open the menu

https://mui.com/material-ui/react-select/#other-props

flaviendelangle:
OK, I don't get the goal of this prop but if it is the same elsewhere

flaviendelangle (approved):

flaviendelangle:
@alexfauquette the tests are failing on Safari, we probably need to early return (see other similar tests)

```
Safari 13.1.2 (Mac OS 10.15.6) <ClockPicker /> should display options, but not update value when readOnly prop is passed FAILED
	undefined is not a constructor (evaluating 'new window.Touch(Object.assign({
	      target,
	      identifier: 0
	    }, opts))')
	/tmp/_karma_webpack_806820/commons.js:91233:64
	map@[native code]
	fireTouchChangedEvent@/tmp/_karma_webpack_806820/commons.js:91233:39
	/tmp/_karma_webpack_806820/commons.js:175764:38
Safari 13.1.2 (Mac OS 10.15.6) <ClockPicker /> should display disabled options when disabled prop is passed FAILED
	undefined is not a constructor (evaluating 'new window.Touch(Object.assign({
	      target,
	      identifier: 0
	    }, opts))')
	/tmp/_karma_webpack_806820/commons.js:91233:64
	map@[native code]
	fireTouchChangedEvent@/tmp/_karma_webpack_806820/commons.js:91233:39
	/tmp/_karma_webpack_806820/commons.js:175787:38
```

PR #4649, opened 2022-04-25
[pickers] Do not apply the current time when no date provided in `DayPicker`

### LINKED ISSUE #4449: [pickers] The minutes and seconds default value is taken from current date instead of being 00
### Duplicates

### Latest version

### Current behavior 😯

If you go to https://mui.com/components/date-time-picker/#api and try selecting a day on the "Clear Initial State" input you will notice that the minutes are initially set to the minutes of the current hour. The seconds of the Date value will also be set to the current seconds of the minute (althought that is not visible unless you inspect the value object).

This is almost never desired and breaks naive integrations with libraries such as React Final Form or React Hook Form.  

Also, if you have set `minutesStep={5}` it will not be respected.

A similar issue has already been addressed on the TimePicker in mui/material-ui#24887 but was not fixed for DateTimePicker (and maybe even DatePicker and DateRangePicker?).

flaviendelangle:
Fixes #4449 

- [x] If we use the `DatePicker` or `DateTimePicker`, when clicking on a date for the 1st time, it should set the time to `00:00:00`not the current time

- [x] Add tests

### DIFF
--- a/packages/x-date-pickers/src/CalendarPicker/DayPicker.test.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/DayPicker.test.tsx
@@ -1,0 +1,56 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { spy } from 'sinon';
+import { userEvent, screen } from '@mui/monorepo/test/utils';
+import { DayPicker } from './DayPicker';
+import { adapterToUse, createPickerRenderer } from '../../../../test/utils/pickers-utils';
+
+describe('<DayPicker />', () => {
+  const { render } = createPickerRenderer({ clock: 'fake' });
+
+  it('should set time to be midnight when selecting a date without a previous date', () => {
+    const onChange = spy();
+
+    render(
+      <DayPicker
+        date={null}
+        focusedDay={null}
+        onFocusedDayChange={() => {}}
+        onMonthSwitchingAnimationEnd={() => {}}
+        isDateDisabled={() => false}
+        isMonthSwitchingAnimating={false}
+        slideDirection="right"
+        reduceAnimations={false}
+        currentMonth={adapterToUse.date('2018-01-01T00:00:00.000')}
+        onChange={onChange}
+      />,
+    );
+
+    userEvent.mousePress(screen.getByLabelText('Jan 2, 2018'));
+    expect(onChange.callCount).to.equal(1);
+    expect(onChange.lastCall.args[0]).toEqualDateTime(adapterToUse.date('2018-01-02T00:00:00.000'));
+  });
+
+  it('should keep the time of the currently provided date', () => {
+    const onChange = spy();
+
+    render(
+      <DayPicker
+        date={adapterToUse.date('2018-01-03T11:11:11.111')}
+        focusedDay={null}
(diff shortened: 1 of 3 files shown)


packages/x-date-pickers/src/CalendarPicker/DayPicker.test.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { expect } from 'chai';
+import { spy } from 'sinon';
+import { userEvent, screen } from '@mui/monorepo/test/utils';
+import { DayPicker } from './DayPicker';
+import { adapterToUse, createPickerRenderer } from '../../../../test/utils/pickers-utils';
+
+describe('<DayPicker />', () => {
```
flaviendelangle, line 8: I added the test directly on `DayPicker`, I think the structure of the pickers favors more unit tests on the smaller components.

packages/x-date-pickers/src/CalendarPicker/DayPicker.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    render(
+      <DayPicker
+        date={null}
+        focusedDay={null}
+        onFocusedDayChange={() => {}}
+        onMonthSwitchingAnimationEnd={() => {}}
+        isDateDisabled={() => false}
+        isMonthSwitchingAnimating={false}
+        slideDirection="right"
+        reduceAnimations={false}
```
alexfauquette, line 23: Why do you define all those props? Is it for performance improvement?

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -139,12 +139,20 @@ export function DayPicker<TDate>(props: PickersCalendarProps<TDate>) {
       if (readOnly) {
         return;
       }
+
       // TODO possibly buggy line figure out and add tests
```
alexfauquette, line 143: Now that it's tested, the comment could be removed, or at least remove 'line' otherwise in a few months we will wonder why `let finalDate` can be buggy

```suggestion
      // TODO possibly buggy figure out and add tests
```

packages/x-date-pickers/src/CalendarPicker/DayPicker.test.tsx
flaviendelangle: No, it's because `DayPicker` being an internal component, it was built with almost all its props required and I did not want to change its api at all.

But some default value could probably be inside `DayPicker` instead of higher in the component chain.

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
flaviendelangle: I found the issue related to this comment ealier today https://github.com/mui/mui-x/issues/4703

I think to problem is that some date library represents dates as array so this check is problematic and we should pass an  `isRange` boolean instead to differentiate ranges and mono dates.

alexfauquette: Good to know 👍

PR #4651, opened 2022-04-25
[license] Allow to limit some packages to a specific license plan

### LINKED ISSUE #3925: [license] Update license key manager to accommodate Premium plan.
# Motivation 

As part of #3135, and to enable users to buy and use the new premium plan, we'll need to update our license key manager to generate and validate the new keys.

flaviendelangle:
- [x] Handle `KEYVERSION=2` in `verifyLicense`
- [x] Handle both `perpetual` and `subscription` sales models
- [x] Do not accept pro licenses for premium features 
- [x] Change the doc demonstration license to be a v2 license with the premium scope*
- [x] Check `salesModel` and `scope` value validity both on license encoding and decoding to avoid typos

Close #3925


packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
```diff
@@ -27,7 +28,11 @@ export function useLicenseVerifier(
       return sharedLicenseStatuses[packageName]!.status;
     }
 
-    const licenseStatus = verifyLicense(releaseInfo, licenseKey);
+    const acceptedScopes: LicenseScope[] = packageName.includes('premium')
```
flaviendelangle, line 31: This logic would have to become smarter if we release other plans

packages/x-license-pro/src/verifyLicense/verifyLicense.test.ts
```diff
@@ -8,36 +8,98 @@ const oneYear = oneDayInMS * 365;
 const RELEASE_INFO = generateReleaseInfo();
 
 describe('License: verifyLicense', () => {
-  const validLicense = generateLicense({
-    expiryDate: new Date(new Date().getTime() + oneYear),
-    orderNumber: 'MUI-123',
-  });
+  describe('key version: 1', () => {
+    const license =
+      '0f94d8b65161817ca5d7f7af8ac2f042T1JERVI6TVVJLVN0b3J5Ym9vayxFWFBJUlk9MTY1NDg1ODc1MzU1MCxLRVlWRVJTSU9OPTE=';
```
flaviendelangle, line 13: We don't have the v1 generation anymore but it's worth testing it
This key is the doc key

packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
DanailH: So to be sure that I understand it - the goal here is to if you have `premium` you also get `pro` right?

packages/x-license-pro/src/utils/licenseScope.ts
```diff
@@ -0,0 +1 @@
+export type LicenseScope = 'pro' | 'premium';
```
DanailH, line 1: To be honest I was thinking to rework this to an `enum` because we are also using the strings themselves and it would be nice to avoid typos. It will also help if we have more licenses in the future. What do you think?

packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
flaviendelangle: Yes
I first did a scale `const scopes = ['community', 'pro', 'premium']` and then compare the index of the current plan with the one of the license
But it would not work with more granular scopes if we have some in the future
Right now it's a very basic check so both options work fine.

mbrookes:
We need to be able to distinguish between Pro perpetual and Pro subscription, and display the appropriate console message/watermark when the license is expiring soon/expired. (Existing perpetual customers will be able to renew their perpetual license.)

flaviendelangle:
> We need to be able to distinguish between Pro perpetual and Pro subscription, and display the appropriate console message/watermark when the license is expiring soon/expired. (Existing perpetual customers will be able to renew their perpetual license.)

@DanailH that's information we currently don't have I think

DanailH:
Nope, we don't have that info in the license. When I was updating it I didn't know what this was a requirement.

flaviendelangle:
Me neither @mbrookes, here is the current v2 format:

```
ORDER=${orderNumber},EXPIRY=${expiryTimestamp},KEYVERSION=2,SCOPE=${scope}`
```

We would need to add a `TIMEOUT_STRATEGY: 'perpetual' | 'annual'` (if you have a shorter key name that would be great @DanailH)
If there are other information we need to add ?

Do we switch the `TIMEOUT_STRATEGY` to `annual` for all new licenses after the next release ?

joserodolfofreitas:
> Do we switch the TIMEOUT_STRATEGY to annual for all new licenses after the next release ?
No, I thought that'd be the case, and then we could simply use the expiryDate. But renewals from perpetual plans will remain perpetual as well. So we'll need an extra attribute for those cases.

flaviendelangle:
> But renewals from perpetual plans will remain perpetual as well. So we'll need an extra attribute for those cases.

Yes, otherwise we could have said "if license creation date > XXX then if's perpetual".

DanailH:
How about `DURATION='perpetual' | 'annual'` instead of `TIMEOUT_STRATEGY`?

mbrookes:
> How about DURATION='perpetual' | 'annual' instead of TIMEOUT_STRATEGY?

`PLAN = 'perpetual' | 'subscription'`

What's `SCOPE=${scope}` ?

flaviendelangle:
> PLAN = 'perpetual' | 'subscription'

We use the term "plan" to differentiate "pro plan" and "premium plan" so I would not use it to describe the duration during which the license is valid.

> What's SCOPE=${scope} ?

`SCOPE` currently contains the plan
I think @DanailH and @oliviertassinari went for scope instead of plan in the license to be more abstract and be able to add scopes that are not plans (maybe an additional license for one specific very high value data grid plugin).

mbrookes:
~~`PLAN`~~ `TERM` ?

mbrookes:
The alternative is that we continue to use the v1 key format for perpetual licenses.

DanailH:
I think it would be better to just use the new version. I guess even if we use v1 for perpetual licenses we would still have the problem of not knowing the SCOPE unless the Premium plan can't be perpetual (initially it was PLAN but as we can have addons in the future we thought that SCOPE is more appropriate).

flaviendelangle:
@DanailH I updated it with `TERM`

flaviendelangle:
By the way, the encoded license if becoming long, should we use initial for keys ?

With full keys: `5abdc6ddfa0eec6e04e57360c6bc5997T1JERVI9TVVJLVN0b3J5Ym9vayxFWFBJUlk9MTY4Mjg0NDU0MjQwNixLRVlWRVJTSU9OPTIsU0NPUEU9cHJlbWl1bSxURVJNPXN1YnNjcmlwdGlvbg==`

With initial keys: `1076e2f8a8dea8d320e2298048600169Tz1NVUktU3Rvcnlib29rLEU9MTY4Mjg0NDY3OTY5NCxLVj0yLFM9cHJlbWl1bSxUPXN1YnNjcmlwdGlvbg==`

mbrookes:
@flaviendelangle 
> We dropped the algo so we would need to re-add it

We have numerous Pro customers who have license keys in v1 format, so we we need to continue to support it regardless. Even if they don't renew, the license is by definition perpetual, so it will need to be supported for the foreseeable future so that they get a meaningful message if they try to upgrade in a year's time with a then out-of-date v1 license key.

@DanailH 
>  I guess even if we use v1 for perpetual licenses we would still have the problem of not knowing the SCOPE unless the Premium plan can't be perpetual.

- Premium will be Subscription only*.
- Pro can be perpetual (customers who purchased before the change), or Subscription*.

*As far as right to develop and use new versions goes. The right to use in production will perpetual.

mbrookes:
> should we use initial for key

Yes. No reason to have the full key in the license string.

mbrookes:
> I updated it with TERM

A quick web search suggests that "model" is the accepted term. Sorry!

flaviendelangle:
> We have numerous Pro customers who have license keys in v1 format, so we we need to continue to support it regardless

We still have the algo to decode it, not the algo to encode it
In which scenario would we need to create a new v1 license (except for test in which case we can checkout a git tag) ?
The future perpetual license will be in the v2 format

flaviendelangle:
> A quick web search suggests that "model" is the accepted term

`model` is not very explicit to me, is this something most dev would understand just by ready the key ?
If we do use the initial for keys, I would use very explicit keys like `VALIDITY_DURATION` (`VD`)

DanailH:
> I updated it with `TERM`

I'm ok with this. One thing - shouldn't the default be `perpetual` as that is the default for `pro` as far as I understood it? That way it won't be breaking for already existing licenses (it's the same thing that we did for the `scope`).

flaviendelangle:
> That way it won't be breaking for already existing licenses

Why would the other default break the existing license ?
Are you talking about the default in encoding or decoding ?

For me we are updating the store to pass this value (so that the store can decide to pass "perpetual" when the user is expanding a perpetual license).
So the default value in encoding, like for "plan" is temporary and just impacts the 1st licenses created before we update the store.
For those first licenses, should we set them to perpetual or to annual ?
If we do "annual", then we must not do any renewal before updating the store
If we do "perpetual", then we should update the store quickly because our terms saying that our licenses are annual will be wrong.

In both cases, I think that we should already have an open PR on the store to update it just after the release. Otherwise, we will stay in this migration phase for to long.

DanailH:
Ok, breaking is probably not the correct term but rather if we do it like this then you should also update the Pro product on the store. With the other approach, one only needs to handle the Premium product in the store.

flaviendelangle:
The new pro licenses will be perpetual ? I thought they were going to be annual just like the premium licenses.

joserodolfofreitas:
What about?
`PERPETUITY = 'production' | 'development'`
After all, licenses will still be perpetual, but in production only. 
And we'll need to "advertise" that (Not to say that this is an argument per se).

We may also consider, in the future, removing the console warning from test environments (and leaving the watermark). We have complaints on [zendesk](https://mui.zendesk.com/agent/tickets/2924) about the annoying volume of log messages related to the license.
In this case, we could later expand the implementation with:
`PERPETUITY = 'production' | 'development' | 'test'`

joserodolfofreitas:
> The new pro licenses will be perpetual ? I thought they were going to be annual just like the premium licenses.

New Pro licenses generated for a previous customer will still be perpetual in development. But for new customers, it will be perpetual in production only.

flaviendelangle:
> New Pro licenses generated for a previous customer will still be perpetual in development. But for new customers, it will be perpetual in production only.

Can we call the "annual in dev but perpetual in prod" just "annual" for simplicity ?

flaviendelangle:
> PERPETUITY = 'production' | 'development' 
> PERPETUITY = 'production' | 'development' | 'test'


Not a huge fan tbh
If we set `PERPETUITY=test", it's not obvious if it means "test" + "development" or just "test".
I would prefer to keep the abstraction and let the code choose on which env to do the check rather than trying to describe the env on the license itself.

And if we decide to remove the warning on test, I guess it should be retroactive, not just for the new licenses. Otherwise we would basically be saying to the user complaining "ok you are right, you will get the fix in one year".

mbrookes:
>> PERPETUITY = 'production' | 'development'
>> PERPETUITY = 'production' | 'development' | 'test'`

> Not a huge fan tbh

Me neither, sorry, just for the name alone.

> Can we call the "annual in dev but perpetual in prod" just "annual" for simplicity ?

That would be "subscription in dev, perpetual in prod", and its proposed to just call it "subscription", so sounds like that's agreed.

joserodolfofreitas:
> > > PERPETUITY = 'production' | 'development'
> > > PERPETUITY = 'production' | 'development' | 'test'`
> 
> > Not a huge fan tbh
> 
> Me neither, sorry, just for the name alone.

Ok, np. But I think "term" is not really exact either. I think it's going to be confused with the "License Term" [we'll use to define how many years the license will be valid](https://codesandbox.io/s/dev-input-hfhqex?file=/demo.js). 

[License Term](https://www.lawinsider.com/dictionary/license-term) means the period of time during which Licensee is authorized to use Program(s) in accordance with the applicable license grant.

flaviendelangle:
`VALIDITY_DURATION` ? `VD` in the license

joserodolfofreitas:
Validity and duration kinda mix two concepts, but I like the idea of using very explicit keys.

But I'm pondering, and `model` seems indeed the best name that we could use for this situation.
If you think model is not clear.

What about:
SALES_MODEL: 'perpetual' | 'subscription'
TERM: number (1-3) ?

flaviendelangle:
`SALES_MODEL` is fine for me

DanailH (approved):

packages/x-license-pro/src/generateLicense/generateLicense.ts
```diff
@@ -1,30 +1,35 @@
...
-  },EXPIRY=${details.expiryDate.getTime()},KEYVERSION=${licenseVersion},SCOPE=${details.scope}`;
+  if (details.scope && !LICENSE_SCOPES.includes(details.scope)) {
+    throw new Error('MUI: Invalid scope');
+  }
+
+  if (details.licensingModel && !LICENSING_MODELS.includes(details.licensingModel)) {
+    throw new Error('MUI: Invalid sales model');
+  }
+
+  return `O=${details.orderNumber},E=${details.expiryDate.getTime()},S=${
```
oliviertassinari, line 26: Without this change, a v2 license key is not usable by < v5.11.0 of the software. Why not keep?

```suggestion
  return `O=${details.orderNumber},EXPIRY=${details.expiryDate.getTime()},S=${
```

it was surfaced in https://groups.google.com/a/mui.com/g/x/c/Nkq2fQbGkn0/. Time will tell if this causes trouble with more users, I suspect it will. If the intent is to shorten the string key, I doubt it's worth the breaking change. If a breaking change needs to happen,  then I think bundling with #4892 could be great.

flaviendelangle: Why revert instead of just supporting both in the decoding script?

oliviertassinari: @joserodolfofreitas we are one week in, how are the upgrades support cases going? I haven't seen new ones, so maybe it's fine.

PR #4663, opened 2022-04-26
[core] Run typescript only on specific package

alexfauquette:
Using typescript takes time which could be reduced by testing only the component library I'm working on

```bash
yarn typescript:grid // 13s
yarn typescript:pickers // 26s
yarn typescript // 58s
```


flaviendelangle (commented):
Maybe we could have `yarn typescript grid`
That way we could list other patterns without flooding the `package.json` files
Thinking about `yarn typescript docs` for instance

PR #4665, opened 2022-04-26
[DateRangePicker] Remove `orientation` prop

### LINKED ISSUE #4656: StaticDateRangePicker orientation prop has no effect
### Duplicates

### Latest version

### Current behavior 😯

By default the StaticDateRangePicker is displayed with 2 calendars side by side (which is landscape orientation).

grafik

If you change the **orientation prop to portrait** the two calendars are still side by side.

grafik

### Expected behavior 🤔

If you change the orientation prop to portrait the calendars should be displayed **one below the other** or the prop should be removed from the component and API reference.

grafik

### Steps to reproduce 🕹

[https://codesandbox.io/s/staticdaterangepickerdemo-material-demo-forked-uor7jq?file=%2Fdemo.js](https://codesandbox.io/s/staticdaterangepickerdemo-material-demo-forked-uor7jq?file=%2Fdemo.js)

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Closes #4656

The purpose of `orientation` in the date range picker is unclear, so I chose to remove it for now since it already does nothing.

In the date picker and time picker, `orientation="landscape"` shows both the calendar and toolbar side to side.

image

Since in the date range multiple calendars are rendered, the user in #4656 was expecting that `orientation` would change the orientation of these calendars, but it doesn't. If we want to make the prop work like in the date picker, then changing the orientation would make the toolbar to render side to side with the calendars, however, this might not be what users expect. `orientation="landscape"` would have this effect on the date range picker:

image

The expected behavior of #4656 can be achieved with CSS if we add slot names to the internal components.


flaviendelangle (approved):
I'm ok with removing it from the range components since it was probably just added by mistake when creating them.

Issue #4669, opened 2021-08-10
[pickers] Add style override for sub components


AsemK:
- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

After migrating pickers to emotion (v4 to v5), I lost the ability to modify the styling of some subcomponents. I think this is due to the fact that `styled` is used without providing the `name` option as can be seen in

https://github.com/mui/mui-x/blob/ebb0e3b1a4d940b02ca20abd61ade86fce9b6e94/packages/x-date-pickers/src/DateTimePicker/DateTimePickerToolbar.tsx#L31

## Expected Behavior 🤔
I should be able to style the subcomponents of pickers. For instance, in RTL view, the `TimeContainer` component of `DateTimePickerToolbar` used in `MobileDateTimePicker` shows the time with `direction: rtl`. I want to change this behavior so that the time is always `ltr` even if the theme is `rtl`. Previously, with JSS, this worked:
```javascript
const theme = createTheme({
  direction: rtl,
  components: {
      MuiDateTimePickerToolbar: {
        styleOverrides: {
          timeContainer: {
            flip: false,
            direction: "ltr",
          },
        },
      },
    },
});
```
Now, with emotion, I would expect the following to work, but it doesn't:
```javascript
const theme = createTheme({
  direction: rtl,
  components: {
      MuiDateTimePickerToolbar: {
        styleOverrides: {
          root: {
            "& .MuiDateTimePickerToolbar-timeContainer": {
              direction: "/* @noflip */ ltr",
            },
          },
        },
      },
    },
});
```

Also, when inspecting the element, the class name is not `MuiDateTimePickerToolbar-timeContainer` or anything similar, but `muirtl-1lomthf`.

Is there currently a work around for this?

m4theushw:
During the migration to emotion we missed to properly convert some internal components. We decided to leave this task for the community. There're a few components where the slot name and overrides resolver is missing. Feel free to open a pull request converting any of the components from the list below. You can use #4672 as example.

Checklist:

- [ ] Set the name, slot and `overridesResolver` in `styled`
- [ ] Generate `classes` using `generateUtilityClasses`
- [ ] Pass the appropriate `classes.XXX` to the `className` prop of the internal components (use `clsx` to merge with user-provided `className` prop, if available)

Components whose slot name is missing:

[...]

PR #4672, opened 2022-04-27
[pickers] Add overrides to `PickersArrowSwitcher`

m4theushw:
Preview: https://deploy-preview-4672--material-ui-x.netlify.app/x/react-date-pickers/date-picker/

An example for #4669 

My goal was to only restore the capability to provide overrides via the theme and add meaningful class names. I didn't add the `classes` prop. 

Some components have "private" class names. Do you agree to make them public? In the DataGrid every new class is public.

https://github.com/mui/mui-x/blob/0d8a55d4f460767942c01398a373028ef2e1e7da/packages/x-date-pickers/src/CalendarPicker/PickersSlideTransition.tsx#L18

Before:

image

After:

image

### DIFF
--- a/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher.tsx
+++ b/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
+import clsx from 'clsx';
 import Typography from '@mui/material/Typography';
+import { generateUtilityClasses } from '@mui/material';
 import { useTheme, styled } from '@mui/material/styles';
 import IconButton from '@mui/material/IconButton';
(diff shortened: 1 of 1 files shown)


oliviertassinari (commented):
The changes seem to match with what we do on the rest of the codebase, especially in Material UI. 

> Some components have "private" class names. Do you agree to make them public? In the DataGrid every new class is public.

Usually, developers ask to be able to customize everything. I think that the only cases when a class name should stay private is if we don't envison a way to use it for customization, or if either is a better way to achieve the same customization.

m4theushw:
I would say to keep only classes of the `XXXTransition` components as private, but even these the users might want to customize. For instance, if I want to override the animation duration or easing, then only with CSS is possible. It seems that the path is to make every class public.

Issue #4516, opened 2022-04-15
[data grid] Set cell focus on closest row after deleted focused row (maybe optional)


fourteenmeister:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Steps to reproduce

https://mui.com/x/react-data-grid/rows/#updating-rows

1. set focus on last row
2. delete a row

expected behavior:
focus on previous line if removed last row or focus on removed index

m4theushw:
Currently, we don't update the focus when a row is removed. Do you have this same problem in the application you're developing? I want to understand if we should only fix the demo or sync the focus when deleting the focused cell.

fourteenmeister:
> Do you have this same problem in the application you're developing?

Yes, in the my application same behavior.
> want to understand if we should only fix the demo or sync the focus when deleting the focused cell.

Need sync the focus when deleting the focused cell.

m4theushw:
We have 3 options to close this issue:

1. Update the demo to focus the cell above when deleting a row
   It's the simplest option.

2. Make the focus be automatically updated when `updateRows` is called
   Also simple to implement, but it won't work if the row is removed from the `rows` prop.
   Only need to check if one of the deleted rows is also focused.
   `apiRef.current.unstable_moveFocusToRelativeCell` can be used to move focus to the cell above

3. Listen to `GridEvents.rowsSet`
   Will work either if the row is removed from the `rows` prop or `updateRows` is called
   We need to iterate through all rows checking if it was removed or not.
   More complex to implement but works in all scenarios
   To be performant, one option is to reuse the for loop used to fill `idRowsLookup`: https://github.com/mui/mui-x/blob/92e2a9fdeb7574d88d87b3eb712e1001c8820a7a/packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts#L61

michelengelen:
Thanks for the report and the discussion. This shipped with #15061. When the focused row is deleted, the grid now moves the focus to the row at the same index, or to the previous row if the last row was removed. It listens to the `rowsSet` event, so it works with `apiRef.current.updateRows` and with `rows` prop updates.

[...]

PR #4683, opened 2022-04-28
[data grid] Check focus validity whenever the rows in state changes

### LINKED ISSUE #4678: [data grid] Focus is not reset when the rows changes due to a change in row grouping model
### Duplicates

### Latest version

### Current behavior 😯

When we configure a grouping model and expand the generated row and then delete this grouping model, clicking on the column header results in an error:

`Uncaught Error: No row with id #auto-generated-row-company/20th Century Fox-director/James Cameron found`

In the local environment it also throws warnings like this:

`MUI: You are calling getValue("__row_group_by_columns_group__") but the column "__row_group_by_columns_group__" is not defined.
Instead, you can access the data from "params.row.__row_group_by_columns_group__".`

### Expected behavior 🤔

We should be able to freely change the grouping model and use the table without errors

### Steps to reproduce 🕹

I will not publish my example because the problem appears also in the documentation.

Steps:

1. Go to the example in docs: [Row Grouping example](https://mui.com/x/react-data-grid/row-grouping/#RowGroupingCustomGroupingColDefCallback) 
2. Select `Group by company and director` tab
3. Expand auto-generated row `by James Cameron`
4. Select `Group by company` tab
5. Click on any column header
6. Then `Uncaught Error: No row with id #auto-generated-row...` error appear.

### Context 🔦

I have a custom row grouping component where I drag and drop the column header name to input which generates a new grouping model, so when I'm changing the grouping model and expanding an auto-generated row and then trying to set a new model then an error appears. 

### Your environment 🌎

System:
    OS: Windows 10 10.0.19044
  Binaries:
    Node: 16.13.2 
    Yarn: 1.22.17 
    npm: 8.1.2 
  Browsers:
    Chrome
  npmPackages:
    @emotion/react: ^11.8.2 => 11.9.0
    @mui/base:  5.0.0-alpha.75
    @mui/icons-material: ^5.5.1 => 5.6.0
    @mui/lab: ^5.0.0-alpha.75 => 5.0.0-alpha.76
    @mui/material: ^5.5.2 => 5.6.0
    @mui/private-theming:  5.6.0
    @mui/styled-engine:  5.6.0
    @mui/system:  5.6.0
    @mui/types:  7.1.3
    @mui/utils:  5.6.0
    @mui/x-data-grid: ^5.8.0 => 5.8.0
    @mui/x-data-grid-generator: ^5.8.0 => 5.8.0
    @mui/x-data-grid-pro:  5.8.0
    @mui/x-license-pro:  5.7.0
    @types/react:  17.0.43
    react: 17.0.2 => 17.0.2
    react-dom: 17.0.2 => 17.0.2
    typescript:  4.6.3

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #4678

Before this PR, we where listening to `props.rows` changes to check the focus validity.
Now I listen to `GridEvents.rowsChange` to also track changes with `apiRef.current.updateRows` or grouping changes.

I don't think we need to listen to `sortedRowChange` and pagination change. If the focus is on a non-visible row, it does not causes any crash.

____

I also removed unused unused `props.rows` from `useGridDimensions` typing
That way only `useGridRows` accesses directly `props.rows`, all other hooks should use the state.


packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
```diff
@@ -236,6 +236,18 @@ export const useGridFocus = (
     [apiRef],
   );
 
+  const handleCleanFocus = React.useCallback<GridEventListener<'rowsSet'>>(() => {
```
m4theushw, line 239:
```suggestion
  const handleRowsSet = React.useCallback<GridEventListener<'rowsSet'>>(() => {
```

packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
```diff
@@ -236,6 +236,18 @@ export const useGridFocus = (
...
+    const cell = gridFocusCellSelector(apiRef);
+
+    // If the focused cell is in a row which does not exist anymore, then remove the focus
+    if (cell && !apiRef.current.getRow(cell.id)) {
+      apiRef.current.setState((state) => ({
+        ...state,
+        focus: { cell: null, columnHeader: null },
```
m4theushw, line 246: In theory, every time a cell loses focus it should publish `cellFocusOut`, but to publish it we call `getCellParams` which will crash because the row doesn't exist. It will be a bug if the removed row was also being edited. If we want to implement the option 3 from https://github.com/mui/mui-x/issues/4516#issuecomment-1102916209 we will likely have to revisit this part. Maybe we change the params passed to `cellFocusOut` to not include the row object. Anyway, until someone complains about `cellFocusOut` not published we can keep the listener to `rowsSet`.

m4theushw (approved):

packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
flaviendelangle: Yes, in v6 we could pass the id instead of the params to allow to publish the focus out event on a destroyed cell.

Issue #4684, opened 2022-04-28
[data grid] Remove `GridEvents` enum


flaviendelangle:
I v6, I propose we drop the `GridEvents` enum in favor of a simple type generated from the keys of `GridEventLookup` (the actual `GridEventsStr`.

That way, we will not have to define the pro-only and premium-only events inside the `gridEvents.ts` file which is located in the community package.
Just like for pre-processor, there is no reference to pro-only processors inside the community package because they are based on the lookup which is augmented in the pro and premium packages.

```diff
-apiRef.current.subscribeEvent(GridEvents.rowGroupingModelChange, handleRowGroupingModelChange)
+apiRef.current.subscribeEvent('rowGroupingModelChange', handleRowGroupingModelChange)

-apiRef.current.publishEvent(GridEvents.rowExpansionChange)
+apiRef.current.publishEvent('rowExpansionChange')

-useGridApiEventHandler(GridEvents.rowGroupingModelChange, handleRowGroupingModelChange)
+useGridApiEventHandler('rowGroupingModelChange', handleRowGroupingModelChange)
```

## Goals

- Do not leak pro-only and premium-only code on the community package
- Needed to allow user to create there own plugins with there own events 

## Work to do

- [x] V5: Stop using `GridEvents` in our codebase and docs (replace with the string which is already supported)
- [x] V5: Mark `GridEvents` as deprecated (I don't see how we could do a runtime warning)
- [x] V5 : Move the comments from `GridEvents` to `GridEventLookup` and update `buildGridEventsDocumentation`
- [x] V6: Remove `GridEvents`

PR #4685, opened 2022-04-28
[data grid] Stop using GridEvents enum in `@mui/x-data-grid-premium`

flaviendelangle:
Part of #4684

This PR demonstrated what I want to do in #4684 but only on `@mui/x-data-grid-premium` to give a small example.
If you are ok with it, I can apply it to all our codebase.


cherniavskii (approved):
Totally agree 👍

DanailH (approved):
I'm ok with this as well.

PR #4688, opened 2022-04-28
[TimePicker] Refacto `isTimeDisabled` method

flaviendelangle:
The old value was way to complicated to read for me

### DIFF
--- a/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
+++ b/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
@@ -268,60 +268,50 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
 
       const isAfter = createIsAfterIgnoreDatePart(disableIgnoringDatePartForTimeValidation, utils);
 
-      const shouldDisableDatePeriod = ({
-        start,
-        end,
-        value,
-      }: {
-        start: TDate;
-        end: TDate;
-        value: number;
-      }) => {
+      const containsValidTime = ({ start, end }: { start: TDate; end: TDate }) => {
         if (minTime && isAfter(minTime, end)) {
-          return true;
+          return false;
         }
 
         if (maxTime && isAfter(start, maxTime)) {
-          return true;
+          return false;
         }
 
+        return true;
+      };
+
+      const isValidValue = (value: number) => {
         if (shouldDisableTime) {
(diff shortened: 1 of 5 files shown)


packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -222,7 +221,7 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
     components,
     componentsProps,
     date,
-    disableIgnoringDatePartForTimeValidation = false,
+    disableIgnoringDatePartForTimeValidation,
```
alexfauquette, line 224: Could be renamed to avoid double negation

```suggestion
    considerDatePartForTimeValidation,
```

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -267,44 +266,63 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
...
-            (maxTime && isAfterComparingFn(getRequestedTimePoint('start'), maxTime)) ||
-            (shouldDisableTime && shouldDisableTime(value, viewType)),
-        );
+      const isAfter = createIsAfterIgnoreDatePart(disableIgnoringDatePartForTimeValidation, utils);
+
+      const shouldDisableDatePeriod = ({
+        start,
+        end,
+        value,
+      }: {
```
alexfauquette, line 275: Doing the two verifications at the same time is disturbing

Could be split into two distinct functions:
- `containsValidTime({start, end})`
- `isValidTime(value)` 

Then `return !containsValidTime({start, end}) || !isValidTime(value)`

alexfauquette (approved):
I made some proposition, not sure they bring clarity to the code, so feel free to merge as it is

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
flaviendelangle: I don't understand this prop, we never set it to `true` and there is no issue on the core mentioning it
I'll leave it as is and try to know what was its initial goal

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
flaviendelangle: Done :+1:

Issue #4463, opened 2021-11-01
[DatePicker] Allow to change order of month and year in PickersCalendarHeader


shiwano:
- [x] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Allow to change the order of year and month in PickersCalendarHeader component for non-Western countries.

IMO: I’d be happy to customize it like this:

```tsx
<DesktopDatePicker
   renderCalendarHeaderTitle={(date) => format(date, 'yyyy/mm')}            
/>
```

The relevant code is here:
https://github.com/mui-org/material-ui/blob/db8a5f536ffb4e0dc0dadffcdfd69eebff62de27/packages/mui-lab/src/CalendarPicker/PickersCalendarHeader.tsx#L177-L200

## Examples 🌈

Actual:
<img width="328" alt="スクリーンショット 2021-11-02 0 52 02" src="https://user-images.githubusercontent.com/1025342/139700877-7a30b986-dbee-4f74-8170-88d5b488b393.png">

Expected:
<img width="335" alt="スクリーンショット 2021-11-02 0 51 46" src="https://user-images.githubusercontent.com/1025342/139700853-c3c692ac-142b-425e-a01c-e9af7c84718f.png">

I'd like to change the actual to the expected.

## Motivation 🔦

I am creating UI components for Japanese using MUI 5, but in Japan, it is not common that year follows month.
It has a merit to make this possible so that the DatePicker can be more useful in non-Western countries.

NguyenPhuonng:
@gky360 
Thanks for amazing answer :heart:

IvanHCChan:
Unfortunately, it only works at DatePicker and DesktopDatePicker but MobileDatePicker and StaticDatePicker.
Desktop:
![image](https://github.com/mui/mui-x/assets/100174803/16c3934c-fc34-4adf-8e69-ba07ff6c65ca)

Mobile:
![image](https://github.com/mui/mui-x/assets/100174803/a260d6eb-ea95-41bd-98f7-f04a9680a90a)

PR #4695, opened 2022-04-29
[DatePicker] Make month year order changeable in header

gky360:
Fixes mui/mui-x#4463.

Allow changing the order of month and year in PickersCalendarHeader by using `monthAndYear` format instead of using `month` and `year` format separately.

Ref https://github.com/mui/material-ui/pull/31888

### DIFF
--- a/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.test.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.test.tsx
@@ -39,8 +39,7 @@ describe('<CalendarPicker />', () => {
       <CalendarPicker date={adapterToUse.date('2019-01-01T00:00:00.000')} onChange={() => {}} />,
     );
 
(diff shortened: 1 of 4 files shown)


gky360:
I'll be happy if someone helps me resolve screenshot diffs detected in "argos" test.

flaviendelangle (commented):

The change makes sense for me
Could you just add a test checking that the picker does take custom `monthAndYear` format into account on this component ?

gky360:
@flaviendelangle Thank you for your reply.
I added a test to check if a custom `monthAndYear` is applied as expected.
Could you check it? https://github.com/mui/mui-x/pull/4695/commits/a4fb1d43c8386deb497567df3438b7b06f0c7d91

flaviendelangle (commented):
Perfect thanks !

flaviendelangle (approved):

PR #4708, opened 2022-04-29
[MonthPicker] New prop `shouldDisableMonth`

someone635:
Reopening here a PR I opened previously on @mui/lab

Took the liberty to add this small functionality, as I needed it in my project. I wonder if there was a rationale behind having shouldDisableYear and not shouldDisableMonth.

Use case: I want users to select editions of a monthly magazine, yet the magazine isn't systematically published every month.

The shouldDisableMonth has exactly the same functionality as shouldDisableYear, it disables the month only on the "month" view. It is still possible to access dates of that month using the "day" view and using arrows, so it may have to be combined with shouldDisableDate for 100% disabling.


packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
```diff
@@ -113,6 +113,23 @@ export interface CalendarPickerProps<TDate>
...
+   */
+  shouldDisableMonth?: (month: TDate) => boolean;
+  /**
+   * Disable specific years dynamically.
+   * Works like `shouldDisableDate` but for year selection view @DateIOType.
+   * @template TDate
+   * @param {TDate} year The year to check.
+   * @returns {boolean} If `true` the year will be disabled.
+   */
+  shouldDisableYear?: (year: TDate) => boolean;
```
flaviendelangle, line 132: We don't need to add `shouldDisableYear` here, it is already defined in `ExportedYearPickerProps`

And we could do the same for months with `ExportedMonthPickerProps`,`onMonthChange` and `shouldDisableMonth` and have `MonthPickerProps` extends it.

flaviendelangle (commented):
The feature implemented make sense, my comments are just about code organization.

packages/x-date-pickers/src/MonthPicker/MonthPicker.tsx
```diff
@@ -9,7 +9,25 @@ import { useUtils, useNow } from '../internals/hooks/useUtils';
...
 
-export interface MonthPickerProps<TDate> {
+export interface ExportedMonthPickerProps<TDate> {
+  /**
+   * Callback firing on month change @DateIOType.
+   * @template TDate
+   * @param {TDate} month The new year.
+   * @returns {void|Promise} -
+   */
+  onMonthChange?: (month: TDate) => void | Promise<void>;
```
flaviendelangle, line 19: Not related to this PR but `onMonthChange` can return a promise and not `onYearChange`, the typing feels wrong.

I will have a look

flaviendelangle (approved):
Looks great !
Thanks for your help, it should be released next week :+1:

PR #4726, opened 2022-05-02
[TimePicker] Disable and invalidate date with minutes not matching `minutesStep`

### LINKED ISSUE #4604: [TimePicker] minutesStep, handle disabled minutes
**Is your feature request related to a problem? Please describe.**
When setting the `minutesStep` prop, there's no indication to the user which minutes are not selectable.  This is nearly a bug, but it's honestly a feature.

**Describe the solution you'd like**
By default minutes that are not eligible for user interaction should be a "disabled" style, which can be overridden in the usual way.

**Describe alternatives you've considered**
I'll go to prod with it, but I'd rather not.

**Additional context**
This component rules.

step

```jsx
import TextField from "@material-ui/core/TextField";
import DateFnsAdapter from "@material-ui/lab/dateAdapter/date-fns";
import LocalizationProvider from "@material-ui/lab/LocalizationProvider";
import TimePicker from "@material-ui/lab/TimePicker";

export default function BasicTimePicker() {
  const [value, setValue] = React.useState<Date | null>(new Date());

  return (
    <LocalizationProvider dateAdapter={DateFnsAdapter}>
      <div style={{ width: 300 }}>
        <TimePicker
          label="12 hours"
          value={value}
          onChange={(newValue) => {
            setValue(newValue);
          }}
          minutesStep={15}
          renderInput={(params) => (
            <TextField {...params} margin="normal" variant="standard" />
          )}
        />
      </div>
    </LocalizationProvider>
  );
}
```

https://codesandbox.io/s/basictimepicker-material-demo-forked-fnrox?file=/demo.tsx

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| @material-ui/lab | v5.0.0-alpha.31  |

flaviendelangle:
Closes #4604

- [x] Visually disable the minutes not matching `minutesStep` in the clock
- [x] Mark the value as invalid if the current minute does not match the `minutesStep` (e.g minute = 34 when minutesStep = 5)
 

[Example](https://codesandbox.io/s/basictimepicker-material-demo-forked-ml8br3?file=/demo.js)

### DIFF
--- a/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
+++ b/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
@@ -34,11 +34,6 @@ export interface ExportedClockPickerProps<TDate> extends ExportedTimeValidationP
    */
   ampm?: boolean;
   /**
-   * Step over minutes.
-   * @default 1
-   */
-  minutesStep?: number;
-  /**
    * Display ampm controls under the clock (instead of in the toolbar).
    * @default false
    */
--- a/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
+++ b/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
@@ -262,10 +262,6 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
 
   const isTimeDisabled = React.useCallback(
     (rawValue: number, viewType: ClockPickerView) => {
-      if (date === null) {
-        return false;
-      }
-
       const isAfter = createIsAfterIgnoreDatePart(disableIgnoringDatePartForTimeValidation, utils);
 
       const containsValidTime = ({ start, end }: { start: TDate; end: TDate }) => {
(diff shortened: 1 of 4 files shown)


packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -262,10 +262,6 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
 
   const isTimeDisabled = React.useCallback(
     (rawValue: number, viewType: ClockPickerView) => {
-      if (date === null) {
```
flaviendelangle, line 265: Even if the current date is null we can check if the value is valid, we just can't check the full time.

alexfauquette (commented):
Looks good. Maybe a test "should disable clock numbers according to `minutesStep`" to check that clock picker show disables options

flaviendelangle:
@alexfauquette done

alexfauquette (approved):

PR #4731, opened 2022-05-02
[pickers] Enable mask by default when using `ampm=true`

### LINKED ISSUE #4488: [TimePicker] "Mask" prop on TimePicker doesn't seem to do anything
Hi,

Thanks for all your hard work, love MUI, please keep it up.

The "mask" prop on TimePicker doesn't seem to enforce any formatting for input the way it does on DatePicker. Here's a codesandbox demonstrating the behavior. As you can see, you can only type numbers into the DatePicker, and they conform to the __/__/____ mask. But you can type anything into the TimePicker.

https://codesandbox.io/s/basicdatepicker-material-demo-forked-dwm8x?file=/demo.js

alexfauquette:
Fix #4488

The default value of `disableMaskedInput` was set to `true` if the component used `ampm=true`

This sounds like it has been partially fixed because `acceptRegex` and `inputFormat` were adapted to accept meridian time. Only the mask was not adapted. Now it's done and tested


flaviendelangle (approved):
Seems to work as expected

PR #4736, opened 2022-05-02
[pickers] Rename prop `date` into `parsedValue` when it can contain a range

### LINKED ISSUE #4703: [pickers] Add support for `@js-joda/core`

## Current Behavior 😯

Currently when using a `DatePicker` in MUI v5 with JsJodaAdapter (from [@date-io/js-joda](https://www.npmjs.com/package/@date-io/js-joda)) an uncaught exception is thrown when changing date. The exception is: `DateTimeException: Unable to obtain LocalTime TemporalAccessor: 2021-10-06, type LocalDate`.

## Expected Behavior 🤔

The picker does not crash and the new value is passed to the `onChange` function.

## Steps to Reproduce 🕹

```
<LocalizationProvider dateAdapter={JsJodaAdapter}>
  <DatePicker
    value={LocalDate.of(2021, 10, 6)}
    onChange={(d) => console.log(d)}
    renderInput={(p) => <TextField {...p} />}
  />
</LocalizationProvider>
```

[Sandbox link](https://codesandbox.io/s/still-rgb-rmyr0?file=/src/Demo.tsx)

## Context 🔦

The curlpit seems to be on [this line](https://github.com/mui-org/material-ui/blob/24015d1e4bdf135ba6c6091141cab2299d824be5/packages/mui-lab/src/CalendarPicker/PickersCalendar.tsx#L126), when calling `utils.mergeDateAndTime`

## Your Environment 🌎

<details>
  
```
  System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 14.17.3 - C:\Program Files\nodejs\node.EXE
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1023.0), Chromium (94.0.992.38)
```

Don't know why it doesn't get detected, but I'm using Chrome 94.0.4606.71
</details>

flaviendelangle:
## Goal

Improve the consistency of the naming. 
A lot of our inner components received a `date` prop even when using the `DateRangePicker`
Things like `date: DateRange<TDate>` (or more suble, `date: TValue` which can be a range).

## Why `parsedValue` ?

We already have a `props.value` which is drilled done even if the typing does not say so 

https://github.com/mui/mui-x/blob/8459a54d94c63cb4a4636ce3b18f70657d4a3c12/packages/x-date-pickers/src/MobileDatePicker/MobileDatePicker.tsx#L46-L47

This trade-off could be re-evaluated at some point

## Work

- [x] Rename the `pickerProps.date` of `usePickerState` into `pickerProps.parsedValue`
- [x] Use `parsedValue` in all the components that are being given the `pickerProps` (`CalendarOrClockPicker` and `DateRangePickerView`)
- [ ] Write the changelog for the breaking change

## Questions

Should we rename `date` into `parsedValue` in deeper views ? Like `CalendarPicker` or `ClockPicker`
When reaching those components, we know that it is not a range so the `date` naming is not incoherent. With that being said, maybe we should aim for consistency.

## What's next ?

In `DayPicker` I kept the `date` name even if it can be given a range.
That's because I want to do deeper changes to this component to fix #4703.


alexfauquette:
> Should we rename date into parsedValue in deeper views?

For me it makes sense to have two namings, one for values which can either be a range of date or single date. And another naming for when we know what type it is

flaviendelangle:
There is a small nuance, the two criteria is "Components in which the value can be a range OR called with a prop spread from a hook (often usePickerState) on which the value can be a range".

And this nuance is ... not great :laughing:

alexfauquette (approved):
It took me some time to understand the nuance. The new naming does not cover the minimal set of variables because of the hook, but at least it covers all the needed ones:

- I see `parsedValue`, maybe it can be date or range, maybe it is only a date. If I have a doubt I take care of handling both
- I see `date` I know it is only a date, I don't need to take care

PR #4743, opened 2022-05-02
[DataGrid] add TypeScript support for pass-through sx prop

### LINKED ISSUE #4742: [DataGrid] TypeScript issue when using prop `sx` on GridToolbarContainer
### Duplicates

### Latest version

### Current behavior 😯

The types of the `GridToolbarContainer` container are declared such that it only supports the props of a native `<div>` element.  The MUI `sx` prop actually works just fine and styles the component properly, but it gives a TypeScript error:
```
TS2322: Type '{ children: Element[]; sx: { justifyContent: string; mb: number; }; }' is not assignable to type 'IntrinsicAttributes & GridToolbarContainerProps & RefAttributes<HTMLDivElement>'.   Property 'sx' does not exist on type 'IntrinsicAttributes & GridToolbarContainerProps & RefAttributes<HTMLDivElement>'.
```

I've looked at the code and I can see the issue.  The props type
```
export type GridToolbarContainerProps = React.HTMLAttributes<HTMLDivElement>;
```
is set *before* the `styled` HOC is applied.  The return type of `const GridToolbarContainerRoot = styled('div'...` is never used anywhere.

The exported `ForwardRefExoticComponent` uses the *unstyled* props type.
```
export const GridToolbarContainer = React.forwardRef<HTMLDivElement, GridToolbarContainerProps>(
```

I looked to see how other components handle this and the support is mixed.  `GridToolbarContainer` can be fixed by copying [line 12 of `GridOverlay`](https://github.com/mui/mui-x/blob/e2c89d2ea05c4b938e3776402a7adf34bf500b77/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx#L12).  I will put in a PR 😃

However in some components like [`GridMainContainer`](https://github.com/mui/mui-x/blob/e2c89d2ea05c4b938e3776402a7adf34bf500b77/packages/grid/x-data-grid/src/components/containers/GridMainContainer.tsx) it appears that `sx` is not supported at all.

### Expected behavior 🤔

I expect `sx` to be a known prop of the `GridToolbarContainer`

### Steps to reproduce 🕹

[TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQokscA3nXHAOJTAAmAKhAgAbAEbIoAYQgA7GMmAysUOgF84REuQACIAK7AA9AA8AtL2TzTAcx69adOgT0z8wWXAAiWEBAAUAJSs7HCUMHpQMnAAPNx8giLiUrLyispwqMYAvCwscABWeqgwwAQAntJyWHIAXOQEwlhm1fYANHAgonUATHCqqoYAfGpAA)

```
import {
  GridToolbarContainer
} from '@mui/x-data-grid';

function Demo() {
  return <GridToolbarContainer sx={{ justifyContent: 'flex-end', mb: 2 }}/>
}
```

### Context 🔦

I would like to avoid the error and get intellisense on the `sx` prop.

### Your environment 🌎

Not relevant as the error is easily reproduced in the TypeScript playground above.

### Order ID 💳 (optional)

_No response_

lindapaiste:
Fixes #4742 

I made this PR to fix the TypeScript issue that I was having where using the prop `sx` on `GridToolbarContainer` causes a TS error (but styles correctly).

I realized that this issue is widespread.  Styling via `sx` is supported in the JavaScript anywhere that `{...other}` is passed down to a ` styled('div')`.  I have found a bunch (but not all) of these places and added `{ sx?: SxProps<Theme> }` to the props type so that it will be supported in the types.

### DIFF
--- a/packages/grid/x-data-grid/src/components/GridRowCount.tsx
+++ b/packages/grid/x-data-grid/src/components/GridRowCount.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import { unstable_composeClasses as composeClasses } from '@mui/material';
-import { styled } from '@mui/material/styles';
+import { styled, SxProps, Theme } from '@mui/material/styles';
 import { useGridApiContext } from '../hooks/utils/useGridApiContext';
 import { getDataGridUtilityClass } from '../constants/gridClasses';
 import { useGridRootProps } from '../hooks/utils/useGridRootProps';
@@ -13,7 +13,10 @@ interface RowCountProps {
   visibleRowCount: number;
 }
 
-type GridRowCountProps = React.HTMLAttributes<HTMLDivElement> & RowCountProps;
+type GridRowCountProps = React.HTMLAttributes<HTMLDivElement> &
+  RowCountProps & {
+    sx?: SxProps<Theme>;
+  };
 
 type OwnerState = { classes: DataGridProcessedProps['classes'] };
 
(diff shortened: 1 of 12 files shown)


m4theushw (approved):
Thanks for the contribution!

Issue #4550, opened 2022-01-30
[DateRangePicker] Calendarview does not toggle on multiple input clicks


faebeee:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Desktop

Currently when I click into the inputfield, the calendarview appears. When I click again, it hides. When I click it a third time, no calendarview appears

### Expected behavior 🤔

When I click a third time into the input field, the calendarview should appear again

### Steps to reproduce 🕹

Steps:

1. Go to https://mui.com/components/date-range-picker/#responsiveness
2. Click inside the `Desktop Start` field -> Cal. View appears
3. Click inside the `Desktop Start` field -> Cal. View disappears
4. Click inside the `Desktop Start` field -> No Cal. View visible

### Your environment 🌎

Browser: Brave & Chrome

```
  System:
    OS: macOS 11.5.2
  Binaries:
    Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
  Browsers:
    Chrome: 97.0.4692.99
    Edge: Not Found
    Firefox: 95.0
    Safari: 14.1.2
  npmPackages:
    @emotion/react: ^11.4.1 => 11.7.1
    @emotion/styled: ^11.3.0 => 11.6.0
    @mui/base:  5.0.0-alpha.63
    @mui/icons-material: ^5.2.1 => 5.2.5
    @mui/lab: ^5.0.0-alpha.59 => 5.0.0-alpha.63
    @mui/material: ^5.2.3 => 5.2.7
    @mui/private-theming:  5.2.3
    @mui/styled-engine:  5.2.6
    @mui/styles: ^5.2.3 => 5.2.3
    @mui/system:  5.2.6
    @mui/types:  7.1.0
    @mui/utils:  5.2.3
    @mui/x-data-grid-pro:  5.3.0
    @mui/x-license-pro:  5.3.0
    @types/react: ^17.0.2 => 17.0.38
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.4.2 => 4.5.4
```
</details>

flaviendelangle:
[...]

PR #4747, opened 2022-05-03
[DateRangePicker] Open view on click, Enter or Space instead of focus

### LINKED ISSUE #4439: [DateRangePicker] `disableCloseOnSelect` does not work

alexfauquette:
Fix #4439

The view did not close on select, because the focus goes back to the input and DateRangePickerInput open the view when focussed, So `onOpen` was called just after `onClose`

I took the same behavior as mobile components: open the view on the click event, or when space/Enter is pressed

I did not managed to test the bug. I thought testing how many times `onOpen` is called would be enough because the bug add a `onOpen` after closing on select, but when running `yarn test:karma` I did not get the last call to `onOpen`

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
+++ b/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
@@ -308,7 +308,9 @@ export const usePickerState = <TInputValue, TValue, TDate>(
       onChange: handleInputChange,
       open: isOpen,
       rawValue: value,
-      openPicker: () => setIsOpen(true),
+      openPicker: () => {
+        setIsOpen(true);
+      },
     }),
     [handleInputChange, isOpen, value, setIsOpen],
   );
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerInput.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerInput.tsx
@@ -3,11 +3,11 @@ import { styled } from '@mui/material/styles';
 import {
   useUtils,
   executeInTheNextEventLoopTick,
-  WrapperVariantContext,
   DateInputProps,
   ExportedDateInputProps,
   MuiTextFieldProps,
   useMaskedInput,
+  onSpaceOrEnter,
 } from '@mui/x-date-pickers/internals';
 import { CurrentlySelectingRangeEndProps, DateRange } from '../internal/models/dateRange';
 import { DateRangeValidationError } from '../internal/hooks/validation/useDateRangeValidation';
@@ -94,7 +94,6 @@ export const DateRangePickerInput = React.forwardRef(function DateRangePickerInp
   const utils = useUtils<TDate>();
   const startRef = React.useRef<HTMLInputElement>(null);
   const endRef = React.useRef<HTMLInputElement>(null);
-  const wrapperVariant = React.useContext(WrapperVariantContext);
 
   React.useEffect(() => {
     if (!open) {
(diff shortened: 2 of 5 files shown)


packages/x-date-pickers/src/internals/hooks/usePickerState.ts
```diff
@@ -308,7 +308,9 @@ export const usePickerState = <TInputValue, TValue, TDate>(
       onChange: handleInputChange,
       open: isOpen,
       rawValue: value,
-      openPicker: () => setIsOpen(true),
+      openPicker: () => {
```
flaviendelangle, line 311: We can revert this change

flaviendelangle (commented):
Unless they are very complicated, I think it is worth trying to fix #4550 and the fact that clicking on the 2nd input while the 1st is focused closes the panel instead of just switching the `currentlySelectingRangeEnd` state.

That would allow us to have a global vision of the range pickers modal opening / closing behavior.

alexfauquette:
This PR fixes #4550 for free since we do not rely anymore on component focus

Not closing the panel when clicking on the second input is a bit more complex.

`DesktopTooltipWrapper` has a special `handleBlur` to check of the new focus is still in the component or not. So I modified the Popper such that `clickAway` calls the `onBlur` if possible, and `onClose` if not provided 
```ts
const handleBlur = () => {
    executeInTheNextEventLoopTick(() => {
      if (
        inputContainerRef.current?.contains(document.activeElement) ||
        popperRef.current?.contains(document.activeElement)
      ) {
        return;
      }

      onDismiss();
    });
  };
```

I'm wondering if the community component shoudl have the similar behavior, becaus for now, we can not have both view and keyboard interaction.

If I use the same code on `DatePicker` it allows to do both keyboard and mouse interactions (light theme is with the modification, dark theme is the current doc)

https://user-images.githubusercontent.com/45398769/167140090-61a85393-1c88-49bb-804a-71aa9e12a9fc.mp4

alexfauquette:
I looked at adding tests, but the focus behavior is too weird, event by setting focus manually on input, it calls `onClose` in karma test, wherase it does not call it in real browser

flaviendelangle:
@alexfauquette I don't know if it was the test you were trying to writ
But the following seems to work.

I am using `userEvent.mousePress` to correctly fire the blur / focus on the events.

```
    it.only('should not close picker when switching focus from start to end input', () => {
      const onChange = spy();
      const onAccept = spy();
      const onClose = spy();

      render(
          <WrappedDesktopDateRangePicker
              onChange={onChange}
              onAccept={onAccept}
              onClose={onClose}
              initialValue={[
                adapterToUse.date('2018-01-01T00:00:00.000'),
                adapterToUse.date('2018-01-06T00:00:00.000'),
              ]}
          />,
      );

      // Open the picker (already tested)
      openPicker({ type: 'date-range', variant: 'desktop', initialFocus: 'start' });

      // Switch to end date
      openPicker({ type: 'date-range', variant: 'desktop', initialFocus: 'end' })
      expect(onChange.callCount).to.equal(0);
      expect(onAccept.callCount).to.equal(0);
      expect(onClose.callCount).to.equal(0);
    })   
    
    it.only('should not close picker when switching focus from end to start input', () => {
      const onChange = spy();
      const onAccept = spy();
      const onClose = spy();

      render(
          <WrappedDesktopDateRangePicker
              onChange={onChange}
              onAccept={onAccept}
              onClose={onClose}
              initialValue={[
                adapterToUse.date('2018-01-01T00:00:00.000'),
                adapterToUse.date('2018-01-06T00:00:00.000'),
              ]}
          />,
      );

      // Open the picker (already tested)
      openPicker({ type: 'date-range', variant: 'desktop', initialFocus: 'end' });

      // Switch to end date
      openPicker({ type: 'date-range', variant: 'desktop', initialFocus: 'start' })
      expect(onChange.callCount).to.equal(0);
      expect(onAccept.callCount).to.equal(0);
      expect(onClose.callCount).to.equal(0);
    })
```

flaviendelangle (approved):

Issue #4761, opened 2022-05-04
x-date-pickers: Incomplete peerDependencies break usage with yarn pnp


nate-summercook:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Yarn PnP (Plug'n'Play) is more strict when it comes to specifying a library's peerDependencies / dependencies correctly.
Currently, @mui/x-date-pickers is missing the following entries in peerDependencies:
- `@babel/runtime`
- `@emotion/react` (because it's a peer dependency of `@mui/system`)
- `@emotion/styled` (because it's a peer dependency of `@mui/system`)

This results in errors like this one:
```
ERROR in resolving fallback for shared module @emotion/react
Module not found: Error: @mui/styled-engine tried to access @emotion/react (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.

Required package: @emotion/react
Required by: @mui/styled-engine@virtual:7a413bb353e548319c90adb82631b0c089acd1a802fdcf34c7a70f0d92986c9f29874633545bb7dc3d02df8b9ae9870617074e94e1a662f0e58596867aac2836#npm:5.6.1 (via /workspace-path/.yarn/__virtual__/@mui-styled-engine-virtual-d4f5f5ea46/0/cache/@mui-styled-engine-npm-5.6.1-06c551ec66-cd17d05902.zip/node_modules/@mui/styled-engine/StyledEngineProvider/)

Ancestor breaking the chain: @mui/lab@virtual:be640854c908ca72c5c45d17e3a5d76a2c579538ed551547b39edc486c90638abf94c25a34fa05a0a9167f796b8fb98e280e0bad9b44c61172fe75f4bcf39869#npm:5.0.0-alpha.80
Ancestor breaking the chain: @mui/x-date-pickers@virtual:a642cd9791514da72a2616a56d7e5a3388c1c9ae1ff46c38c21105bc127f858e9e9b8bf21f1f43f2292c03fb16041526691dbd0ab6f7ba072e2c24ae04f50e11#npm:5.0.0-alpha.2
```

### Expected behavior 🤔

If the mentioned packages were added to peerDependencies, these errors would not occur with yarn pnp.

### Steps to reproduce 🕹

Steps:

1. Set up a new react app (i.e. via CRA) `npx create-react-app mui-x-date-pickers`
2. Make sure you have yarn installed
[...]

PR #4763, opened 2022-05-04
PR for #4761: Added missing peerDependencies

nate-summercook:
Added missing peerDependencies in order to not break projects using yarn pnp, issue #4761.

### DIFF
--- a/packages/x-date-pickers/package.json
+++ b/packages/x-date-pickers/package.json
@@ -44,6 +44,7 @@
     "directory": "packages/x-date-pickers"
   },
   "dependencies": {
+    "@babel/runtime": "^7.17.2",
     "@date-io/date-fns": "^2.13.1",
     "@date-io/dayjs": "^2.13.1",
     "@date-io/luxon": "^2.13.1",
@@ -55,7 +56,6 @@
     "rifm": "^0.12.1"
   },
   "peerDependencies": {
-    "@babel/runtime": "^7.16.3",
     "@emotion/react": "^11.9.0",
     "@emotion/styled": "^11.8.1",
     "@mui/material": "^5.2.3",
@@ -72,6 +72,12 @@
       "optional": true
     },
     "dayjs": {
+      "optional": true
+    },
+    "@emotion/react": {
+      "optional": true
+    },
+    "@emotion/styled": {
       "optional": true
     },
     "luxon": {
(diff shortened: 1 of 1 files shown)


packages/x-date-pickers/package.json
```diff
@@ -55,6 +55,9 @@
     "rifm": "^0.12.1"
   },
   "peerDependencies": {
+    "@babel/runtime": "^7.16.3",
```
flaviendelangle, line 58: On the core, `@babel/runtime` is a dep, not a peer dep

I wonder why we never needed it on the data grid packages...
@mnajdova do you know what it is being used for on the core ?

flaviendelangle (commented):

packages/x-date-pickers/package.json
nate-summercook: Would work for me too to put `@babel/runtime` into dependencies. But it needs to be defined somewhere.

nate-summercook: Looks like most libraries would put that into the dependencies section. Went ahead and changed it in the PR. Also I made the emotion imports optional, as mui can also be used with styled-components.

nate-summercook:
Maybe give your TimePicker and DateTimePickers (or actually all date related components) a fix time in your tests, so they don't fail randomly ;)

flaviendelangle:
> Maybe give your TimePicker and DateTimePickers (or actually all date related components) a fix time in your tests, so they don't fail randomly ;)

We did, but it broke clearly
I must spend some time investigating...

flaviendelangle (approved):
We will need to do the same for emotion on the datagrid packages
I would like to know why we need this babel plugin on the pickers and not on the data grid

But for the pickers, the modification make sense.
I will also do the same on `@mui/x-date-pickers-pro` in a follow up PR :+1:

flaviendelangle:
> have you tried my reproduction approach (with yarn 3 & pnp) with the datagrid?

No, I will do it before trying to apply a fix :+1: 
I am just surprised nobody raised it before. But maybe yarn 3 & pnp is not used a lot yet.

packages/x-date-pickers/package.json
```diff
@@ -55,6 +56,8 @@
     "rifm": "^0.12.1"
   },
   "peerDependencies": {
+    "@emotion/react": "^11.9.0",
+    "@emotion/styled": "^11.8.1",
```
oliviertassinari, line 60: Outside of the scope that I think should have been as low as possible, I would personally assume that this change is going in the wrong direction: #32668.

PR #4772, opened 2022-05-05
[core] Update monorepo

flaviendelangle:
Will automerge

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2601,7 +2601,7 @@
 
 "@mui/monorepo@https://github.com/mui/material-ui.git#master":
   version "5.6.3"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -2601,7 +2601,7 @@
 
 "@mui/monorepo@https://github.com/mui/material-ui.git#master":
   version "5.6.3"
```
oliviertassinari, line 2603: I would have expected a different version here: https://github.com/mui/material-ui/blob/f622fa9efeea553a1f250ef28b884953b96c1543/package.json#L2

```suggestion
  version "5.6.4"
```

I personally remove the whole `"@mui/monorepo@https://github.com/mui/material-ui.git#master":` block and then run `yarn`. It seems to take care of this automatically, it could be a good tip for the future.

PR #4775, opened 2022-05-05
Update DateTimePicker.tsx

krychaxp:


alexfauquette:
Hi, thanks for pointing this error in the default. However it's a bit more complex than that. The default value seems to be computed according to the locale

The code responsible for that is here:
Here is an example with french/english locales
https://codesandbox.io/s/basicdatepicker-material-demo-forked-et5fvx?file=/demo.tsx

https://github.com/alexfauquette/mui-x/blob/33f513a8a406a550e1c06bbd013dbbb2f137edea/packages/x-date-pickers/src/DateTimePicker/shared.ts#L106:L106
```js
  const ampm = themeProps.ampm ?? utils.is12HourCycleInCurrentLocale();
```

If you want to correct the types, you will have to modify comments on typescripts, and run `yarn proptypes`. As explained in the code comment

> These PropTypes are generated from the TypeScript type definitions

krychaxp:
> Hi, thanks for pointing this error in the default. However it's a bit more complex than that. The default value seems to be computed according to the locale
> 
> The code responsible for that is here: Here is an example with french/english locales https://codesandbox.io/s/basicdatepicker-material-demo-forked-et5fvx?file=/demo.tsx
> 
> https://github.com/alexfauquette/mui-x/blob/33f513a8a406a550e1c06bbd013dbbb2f137edea/packages/x-date-pickers/src/DateTimePicker/shared.ts#L106:L106
> 
> ```js
>   const ampm = themeProps.ampm ?? utils.is12HourCycleInCurrentLocale();
> ```
> 
> If you want to correct the types, you will have to modify comments on typescripts, and run `yarn proptypes`. As explained in the code comment
> 
> > These PropTypes are generated from the TypeScript type definitions

Thanks 👏

PR #4778, opened 2022-05-05
[pickers] Add component slot for action bar

alexfauquette:
### Breaking changes

The following props have been removed: `clearable`, `showTodayButton`, `cancelText`, `okText`. 
To control the presence of action buttons, you can pass props to the `Actionbar` as follow

```jsx
<DatePicker
	componentsProps={{ actionBar: { actions: ['clear'] } }}
/>	
```

The default `ActionBar` accept a prop `actions` which indicates buttons to display and their order.
Default values are `['cancel', 'accept']` for mobile and `[]` for desktop.

If you want to have a different set of actions for mobile and desktop components, actions can be a callback taking as an input the component variant and returning the array of actions to display.

```jsx
<DatePicker
	// This example show no action on desktop, and 'clear' action otherwise
	componentsProps={{ actionBar: { actions: (variant) => variant === 'desktop' ? [] : ['clear'] } }}
/>	
```

Available actions are 'clear', 'cancel', 'accept', and 'today'. You can add other ones by overriding the `ActionBar` slot.


packages/x-date-pickers/src/internals/components/wrappers/MobileWrapper.tsx
```diff
@@ -1,56 +1,62 @@
...
-    todayText,
+    components,
+    componentsProps,
     ...other
   } = props;
 
   return (
     <WrapperVariantContext.Provider value="mobile">
-      <PureDateInputComponent {...other} {...DateInputProps} />
+      <PureDateInputComponent components={components} {...other} {...DateInputProps} />
```
flaviendelangle, line 49: We don't pass the `componentsProps` here ?
Even if none are used right now, it's maybe better to pass it to make sure we don't forget it when adding a new slot props

packages/x-date-pickers-pro/src/DesktopDateRangePicker/DesktopDateRangePicker.tsx
```diff
@@ -6,8 +6,14 @@ import {
   usePickerState,
   DateInputPropsLike,
   DesktopWrapperProps,
+  DesktopWrapperSlotsComponent,
+  DesktopWrapperSlotsComponentProps,
```
flaviendelangle, line 10: Should be named `DesktopWrapperSlotsComponentsProps`

x other variables named ` XXXComponentProps`

packages/x-date-pickers/src/DatePicker/DatePicker.tsx
```diff
@@ -2,18 +2,38 @@ import * as React from 'react';
...
+  extends Omit<DesktopDatePickerProps<TInputDate, TDate>, 'components' | 'componentsProps'>,
+    Omit<MobileDatePickerProps<TInputDate, TDate>, 'components' | 'componentsProps'> {
   /**
    * CSS media query when `Mobile` mode will be changed to `Desktop`.
    * @default '@media (pointer: fine)'
    * @example '@media (min-width: 720px)' or theme.breakpoints.up("sm")
    */
   desktopModeMediaQuery?: string;
+  components?: Partial<DatePickerSlotsComponent>;
+  componentProps?: Partial<DatePickerSlotsComponentProps>;
```
flaviendelangle, line 36: Should be named `componentsProps`

packages/x-date-pickers/src/TimePicker/TimePicker.tsx
```diff
@@ -2,18 +2,38 @@ import * as React from 'react';
...
+  extends Omit<DesktopTimePickerProps<TInputDate, TDate>, 'components' | 'componentsProps'>,
+    Omit<MobileTimePickerProps<TInputDate, TDate>, 'components' | 'componentsProps'> {
   /**
    * CSS media query when `Mobile` mode will be changed to `Desktop`.
    * @default '@media (pointer: fine)'
    * @example '@media (min-width: 720px)' or theme.breakpoints.up("sm")
    */
   desktopModeMediaQuery?: string;
+  components?: Partial<TimePickerSlotsComponent>;
+  componentProps?: Partial<TimePickerSlotsComponentProps>;
```
flaviendelangle, line 36: Should be named `componentsProps`

packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onAccept: () => void;
+  onClear: () => void;
+  onCancel: () => void;
+  onSetToday: () => void;
+}
+
+const PickersActionBarRoot = styled(DialogActions)<{
+  ownerState: PickersActionBarProps;
+}>(({ ownerState }) => ({
+  ...((ownerState.actions?.includes('clear') || ownerState.actions?.includes('today')) && {
```
flaviendelangle, line 25: Why do we only add this CSS override for some actions ?

alexfauquette: To have the same behavior as the `<PickersModalDialogActions />` I'm replacing

```jsx
const PickersModalDialogActions = styled(DialogActions)<{
  ownerState: PickersModalDialogProps;
}>(({ ownerState }) => ({
  ...((ownerState.clearable || ownerState.showTodayButton) && {
    // set justifyContent to default value to fix IE11 layout bug
    // see https://github.com/mui-org/material-ui-pickers/pull/267
    justifyContent: 'flex-start',
    '& > *:first-of-type': {
      marginRight: 'auto',
    },
  }),
}));
```

Makes me realize that desktop has a different behavior

```jsx
const PickersPopperAction = styled(DialogActions)<{
  ownerState: PickerPopperProps;
}>(({ ownerState }) => ({
  ...(ownerState.clearable
    ? {
        justifyContent: 'flex-start',
        '& > *:first-of-type': {
          marginRight: 'auto',
        },
      }
    : { padding: 0 }),
}));
```

And the `marginRight: 'auto'` does not have any more sens now that users can modify buttons order.

I think I missed a big blocker: the difference between mobile and desktop which have different style and behavior

https://codesandbox.io/s/basicdatepicker-material-demo-forked-8gvp4h?file=/demo.tsx

So I propose to add 3 props to this component: 
- `isDesktop` such that developers can use `style(PickersActionBar)` to apply different style on descktop/mobile
- add `actionsForMobile` and `actionsForDesktop` by default to `undefined` if the dev specify them it overrides the `actions` props

flaviendelangle: You can use `WrapperVariantContext` instead of adding a `isDesktop` prop.

> add actionsForMobile and actionsForDesktop by default to undefined if the dev specify them it overrides the actions props

I propose to rather implement https://github.com/mui/material-ui/issues/32144 in a follow up PR with the variant as a param.

packages/x-date-pickers/src/internals/components/wrappers/MobileWrapper.tsx
alexfauquette: That's a bit tricky. The components are already in `DateInputProps`

Maybe it will be easier to maintain if I move it from `DateInputProps` to the one you propose. But for now, I prefer to do this correction in a dedicated PR. The reason of the `DateInputProps` is still unclear to me

packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
alexfauquette: I added the callback for `actions` + some tests

pganster:
Thanks for another big PR merged. I'm currently wondering, how is it now intended to translate the `accept` and `clear` etc.  button texts, since their corresponding text properties were removed? 

I'm guessing by providing a `CustomActionBar`, as mentioned in the release notes. But I've not yet found an example how to implement such an action bar (especially with the same design and logic as the normal buttons). Is there a documentation on how to implement this?

alexfauquette:
I did not had time to write a doc page for this yet. I'm not sure to understand your question

For translating texts, you can use [localization](https://mui.com/x/react-date-pickers/localization/) (more keys are coming)

For creating a custom component, the easiest, for now, is maybe to have a look at how the default component is made 
https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx

flaviendelangle:
I started a doc page but I implemented #5015 before to be able to do static examples (which are better to display customization in the doc I think).

But yes, the doc page should be done quickly

flaviendelangle:
> I need to wrap LocalizationProvider with ThemeProvider (I had it the other way around)

For that one, it would be nice to add a note on the doc saying that the `ThemeProvider` must be higher in the tree than `LocalizationProvider` @alexfauquette

PR #4783, opened 2022-05-06
[core] Rework `DayPicker` api

flaviendelangle:
Follow up on #4736 with a small potential bug fix.

- [x] Always pass an array of dates to `DayPicker` and rename is `selectedDays` instead of `date`
- [x] Rename `onChange` into `onSelectedDayChange` since it's not a callback of the `value` prop (we only pass the modified date and let the upper component decide how it impacts the list of selected days)

## What's next

- Test the `DayPicker` component independently of the context


DanailH:
I checked it and it looks good. I also took the time to run it locally. 
One note - since this is a breaking change (changing the API) I propose to add a small migration guide like we used to do for the DataGrid explaining what has changed in the PR so that whoever is doing the release can follow up and write the correct changelog.

DanailH (approved):

alexfauquette (approved):

PR #4790, opened 2022-05-07
[core] Remove docs duplication with monorepo

oliviertassinari:
This is a proposal to fix the current duplication of files between the monorepo and this one. It's meant to help better isolate each repository. Here would find it OK to have to update the MUI X repository once a bug is fixed in MUI Core? The discussion started in https://github.com/mui/mui-x/pull/4711/files#r862923284

For instance, the current CSS we have for code formatting is broken, it looks like this on Safari (KO):

Screenshot 2022-05-07 at 20 20 52

vs. on Chrome (OK)

Screenshot 2022-05-07 at 20 21 22

https://mui.com/material-ui/getting-started/installation/

This is because of some subtle way `<pre>` font size behaves in Safari. Now to fix this, would say it's OK that we have to update `docs/public/static/styles/prism-okaidia.css` in both repositories? I don't know, but it's clearly not ideal. I'm saying I don't know because there is a cost to the change I'm proposing here: contributors have to run the docs of the mono repository to get things displayed correctly. It will fly with maintainers, but contributors?

If we decide against this approach, then I think that we can still remove some folders "sponsors", "branding", "images/users", etc. and:

https://github.com/mui/mui-x/blob/989f8c170ab7d67e3ce48eb6bc581c06e308b5e4/docs/public/_redirects#L4

We could also write a script to run that copy & paste the right files.

---

Personally, I'm more for the current duplication, but I wanted to bring this option to the table 😁


oliviertassinari:
Actually, forget about this, it's probably a bad idea. Closing 😁

PR #4794, opened 2022-05-07
[core] System has no singleton so should be direct deps

oliviertassinari:
Something I have noticed as wrong having a quick look at #4761.


packages/x-date-pickers/package.json
```diff
@@ -49,6 +49,7 @@
     "@date-io/dayjs": "^2.13.1",
     "@date-io/luxon": "^2.13.1",
     "@date-io/moment": "^2.13.1",
+    "@mui/system": "^5.2.3",
```
oliviertassinari, line 52:
```suggestion
    "@mui/system": "^5.2.8",
```

yarn.lock
```diff
@@ -2644,7 +2644,7 @@
     jss-plugin-vendor-prefixer "^10.8.2"
     prop-types "^15.7.2"
 
-"@mui/system@^5.6.4":
+"@mui/system@^5.2.3", "@mui/system@^5.2.8", "@mui/system@^5.6.4":
```
oliviertassinari, line 2647:
```suggestion
"@mui/system@^5.2.8", "@mui/system@^5.6.4":
```

yarn.lock
```diff
@@ -2644,7 +2644,7 @@
     jss-plugin-vendor-prefixer "^10.8.2"
     prop-types "^15.7.2"
 
-"@mui/system@^5.6.4":
+"@mui/system@^5.2.8", "@mui/system@^5.6.4":
```
flaviendelangle, line 2647: If it's a dep, we can put it in its latest version like `@mui/utils` right ?

oliviertassinari: I think that the wider the range we set, the better, it will minimize version duplication among developers.

oliviertassinari:
My initial statement is false, @mui/material doesn't have a singleton, it's @mui/system that has it. So in theory, I believe that the instructions in https://mui.com/x/react-data-grid/getting-started/#installation are not correct. It should be `@mui/material` as a dependency and `@mui/system` as a peer dependency. But then if we do this, we break everything as `@mui/material` doesn't correctly list `@mui/system` as a peer dependency. So in the meantime to keep the npm dependency list sound, e.g. to work with yarn PnP, I would propose we remove the dependency on @mui/system.

Issue #2998, opened 2021-10-26
[DataGrid] Checkbox ripple not consistent when navigating with arrows


alexfauquette:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When navigating with arrows on the select column, checkbox's Ripple are not removed when going up

### Steps to reproduce 🕹

Steps:

1. Go on [https://mui.com/x/](https://mui.com/x/)
2. click on a checkbox
3. use arrows to go down, everything is normal
4. use arrows to go up, each checkbox keep its Ripple after being blured
5. repeat up and dow to increase the number of ripples

### Your environment 🌎

Chrome and Firefox

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  System:
    OS: Linux 5.13 Ubuntu 21.10 21.10 (Impish Indri)
    CPU: (8) x64 Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
    Memory: 3.50 GB / 15.35 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 16.8.0 - ~/.nvm/versions/node/v16.8.0/bin/node
    Yarn: 1.21.1 - /usr/bin/yarn
    npm: 7.21.0 - ~/.nvm/versions/node/v16.8.0/bin/npm
  Managers:
    Apt: 2.3.9 - /usr/bin/apt
    Composer: 1.9.0 - /usr/local/bin/composer
    pip2: 19.3.1 - ~/.local/bin/pip2
    pip3: 20.1.1 - ~/anaconda3/bin/pip3
    RubyGems: 3.2.5 - /usr/bin/gem
  Utilities:
    CMake: 3.18.4 - /usr/bin/cmake
    Make: 4.3 - /usr/bin/make
    GCC: 11.2.0 - /usr/bin/gcc
    Git: 2.32.0 - /usr/bin/git
    FFmpeg: 4.4 - /usr/bin/ffmpeg
  Virtualization:
    Docker: 20.10.8 - /usr/bin/docker
  IDEs:
    Nano: 5.6.1 - /usr/bin/nano
    VSCode: 1.61.2 - /snap/bin/code
    Vim: 8.2 - /usr/bin/vim
  Languages:
    Bash: 5.1.8 - /usr/bin/bash
    Go: 1.14.1 - /usr/local/go/bin/go
    Java: 1.8.0_302 - /usr/bin/javac
[...]

PR #4795, opened 2022-05-07
[data grid] Better reflect the dependency on Material UI

oliviertassinari:
Per https://github.com/mui/mui-x/issues/2998#issuecomment-1036337411 our dependency on Material UI is `    "@mui/material": "^5.4.1",`. Now, since we set the dependency as a peer dependency, we should keep the scope as broad as possible on MUI Core to avoid bundle size duplication.

Going forward, I think that we will need to revert changes like https://github.com/mui/mui-x/pull/4524/files#diff-a8bc128b9870c999f703a361035d8e5dd7e6283f719a3dc2db38d8658ce21f28R45 that touches packages.json files in the `/packages/` folder.


packages/x-date-pickers-pro/package.json
```diff
@@ -42,11 +42,11 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.17.2",
-    "@date-io/date-fns": "^2.14.0",
-    "@date-io/dayjs": "^2.14.0",
-    "@date-io/luxon": "^2.14.0",
-    "@date-io/moment": "^2.14.0",
-    "@mui/utils": "^5.7.0",
+    "@date-io/date-fns": "^2.11.0",
```
flaviendelangle, line 45: Please don't downgrade `@date-io`, I'm doing releases on it so I need the latest features at any time

oliviertassinari: Right, OK it matches with this part of the PR description:

> For @date-io/* it's less obvious to me if my proposed change is a step forward or backward. The big difference is that developers won't install it on their own, so there are no duplication upsides, and the downside is that for fixes done in @date-io/ it's harder for developers to upgrade the access to them. So, maybe to revert?

flaviendelangle: More than fixes, I need the new features, the pickers will not work at all without 2.14
And I suspect it will be the same for the next upgrades of `@date-io`

packages/grid/x-data-grid-generator/package.json
```diff
@@ -45,7 +45,7 @@
   },
   "peerDependencies": {
     "@mui/icons-material": "^5.2.5",
-    "@mui/material": "^5.2.8",
+    "@mui/material": "^5.2.3",
```
oliviertassinari, line 48: It sounds like it should be 
```suggestion
    "@mui/material": "^5.4.1",
```

per https://github.com/mui/mui-x/issues/2998#issuecomment-1036337411

packages/x-date-pickers-pro/package.json
oliviertassinari: Yeah OK, agree, let's do this. I'm updating the PR

flaviendelangle (approved):
I would like @m4theushw review on this one

packages/grid/x-data-grid-generator/package.json
```diff
@@ -44,8 +44,8 @@
     "yargs": "^17.5.1"
   },
   "peerDependencies": {
-    "@mui/icons-material": "^5.2.5",
-    "@mui/material": "^5.2.8",
+    "@mui/icons-material": "^5.4.1",
+    "@mui/material": "^5.4.1",
```
oliviertassinari, line 48: This might be considered a breaking change, but I would argue that the breaking change was already released in https://github.com/mui/mui-x/issues/2998#issuecomment-1036337411 so we might as well be upfront with it.

packages/grid/x-data-grid-pro/package.json
```diff
@@ -52,8 +52,8 @@
     "reselect": "^4.1.5"
   },
   "peerDependencies": {
-    "@mui/material": "^5.2.8",
-    "@mui/system": "^5.2.8",
+    "@mui/material": "^5.4.1",
+    "@mui/system": "^5.4.1",
```
oliviertassinari, line 56: This might be considered a breaking change, but I would argue that the breaking change was already released in https://github.com/mui/mui-x/issues/2998#issuecomment-1036337411 so we might as well be upfront with it.

Issue #4737, opened 2022-05-02
[data grid] `onPreferencePanelClose` callback gets called twice when closing columns/filters Preference panel


tielushko:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

1. When closing the toolbar columns or filter panel by clicking the filter/columns buttons, the onPreferencePanelClose callback gets called twice. 

2. Clicking away from filter/columns panel only calls it once.

3. Density and export buttons don't trigger the callback at all.

### Expected behavior 🤔

1. All the preference panel items trigger the onPreferencePanelClose callback once in all three cases above.

### Steps to reproduce 🕹

Steps:

1. Access this sandbox. https://codesandbox.io/s/datagridprodemo-material-demo-forked-2lswmm?file=/demo.tsx:0-711 
2. Open the console. 
3. Go through steps 1-3 mentioned in "current behavior" and observe the output.

### Context 🔦

Trying to export the state using the apiRef.current.exportState() into the preferences Database for storing and retrieving user state at a later time. 

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

### Order ID 💳 (optional)

31986

DanailH:
Hi, @tielushko thanks for raising this. Your concern is partially valid.
In the first case when you for example open the Filter panel and then close it by clicking the `FILTERS` button does call the `onPreferencePanelClose` twice. This should not happen.
Regarding the second concern where the Density and Export menus when closed does not call the `onPreferencePanelClose` - these 2 buttons open a `Menu` component, not a preference panel. I think a correct solution would be to potentially add a callback like `onGridMenuClose` to allow the user to react to the close. It wouldn't be possible to call the `onPreferencePanelClose` in that case as the component is not a Preference Panel.

flaviendelangle:
@DanailH according to the label description, `plan: Pro` is for issues that impact pro users, even if they are on the community codebase

flaviendelangle:
(sorry I did not saw your assignment :grimacing:)

tielushko:
The other concern I have found is when the Column Menu panel gets closed by clickAway, the even is raised, but the apiRef.current.exportState() function still has the preferencePanel property set to true, so that if the page is reloaded and that state  gets restored, it will open that panel right away.

shaselboeck:
[...]

PR #4810, opened 2022-05-09
[data grid] Only publish `preferencePanelClose` event once when closing panel by clicking on another panel button

flaviendelangle:
Part of #4737

The check in `hidePreferences` to know if the panel is opened was based on the state from the state render.
When calling `hidePreferences` synchronously after `showPreferences`, we could have an outdated state and thus an invalid check leading to an unwanted event publication.


DanailH (approved):

DanailH:
Thanks for opening a PR. I changed it to "Part of #". I'll create one more to add a callback for the `GridMenu` as that was the second problem described in the issue.

PR #4807, opened 2022-05-09
[DatePicker] Fix keyboard accessibility to first/last year

alexfauquette:
Fix #4806

Inspired from the `isTimeDisabled` in `<ClockPicker />`

TODO: 
- [x] add tests

Issue #4806, opened 2022-05-09
Last allowed year in Year view is not accessible with keyboard


fymmot:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When interacting with the DatePicker using keyboard, I can tab to the view-switching button and open the Year view.
I can then change my selection using the arrow keys, and confirm my choice using SPACE.

However, if I have entered a maxDate such as 2024-01-25, there is a bug where the picker does not allow me to select the last allowed year, i.e. 2024. 

![image](https://user-images.githubusercontent.com/6492106/167360902-53597f26-2bbf-4f58-8500-147beb54f410.png)

### Expected behavior 🤔

A keyboard user should be able to access all the displayed year buttons in the Year view.

### Steps to reproduce 🕹

Steps:

1. Open the reproduction CodeSandbox: https://codesandbox.io/s/basicdatepicker-material-demo-forked-s66xgc 
2. Open the calendar
3. Click on the button next to the year to open Year view.
4. Press tab once to move focus to the currently selected year button (the focus indicator is very subtle)
5. Use arrow keys to move between the allowed years (2022, 2023 and 2024)
6. You will not be able to reach 2024.

### Context 🔦

We encountered this issue when evaluating the accessibility of the Datepicker for use on a major Swedish travel website. 

Being able to access all features of the datepicker using the keyboard is very important.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

Forked Basic usage example from https://mui.com/x/react-date-pickers/date-picker/
</details>

alexfauquette:
Thanks the nice description

The problem seems to come from year validation which complete the date with the current day by default. So it compares you max date `2024-01-24` with `2024-05-09` (today day and month and 2024)

The Behavior should be the same as the clock: A Year is valid if it contains at least a valid date, even if the current month/day are not valid

PR #4814, opened 2022-05-09
[pickers] Restructure props in `MonthPicker` / `YearPicker` and `DayPicker`

flaviendelangle:
## Goal 

- Have `MonthPicker` / `YearPicker` / `DayPicker` usable in a standalone way. They should not rely on abstractions in `CalendarPicker` and they should have a coherent API when used alone (no required prop that could be nullable for instance).

- Avoid JSDoc duplication for documented props like `disablePast`, it leads to inconsistent descriptions / default value.

### Packages

- [x] Create new centralized interfaces `DayValidationProps`, `MonthValidationProps`, `YearValidationProps` to avoid defining props like `shouldDisableMonth` or `disablePast` in several interfaces with inconsistent descriptions.

- [x] Remove `onMonthChange` prop from `MonthPicker` as it is equivalent to `onChange` and instead wrap the `onChange` passed to `MonthPicker` from `CalendarPicker` to call `onMonthChange`.

- [x] Remove `onYearChange` prop from `YearPicker` as it is equivalent to `onChange` and instead wrap the `onChange` passed to `YearPicker` from `CalendarPicker` to call `onYearChange`.

- [x] When changing year, do not go to the closest non-disabled date in `YearPicker` but in `CalendarPicker` because `YearPicker` should not care about the availability of days inside the year.

- [x] When changing month, go to the closest non-disabled date in `CalendarPicker` (was not done at all)

- [x] Remove the `isDateDisabled` prop from `YearPicker` and `MonthPicker` and instead add `maxDate` / `minDate` / `disableFuture` / `disablePast` to be able to use those components standalone instead of relying on an outside abstraction.

- [x] Apply default `minDate` / `maxDate` directly in the validation function to avoid missing it in some components

### Tests

- [x] Write test where we select a non disabled month but the day is disabled (should go to the closest non-disabled date)
- [x] Write test with `minDate` / `maxDate`  in `YearPicker` and `MonthPicker` 

## What's next 

- Clarify why on `CalendarProps` we sometime spread `other` and / or `calendarState`. It makes it difficult to follow props and I don't think the sub-components have enough props to justify those spreads.

- Move the [Sub-components](https://mui.com/x/react-date-pickers/date-picker/#sub-components) section into standalone pages for `YearPicker`, `MonthPicker` and `DayPicker`

## Breaking change

- The props of `MonthPicker` / `YearPicker` and `DayPicker` have been reworked to make them more consistent for a standalone usage (#4814) @flaviendelangle

  **MonthPicker**: The prop `onMonthChange` has been removed, you can use `onChange` instead since every change is a month change

  **YearPicker**: The prop `onYearPicker` has been removed, you can use `onChange` instead since every change is a year change

  **DayPicker**: The prop `isDateDisabled` has been removed, you can now use the same validation props as for the other components (`maxDate`, `minDate`, `shouldDisableDate`, `disableFuture` and `disablePast`)


packages/x-date-pickers/src/CalendarPicker/useCalendarState.tsx
```diff
@@ -152,36 +152,32 @@ export const useCalendarState = <TDate extends unknown>({
...
-      validateDate(utils, day, {
-        disablePast,
-        disableFuture,
-        minDate,
-        maxDate,
-        shouldDisableDate,
-      }) !== null,
-    [disableFuture, disablePast, maxDate, minDate, shouldDisableDate, utils],
-  );
+  const isDayDisabled = useIsDayDisabled({
```
flaviendelangle, line 155: Just a small wrapper to be re-used in `DayPicker`

packages/x-date-pickers/src/YearPicker/YearPicker.tsx
```diff
@@ -83,29 +66,44 @@ export const YearPicker = React.forwardRef(function YearPicker<TDate>(
...
-  const theme = useTheme();
-  const utils = useUtils<TDate>();
-
   const selectedDate = date || now;
   const currentYear = utils.getYear(selectedDate);
   const wrapperVariant = React.useContext(WrapperVariantContext);
   const selectedYearRef = React.useRef<HTMLButtonElement>(null);
   const [focusedYear, setFocusedYear] = React.useState<number | null>(currentYear);
 
+  const isYearDisabled = React.useCallback(
```
flaviendelangle, line 85: Copies from #4807

alexfauquette (commented):
It looks good

I had some difficulties with `isMonthDisabled` because as it is I would imagine it checking "is January disabled?" but instead it checks "is January 1997 disabled?" but I'm not sure we can clarify this point

flaviendelangle:
> I had some difficulties with isMonthDisabled because as it is I would imagine it checking "is January disabled?" but instead it checks "is January 1997 disabled?" but I'm not sure we can clarify this point

I guess you can have use cases where you want to disable January in all years and use cases where you want to disable January just in one year
The good thing with the current `isMonthDisabled` is that you can do both

packages/x-date-pickers/src/internals/hooks/validation/useTimeValidation.ts
```diff
@@ -47,20 +47,31 @@ export type TimeValidationError =
   | 'shouldDisableTime-seconds'
   | null;
 
-export const validateTime: Validator<any, TimeValidationProps<any, any>> = (
-  utils,
+export const validateTime: Validator<any, TimeValidationProps<any, any>> = ({
+  adapter,
```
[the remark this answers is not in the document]
flaviendelangle: This change is usefull for the other validators.
`validateTime` is the only validator not using `adapter.defaultDates`, but I keep the same signature to be able to use the same wrappers.

packages/x-date-pickers/src/YearPicker/YearPicker.tsx
alexfauquette: I let you add the "Fix #4806" in PR description, and I close #4807?

alexfauquette (approved):

packages/x-date-pickers/src/internals/hooks/useViews.tsx
```diff
@@ -9,6 +9,11 @@ export type PickerOnChangeFn<TDate> = (
   selectionState?: PickerSelectionState,
 ) => void;
 
+export type NonNullablePickerChangeHandler<TDate> = (
```
flaviendelangle, line 12: I will rename `PickerOnChangeFn` `NullablePickerChangeHandler` in a follow up PR

packages/x-date-pickers/src/CalendarPicker/useCalendarState.tsx
```diff
@@ -30,7 +30,7 @@ export const createCalendarStateReducer =
     action:
       | ReducerAction<'finishMonthSwitchingAnimation'>
       | ReducerAction<'changeMonth', ChangeMonthPayload<TDate>>
-      | ReducerAction<'changeFocusedDay', { focusedDay: TDate }>,
+      | ReducerAction<'changeFocusedDay', { focusedDay: TDate | null }>,
```
flaviendelangle, line 33: When going to a view without changing the current date, we remove the focus

packages/x-date-pickers/src/internals/utils/date-utils.ts
```diff
@@ -53,21 +53,21 @@ export const findClosestEnabledDate = <TDate>({
...
-      if (!shouldDisableDate(backward)) {
+      if (!isDateDisabled(backward)) {
         return backward;
       }
       backward = utils.addDays(backward, -1);
     }
   }
 
-  return today;
+  return null;
```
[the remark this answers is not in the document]
alexfauquette: I agree it was a bad idea because we need to keep in mind that there is not necessarily an enabled date, and dealing with this edge case is not the responsibility of this function

For the second point I think it makes more sens to move to `null` because having `onChange` called with a disabled date sounds like a bug

alexfauquette:
Do you need a review or do you have stuff to add in this PR?

flaviendelangle:
I think I'm good for a review :+1:

packages/x-date-pickers/src/internals/hooks/validation/useDateValidation.ts
```diff
@@ -41,42 +18,76 @@ export type DateValidationError =
...
-    case Boolean(maxDate && utils.isAfterDay(date, maxDate)):
+    case Boolean(maxDate && adapter.utils.isAfterDay(date, maxDate)):
       return 'maxDate';
 
     default:
       return null;
   }
 };
 
+export const useIsDayDisabled = <TDate>({
```
alexfauquette, line 65: Should we put in this file all the validations (adding month and year validation)

Such that in this folder we test the edge cases, and in components only do one test to check diasble has been applied to the component

packages/x-date-pickers/src/MobileDatePicker/MobileDatePicker.test.tsx
```diff
@@ -22,29 +22,6 @@ const WrappedMobileDatePicker = withPickerControls(MobileDatePicker)({
 describe('<MobileDatePicker />', () => {
   const { clock, render } = createPickerRenderer({ clock: 'fake', clockConfig: new Date() });
 
-  it('selects the closest enabled date if selected date is disabled', () => {
```
alexfauquette, line 25: This test is still passing

https://codesandbox.io/s/basicdatepicker-material-demo-forked-kegr9m?file=/demo.tsx

flaviendelangle: Yes, I just moved it lower in the component tree to not just test it on `MobileDatePicker` :+1:

alexfauquette: My bad, just found it in the Calendar tests

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
```diff
@@ -260,14 +337,14 @@ const CalendarPicker = React.forwardRef(function CalendarPicker<TDate>(
 
   React.useEffect(() => {
     if (date && isDateDisabled(date)) {
-      const closestEnabledDate = findClosestEnabledDate<TDate>({
+      const closestEnabledDate = findClosestEnabledDate({
         utils,
         date,
-        minDate,
-        maxDate,
+        minDate: minDate ?? defaultDates.minDate,
```
alexfauquette, line 343: default values are already set when doing 
```jsx
{
  minDate = defaultDates.minDate,
  maxDate = defaultDates.maxDate,
  ...other
} = props;
```

alexfauquette (approved):
It looks good

packages/x-date-pickers/src/internals/hooks/validation/useDateValidation.ts
flaviendelangle: Yes I think it would be a good idea to move all date-related validations in the `validation` folder

PR #4826, opened 2022-05-10
[DataGrid] Allows to customize variant of value input in filter panel

### LINKED ISSUE #4822: [data grid] Allow GridFilterPanel filterFormProps valueInputProps variant customization
### Duplicates

### Latest version

### Summary 💡

Hi,

I found when customizing the `GridFilterPanel` by setting `filterFormProps` the `variant` prop would not have any effect in the `valueInputProps`, but work for `operatorInputProps` and `columnInputProps`. Since I am not sure if this is by design I set this as a feature request.

### Examples 🌈

```
<GridFilterPanel
  filterFormProps={{
    valueInputProps: {variant: 'outlined'},     // <- Has no effect
    columnInputProps: {variant: 'outlined'},    // <- Works
    operatorInputProps: {variant: 'outlined'},  // <- Works
  }}
/>
```
Screenshot from 2022-05-10 12-56-10

### Motivation 🔦

On the `<DataGridPro/>` component we leverage the `components` prop to inject a custom dialog (by replacing the `<GridFooter/>` with `<><GridFooter/><CustomDialog/></>`) in which we want to reuse the `<GridFilterPanel/>` but with different theming. We inject the dialog in the components prop so we have access to the context object which is needed to render the GridFilterPanel. 

### Order ID 💳 (optional)

39800

alexfauquette:
Fix #4822

The idea for know is to pass customization props to the filter input

Special case for Autocomplet input for which TextFields are not the root component


packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx
```diff
@@ -476,6 +498,7 @@ const GridFilterForm = React.forwardRef<HTMLDivElement, GridFilterFormProps>(
               applyValue={applyFilterChanges}
               focusElementRef={valueRef}
               {...currentOperator.InputComponentProps}
+              {...valueInputTextFieldProps}
```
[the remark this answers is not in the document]
alexfauquette: 👍That was another option I had in mind

About the naming, what do you think of `TextFieldProps` instead of `InputComponentProps` to be more specific?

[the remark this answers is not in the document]
m4theushw: We can't know if the input component is a text field or not, it could be anything. The filter operator already has `InputComponentProps`. I was considering this name for consistency.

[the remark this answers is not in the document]
alexfauquette: But `variant` is not part of the `InputComponentProps`.
The InputComponent will be a `FilledInput`, `OutlinedInput` or `Input` component depending on the `variant` prop value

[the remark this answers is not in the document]
m4theushw: My idea was to allow any prop in `InputComponentProps`, not restricting to only those from the text  field variants. It's the role of the current input component to use the `variant` received.

As example, the input component of the operators of a `singleSelect` column is not a TextField, but an Autocomplete. We need to forward the prop to the element that can use it.

```diff
diff --git a/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx b/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
index e757992be..d3fa254ca 100644
--- a/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
@@ -24,7 +24,7 @@ const isOptionEqualToValue: GridFilterInputMultipleSingleSelectProps['isOptionEq
 const filter = createFilterOptions<any>();
 
 function GridFilterInputMultipleSingleSelect(props: GridFilterInputMultipleSingleSelectProps) {
-  const { item, applyValue, type, apiRef, focusElementRef, ...other } = props;
+  const { item, applyValue, type, apiRef, focusElementRef, variant = 'standard', ...other } = props;
   const id = useId();
 
   const resolvedColumn = item.columnField ? apiRef.current.getColumn(item.columnField) : null;
@@ -118,7 +118,7 @@ function GridFilterInputMultipleSingleSelect(props: GridFilterInputMultipleSingl
           }}
           inputRef={focusElementRef}
           type={'singleSelect'}
-          variant="standard"
+          variant={variant}
         />
       )}
       {...other}
```

docs/data/data-grid/filtering/filtering.md
```diff
@@ -299,6 +299,11 @@ More details are available in the demo.
 | `operatorInputProps`     | `MuiDataGrid-filterFormOperatorInput`     |
 | `valueInputProps`        | `MuiDataGrid-filterFormValueInput`        |
 
+The value input is a spacial case, because it can be any input component.
```
cherniavskii, line 302:
```suggestion
The value input is a special case, because it can be any input component.
```

docs/data/data-grid/filtering/CustomFilterPanelContent.js
```diff
@@ -69,6 +69,10 @@ export default function CustomFilterPanelContent() {
                 size: 'small',
                 sx: { mt: 'auto' },
               },
+              InputComponentProps: {
```
cherniavskii, line 72: I wonder if we can reflect `FilterFormValueInput > currentOperator.InputComponent` hierarchy in here?
I would expect something like this:

```ts
filterFormProps: {
  valueInputProps: {
    // will be passed to `FilterFormValueInput`
    // ...
    
    InputComponentProps: {
      // will be passed to `currentOperator.InputComponent`
    }
  }
}
```

Especially since `InputComponentProps` are only passed to value input, not to other inputs

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputValueProps.ts
```diff
@@ -1,10 +1,11 @@
 import * as React from 'react';
+import { TextFieldProps } from '@mui/material/TextField';
 import { GridFilterItem } from '../../../models/gridFilterItem';
 
-export interface GridFilterInputValueProps {
+export type GridFilterInputValueProps<T> = {
```
cherniavskii, line 5: Can we type it better?

```ts
import type { GridApiCommon } from '../../../models/api/gridApiCommon';
import type { GridApiCommunity } from '../../../models/api/gridApiCommunity';

export type GridFilterInputValueProps<Api extends GridApiCommon = GridApiCommunity> = {
  // ...
  apiRef: React.MutableRefObject<Api>;
```

Maybe something like this?
The `Api` is optional, so no need for `GridFilterInputValueProps<any>` in other places
What do you think?

docs/data/data-grid/filtering/filtering.md
```diff
@@ -299,6 +299,11 @@ More details are available in the demo.
...
+You can either [create your own `InputComponent`](#create-a-custom-operator).
+You can provide `InputComponentProps` which will be passed directly to `InputComponent`.
```
cherniavskii, line 304:
```suggestion
You can either [create your own `InputComponent`](#create-a-custom-operator) or provide `InputComponentProps` that will be passed directly to `InputComponent`.
```

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputValueProps.ts
cherniavskii: Correct

docs/data/data-grid/filtering/CustomFilterPanelContent.js
alexfauquette: done

packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputMultipleSingleSelect.tsx
```diff
@@ -2,30 +2,28 @@ import * as React from 'react';
...
+      | 'onChange'
+      | 'value'
+      | 'id'
+      | 'filterOptions'
+      | 'isOptionEqualToValue'
+      | 'limitTags'
+      | 'multiple'
+      | 'color'
+    >,
+    GridFilterInputValueProps<GridApiCommon> {
```
cherniavskii, line 26: Is `GridApiCommon` necessary here?

alexfauquette: effectively it's not

m4theushw (approved):

PR #4835, opened 2022-05-11
[data grid] Temporarily add back the row grouping on the Pro package

flaviendelangle:
See [this Slack discussion](https://mui-org.slack.com/archives/G011VC970AW/p1652208783363809)

I duplicated everything, except the module augmentation which caused conflict (so I removed it from premium)


packages/grid/x-data-grid-pro/src/DataGridPro/useDataGridProProps.ts
```diff
@@ -27,12 +27,14 @@ export const DATA_GRID_PRO_PROPS_DEFAULT_VALUES: DataGridProPropsWithDefaultValu
...
   getDetailPanelHeight: () => 500,
+  disableRowGrouping: false,
+  rowGroupingColumnMode: 'single',
 };
 
 const rowGroupingWarning = buildWarning(
   [
     'MUI: The row grouping has been moved to the new `@mui/x-data-grid-premium` package.',
-    'Your application should continue to work with `@mui/x-data-grid-pro` but all the row-grouping related features will not be present.',
+    'It will soon be removed from `@mui/x-data-grid-pro`.',
```
joserodolfofreitas, line 37:
```suggestion
    'The feature is deprecated in `@mui/x-data-grid-pro` and it will soon be removed from this package.',
```

packages/grid/x-data-grid-premium/src/typeOverloads/index.ts
```diff
@@ -1 +1 @@
-export * from './modules';
+import '@mui/x-data-grid-pro/typeOverloads';
```
DanailH, line 1: Just to be clear we went with duplicating the row grouping required files over to the `x-data-grid-pro` package, right? I'm asking because I don't see them being removed from the `x-data-grid-premium`. That's ok, but can we add somewhere, maybe a TODO explaining what needs to happen (i guess as simple as reverting this exact commit) once the transition period is over?

flaviendelangle: Yes I went for duplication to avoid spending to much time on a temporary solution
I am open to discussion of course.

m4theushw (approved):
I ran some demos from https://mui.com/x/react-data-grid/row-grouping/ with the Pro package generated here and it worked. I won't spend too much time checking everything. The feature was always experimental, so developers should pay attention to the release notes before updating.

Issue #4653, opened 2022-04-25
[DateTimePicker] Allow to pass props to the `TrapFocus` with `TrapFocusProps`


Dodie324:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Currently using the following versions:
```
  "@mui/lab": "^5.0.0-alpha.53",
  "@mui/material": "^5.0.1",
 ```
 I have a form rendered inside of `Dialog` that consists of a couple `TextField` inputs (a mix of 'text' and 'select' types) and a `DesktopDateTimePicker` that's rendered inside of a `LocalizationProvider`.
 
If a user opens the date time picker, navigates between different months using the left and right arrows, and then clicks into a select input field while the date time picker is open a `RangeError: Maximum call stack size exceeded.` error from `Unstable_TrapFocus.js` gets thrown.
 
While this doesn't seem to break the UI functionality it is causing an unhandled error in our Sentry logs and kicks-off a PagerDuty alert.

Seems like it's relevant to https://github.com/mui/material-ui-pickers/issues/1852
 
 

### Expected behavior 🤔

I'd expect a call stack error not to exist.

### Steps to reproduce 🕹

Steps:

1. Open `DesktopDateTimePicker` from within `Dialog`
2. Navigate between months using left and right arrows
3. While date time picker is open click into a select input field
4. Notice the error in the console

### Context 🔦

<img width="462" alt="Screen Shot 2022-04-25 at 16 03 00" src="https://user-images.githubusercontent.com/9405172/165165864-e031606e-bb26-47d3-93d1-f87331d671d2.png">

### Your environment 🌎

```
    "@emotion/react": "^11.4.1",
    "@emotion/styled": "^11.3.0",
    "@mui/lab": "^5.0.0-alpha.53",
    "@mui/material": "^5.0.1",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
```

Tested in Chrome Version 100.0.4896.127 (Official Build) (arm64)

Dodie324:
@m4theushw I hope this example is useful: https://codesandbox.io/s/data-grid-community-forked-8inrq5?file=/src/App.tsx

Steps to reproduce in example:
1. Open up the console once the page loads
2. Toggle through the different months using the left and right arrow in the date picker
3. Click into the select field and the out of the select field
4. Notice the error in the console: `Uncaught RangeError: Maximum call stack size exceeded.`

[...]

PR #4850, opened 2022-05-11
[DateTimePicker] Infinite loop in TrapFocus

### LINKED ISSUE #4595: [DateTimePicker] Uncaught RangeError when Date Time Picker focus changes to a newly opened Dialog
### Duplicates

### Latest version

### Current behavior 😯

"Uncaught RangeError: Maximum call stack size exceeded. " is thrown in the chrome devtools console when TimePicker focus changes to the newly opened Dialog.

### Expected behavior 🤔

I would expect no "Uncaught RangeError"

### Steps to reproduce 🕹

codesandbox: https://codesandbox.io/s/laughing-tharp-k47gd

Steps:

in Chrome...

1. Open the datetimepicker to the timepicker view
2. Click the "click me" button next to the datetimepicker field
3. Notice in the browser devtools console that a "RangeError" is thrown. (You'll have to open the devtools, the codesandbox console doesn't show this error for some reason)

### Context 🔦

_No response_

### Your environment 🌎

Chrome

flaviendelangle:
Fixes #4595

[Behavior before](https://tbmhyd.csb.app/)
[Behavior after](https://3jikig.csb.app/)

To test you need to open the picker, select a date to go to the hour view, then click on "Click me".
You should then have dozens of log in the console.

___

Working with `TrapFocus` is always the far-west.
From what I understood, when closing the picker while on clock view and opening a `Dialog` (with it own `TrapFocus`) at the same time, then we ended up in a situation where 2 `TrapFocus` were fighting for the focus.

The problem being that in `TrapFocus`, the `useEffect` tracking `isEnabled` and `open` is not fired after the infinite loop.

My fix is just to use `isEnabled` with a ref to have the latest version of `open` and break this infinite chain.

___

I guess in the future we could ask users to use `React.useEvent` for `isEnabled` rather than relying on a `useEffect` update.
(not sure if it will work, it depends on the implementation on React I guess)

@michaldudak if you have some insights.

### DIFF
--- a/packages/x-date-pickers/src/internals/components/PickersPopper.tsx
+++ b/packages/x-date-pickers/src/internals/components/PickersPopper.tsx
@@ -285,6 +285,10 @@ export const PickersPopper = (props: PickerPopperProps) => {
     ...otherPaperProps
   } = PaperProps;
 
+  const openRef = React.useRef(true);
+  openRef.current = open;
+  const isTrapFocusEnabled = React.useCallback(() => openRef.current, []);
+
   return (
     <PickersPopperRoot
       transition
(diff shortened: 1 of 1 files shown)


m4theushw:
Would make sense to let the user configure this? I proposed a different solution in https://github.com/mui/mui-x/issues/4653

flaviendelangle:
I closed #4595 to keep a single conversation
I will have a look at your proposal :+1:

LukasTy:
Correct me if I'm wrong, but testing the `behaviour after` sill does produce call stack errors 🤔 

https://user-images.githubusercontent.com/4941090/188112527-fa8cc921-0e42-49d4-a673-cde282cc0b20.mov

PR #4864, opened 2022-05-13
Update ruRU.ts (Quick filter toolbar field)

arvkonstantin:
add "Quick filter toolbar field" translation


flaviendelangle (approved):

PR #4867, opened 2022-05-13
[TimePicker] Fix time validation when current date is null

### LINKED ISSUE #4866: [TimePicker] Min/max is not respected if it doesn't have a default value
### Duplicates

### Latest version

### Current behavior 😯

the time range min/max isn't respected for a  Time component that has the value = null;
it's only after first selected a time that the time range is respected.

https://stackblitz.com/edit/react-ix5es1?file=demo.tsx

### Expected behavior 🤔

should always respect the min/max

### Steps to reproduce 🕹

Steps:

1. https://stackblitz.com/edit/react-ix5es1?file=demo.tsx
2. click to add time
3. you will see you it displays all the time options and not limited to the min/max

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

flaviendelangle:
Fixes #4866

When we migrated the project, there was 0 validation on `ClockPicker` if the current time was null.
In #4726 I applied the value validity check (is hours "5" valid).
And in this PR I apply the full date validity (is 24th of May 2022 at 5PM valid)

## What's next ?

- Refacto `ClockPicker.test.tsx` to group the validity tests

### DIFF
--- a/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
+++ b/packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
@@ -258,8 +258,8 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
 
   const now = useNow<TDate>();
   const utils = useUtils<TDate>();
-  const midnight = utils.setSeconds(utils.setMinutes(utils.setHours(now, 0), 0), 0);
-  const dateOrMidnight = date || midnight;
+
+  const dateOrMidnight = date || utils.setSeconds(utils.setMinutes(utils.setHours(now, 0), 0), 0);
 
   const { meridiemMode, handleMeridiemChange } = useMeridiemMode<TDate>(
     dateOrMidnight,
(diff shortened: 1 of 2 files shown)


packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -258,8 +258,11 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
 
   const now = useNow<TDate>();
   const utils = useUtils<TDate>();
-  const midnight = utils.setSeconds(utils.setMinutes(utils.setHours(now, 0), 0), 0);
-  const dateOrMidnight = date || midnight;
+
+  const dateOrMidnight = React.useMemo(
```
flaviendelangle, line 262: Without this `useMemo`, the `useCallback` and `useMemo` below are useless when `date` is null

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -298,36 +301,23 @@ export const ClockPicker = React.forwardRef(function ClockPicker<TDate extends u
       switch (viewType) {
         case 'hours': {
           const value = convertValueToMeridiem(rawValue, meridiemMode, ampm);
-
-          if (date == null) {
-            return !isValidValue(value);
-          }
-
-          const dateWithNewHours = utils.setHours(date, value);
+          const dateWithNewHours = utils.setHours(dateOrMidnight, value);
```
flaviendelangle, line 304: `dateOrMidnight` is the variable used on the `onChange` callbacks.
So using the same value should keep the validation coherent.

Even if `disableIgnoringDatePartForTimeValidation` is set to `true` (i.e: when used in `DateTimePicker` and there is a `minDateTime` or `maxDateTime` provided).

alexfauquette (approved):
I was looking for a way to handle minutes/seconds because tests looked strange. 

But I don't think we have a better way expect a doc page on validation

flaviendelangle:
> I was looking for a way to handle minutes/seconds because tests looked strange.

What was looking strange in the tests ?

alexfauquette:
That selecting minutes to 20 is impossible when `minValue=12:15` and `maxValue=15:45`

flaviendelangle:
Indeed, maybe we will find a better behavior later

PR #4871, opened 2022-05-13
[pickers] Allow to reset the input when the previous date was invalid

### LINKED ISSUE #4870: [pickers] Impossible to clear an invalid date

flaviendelangle:
Related to #4486
Fixes #4870

@alexfauquette I don't see why we are comparing `currentInputValue` with `previousInputValueRef.current` (which is essentially the previous version of `currentInputValue`.

This variables holds the stringified version of the date. So for an invalid date it is an empty string, like for a null date.
But when we type by hand an invalid date, there is a de-synchronization between `currentInputValue` (which has an empty string) and `innerInputValue` which has the invalid date (`33/33/3333` for instance).

Is there a scenario where we don't have the input focus AND when the `props.date` changes we don't want to update the input with the stringified version ?

It breaks all the tests where we don't focus the input, but in a real usage, it will be focused.

____

If you feel like it is to tightly linked to #4486, I can remove the behavior change and just add a `.skip` to the new test (your PR would have to remove the `.skip`).

### DIFF
--- a/packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
+++ b/packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
@@ -13,6 +13,7 @@ import {
   withPickerControls,
   openPicker,
 } from '../../../../test/utils/pickers-utils';
+import { DatePickerProps } from '@mui/x-date-pickers';
 
 const WrappedDesktopDatePicker = withPickerControls(DesktopDatePicker)({
   DialogProps: { TransitionComponent: FakeTransitionComponent },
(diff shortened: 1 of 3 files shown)


alexfauquette:
I reached the same point as you in my PR

You can see that tests are failing, and the reason is the next one: the component has a default value which is `01/01/2019` but it does not handle `onChange` by default:

```jsx
function TestKeyboardDatePicker(
  PickerProps: MakeOptional<DesktopDatePickerProps<any, any>, 'value' | 'onChange' | 'renderInput'>,
) {
  const [value, setValue] = React.useState<unknown>(
    PickerProps.value ?? adapterToUse.date('2019-01-01T00:00:00.000'),
  );

  return (
    <DesktopDatePicker
      value={value}
      onChange={(newDate) => setValue(newDate)}
      renderInput={(props) => <TextField placeholder="10/10/2018" {...props} />}
      {...PickerProps}
    />
  );
}
```

When you are typing, you cause a rerender because internal state is updated due to `handleChange` but there is no way to know if it is because of a modification of `rawValue` or `innerInputValue`. The ref was here to remove the ambiguity

```js
const valueChanged = previousInputValueRef.current !== currentInputValue;
```

If `valueChanged===true` the developed is controlling the component, so we trust the dev (as long as she provides valid date). Otherwise, we keep what is written in the input

flaviendelangle:
Closing this one then

PR #4891, opened 2022-05-15
[docs] Avoid confusion with license key installation

oliviertassinari:
My initial trigger to work on this PR is that the license we share in https://mui.com/x/advanced-components/#license-key-installation is a working one, proof:

Screenshot 2022-05-15 at 22 34 15

https://master--toolpad.mui.com/_toolpad/app/cl1c2j0l512939zo6a43575en/editor/pages/cl2kfa28r00013f69amqv1prv

Now considering #1364, I think that it makes it too easy to cheat and more importantly will lead to many customers to install **the wrong key**. I think that the change I propose here will make things clear. Note that I have used a key that can be quicky double-clicked selected for replacement:

https://user-images.githubusercontent.com/3165635/168492813-72522ed8-68df-47be-b704-043a0fe45231.mp4

e.g. [not like done here](https://www.ag-grid.com/react-data-grid/licensing/#via-grid-packages-2)…

*This PR is a part of a larger effort to polish the angles of the infrastructure to support selling MUI X, post Premium plan launch.*

### DIFF
--- a/docs/data/advanced-components/overview.md
+++ b/docs/data/advanced-components/overview.md
@@ -92,9 +92,7 @@ This key removes all watermarks and console warnings.
 ```jsx
 import { LicenseInfo } from '@mui/x-license-pro';
 
-LicenseInfo.setLicenseKey(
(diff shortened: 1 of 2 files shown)


flaviendelangle (commented):
Agree

flaviendelangle (approved):

Issue #4851, opened 2022-05-12
[RFC] Internal row structure for partial filtering / sorting and lazy-loading


flaviendelangle:
This is a draft to start discussing about this subject and prepare for v6

## Introduction

We currently store the rows in a flat structure (in `state.rows.ids`) and then create a tree from it if needs be.
Any update to the rows is at first an update to the flat row list followed by a full re-generation of the tree
This has several implications:

- When adding / removing / modifying a row, the grid always triggers a re-generation of the whole tree

- When adding / removing / modifying a row, the grid always re-applies the sorting, filtering and aggregation on all the rows, even those in groups not impacted at all by the modification

- When toggling a group expansion, the grid always re-applies the filtering on all the rows

- In the tree, the group are represented by there parent row. But for the top level rows, there is no notion of "top level group" since the top level rows do not have a parent. In #4208, I had to invent a fake row ID to store aggregated data of the top level rows.

- In the tree, the group are represented by there parent row. It is therefore impossible to do lazy loading with *Row Grouping* without having at least one row of each group on the initial fetch.

- The auto generated rows (grouping rows for *Tree Data* or *Row Grouping* and aggregation footer) are regular rows in the state. They can be updated / removed by `apiRef.current.updateRows` which makes no sense.

**Note**: It is unclear to me exactly how far we can optimize to avoid re-computing sorting / filtering / aggregation on non-impacted groups. But I am sure that we can do non-negligible optimizations and that we need a better internal state structure for it. 

In summary, our current row state is not suited for large datasets, especially when it comes to Row Grouping because any change in the dataset triggers computation on the whole dataset, even when it is a localized update.

## Proposal

[AG-Grid paragraph](https://www.ag-grid.com/javascript-data-grid/data-update-transactions/#localised-changes-in-grouped-data)

In the presentation below, *row* means "real row". The auto generated row are no longer stored as regular row in the state. They can't be updated / removed using the same API as the regular row.

We do have a slight nomenclature issue because the word *row* can have two meanings.

1. An element in the data given by the user (does not include the groups)
2. A row in the HTML table (does include the group)

This can create confusion, but AG-Grid also calls the data given by the user "rows" so it's probably not that problematic.

### New state structure

```ts
/**
 * Equivalent of the current GridRowTreeNodeConfig for real rows
 */
interface GridRowConfig {
  /**
   * The id of the row
   */
  id: GridRowId;

  /**
   * The id of the group in which this row id
   */
  groupId: string;

  /**
[...]

PR #4896, opened 2022-05-16
[core] New pipe processing: row hydration

flaviendelangle:
Extracted from #4208

## Goal

Allow plugin to create custom rows.
The main use case is to create footers for the aggregation.
This logic will probably be impacted by #4851


m4theushw (approved):

packages/grid/x-data-grid/src/hooks/features/rows/gridRowsState.ts
```diff
@@ -51,4 +51,8 @@ export interface GridRowsState extends GridRowTreeCreationValue {
    * It does not count the expanded children rows.
    */
   totalTopLevelRowCount: number;
+
```
m4theushw, line 54: Description missing.

```suggestion
```

PR #4902, opened 2022-05-16
[core] Add new script to generate tree data rows from file tree

flaviendelangle:
I am working on the "Aggregation + Tree Data" doc.
The current dataset we use for the Tree Data is not well suited for aggregation.
I think it would be nice to have example like AG-Grid with a file explorer.

I created a small script to turn a file tree into a list of tree data rows that can be used for Tree Data examples.
Here are a few example from the core repo

- [create-react-app template](https://codesandbox.io/s/data-grid-pro-forked-ky7swv?file=/src/App.tsx)
- [tailwind-css template](https://codesandbox.io/s/data-grid-pro-forked-jls38k?file=/src/App.tsx)
- [material-next package](https://codesandbox.io/s/data-grid-pro-forked-qdli8h?file=/src/App.tsx)

For now I only store the hierarchy, the size and the modification date.
And of course there is no aggregation for now

With the Github API we could even do a real file explorer of our repo which would give a very good example of the Tree Data


scripts/treeDataFromFileTree.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+yargs
+  .command({
+    command: '$0',
+    describe: 'Generate tree data rows from a folder.',
+    builder: (command) =>
+      command.option('path', {
+        describe: 'Root of the hierarchy to generate',
+        type: 'string',
```
m4theushw, line 71: It could have `process.cwd()` as default value. Initially it didn't understand if path meant the input folder or path to write the output file (if this was supported).

scripts/treeDataFromFileTree.ts
```diff
@@ -0,0 +1,78 @@
...
+  };
+};
+
+const getSubTree = async (nodePath: string, parentHierarchy: string[] = []) => {
+  const nodes: Node[] = [];
+  const stats = await fs.lstat(nodePath);
+
+  if (stats.isDirectory()) {
+    const children = await fs.readdir(nodePath);
+    const directory = path.basename(nodePath);
```
m4theushw, line 27: We need to ignore `node_modules` and `build`. If I run `yarn dataset:file-tree --path .` it passes through these folders.

scripts/treeDataFromFileTree.ts
flaviendelangle: Our whole repo is too big to be logged in the console
But if I store it in a file it works well

https://codesandbox.io/s/data-grid-pro-forked-x1gygw?file=/src/App.tsx

scripts/treeDataFromFileTree.ts
```diff
@@ -0,0 +1,86 @@
...
+  for (let i = 0; i < children.length; i += 1) {
+    const childPath = path.join(argv.path, children[i]);
+
+    // eslint-disable-next-line no-await-in-loop
+    const childNodes = await getSubTree(childPath);
+    nodes.push(...childNodes);
+  }
+
+  return fs.writeFile(
+    './treeDataFromFileTreeResponse.txt',
```
m4theushw, line 65: It won't write to the root when executed from inside `scripts`.

```suggestion
    path.join(__dirname, '../treeDataFromFileTreeResponse.json'),
```

m4theushw (approved):

PR #4908, opened 2022-05-16
[core] Fix React 18 peer dependency

oliviertassinari:
Screenshot 2022-05-16 at 22 34 04

### DIFF
--- a/packages/grid/x-data-grid-premium/package.json
+++ b/packages/grid/x-data-grid-premium/package.json
@@ -55,7 +55,7 @@
     "@mui/material": "^5.2.8",
     "@mui/system": "^5.2.8",
     "exceljs": "^4.3.0",
-    "react": "^17.0.2"
+    "react": "^17.0.2 || ^18.0.0"
   },
   "peerDependenciesMeta": {
     "exceljs": {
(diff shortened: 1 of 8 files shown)


packages/grid/x-data-grid-premium/package.json
```diff
@@ -55,7 +55,7 @@
     "@mui/material": "^5.2.8",
     "@mui/system": "^5.2.8",
     "exceljs": "^4.3.0",
-    "react": "^17.0.2"
+    "react": "^17.0.2 || ^18.0.0"
```
oliviertassinari, line 58: What I wanted to change

m4theushw (approved):

packages/x-date-pickers/package.json
```diff
@@ -64,8 +64,7 @@
     "dayjs": "^1.10.7",
     "luxon": "^1.28.0 || ^2.0.0",
     "moment": "^2.29.1",
-    "react": "^17.0.2 || ^18.0.0",
-    "react-dom": "^17.0.2 || ^18.0.0"
```
oliviertassinari, line 68: We never import from "react-dom" so I assume we don't need it.

PR #4933, opened 2022-05-19
[core] Simplify rows cache management

flaviendelangle:
While working on #4927, I noticed that our cache structure had some complex legacy structure that was not useful anymore.

- [x] Rename `convertRowsPropToState` into `createRowsInternalCache`
- [x] Don't have branch behavior in `createRowsInternalCache` that was basically "do nothing", only call it when we do want to regenerate the cache
- [x] Move all the utils in `gridRowUtils`
- [x] Make `apiRef.current.unstable_caches.rows` non nullable, like for the state we can rely on the fact that the stat init is being done before the rest
- [x] Flatten `GridRowsInternalCache` instead of having a `value` key storing almost everything
- [x] Clean the naming, no variable related to the cache should be named `state` (otherwise we can have misunderstandings between what leaves in the state and in the cache)

### DIFF
--- a/packages/grid/x-data-grid/src/hooks/features/rows/gridRowsUtils.ts
+++ b/packages/grid/x-data-grid/src/hooks/features/rows/gridRowsUtils.ts
@@ -1,4 +1,14 @@
-import { GridRowId, GridRowModel, GridRowTreeConfig } from '../../../models';
+import * as React from 'react';
+import {
+  GridRowId,
+  GridRowIdGetter,
+  GridRowModel,
+  GridRowsProp,
+  GridRowTreeConfig,
+} from '../../../models';
+import { DataGridProcessedProps } from '../../../models/props/DataGridProps';
+import { GridApiCommunity } from '../../../models/api/gridApiCommunity';
+import { GridRowsInternalCache, GridRowsState } from './gridRowsState';
 
 /**
  * A helper function to check if the id provided is valid.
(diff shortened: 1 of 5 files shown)


alexfauquette (approved):
Looks good to me but I never had to use the cache

packages/grid/x-data-grid/src/hooks/features/rows/gridRowsUtils.ts
```diff
@@ -23,6 +33,81 @@ export function checkGridRowIdIsValid(
...
+export const getRowsStateFromCache = ({
+  apiRef,
+  previousTree,
+  rowCountProp,
+  loadingProp,
+}: {
+  apiRef: React.MutableRefObject<GridApiCommunity>;
+  previousTree: GridRowTreeConfig | null;
+  rowCountProp: number | undefined;
+  loadingProp: boolean | undefined;
```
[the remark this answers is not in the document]
m4theushw: I only added `undefined` to snooze the TypeScript error, but a default value can be added.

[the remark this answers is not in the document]
flaviendelangle: I will add a default value to `DATA_GRID_PROPS_DEFAULT_VALUES` in another PR then :+1:

packages/grid/x-data-grid/src/hooks/features/rows/gridRowsUtils.ts
```diff
@@ -23,6 +33,82 @@ export function checkGridRowIdIsValid(
...
+};
+
+export const createRowsInternalCache = ({
+  rows,
+  getRowId,
+  loadingProp,
+}: {
+  getRowId: DataGridProcessedProps['getRowId'];
+  loadingProp: DataGridProcessedProps['loading'];
+  rows: GridRowsProp;
```
m4theushw, line 53: What do you think of making explicit that they are props? `Pick` could also be used.

```suggestion
  getRowIdProp: DataGridProcessedProps['getRowId'];
  loadingProp: DataGridProcessedProps['loading'];
  rowsProp: DataGridProcessedProps['rows'];
```

m4theushw (approved):
Much better with the values flattened.

packages/grid/x-data-grid/src/hooks/features/rows/gridRowsUtils.ts
flaviendelangle: `rows` can come from `setRows` so it would not be 100% accurate
But maybe that's not a real issue though

flaviendelangle: Done

PR #4936, opened 2022-05-19
[docs] Fix TS typo in ColumnSpanningFunction example

lindapaiste:
Change TypeScript type of `subtotal` from number literal `624` to type `number`.


cherniavskii (approved):
Thanks!

PR #4945, opened 2022-05-20
[core] Remove dead code on row grouping tree creation

flaviendelangle:
Noticed it while working on #4927 
This lookup is never used.


alexfauquette (approved):
Yes, that sounds like an helper that has been abandoned and not deleted

Issue #4925, opened 2022-05-18
[DataGrid] GridRow throws Hooks exception when I try to edit a custom cell.


PedroLucasSilva:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When I click on an editable cell that has a custom component I get an error in the console about the rendering order of the hooks.

![image](https://user-images.githubusercontent.com/23297731/169042613-43ee88ae-fd6a-4272-bdaa-ae1d7fdce18b.png)

### Expected behavior 🤔

I should be able to edit without any errors being thrown to the console.

### Steps to reproduce 🕹

Steps:

1. Open this example https://codesandbox.io/s/data-grid-community-forked-xmherr?file=/src/App.tsx
2. Try editing a cell from the first column
3. Check the console for errors

### Context 🔦

In our project, we have an input a little more complex than this, made to format dates as the user writes. After switching to mui 5 and the latest version of DataGrid we have faced this problem.

flaviendelangle:
Hi,

I can't reproduce the crash when editing a cell of the 1st column in your example linked above.
Are you sure this example is complete ?

PedroLucasSilva:
Hi @flaviendelangle! Did you double click on the cell and look at the console? I asked other people on my team to test it and the result was the same as mine.

![image](https://user-images.githubusercontent.com/23297731/169105429-258a1d84-71c1-4270-8bea-07cf06aac3fb.png)

flaviendelangle:
I was not looking at the console sorry...

But I agree with the fix proposer by Matheus.

PR #4972, opened 2022-05-22
Variable change on "Create your own edit component"

ansarisufiyan777:
CustomEditComponent was having props as a parameter, but in scope you are using params.

### DIFF
--- a/docs/data/data-grid/editing/editing.md
+++ b/docs/data/data-grid/editing/editing.md
@@ -296,7 +296,7 @@ To customize column types, or override the existing components, you can provide 
 This property works like the `renderCell` property, which is rendered while cells are in view mode.
 
 ```tsx
-function CustomEditComponent(props: GridRenderEditCellParams) {
+function CustomEditComponent(params: GridRenderEditCellParams) {
   return <input type="text" value={params.value} onValueChange={...}>;
 }
 


flaviendelangle (commented):
Thanks for your contribution
In a React component I would probably do the opposite and name it `props`

m4theushw:
We usually use `props` for the first argument of a functional component. To not lose this contribution you could fix the typo in https://github.com/mui/mui-x/issues/4925#issuecomment-1130958009 Side note: in this example `params` is correct because it's a render function.

flaviendelangle:
> Side note: in this example params is correct because it's a render function.

If it's a render functiion, shouldn't it be named `renderCustomEdit` or something similar ?

Having `const XXXComponent = (params) => {}` feels weird to me.

flaviendelangle:
@m4theushw I'm not sure we are talking about the same function

I am talking about the one modified on this PR, which is currently not anonymous

https://github.com/mui/mui-x/pull/4972/files#diff-f4cb3f2893a5afd3c53698f703491e039bb334ad40056c968000eb6498b86cdbR299

m4theushw:
@flaviendelangle For me, this is a component and should be named as `XXXComponent`. I have the following in mind:

Render function: `const renderSomething = () => {}` (it's anonymous in the docs)
Functional component: `const SomethingComponent = () => {}`

flaviendelangle:
Then if it's a component, do you agree that it's argument should be named `props` and not `params` ?

m4theushw:
Yes, it should be `props` in `CustomEditComponent`. I said that in https://github.com/mui/mui-x/pull/4972#issuecomment-1134593692. My comment may have looked confusing since I used "this example", but I was referring to the example in the docs where a render function is used. If the change here is reverted, this PR will be closed so I suggested to fix the typo with `(`.

m4theushw:
I'm closing since we have not received feedback about our suggestions.

PR #4974, opened 2022-05-23
[l10n] Add deDE translations for x-date-pickers

felixh10r:
Following up on https://github.com/mui/mui-x/issues/3211, I'd like to provide German translations for the Date Picker component.


flaviendelangle (approved):

PR #4982, opened 2022-05-23
[DataGrid] Add generics to `GridActionsColDef` to match `GridColDef`

### LINKED ISSUE #4660: [data grid] getActions generic not working
### Duplicates

### Latest version

### Current behavior 😯

`getActions` is not working with generic like `valueGetter`

### Expected behavior 🤔

It should provide correct type

### Steps to reproduce 🕹

```
import {
  GridColumns
} from '@mui/x-data-grid'

interface User {
  name: string;
  age: number
}

const columns: GridColumns<User> = [
  {
    field: "name",
    valueGetter: (params) => params.row.name,
  }, 
  {
    field: "actions",
    type: 'actions',
    getActions: (params) => params.row.age && []
  },
]
```

Bildschirmfoto 2022-04-26 um 12 52 00
Bildschirmfoto 2022-04-26 um 12 52 10

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

subvertallchris:
Closes #4660 

This adds generic arguments to the `GridActionsColDef` interface. They match those provided to `GridValidRowModel` and feed into it in the definition as well as in the params given to the `getActions` callback.

The immediate benefit here is that it provides an added layer of type safety to `getActions` and anywhere else in the base `GridColDef` that those arguments are used. It should be fully backwards compatible with all existing uses of this interface since the generics are optional and default to `any`. 

I added a simple test that puts the compiler through some very basic paces and also tests a `getActions` callback that makes use of those arguments. I see there is another test that also checks the `getActions` callback params, so maybe this second portion isn't necessary, but I figured I might as well demonstrate the connection between the interface and the implementation.

### DIFF
--- a/packages/grid/x-data-grid/src/models/colDef/gridColDef.ts
+++ b/packages/grid/x-data-grid/src/models/colDef/gridColDef.ts
@@ -249,7 +249,8 @@ export interface GridColDef<R extends GridValidRowModel = any, V = any, F = V> {
   colSpan?: number | ((params: GridCellParams<V, R, F>) => number | undefined);
 }
 
-export interface GridActionsColDef extends GridColDef {
+export interface GridActionsColDef<R extends GridValidRowModel = any, V = any, F = V>
+  extends GridColDef<R, V, F> {
   /**
    * Type allows to merge this object with a default definition [[GridColDef]].
    * @default 'actions'
@@ -260,12 +261,12 @@ export interface GridActionsColDef extends GridColDef {
    * @param {GridRowParams} params The params for each row.
    * @returns {React.ReactElement<GridActionsCellItemProps>[]} An array of [[GridActionsCell]] elements.
    */
-  getActions: (params: GridRowParams) => React.ReactElement<GridActionsCellItemProps>[];
+  getActions: (params: GridRowParams<R>) => React.ReactElement<GridActionsCellItemProps>[];
 }
 
 export type GridEnrichedColDef<R extends GridValidRowModel = any, V = any, F = V> =
   | GridColDef<R, V, F>
-  | GridActionsColDef;
+  | GridActionsColDef<R, V, F>;
 
 export type GridColumns<R extends GridValidRowModel = any> = GridEnrichedColDef<R>[];
 
--- a/packages/grid/x-data-grid/src/tests/columns.spec.tsx
+++ b/packages/grid/x-data-grid/src/tests/columns.spec.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { DataGrid, GridRenderCellParams } from '@mui/x-data-grid';
 import { GridCellParams } from '../models/params/gridCellParams';
+import { GridActionsColDef, GridColDef, GridColumns, GridRowParams } from '../models';
 
 const RenderCellParamsExplicitTyping = () => {
   return (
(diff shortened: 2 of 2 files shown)


packages/grid/x-data-grid/src/models/colDef/gridColDef.ts
```diff
@@ -260,7 +261,7 @@ export interface GridActionsColDef extends GridColDef {
    * @param {GridRowParams} params The params for each row.
    * @returns {React.ReactElement<GridActionsCellItemProps>[]} An array of [[GridActionsCell]] elements.
    */
-  getActions: (params: GridRowParams) => React.ReactElement<GridActionsCellItemProps>[];
+  getActions: (params: GridRowParams<R>) => React.ReactElement<GridActionsCellItemProps>[];
 }
 
 export type GridEnrichedColDef<R extends GridValidRowModel = any, V = any, F = V> =
```
m4theushw, line 267: You also need to pass the generics to `GridActionsColDef` to make the typings work below:

```ts
const columns: GridColumns<TestRowModel> = [
  {
    field: 'actions',
    type: 'actions',
    getActions: ({ id, row }) => {},
  },
];
```

The diff in https://github.com/mui/mui-x/issues/4660#issuecomment-1109929700 shows how to do.

packages/grid/x-data-grid/src/tests/rows.DataGrid.test.tsx
```diff
@@ -145,6 +146,77 @@ describe('<DataGrid /> - Rows', () => {
...
+                icon={<span />}
+              />,
+            ];
+          },
+        };
+      }, []);
+
+      return (
+        <div>
+          <div data-testid="id">{id || 'none'}</div>
```
m4theushw, line 190: We usually don't use DOM elements to test if a function received a specific argument. Instead, we use spy functions.

Example:

https://github.com/mui/mui-x/blob/23a57c3665ff00675c976cc89f8aaeee54cefe69/packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.new.test.tsx#L83-L89

Could you use this same approach here? It reduces the boilerplate.

m4theushw (commented):
Thanks for the contribution. I linked #4660 in the description.

> I added a simple test that puts the compiler through some very basic paces and also tests a getActions callback that makes use of those arguments.

Usually we put these kind of tests in https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/tests/columns.spec.tsx

We type the generic so it throws a TS error if we use an invalid prop. There're some examples in this file. If you could add one there it would be awesome.

packages/grid/x-data-grid/src/tests/rows.DataGrid.test.tsx
subvertallchris: I think I can go ahead and pull the tests out of this file since the behavior is already tested and the compiler tests are covered elsewhere.

subvertallchris:
> Usually we put these kind of tests in https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/tests/columns.spec.tsx

Sounds good! I looked for something like this but didn't find it, it's how I'm used to testing type defs for definitely-typed. I'll make the changes. I hadn't come across `GridEnrichedColDef` before. That behavior is really cool, I'm glad I did this exercise.

packages/grid/x-data-grid/src/tests/columns.spec.tsx
```diff
@@ -40,6 +41,29 @@ const RenderCellParamsExplicitTyping = () => {
             return params.formattedValue.toUpperCase();
           },
         },
+        {
+          field: 'price6',
+          type: 'actions',
+          getActions: (params: GridRowParams) => {
+            return params.row.foo.bar.toExponential();
+          },
+        },
```
m4theushw, line 50: I don't think we need this test. The `row` in `GridRowParams` already has `any` so it accepts everything. Instead, you could create a different function using `GridColumns<{ ... }>`, then check if `getActions` receives the same row. Something like:


```ts
const columns: GridColumns<{ price: number }> = [{
  getActions: (params) => {
    return params.row.price.toUpperCase(); // fails
  }
}];
```

m4theushw (approved):
Thanks for going the extra mile.

PR #4987, opened 2022-05-24
[data grid] Should not call setRowIndex when dragging a column over a row

flaviendelangle:
Currently if `props.rowReorder = true`, when dragging a column over a row, we call `setRowIndex` with `rowId === ''`
This does not cause any bug, but `setRowIndex` is doing an useless lookup over the rows to find the index of the non existing row we are trying to move.


DanailH (approved):
Nice catch!

PR #4994, opened 2022-05-24
[test] Skip Safari and Firefox on broken tests

alexfauquette:
From https://github.com/mui/mui-x/pull/4645#issuecomment-1132887003

Skip tests on Firefox and Safari because they fail without apparent reason

### DIFF
--- a/packages/x-date-pickers/src/ClockPicker/ClockPicker.test.tsx
+++ b/packages/x-date-pickers/src/ClockPicker/ClockPicker.test.tsx
@@ -195,7 +195,8 @@ describe('<ClockPicker />', () => {
   });
 
   it('should display options, but not update value when readOnly prop is passed', function test() {
-    if (/safari/i.test(window.navigator.userAgent) || /firefox/i.test(window.navigator.userAgent)) {
+    // Only run in supported browsers
+    if (typeof Touch === 'undefined') {
       this.skip();
     }
     const selectEvent = {
@@ -228,7 +229,8 @@ describe('<ClockPicker />', () => {
   });
 
   it('should display disabled options when disabled prop is passed', function test() {
-    if (/safari/i.test(window.navigator.userAgent) || /firefox/i.test(window.navigator.userAgent)) {
+    // Only run in supported browsers
+    if (typeof Touch === 'undefined') {
       this.skip();
     }
     const selectEvent = {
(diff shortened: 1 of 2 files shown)


packages/x-date-pickers/src/StaticTimePicker/StaticTimePicker.test.tsx
```diff
@@ -46,7 +46,10 @@ describe('<StaticTimePicker />', () => {
     }),
   );
 
-  it('should allows view modification, but not update value when `readOnly` prop is passed', () => {
+  it('should allows view modification, but not update value when `readOnly` prop is passed', function test() {
+    if (/safari/i.test(window.navigator.userAgent) || /firefox/i.test(window.navigator.userAgent)) {
+      this.skip();
```
oliviertassinari, line 51: For each test, could we leave a comment on what's missing to make it work? I believe we have done this systematically in the past. 

For instance:
```suggestion
      // Only run in supported browsers
      if (typeof Touch === 'undefined') {
        this.skip();
      }
```

packages/x-date-pickers/src/ClockPicker/ClockPicker.test.tsx
```diff
@@ -194,7 +194,10 @@ describe('<ClockPicker />', () => {
     expect(Math.max(...hours)).to.equal(23);
   });
 
-  it('should display options, but not update value when readOnly prop is passed', () => {
+  it('should display options, but not update value when readOnly prop is passed', function test() {
+    if (/safari/i.test(window.navigator.userAgent) || /firefox/i.test(window.navigator.userAgent)) {
+      this.skip();
+    }
```
oliviertassinari, line 200: For each test, could we leave a comment on what's missing to make it work? I believe we have done this systematically in the past. 

For instance:
```suggestion
    // Only run in supported browsers
    if (typeof Touch === 'undefined') {
      this.skip();
    }
```

alexfauquette: 👍 Sounds easier to read

oliviertassinari: And more reliable, if tomorrow the API is available, then the test starts running.

Issue #409, opened 2020-10-07
[DataGrid] Server-side pagination for an unknown number of items


thomasdowell:
To enable server-side pagination for an unknown number of items in the TablePagination component, the value of the count prop must be -1. When the value of the DataGrid component's rowCount prop is set to -1, the server-side pagination functionality is not complete.

## Current Behavior 😯

"Total rows: -1" is rendered and the onPageChange function is not triggered.

## Expected Behavior 🤔

"Total rows: -1" should not render and the onPageChange function should trigger until there are no more pages to get from the server.

## Steps to Reproduce 🕹

Steps:

1. Go to [this CodeSandbox](https://codesandbox.io/s/1gjqg?file=/demo.js) which demos [this material-ui example](https://material-ui.com/components/data-grid/pagination/#server-side-pagination)
2. Navigate to demo.js
3. Change `rowCount={100}` to `rowCount={-1}` on line 55

## Context 🔦

Server-side pagination

## Benchmark

- https://www.ag-grid.com/javascript-grid/server-side-model-pagination/#server-pagination

oliviertassinari:
@thomasdowell Thanks for the feature request. The grid doesn't support the unknown last row case yet. This request is close to #404.

ahurlburt:
Are there any known workarounds to this? I'm using dynamo db and item count is either estimated (to get quickly) or expensive to get accurately. It would be nice if the data grid just allowed the next page to load and disable showing the total count in this case.

I see the infinite scroll for client side was marked important and closed, @oliviertassinari I see you marked that issue as important. Seeing dynamo's (and nosql in general) growth in popularity I would assume many people are going to start running into this issue (unless there is a valid workaround I'm not aware of)?

oliviertassinari:
> I'm using DynamoDB and item count is either estimated (to get quickly) or expensive to get accurately.

@ahurlburt I had the same experience with PostgreSQL after 1m rows on a table.

Regarding the possible solution of this problem, I think that there are two different paths that we could follow (complementary):

- Option 1. We make it work with the props only. Meaning, we can ask developers to set `rowCount={-1}` when the don't know how many rows they have, then use the number of rows in the `rows` prop (if server-side mode) and the size of the page to know where the displayed rows are, relative to the full data set.
[...]

PR #4998, opened 2022-05-24
[docs] Section for overwriting core components i18n keys

### LINKED ISSUE #4323: [question] Are there plans to implement a DataGridPro footer that doesn't require a finite rowCount
### Duplicates

### Latest version

### Summary 💡

Say I have no way of knowing the total row count that the server has, thus not being able to provide a definite rowCount, is there a way to display the total rows differently? Similar to a "see more" or have the ">" button clickable until no records can be fetched from the server?

### Examples 🌈

image

### Motivation 🔦

I have a use case where the exact total count of data/rows is not known--i.e., DynamoDB table without scanning the whole table.

### Order ID 💳 (optional)

_No response_

DanailH:
Part of #4323
Reference: https://github.com/mui/mui-x/issues/4323#issuecomment-1086167533

Preview: https://deploy-preview-4998--material-ui-x.netlify.app/x/react-data-grid/localization/#translation-keys

The goal is to add a section that explains that since the grid uses MUI Core components some translation keys need to be accessed through the MUI Core component.


docs/data/data-grid/localization/localization.md
```diff
@@ -17,6 +17,22 @@ In the following example, the labels of the density selector are customized.
 
 {{"demo": "CustomLocaleTextGrid.js", "bg": "inline"}}
 
+> ⚠️ It's important to note that because the data grid uses components from the Material UI Core library some translation keys need to be accessed using that component key.
```
oliviertassinari, line 20: I'm not aware we ever reference it this way

```suggestion
> ⚠️ It's important to note that because the data grid uses components from the Material UI library some translation keys need to be accessed using that component key.
```

oliviertassinari: Also, I think that we should use specialized callouts https://github.com/mui/material-ui/pull/32402. 

@siriwatknp did we ever announce the change to the other teams? I believe we didn't, maybe we should?

docs/data/data-grid/localization/localization.md
```diff
@@ -17,6 +17,22 @@ In the following example, the labels of the density selector are customized.
...
+>
+> One example is the `TablePagination` component used in the grid footer when pagination is enabled.
```
oliviertassinari, line 22: Would this help? It wasn't clear to me:

```suggestion
> One example is the table pagination component used in the grid footer when pagination is enabled. All the keys provided to the `MuiTablePagination` object are applied as props directly to the [`TablePagination`](/material-ui/api/table-pagination/) component.
```

oliviertassinari (commented):
> Fixes https://github.com/mui/mui-x/issues/4323

I thought #4323 was a duplicate of #409. What's the relationship between this PR and the issue?

docs/data/data-grid/localization/localization.md
flaviendelangle: https://github.com/mui/mui-x/pull/5008

docs/data/data-grid/localization/localization.md
DanailH: There are 2 problems raised in this issue. The first is the one you pointed out. The second is regarding how to provide your own string so you overwrite the OTB text we have. This PR is related to the second problem. The thing is that we don't explain how to overwrite texts in MUI Core components we use.
It's my mistake that I marked that this PR fixed the linked issue. Its related to what has been discussed there.

docs/data/data-grid/localization/localization.md
DanailH: So the correct one we should use it `:::info`?

DanailH: > I'm not aware we ever reference it this way

We never explicitly said it. I think the initial assumption was that people will use the theme to overwrite those texts but if you want to use the `localeText` only then this is the way.

m4theushw (approved):

PR #5001, opened 2022-05-24
[l10n] Add Norwegian (Bokmål) (nb-NO) locale

spiftire:


packages/grid/x-data-grid/src/locales/nbNO.ts
```diff
@@ -0,0 +1,144 @@
+import { GridLocaleText } from '@mui/x-data-grid';
```
oliviertassinari, line 1: Don't we need this one too?

```suggestion
import { nbNO as nbNOCore } from '@mui/material/locale';
import { GridLocaleText } from '@mui/x-data-grid';
```

flaviendelangle (approved):

flaviendelangle:

PR #5034, opened 2022-05-27
[docs] Fix demo currency format

oliviertassinari:
My locale is configured to be en-US, which is the locale [we use](https://www.notion.so/mui-org/Technical-writing-7e55b517ac2e489a9ddb6d0f6dd765de#9494c1a7521146c2ab571969b21955ce) to write the documentation. In this context, when opening https://mui.com/x/react-data-grid/row-grouping/, you will see:

Screenshot 2022-05-27 at 23 34 52

which I think is wrong:

```jsx
new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD', maximumFractionDigits: 0 }).format(2847246203);
// = $2,847,246,203
```

https://deploy-preview-5034--material-ui-x.netlify.app/x/react-data-grid/row-grouping/

### DIFF
--- a/docs/data/data-grid/column-definition/CustomColumnTypesGrid.js
+++ b/docs/data/data-grid/column-definition/CustomColumnTypesGrid.js
@@ -32,7 +32,7 @@ const currencyFormatter = new Intl.NumberFormat('en-US', {
 const usdPrice = {
   type: 'number',
   width: 130,
-  valueFormatter: ({ value }) => currencyFormatter.format(Number(value)),
+  valueFormatter: ({ value }) => currencyFormatter.format(value),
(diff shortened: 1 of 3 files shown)


oliviertassinari:
Actually, from a performance perspective, we might want to do this:

https://github.com/mui/mui-x/blob/77e7de44479d010e099ea60b7c80d76de1a48548/docs/data/data-grid/column-definition/CustomColumnTypesGrid.tsx#L27-L30

instead.

oliviertassinari:
cc @samuelsycamore here is one

samuelsycamore (approved):
Good catch @oliviertassinari!

PR #5050, opened 2022-05-30
[data grid] Use column visibility model on Hide All / Show All when enabled

### LINKED ISSUE #5047: [data grid] Columns disappear after "Hide All" and resize
### Duplicates

### Latest version

### Current behavior 😯

After clicking "Hide All"(in ToolbarColumns) and showing some columns without using the "Show All" button, columns disappear after resizing. It only happens when the columnVisibilityModel is chosen in initialState. Is it a bug or just a problem with my code?  

The problem comes after adding this:

image_2022-05-29_213609791

Example(https://stackblitz.com/edit/react-ts-5pfnmd?file=App.tsx): 

https://user-images.githubusercontent.com/53148689/170888472-479eb895-42f5-4300-997f-50411f3983bb.mp4

### Expected behavior 🤔

_No response_

### Steps to reproduce 🕹

Steps:

1. Open: https://stackblitz.com/edit/react-ts-5pfnmd?file=App.tsx.
2. Click “Columns”
3. Click “Hide All”
4. Click “name” checkbox to show that column
5. Try resize the “name” column

### Context 🔦

I ran into the problem when I tried to do a save and reset of the current state of the grid

### Your environment 🌎

  Opera GX
  "@mui/core": "^5.0.0-alpha.54",
   "@mui/icons-material": "^5.6.2",
   "@mui/material": "^5.5.1",
   "@mui/x-data-grid": "^5.6.1",
   "@mui/x-data-grid-pro": "^5.11.1",

### Order ID 💳 (optional)

240359

flaviendelangle:
Fixes #5047

- [x] Add test

@m4theushw it's a temporary fix before v6
The problem was that we were updating `colDef.hide` when clicking on "Hide All", even when using the column visibility model.
So the next time the column was updated (when resizing in #5047), it was re-syncinq the column (which had `hide: true`) and the visibility model.

My solution is to update directly the visibility model instead of calling `updateColumns` when we are using the visibility model.
It requires to expose the `isUsingColumnVisibilityModel` ref or to expose a new private method to have this logic internally in `useGridColumns`.
I think exposing the ref is better.

If you can confirm that the logic seems good to you, I'll add a test after.

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/columnsVisibility.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/columnsVisibility.DataGridPro.test.tsx
@@ -15,7 +15,7 @@ import {
   GridRowsProp,
   useGridApiRef,
 } from '@mui/x-data-grid-pro';
-import { getCell, getColumnHeaderCell, getColumnHeadersTextContent } from 'test/utils/helperFn';
+import { getColumnHeadersTextContent } from 'test/utils/helperFn';
 
 const isJSDOM = /jsdom/.test(window.navigator.userAgent);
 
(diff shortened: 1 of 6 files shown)


m4theushw (approved):
The logic is good. You can add the test and merge.

Issue #4890, opened 2022-05-15
[pickers] TimePicker ignoring mask when used AdapterDayjs on `LocalizationProvider`


josejimenez-swipejobs:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

I recently installed the latest version of `@mui/x-date-pickers`(5.0.0-alpha.3) and the latest of dayjs library `1.11.2`. But, when I try to use the component `TimePicker` with the ampm as `true`, it ignores the mask and allows any format on the input.

![Screen Shot 2022-05-16 at 5 28 01 am](https://user-images.githubusercontent.com/58408953/168490506-99d37a69-3e72-421a-b4f0-8b0a338c8d96.png)

I created this CodeSandbox to reproduce the issue https://codesandbox.io/s/input-format-adapter-dayjs-fot7g4?file=/src/App.js.

### Expected behavior 🤔

It will be great that the `TimePicker` works with the adapter: `AdapterDayjs` for the `LocalizationProvider`.

### Steps to reproduce 🕹

Steps:

1. Install the latest `@mui/x-date-pickers`,
2. Use the adapter `AdapterDayjs` on the LocalizationProvider Context,
3. Use the prop `ampm` as true in the component `TimePicker`.

### Context 🔦

Seems to be that this issue comes from here: https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/TimePicker/shared.ts#L81

This is setting the value from the context as `utils.formats.fullTime12h` for `ampm` true. This value comes from as "hh:mm A" rather than "hh:mm a".

I'm currently fixing this issue by forcing the `TimePicker` to use the prop `inputFormat` as "hh:mm a" when `ampm` is true.

```
<TimePicker
  ampm
  inputFormat="hh:mm a"
  value={value}
  onChange={setValue}
  renderInput={(params) => <TextField {...params} />}
/>
```

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
System:
    OS: macOS 11.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 38.00 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
    npm: 7.24.2 - ~/Develop/fe-core-elements/node_modules/.bin/npm
[...]

PR #1665, opened 2021-05-14
[DataGrid] Fix filter with object as value and value getter

dtassone:
fix #1615

PR #5055, opened 2022-05-31
[core] Allows to run tests with different date adapters

alexfauquette:
Allows to test bugs such as #4890 locally for now

to run unit test with a specific adapter, use `yarn test:unit  --date-adapter moment` (or `luxon`, `day-js`, `date-fns`)

### DIFF
--- a/packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.test.tsx
+++ b/packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.test.tsx
@@ -22,7 +22,7 @@ const WrappedDesktopDateTimePicker = withPickerControls(DesktopDateTimePicker)({
 describe('<DesktopDateTimePicker />', () => {
   const { render } = createPickerRenderer({
     clock: 'fake',
-    clockConfig: adapterToUse.date('2018-01-01T00:00:00.000').getTime(),
+    clockConfig: new Date('2018-01-01T00:00:00.000'),
   });
 
   ['readOnly', 'disabled'].forEach((prop) => {
--- a/test/utils/pickers-utils.tsx
+++ b/test/utils/pickers-utils.tsx
@@ -8,16 +8,20 @@ import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
 import { AdapterMoment } from '@mui/x-date-pickers/AdapterMoment';
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 
-// Add parameter `--date-adapter luxon` to use AdapterLuxon for running tests
-// adapter available : date-fns (default one), day-js, luxon, moment
-const argv = require('yargs/yargs')(process.argv.slice(2)).argv;
-
 const availableAdapters = ['date-fns', 'day-js', 'luxon', 'moment'];
+let adapter = 'date-fns';
 
-const adapter =
-  argv['date-adapter'] && availableAdapters.includes(argv['date-adapter'])
-    ? argv['date-adapter']
-    : 'date-fns';
+// Check if we are in unit tests
+if (typeof window === 'undefined') {
+  // Add parameter `--date-adapter luxon` to use AdapterLuxon for running tests
+  // adapter available : date-fns (default one), day-js, luxon, moment
+  const args = process.argv.slice(2);
+  const flagIndex = args.findIndex((element) => element === '--date-adapter');
+  const potentialAdapter = flagIndex + 1 < args.length ? args[flagIndex + 1] : null;
+  if (potentialAdapter && availableAdapters.includes(potentialAdapter)) {
+    adapter = potentialAdapter;
+  }
+}
 
 let AdapterClassToExtend;
 switch (adapter) {
--- a/test/utils/pickers-utils.tsx
+++ b/test/utils/pickers-utils.tsx
@@ -12,7 +12,7 @@ const availableAdapters = ['date-fns', 'day-js', 'luxon', 'moment'];
 let adapter = 'date-fns';
 
 // Check if we are in unit tests
-if (typeof window === 'undefined') {
+if (/jsdom/.test(window.navigator.userAgent)) {
   // Add parameter `--date-adapter luxon` to use AdapterLuxon for running tests
   // adapter available : date-fns (default one), day-js, luxon, moment
   const args = process.argv.slice(2);
--- a/test/utils/pickers-utils.tsx
+++ b/test/utils/pickers-utils.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { parseISO } from 'date-fns';
 import { createRenderer, screen, RenderOptions, userEvent } from '@mui/monorepo/test/utils';
 import { CreateRendererOptions } from '@mui/monorepo/test/utils/createRenderer';
 import { TransitionProps } from '@mui/material/transitions';
@@ -7,29 +6,9 @@ import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 
 // TODO make possible to pass here any utils using cli
-/**
- * Wrapper around `@date-io/date-fns` that resolves https://github.com/dmtrKovalenko/date-io/issues/479.
- * We're not using `adapter.date` in the implementation which means the implementation is safe.
- * But we do use it in tests where usage of ISO dates without timezone is problematic
- */
-export class AdapterClassToUse extends AdapterDateFns {
-  // Inlined AdapterDateFns#date which is not an instance method but instance property
-  // eslint-disable-next-line class-methods-use-this
-  date = (value?: any): Date => {
-    if (typeof value === 'string') {
-      return parseISO(value);
-    }
-    if (typeof value === 'undefined') {
-      return new Date();
-    }
-    if (value === null) {
-      // @ts-expect-error AdapterDateFns#date says it returns NotNullable but that's not true
-      return null;
-    }
-    return new Date(value);
-  };
-}
-export const adapterToUse = new AdapterClassToUse();
+export class AdapterClassToUse extends AdapterDateFns { }
+
+export const adapterToUse = new AdapterDateFns();
 
 export const FakeTransitionComponent = React.forwardRef<HTMLDivElement, TransitionProps>(
   function FakeTransitionComponent({ children }, ref) {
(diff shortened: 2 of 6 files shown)


packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.test.tsx
```diff
@@ -22,7 +22,7 @@ const WrappedDesktopDateTimePicker = withPickerControls(DesktopDateTimePicker)({
 describe('<DesktopDateTimePicker />', () => {
   const { render } = createPickerRenderer({
     clock: 'fake',
-    clockConfig: adapterToUse.date('2018-01-01T00:00:00.000').getTime(),
+    clockConfig: new Date('2018-01-01T00:00:00.000'),
```
alexfauquette, line 25: `getTime` is a date-fns method (note date-io), so this line was silently bugging when trying to run tests with another library

test/utils/pickers-utils.tsx
```diff
@@ -1,34 +1,48 @@
...
-      // @ts-expect-error AdapterDateFns#date says it returns NotNullable but that's not true
-      return null;
-    }
-    return new Date(value);
-  };
+const availableAdapters = ['date-fns', 'day-js', 'luxon', 'moment'];
+let adapter = 'date-fns';
+
+// Check if we are in unit tests
+if (typeof window === 'undefined') {
```
cherniavskii, line 15:
```suggestion
if (/jsdom/.test(window.navigator.userAgent)) {
```

Should we check for jsdom instead?
When I run `yarn test:unit` locally, `typeof window === 'undefined'` returns false

alexfauquette: That would be nice, but for another PR Could be nice to make test pass with other libraries on master 
And before that we should make unit test pass with the current code base which not the case

cherniavskii: Yeah, it can be addressed in separate PR

> And before that we should make unit test pass with the current code base which not the case

You mean to make unit tests pass with each of the adapters?

alexfauquette: Yes. For example some tests do not pass because some libraries use uppercase `AM/PM` and others use `am/pm`

flaviendelangle (approved):

alexfauquette:
Probably the `clockConfig: new Date('2018-01-01T00:00:00.000')` that provide different results in Safari

What about using `new Date(2018, 0, 1)`?

cherniavskii:
@alexfauquette let me check it in Safari

cherniavskii:
Ok, few observation from me:
1. I couldn't reproduce the issue locally in Safari 15 (in BrowserStack we use Safari 13). There's no way to install older version of Safari on MacOS, since Webkit is built into OS.
2. I've tried to run one test locally through BrowserStack Safari and log the start date that we pass to the picker:
	  ```diff
	  diff --git a/packages/x-date-pickers-pro/src/DesktopDateRangePicker/DesktopDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/DesktopDateRangePicker/DesktopDateRangePicker.test.tsx
	  index 0032283579..fcd8c96333 100644
	  --- a/packages/x-date-pickers-pro/src/DesktopDateRangePicker/DesktopDateRangePicker.test.tsx
	  +++ b/packages/x-date-pickers-pro/src/DesktopDateRangePicker/DesktopDateRangePicker.test.tsx
	  @@ -104,14 +104,14 @@ describe('<DesktopDateRangePicker />', () => {
	   
	     // TODO: Move to DayPicker test file ?
	     describe('selection behavior', () => {
	  -    it('should select the range from the next month', () => {
	  +    it.only('should select the range from the next month', () => {
	         const handleChange = spy();
	   
	  +      const startDate = adapterToUse.date('2019-01-01T00:00:00.000');
	  +      console.log('startDate', startDate);
	  +
	         render(
	  -        <WrappedDesktopDateRangePicker
	  -          onChange={handleChange}
	  -          initialValue={[adapterToUse.date('2019-01-01T00:00:00.000'), null]}
	  -        />,
	  +        <WrappedDesktopDateRangePicker onChange={handleChange} initialValue={[startDate, null]} />,
	         );
	  ```
	This resulted in the following output:
	```sh
	Safari 13.1.2 (Mac OS 10.15.6) WARN: 'including inaccessible elements by default'
	  LOG: 'startDate', Tue Jan 01 2019 11:00:00 GMT+1100 (AEDT)
	  Safari 13.1.2 (Mac OS 10.15.6) <DesktopDateRangePicker /> selection behavior should select the range from the next month FAILED
	          expected '2018-12-31T13:00:00.000Z' to equal '2019-01-01T00:00:00.000Z'
	          AssertionError
	  
	          assertEqual
	          methodWrapper
	          [native code]
	          toEqualDateTime
	          methodWrapper
	          [native code]
	  
	  Safari 13.1.2 (Mac OS 10.15.6): Executed 1 of 1438 (1 FAILED) (1.338 secs / 1.308 secs)
	```
	Which shows that the timezone in Safari is AEDT.
3. I've tried to change timezone on my machine to AEDT and run the same test through BrowserStack Safari, but it resulted in the same output as in 2 above.

I'm not sure why it's failing, but I tend to think it's related to the timezone somehow (maybe related to this? https://github.com/karma-runner/karma-browserstack-launcher/issues/187)

cherniavskii:
According to our browserslist we support Safari 14: https://github.com/mui/mui-x/blob/b5be28e1640cd7e44fea1ef89823be7b6b5dc56f/.browserslistrc#L29

Same with Supported platforms on MUI website: https://mui.com/material-ui/getting-started/supported-platforms/#browser

Screenshot 2022-06-09 at 13 51 52

So I think it's safe to update Safari version in BrowserStack.

cherniavskii:
I've run all test in Safari 14 and some Data Grid tests fail:

```
Safari 14.1.2 (Mac OS 10.15.7) WARN: 'including inaccessible elements by default'
................................................................................
................................................................................
................................................................................
................................................................................
.........................................
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call setEditCellValue with the value converted to Date FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call setEditCellValue with null when entered an empty value FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
.
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should handle correctly dates with partial years FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call onValueChange if defined FAILED
        expected +0 to equal 1
        AssertionError

        assertEqual
        methodWrapper
        [native code]

        asyncFunctionResume@[native code]
        [native code]
        promiseReactionJobWithoutPromise@[native code]
        promiseReactionJob@[native code]
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should call setEditCellValue with the value converted to Date FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should call setEditCellValue with null when entered an empty value FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
.
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should handle correctly dates with partial years FAILED
        null is not an object (evaluating 'spiedSetEditCellValue.lastCall.args')
....................
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call onEditCellPropsChange with the value entered as a Date FAILED
        undefined is not an object (evaluating 'onEditCellPropsChange.args[0][0]')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call onEditCellPropsChange with null when entered an empty value FAILED
        undefined is not an object (evaluating 'onEditCellPropsChange.args[0][0]')
..
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should handle all the intermediate dates while editing the value FAILED
        null is not an object (evaluating 'onEditCellPropsChange.lastCall.args')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: date should call preProcessEditCellProps once if it resolves with an error and preventCommitWhileValidating=true FAILED
        expected 1625407200000 to equal 1641906000000
        AssertionError

        assertEqual
        methodWrapper
        [native code]

        asyncFunctionResume@[native code]
        [native code]
        promiseReactionJobWithoutPromise@[native code]
        promiseReactionJob@[native code]
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should call onEditCellPropsChange with the value entered as a Date FAILED
        undefined is not an object (evaluating 'onEditCellPropsChange.args[0][0]')
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should call onEditCellPropsChange with null when entered an empty value FAILED
        undefined is not an object (evaluating 'onEditCellPropsChange.args[0][0]')
..
Safari 14.1.2 (Mac OS 10.15.7) <DataGridPro /> - Edit Components column type: dateTime should handle all the intermediate dates while editing the value FAILED
        null is not an object (evaluating 'onEditCellPropsChange.lastCall.args')
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
..........................................
Safari 14.1.2 (Mac OS 10.15.7): Executed 1403 of 1438 (14 FAILED) (skipped 35) (3 mins 30.967 secs / 3 mins 8.414 secs)
```

m4theushw:
We had some problems in the past with BrowserStack using a different locale on Edge: https://github.com/mui/mui-x/pull/1665#discussion_r633681113

It might be a similar reason here. The date passed to `adapterToUse.date` doesn't have the timezone. Does it assume GMT?

cherniavskii:
> We had some problems in the past with BrowserStack using a different locale on Edge: https://github.com/mui/mui-x/pull/1665#discussion_r633681113

> It might be a similar reason here

It looks like it was a Safari issue, since tests that failed in Safari 13 passed in Safari 14 (but other tests started to fail in Safari 14 😀)

PR #5060, opened 2022-05-31
[pickers] Infer mask from `inputFormat`

### LINKED ISSUE #4756: [pickers] Adapt masks to the locale

## Summary 💡

The Date Picker should provide the correct mask for the set locale.

## Motivation 🔦

The [material-ui docs](https://mui.com/components/date-picker/#localization) show how to create locale-specific masks. Yet, this still requires to map the browser locale to all kind of masks, and that's actually quite complex (see [Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country)).

Alternatively, we also considered using the MobileDataPicker (which doesn't need masks for entry). But this doesn't work because it uses masks to show the currently selected value. This would work if the Date Picker would allow to change the rendering of the selected value (e.g. to use the browser built-in `Date.prototype.toLocaleDateString()`).

### LINKED ISSUE #4890: [pickers] TimePicker ignoring mask when used AdapterDayjs on `LocalizationProvider`

alexfauquette:
Fix #4890 
Fix #4756

The idea is that is mask is not defined, we try to infer it from the `inputFormat` this allows to support at the same time libraries that use uppercase letters for `ampm` and library using lowercase

To run test, I based this PR on #5055

### Changelog

The mask can now be inferred from the `inputFormat`.
This allows supporting different separators like `dd/MM/YYYY`, `dd-MM-YYYY`, `dd.MM.YYYY` and different orders such as `MM-dd-YYYY`, `YYYY-MM-dd` without having to specify the `mask` prop
To be accepted, the input format needs to b made of digits (months  can not be `Jan`, `Fev`, ...) and with fix length (January must be `01` and not `1`).
You can still use `disableMaskedInput` to disable this feature and provide `mask` prop to override the inferred mask.

### DIFF
--- a/docs/data/date-pickers/date-picker/LocalizedDatePicker.tsx
+++ b/docs/data/date-pickers/date-picker/LocalizedDatePicker.tsx
@@ -17,15 +17,8 @@ const localeMap = {
   de: deLocale,
 };
 
-const maskMap = {
-  fr: '__/__/____',
-  en: '__/__/____',
-  ru: '__.__.____',
-  de: '__.__.____',
-};
-
 export default function LocalizedDatePicker() {
-  const [locale, setLocale] = React.useState<keyof typeof maskMap>('ru');
+  const [locale, setLocale] = React.useState<keyof typeof localeMap>('ru');
   const [value, setValue] = React.useState<Date | null>(new Date());
 
   const selectLocale = (newLocale: any) => {
@@ -50,7 +43,6 @@ export default function LocalizedDatePicker() {
           ))}
         </ToggleButtonGroup>
         <DatePicker
-          mask={maskMap[locale]}
           value={value}
           onChange={(newValue) => setValue(newValue)}
           renderInput={(params) => <TextField {...params} />}
--- a/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
+++ b/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
@@ -70,12 +70,7 @@ describe('text-field-helper', () => {
       if (isValid) {
         expect(runMaskValidation()).to.be.equal(true);
       } else {
-        expect(runMaskValidation).toWarnDev(
-          [
-            `The mask "${mask}" you passed is not valid for the format used ${format}.`,
-            `Falling down to uncontrolled no-mask input.`,
-          ].join('\n'),
-        );
+        expect(runMaskValidation).toWarnDev('Falling down to uncontrolled no-mask input.');
       }
     });
   });
--- a/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
+++ b/packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
@@ -26,18 +26,42 @@ describe('text-field-helper', () => {
   });
 
   [
-    { mask: '__.__.____', format: adapterToUse.formats.keyboardDate, isValid: false },
-    { mask: '__/__/____', format: adapterToUse.formats.keyboardDate, isValid: true },
-    { mask: '__:__ _m', format: adapterToUse.formats.fullTime, isValid: false },
-    { mask: '__/__/____ __:__ _m', format: adapterToUse.formats.keyboardDateTime, isValid: false },
-    { mask: '__/__/____ __:__', format: adapterToUse.formats.keyboardDateTime24h, isValid: true },
-    { mask: '__/__/____', format: 'MM/dd/yyyy', isValid: true },
-    { mask: '__/__/____', format: 'MMMM yyyy', isValid: false },
+    // Time picker
+    // - with ampm = true
+    { mask: '__:__ _m', format: adapterToUse.formats.fullTime12h, isValid: true },
+    // - with ampm=false
+    { mask: '__:__', format: adapterToUse.formats.fullTime24h, isValid: true },
+    // Date Picker
+    {
+      mask: '__/__/____',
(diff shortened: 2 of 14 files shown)


packages/x-date-pickers/src/internals/utils/text-field-helper.test.ts
```diff
@@ -46,12 +70,7 @@ describe('text-field-helper', () => {
       if (isValid) {
         expect(runMaskValidation()).to.be.equal(true);
       } else {
-        expect(runMaskValidation).toWarnDev(
-          [
-            `The mask "${mask}" you passed is not valid for the format used ${format}.`,
-            `Falling down to uncontrolled no-mask input.`,
-          ].join('\n'),
-        );
```
alexfauquette, line 54: I removed the `The mask ... you passed` because when masked is inferred, it does not make sens to show this error message to the dev

docs/data/date-pickers/date-picker/LocalizedDatePicker.tsx
```diff
@@ -17,15 +17,8 @@ const localeMap = {
   de: deLocale,
 };
 
-const maskMap = {
-  fr: '__/__/____',
-  en: '__/__/____',
-  ru: '__.__.____',
-  de: '__.__.____',
-};
```
alexfauquette, line 25: Since default mask is inferred, this can be done

packages/x-date-pickers/src/internals/utils/text-field-helper.ts
```diff
@@ -39,12 +39,53 @@ const MASK_USER_INPUT_SYMBOL = '_';
...
+  const inferredFormatPatternWith2Digits = utils
+    .formatByString(utils.date(staticDateWith2DigitTokens)!, format)
+    .replace(acceptRegex, '_');
+
+  if (inferredFormatPatternWith1Digits === inferredFormatPatternWith2Digits) {
+    return inferredFormatPatternWith1Digits;
+  }
+
+  console.warn(
+    [
```
flaviendelangle, line 70: Do we want to warn even in production ?

alexfauquette: of course not 🙈

flaviendelangle (commented):
I don't feel very confident approving this one because it's a very touchy part
But if the test pass for all adapters, I think we can give it a try

flaviendelangle (approved):

PR #5067, opened 2022-05-31
[core] Improve GitHub bug reproduction template

joserodolfofreitas:
Instead of adding templates for the different components, we'll suggest to fork a demo close to the user use case.


alexfauquette (approved):
Sounds more scalable with the number of components 👍

flaviendelangle (commented):
I would suggest to propose both options in the template.

The Codesandbox template is a link, people do not have to look for the codesandbox button on the doc (which can be non-trivial to find). So for a basic example, I think it's easier.

.github/ISSUE_TEMPLATE/1.bug.yml
```diff
@@ -34,17 +34,18 @@ body:
...
       label: Steps to reproduce 🕹
       description: |
-        Provide a link to a live example (you can use codesandbox.io) and an unambiguous set of steps to reproduce this bug.
-        Include code to reproduce, if relevant (which it most likely is).
-
-        You can use the official codesandbox template of the [DataGrid](https://codesandbox.io/s/datagrid-v5-quick-start-54iz1) or [DataGridPro](https://codesandbox.io/s/datagridpro-v5-quick-start-5pjhc) as a starting point.
+        Please provide a link to a live example and an unambiguous set of steps to reproduce this bug.
+        
+        You can use codesandbox.io to build a reproduction case.
+        As a starting point, we recommend you browse our [documentation](https://mui.com/x/advanced-components/), and fork the closest example to your use case.
```
oliviertassinari, line 40: Maybe we should explain how to "fork"?

flaviendelangle (commented):
Great !

flaviendelangle (approved):

.github/ISSUE_TEMPLATE/1.bug.yml
```diff
@@ -34,17 +34,19 @@ body:
     attributes:
       label: Steps to reproduce 🕹
       description: |
-        Provide a link to a live example (you can use codesandbox.io) and an unambiguous set of steps to reproduce this bug.
-        Include code to reproduce, if relevant (which it most likely is).
+        Please provide a link to a live example and an unambiguous set of steps to reproduce this bug.
+             
+        As a starting point, we recommend you browse our [documentation](https://mui.com/x/advanced-components/), and [select](https://mui.com/static/x/forking-an-example.png) the closest example to your use case.
```
oliviertassinari, line 39: Maybe we should make this a general rule? So first do the change in http://github.com/mui/material-ui, add t the screenshot and then come back here?

```suggestion
        As a starting point, we recommend you browse our [documentation](https://mui.com/x/advanced-components/), and [select](https://mui.com/static/core/forking-an-example.png) the closest example to your use case.
```

.github/ISSUE_TEMPLATE/1.bug.yml
```diff
@@ -34,17 +34,19 @@ body:
...
 
-        You can use the official codesandbox template of the [DataGrid](https://codesandbox.io/s/datagrid-v5-quick-start-54iz1) or [DataGridPro](https://codesandbox.io/s/datagridpro-v5-quick-start-5pjhc) as a starting point.
+        Or you can use a [basic template](https://codesandbox.io/s/mui-x-basic-template-uh3pm2?file=/src/App.tsx) to build a reproduction case.
```
oliviertassinari, line 41: Looking at this demo, I wonder if it shouldn't go in the opposite direction compared to the codesandbox edit buttons. The alternative would be to keep it an open pristine playground. Taking https://mui.com/r/issue-template and adding all the dependencies that developer might need, joy, system, base, data grid, date picker, tree view, etc.

oliviertassinari (commented):
Two thoughts:

.github/ISSUE_TEMPLATE/1.bug.yml
joserodolfofreitas: Yeah, I was wondering what would work better for the devs on this case. I considered it'd be faster to have a couple of commented lines to support a "ready-to-start" template with different components. On MUI X this is still viable but with MUI Core also in mind, it's definitely not a good start ^^.

oliviertassinari: >  I considered it'd be faster to have a couple of commented lines to support a "ready-to-start" template with different components

My flow of thought was: the codesandbox leads to the same outcome as the first proposed solution, so it's redundant so better remove it. But then, I was almost like, well, it's also a pain to have to open a codesandbox just to remove all the content on it, so I can copy & paste code from my own codebase.

.github/ISSUE_TEMPLATE/1.bug.yml
```diff
@@ -34,17 +34,17 @@ body:
     attributes:
       label: Steps to reproduce 🕹
       description: |
-        Provide a link to a live example (you can use codesandbox.io) and an unambiguous set of steps to reproduce this bug.
-        Include code to reproduce, if relevant (which it most likely is).
+        Please provide a link to a live example and an unambiguous set of steps to reproduce this bug.
+             
+        As a starting point, we recommend you browse our [documentation](https://mui.com/x/advanced-components/), and [select](https://mui.com/static/docs/forking-an-example.png) the closest example to your use case.
```
oliviertassinari, line 39: This link is a 301

```suggestion
        As a starting point, we recommend you browse our [documentation](https://mui.com/x/introduction/), and [select](https://mui.com/static/docs/forking-an-example.png) the closest example to your use case.
```

docs/public/_redirects
```diff
@@ -8,6 +8,7 @@
 /r/x-license-eula https://mui.com/store/legal/mui-x-eula/ 302
 /r/x-license-key-installation https://mui.com/x/advanced-components/#license-key-installation 302
 /r/x-data-grid-no-dimensions https://mui.com/x/react-data-grid/layout/ 302
+/r/x-issue-template https://codesandbox.io/s/mui-x-issue-template-uh3pm2 302
```
oliviertassinari, line 11: Please move the codesandbox under the MUI organization, so others can modify it later on.

Screenshot 2022-07-09 at 11 43 53

.github/ISSUE_TEMPLATE/1.bug.yml
```diff
@@ -34,17 +34,17 @@ body:
...
 
-        You can use the official codesandbox template of the [DataGrid](https://codesandbox.io/s/datagrid-v5-quick-start-54iz1) or [DataGridPro](https://codesandbox.io/s/datagridpro-v5-quick-start-5pjhc) as a starting point.
-
-        If you have an issue concerning TypeScript please start from this TypeScript playground: https://mui.com/r/ts-issue-template.
+        Or you can use a [basic template](https://mui.com/r/x-issue-template) to build your reproduction case.
```
oliviertassinari, line 41: Would it feel better without the blank line before "Or you can use"?

Screenshot 2022-07-09 at 11 45 54

docs/public/_redirects
oliviertassinari: For example, we likely want to fix:

Screenshot 2022-07-09 at 11 46 47

oliviertassinari (commented):
A few notes

.github/ISSUE_TEMPLATE/1.bug.yml
mbrookes: "the documentation". We document ourselves [here](https://mui.com/about/) 😁

```suggestion
        As a starting point, we recommend you [browse the documentation](https://mui.com/x/introduction/), and [select](https://mui.com/static/docs/forking-an-example.png) the closest example to your use case.
```

PR #5073, opened 2022-06-01
[DayPicker] Remove empty space at the bottom of the day view

### LINKED ISSUE #4819: [DayPicker] Excessive empty space at the bottom of the calendar
Hi everyone,

We noticed an empty space at the bottom of the Calendar, everywhere it is used, even when all 6 possible week rows are used.
image

It looks that it comes from the fact that [the minHeight is set here to 288px](https://github.com/mui/material-ui/blob/efa9620d0c4c6ee62076c9592dab5f4e448f7128/packages/mui-lab/src/DateRangePicker/DateRangePickerViewDesktop.tsx#L60
), overriding the 264px that would be enough.

Question: what is the reason of this empty space?

Thanks for your help.

flaviendelangle:
Fixes #4819

### DIFF
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
@@ -12,6 +12,7 @@ import {
   DayPicker,
   buildDeprecatedPropsWarning,
   DayPickerProps,
+  DAY_MARGIN,
 } from '@mui/x-date-pickers/internals';
 import { calculateRangePreview } from './date-range-manager';
 import { DateRange } from '../internal/models';
@@ -58,9 +59,13 @@ const DateRangePickerViewDesktopContainer = styled('div')(({ theme }) => ({
   },
 }));
 
+const DAY_RANGE_SIZE = 40;
+
+const weeksContainerHeight = (DAY_RANGE_SIZE + DAY_MARGIN * 2) * 6;
+
 const DateRangePickerViewDesktopCalendar = styled(DayPicker)({
   minWidth: 312,
-  minHeight: 288,
+  minHeight: weeksContainerHeight,
 }) as typeof DayPicker;
 
 const DateRangePickerViewDesktopArrowSwitcher = styled(PickersArrowSwitcher)({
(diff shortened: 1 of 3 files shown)


packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -62,7 +62,7 @@ export interface DayPickerProps<TDate> extends ExportedDayPickerProps<TDate> {
   TransitionProps?: Partial<SlideTransitionProps>;
 }
 
-const weeksContainerHeight = (DAY_SIZE + DAY_MARGIN * 4) * 6;
+const weeksContainerHeight = (DAY_SIZE + DAY_MARGIN * 2) * 6;
```
flaviendelangle, line 65: No idea why we were adding 4 time the margin, we only to add it one time for margin-top and one time for margin-bottom

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -58,9 +59,13 @@ const DateRangePickerViewDesktopContainer = styled('div')(({ theme }) => ({
   },
 }));
 
+const DAY_RANGE_SIZE = 40;
+
+const weeksContainerHeight = (DAY_RANGE_SIZE + DAY_MARGIN * 2) * 6;
```
flaviendelangle, line 64: Same formula as for the regular pickers but with a different day size

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
alexfauquette: It seems that margin top and margin bottom are merged, leading to only `2px` margin between weeks

Then the correct height should be

```suggestion
const weeksContainerHeight = (DAY_SIZE + DAY_MARGIN) * 6 + DAY_MARGIN;
```

Or the margin should be modified to put `4px` beween weeks and get a total of `40px` as explained in the [specification](https://material.io/components/date-pickers#specs)

flaviendelangle: The size of the day between range and non-range is weird
I think I'll clean that to have a better vision of the spacing.

eggplants:
Will this change be released in the next version?

flaviendelangle:
@eggplants yes, I wanted to do a better fix but I don't have the time, let's release this quick win and improve it later

sjdeak:
Hi @flaviendelangle  I found DateCalendar component also has this problem, so created a new issue for tracking: #10662 , please check it when you're convenient, thanks a lot!

PR #5074, opened 2022-06-01
[docs] Instruction to deal with v2 invalid license message

joserodolfofreitas:
* Add warning in the license documentation to instruct about the issue with v2-v1 licenses

https://deploy-preview-5074--material-ui-x.netlify.app/x/advanced-components/#license-key-installation

### DIFF
--- a/docs/data/advanced-components/overview.md
+++ b/docs/data/advanced-components/overview.md
@@ -97,7 +97,6 @@ To use the v2 licenses, you'll need to update the MUI X packages to the latest o
 If updating is not practicable or viable, please contact our support at sales@mui.com
 :::
 
-
 ### How and where to install the key?
 
 ```jsx
--- a/docs/data/advanced-components/overview.md
+++ b/docs/data/advanced-components/overview.md
@@ -86,16 +86,25 @@ You will need to purchase a commercial license in order to remove the watermarks
 
 ## License key installation
 
-When you purchase a commercial license, you'll receive a license key by email.
+When you purchase a commercial license, you'll receive your license key by email.
 This key removes all watermarks and console warnings.
+
+:::warning
+⚠️ If you purchased your license **after May 13, 2022**, You'll obtain a **v2** license key.
+
+To use the v2 licenses, you'll need to update the MUI X packages to the latest or at least the `5.11` version.
+
+If updating is not practicable or viable, please contact our support at sales@mui.com
+:::
+
+
+### How and where to install the key?
 
 ```jsx
 import { LicenseInfo } from '@mui/x-license-pro';
 
 LicenseInfo.setLicenseKey('YOUR_LICENSE_KEY');
 ```
-
-### Where to install the key?
 
 You must call `setLicenseKey` before React renders the first component.
 You only need to install the key once in your application.
@@ -143,7 +152,7 @@ But you will not be able to install a newer version released two years from now,
 
 #### Invalid license key
 
-This error indicates that your license key doesn't match what was issued by MUI—this is likely a typo.
+This error indicates that your license key doesn't match what was issued by MUI (this is likely a typo) or that you may be trying to validate a [v2 license key with an older package](#license-key-installation).
 
 ## Support
 
--- a/docs/data/advanced-components/overview.md
+++ b/docs/data/advanced-components/overview.md
@@ -90,7 +90,7 @@ When you purchase a commercial license, you'll receive your license key by email
 This key removes all watermarks and console warnings.
 
 :::warning
-⚠️ Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
+Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
 
 Please update your packages if you are using an earlier version.
 
--- a/docs/data/advanced-components/overview.md
+++ b/docs/data/advanced-components/overview.md
@@ -90,9 +90,9 @@ When you purchase a commercial license, you'll receive your license key by email
 This key removes all watermarks and console warnings.
 
 :::warning
-⚠️ All licenses purchased after **May 13, 2022** are _v2_ licenses.
+⚠️ Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
 
-The v2 license key is only compatible with MUI X version `5.11` or later.
+Please update your packages if you are using an earlier version.
 
 If this is not a viable solution for your use case, please contact sales@mui.com.
 :::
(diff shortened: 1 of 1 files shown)


docs/data/advanced-components/overview.md
```diff
@@ -86,17 +86,25 @@ You will need to purchase a commercial license in order to remove the watermarks
...
-When you purchase a commercial license, you'll receive a license key by email.
+When you purchase a commercial license, you'll receive your license key by email.
 This key removes all watermarks and console warnings.
 
+:::warning
+⚠️ If you purchased your license **after May 13, 2022**, You'll obtain a **v2** license key.
+
+To use the v2 licenses, you'll need to update the MUI X packages to the latest or at least the `5.11` version.
+
+If updating is not practicable or viable, please contact our support at sales@mui.com
```
samuelsycamore, line 97: Here's how I would phrase this:

```suggestion
All licenses purchased after **May 13, 2022** are _v2_ licenses.

The v2 license key is only compatible with MUI X v5.11 or later.
Please update your packages if you are using an earlier version.

If this is not a viable solution for your use case, please contact us at sales@mui.com.
```

oliviertassinari:
A question, if we do this change:

```diff
diff --git a/packages/x-license-pro/src/generateLicense/generateLicense.ts b/packages/x-license-pro/src/generateLicense/generateLicense.ts
index b1c4b243..8d4ab44c 100644
--- a/packages/x-license-pro/src/generateLicense/generateLicense.ts
+++ b/packages/x-license-pro/src/generateLicense/generateLicense.ts
@@ -23,7 +23,7 @@ function getClearLicenseString(details: LicenseDetails) {
     throw new Error('MUI: Invalid sales model');
   }

-  return `O=${details.orderNumber},E=${details.expiryDate.getTime()},S=${
+  return `O=${details.orderNumber},EXPERY=${details.expiryDate.getTime()},S=${
     details.scope ?? 'pro'
   },LM=${details.licensingModel ?? 'perpetual'},KV=${licenseVersion}`;
 }
```

Do we need the changes in this PR?

flaviendelangle:
If we do this change, the only problem would come with people using v2 licenses and v5.10 - v5.12.
Because the fix will not be applied to those versions.

joserodolfofreitas:
I think we can officially version our licenses and be transparent with it in our docs. Perhaps it makes sense to establish a section for it in this PR, even if we can fix it for 5.13.

On the technical side, It wasn't clear to me that the "E" vs "Expiry" was actually going to fix this. In this case we already took to long to do it. +1 to go for it, but I have a question:
Why the change on the `generateLicense` should be a few weeks later, aren't we doing this so old packages can also validate v2 licenses?

oliviertassinari:
> Why the change on the generateLicense should be a few weeks later, aren't we doing this so old packages can also validate v2 licenses?

@joserodolfofreitas The second diff makes the license key only compatible from v0 to v5.10.0. It's would be broken for developers that just installed MUI X. My proposal is to make the license key change at a point in time where we have at least 4 weeks of retro compatibility in the past to avoid the same problem that we had once we did the breaking license change. 

I think that to decide, we need to compare the scenarios based on the speed developers upgrade and the size of the market for each version https://www.npmjs.com/package/@mui/x-data-grid-pro

Screenshot 2022-06-01 at 18 49 19

docs/data/advanced-components/overview.md
```diff
@@ -90,9 +90,9 @@ When you purchase a commercial license, you'll receive your license key by email
 This key removes all watermarks and console warnings.
 
 :::warning
-⚠️ All licenses purchased after **May 13, 2022** are _v2_ licenses.
+⚠️ Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
```
flaviendelangle, line 93: We decided that the `:::warning` sections did not need the icon (I can find the discussion on the core if you want)

joserodolfofreitas:
```suggestion
Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
```

docs/data/advanced-components/overview.md
```diff
@@ -143,7 +153,21 @@ But you will not be able to install a newer version released two years from now,
 
 #### Invalid license key
 
-This error indicates that your license key doesn't match what was issued by MUI—this is likely a typo.
+This error indicates that your license key doesn't match what was issued by MUI (this is likely a typo).
+
+#### Invalid license key (Error: extracting license expiry timestamp)
+
+This is a more specific situation of the error above. If additionally to the invalid license message, you see an error on the console that looks like:
```
oliviertassinari, line 160: "If additionally" is strange, maybe?

```suggestion
The following JavaScript exception indicates that you may be trying to validate [the new license's key format on an older version of the npm package](#license-key-installation).
```

oliviertassinari (commented):
Look good 👍 , a few thoughts:,

docs/data/advanced-components/overview.md
```diff
@@ -143,7 +153,21 @@ But you will not be able to install a newer version released two years from now,
...
+
+:::error
+Error extracting license expiry timestamp.
+
+TypeError: Cannot read properties of null (reading '1') at verifyLicense.
+:::
+
+It indicates that you may be trying to validate [the new license's key format on an older package](#license-key-installation).
+
+In this case, please update to MUI X `5.11` or later.
```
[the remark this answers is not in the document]
flaviendelangle:
```diff
-to `v5.11.0` or later.
+to `v5.11.0` or above.
```

Not sure

docs/data/advanced-components/overview.md
```diff
@@ -86,9 +86,19 @@ You will need to purchase a commercial license in order to remove the watermarks
 
 ## License key installation
 
-When you purchase a commercial license, you'll receive a license key by email.
+When you purchase a commercial license, you'll receive your license key by email.
```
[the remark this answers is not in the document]
joserodolfofreitas: Alright. That makes sense

docs/data/advanced-components/overview.md
```diff
@@ -86,9 +86,19 @@ You will need to purchase a commercial license in order to remove the watermarks
...
 This key removes all watermarks and console warnings.
 
+:::warning
+Licenses purchased after **May 13, 2022** are only compatible with MUI X version `5.11` or later.
```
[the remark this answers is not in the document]
joserodolfofreitas:
```suggestion
Licenses purchased after **May 13, 2022** are only compatible with MUI X `v5.11.0` or later.
```

PR #5097, opened 2022-06-02
[pickers] Use new localization for doc examples

flaviendelangle:


docs/data/date-pickers/date-time-picker/CustomDateTimePicker.tsx
```diff
@@ -18,28 +18,34 @@ export default function CustomDateTimePicker() {
...
-          disableFuture
-          hideTabs
-          openTo="hours"
-          value={value}
-          onChange={(newValue) => {
-            setValue(newValue);
+        <LocalizationProvider
+          dateAdapter={AdapterDateFns}
+          localeText={{
+            openPreviousView: 'Open previous view',
```
flaviendelangle, line 24: This one was weird, we were setting a label related to months on a picker where we can never access a day view.

I don't even now why we do this example here, it should be using `TimePicker`.
But I don't want to do deeper changes on a l10n pull request.

docs/data/date-pickers/date-range-picker/BasicDateRangePicker.js
```diff
@@ -9,10 +9,11 @@ export default function BasicDateRangePicker() {
   const [value, setValue] = React.useState([null, null]);
 
   return (
-    <LocalizationProvider dateAdapter={AdapterDateFns}>
+    <LocalizationProvider
+      dateAdapter={AdapterDateFns}
+      localeText={{ start: 'Check-in', end: 'Check-out' }}
```
alexfauquette, line 14: I'm wondering if moving this `start` and `end` in the l10n was a good idea. This could be "Check-in" "Departure", "Arrival", "Begining", ... according to the context of the application

Maybe it is better to rename them `startLabel` since it is the label applied to the `TextField` for the start date

flaviendelangle: Maybe we could add a `localeText` key on the components as well and if they are defined, it wrap the component with a context (with a small HOC).

flaviendelangle: Is it used elsewhere ?
If not I indeed don't see the point of this prop in the 1st place

alexfauquette: My bad, it is used for the toolbar of the mobile date range picker when no value is selected

docs/data/date-pickers/date-time-picker/CustomDateTimePicker.tsx
alexfauquette: Why not just remove it? Those texts are the default ones. Modifying `leftArrowButtonText` in this doc example was helpful because the component was responsible for translations. 
Now we have a dedicated page to document how text are modified

flaviendelangle: Indeed
Removed

alexfauquette (approved):

PR #5101, opened 2022-06-03
[core] Upgrade monorepo

oliviertassinari:
Try https://github.com/mui/material-ui/pull/32939 out. 

An example of a problem this solves: search for "column definition" in Algolia: you will get a duplicate result:

Screenshot 2022-06-04 at 00 50 27

one with the leading slash, the other without, proof: https://www.algolia.com/apps/TZGZ85B9TB/explorer/browse/material-ui?query=column%20definition&searchMode=search 🙃.

### DIFF
--- a/docs/data/data-grid/filtering/filtering.md
+++ b/docs/data/data-grid/filtering/filtering.md
@@ -280,7 +280,7 @@ You can customize the rendering of the filter panel as shown in [the component s
 
 ### Customize the filter panel content
 
-The customization of the filter panel content can be performed by passing props to the default [`<GridFilterPanel />`](/x/api/data-grid/grid-filter-panel) component.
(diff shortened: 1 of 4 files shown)


oliviertassinari:
Niiice, it works:

```
12:42:18 PM: Error: Missing trailing slash. The following link:
12:42:18 PM: [<code>&lt;GridFilterPanel /&gt;</code>](/x/api/data-grid/grid-filter-panel) in /docs/data/data-grid/filtering/filtering.md is missing a trailing slash, please add it.
```

It's especially helpful for MUI X because of https://answers.netlify.com/t/no-trailing-slash-when-using-proxies/48790/18

flaviendelangle (approved):

PR #5104, opened 2022-06-03
[docs] Fix redundant headers

oliviertassinari:
I have found this from https://app.ahrefs.com/site-audit/3523498/13/data-explorer?columns=pageRating%2Curl%2Ctraffic%2Ctitle%2Cserp_title%2Cis_page_title_used_in_serp%2Ctop_keyword%2Ctop_keyword_position&filterCollapsed=true&filterId=207cf10d892e807e0aaeafdf8fc70712&issueId=d69246c2-225a-11ec-8456-06d2f2f613d8

Screenshot 2022-06-03 at 14 48 15

I don't see the logic with heaving MUI X to prefix.

### DIFF
--- a/docs/data/data-grid/getting-started/migration-v4.md
+++ b/docs/data/data-grid/getting-started/migration-v4.md
@@ -1,8 +1,4 @@
----
-title: MUI X - Migration from v4 to v5
----
-
-# MUI X - Migration from v4 to v5
+# Migration from v4 to v5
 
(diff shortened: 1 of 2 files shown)


oliviertassinari:
The issue was impacting the migration pages, which makes me think of a related matter: it would be much better to group the migration steps for each component under a single place as done in https://mui.com/material-ui/migration/migration-v4/.

PR #5121, opened 2022-06-05
[data grid] Fix implicit dependency on react-dom

oliviertassinari:
I was on https://github.com/mui/material-ui/pull/33015#issuecomment-1146631891, having a quick look at what could explain the jump in bundle size in the data grid at version v5.9.0. We went [from 78.1 to 122.1 KB](https://bundlephobia.com/package/@mui/x-data-grid@5.12.0). Compared to react-table [15.7 KB](https://bundlephobia.com/package/react-table@7.8.0) it can be intimidating. I could identify the origin down to #4332, we depend on `react-dom`:

https://github.com/mui/mui-x/blob/23aaa6bad1a0398ebc7677337d1d46226c2b70f6/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx#L2

but we don't list the dependency.

### DIFF
--- a/benchmark/browser/index.js
+++ b/benchmark/browser/index.js
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import ReactDOM from 'react-dom';
+import * as ReactDOM from 'react-dom';
 
 // Get all the scenarios
 const requirePerfScenarios = require.context('./scenarios', true, /(js|ts|tsx)$/);
--- a/test/regressions/index.js
+++ b/test/regressions/index.js
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import ReactDOM from 'react-dom';
+import * as ReactDOM from 'react-dom';
 import { BrowserRouter as Router, Routes, Route, Link } from 'react-router-dom';
 import { LicenseInfo } from '@mui/x-data-grid-pro';
 import TestViewer from 'test/regressions/TestViewer';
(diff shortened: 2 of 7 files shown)


benchmark/browser/index.js
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import ReactDOM from 'react-dom';
+import * as ReactDOM from 'react-dom';
```
oliviertassinari, line 2: A bonus refactoring, to match the docs: https://reactjs.org/docs/react-dom.html.

m4theushw (approved):

oliviertassinari:
It worked, it's more accurate now 😁

Screenshot 2022-06-13 at 14 41 59

https://bundlephobia.com/package/@mui/x-data-grid@5.12.0

test/regressions/index.js
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import ReactDOM from 'react-dom';
+import * as ReactDOM from 'react-dom';
```
oliviertassinari, line 2: @m4theushw When do we plan to use React 18? It would then be 

```suggestion
import * as ReactDOM from 'react-dom/client';
```

per https://reactjs.org/blog/2022/03/29/react-v18.html#new-client-and-server-rendering-apis

m4theushw: I'm only waiting for the core repo to adopt React 18 in the infra. But nothing stops us from adopting earlier. https://github.com/mui/mui-x/pull/4155 is open.

PR #5122, opened 2022-06-05
[l10n] Improve Japanese (ja-JP) locale

hikotq:
Adds missing translations for Japanese (ja-JP) locale - https://github.com/mui/mui-x/issues/3211.


flaviendelangle (approved):

PR #5123, opened 2022-06-05
[date picker] Fix `shouldDisableDate` in range pickers

### LINKED ISSUE #5111: [DateRangePicker] Disabled dates are still selectable
### Duplicates

### Latest version

### Current behavior 😯

1. Use the `shouldDisableDate` function or `minDate` & `maxDate` props to create disabled dates. 
2. Choose an intended disabled date as a start or end date.
3. Notice that the calendar shows the date with a normal appearance (not greyed out) and is selectable.
4. After selecting both a start and end date, notice that the textfield with the intended disabled date now has an error state. 

Note: this is the behavior in Alpha.5; Alpha.4 does not seem to have the same problem.

### Expected behavior 🤔

The calendar days should be greyed out and not selectable.

### Steps to reproduce 🕹

https://codesandbox.io/s/daterangepicker-alpha-5-react-17-0-2-d7fp9x

Choose one of the dates from `disableTheseDates`, or outside the bounds of the `minDate` & `maxDate`, as a start or end date. Then follow the steps in 'Current behavior.'

### Context 🔦

_No response_

### Your environment 🌎

Accessed the codesandbox above using Chrome: `101.0.4951.64 (Official Build) (arm64)` 

### Order ID 💳 (optional)

39843

flaviendelangle:
Fixes #5111

### DIFF
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
@@ -229,6 +229,7 @@ function DateRangePickerViewRaw<TInputDate, TDate>(
       disablePast,
       minDate,
       maxDate,
+      shouldDisableDate,
       ...calendarState,
       ...other,
     };
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewMobile.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewMobile.tsx
@@ -93,6 +93,8 @@ export function DateRangePickerViewMobile<TDate>(props: DesktopDateRangeCalendar
       />
       <DayPicker<TDate>
         {...other}
+        minDate={minDate}
+        maxDate={maxDate}
         disabled={disabled}
         readOnly={readOnly}
         selectedDays={parsedValue}


packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
```diff
@@ -229,6 +229,7 @@ function DateRangePickerViewRaw<TInputDate, TDate>(
       disablePast,
       minDate,
       maxDate,
+      shouldDisableDate,
```
flaviendelangle, line 232: This part of the code is very error-prone, maybe we could improve it later

alexfauquette: Do you have an idea about how to improve it? I only see two ways:

- Use providers to avoid having to propagate the information, and forget them somewhere
- make tests that loop on all the exported components to check we do not break props propagation

alexfauquette: Do you have an idea about how to improve it? I only see two ways:

- Use providers to avoid having to propagate the information, and forget them somewhere
- make tests that loop on all the exported components to check we do not break props propagation

alexfauquette (approved):

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
flaviendelangle: I did not think about it in depth tbh
We could indeed have a conformance test taking a component, 2-3 callbacks to interact with the DOM and which tests all the validation props (it would at least be a subset of our problem)

PR #5135, opened 2022-06-07
[DataGrid] Use exponential search to render non-measured rows

m4theushw:
Fixes https://mui.zendesk.com/agent/tickets/3610

Before: https://codesandbox.io/s/dynamicrowheightgrid-demo-mui-x-forked-7t4h90?file=/demo.tsx
After: https://codesandbox.io/s/dynamicrowheightgrid-demo-mui-x-forked-qfiwjv?file=/demo.tsx

The performance problem related in this issue is because we had a logic where the new rows rendered would always begin after the last measured row. This was added to prevent the scrollbar from changing the size when passing through groups of measured rows + non-measured row + measured rows. However, it didn't work as expected and generated a big performance bottleneck when pulling the scrollbar to a random position. This PR removes this logic and replaces it with a exponential search, like [react-virtualized](https://github.com/bvaughn/react-virtualized/blob/005be24a608add0344284053dae7633be86053b2/source/Grid/utils/CellSizeAndPositionManager.js#L291) does. As consequence, the scrollbar may jump now when pulling to a position not measured yet. This side effect can be reduced by tweaking `getEstimatedRowHeight`.

msedge_pOJ4GQTIFj


packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -104,6 +115,18 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
   const [containerWidth, setContainerWidth] = React.useState<number | null>(null);
   const prevTotalWidth = React.useRef(columnsTotalWidth);
 
+  const getNearestIndexToRender = React.useCallback(
```
flaviendelangle, line 118: I would be in favor of extracting those methods in an utility file and document them properly (what is offset ? Why we sometime use binary and someting exponential ?)

packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -116,13 +139,13 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
 
     const { top, left } = scrollPosition.current!;
 
-    const firstRowIndex = Math.min(
-      Math.max(0, apiRef.current.unstable_getLastMeasuredRowIndex()),
-      getIndexFromScroll(top, rowsMeta.positions),
-    );
+    // Clamp the value because the search may return an index out of bounds.
+    // In the last index, this is not needed because Array.slice doesn't include it.
+    const firstRowIndex = Math.min(getNearestIndexToRender(top), rowsMeta.positions.length - 1);
```
flaviendelangle, line 144: Couldn't `exponentialSearch` be responsible for returning a clamped value ?

flaviendelangle:
The trade-off performances vs small scrollbar jump make total sense to me

packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
m4theushw: I added some comments explaining. This method is a couple of lines of where it's used. IMO moving it to a separate file will cause more harm than good.

packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
m4theushw: Both search algorithms should return the same values when receiving the same arguments. Besides that we do can fix them to return the value inclusive, then do `+1` for the last index. However, I don't want to touch this logic now because it's critical. This `Math.min` is fixing a bug I discovered when rendering the grid with a very small height and no buffer: https://codesandbox.io/s/datagriddemo-demo-mui-x-forked-g9247r?file=/demo.tsx (scroll all the way to the bottom, no rows is rendered).

```diff
diff --git a/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx b/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
index b71c895f6..a7be89d11 100644
--- a/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
+++ b/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
@@ -27,14 +27,14 @@ export function binarySearch(
   offset: number,
   positions: number[],
   sliceStart = 0,
-  sliceEnd = positions.length,
+  sliceEnd = positions.length - 1,
 ): number {
   if (positions.length <= 0) {
     return -1;
   }
 
   if (sliceStart >= sliceEnd) {
-    return sliceStart;
+    return sliceStart - 1;
   }
 
   const pivot = sliceStart + Math.floor((sliceEnd - sliceStart) / 2);
```

seems to be sufficient to return the value inclusive. We always had this wrong `position.length`.

flaviendelangle (approved):

tr-joe:
**Overview**

When using the `getRowHeight={()=>'auto'}` property, the renderCell only evaluates height after an animation is complete. Consequently any components using a transition, such as Accordion or Collapse are affected. After a transition animation is finished, a height jump of the overall Table and current row is shown. This bug was introduced with this commit, but unfortunately it remains up to and including the latest release. I have attached a simple example to replicate this unresolved issue. After some investigation I believe I have isolated it to the `useGridRowsMeta` file and the following lines highlighted below. I am not 100% au fait on this as I am not a developer who has contributed, but it appears to be the `debounce` dependency. I would really appreciate a solution to this ASAP and Ideally a fix included in this version.

**Affected Code**

**Before - Smooth rendering**
```
    if (needsHydration) {
      hydrateRowsMeta();
    }
  }, [hydrateRowsMeta]);
```

**After - Jumpy rendering**
```
      if (needsHydration) {
      debouncedHydrateRowsMeta();
    }
  }, [debouncedHydrateRowsMeta]);
```

**Example code to replicate bug**

```
import { memo, useState } from "react";
import { Accordion, AccordionDetails, AccordionSummary, Box, Grid } from "@mui/material";
import { DataGrid, GridColDef, GridRenderCellParams } from "@mui/x-data-grid";

const ExpansionRow = memo(({ value }: GridRenderCellParams<string, any, string>) => {

    const [count, setCount] = useState<number>(0);

    return (
        <Box width={'100%'}>
            <Accordion onChange={() => setCount((prev) => prev + 1)} expanded={count % 2 !== 0}>

                <AccordionSummary>
                    {value}
                </AccordionSummary>

                <AccordionDetails>
                    <Grid
                        display={'flex'}
                        justifyContent={'center'}
                        alignItems={'center'}
                        sx={{ background: 'silver', height: 100 }}
                    >
                        EXPANSION DETAILS
                    </Grid>
                </AccordionDetails>

            </Accordion>
        </Box>
    );
});

const App = (): JSX.Element => {

        {
            field: 'name',
            headerName: 'Name',
            flex: 1,
            renderCell: (params: GridRenderCellParams<string, any, string>) =>
                <ExpansionRow {...params} />
        }
    ];

    const rows = [
        { id: 1, name: 'Jon', },
        { id: 2, name: 'Cersei', },
        { id: 3, name: 'Jaime', },
        { id: 4, name: 'Arya', },
        { id: 5, name: 'Daenerys', },
        { id: 6, name: 'Ferrara', },
        { id: 7, name: 'Rossini', },
        { id: 8, name: 'Harvey', },
    ];

    return (

        <Box width={'100%'} sx={{ background: 'gold' }}>
            <DataGrid
                rows={rows}
                columns={columns}
                autoHeight
                getRowHeight={() => 'auto'}
            />
        </Box>
    );
}

export default App;
```

PR #5142, opened 2022-06-08
[docs] Improve movie dataset

flaviendelangle:


flaviendelangle:
> What do you want to do with imbdRating data?

In the aggregation it's the example for aggregation on a cell with a custom renderer (the `Rating` component).
I'll probably add it to regular grouping demo later, as it creates more good-looking grids with some color.

Issue #4446, opened 2022-03-24
[pickers] `getViewSwitchingButtonText` prop does not work


jazhall:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

**Note:** There are different issues for the **DatePicker**, **DateRangePicker**, **DateTimePicker**, and **TimePicker** components. See notes below. Let me know if this issue should be split into different bug tickets for each bug occurrence.

In the **DatePicker** and **DateRangePicker** components, attempting to override the view switching button's aria-label via `getViewSwitchingButtonText` property does not currently change the aria-label at all.

In the **DateTimePicker** component, the view switch button does not show up at all, even though `showToolbar` is set to true. This results in `getViewSwitchingButtonText` not doing anything at al.

In the **TimePicker** component, `getViewSwitchingButtonText` isn't a supported prop despite the component having a view switch button when `showToolbar` is set to true.

### Expected behavior 🤔

**DatePicker** & **DateRangePicker**: Supplying a `getViewSwitchingButtonText` prop should override the aria-label in the view switching button.

**DateTimePicker**: The view switching button needs to show up so `getViewSwitchingButtonText` can be usable.

**TimePicker**: `getViewSwitchingButtonText` needs to be a supported prop, as this component has a view switching button when `showToolbar` is set to true.

### Steps to reproduce 🕹

Sample: https://codesandbox.io/embed/sharp-ramanujan-3topgp?fontsize=14&hidenavigation=1&theme=dark

Steps:

1. Open the DatePicker.
2. Use the browser console and inspect the view switching button.
3. Find the aria-label for the <button> element in the browser console.

### Context 🔦

I am adding these date/time pickers to my application that supports multiple languages. I'm using `useViewSwitchingButtonText` in each of these components to translate the aria-label of these buttons for each language.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.

Browser: Google Chrome Version 99.0.4844.82 (Official Build) (64-bit)
System:
    OS: Windows 10 10.0.19041
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD
  Browsers:
[...]

PR #5149, opened 2022-06-08
[pickers] Fix am/pm buttons position and responsiveness

### LINKED ISSUE #4947: [TimePicker] Action bar buttons overlap with PM button
### Duplicates

### Latest version

### Current behavior 😯

When using action bar buttons with AM / PM, the two are rendered on top of each other

https://codesandbox.io/s/date-and-time-pickers-forked-9z1n5c?file=/src/App.tsx

image

### Expected behavior 🤔

The actions should come below the AM / PM buttons

### LINKED ISSUE #7279: [DateTimePicker] Fully support Toolbar in landscape mode
### Summary 💡

Follow up of #6900

For consistency, the Toolbar of `DateTimePicker` has been added in landscape mode. However, it's support is still incomplete.

To add:
- Better width to let every date fit in it
- Add AM/PM selector when `ampmInClock` is `false`

alexfauquette:
Fix #4947
Fix #7279
Second part of #4446

Explanation commit by commit:

1. `ampmInClock` is used to know if buttons are in there clock or the toolbar (not the appbar). So I rewrite the `ampmInClock` behavior as follow:
	- If there is no toolbar, ampm buttons are in the clock whatever is the value of `ampmInClock`. Otherwise, end-users would not have access to ampm
	- If there is a toolbar, ampm buttons are either in the toolbar or the clock (not both) according to `ampmInClock`	
2. I forgetted to propagate `ampmInClock` to the `ToolbarComponent`
3. `MobileDateTimePicker` neither display the toolbar eithen if `showToolbar` is set to `true` fixing that was asked in #4446 
4. Add ampm control in the toolbar of MobileDateTimePicker
5. Inverse the order between components such that the view switcher is over the clock which is stealing the click

Before: [with code](https://codesandbox.io/s/responsivetimepickers-material-demo-forked-02fxr4?file=/demo.tsx) | [without code](https://02fxr4.csb.app/) 
After: [with code](https://codesandbox.io/s/responsivetimepickers-material-demo-forked-k405oi?file=/demo.tsx) | [without code](https://k405oi.csb.app/)

## Changelog

On desktop, `DateTimePicker` shows the ampm controls in the toolbar instead of the clock by default.
It can be overridden by specifying `ampmInClock` prop.


alexfauquette:
It still has one problem: the pen icon in the toolbar is at the same location as the view switcher of the clock
image

flaviendelangle:
image

On the `DateTimePicker` (both mobile and desktop) with both props true.

alexfauquette:
Strange because I do not manage to reproduce this issue which looks like an encoding one

image

packages/x-date-pickers/src/DateTimePicker/DateTimePickerToolbar.tsx
```diff
@@ -34,6 +36,8 @@ const useUtilityClasses = (ownerState: DateTimePickerToolbarProps<any> & { theme
     dateContainer: ['dateContainer'],
     timeContainer: ['timeContainer', theme.direction === 'rtl' && 'timeLabelReverse'],
     separator: ['separator'],
+    ampmSelection: ['ampmSelection'],
+    ampmLabel: ['ampmLabel'],
```
alexfauquette, line 40: This naming (like most of the code) is taken from `TimePickerToolbar`

alexfauquette:
To test all to configuration I made a demo:

https://codesandbox.io/s/nervous-robinson-05gnne?file=/demo.tsx

To compare with the same demo based on the latest version
https://codesandbox.io/s/gracious-edison-qs4mn8

packages/x-date-pickers/src/DateTimePicker/dateTimePickerToolbarClasses.ts
```diff
@@ -10,10 +10,18 @@ export interface DateTimePickerToolbarClasses {
...
   timeContainer: string;
+  /** Styles applied to the time (except am/pm) container element. */
+  timeDigitsContainer: string;
   /** Styles applied to the time container if rtl. */
   timeLabelReverse: string;
   /** Styles applied to the separator element. */
   separator: string;
+  /** Styles applied to the am/pm section. */
+  ampmSelection: string;
+  /** Styles applied to am/pm section in landscap mode. */
```
LukasTy, line 21:
```suggestion
  /** Styles applied to am/pm section in landscape mode. */
```

LukasTy, line 21:
```suggestion
  /** Styles applied to am/pm section in landscape mode. */
```

packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/CalendarOrClockPicker.tsx
```diff
@@ -72,6 +72,11 @@ export interface CalendarOrClockPickerProps<TDate, View extends DateOrTimeView>
     PickerStatePickerProps<TDate | null> {
   autoFocus?: boolean;
   DateInputProps: DateInputPropsLike;
+  /**
+   * Display ampm controls under the clock (instead of in the toolbar).
+   * @default false if toolbar is displayed
```
LukasTy, line 77:
```suggestion
   * @default `false` if toolbar is displayed
```

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
```diff
@@ -59,6 +60,11 @@ export interface UsePickerViewsBaseProps<
    * If `undefined`, internally defined view will be the used.
    */
   viewRenderers: PickerViewRendererLookup<TValue, TView, TExternalProps, TAdditionalProps>;
+  /**
+   * If `true`, the toolbar will be visible.
+   * @default `true` for mobile, `false` for desktop
+   */
+  showToolbar?: boolean;
```
LukasTy, line 67: Heads-up: This is either going to conflict with https://github.com/mui/mui-x/pull/7498

LukasTy (approved):
Great job! 🚀 
Such an annoying issue finally fixed. 💪 

I couldn't find any issues testing what I could think of. 😉

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
alexfauquette: An initial idea was to show view switcher only if:
- There is view to swich (more than 1 time view)
- The swich is not already available in the toolbar

Since `showToolbar` is not available any more, and using `slotProps.toolbar.hidden` does not seem reliable (as a developer it would be strange that a slot props impact another one) I propose to go to the safest way: show the view swich even if there is a toolbar. Leading to the following argos diff

@LukasTy does it seems good for you?

image

LukasTy: > Since `showToolbar` is not available any more, and using `slotProps.toolbar.hidden` does not seem reliable (as a developer it would be strange that a slot props impact another one)

Hm. Why do you think like that? `showToolbar` was being used to understand if this view switcher need to be shown, but the only thing that changed is the location this prop has moved to.
If we want/need the view switcher logic to depend on toolbar prop, I do not see much problem with it. 🤔 
Maybe adding a mention about this relationship somewhere would be useful, but I don't know the best place for it. 🙈

flaviendelangle: Note that this kind of scenario won't work well if we support the callback format at somepoint.
Because we would not have the `ownerState` outside of the layout component, so we could not correctly resolve the `toolbar.hidden` prop.

alexfauquette: Yes, I did not thaught about that edge case

LukasTy (commented):
Looking great! 👌 
Had another go after all the recent changes in API and couldn't find issues with it.
Looks like a great improvement and fixes the shortcoming of not having meridiem switcher in `DateTimePickerToolbar`. 🙌 

There are a few final issues to solve before we could merge this. 😉 

FYI. Been using [this sandbox](https://codesandbox.io/s/responsivetimepickers-material-demo-forked-29xxhz?file=/package.json) as well as locale changes.

packages/x-date-pickers/src/DateTimePicker/shared.tsx
```diff
@@ -74,6 +74,11 @@ export interface BaseDateTimePickerProps<TDate>
    * @default `utils.is12HourCycleInCurrentLocale()`
    */
   ampm?: boolean;
+  /**
+   * Display ampm controls under the clock (instead of in the toolbar).
+   * @default true
```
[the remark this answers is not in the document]
alexfauquette: I forget it was also fixing it

packages/x-date-pickers/src/DateTimePicker/shared.tsx
```diff
@@ -69,11 +69,9 @@ export interface BaseDateTimePickerProps<TDate>
   extends BasePickerInputProps<TDate | null, TDate, DateOrTimeView, DateTimeValidationError>,
     Omit<ExportedDateCalendarProps<TDate>, 'onViewChange'>,
     ExportedTimeClockProps<TDate> {
-  // TODO: Change default to be false on mobile, when `DateTimePickerToolbar` will support `ampm` buttons
-  // https://github.com/mui/mui-x/issues/7279
   /**
    * Display ampm controls under the clock (instead of in the toolbar).
-   * @default true
+   * @default true on desktop, false on mobile
```
LukasTy, line 74: Could you please apply the following diff to actually make this work [as mentioned here](https://github.com/mui/mui-x/issues/7279#issuecomment-1450084374)? 🤔 
Because currently, this place overrides defaults coming from `Static`, `Desktop`, and `Mobile` pickers. 🙈 

```diff
diff --git a/packages/x-date-pickers/src/DateTimePicker/shared.tsx b/packages/x-date-pickers/src/DateTimePicker/shared.tsx
index 8bc182198..68e0c1910 100644
--- a/packages/x-date-pickers/src/DateTimePicker/shared.tsx
+++ b/packages/x-date-pickers/src/DateTimePicker/shared.tsx
@@ -129,7 +129,6 @@ type UseDateTimePickerDefaultizedProps<
     | 'views'
     | 'openTo'
     | 'orientation'
-    | 'ampmInClock'
     | 'ampm'
     | keyof BaseDateValidationProps<TDate>
     | keyof BaseTimeValidationProps
@@ -163,7 +162,6 @@ export function useDateTimePickerDefaultizedProps<
     };
   }, [themeProps.localeText]);
 
-  const ampmInClock = themeProps.ampmInClock ?? true;
   const slots = themeProps.slots ?? uncapitalizeObjectKeys(themeProps.components);
   const slotProps = themeProps.slotProps ?? themeProps.componentsProps;
   return {
@@ -177,7 +175,6 @@ export function useDateTimePickerDefaultizedProps<
     ampm,
     localeText,
     orientation: themeProps.orientation ?? 'portrait',
-    ampmInClock,
     // TODO: Remove from public API
     disableIgnoringDatePartForTimeValidation:
       themeProps.disableIgnoringDatePartForTimeValidation ??
@@ -205,7 +202,6 @@ export function useDateTimePickerDefaultizedProps<
       ...slotProps,
       toolbar: {
         ampm,
-        ampmInClock,
         ...slotProps?.toolbar,
       },
     },
```

packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.types.ts
```diff
@@ -26,6 +26,11 @@ export interface DesktopDateTimePickerSlotsComponentsProps<TDate>
 export interface DesktopDateTimePickerProps<TDate>
   extends BaseDateTimePickerProps<TDate>,
     DesktopOnlyPickerProps<TDate> {
+  /**
+   * Display ampm controls under the clock (instead of in the toolbar).
+   * @default true
```
LukasTy, line 31: Are you sure it's worth going with these overrides trying to provide a more correct default? 🤔 
It's nice, I agree, but in that case, we'd need to fix an issue with `DateTimePicker`, maybe omit `ampmInClock` when creating `DateTimePickerProps`?
Or just go without overriding the default and show `true on desktop, false on mobile` 🙈 
What's your take on it @flaviendelangle?

LukasTy (commented):
Leaving a few comments to finalize a few points. 😉

packages/x-date-pickers/src/DesktopDateTimePicker/DesktopDateTimePicker.types.ts
alexfauquette: I did not realize it would also affect all the other component. I moved back to `true on desktop, false on mobile`  for every one to avoid side effects

PR #5156, opened 2022-06-09
[docs] Improve the getRowId doc section

flaviendelangle:
- Don't talk about "id field" to avoid confusion between the "id column" and the "id property in the row model".

- Create a standalone section instead of putting everything in a warning

- Add a warning saying that getRowId must be memoized

Should we add a note saying that the grid does not require an "id" column ? 
I think it's what the `Note that [column definition](https://mui.com/x/react-data-grid/column-definition) for id field is not required.` tries to say but it's not very comprehensible in its current form.

### DIFF
--- a/docs/data/data-grid/rows/rows.md
+++ b/docs/data/data-grid/rows/rows.md
@@ -17,17 +17,26 @@ Otherwise, the grid will re-apply heavy work like sorting and filtering.
 
 {{"demo": "RowsGrid.js", "bg": "inline"}}
 
-:::warning
-Each row object should have a field that uniquely identifies the row.
-By default, the grid will use the `id` property of the row. Note that [column definition](/x/react-data-grid/column-definition/) for `id` field is not required.
+## Row identifier
 
-When using dataset without a unique `id` property, you can use the `getRowId` prop to specify a custom id for each row.
+Each row must have a unique id.
+
+This id is used internally to identify the row in the various models (for instance the row selection model) and to track the row across updates.
+By default, the grid will use the `id` property of the row.
+
+When using a dataset without a unique `id` property, you can use the `getRowId` prop to specify the id of each row.
+This can be useful if your id is stored in another property like shown below:
 
 ```tsx
 <DataGrid getRowId={(row) => row.internalId} />
 ```
 
 {{"demo": "RowsGridWithGetRowId.js", "bg": "inline", "defaultCodeOpen": false}}
+
+:::warning
+Just like the `rows` prop, the `getRowId` prop should keep the same reference between two renders.
+Otherwise, the grid will re-apply heavy work like sorting and filtering.
+:::
 
 ## Updating rows
 


samuelsycamore:
> Should we add a note saying that the grid does not require an "id" column ? I think it's what the `Note that [column definition](https://mui.com/x/react-data-grid/column-definition) for id field is not required.` tries to say but it's not very comprehensible in its current form.

I'm a bit confused by this situation. Tell me if this is correct:
- Each row must have a unique ID
- But the grid does not require an `id` column specifically - the data set may not have unique IDs, or they may be stored in some other property not called `id`
- If a data set doesn't have an `id` field, use the `getRowId` to assign (?) an ID to the row that the grid can use

flaviendelangle:
> Each row must have a unique ID

To rephrase it, the grid must be able to get a unique id for each row.
Before even talking about where the id is and how the grid knows how to get it, the first thing is that the grid needs a unique id for each row.

> But the grid does not require an id column specifically - the data set may not have unique IDs, or they may be stored in some other property not called id

The "id" column, if is exists, has nothing to do with the identifier of the grid. That's a confusion José had and I think he"s not the only one at all.
You can create an column "id" if you want, but the grid will not use it to identify the rows, it will be a regular column to display the id.

> If a data set doesn't have an id field, use the getRowId prop to assign (?) an ID to the row that the grid can use

Yes. I just avoid talking about "field" because it's a name tightly linked to the columns (to avoid the confusion described above).
If `myRow.id` does not contain an unique ID, `getRowId` allows the dev to say "here is your id for this row".
We can distinguish two main use case I think:

1. There is a unique identifier in another property of the row model (`myRow._id`, `myRow.index`, etc...)

2. There is no unique identifier and the user must create it (could be by incrementing a global variable, may be by concatenating all the field of the model, ...). This should be avoided because it's can't work well with features like editing or external update of the row, but in some scenario it can work.

samuelsycamore:
Thanks for the clarification @flaviendelangle! It's starting to make more sense to me. The use cases you've described are particularly helpful. My first thought after reading the documentation as it stands was, "what if there is no unique ID in the data set to point to with `getRowId`?" It sounds like that scenario should be avoided. That might be worth spelling out here.

Here's how I understand it now:

Each row much have a unique identifier. 
By default, the grid looks for `id` in the data set to get that identifier. 
If the row's identifier is not called `id`, you need to use the `getRowId` to tell the grid where to find that information.
If no such unique identifier exists in the data set, then the developer must create it by some other means, but this scenario should be avoided because it leads to other issues.
In any case, it makes no difference whether or not there is a column in the grid to display the unique identifier data.

Is that correct?

As a side note—how do you feel about "unique identifier" here instead of "id" when not referring specifically to `id`? I think it helps to clarify things.

docs/data/data-grid/rows/rows.md
```diff
@@ -17,18 +17,27 @@ Otherwise, the grid will re-apply heavy work like sorting and filtering.
...
+## Row identifier
+
+Each row must have a unique id.
 
-When using dataset without a unique `id` property, you can use the `getRowId` prop to specify a custom id for each row.
+This id is used internally to identify the row in the various models (for instance the row selection model) and to track the row across updates.
+By default, the grid will use the `id` property of the row.
+
+When using a dataset without a unique `id` property, you can use the `getRowId` prop to specify the id of each row.
+This can be useful if your id is stored in another property like shown below:
```
samuelsycamore, line 28:
```suggestion
This identifier is used internally to identify the row in the various models—for instance, the row selection model—and to track the row across updates.

By default, the grid looks for a property named `id` in the data set to get that identifier.

If the row's identifier is not called `id`, then you need to use the `getRowId` prop to tell the grid where it's located.

The following demo shows how to use `getRowId` to grab the unique identifier from a property named `internalId`:
```

docs/data/data-grid/rows/rows.md
```diff
@@ -17,18 +17,27 @@ Otherwise, the grid will re-apply heavy work like sorting and filtering.
...
 
 ```tsx
 <DataGrid getRowId={(row) => row.internalId} />
 ```
 
 {{"demo": "RowsGridWithGetRowId.js", "bg": "inline", "defaultCodeOpen": false}}
```
samuelsycamore, line 35:
```suggestion

If no such unique identifier exists in the data set, then you must create it by some other means, but this scenario should be avoided because it leads to issues with other features of the grid.

Note that it is not necessary to create a column to display the unique identifier data. 
The data grid pulls this information directly from the data set itself, not from anything that is displayed on the screen.

```

samuelsycamore (commented):
Here's how I'd write this section, now that I understand how it works! 😁

docs/data/data-grid/rows/rows.md
```diff
@@ -17,18 +17,34 @@ Otherwise, the grid will re-apply heavy work like sorting and filtering.
...
+Each row must have a unique identifier.
+
+This identifier is used internally to identify the row in the various models—for instance, the row selection model—and to track the row across updates.
 
-When using dataset without a unique `id` property, you can use the `getRowId` prop to specify a custom id for each row.
+By default, the data grid looks for a property named `id` in the data set to get that identifier.
+
+If the row's identifier is not called `id`, then you need to use the `getRowId` prop to tell the grid where it's located.
+
+The following demo shows how to use `getRowId` to grab the unique identifier from a property named `internalId`:
```
oliviertassinari, line 30: Are we sure about the number of line breaks? It looks dense and hard to skim read:

Screenshot 2022-06-14 at 13 07 20

flaviendelangle: We can improve that one in https://github.com/mui/mui-x/pull/5195 
I felt like it was already an improvement and could be shiped as is if needed

oliviertassinari: I would personally enjoy reading this:

Screenshot 2022-06-14 at 13 09 23

oliviertassinari: > I felt like it was already an improvement and could be shiped as is if needed

@flaviendelangle For sure 👍

samuelsycamore: Both of these look good to me, so I'm happy to accommodate either style. I would say that in general, I don't think paragraphs should be any longer than they are in the latter version. With dense material like this, it's easy to get lost in the middle of a paragraph.

I think this paragraph is the ideal length for text like this:

Screen Shot 2022-06-14 at 9 11 03 AM

So we could say that as a general rule, you should consider a line break after 2 sentences or ~200 characters.

oliviertassinari: cc @mbrookes for perspective on the matter.

---

If we were to look for rules that could work in practice, I would personally imagine that above 500 characters, a paragraph is either too verbose and should be broken down, or it's actually covering a different matter and hence should be broken down too. In the docs of Stripe (one I would assume receives a lot of care), it's common to find https://stripe.com/docs/products-prices/how-products-and-prices-work paragraphs of almost 400 characters but It's hard to find find one above 500.

mbrookes: I'm with @oliviertassinari on this one. The second example is superior to the first. But if they needed to be to convey the concept, each paragraph could comfortably be twice as long. I don't think that there should be an arbitrary rule on paragraph length – paragraphs aren't measured in words or sentences; but I agree that verbosity should be eliminated, and that paragraphs should cover one narrow topic.

PR #5195, opened 2022-06-14
[docs] Split Rows doc page

flaviendelangle:
[Doc preview](https://deploy-preview-5195--material-ui-x.netlify.app/x/react-data-grid/row-definition/)

I merged #5156 but we can improve the content here if needed

Some of those pages have outdated content, we can improve them on this PR but if it's heavy changes I would suggest to split and then improve.

I don't really know where to put the "Styling rows" small section.

@m4theushw I moved the Master Detail doc into the "Row" folder instead of "Group & Pivot" as discussed a few weeks ago.


docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
...
+title: Data Grid - Row ordering
+---
+
+# Data grid - Row ordering [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+<p class="description">Drag and drop your rows to reorder them.</p>
+
+Row reordering allows to rearrange rows by dragging the special reordering cell.
+
+By default, row reordering is disabled.
```
flaviendelangle, line 11: I would add a paragraph describing the default behavior with a link to the Sorting page to let people know how those two features interplay.

I think this information is worth displaying up-front instead of in a warning at the end of the page

docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
...
+  ]}
+/>
+```
+
+This approach can also be used to change the location of the toggle column.
+
+:::warning
+For now, row reordering is disabled if sorting is applied to the grid.
+
+In addition, if row grouping or tree data is being used, the row reordering is also disabled.
```
flaviendelangle, line 79: I would add a section "Reordering rows with Tree Data" with a "Wait for upvote" issue
And maybe same for Row Grouping.

docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
...
+  columns={[
+    {
+      ...GRID_REORDER_COL_DEF, // Already contains the right field
+      width: 40,
+    },
+  ]}
+/>
+```
+
+This approach can also be used to change the location of the toggle column.
```
flaviendelangle, line 74: @DanailH What do you think about copying the structure of [this example](https://mui.com/x/react-data-grid/selection/#custom-checkbox-column) ?

To display both changes instead of adding an explanation after.

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,123 @@
...
+title: Data Grid - Row dimensions
+---
+
+# Data grid - Row dimensions
+
+<p class="description">Customize the dimensions of your rows.</p>
+
+## Static row height
+
+By default, the rows have a height of 52 pixels.
```
flaviendelangle, line 11: @samuelsycamore I feel like we could be more explicit here about the difference between rowHeight and density.

docs/data/data-grid/row-updates/row-updates.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Data Grid - Row updates
+---
+
+# Data grid - Row updates
```
flaviendelangle, line 5: This page will require a lot more content in v6 with #4927 and with all the lazy-loading related topics that @DanailH is handling.

It could even require a dedicated page folder at some point with a page for each approach.
But for now, we can improve this one.

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,123 @@
...
+
+If you want to create a more / less compact grid and not only set the row height, take a look at our [Density documentation](/x/react-data-grid/accessibility/#density-selector)
+
+To change the row height for the whole grid, set the `rowHeight` prop:
+
+{{"demo": "DenseHeightGrid.js", "bg": "inline"}}
+
+## Variable row height
+
+If you need some rows to have different row heights this can be achieved using the `getRowHeight` prop. This function is called for each visible row and if the return value is a `number` then that `number` will be set as that row's `rowHeight`. If the return value is `null` or `undefined` then the `rowHeight` prop will take effect for the given row.
```
m4theushw, line 22:
```suggestion
If you need some rows to have different row heights this can be achieved using the `getRowHeight` prop. This function is called for each visible row and if the return value is a `number` then that number will be set as that row's `rowHeight`. If the return value is `null` or `undefined`, then the `rowHeight` prop will take effect for the given row.
```

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,123 @@
+---
+title: Data Grid - Row dimensions
+---
+
+# Data grid - Row dimensions
```
m4theushw, line 5: What about renaming to "Row height"? The row has only one dimension: height.

docs/data/data-grid/row-updates/row-updates.md
m4theushw: Does it make sense to merge this page with "row definition", to keep consistency with https://mui.com/x/react-data-grid/column-definition/#providing-content? In essence, we're talking here about different approachs to feed the grid with rows. One of the approachs also allows to update rows (`updateRows`), but it could also be understood as replacing one row with another (updated) row, so "providing rows via API".

docs/data/pages.ts
```diff
@@ -31,7 +31,27 @@ const pages: MuiPage[] = [
...
+          '/x/react-data-grid/row-spanning',
+        ],
+        children: [
+          { pathname: '/x/react-data-grid/row-definition' },
+          { pathname: '/x/react-data-grid/row-updates' },
+          { pathname: '/x/react-data-grid/row-dimensions' },
+          { pathname: '/x/react-data-grid/row-ordering', plan: 'pro' },
+          { pathname: '/x/react-data-grid/master-detail', plan: 'pro' },
+          { pathname: '/x/react-data-grid/row-pinning', plan: 'pro', title: 'Row pinning 🚧' },
+          { pathname: '/x/react-data-grid/row-spanning', title: 'Row spanning 🚧' },
```
m4theushw, line 52: "master detail" in the middle of "rows" breaks the flow.

```suggestion
          { pathname: '/x/react-data-grid/row-pinning', plan: 'pro', title: 'Row pinning 🚧' },
          { pathname: '/x/react-data-grid/row-spanning', title: 'Row spanning 🚧' },
          { pathname: '/x/react-data-grid/master-detail', plan: 'pro' },
```

flaviendelangle: I think we can see it in two ways

1. We put *Master Detail* quite high because it's an important feature compared to some existing or upcoming row features

2. We put *Master Detail* at the end because it's the only one that does not start with *Row*


@samuelsycamore

docs/data/data-grid/row-definition/row-definition.md
```diff
@@ -0,0 +1,54 @@
+---
+title: Data Grid - Row definition
```
samuelsycamore, line 2: This feels vague—it isn't clear what to me what it means to "define" a row. It seems like the page is all about defining row identifiers, so I would go with that.

```suggestion
title: Data Grid - Defining row identifiers
```

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,125 @@
...
+<p class="description">Customize the dimensions of your rows.</p>
+
+## Static row height
+
+By default, the rows have a height of 52 pixels.
+This matches the normal height in the [Material Design guidelines](https://material.io/components/data-tables).
+
+If you want to create a more / less compact grid and not only set the row height, take a look at our [Density documentation](/x/react-data-grid/accessibility/#density-selector)
+
+To change the row height for the whole grid, set the `rowHeight` prop:
```
samuelsycamore, line 16:
```suggestion
Use the the `rowHeight` prop to change this default value, as shown below:
```

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,125 @@
...
+# Data grid - Row dimensions
+
+<p class="description">Customize the dimensions of your rows.</p>
+
+## Static row height
+
+By default, the rows have a height of 52 pixels.
+This matches the normal height in the [Material Design guidelines](https://material.io/components/data-tables).
+
+If you want to create a more / less compact grid and not only set the row height, take a look at our [Density documentation](/x/react-data-grid/accessibility/#density-selector)
```
samuelsycamore, line 14: Cut this to move it further down in this section.

```suggestion
```

docs/data/data-grid/row-dimensions/row-dimensions.md
```diff
@@ -0,0 +1,125 @@
...
+  sx={{
+    '&.MuiDataGrid-root--densityCompact .MuiDataGrid-cell': { py: '8px' },
+    '&.MuiDataGrid-root--densityStandard .MuiDataGrid-cell': { py: '15px' },
+    '&.MuiDataGrid-root--densityComfortable .MuiDataGrid-cell': { py: '22px' },
+  }}
+/>
+```
+
+:::
+
```
samuelsycamore, line 86: I think this deserves its own section.

```suggestion

## Row density

Give your users the option to change the default row density to match their preferences—compact, standard, or comfortable. 
Density is calculated based on the `rowHeight` and/or `headerHeight` props, when present.
See [Density](https://mui.com/x/react-data-grid/accessibility/#density) for details.

```

docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Data Grid - Row ordering
+---
+
+# Data grid - Row ordering [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+<p class="description">Drag and drop your rows to reorder them.</p>
+
+Row reordering allows to rearrange rows by dragging the special reordering cell.
```
samuelsycamore, line 9:
```suggestion
Row reordering lets users rearrange rows by dragging the special reordering cell.
```

docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
...
+  components={{
+    RowReorderIcon: CustomMoveIcon,
+  }}
+/>
+```
+
+Another way to customize is to add a column with `field: __reorder__` to your set of columns.
+That way, you can overwrite any of the properties from the `GRID_REORDER_COL_DEF` column.
+The grid will detect that there is already a reorder column defined and it will not add another one in the default position.
+By only setting the `field`, is up to you to configure the remaining options (e.g. disable the column menu, filtering, sorting).
```
samuelsycamore, line 60:
```suggestion
If you only set the `field`, then it is up to you to configure the remaining options (e.g. disable the column menu, filtering, sorting).
```

docs/data/data-grid/row-ordering/row-ordering.md
```diff
@@ -0,0 +1,86 @@
...
+To already start with a few suggested options configured, spread `GRID_REORDER_COL_DEF` when defining the column.
```
samuelsycamore, line 61:
```suggestion
To start with our suggested configuration, spread `GRID_REORDER_COL_DEF` when defining the column.
```

docs/data/data-grid/row-pinning/row-pinning.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Data Grid - Row pinning
+---
+
+# Data grid - Row pinning [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+<p class="description">Pin rows to keep them visible at all time.</p>
```
samuelsycamore, line 7:
```suggestion
<p class="description">Pin rows to keep them visible at all times.</p>
```

docs/data/data-grid/row-spanning/row-spanning.md
```diff
@@ -0,0 +1,24 @@
...
+
+:::warning
+This feature isn't implemented yet. It's coming.
+
+👍 Upvote [issue #207](https://github.com/mui/mui-x/issues/207) if you want to see it land faster.
+:::
+
+Each cell takes up the width of one row.
+Row spanning allows to change this default behavior.
+It allows cells to span multiple rows.
```
samuelsycamore, line 17:
```suggestion
Row spanning lets you change this default behavior, so cells can span multiple rows.
```

docs/data/data-grid/row-updates/row-updates.md
flaviendelangle: @joserodolfofreitas @samuelsycamore I'm interested in your opinion here

I agree that this section could be on the same page as the row update sections

image


But having the "Row identifier" section in the same page as well will start to make the page very long.

docs/data/pages.ts
flaviendelangle: @joserodolfofreitas as well

docs/data/data-grid/row-dimensions/row-dimensions.md
flaviendelangle: Would you rename the "Column dimensions" page "Columns width" ?

docs/data/data-grid/row-ordering/row-ordering.md
DanailH: Interesting. I was thinking in terms that this is something that a user can do so I wasn't sure what would be the benefit of showcasing this, but we can add a demo with the dragged column be the last one. The problem with this is that if you have a horizontal scroll then the dragged column will be out of view.

docs/data/data-grid/row-updates/row-updates.md
DanailH: Would it make sense to split that page into 2, one for `Row updates` and one for `Row loading`? The `Row loading` one will have the infinite + lazy loading and anything else related to loading rows on the fly.

joserodolfofreitas: Yeah, I agree it makes sense to move the lazy loading to its own page.
We could probably use one more depth level on the side bar to hold everything about "feeding data", but I agree it makes sense to separate `Row updates` from `Row Loading` (and merge `Row definitions` with `Row updates`).

But maybe we could rename this menus?

What about

- Data binding
- Lazy loading

?

DanailH: It can work but calling the page that is about loading rows "Lazy loading" won't be ideal as my idea was to have lazy and infinite loading there.

docs/data/pages.ts
joserodolfofreitas: I think it's enough if `Master detail` is the first Pro-listed feature. The break on the pattern name (+ Pro indication) will probably be enough to grab some extra attention.

docs/data/data-grid/row-updates/row-updates.md
joserodolfofreitas: then:

- Data binding
- Lazy and infinite loading

?

m4theushw: @joserodolfofreitas I would add items to the **left sidebar**. Keeping "lazy loading" inside https://deploy-preview-5195--material-ui-x.netlify.app/x/react-data-grid/row-updates/ will make the page very long. Note that we need to consider that "Row updates" and "Row definition" will be merged.

DanailH (approved):

docs/data/pages.ts
cherniavskii: I've moved master detail to be the first Pro feature in the navbar

docs/data/data-grid/row-definition/row-definition.md
cherniavskii: Apart from row identifiers, this page also mentions `rows` prop structure and has a warning about keeping the same reference.
Also, "Row definition" is consistent with "Column definition" page https://mui.com/x/react-data-grid/column-definition/

docs/data/data-grid/row-dimensions/row-dimensions.md
cherniavskii: I think "Row height" makes more sense here. We can discuss renaming "Column dimensions" later.

docs/data/data-grid/row-ordering/row-ordering.md
cherniavskii: Done

docs/data/data-grid/row-updates/row-updates.md
cherniavskii: @DanailH @m4theushw @joserodolfofreitas 
Did we reach an agreement on this topic? Or should I create a separate issue for discussion to unlock this PR?

docs/data/data-grid/row-definition/row-definition.md
cherniavskii: (Adding missing "Feeding data" section title to make it more clear)

joserodolfofreitas (approved):
Regardless of the improvements we may still want to do (like renaming some of the row sub-sections, or reorganizing the content), I think this PR serves its purpose already. It splits the row documentation and it's already a great improvement over what we had before.

PR #5200, opened 2022-06-14
[docs] Add plan badge next to event name in the Events page

flaviendelangle:
Same idea as for the api methods, the doc now displays the plan badge of the lowest plan on which the event is available

image


joserodolfofreitas (commented):
Great idea! And good timing too, given we consider opening the apiRef to the community plan.

I think we could try aligning the plans' icons. Perhaps with a specific column for it, or maybe aligned to left before the API's names?

In pages like the [Excel's export API](https://deploy-preview-5200--material-ui-x.netlify.app/x/api/data-grid/grid-excel-export-options/) it can feel a bit messy, and when we open to the community and also mark the Pro APIs, this will happen in some scale on most pages.

flaviendelangle:
> I think we could try aligning the plans' icons. Perhaps with a specific column for it, or maybe aligned to left before the API's names?
>
> In pages like the [Excel's export API](https://deploy-preview-5200--material-ui-x.netlify.app/x/api/data-grid/grid-excel-export-options/) it can feel a bit messy

True
For these api tables, it would be nice to have a redesign because the current shape is not very optimal (related to some discussions in https://github.com/mui/material-ui/issues/31745)

> and also mark the Pro APIs

It is already the case
For instance, [here](https://deploy-preview-5200--material-ui-x.netlify.app/x/api/data-grid/grid-api/) the detail-panel related apis are marked with the pro plan.

PR #5201, opened 2022-06-14
[core] Use the scrollbar size from `useGridDimensions` on `DataGridProColumnHeaders`

flaviendelangle:
If was looking at the `TODO` in our codebase and that one seemed pretty straightforward.


m4theushw (approved):
Thanks! I think this TODO was for me to address before opening https://github.com/mui/mui-x/pull/2946.

Merging with `master` should fix the types.

PR #5204, opened 2022-06-15
[pickers] Export adapters from both `@mui/x-date-pickers` and `@mui/x-date-pickers-pro`

### LINKED ISSUE #5132: [pickers] How to import `AdapterDateFns` when using pro plan ?
### Order ID 💳

44828

### Duplicates

### Latest version

### The problem in depth 🔍

From where do I import AdapterDateFns, since we have the pro version I think it should be like - 
`import { AdapterDateFns } from "@mui/x-date-pickers-pro/AdapterDateFns";`

But this is not working.

### Your environment 🌎

<details>

```
```
</details>

flaviendelangle:
Fixes #5132

My comment on the issue above

> The basic idea was: we don't want to create dozens of folders inside the pro package just re-exporting the community components just to be able to do `import { DatePicker } from '@mui/x-date-pickers-pro/DatePicker'`. Users could do `import { DatePicker } from '@mui/x-date-pickers-pro'` or if they really want dev tree shaking `import { DatePicker } from '@mui/x-date-pickers/DatePicker'`.
>
> But the adapters (I think they are the only one), we don't export them from the root of `@mui/x-date-pickers` and thus they are not re-exported from the root of `@mui/x-date-pickers-pro`.

### DIFF
--- a/docs/data/date-pickers/date-range-picker/BasicDateRangePicker.js
+++ b/docs/data/date-pickers/date-range-picker/BasicDateRangePicker.js
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import TextField from '@mui/material/TextField';
+import { LocalizationProvider } from '@mui/x-date-pickers-pro';
 import { DateRangePicker } from '@mui/x-date-pickers-pro/DateRangePicker';
-import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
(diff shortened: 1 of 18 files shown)


alexfauquette (approved):

alexfauquette:
A follow-up PR would be to modify demo mixing pro and community plans

PR #5210, opened 2022-06-15
[l10n] Add Swedish (sv-SE) locale

mrxdst:
Swedish locale already exists in Core.  
10M speakers.


alexfauquette (approved):
Thanks for sharing all those translations 🎉🇸🇪

Issue #1247, opened 2021-03-17
[DataGrid] Make infinite loader support rowCount


DanailH:
Make infinite loader work together with the rowCount prop

- [ ] I have searched the [issues](https://github.com/mui-org/material-ui-x/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

In order for the Infinite loader feature to be complete, we need to make it work in combination with the `rowCount ` prop. This means that a developer can set the total number of rows beforehand. The grid will then allocate space for all the rows and size this scrollbar correctly. 

Once the user starts scrolling the `onRowsScrollEnd` callback will be called requestion the rows within an area of [a, y] and load them in their appropriate place.

## Examples 🌈

```
 <XGrid
  {...data}
  rowCount={500}
  onRowsScrollEnd={handleOnRowsScrollEnd}
 />
```

## Motivation 🔦

Alow for a smoother loading when the total number of rows is known.

sebastianfrey:
Any plans when this feature will be released?

oliviertassinari:
> Any plans when this feature will be released?

@sebastianfrey Unclear, but we would love to hear more about the problem you wish to solve with this feature. Could you expand? :)

sebastianfrey:
@oliviertassinari Sure :).

I use the grid in a currently new developed GIS application to visualize the attributes of various heterogeneous spatial data sources with support for paginated queries. The amount of data this sources have stored goes from a few hundred records to a few hundred thousand and somtimes up to a million or more.

So while having a few hundered to thousand records the current approach by just loading the next batch of rows when the user reaches the bottom end of the scroll area is acceptable. With an increasing amount of records this becomes unwieldy.

Sure users can always filter the records by attributes or spatial conditions, but since in the current version of our application, the here proposed feature is already supported and there are likely some users, which will miss that feature.

It would be nice, to have this feature supported by XGrid as well.

FYI:

In the current version of our GIS application we use the Dojo Toolkit library [dgrid](https://github.com/SitePen/dgrid) in conjunction with a custom [dstore](https://github.com/SitePen/dstore) implementation for data fetching.

erwelch:
[...]

PR #5214, opened 2022-06-15
[DataGridPro] Implement Lazy loading

DanailH:
#1247 

Preview -> https://deploy-preview-5214--material-ui-x.netlify.app/x/react-data-grid/row-updates/#lazy-loading

## How does it work?
The approach is fairly straightforward. The new virtualization does make solving this problem more easily. The solution still doesn't on the `scrollTop`. It reacts when there are changes of the`renderContext` and emits an event that is handled by the `useGridLazyLoader` feature hook.
There is a small logic that finds the section between 2 given indexes of the rows array and only passes down to the user the start and end index of the rows that hadn't been loaded yet (skeleton rows).

The more interesting part is how the skeleton rows are being rendered. There is a pre-processor that adds logic to the `hydrateRows`. When the lazy load feature is enabled the pre-processor runs and it checks if there is a difference between the number of provided rows and the `rowCount` set by the user. If there is a difference then "empty" (skeleton rows, with only auto-generated `id` and no `model` added to the state). There is a small logic in the `GridRow.tsx` that handles the case where a row has no `model` and renders skeleton cells instead of `GridCell.tsx` component.

Last but not least, there is a new API method in the `useGridRows` that replaces a set of rows between 2 given indexes with a user-provided set. This is how the new rows are "loaded".

### ToDo:

- [x] add tests
- [x] polush the docs
- [x] make the feature exprimental


packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,140 @@
...
+        (rowId) => `${rowId}`.indexOf(GRID_SKELETON_ROW_ROOT_ID) >= 0,
+      );
+
+      if (!hasSkeletonRowIds) {
+        return;
+      }
+
+      renderedRowsIntervalCache.current = params;
+      const fetchRowsParams: GridFetchRowsParams = {
+        firstRowToRender: params.firstRowToRender,
```
m4theushw, line 70: I don't know if you plan to improve this, but the `firstRowToRender` might be the index of a row already loaded, making the server to return data not needed. I think we should pass to the callback only the indexes, between `firstRowToRender` and `lastRowToRender`, that contains unloaded rows.

Adding `rowBuffer=0` and `rowThresold=0` to a demo makes easier to see this problem.

docs/data/data-grid/rows/LazyLoadingGrid.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export default function LazyLoadingGrid() {
+  const { data } = useDemoData({
+    dataSet: 'Commodity',
+    rowLength: 10,
+    maxColumns: 6,
+  });
+
+  const handleFetchRows = async (params, event, details) => {
```
flaviendelangle, line 32: We should avoid using `details.api` since we want to remove it

docs/data/data-grid/rows/LazyLoadingGrid.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
+      <DataGridPro
+        {...data}
+        hideFooterPagination
+        rowCount={50}
+        sortingMode="server"
+        filterMode="server"
+        rowsLoadingMode="server"
```
flaviendelangle, line 52: I would be more precise than `server` here

https://www.ag-grid.com/javascript-data-grid/viewport/

AG-Grid calls it the *Viewport Row Model* which is one of the *Server-Side Data* models along with *Infinite Row Model* or *Server Side Row Model*.

docs/data/data-grid/rows/rows.md
```diff
@@ -80,6 +82,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
 
 {{"demo": "InfiniteLoadingGrid.js", "bg": "inline", "disableAd": true}}
 
+### Lazy loading [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+By default, infinite loading works on the client-side.
```
flaviendelangle, line 87: I would not talk about *infinite loading* which is mostly use to describe the scenario where the scrollbar shrink as you scroll (not what we have here if I understand correctly, the grid does not support infinite loading at all).


The problem might come from the name of `useGridInfiniteLoader` which is, I think mis-named

docs/data/data-grid/rows/rows.md
```diff
@@ -80,6 +82,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
...
+
+To switch it to server-side, set `rowsLoadingMode="server"`.
+Then the `rowCount` needs to be set and the number of initially loaded rows needs to be less than the `rowCount` value.
+In addition, you need to handle the `onFetchRows` callback to fetch the rows for the corresponding index.
+Finally, you need to use the `apiRef.current.insertRows()` to tell the DataGrid where to insert the newly fetched rows.
+
+**Note**: in order for the filtering and sorting to work you need to set their modes to `server`.
```
flaviendelangle, line 94: You can use `:::info`

packages/grid/x-data-grid-pro/src/DataGridPro/useDataGridProProps.ts
```diff
@@ -25,6 +26,7 @@ export const DATA_GRID_PRO_PROPS_DEFAULT_VALUES: DataGridProPropsWithDefaultValu
   disableChildrenFiltering: false,
   disableChildrenSorting: false,
   rowReordering: false,
+  rowsLoadingMode: GridFeatureModeConstant.client,
```
flaviendelangle, line 29: I would personally use a typed string (`type GridFeatureModeConstant = 'client' | 'thing1' | 'thing2'` instead of an enum to be more flexible and avoid forcing an import.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoaderPreProcessors.tsx
```diff
@@ -0,0 +1,47 @@
...
+  props: Pick<DataGridProProcessedProps, 'rows' | 'rowCount' | 'rowsLoadingMode'>,
+) => {
+  const addSkeletonRows = React.useCallback<GridPipeProcessor<'hydrateRows'>>(
+    (groupingParams) => {
+      if (
+        props.rowsLoadingMode === 'server' &&
+        props.rowCount &&
+        props.rows.length < props.rowCount
+      ) {
+        const newRowsIds: Array<GridRowId> = [...groupingParams.ids];
```
flaviendelangle, line 28:
```suggestion
        const newRowsIds: GridRowId[] = [...groupingParams.ids];
```

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
```diff
@@ -192,4 +200,11 @@ export interface DataGridProPropsWithoutDefaultValue<R extends GridValidRowModel
    * @param {GridCallbackDetails} details Additional details for this callback.
    */
   onRowOrderChange?: GridEventListener<'rowOrderChange'>;
+  /**
+   * Callback fired when rowCount is set and the next batch of virtualized rows is rendered.
+   * @param {GridFetchRowsParams} params With all properties from [[GridFetchRowsParams]].
+   * * @param {MuiEvent<{}>} event The event object.
```
flaviendelangle, line 206:
```suggestion
   * @param {MuiEvent<{}>} event The event object.
```

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -325,6 +326,63 @@ export const useGridRows = (
...
+      const updatedIdToIdLookup = { ...idToIdLookup };
+      const updatedTree = { ...tree };
+      const rowIdsToBeReplaced = allRows.slice(firstRowToRender, lastRowToRender);
+      rowIdsToBeReplaced.forEach((id) => {
+        delete updatedIdRowsLookup[id];
+        delete updatedIdToIdLookup[id];
+        delete updatedTree[id];
+      });
+
+      const newIdRowsLookup = newRows.reduce(
```
flaviendelangle, line 358: We should avoid looping several time the list of the rows

docs/data/data-grid/rows/LazyLoadingGrid.js
DanailH: it's called `server` because we use the same terminology for the filter and sorting feature modes. It uses the type `GridFeatureMode`

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
DanailH: The size of the `newRows` array will never be too big (depends on the viewport size and the buffer). Either way, I updated it from 3 loops to 1. It looks more condensed that way.

docs/data/data-grid/rows/LazyLoadingGrid.js
DanailH: I agree. we can revisit this while doing the breaking changes for v6.

docs/data/data-grid/rows/rows.md
```diff
@@ -33,6 +33,8 @@ The following demo shows how to use `getRowId` to grab the unique identifier fro
 <DataGrid getRowId={(row) => row.internalId} />
 ```
 
+:::
+
```
alexfauquette, line 37:
```suggestion
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,179 @@
...
+      ) {
+        return;
+      }
+
+      const renderedRowsIds: Array<GridRowId | null> = [...rowIds].splice(
+        params.firstRowToRender,
+        params.lastRowToRender - params.firstRowToRender,
+      );
+      const hasSkeletonRowIds = renderedRowsIds.some(
+        (rowId) => `${rowId}`.indexOf(GRID_SKELETON_ROW_ROOT_ID) >= 0,
```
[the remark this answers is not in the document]
alexfauquette: By the way, instead of `some`, you can use `findIndex` such that you get at the same time the start index of `skeletonRowSection`. And to get the last one you can use `findLastIndex`

docs/data/data-grid/rows/rows.md
```diff
@@ -80,6 +82,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
...
+Then the `rowCount` needs to be set and the number of initially loaded rows needs to be less than the `rowCount` value.
+In addition, you need to handle the `onFetchRows` callback to fetch the rows for the corresponding index.
+Finally, you need to use the `apiRef.current.insertRows()` to tell the DataGrid where to insert the newly fetched rows.
+
+:::info
+In order for the filtering and sorting to work you need to set their modes to `server`.
+You can find out more information about how to do that on the [server-side filter page](/components/data-grid/filtering/#server-side-filter) and on the [server-side sorting page](/components/data-grid/sorting/#server-side-sorting).
+:::
+
+{{"demo": "LazyLoadingGrid.js", "bg": "inline", "disableAd": true}}
```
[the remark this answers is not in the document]
DanailH: If all the rows are loaded the `onFetchRows` won't be called.
One optimization that is needed by the user is to implement throttle because now this callback is being called a lot while scrolling.

packages/grid/x-data-grid-premium/src/DataGridPremium/DataGridPremium.tsx
```diff
@@ -825,6 +832,13 @@ DataGridPremiumRaw.propTypes = {
    * Set of rows of type [[GridRowsProp]].
    */
   rows: PropTypes.array.isRequired,
+  /**
+   * Loading rows can be processed on the server or client-side.
+   * Set it to 'client' if you would like to handle the infnite loading on the client-side.
```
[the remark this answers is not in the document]
DanailH: updated

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,179 @@
...
+function findSkeletonRowsSection(
+  visibleRows: GridRowEntry<any>[],
+  range: { firstRowIndex: number; lastRowIndex: number },
+): { firstRowIndex: number; lastRowIndex: number } {
+  const visibleRowsSection = visibleRows.slice(range.firstRowIndex, range.lastRowIndex);
+  let firstRowIndex = range.firstRowIndex;
+  let lastRowIndex = range.lastRowIndex;
+  let startIndex = 0;
+  let endIndex = visibleRowsSection.length - 1;
+  let isSkeletonSectionFound = false;
```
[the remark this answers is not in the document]
DanailH: I need to know if at least 1 row in a range hasn't been loaded yet. That's why I chose this approach.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,220 @@
...
+    (newFilterModel) => {
+      const dimensions = apiRef.current.getRootDimensions();
+
+      if (
+        !(props.experimentalFeatures as GridExperimentalProFeatures)?.lazyLoading ||
+        !dimensions ||
+        props.rowsLoadingMode !== GridFeatureModeConstant.server
+      ) {
+        return;
+      }
```
m4theushw, line 187: It's a bit complex to read all these conditions. You could split them:

```suggestion
      const { lazyLoading } = (props.experimentalFeatures ?? {}) as GridExperimentalProFeatures;

      if (!lazyLoading || !dimensions) {
        return;
      }

      if (props.rowsLoadingMode !== GridFeatureModeConstant.server) {
        return;
      }

      if (
        renderedRowsIntervalCache.current.firstRowToRender === params.firstRowToRender &&
        renderedRowsIntervalCache.current.lastRowToRender === params.lastRowToRender
      ) {
        return;
      }
```

The two first `if` could even be moved to a function and reused in other places of this hook.

packages/grid/x-data-grid/src/components/GridRow.tsx
```diff
@@ -288,110 +288,141 @@ function GridRow(props: React.HTMLAttributes<HTMLDivElement> & GridRowProps) {
       ? rootProps.showCellRightBorder
       : !removeLastBorderRight && rootProps.disableExtendRowFullWidth;
 
-    const cellParams = apiRef.current.getCellParams(rowId, column.field);
+    if (row) {
```
m4theushw, line 291: Could move the content of this `if` to a function, e.g. `getCell`? Each branch is very long to read.

docs/data/data-grid/rows/LazyLoadingGrid.tsx
```diff
@@ -0,0 +1,66 @@
...
+async function sleep(duration: number) {
+  return new Promise<void>((resolve) => {
+    setTimeout(() => {
+      resolve();
+    }, duration);
+  });
+}
+
+const loadServerRows = async (newRowLength: number) => {
+  const newData = await getRealGridData(newRowLength, getCommodityColumns());
```
m4theushw, line 22: Maybe call `getRealGridData` only once, before the grid is loaded, then slice it here to get only the portion needed. This is to avoid replacing already loaded rows with different data.

DanailH:
> msedge_O89UHeNdw0.mp4

@m4theushw I've updated the demo so the problem is not visible anymore.

packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -278,9 +278,23 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
...
+
+      const [firstRowToRender, lastRowToRender] = getRenderableIndexes({
+        firstIndex: nextRenderContext.firstRowIndex,
+        lastIndex: nextRenderContext.lastRowIndex,
+        minFirstIndex: 0,
+        maxLastIndex: currentPage.rows.length,
+        buffer: rootProps.rowBuffer,
+      });
+
+      apiRef.current.publishEvent('renderedRowsIntervalChange', {
```
m4theushw, line 290: We already calculate `firstRowToRender` inside `getRows`. Would it work if we change to publish the event inside this method? You could use a ref to keep track of the last render context, to avoid publishing the event when nothing has changed.

The `buffer` param in `getRows` also considers if virtualization is disabled. This may not be relevant here but reduces duplicated code.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,256 @@
...
+
+      apiRef.current.publishEvent('fetchRows', fetchRowsParams);
+    },
+    [
+      apiRef,
+      props.rowsLoadingMode,
+      rowIds,
+      sortModel,
+      filterModel,
+      visibleRows,
```
m4theushw, line 163:
```suggestion
      visibleRows.rows,
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,256 @@
...
+      props.experimentalFeatures,
```
m4theushw, line 164: Same as comment.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,256 @@
...
+        buffer: props.rowBuffer,
+      });
+      const fetchRowsParams: GridFetchRowsParams = {
+        firstRowToRender,
+        lastRowToRender,
+        sortModel,
+        filterModel: newFilterModel,
+      };
+
+      apiRef.current.publishEvent('fetchRows', fetchRowsParams);
```
m4theushw, line 240: Could part of `handleRenderedRowsIntervalChange` be reused here? Same for `handleGridSortModelChange`. This code seems duplicated.

docs/data/data-grid/row-updates/row-updates.md
```diff
@@ -39,6 +39,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
 
 {{"demo": "InfiniteLoadingGrid.js", "bg": "inline", "disableAd": true}}
 
+## Lazy loading [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+To allow the grid to lazy load data, set `rowsLoadingMode="server"`.
+Then the `rowCount` needs to be set to the number of available rows on the server
+In addition, you need to handle the `onFetchRows` callback to fetch the rows for the corresponding index.
+Finally, you need to use the `apiRef.current.insertRows()` to tell the DataGrid where to insert the newly fetched rows.
```
m4theushw, line 47: The demo is using `unstable_replaceRows`. Which one is the right method? BTW, if we say to use a method it shouldn't have "unstable". IHMO, "replace" is better because it can be reused for other things too.

docs/data/data-grid/rows/rows.md
```diff
@@ -0,0 +1,323 @@
+---
+title: Data Grid - Rows
```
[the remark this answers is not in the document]
DanailH: I thought I removed it, sorry.

docs/data/data-grid/row-updates/row-updates.md
DanailH: My bad, I forgot to update it after changing the name of the method. I was thinking whether it should be unstable or not but since the feature is that's why I added the `unstable_` prefix.
The method itself is stable so I can remove the prefix and people can use it without the lazy loading.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
DanailH: `handleRenderedRowsIntervalChange ` does more than it's needed for the filtering and sorting cases.
You are right, the code is very similar, the difference is that in the filtering case the filter model that is being passed is from the function parameters and so the sorting model is from the state and vice versa for the sorting case.
I can make it into a single event handler but having them separated makes it easier to read and reason.

packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
DanailH: Doesn't this introduce additional complexity? I remember that I initially publish the event from `getRows` but then it fired more ofthen than needed so I adjusted it to be fired only when the context changes and this seemed the correct place to do it from, no?

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -325,6 +326,57 @@ export const useGridRows = (
     [apiRef, logger],
   );
 
+  const replaceRows = React.useCallback<GridRowApi['replaceRows']>(
+    (firstRowToRender, lastRowToRender, newRows) => {
+      if (props.signature === GridSignature.DataGrid) {
+        throw new Error(
+          [
+            "MUI: You can't replace rows using `apiRef.current.replaceRows` on the DataGrid.",
+            'You need to upgrade to the DataGridPro component to unlock this feature.',
```
cherniavskii, line 335:
```suggestion
            'You need to upgrade to DataGridPro or DataGridPremium component to unlock this feature.',
```

To make it consistent with other errors https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/DataGrid/DataGrid.tsx#L103

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -325,6 +326,57 @@ export const useGridRows = (
...
+          ].join('\n'),
+        );
+      }
+
+      const newRowsIds = newRows.map((row) => row.id);
+      const allRows = gridRowIdsSelector(apiRef);
+      const updatedRows = [...allRows];
+      updatedRows.splice(firstRowToRender, lastRowToRender - firstRowToRender, ...newRowsIds);
```
cherniavskii, line 343: What if `newRows.length !== lastRowToRender - firstRowToRender `?
Should we throw an error in that case?

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -325,6 +326,57 @@ export const useGridRows = (
...
+
+      const idRowsLookup = gridRowsLookupSelector(apiRef);
+      const idToIdLookup = gridRowsIdToIdLookupSelector(apiRef);
+      const tree = gridRowTreeSelector(apiRef);
+      const updatedIdRowsLookup = { ...idRowsLookup };
+      const updatedIdToIdLookup = { ...idToIdLookup };
+      const updatedTree = { ...tree };
+      const rowIdsToBeReplaced = allRows.slice(firstRowToRender, lastRowToRender);
+
+      rowIdsToBeReplaced.forEach((id) => {
```
cherniavskii, line 353: There's room for optimization here - `replaceRows` currently loops over `newRows` 3 times, we can reduce it to 2 loops:

```diff
diff --git a/packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts b/packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
index 38446fe86..326ecf52d 100644
--- a/packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
+++ b/packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
@@ -337,10 +337,8 @@ export const useGridRows = (
         );
       }
 
-      const newRowsIds = newRows.map((row) => row.id);
       const allRows = gridRowIdsSelector(apiRef);
       const updatedRows = [...allRows];
-      updatedRows.splice(firstRowToRender, lastRowToRender - firstRowToRender, ...newRowsIds);
 
       const idRowsLookup = gridRowsLookupSelector(apiRef);
       const idToIdLookup = gridRowsIdToIdLookupSelector(apiRef);
@@ -348,14 +346,17 @@ export const useGridRows = (
       const updatedIdRowsLookup = { ...idRowsLookup };
       const updatedIdToIdLookup = { ...idToIdLookup };
       const updatedTree = { ...tree };
-      const rowIdsToBeReplaced = allRows.slice(firstRowToRender, lastRowToRender);
 
-      rowIdsToBeReplaced.forEach((id) => {
-        delete updatedIdRowsLookup[id];
-        delete updatedIdToIdLookup[id];
-        delete updatedTree[id];
+      newRows.forEach((row, index) => {
+        const rowId = row.id;
+        const [replacedRowId] = updatedRows.splice(firstRowToRender + index, 1, rowId);
+
+        delete updatedIdRowsLookup[replacedRowId];
+        delete updatedIdToIdLookup[replacedRowId];
+        delete updatedTree[replacedRowId];
       });
 
+      // update lookups in separate loops to avoid deleting rows when sorting has changed
       newRows.forEach((row) => {
```

Initially I wanted to reduce it to a single loop, but then realized that it won't work with sorting. So 2 loops are needed here.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -325,6 +326,57 @@ export const useGridRows = (
...
+        delete updatedIdRowsLookup[id];
+        delete updatedIdToIdLookup[id];
+        delete updatedTree[id];
+      });
+
+      newRows.forEach((row) => {
+        updatedIdRowsLookup[row.id] = row;
+        updatedIdToIdLookup[row.id] = row.id;
+        updatedTree[row.id] = row;
```
cherniavskii, line 362: Tree should contain `GridRowTreeNodeConfig`, not `GridRowModel`.
Unfortunately Typescript doesn't catch that, because we use `any` as default type for `GridRowModel`.

Maybe there's a way to fix this typing issue in v6?
cc @flaviendelangle

flaviendelangle: Nice catch !

We had a discussion about setting `unknown` as the default row model instead of `any` but it was a pain when I tried to do it
We could probably try again at some point

cherniavskii: I've opened #5734 to address it

packages/grid/x-data-grid/src/components/GridRow.tsx
```diff
@@ -292,110 +400,51 @@ function GridRow(props: React.HTMLAttributes<HTMLDivElement> & GridRowProps) {
...
+      if (cellColSpanInfo && !cellColSpanInfo.spannedByColSpan) {
+        const { colSpan, width } = cellColSpanInfo.cellProps;
+        const cellProps = { width, colSpan, showRightBorder, indexRelativeToAllColumns };
+        cells.push(getCell(column, cellProps));
+      }
+    } else {
+      const cellColSpanInfo = apiRef.current.unstable_getCellColSpanInfo(
+        rowId,
+        indexRelativeToAllColumns,
       );
```
cherniavskii, line 418: `unstable_getCellColSpanInfo` call can be deduplicated and moved up outside of `if (row) {` statement

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,237 @@
...
+): { firstRowIndex: number; lastRowIndex: number } {
+  const visibleRowsSection = visibleRows.slice(range.firstRowIndex, range.lastRowIndex);
+  let firstRowIndex = range.firstRowIndex;
+  let lastRowIndex = range.lastRowIndex;
+  let startIndex = 0;
+  let endIndex = visibleRowsSection.length - 1;
+  let isSkeletonSectionFound = false;
+
+  while (!isSkeletonSectionFound && firstRowIndex < lastRowIndex) {
+    if (!visibleRowsSection[startIndex].model && !visibleRowsSection[endIndex].model) {
```
cherniavskii, line 39: Side note: at some point we might need to introduce invalidation mechanizm, so that rows that were already fetched won't get fetched again, unless certain amount of time have passed.

cherniavskii (commented):
Great job! 🎉

docs/data/data-grid/row-updates/row-updates.md
```diff
@@ -39,6 +39,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
 
 {{"demo": "InfiniteLoadingGrid.js", "bg": "inline", "disableAd": true}}
 
+## Lazy loading [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+To allow the grid to lazy load data, set `rowsLoadingMode="server"`.
+Then the `rowCount` needs to be set to the number of available rows on the server.
+In addition, you need to handle the `onFetchRows` callback to fetch the rows for the corresponding index.
+Finally, you need to use the `apiRef.current.replaceRows()` to tell the DataGrid where to insert the newly fetched rows.
```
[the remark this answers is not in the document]
joserodolfofreitas:
```suggestion
Finally, replace the empty rows with the newly fetched ones using `apiRef.current.replaceRows()` like in the demo below.
```

docs/data/data-grid/row-updates/row-updates.md
```diff
@@ -39,6 +39,20 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
...
+
+:::info
```
joserodolfofreitas, line 49: Would it make sense to display this info after the demo?

joserodolfofreitas (commented):
Nice job!

Two questions about the API:

1. Is there other values available for rowsLoadingMode ?
2. is it really necessary to set rowsLoadingMode? maybe the grid can apply lazy loading when the user sets a rowCount and the callback method (onFetchRows)?

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
DanailH: This is defined correctly. The previous line of `const updatedTree = { ...tree };` forces the correct type.

packages/grid/x-data-grid-pro/src/DataGridPro/useDataGridProProps.ts
```diff
@@ -6,6 +6,7 @@ import {
   GridSlotsComponent,
   DATA_GRID_PROPS_DEFAULT_VALUES,
   GridValidRowModel,
+  GridFeatureModeConstant,
```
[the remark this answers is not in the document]
DanailH: It is used in all places where there is server-side functionality: filtering, sorting, and pagination.
I remember we discussed removing it but no decision was made. I personally am in favor of keeping it but I'm interested in putting it into a discussion.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoaderPreProcessors.tsx
```diff
@@ -0,0 +1,53 @@
...
+  GridExperimentalProFeatures,
+} from '@mui/x-data-grid-pro/models/dataGridProProps';
+import { GridApiPro } from '@mui/x-data-grid-pro/models/gridApiPro';
+import { GridPipeProcessor, useGridRegisterPipeProcessor } from '@mui/x-data-grid/internals';
+import { GridFeatureModeConstant, GridRowId } from '@mui/x-data-grid';
+
+export const GRID_SKELETON_ROW_ROOT_ID = 'auto-generated-skeleton-row-root';
+
+const getSkeletonRowId = (index: GridRowId | null) => {
+  if (index == null) {
```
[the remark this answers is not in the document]
DanailH: In the initial render I think it does, yes.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,237 @@
...
+      );
+
+      if (!hasSkeletonRowIds) {
+        return;
+      }
+
+      const { firstRowIndex, lastRowIndex } = findSkeletonRowsSection(visibleRows.rows, {
+        firstRowIndex: params.firstRowToRender,
+        lastRowIndex: params.lastRowToRender,
+      });
```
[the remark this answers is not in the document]
DanailH: It can become complicated. The `hasSkeletonRowIds` pourpos is to know if there is a skeleton row in the current rendered and sorted `rowId`'s.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
DanailH: That is currently the case with the exception of the time variable. It is a good suggestion for improvement.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
cherniavskii: I don't think it's defined correctly. `tree` is of type `GridRowTreeConfig`:

```ts
export type GridRowTreeConfig = Record<GridRowId, GridRowTreeNodeConfig>;

export interface GridRowTreeNodeConfig {
  /**
   * The grid row id.
   */
  id: GridRowId;
  /**
   * The id of the row children.
   * @default []
   */
  children?: GridRowId[];
  /**
   * The id of the footer
   */
  footerId?: GridRowId | null;
  /**
   * The row id of the parent (null if this row is a top level row).
   */
  parent: GridRowId | null;
  /**
   * Current expansion status of the row.
   * @default false
   */
  childrenExpanded?: boolean;
  /**
   * 0-based depth of the row in the tree.
   */
  depth: number;
  /**
   * The key used to group the children of this row.
   */
  groupingKey: GridKeyValue | null;
  /**
   * The field used to group the children of this row.
   * Is `null` if no field has been used to group the children of this row.
   */
  groupingField: string | null;
  /**
   * If `true`, this node has been automatically added to fill a gap in the tree structure.
   * @default false
   */
  isAutoGenerated?: boolean;
  /**
   * Position of the row among its sibling.
   * @default 'body'
   */
  position?: 'body' | 'footer';
  /**
   * If `true`, this row is pinned.
   * @default false
   */
  isPinned?: boolean;
}
```

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -331,6 +332,69 @@ export const useGridRows = (
...
+            "MUI: You can't replace rows using `apiRef.current.replaceRows` on the DataGrid.",
+            'You need to upgrade to DataGridPro or DataGridPremium component to unlock this feature.',
+          ].join('\n'),
+        );
+      }
+
+      if (newRows.length && newRows.length !== lastRowToRender - firstRowToRender) {
+        throw new Error(
+          [
+            'MUI: The number of rows you want update is more than what was indicated',
```
cherniavskii, line 349: It could be less as well
```suggestion
            'MUI: The number of rows you want update is different than what was indicated',
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,233 @@
...
+  DataGridProProcessedProps,
+  GridExperimentalProFeatures,
+} from '../../../models/dataGridProProps';
+import { GRID_SKELETON_ROW_ROOT_ID } from './useGridLazyLoaderPreProcessors';
+import { GridFetchRowsParams } from '../../../models/gridFetchRowsParams';
+
+function findSkeletonRowsSection(
+  visibleRows: GridRowEntry<any>[],
+  range: { firstRowIndex: number; lastRowIndex: number },
+): { firstRowIndex: number; lastRowIndex: number } {
```
cherniavskii, line 30:
```suggestion
) {
```

to infer the return type

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,233 @@
...
+import { GridApiPro } from '../../../models/gridApiPro';
+import {
+  DataGridProProcessedProps,
+  GridExperimentalProFeatures,
+} from '../../../models/dataGridProProps';
+import { GRID_SKELETON_ROW_ROOT_ID } from './useGridLazyLoaderPreProcessors';
+import { GridFetchRowsParams } from '../../../models/gridFetchRowsParams';
+
+function findSkeletonRowsSection(
+  visibleRows: GridRowEntry<any>[],
```
cherniavskii, line 28:
```suggestion
  visibleRows: GridRowEntry[],
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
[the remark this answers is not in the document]
cherniavskii: Yes, but `findSkeletonRowsSection` would also tell whether there are skeleton rows or not.
It operates in the same bounds of `[findSkeletonRowsSection, lastRowToRender]` as `hasSkeletonRowIds`.

Consider the use case I've described above - it would be much faster to use `findSkeletonRowsSection` without even trying to evaluate `hasSkeletonRowIds`.

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoaderPreProcessors.tsx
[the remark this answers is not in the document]
cherniavskii: It looks like `getSkeletonRowId` will only be called if there are skeleton rows:

```ts
for (let i = 0; i < props.rowCount - groupingParams.ids.length; i += 1) {
  const skeletonId = getSkeletonRowId(i);
  newRowsIds.push(skeletonId);
}
```

So `index` will always be a number, `null` shouldn't be the case:
```diff
- const getSkeletonRowId = (index: GridRowId | null) => {
+ const getSkeletonRowId = (index: number) => {
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
```diff
@@ -0,0 +1,220 @@
...
+
+function findSkeletonRowsSection(
+  visibleRows: GridRowEntry[],
+  range: { firstRowIndex: number; lastRowIndex: number },
+) {
+  let { firstRowIndex, lastRowIndex } = range;
+  let isSkeletonSectionFound = false;
+
+  while (!isSkeletonSectionFound && firstRowIndex < lastRowIndex) {
+    if (!visibleRows[firstRowIndex].model && !visibleRows[lastRowIndex].model) {
```
m4theushw, line 32: The first demo in https://deploy-preview-5214--material-ui-x.netlify.app/x/react-data-grid/row-updates/#lazy-loading is crashing here when I scroll all the way to the bottom.

docs/data/data-grid/row-updates/row-updates.md
```diff
@@ -39,6 +39,35 @@ In addition, the area in which `onRowsScrollEnd` is called can be changed using
...
+:::warning
+This feature is experimental, it needs to be explicitly activated using the `lazyLoading` experimental feature flag.
+
+```tsx
+<DataGridPro experimentalFeatures={{ lazyLoading: true }} {...otherProps} />
+```
+
+:::
+
+Lazy Loading works like a pagination system, but instead of loading new rows based on pages, it loads them as the user scroll through the grid and reveals empty rows.
```
m4theushw, line 53:
```suggestion
Lazy Loading works like a pagination system, but instead of loading new rows based on pages, it loads them as the user scrolls through the grid and reveals empty rows.
```

packages/grid/x-data-grid-pro/src/hooks/features/lazyLoader/useGridLazyLoader.ts
DanailH: I knew there was a reason why I used an additional set of indexes to track this loop. The tests were failing last night. I reverted the change and now it is working as expected.

CC @cherniavskii

m4theushw (approved):
I'm happy with the result. Well done!

We may need to think about later how to support properly the Select All checkbox. Currently, if I select all rows, it selects all rows, but then, after scrolling, it sanitizes the selected rows keeping only the ones in the `rows` prop. Maybe Select All shouldn't select skeleton rows.

oliviertassinari (commented):
A first quick pass

packages/grid/x-data-grid/src/utils/utils.ts
```diff
@@ -170,3 +170,20 @@ export function isDeepEqual(a: any, b: any) {
...
+    let t = (a += 0x6d2b79f5);
+    t = Math.imul(t ^ (t >>> 15), t | 1);
+    t ^= t + Math.imul(t ^ (t >>> 7), t | 61);
+    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
+    /* eslint-enable */
+  };
+}
+
+export function randomNumberBetween(seed: number, min: number, max: number): () => number {
+  const random = mulberry32(seed);
```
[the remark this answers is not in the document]
DanailH: With just math random on each scroll, the length of the skeleton cells is different. Passing a specific seed makes it stable, otherwise, it looks a bit buggy.

cherniavskii:
Oops, sorry!
I didn't expect https://github.com/mui/mui-x/pull/5734 to close this PR :D

docs/data/data-grid/row-updates/LazyLoadingGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const ROW_COUNT = 10000;
+const DATASET_OPTION: UseDemoDataOptions = {
+  dataSet: 'Employee',
+  rowLength: ROW_COUNT,
+};
+
+const { columns, useQuery } = createFakeServer(DATASET_OPTION);
+
+export default function LazyLoadingGrid() {
```
oliviertassinari, line 17: This demo has a couple of bugs:

1. Filtering is not supported
2. Sorting gets outdated

I have looked into a fix, it can be done with:

```diff
diff --git a/docs/data/data-grid/row-updates/LazyLoadingGrid.tsx b/docs/data/data-grid/row-updates/LazyLoadingGrid.tsx
index 79dbe9813..fa22ebcf6 100644
--- a/docs/data/data-grid/row-updates/LazyLoadingGrid.tsx
+++ b/docs/data/data-grid/row-updates/LazyLoadingGrid.tsx
@@ -4,30 +4,65 @@ import {
   GridFetchRowsParams,
   useGridApiRef,
 } from '@mui/x-data-grid-pro';
-import { createFakeServer, UseDemoDataOptions } from '@mui/x-data-grid-generator';
+import {
+  createFakeServer,
+  QueryOptions,
+  loadServerRows,
+  UseDemoDataOptions,
+} from '@mui/x-data-grid-generator';

-const ROW_COUNT = 10000;
 const DATASET_OPTION: UseDemoDataOptions = {
   dataSet: 'Employee',
-  rowLength: ROW_COUNT,
+  rowLength: 10000,
 };

-const { columns, useQuery } = createFakeServer(DATASET_OPTION);
+const { columns, columnsWithDefaultColDef, useQuery } =
+  createFakeServer(DATASET_OPTION);

 export default function LazyLoadingGrid() {
   const apiRef = useGridApiRef();
-  const cachedQuery = React.useMemo(() => ({}), []);
-  const { data, loadServerRowsInterval } = useQuery(cachedQuery);
-  const cachedRows = React.useMemo(() => data.slice(0, 10), [data]);
+  const [queryOptions, setQueryOptions] = React.useState<QueryOptions>({});
+  const { data } = useQuery(queryOptions);
+
+  const initialRows = React.useMemo(() => data.slice(0, 10), [data]);
   const cachedColumns = React.useMemo(() => columns, []);

   const handleFetchRows = async (params: GridFetchRowsParams) => {
-    const newRowsBatch = await loadServerRowsInterval(params);
+    // Bug, it shouldn't be triggered for the initial rendering
+    if (data.length === 0) {
+      return;
+    }
+
+    setQueryOptions((prev) => {
+      if (
+        prev.filterModel === params.filterModel &&
+        prev.sortModel === params.sortModel
+      ) {
+        return prev;
+      }
+
+      return {
+        filterModel: params.filterModel,
+        sortModel: params.sortModel,
+      };
+    });
+
+    const serverRows = await loadServerRows(
+      data,
+      {
+        filterModel: params.filterModel,
+        sortModel: params.sortModel,
+      },
+      {
+        useCursorPagination: false,
+      },
+      columnsWithDefaultColDef,
+    );

     apiRef.current.replaceRows(
       params.firstRowToRender,
       params.lastRowToRender,
-      newRowsBatch.returnedRows,
+      serverRows.returnedRows.slice(params.firstRowToRender, params.lastRowToRender),
     );
   };

@@ -35,10 +70,10 @@ export default function LazyLoadingGrid() {
     <div style={{ height: 400, width: '100%' }}>
       <DataGridPro
         columns={cachedColumns}
-        rows={cachedRows}
+        rows={initialRows}
         apiRef={apiRef}
         hideFooterPagination
-        rowCount={ROW_COUNT}
+        rowCount={data.length}
         sortingMode="server"
         filterMode="server"
         rowsLoadingMode="server"
diff --git a/packages/grid/x-data-grid-generator/src/hooks/useQuery.ts b/packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
index 84648e349..87478a253 100644
--- a/packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
+++ b/packages/grid/x-data-grid-generator/src/hooks/useQuery.ts
@@ -107,7 +107,7 @@ const getFilteredRows = (
 /**
  * Simulates server data loading
  */
-const loadServerRows = (
+export const loadServerRows = (
   rows: GridRowModel[],
   queryOptions: QueryOptions,
   serverOptions: ServerOptions,
@@ -170,12 +170,14 @@ interface PageInfo {
   pageSize?: number;
 }

-export interface ServerOptions {
+interface DefaultServerOptions {
   minDelay: number;
   maxDelay: number;
   useCursorPagination?: boolean;
 }

+type ServerOptions = Partial<DefaultServerOptions>;
+
 export interface QueryOptions {
   cursor?: GridRowId;
   page?: number;
@@ -196,7 +198,7 @@ const DEFAULT_DATASET_OPTIONS: UseDemoDataOptions = {
 declare const DISABLE_CHANCE_RANDOM: any;
 const disableDelay = typeof DISABLE_CHANCE_RANDOM !== 'undefined' && DISABLE_CHANCE_RANDOM;

-const DEFAULT_SERVER_OPTIONS: ServerOptions = {
+const DEFAULT_SERVER_OPTIONS: DefaultServerOptions = {
   minDelay: disableDelay ? 0 : 100,
   maxDelay: disableDelay ? 0 : 300,
   useCursorPagination: true,
@@ -204,7 +206,7 @@ const DEFAULT_SERVER_OPTIONS: ServerOptions = {

 export const createFakeServer = (
   dataSetOptions?: Partial<UseDemoDataOptions>,
-  serverOptions?: Partial<ServerOptions>,
+  serverOptions?: ServerOptions,
 ) => {
   const dataSetOptionsWithDefault = { ...DEFAULT_DATASET_OPTIONS, ...dataSetOptions };
   const serverOptionsWithDefault = { ...DEFAULT_SERVER_OPTIONS, ...serverOptions };
@@ -275,43 +277,11 @@ export const createFakeServer = (
     // We use queryOptions pointer to be sure that isLoading===true as soon as the options change
     const effectShouldStart = queryOptionsRef.current !== queryOptions;

-    const loadServerRowsInterval = (
-      query: Pick<
-        QueryOptions,
-        'filterModel' | 'sortModel' | 'firstRowToRender' | 'lastRowToRender'
-      >,
-      server = serverOptionsWithDefault,
-    ): Promise<FakeServerResponse> => {
-      const { minDelay = 100, maxDelay = 300 } = server;
-
-      if (maxDelay < minDelay) {
-        throw new Error('serverOptions.minDelay is larger than serverOptions.maxDelay ');
-      }
-
-      const delay = Math.random() * (maxDelay - minDelay) + minDelay;
-      const { firstRowToRender, lastRowToRender } = query;
-      let filteredRows = getFilteredRows(rows, query.filterModel, columnsWithDefaultColDef);
-      const rowComparator = getRowComparator(query.sortModel, columnsWithDefaultColDef);
-      filteredRows = [...filteredRows].sort(rowComparator);
-      const totalRowCount = filteredRows.length;
-      const res: FakeServerResponse = {
-        returnedRows: filteredRows.slice(firstRowToRender, lastRowToRender),
-        totalRowCount,
-      };
-
-      return new Promise<FakeServerResponse>((resolve) => {
-        setTimeout(() => {
-          resolve(res);
-        }, delay); // simulate network latency
-      });
-    };
-
     return {
       isLoading: isLoading || effectShouldStart,
-      loadServerRowsInterval,
       ...response,
     };
   };

-  return { columns, initialState, useQuery };
+  return { columns, columnsWithDefaultColDef, initialState, useQuery };
 };
 ```

https://user-images.githubusercontent.com/3165635/186415286-ee025157-9c95-4d7b-8783-0a9a191a3bab.mov

Now, there is still something a bit off. When the network request is too fast, the Skeleton creates a blink.

**Review by oliviertassinari — COMMENTED** [review]
The second pass. I'm running out of time budget 😁

File: packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts [hunk 42]
```diff
@@ -331,6 +332,69 @@ export const useGridRows = (
...
+
+        delete updatedIdRowsLookup[replacedRowId];
+        delete updatedIdToIdLookup[replacedRowId];
+        delete updatedTree[replacedRowId];
+      });
+
+      newRows.forEach((row) => {
+        updatedIdRowsLookup[row.id] = row;
+        updatedIdToIdLookup[row.id] = row.id;
+        updatedTree[row.id] = row as GridRowTreeNodeConfig;
```
cherniavskii, line 380: This is what I mentioned in https://github.com/mui/mui-x/pull/5214#discussion_r939047535
updatedTree should contain `GridRowTreeNodeConfig`, not `GridRowModel`

packages/grid/x-data-grid/src/utils/utils.ts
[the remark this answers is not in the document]
DanailH: For now, we can keep it in the helper's file, yes.

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
```diff
@@ -194,6 +209,13 @@ export interface DataGridProPropsWithoutDefaultValue<R extends GridValidRowModel
    * @param {GridCallbackDetails} details Additional details for this callback.
    */
   onRowOrderChange?: GridEventListener<'rowOrderChange'>;
+  /**
+   * Callback fired when rowCount is set and the next batch of virtualized rows is rendered.
+   * @param {GridFetchRowsParams} params With all properties from [[GridFetchRowsParams]].
+   * @param {MuiEvent<{}>} event The event object.
+   * @param {GridCallbackDetails} details Additional details for this callback.
+   */
+  onFetchRows?: GridEventListener<'fetchRows'>;
```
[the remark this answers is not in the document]
DanailH: It was like this in the earlier iterations but based on peer feedback I've removed the `unstable_`. The API itself is stable, its combination with the lazy loading feature might make it unstable.

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
```diff
@@ -104,6 +112,13 @@ export interface DataGridProPropsWithDefaultValue extends DataGridPropsWithDefau
    * @default false
    */
   rowReordering: boolean;
+  /**
+   * Loading rows can be processed on the server or client-side.
+   * Set it to 'client' if you would like enable infnite loading.
+   * Set it to 'server' if you would like to enable lazy loading.
+   * * @default "client"
+   */
+  rowsLoadingMode: GridFeatureMode;
```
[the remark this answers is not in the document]
DanailH: This is also used in infinite loading. The combination with the lazy loading is what might cause issues.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -331,6 +332,69 @@ export const useGridRows = (
...
+      if (props.signature === GridSignature.DataGrid) {
+        throw new Error(
+          [
+            "MUI: You can't replace rows using `apiRef.current.replaceRows` on the DataGrid.",
+            'You need to upgrade to DataGridPro or DataGridPremium component to unlock this feature.',
+          ].join('\n'),
+        );
+      }
+
+      if (newRows.length && newRows.length !== lastRowToRender - firstRowToRender) {
```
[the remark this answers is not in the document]
DanailH: Adding the `+ 1` causes it to always be true.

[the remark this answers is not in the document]
oliviertassinari: If it's always true, it's maybe because the params are always wrong.

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
```diff
@@ -94,4 +94,15 @@ export interface GridRowApi {
    * @returns {GridRowId[]} The id of each row in the grouping criteria.
    */
   getRowGroupChildren: (params: GridRowGroupChildrenGetterParams) => GridRowId[];
+  /**
+   * Replace a set of rows with new rows.
```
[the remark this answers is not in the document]
DanailH: I'm not sure. The names are quite self explanatory but still, if we see that people are getting confused we can think of where to add that explanation.

packages/grid/x-data-grid/src/components/GridRow.tsx
```diff
@@ -295,110 +410,44 @@ function GridRow(props: React.HTMLAttributes<HTMLDivElement> & GridRowProps) {
...
+        const { colSpan, width } = cellColSpanInfo.cellProps;
+        const cellProps = { width, colSpan, showRightBorder, indexRelativeToAllColumns };
+        cells.push(getCell(column, cellProps));
+      } else {
+        const { width } = cellColSpanInfo.cellProps;
+        const contentWidth = Math.round(randomNumber());
+
+        cells.push(
+          <rootProps.components.SkeletonCell
+            key={`col-${column.field}-${i}`}
```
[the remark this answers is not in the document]
DanailH: But doing it that way will result in all skeleton cells in a column having the same keys, no? The index is what differentiates them.

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
```diff
@@ -94,4 +94,15 @@ export interface GridRowApi {
...
+   * @param {number} firstRowToReplace The index of the first row to be replaced.
+   * @param {number} lastRowToReplace The index of the last row to be replaced.
+   * @param {GridRowModel[]} newRows The new rows.
+   */
+  replaceRows: (
```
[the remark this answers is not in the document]
DanailH: Same answer. The grid row API is stable. You can use this method for other purposes.

packages/grid/x-data-grid/src/components/GridRow.tsx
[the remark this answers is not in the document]
oliviertassinari: > But doing it that way will result in all skeleton cells in a column having the same keys, no?

I don't see why. Each row is a div. The parent of the Skeleton is a that div. Another way to look at if. If this was true, then how would it with with the non Skeleton cells that are using the proposed diff?

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
```diff
@@ -326,11 +327,84 @@ export const useGridRows = (
...
+        const [replacedRowId] = updatedRows.splice(firstRowToRender + index, 1, rowId);
+
+        delete updatedIdRowsLookup[replacedRowId];
+        delete updatedIdToIdLookup[replacedRowId];
+        delete updatedTree[replacedRowId];
+      });
+
+      newRows.forEach((row) => {
+        const rowTreeNodeConfig: GridRowTreeNodeConfig = {
+          ...row,
```
cherniavskii, line 381:
```suggestion
```

No need for it

cherniavskii (approved):

packages/grid/x-data-grid/src/components/cell/GridSkeletonCell.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+function GridSkeletonCell(props: React.HTMLAttributes<HTMLDivElement> & GridSkeletonCellProps) {
+  const { field, align, width, contentWidth, ...other } = props;
+  const rootProps = useGridRootProps();
+  const ownerState = { classes: rootProps.classes, align };
+  const classes = useUtilityClasses(ownerState);
+
+  return (
+    <div className={classes.root} style={{ width }} {...other}>
+      <Skeleton sx={{ mx: 1 }} width={`${contentWidth}%`} />
```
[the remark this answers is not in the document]
gerdadesign: agreed that after is better!

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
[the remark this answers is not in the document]
oliviertassinari: Is there a use case for `onFetchRows` without lazy loading? It seems that it's not the case, considering the logic. So from my perspective, if lazy loading is unstable, so is this API.

Personally, I wonder if doesn't make sense to drop `onFetchRows` and replace it with a new API that only focuses on making it possible for developers to subscribe to the visible rows, regardless of a skeleton cache.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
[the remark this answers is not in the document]
oliviertassinari: > If I want the rows between index 5 and 10 to be replaced

Here is how I under it:

5
6
7
8
9
10

This is 6 rows

docs/data/data-grid/rows/rows.md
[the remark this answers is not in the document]
oliviertassinari: It seems to fetch for each row that becomes visible in the viewport. In practice, it's probably not usable in production in the current shape as it would flood the server API.

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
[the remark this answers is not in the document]
oliviertassinari: But then I'm also wondering. Why is `updateRows` not enough for our use case?

docs/data/data-grid/row-updates/LazyLoadingGrid.tsx
oliviertassinari: Ok thanks.

oliviertassinari:
I have pushed a simplified version of the demo, aiming to be as close as possible to how developers might fetch the data. It fixes the double fetch flickers. However, it seems that we have a couple of issues. A summary of what I think is off, by order of importance (to complement the existing open comments on lines of code):

1. `onFetchRows` is called every time a new row is in the viewport, so when you scroll, you can easily throw 50 requests to your backend. So IMHO, this can't be used in production by a developer. I think that we should either change its behavior, for example, call it something like `onViewportChange` or we should batch the calls by introducing a notion of chunk size so it can be used for data fetching.
1. When using the filter in the demo, there are a lot of strange behaviors: flickering of row rendering crashes of out-of-range row updates crashes of duplicated React keys. I don't know why.
1. The `onFetchRows` API itself has no notion of cancellable requests. It assumes that calls are resolved instantly. But if you fetch, sort, and request a previous sortModel resolves, it breaks your data. So IMHO, this can't be used in production by a developer. The issue can be seen when filtering the demo. The results displayed are the ones from an incomplete search because they resolve after the query will the full request.
1. For the skeleton, add a small delay for appearing, like 100ms, to avoid flashes.

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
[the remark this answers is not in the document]
DanailH: The main difference is that `updateRows` does not update the `id` of the row, only the data. I entertained the idea of updating the `updateRows` to handle that use case as well but ultimately decided to have a separate API for it.
PS: the skeleton rows have autogenerated IDs so when a user loads rows that need to replate the skeleton rows their IDs will also be different.

packages/grid/x-data-grid-pro/src/models/dataGridProProps.ts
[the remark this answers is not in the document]
DanailH: Valid arguments.
- Right now I wouldn't drop the `experimentalFeatures` as we are already using it for other features and it might get messy. We can think of dropping it for v6.
- Regarding dropping the `onFetchRow`: an event that the user subscribes to can also work. I followed the already established way we have. Having a callback seems easier to implement and understand (at least for me).

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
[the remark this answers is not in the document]
DanailH: The way it is now is without the last row? So from 5 to 9 included. A way around this is to actually remove the second param (`lastRowToRender `) in the `replaceRows` method. That way the developer is free to decide how many rows they would want to load each time (set the threshold themselves). The `onFetchRows` will still pass the first and last visible rows but for replacing one just say from which index they want to start replacing and how many rows in total, what do you think?

On second thought: adding `+ 1` is not really a problem. The bug is not with the method itself but with how `splice` works. So a simple fix is to just do `serverRows.returnedRows.slice(params.firstRowToRender, params.lastRowToRender + 1)` in the demo.

DanailH:
Ok, I've pushed the solution a bit more and cleaned most of the issues.

> 1. When using the filter in the demo, there are a couple of strange behaviors: flickering of row rendering crashes of out-of-range row updates, and crashes of duplicated React keys. I don't know why.

Should be fixed now

> 3. onFetchRows is called every time a new row is in the viewport, so when you scroll, you can easily throw 50 requests to your backend.

I left it like this intentionally. We can add a line in the doc that advises the developers to implement either throttle or debounce themselves. Adding this functionality as part of the feature increases the complexity quite a lot.

> 4. For the skeleton, add a small delay for appearing, like 100ms, to avoid flashes

I guess this will be added once the Core adds that prop?

> 2. The onFetchRows API itself has no notion of cancellable requests. It assumes that calls are resolved instantly. But if you fetch, sort, and request a previous sortModel resolves, it breaks your data.

Can't this be solved the same way `3.` is?

We can merge this version and see what the feedback will be. We can continue to polish it more but it's been quite a long time and I feel we are operating based on assumptions. Not that it is wrong but I would love to start a clean PR after we get some real-world use cases/bugs 😅

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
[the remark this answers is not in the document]
oliviertassinari: Ah right OK. So one updates a row id, the other a row index.

packages/grid/x-data-grid/src/hooks/features/rows/useGridRows.ts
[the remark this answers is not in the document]
oliviertassinari: Ok, so the options are:

1. Remove the second param (lastRowToRender ) in the replaceRows method
2. Document that `lastRowToRender` is not the last row to render, but the one just before
3. Make the `lastRowToRender` the last row to render

Option 1. sounds good, a better DX, it's also closer to how `updateRows` behaves.

oliviertassinari:
> Should be fixed now

Great, thanks for looking into it. I have found a few more issues on https://deploy-preview-5214--material-ui-x.netlify.app/x/react-data-grid/row-updates/#lazy-loading.

a. When sorting, the rows updates twice:

https://user-images.githubusercontent.com/3165635/186713292-80815809-ae33-4e67-97c8-c61d3585ffc4.mov

b. When opening the filtering panel with an active sort, the rows change:

https://user-images.githubusercontent.com/3165635/186713631-f7aa338c-0fb3-459d-8a42-e22210f20a1d.mov

c. When filtering with an active sort, there are rows that shouldn't match

Uploading Screen Recording 2022-08-25 at 18.02.30.mov…

And I think that it's :). It might be an issue with the demo, and not the source.

> Can't this be solved the same way 3. is?

> I left it like this intentionally. We can add a line in the doc that advises the developers to implement either throttle or debounce themselves. Adding this functionality as part of the feature increases the complexity quite a lot.

It's definitely more complex, but I also think why we should care. It's precisely for the hard problems that developers will come to use our stuff.

My main concern is: Are we shipping a demo that can't be used in production and with APIs that are not `unstable_` so that we can't change without waiting for v6?

From what I understand, if a developer uses https://deploy-preview-5214--material-ui-x.netlify.app/x/react-data-grid/row-updates/#lazy-loading as is, he will DDoS his backend and will show rows to his end-users that don't match what they filter for (out-of-sync). So I would expect that he will need to do further work to solve 2. and 3 on his own. Since we know it, I think that we should fix it before we tell them to use the demo. I'm afraid that if we don't, we might harm the trust that developers have in their ability to copy & paste from the demo in the docs. Since the problem is complex, 💯 to only fix it in the demo. It will give us time to find an API that can then accommodate tree data, pivoting, etc.

> I guess this will be added once the Core adds that prop?

We could wait for https://github.com/mui/material-ui/issues/34057. However, I think that it would be great to first dogfood this idea inside MUI X, both to find an implementation that works well, and to prove that it brings value, and that it deserves to be in the core. I think that it could be a follow-up PR. It's kind of detail, a lot less important than the previous point.

DanailH:
Alright, I added the last changes:
- add the `unstable_` prefix for the `replaceRows` API
- add a debounce to the `onFetchRows` callback
- add a warning in the docs to highlight that it's best to implement a debounce so you limit the number of requests since sends to their BE.

Regarding the problems with the demo - I guess it's a limitation with what we have in terms of the fake server. It needs additional polishing but that can be done later on in a clean PR. I will merge this PR as it states it is stable in terms of the lazy loading feature and the replace rows API.

Thank you @oliviertassinari for the detailed review and feedback.

packages/grid/x-data-grid/src/models/api/gridRowApi.ts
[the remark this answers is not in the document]
DanailH: I'm not sure about those as well. I agree it's probably best to move the 2 api's under 1 but that we can do once we start v6. For now it would be best to keep the existing one (`updateRows`) and add the new one as `unstable_replaceRows`.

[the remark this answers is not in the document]
flaviendelangle: > Ah right OK. So one updates a row ID, the other a row index. So maybe we could transpire this in their name grin:

Not exactly
One update the row based on the row id
The other update the row based on the row index

flaviendelangle (approved):

PR #5238, opened 2022-06-17
[DataGridPremium] Fix error with quick filter and grouping rows

### LINKED ISSUE #5234: [data grid] Quick filter is not working with row grouping
### Duplicates

### Latest version

### Current behavior 😯

I noticed that QuickFilter does not work with row grouping: when typing anything into the QuickFilter input, an error will be thrown saying: "Cannot destructure property 'value' of 'undefined' as it is undefined." and no filter will be applied.

I created a demo codesandbox starting with the basic example for row grouping from the documentation (https://mui.com/x/react-data-grid/row-grouping/) and added the toolbar with QuickFilter: https://codesandbox.io/s/rowgroupingbasicexample-with-quickfilter-oblz92?file=/demo.tsx

### Expected behavior 🤔

QuickFilter should filter elements also if row grouping is applied.

### Steps to reproduce 🕹

Steps:

1. Create a simple data grid with row grouping
2. Add a toolbar with QuickFilter
3. Type anything into the QuickFilter input and an error will be thrown

See demo: https://codesandbox.io/s/rowgroupingbasicexample-with-quickfilter-oblz92?file=/demo.tsx

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #5234

The quick filter is using `shouldApplyFilter` parameters which was creating errors in quick filter engine

I also observed that quick filtering is not adapted to the row grouping filter logic, so I modified the behavior of quick filter on autogenerated rows to make it a bit more intuitive


packages/grid/x-data-grid/src/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -239,18 +239,24 @@ export const buildAggregatedQuickFilterApplier = (
   );
 
   return (rowId, shouldApplyFilter) => {
+    const rowNode = apiRef.current.getRowNode(rowId);
+    if (rowNode?.isAutoGenerated) {
+      return true;
+    }
```
alexfauquette, line 245: The row grouping filter is expecting that autogenerated rows pass the filtering.

For normal filtering, it makes sense. If you set "director must contain David" and you groupes your table with "director" you don't want rows which are not a child of a director named "David"


For the quick filter, the logic is a bit different since the criterion is applied without specifying the logic.
So for the following structure
- david
	- 1
	- 2
	
The quick search "1" will return nothing because the grouping row "david" does not contain "1"
The quick search "David" will return nothing because the grouping row "david" does not contain row passing the quick filter

I made a fix saying that all grouping rows pass quick filtering assuming the user will quick search in the child more often than the grouping columns

packages/grid/x-data-grid/src/hooks/features/filter/gridFilterUtils.ts
```diff
@@ -239,18 +239,24 @@ export const buildAggregatedQuickFilterApplier = (
...
+
     const usedCellParams: { [field: string]: GridCellParams } = {};
+    const columnsFieldsToFilter: string[] = [];
 
     Object.keys(appliersPerColumnField).forEach((columnField) => {
       if (!shouldApplyFilter || shouldApplyFilter(columnField)) {
         usedCellParams[columnField] = apiRef.current.getCellParams(rowId, columnField);
+        columnsFieldsToFilter.push(columnField);
```
alexfauquette, line 253: The error comes from the use of `shouldApplyFilter`

the `usedCellParams[columnField]` was not set, if not necessary. But in the next block of code, I'm looping on `Object.keys(appliersPerColumnField)` to get the field

So I was calling `usedCellParams[columnField]` with some `columnField` not set

packages/grid/x-data-grid/src/hooks/features/filter/gridFilterUtils.ts
flaviendelangle: What should be the behavior on Tree Data ?
You have different filtering logic for Row Grouping and for Tree Data

alexfauquette: From what I understood, with tree data we only filter leaf rows and auto-generate rows to show the tree structure

So letting autogenerated rows pass the filter should not ba a problem

flaviendelangle: I think we could end up with all the children of an auto generated group filtered-out and the group still visible
But I'm not sure

alexfauquette: I confirm it lets group without leafs 

https://codesandbox.io/s/treedatasimple-demo-mui-x-forked-3u7zey?file=/demo.tsx

alexfauquette: I added a test to be sure not breaking the behavior the same way for tree

For the initial problem on row grouping, I just reverted the modification. It should be done in another PR. For now I don't know how to do it

flaviendelangle (commented):
OK to first fix the bug and then talk about behavior improvement

Can we create a follow up issue ?

alexfauquette:
Issue created: #5278

PR #5242, opened 2022-06-17
Update getting-started.md

dorobiro:
After trying to get the Date Picker working I found a command on Stack Overflow, which is not mentioned in the MUI Documentation! Please add it.
npm install --save date-fns
Reference: https://stackoverflow.com/questions/71037974/module-not-found-error-cant-resolve-date-fns-adddays-in-c-users

### DIFF
--- a/docs/data/date-pickers/getting-started/getting-started.md
+++ b/docs/data/date-pickers/getting-started/getting-started.md
@@ -36,7 +36,7 @@ First you have to install the adapter package for the date-library you want to u
 
 ```sh
 // date-fns
-npm install @date-io/date-fns
+npm install @date-io/date-fns --> then "npm install --save date-fns"
 // or for Day.js
 npm install @date-io/dayjs
 // or for Luxon


oliviertassinari:
@dorobiro Thanks for raising! Yes, this docs is completely off:

1. "First you have to install the adapter package for the date-library you want to use:" No, this doesn't make sense, the adapters are already dependencies:

https://github.com/mui/mui-x/blob/38ba9cad0586136db103404027053358f1c77034/packages/x-date-pickers/package.json#L46-L51

2. In "We currently support 4 different date-libraries", I'm surprised moment is last, it's still the most popular https://npm-stat.com/charts.html?package=date-fns&package=dayjs&package=luxon&package=moment&from=2019-06-17&to=2022-06-17.

3. In https://mui.com/x/react-date-pickers/date-picker/#jalali-calendar-system. We say "Install date-fns-jalali and use @date-io/date-fns-jalali". However, @date-io/date-fns-jalali also needs to be installed. It was raised in https://mui.zendesk.com/agent/tickets/1529.

alexfauquette:
> the adapters are already dependencies

I'm wondering if `date-io/xxx` should be in the dependencies. They could be optional dependencies since it does not make sense to have at the same time `@date-io/date-fns`, `@date-io/dayjs`, `@date-io/luxon` and `@date-io/moment`.

If we move them to optional, the instruction should be to install the `@date-io` package and the library itself

> I'm surprised moment is last

They are in alphabetical order

oliviertassinari:
> I'm wondering if date-io/xxx should be in the dependencies.

@alexfauquette I think that the thought initially was that they are not bundled nor singletons, we could improve the DX by controlling the exact version used installed and reducing then number of steps to get started.

alexfauquette:
@dorobiro I tried to take your feedback into account when doing #5293 ([doc preview](https://deploy-preview-5293--material-ui-x.netlify.app/x/react-date-pickers/getting-started/))

Do not hesitate to reopen this PR or open an issue if you still see unclear parts

PR #5110, opened 2022-06-03
[DataGridPro] Refresh detail panel caches when props change

m4theushw:
Fixes #4784

Before: https://codesandbox.io/s/basicdetailpanels-material-demo-forked-t604oh?file=/demo.tsx (CodeSandbox from Andrew)
After: https://codesandbox.io/s/basicdetailpanels-material-demo-forked-7b5ujk?file=/demo.tsx

Previously, the cache for the content and height of the detail panels was always calculated when `getDetailPanelContent` or `getDetailPanelHeight` is changed, but the recalculation didn't trigger a new rows hydration. As consequence, the cache was updated but the height of the panel was the same, unless the panel is closed and reopened. This PR changes to trigger a rows hydration, and also a cache recalculation, when any the props is changed.

PR #5264, opened 2022-06-20
[data grid] Fix detail panel on DataGridPremium

### LINKED ISSUE #5262: [DataGrid] MasterDetail grid inside tab throws error on tab switch
### Duplicates

### Latest version

### Current behavior 😯

When the DataGrid's MasterDetail is used inside Tabs, the page will crash when trying to access the tab containing the master detail.
The following error is thrown when switching from any tab to the tab with the master detail grid:
**`apiRef.current.getRowParams is not a function`**

With "regular" grids, this error doesn't occur.

I created a sandbox to reproduce this error: https://codesandbox.io/s/masterdetailgrid-tabs-7tlzx7

(The MasterDetailGrid component content was taken from the first example on https://mui.com/x/react-data-grid/master-detail/, the simple DataGrid example from https://mui.com/x/react-data-grid/getting-started/)

### Expected behavior 🤔

When switching between tabs, grids containing master details should be rendered correctly.

### Steps to reproduce 🕹

Demo: https://codesandbox.io/s/masterdetailgrid-tabs-7tlzx7
Steps:
1. switch from any tab to the "MasterDetail Grid" tab
2. the error will be thrown

### Context 🔦

I am using various grids in tabs (which don't cause any problems) and in one specific case, I need a grid containing master detail. When switching from any tab to the master detail grid tab, an error will be thrown which causes the whole page to crash.

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #5262

In #5110 the plugin order was only changed for the pro plan

### DIFF
--- a/packages/grid/x-data-grid-premium/src/DataGridPremium/useDataGridPremiumComponent.tsx
+++ b/packages/grid/x-data-grid-premium/src/DataGridPremium/useDataGridPremiumComponent.tsx
@@ -118,11 +118,11 @@ export const useDataGridPremiumComponent = (
   useGridTreeData(apiRef);
   useGridKeyboardNavigation(apiRef, props);
   useGridSelection(apiRef, props);
-  useGridDetailPanel(apiRef, props);
   useGridColumnPinning(apiRef, props);
   useGridColumns(apiRef, props);
   useGridRows(apiRef, props);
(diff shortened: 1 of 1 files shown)


m4theushw (approved):
Oops, it happens.

PR #5267, opened 2022-06-20
[docs] New page for the action bar customization

flaviendelangle:


flaviendelangle:
@alexfauquette 

> What about using StatikDatePicker instead of StatikDateTimePicker for the demonstrations such that we can focus on the action bar without being distracted by the view switching between calender/clock?

Done

> About the second demo, the code is maybe over complicated. What about a CustomActionBar that displays a cancel/validate buttons made with <IconButton />

That's a recurrent topic
How complete and powerfull should the demo be ?
If we go for a simpler demo, I would at least add a note linking to our code to help the user building a real working example on application with several pickers that might have various actions.

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
+---
+product: date-pickers
+title: Date and Time pickers - Custom components
+---
+
+# Custom components
+
+<p class="description">The date picker let the user select a date.</p>
```
samuelsycamore, line 8:
```suggestion
<p class="description">The date picker lets users select a date from a menu.</p>
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Overriding components
+
+The picker components allows you to override internal components with the `components` prop.
```
samuelsycamore, line 12:
```suggestion
You can override the internal elements of the component (known as "slots") using the `components` prop.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+
+If you wish to pass additional props in a component slot, you can do it using the `componentsProps` prop.
```
samuelsycamore, line 14:
```suggestion
Use the `componentsProps` prop if you need to pass additional props to a component slot.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+
+As an example, you could override the action bar and pass additional props as below.
```
samuelsycamore, line 16:
```suggestion
As an example, you could override the `ActionBar` and pass additional props to the custom component as shown below:
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+  components={{
+    ActionBar: CustomActionBar,
+  }}
+  componentsProps={{
+    actionBar: ['clear'],
+  }}
+/>
+```
+
+:::info
```
samuelsycamore, line 30: Maybe this needs to be a more severe callout?

```suggestion
:::warning
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+The casing is different between the `components` (ActionBar) and `componentsProps` (actionBar) props.
```
samuelsycamore, line 31: To be more specific.

```suggestion
The `components` prop uses Pascal case (`ActionBar`), while `componentsProps` uses camel case (`actionBar`).
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+:::
+
+## Action bar
+
+### Component props
+
+The action bar is available on all the pickers component and is located at the bottom of the picker's views.
```
samuelsycamore, line 38:
```suggestion
The action bar is available on all picker components.
It is located at the bottom of the picker's views.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+By default, it contains no action on desktop and the action _Cancel_ and _Accept_ on mobile.
+
+You can override the action displayed using the `actions` prop passed to this component.
```
samuelsycamore, line 41: To be more specific and tell the reader exactly what to look for.

```suggestion
You can override the action displayed by passing the `actions` prop to the `actionBar` within `componentsProps`, as shown here:
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -0,0 +1,79 @@
...
+/>
+```
+
+In the example below, the action bar contains only one button to reset to today's date:
+
+{{"demo": "ActionBarComponentProps.js"}}
+
+#### Available actions
+
+The build-in `ActionBar` component supports 4 different actions:
```
samuelsycamore, line 63:
```suggestion
The built-in `ActionBar` component supports four different actions:
```

samuelsycamore:
> Maybe it would be nice to have a well written page shared by all our components (core and X) explaining the `components` and `componentsProps` instead.

This is something I've been thinking about as well (https://github.com/mui/material-ui/pull/33156#issuecomment-1161797141) as I'm polishing up the Base docs. For those pages, I think I'm going to document the usage of `components` and `componentsProps` for each component individually. I don't want to make the dev leave the page they're on to go learn a new concept before putting this into use. And once they see this on a few pages, I think most people will start to pick up the pattern intuitively. But I do also think we need a higher-level guide explaining how these work.

docs/data/date-pickers/custom-components/custom-components.md
flaviendelangle: We should probably do the same change on the data grid equivalent page when improving it then

alexfauquette (approved):

Issue #5109, opened 2022-06-03
[DatagridPro] How to save apiRef.current right before DataGrid component unmounts


KhachDavid:
### Order ID 💳

#44112

### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### The problem in depth 🔍

As the title says, my goal is to save the apiRef right before DataGridPro component unmounts in the demo linked below. Currently, I am using the fallback function of useEffect in a wrapper component. However, at the time when fallback is being called, apiRef.current is already null (see the console log). 

A solution, I have in mind is using listeners to keep track of the apiRef changes that I care about. A concern about this solution is that it might impact the performance, as callback listeners might be constantly called (e.g. columnWdithResizeListener). 

What might be a better way of keeping track of apiRef? 

[Demo](https://stackblitz.com/edit/react-ts-rce4tb?file=demo.jsx)

<details>
  <summary>Console Log</summary>
<img width="514" alt="image" src="https://user-images.githubusercontent.com/65775837/171881400-3819e486-f530-4a7d-b398-f527e2d67966.png">

</details

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

m4theushw:
The `apiRef` is null because it's destroyed even before the unmount function of the effect runs. As alternative I would suggest to not try to track when the grid is unmounted, but instead run the action you want before updating the state that unmounts the grid.

Example: https://stackblitz.com/edit/react-ts-2y46jb?file=demo.jsx,App.jsx

Basically we do:

```tsx
<Button 
  onClick={() => {
    if (isMounted) {
      // access apiRef.current
    }
    setMounted(!isMounted);
  })
/>
```

KhachDavid:
Hey Matheus, 

Thank you for the answer. I think this will work. Do you think this solution is scalable in a big project, when there are multiple buttons that can unmount the DataGrid wrapper component?

PR #5273, opened 2022-06-21
[docs] New page presenting the `apiRef`

### LINKED ISSUE #5269: [data grid] How to instanciate the api object
### Duplicates

### Latest version

### Current behavior 😯

```
const apiRef = useRef<GridApiPro>();

<DataGridPro apiRef={apiRef} ....... />
```
produces an TS error 
```
TS2322: Type 'MutableRefObject<GridApiPro | undefined>' is not assignable to type 'MutableRefObject<GridApiPro>'.   Type 'GridApiPro | undefined' is not assignable to type 'GridApiPro'.     Type 'undefined' is not assignable to type 'GridApiPro'. 
```

essentially, as far as i understand it is impossible to create the required type without the correct initial ref value or forcing it with casting.

### Expected behavior 🤔

apiRef prop type should be 
```
MutableRefObject<GridApiPro | undefined>
```

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Part of #1826 (12)
Fixes #5269

Doc preview: https://deploy-preview-5273--material-ui-x.netlify.app/x/react-data-grid/api-object/

- [x] Group the advanced pages in a doc folder (maybe other could go there too, I'd like to create a *Typescript* page in the future)
- [x] Create a new "API object" doc page presenting how to access the `apiRef` and linking to the basic examples

We could probably use this new page to link to other basic use cases.
For instance a section *Accessing the row model* inside one of the row pages and linked here would make a lot of sense.


docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Data Grid - API object
+---
+
+# Data grid - API object
+
+<p class="description">Interact with the grid using its api.</p>
```
m4theushw, line 7:
```suggestion
<p class="description">Interact with the grid using its API.</p>
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+
+## What is the API object ?
+
+The API object is an interface containing the state and all the methods needed to interact imperatively with the grid.
+
+You can find the list of all the available API methods on the [GridApi page](/x/api/data-grid/grid-api/).
+
+:::warning
+All the methods prefixed by `unstable_` are private.
+They should not be used by third party user and can be removed / renamed / reworked at any time.
```
m4theushw, line 17:
```suggestion
They should not be used by third party users and can be removed, renamed, reworked at any time.
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+:::
+
+## How to use the API object ?
+
+The API object is accessible through the `apiRef` variable.
+Depending on where you are trying to access this variable, you will have to use either `useGridApiContext` or `useGridApiRef`.
+
+### Use it inside the grid
+
+If you need to access the api object inside component slots or inside renders (e.g: `renderCell`, `renderHeader`),
```
m4theushw, line 30:
```suggestion
If you need to access the API object inside component slots or inside renders (e.g: `renderCell`, `renderHeader`),
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+you can use the `useGridApiContext` hook.
+
+```tsx
+const apiRef = useGridApiContext();
+
+<Button onClick={() => apiRef.current.setPage(1)}>Go to page 1</Button>;
+```
+
+:::info
+You don't have to initialize the API object using `useGridApiRef` to be able to it inside the grid components.
```
m4theushw, line 40: "you don't have to" seems like the user shouldn't initialize, but he still can.

```suggestion
It's not needed to initialize the API object using `useGridApiRef` to be able to use it inside the grid components.
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+
+{{"demo": "UseGridApiContext.js", "bg": "inline", "defaultCodeOpen": false}}
+
+### Use it outside the grid [<span class="plan-pro"></span>](https://mui.com/store/items/mui-x-pro/)
+
+When using the API object outside the grid components, you need to initialize it using the `useGridApiRef` hook.
+You can then pass it to the `apiRef` prop of the grid.
+
+```tsx
+const apiRef = useGridApiRef();
```
m4theushw, line 51: Maybe wrap this code in a function to give sense that it only works inside React components. Also, add the `return`.

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+
+<div>
+  <Button onClick={() => apiRef.current.setPage(1)}>Go to page 1</Button>
+  <DataGridPro apiRef={apiRef} {...other} />
+</div>;
+```
+
+:::warning
+The API object will be populated by the various plugins of the grid during the 1st render of our component.
+If you try to use it in the 1st render of your component, all the methods will be undefined.
```
m4theushw, line 61:
```suggestion
If you try to use it in the 1st render of your component, it will crash since all methods are not registered yet.
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+
+{{"demo": "UseGridApiRef.js", "bg": "inline", "defaultCodeOpen": false}}
+
+## Common use cases
+
+### Retrieve data from the state
+
+You can find a detailed example in the [State page](/x/react-data-grid/state/#access-the-state).
+
+### Listen to grid events
```
m4theushw, line 72: Another suggestion is https://github.com/mui/mui-x/issues/5109

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,81 @@
...
+<p class="description">Interact with the grid using its api.</p>
+
+## What is the API object ?
+
+The API object is an interface containing the state and all the methods needed to interact imperatively with the grid.
+
+You can find the list of all the available API methods on the [GridApi page](/x/api/data-grid/grid-api/).
+
+:::warning
+All the methods prefixed by `unstable_` are private.
```
joserodolfofreitas, line 16: I thought `unstable_` was used more in the direction of an `alpha` version of an API method.
Would it make sense to consider `private_` in these cases?

docs/data/data-grid/api-object/api-object.md
joserodolfofreitas, line 17: further iteration
```suggestion
We don't recommend using them as they can be removed, renamed, or reworked at any time.
```

docs/data/data-grid/api-object/api-object.md
joserodolfofreitas: Question: can the user use the `apiRef` inside the grid with `useGridApiRef`?
If that's the case, why not have a single hook?

joserodolfofreitas (commented):
A few thoughts:

docs/data/data-grid/api-object/api-object.md
flaviendelangle: `useGridApiRef` creates a new `apiRef` that must be passed to the grid
So if they you it, it will just return something that they can't pass to the grid (since they are inside the grid)

docs/data/data-grid/api-object/api-object.md
flaviendelangle: Right now it's the same prefix
And right now no `unstable_` is documented in a markdown file, even if theoretically we could indeed be used for features in alpha phase.

I suggest we leave it as is (with rewording if you want).
And we discuss if it's worth splitting in two prefixes in another PR.

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
...
+The API object is accessible through the `apiRef` variable.
+Depending on where you are trying to access this variable, you will have to use either `useGridApiContext` or `useGridApiRef`.
+
+### Use it inside the grid
+
+If you need to access the API object inside component slots or inside renders (e.g: `renderCell`, `renderHeader`),
+you can use the `useGridApiContext` hook.
+
+```tsx
+const CustomFooter = () => {
```
oliviertassinari, line 34: What do you think about using the function syntax for component declarations? 

```suggestion
function CustomFooter() {
```

docs/data/data-grid/api-object/UseGridApiRef.js
```diff
@@ -0,0 +1,27 @@
...
+    rowLength: 100,
+    maxColumns: 6,
+  });
+
+  const apiRef = useGridApiRef();
+
+  const handleGoToPage1 = () => apiRef.current.setPage(1);
+
+  return (
+    <Stack spacing={2} sx={{ width: '100%' }} alignItems="flex-start">
```
oliviertassinari, line 20: We have 1 in https://deploy-preview-5273--material-ui-x.netlify.app/x/react-data-grid/rows/#UpdateRowsProp.tsx and the other demos. Would this look better?

```suggestion
    <Stack spacing={1} sx={{ width: '100%' }} alignItems="flex-start">
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Data Grid - API object
+---
+
+# Data grid - API object
+
+<p class="description">Interact with the grid using its API.</p>
+
+## What is the API object ?
```
oliviertassinari, line 9:
```suggestion
## What is the API object?
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
...
+
+:::warning
+All the methods prefixed by `unstable_` are private.
+We don't recommend using them as they can be removed, renamed, or reworked at any time.
+
+If you need to use a private method, please open a GitHub issue describing your use case.
+We will help you to achieve the same goal with public methods, or we will discuss making this specific method public.
+:::
+
+## How to use the API object ?
```
oliviertassinari, line 23:
```suggestion
## How to use the API object?
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
...
+
+The API object is accessible through the `apiRef` variable.
+Depending on where you are trying to access this variable, you will have to use either `useGridApiContext` or `useGridApiRef`.
+
+### Use it inside the grid
+
+If you need to access the API object inside component slots or inside renders (e.g: `renderCell`, `renderHeader`),
```
oliviertassinari, line 30:
```suggestion
If you need to access the API object inside component slots or inside renders (e.g.: `renderCell`, `renderHeader`),
```

docs/data/data-grid/api-object/api-object.md
flaviendelangle: What would you like to document exactly here ?

docs/data/data-grid/api-object/api-object.md
samuelsycamore, line 23:
```suggestion
## How to use the API object
```

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
...
+```tsx
+const CustomFooter = () => {
+  const apiRef = useGridApiContext();
+
+  return <Button onClick={() => apiRef.current.setPage(1)}>Go to page 1</Button>;
+};
+```
+
+:::info
+It's not needed to initialize the API object using `useGridApiRef` to be able to use it inside the grid components.
```
samuelsycamore, line 42:
```suggestion
You don't need to initialize the API object using `useGridApiRef` to be able to use it inside the grid components.
```

docs/data/data-grid/api-object/api-object.md
flaviendelangle: Would you agree to add a section on the `State` page in a follow up PR and link it here ?

m4theushw: Yes.

docs/data/data-grid/api-object/UseGridApiRef.js
m4theushw: It's still 2.

image

docs/data/data-grid/api-object/api-object.md
```diff
@@ -0,0 +1,87 @@
...
+## What is the API object
+
+The API object is an interface containing the state and all the methods available to programmatically interact with the grid.
+
+You can find the list of all the API methods on the [GridApi page](/x/api/data-grid/grid-api/).
+
+:::warning
+All the methods prefixed by `unstable_` are private.
+We don't recommend using them as they can be removed, renamed, or reworked at any time.
+
```
m4theushw, line 18: Paragraphs inside callouts have `margin: 0`. I found weird to not have any space between them, so better to keep inside the same paragraph.

```suggestion
```

PR #5279, opened 2022-06-21
[DataGrid] Remove trailing spaces in filter input

### LINKED ISSUE #5275: [data grid] Filter value with trailing spaces
### Duplicates

### Latest version

### Current behavior 😯

Currently, filter value with trailing spaces doesn't remove the trailing spaces like below.

1. "sampl3e " passed as "sample " and not "sample"
2. "sample filter value" passed as "sample filter value" which is the expected behavior

### Expected behavior 🤔

Filter value for string should removed the trailing space when filtering.

1. "sample " should pass "sample" as value
2. "sample filter value" passed as "sample filter value" which is the expected behavior

### Steps to reproduce 🕹

Steps:

1. Choose any string filter
2. try to put "sample " as the filter value
3. The filter value "sample " doesn't remove the trailing space
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

33625

alexfauquette:
Fix #5275

### DIFF
--- a/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputValue.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterInputValue.tsx
@@ -109,6 +109,9 @@ function GridFilterInputValue(props: GridTypeFilterInputValueProps & TextFieldPr
       setIsApplying(true);
       // TODO singleSelect doesn't debounce
       filterTimeout.current = setTimeout(() => {
+        if (type !== 'singleSelect') {
+          value = value.trim();
+        }
         applyValue({ ...item, value });
         setIsApplying(false);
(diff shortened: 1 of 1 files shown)


m4theushw (commented):
Wouldn't be better to trim inside the operator? If `filterModel.items.value` contains space it shows no result: https://codesandbox.io/s/controlledfilters-demo-mui-x-forked-84qzhq?file=/demo.tsx

alexfauquette:
@m4theushw My concern was about how to disable the `trim` if I want to take into account the last space?

We can simply pass prop to the Input component, but accessing to the filtering engine seems a bit more complicated

m4theushw (approved):
Much cleaner now.

PR #5293, opened 2022-06-22
[docs] Improve getting-started page

### LINKED ISSUE #5290: [pickers] Usage with various calendars (Hijri, Hebrew, Chinese, ...)
### Duplicates

### Latest version

### Summary 💡

As a user who wants to use `DatePicker` I need to see more examples and clearer explanations.

For example, I want to use **hijri** calendar, but I don't want to use `moment.js` because it's an old library with old design and their official recommendation is that we should not use them for our new projects.

Based on your [current docs](https://mui.com/x/react-date-pickers/getting-started/) we should use `@date-io` to abstract adapters.

However, when I install `@date-io/hijri`, the project won't compile and needs `moment-hijri` which in turn might need `moment.js`. 

What should I do if I want to use `day.js` for the purpose of date mapping? 

This is not clear in your documentation.

### Examples 🌈

In the [current docs](https://mui.com/x/react-date-pickers/date-picker/#localization) there is only one example of how to localize MUI DatePicker for **Jalali** calendar. That's it. No more examples. What if we want to change it to Hebrew? How about the Chinese calendar?

What should we do if we can't find a calendar there? Where to search? What to do?

Please provide more examples there.

### Motivation 🔦

A website design company that works for many customers across the globe, needs to support different locales.

While it's true that a library like MUI can not possibly support all locales, it's required to be clear on its borders and scope.

For example, a simple table can be created to show the list of locales supported and for each one a simple example can be given.

This way, users won't be puzzled.

alexfauquette:
Fix #5290

main modifications:

- explain briefly how libraries interact
- replace @date-io installation by the date-library installation (https://github.com/mui/mui-x/pull/5242#issuecomment-1159544303)
- replace long cli instructions by an interactive one

before: https://mui.com/x/react-date-pickers/getting-started/
after: https://deploy-preview-5293--material-ui-x.netlify.app/x/react-date-pickers/getting-started/


docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -21,31 +21,38 @@ packageName: '@mui/x-date-pickers'
...
-We currently support 4 different date-libraries:
+You need to install 3 different types of package to make the pickers work:
 
-- [date-fns](https://date-fns.org/)
-- [Day.js](https://day.js.org/)
-- [Luxon](https://moment.github.io/luxon/#/)
-- [Moment.js](https://momentjs.com/)
+1. **The component** (`@mui/x-date-pickers` or `@mui/x-date-pickers-pro`) which manages the rendering.
+2. **The date-library** ([`moment`](https://momentjs.com/), [`dayjs`](https://day.js.org/), ...) which manages the date manipulation.
+3. **The adapter** ([`@date-io`](https://github.com/dmtrKovalenko/date-io#projects)) expose your favorite **date-library** under a unified api used by **component**.
```
flaviendelangle, line 30:
```suggestion
3. **The adapter** ([`@date-io`](https://github.com/dmtrKovalenko/date-io#projects)) which exposes your favorite **date-library** under a unified api used by **component**.
```

Or remove the `which` from the other two ?

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -21,31 +21,38 @@ packageName: '@mui/x-date-pickers'
...
+- [Luxon](https://moment.github.io/luxon/#/) adapted by `@date-io/luxon`.
+- [Moment.js](https://momentjs.com/) adapted by `@date-io/moment`.
+
+If you need to use `js-joda`, `date-fns-jalali`, `moment-jalaali`, or `moment-hijri` library, you should be able to find the corresponding date-library from [`@date-io`](https://github.com/dmtrKovalenko/date-io#projects).
+In such a case, you will have to install both the _date-library_ and the corresponding `@date-io` adapter.
+
+```jsx
+// To use moment-jalaali
+yarn add moment-jalaali
+yarn add @date-io/jalaali
```
flaviendelangle, line 49: Maybe add a section to fully cover https://github.com/mui/mui-x/issues/5290 saying that if date-io does not contain any adapter for your date library, you can create one and the date-io team will help you ?

flaviendelangle (commented):
A lot better than today !

docs/data/date-pickers/getting-started/getting-started.md
alexfauquette: I added the section

oliviertassinari:
The layout is a bit broken on mobile.

oliviertassinari:
> Fix #5290

Are we sure? I understood the issue as the user doesn't want to use moment.

alexfauquette:
From the answer, I assume that adding explanations in the doc fix the pain of the user

> @flaviendelangle, your explanation is exactly what I mean to be added to MUI docs.
>
>Your explanation made it much clearer. You can even extend it. Now I have a better understanding of what should I search for.

flaviendelangle (approved):
We will be able to improve it later based on feedback, but it's already 100x better than the existing

Nefcanto:
@alexfauquette, thank you so much. 

I read the improved docs and now I can understand the architecture of date-time handling in MUI. 

Still, a couple of demos would be great. But this is good enough. 

Now everything is clear and thank you so much for this clarification.

oliviertassinari:
@alexfauquette OK great.

@Nefcanto :+1: it could make sense to open a new issue about the Chinese calandar. If we want to attract this audience, it should like a good opportunity to communicate them that we give it value, and to more easily test that it works as expected.

Nefcanto:
@oliviertassinari it would be awesome. But I'm not from China. I'm from Iran.

However, I would gladly help. I can create the issue, but in English. And I would test them in my own applications.

Please let me know if I should open the issue or not.

PR #5373, opened 2022-07-02
[DatePicker] Customize day formatter in the calendar

### LINKED ISSUE #4605: [DatePicker] Customize day headers
I'm trying to customize the DatePicker designs as close as possible to my requirements. We are using 3 letters(mon - tue - wed...) to describe our week days but I can't see any option where I can customize the default headers(M - T - W...).

Screen Shot 2020-05-22 at 19 09 45

### LINKED ISSUE #5222: [pickers] dateFormats doesn't work on weekday
### Duplicates

### Latest version

### Current behavior 😯

This is the current weekdays text in Chinese on date picker.
1971655362607_ pic
 I try to use dateFormats, but it doesn't work.
1981655362657_ pic

### Expected behavior 🤔

I want the weekdays text can be displayed as “ 日 一 二 三 四 五 六" or “ 一 二 三 四 五 六 日" , depending on the start day of the week. In fact, most of my users want the latter pattern.
Thank you very much!

### Steps to reproduce 🕹

Codesandbox : https://codesandbox.io/s/nervous-lalande-uj9h5s?file=/src/App.js

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

### LINKED ISSUE #5330: Week days name in jalali calendar
### Duplicates

### Latest version

### Current behavior 😯

Hi
As you see week day name use of day number with ش character which in the datepicker just display first letter of the week day name.

image

### Expected behavior 🤔

I want to config the `@date-io/date-fns-jalali` for change default week day names to full name. for example convert:

- ش to شنبه
- ۱ to یکشنبه
- ۲ to دوشنبه
.
.
.

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
partial for #5222 (`dateFormat.weekdayShort` is not applied, but that's because of the date-io behavior)
Fix #4605
Fix #5330

### Open question

Maybe we should add a second argument, for example, the column index such that the user can customize it without having to know what is `utils.getWeekdays()`

Also wondering, if we should break the behavior and stop formating with `.charAt(0).toUpperCase()` which is kind of breaking the i18n. For example chines use two characters meaning week + number od the day in the week, so applying the formater only keeps the character meaning "week"

### TODO
- [x] Document how to format the date ([preview](https://deploy-preview-5373--material-ui-x.netlify.app/x/react-date-pickers/localization/#advanced-customization))
- [x] Add test to verify the calendar is updated

### Remaining problem with date-io

We allow to pass a configuration object `dateFormats` which allows specifying the format of each date display of date-io `weekday`, `weekdayShort`, etc. (see #5222 for example)
But this seems to not be considered by `getWeekdays` because each library adapter has it's own behavior (see example bellow). I assume this might be fixed in date-io instead of mui-x

```js
// for moment.js
public getWeekdays = () => {
  return this.moment.weekdaysShort(true);
};

// for date-fns
public getWeekdays = () => {
  const now = new Date();
  return eachDayOfInterval({
    start: startOfWeek(now, { locale: this.locale }),
    end: endOfWeek(now, { locale: this.locale }),
  }).map((day) => this.formatByString(day, "EEEEEE"));
};
```


packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -43,6 +43,13 @@ export interface ExportedDayPickerProps<TDate>
    * @default () => "..."
    */
   renderLoading?: () => React.ReactNode;
+  /**
+   * Format the day name displayed in the calendar header
+   * @param {string} day The day name provided by the adapter's method `getWeekdays`
+   * @returns {string} The name to display
```
m4theushw, line 49:
```suggestion
   * Formats the day of week displayed in the calendar header.
   * @param {string} day The day of week provided by the adapter's method `getWeekdays`.
   * @returns {string} The name to display.
```

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -43,6 +43,13 @@ export interface ExportedDayPickerProps<TDate>
...
+   * @default (day) => day.charAt(0).toUpperCase()
+   */
+  headerDayFormatter?: (day: string) => string;
```
m4theushw, line 52: The name of the days are rendered in the header but we may allow users to change this or even repeat them in the footer too, so I would avoid "header". Other props that accept function have the "get" prefix.

```suggestion
  getDayOfWeekFormatter?: (day: string) => string;
```

alexfauquette: I applied the modification but used `dayOfWeekFormatter` because the idea that this prop is a function is already implied by the `Formatter` suffix. Like we say `valueFormatter` and not `getValueFormatter` in the data grid

PR #5390, opened 2022-07-04
[DataGrid] Avoid inconsistent state export

### LINKED ISSUE #4620: [data grid] Export state with hidden columns
### Duplicates

### Latest version

### Current behavior 😯

When there is hidden column, the apiRef.current.exportState() does not include columnVisibilityModel.
It only works when there is at least one hidden column added to the initialState.

### Expected behavior 🤔

Hide columns should be exported and restored correctly without initialState.

### Steps to reproduce 🕹

Steps:

1. go to https://mui.com/x/react-data-grid/state/#save-and-restore-the-state
2. click on any three dots at the first table header
3. go to "show columns"
4. click "show all"
5. click "recreate the 2nd grid"
6. the ID column is not added to the second grid

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

### LINKED ISSUE #4779: [data grid] restoreState doesn't always have same result when passing an exportedState
### Duplicates

### Latest version

### Current behavior 😯

When using exportState then passing the returned value to the restoreState function, the resulting DataGrid isn't always the same as when exported.

Currently when you export the state of a table that has all columns visible the ColumnVisibilityModel is not defined as an attribute of the exported State. When performing a restoreState with the returned export, it does not update the visible columns because ColumnVisibilityModel isn't defined unless you manually add it as an empty object in the state structure. 

**All visible**
image

**Columns Hidden**
image

**All selected but remain hidden**
image

### Expected behavior 🤔

When restoring to a state that has all columns visible, it should always restore all columns as visible.
The ColumnVisibilityModel should always be returned when the state is exported, even as an empty object, or the restoreState should interpret an undefined ColumnVisibilityModel to mean all columns are visible.

### Steps to reproduce 🕹

Steps: This can be performed on the website at [Restore The State With Apiref]( https://mui.com/x/react-data-grid/state/#restore-the-state-with-apiref
 )

1. exportState of a DataGrid with all columns visible
2. Hide some columns on the table
3. use restoreState and pass in the previously exported state
4. columns remain hidden

### Context 🔦

I'm attempting to do a multi-view table

### Your environment 🌎

<details>

 System:
    OS: Windows 10 10.0.19043
  Binaries:
    Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1266.0), Chromium (101.0.1210.32)
  npmPackages:
    @emotion/react: ^11.7.1 => 11.9.0 
    @emotion/styled: ^11.6.0 => 11.8.1 
    @mui/base:  5.0.0-alpha.78 
    @mui/icons-material: ^5.3.1 => 5.6.2 
    @mui/lab: ^5.0.0-alpha.67 => 5.0.0-alpha.79 
    @mui/material: ^5.4.0 => 5.6.3 
    @mui/private-theming:  5.6.2 
    @mui/styled-engine:  5.6.1 
    @mui/styles: ^5.3.0 => 5.6.2 
    @mui/system:  5.6.3 
    @mui/types:  7.1.3 
    @mui/utils:  5.6.1 
    @mui/x-data-grid: ^5.10.0 => 5.10.0 
    @mui/x-data-grid-pro: ^5.10.0 => 5.10.0
    @mui/x-license-pro:  5.10.0
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.1 => 17.0.2
    styled-components: ^5.2.1 => 5.3.3
    typescript: ^4.1.3 => 4.5.4
</details>

### Order ID 💳 (optional)

#39752

### LINKED ISSUE #5211: [DataGridPremium] Row grouping column does not dissapear when restoring state.
### Duplicates

### Latest version

### Current behavior 😯

When grouping by a certain column and then restoring the state to a previously exported state that does not include the grouping column the, grouping column is pushed to the end of the grid. 

### Expected behavior 🤔

If I export a state with 2 columns which are not grouped then the state should restore it to these 2 columns regardless of the current state.

### Steps to reproduce 🕹

Steps: This is a copy from the mui state export and restore demo
https://codesandbox.io/s/inspiring-tess-z5qr34?file=/src/App.tsx

1. Create a view with all cols
2. Create a view with just 2 cols
3. Group the rows by Commodity
4. Select one of the previously created views 
5. Look at the last column and the grouped column will be there.

### Context 🔦

I'm attempting to do a multi-view table. 

This is a similar issue to [#4779 ](https://github.com/mui/mui-x/issues/4779)

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #5211
Fixes #4779
Fixes #4620

## Behavior currently in the PR

By default, we export all the models.

```tsx
apiRef.current.exportState();
```

And there is a new `exportOnlyUsedModels` prop that only exports the following:

- initialized (in `props.initialState`)
- controlled 
- modified (the value when exporting does not equal the one passed to the state on mount)

```tsx
apiRef.current.exportState({ exportOnlyUsedModels: true });
```

It is a behavior change but I would not consider it as a breaking change.
For me not exporting everything by default can be considered as a broken behavior.

### New behavior for exported states

The new built-in behavior is to export all the models that are either:
- initialized (in `props.initialState`)
- controlled 
- modified (the value when exporting does not equal the one passed to the state on mount)

### Limits

We have on scenario where the behavior on `master` and the behavior of this PR don't work:

1. Display a grid with a model not controlled or initialized (for instance `filterModel`)
2. Export the state
4. Add a filter item 
5. Restore the state

The filter item won't go away since the empty model was not exported in 2.

You can test it [on master](https://codesandbox.io/s/data-grid-premium-forked-v4eos2) or [on this branch](https://codesandbox.io/s/data-grid-premium-forked-4udesl), just click on the button from left to right.

### Implemented solution

I added an option `shouldExportUnusedModels` (name to be discuss) to switch to a mode where we export everything ([same example with this option](https://codesandbox.io/s/data-grid-premium-forked-b12s54))

### Alternate solutions

I am honestly wondering if we should not always export everything.
We have all these bugs and discussions before we want to be smart on what is exported.
But if a model is skipped, it's because it is empty so it would probably not be a big deal to export it anyway.

Alternatively, we could export `null` to say "pick the default value" when restoring.

## Tasks

- [x] Introduce a new param to force exporting all the states even if untouched `apiRef.current.exportState({ shouldExportUnusedModels: true })`

- [x] Always have the same export behavior across models (always export if the model is controlled / initialized or modified)

- [x] Write additional tests

- [ ] Improve doc page


alexfauquette:
I like the introduction of `shouldExportUnusedModels`

I think the logic should go in the opposite. By default export all, and allow to restrict with `exportOnlyUsedModels`. Because it would provide a save/restore working out of the box, and the possibility to customize for more advanced use cases

flaviendelangle:
> I think the logic should go in the opposite. By default export all, and allow to restrict with exportOnlyUsedModels. Because it would provide a save/restore working out of the box, and the possibility to customize for more advanced use cases

Would you keep the same cherry picking logic ? 

```ts
      const shouldExportFilterModel =
        // Always export if the `shouldExportUnusedModels` property is activated
        context.shouldExportUnusedModels ||
        // Always export if the model is controlled
        props.filterModel != null ||
        // Always export if the model has been initialized
        props.initialState?.filter?.filterModel != null ||
        // Export if the model is not equal to the default value
        !isDeepEqual(filterModelToExport, getDefaultGridFilterModel());
```

alexfauquette:
Yes, it would be a good starting point to get feedbacks from users about what they would like to not export

flaviendelangle:
I applied the logic and added tests
Just have to add docs if we agree on the behavior

packages/grid/x-data-grid-pro/src/tests/statePersistence.DataGridPro.test.tsx
```diff
@@ -112,11 +113,33 @@ describe('<DataGridPro /> - State Persistence', () => {
...
+        preferencePanel: {
+          open: false,
+        },
+        sorting: {
+          sortModel: [],
+        },
+      });
+    });
+
+    it('should export not the default values of the models when using exportOnlyUsedModels', () => {
```
m4theushw, line 140:
```suggestion
    it('should not export the default values of the models when using exportOnlyUsedModels', () => {
```

packages/grid/x-data-grid-pro/src/tests/statePersistence.DataGridPro.test.tsx
```diff
@@ -128,6 +151,57 @@ describe('<DataGridPro /> - State Persistence', () => {
...
+              dimensions: FULL_INITIAL_STATE.columns?.dimensions,
+            },
+            preferencePanel: FULL_INITIAL_STATE.preferencePanel,
+          }}
+        />,
+      );
+      expect(apiRef.current.exportState()).to.deep.equal(FULL_INITIAL_STATE);
+    });
+
+    it('should export the controlled values of the models when using exportOnlyUsedModels', () => {
```
m4theushw, line 176: `exportOnlyUsedModels` is not used here. Was that the intention?

```suggestion
    it('should export only the controlled values of the models when not using exportOnlyUsedModels', () => {
```

m4theushw (approved):
Instead of `exportOnlyUsedModels`, what about `exportOnlyDirtyModels`? [Angular](https://angular.io/guide/form-validation) used this term for form input modified by the user.

packages/grid/x-data-grid-pro/src/tests/statePersistence.DataGridPro.test.tsx
flaviendelangle: The goal was to check that if we pass some controlled models, they are correcftly exported
I'm not puttin "only" because I'm passing all possible controlled model (to test that all of them are exported).
But yes I forgot to pass `exportOnlyUsedModels`, fixed

PR #5397, opened 2022-07-05
[pickers] Should fallback on default minDate / maxDate when null is passed

### LINKED ISSUE #5190: [pickers] Broken behavior when minDate or maxDate set to null
### Duplicates

### Latest version

### Current behavior 😯

When `minDate` or `maxDate` are `null` the picker exhibits odd behavior as opposed to when they are set to `undefined` or not provided.

The month picker component will be broken and the back or forward arrows will be disabled.

### Expected behavior 🤔

Setting `minDate` or `maxDate` to `null` should have the same behavior as setting `undefined` or  not suppling the props.

### Steps to reproduce 🕹

Steps:

1. Render DatePicker
2. Set minDate or maxDate as null
3. Picker will behave as if it has a min or max Date and the month picker will not be functional.

You can find a codesanbox example [here](https://codesandbox.io/s/basicdatepicker-material-demo-forked-oru8f2?file=/demo.js)

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
 System:
    OS: macOS 12.4
  Binaries:
    Node: 16.14.2 - ~/.asdf/installs/nodejs/16.14.2/bin/node
    Yarn: 1.22.17 - ~/.asdf/shims/yarn
    npm: 8.5.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 102.0.5005.115
    Edge: Not Found
    Firefox: 101.0.1
    Safari: 15.5
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #5190

- Support `null` as a valid value for `minDate` / `maxDate` / `minTime` / `maxTime`
- Improve typing of defaultized props hooks
- Use defaultized props hooks pattern on some other components


packages/x-date-pickers/src/internals/hooks/validation/useDateValidation.ts
```diff
@@ -25,14 +26,16 @@ export const validateDate: Validator<any, DateValidationProps<any, any>> = ({
...
   const date = adapter.utils.date(value);
-
-  const {
-    shouldDisableDate,
-    minDate = adapter.defaultDates.minDate,
-    maxDate = adapter.defaultDates.maxDate,
-    disableFuture,
-    disablePast,
-  } = props;
+  const minDate = parsePickerInputValueWithDefault(
```
alexfauquette, line 29: This value should already be normalised, because hooks are called after `use[...]PickerDefaultizedProps`

alexfauquette (commented):
The logic looks good

I'm wondering if we could unify all that with a wrapper that takes as an input the parameters of `useDateValidation` and provide an API `firstValidDate`, `firstValidMonth`, `isValidDate()`, ... such that all the validation logic is isolated and do not need to be  propagated with props (it's an idea for v6 not for this PR of course)

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
```diff
@@ -444,7 +477,6 @@ const CalendarPicker = React.forwardRef(function CalendarPicker<TDate>(
               reduceAnimations={reduceAnimations}
               selectedDays={[date]}
               onSelectedDaysChange={onSelectedDayChange}
-              loading={loading}
```
alexfauquette, line 447: What's the reason for removing `loading`?

packages/x-date-pickers/src/YearPicker/YearPicker.tsx
```diff
@@ -21,6 +23,24 @@ const useUtilityClasses = (ownerState: any) => {
...
+  const themeProps = useThemeProps({
+    props,
+    name,
+  });
+
+  return {
+    ...themeProps,
+    minDate: parseNonNullablePickerDate(utils, themeProps.minDate, defaultDates.minDate),
+    maxDate: parseNonNullablePickerDate(utils, themeProps.maxDate, defaultDates.maxDate),
+  };
```
alexfauquette, line 41: No `disablePast: false, disableFuture: false,`?

alexfauquette (commented):
I will read it a second time to take care about the types

LukasTy (approved):
LGTM 👍

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
LukasTy: It's not used anywhere else in the component and is not coming from `...other` where all the defaultized props live.

packages/x-date-pickers/src/YearPicker/YearPicker.tsx
flaviendelangle: Indeed, added

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
alexfauquette: I was looking for `{...other}` at the end of the props 😅

Maybe could do the same with `renderLoading` such that they are both handled the same way

packages/x-date-pickers/src/MonthPicker/MonthPicker.tsx
```diff
@@ -48,6 +55,24 @@ const useUtilityClasses = (ownerState: MonthPickerProps<any>) => {
...
+  name: string,
+): DefaultizedProps<MonthPickerProps<TDate>, 'minDate' | 'maxDate'> {
+  const utils = useUtils<TDate>();
+  const defaultDates = useDefaultDates<TDate>();
+  const themeProps = useThemeProps({
+    props,
+    name,
+  });
+
+  return {
```
alexfauquette, line 69:
```suggestion
  return {
      disableFuture: false,
      disablePast: false,
```

packages/x-date-pickers/src/YearPicker/YearPicker.tsx
```diff
@@ -21,6 +23,26 @@ const useUtilityClasses = (ownerState: any) => {
...
+    props,
+    name,
+  });
+
+  return {
+    ...themeProps,
+    minDate: parseNonNullablePickerDate(utils, themeProps.minDate, defaultDates.minDate),
+    maxDate: parseNonNullablePickerDate(utils, themeProps.maxDate, defaultDates.maxDate),
+    disablePast: false,
+    disableFuture: false,
```
alexfauquette, line 42: Such that theme  Props can override the `disableXxx`
```suggestion
  return {
    disablePast: false,
    disableFuture: false,
    ...themeProps,
    minDate: parseNonNullablePickerDate(utils, themeProps.minDate, defaultDates.minDate),
    maxDate: parseNonNullablePickerDate(utils, themeProps.maxDate, defaultDates.maxDate),
```

alexfauquette (approved):

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
flaviendelangle: Indeed, `renderLoading` is only used in `DayPicker` so the default value is useless here

flaviendelangle: Done

flaviendelangle:
@alexfauquette if you have other feedbacks on the typing don't hesitate, I will open a new PR since it's 100% internal

PR #5399, opened 2022-07-05
[docs] Add warning clarifications

alexfauquette:
I struggled to understand a warning in [row grouping documentation](https://mui.com/x/react-data-grid/row-grouping/#hide-the-grouped-columns) about `useKeepGroupedColumnsHidden`

> This hook is only compatible with the deprecated column property `hide` or with the controlled `columnVisibilityModel` prop.
>
> You must use the `columnVisibilityModel` in the `initialState` instead.

Not sure my correction is exactly what the hook does

### DIFF
--- a/docs/data/data-grid/row-grouping/row-grouping.md
+++ b/docs/data/data-grid/row-grouping/row-grouping.md
@@ -142,9 +142,9 @@ return (
 ```
 
 :::warning
-This hook is only compatible with the deprecated column property `hide` or with the controlled `columnVisibilityModel` prop.
+This hook is not compatible with the deprecated column property `hide`, and you should not control the `columnVisibilityModel`.
 
-You must use the `columnVisibilityModel` in the `initialState` instead.
+To hide some columns, use the `columnVisibilityModel` in the `initialState` instead.
 :::
 
 {{"demo": "RowGroupingUseKeepGroupedColumnsHidden.js", "bg": "inline", "defaultCodeOpen": false}}
--- a/docs/data/data-grid/row-grouping/row-grouping.md
+++ b/docs/data/data-grid/row-grouping/row-grouping.md
@@ -142,9 +142,18 @@ return (
 ```
 
 :::warning
-This hook is not compatible with the deprecated column property `hide`, and you should not control the `columnVisibilityModel`.
+This hook is not compatible with the deprecated column property `hide`.
+It can be used with controlled `columnVisibilityModel` and `initialState`.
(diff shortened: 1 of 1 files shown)


docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -142,9 +142,9 @@ return (
 ```
 
 :::warning
-This hook is only compatible with the deprecated column property `hide` or with the controlled `columnVisibilityModel` prop.
+This hook is not compatible with the deprecated column property `hide`, and you should not control the `columnVisibilityModel`.
```
flaviendelangle, line 145: Wow my sentence made no sense...
It should be compatible with the controlled `columnVisibilityModel` prop if you also pass its value to the hook.

See the `RowGroupingCustomGroupingColDefCallback` example

docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -142,9 +142,18 @@ return (
 ```
 
 :::warning
-This hook is not compatible with the deprecated column property `hide`, and you should not control the `columnVisibilityModel`.
+This hook is not compatible with the deprecated column property `hide`.
+It can be used with controlled `columnVisibilityModel` and `initialState`.
+To do so, provide the controlled value and the initial state to the hook as follow:
```
flaviendelangle, line 147: You just have to provide one of the two, and the example below is redundant with what is currently existing

docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -91,6 +91,15 @@ For instance if you group by `"director"`, you will have two columns titled _Dir
...
 The hook will automatically hide the columns when added to the model and show them back when removed from it.
 
+:::warning
+This hook is not compatible with the deprecated column property `hide`.
+
+To manage column visibility with `columnVisibilityModel`, `initialState`, or both together, you must pass them to hook parameters.
+:::
+
+Bellow are two examples about how to use `columnVisibilityModel`, or `initialState` with `useKeepGroupedColumnsHidden` hook.
+You can mix the two examples to support both at the same time.
```
flaviendelangle, line 101: In which scenario would you use both an initial state and a control state ? The control state would totally override the initial state right ?

flaviendelangle: Oh ok, I thought you were talking about setting the same model in both control and initial mode

PR #5405, opened 2022-07-06
[DataGrid] Start editing with uppercase letter and Ctrl+V

### LINKED ISSUE #5228: Cannot start editing with Shift+[key] while using newEditingApi?
### Order ID 💳

43201

### Duplicates

### Latest version

### The problem in depth 🔍

According to the documentation, one can start editing by [Pressing any printable key, such as a, E, 0, or $](https://mui.com/x/react-data-grid/editing/#start-editing); but this doesn't work with the newEditingApi.  

While using `experimentalFeatures={{ newEditingApi: true }}`, click in a cell, type E (Shift+e) or $ (Shift+4).  Nothing happens.

### Your environment 🌎

<details>

```
  Don't forget to mention which browser you used:
Chrome 102.0.5005.63
Edge 102.0.1245.39

  Output from `npx @mui/envinfo`:
  System:
    OS: Linux 5.10 Ubuntu 20.04 LTS (Focal Fossa)
  Binaries:
    Node: 16.13.2 - ~/.nvm/versions/node/v16.13.2/bin/node
    Yarn: Not Found
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.2/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
```
</details>

alexfauquette:
Fix #5228

The editing can now be start with 

- <kbd>Shift</kbd> + a letter
- <kbd>Ctrl</kbd> + V

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.new.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.new.test.tsx
@@ -689,7 +689,7 @@ describe('<DataGridPro /> - Cell Editing', () => {
         expect(listener.callCount).to.equal(0);
       });
 
-      ['shiftKey', 'ctrlKey', 'metaKey', 'altKey'].forEach((key) => {
+      ['ctrlKey', 'metaKey', 'altKey'].forEach((key) => {
         it(`should not publish 'cellEditStart' if ${key} is pressed`, () => {
           render(<TestCase />);
           const listener = spy();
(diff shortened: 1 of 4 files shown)


DanailH (approved):
I guess this is mostly if you want to start with a capital letter. 👍

oliviertassinari:
I feel that we are introducing a regression in terms of UX cc @gerdadesign. See the current behavior:

https://user-images.githubusercontent.com/3165635/178032243-c720e496-def9-43a0-b8a1-0849289a32a7.mov

https://deploy-preview-5405--material-ui-x.netlify.app/x/react-data-grid/editing/#making-a-column-editable

But in all the other data grids I can benchmark against, <kbd>Ctrl</kbd> + <kbd>V</kbd> edits the cell **without** entering in edit mode. You can reproduce this on: Notion, AG Grid, Airtable. So, I would propose that we ignore the shortcut for now, delaying it for #199.

alexfauquette:
@oliviertassinari No problem for removing it. I added the <kbd>Ctrl</kbd> + <kbd>V</kbd> because I felt it would be a quick win to be able to past something in a cell without having to start edit mode but it can wait the implementation of the clipboard feature

oliviertassinari:
On a related note, <kbd>⌥ (option)</kbd> + <kbd>p</kbd> that prints π does not work on this PR as well as before this PR, nor on AG Grid, but it works on Excel. 

Maybe this would do it:

```diff
diff --git a/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts b/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts
index 243ce0d8..8e984af2 100644
--- a/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts
+++ b/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts
@@ -152,11 +152,7 @@ export const useGridCellEditing = (
         let reason: GridCellEditStartReasons | undefined;

         if (isPrintableKey(event.key)) {
-          if (
-            (event.ctrlKey && event.key !== 'v') ||
-            (event.metaKey && event.key !== 'v') ||
-            event.altKey
-          ) {
+          if (event.ctrlKey || event.metaKey) {
             return;
           }
           reason = GridCellEditStartReasons.printableKeyDown;
```

I guess this is the kind of logic that is painful to get right, full of hedge cases. +1 to add a test for event.altKey = true, and event.key = 'π' that is Option + P on macOS.

alexfauquette:
@m4theushw Do you know what is the original reason of this condition?

The only problematic key combo I can imagine is <kbd>Ctrl</kbd>+<kbd>C</kbd> because when user select a box content and copy it, they don't want to go to edit mode. 

Otheriwise mot of the combinations I know are to write spécial letters which could start the édit mode. For the <kbd>Shift</kbd> we should keep ignoring it. For editing with capital letters but also for users who do not have a num pad and try to [edit number columns](https://mui.com/x/react-data-grid/editing/#making-a-column-editable)

m4theushw:
> Do you know what is the original reason of this condition?

@alexfauquette I think that the condition to not start the edit mode was added mainly because of <kbd>Ctrl</kbd> + <kbd>a</kbd>. But then it also blocked other key combinations.

> But in all the other data grids I can benchmark against, Ctrl + V edits the cell without entering in edit mode. You can reproduce this on: Notion, Excel, Google Spreadsheet, AG Grid, and Airtable. It breaks the learnings end-users have built on these tools.

@oliviertassinari If we allow users to paste any content without entering the edit mode, then we won't be able to validate the new data entered which may lead to inconsistencies. For instance, AG Grid has this problem too, if I try to copy from the Name column and paste into the Country column, the content gets messed up because it doesn't have the country flag to show.

oliviertassinari:
> we won't be able to validate the new data entered which may lead to inconsistencies.

@m4theushw Interesting. My issue as an end-user is that in the very large majority of the cases when I use <kbd>Ctrl</kbd>+<kbd>V</kbd>, I don't want to further edit the value. Having to systematically press <kbd>Escape</kbd> is a bit painful.

Issue #4687, opened 2022-04-28
[data grid] Broken "start editing" integration with Japanese


vuduy97:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

I create a table that allows to edit cells but have problem entering values, if you don't enter edit mode before entering text but only focus on the cell you want to edit and enter values ​​too, i have problem using keyboard japanese, the first value doesn't get the value of the language I use, it only accepts the value of the keyboard that I press

https://user-images.githubusercontent.com/72425044/165739433-82a47833-cb41-4960-afbd-31bcdcefb93f.mp4

### Expected behavior 🤔

I would like to be able to get the first value which is the one using the keyboard type i use

https://user-images.githubusercontent.com/72425044/165739663-25463ae0-5824-48b5-b10c-cfc21bada40f.mp4

### Steps to reproduce 🕹

Steps:

1.focus the cell you want edit (not enter the edit mode in this cell)
2.change the keyboard in use to Japanese
3.enter input value (like ああああああ)

### Context 🔦

I get the value of　aああああ, instead of　あああああ

### Your environment 🌎

System:
    OS: Linux 5.13 Ubuntu 20.04.3 LTS (Focal Fossa)
  Binaries:
    Node: 14.17.1 - ~/.nvm/versions/node/v14.17.1/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v14.17.1/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v14.17.1/bin/npm
  Browsers:
    Chrome: 97.0.4692.71
    Firefox: 99.0
  npmPackages:
    @emotion/react: ^11.9.0 => 11.9.0 
    @emotion/styled: ^11.8.1 => 11.8.1 
    @mui/base: ^5.0.0-alpha.74 => 5.0.0-alpha.77 
    @mui/icons-material: ^5.6.0 => 5.6.2 
    @mui/lab: ^5.0.0-alpha.76 => 5.0.0-alpha.78 
    @mui/material: ^5.6.0 => 5.6.2 
    @mui/private-theming:  5.6.2 
    @mui/styled-engine:  5.6.1 
    @mui/styles: ^5.6.0 => 5.6.2 
    @mui/system:  5.6.2 
    @mui/types:  7.1.3 
    @mui/utils:  5.6.1 
    @mui/x-data-grid: ^5.7.0 => 5.9.0 
[...]

PR #1409, opened 2021-04-13
[DataGrid] Fix printable keys to match ag

dtassone:
Following issue mentioned in #1287

PR #5414, opened 2022-07-07
[DataGrid] Fix broken "start editing" integration with Japanese

mnajdova:
Fixes https://github.com/mui/mui-x/issues/4687

I also need this for fixing the build in https://github.com/mui/material-ui/pull/33196

### DIFF
--- a/packages/grid/x-data-grid/src/utils/keyboardUtils.ts
+++ b/packages/grid/x-data-grid/src/utils/keyboardUtils.ts
@@ -13,9 +13,6 @@ export const isHomeOrEndKeys = (key: string): boolean => key === 'Home' || key =
 export const isPageKeys = (key: string): boolean => key.indexOf('Page') === 0;
 export const isDeleteKeys = (key: string) => key === 'Delete' || key === 'Backspace';
 
-const printableCharRegex = /^(\p{L}|\p{M}\p{L}|\p{M}|\p{N}|\p{Z}|\p{S}|\p{P})$/iu;
-export const isPrintableKey = (key: string) => printableCharRegex.test(key);
-
 export const GRID_MULTIPLE_SELECTION_KEYS = ['Meta', 'Control', 'Shift'];
 export const GRID_CELL_EXIT_EDIT_MODE_KEYS = ['Enter', 'Escape', 'Tab'];
 export const GRID_CELL_EDIT_COMMIT_KEYS = ['Enter', 'Tab'];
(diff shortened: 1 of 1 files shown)


DanailH (approved):

oliviertassinari (commented):
It could be great to add a test case for this. It's the second time we iterate on the logic to fix a bug. The first time was in #1409. We also fix the logic with, say when typing "$12".

mnajdova:
> It could be great to add a test case for this. It's the second time we iterate on the logic to fix a bug. The first time was in https://github.com/mui/mui-x/pull/1409. We also fix the logic with, say when typing "$12" into the cell.

Sure, makes sense. I was looking a bit at the test in `cellEditing.DataGridPro.new.test.tsx`, but looks like the test used directly the `apiRef.current.startCellEditMode` API. I may not have much context, but I was expecting that the tests would focus on the cell, simulate some key-down and validate that the editing has started. I am not sure where/what exactly the test should test out.

flaviendelangle:
Feel free to open another PR for the test
I'm merging to unlock you in this release :+1:

mnajdova:
> Feel free to open another PR for the test
I'm merging to unlock you in this release 👍

Thanks @flaviendelangle

m4theushw:
> Sure, makes sense. I was looking a bit at the test in cellEditing.DataGridPro.new.test.tsx, but looks like the test used directly the apiRef.current.startCellEditMode API. I may not have much context, but I was expecting that the tests would focus on the cell, simulate some key-down and validate that the editing has started. I am not sure where/what exactly the test should test out.

This test file is divided into groups whose tests inside reproduce how the user would interact with the editing functionality. The first group is `apiRef` and it contains only tests checking the methods from the editing API. Here we test the behavior of `apiRef.current.startCellEditMode`, for example. It might be controversy to call API methods here but they're part of the public API, so users can call them directly. The group `start edit mode` is where we have tests simulating user interactions (e.g. keydown). Since we're already tested what the API methods do in the previous group, here we only check that they are called with the right params. Inside this group there're different groups equivalent to the different ways that users can interact with the API, e.g. by double-click, by pressing <kbd>Enter</kbd>, by **pressing a printable character** and so on.

The new test could be inside

https://github.com/mui/mui-x/blob/21c1f6c72838719536ba286551c56aa23f305109/packages/grid/x-data-grid-pro/src/tests/cellEditing.DataGridPro.new.test.tsx#L586-L587

PR #5422, opened 2022-07-07
[DataGrid] Added support for `time` column type

### LINKED ISSUE #5421: [data grid] Add native data grid `time` column type
### Duplicates

### Latest version

### Summary 💡

We should be able to use `time` as a native column type.

The filters can stay the same, the only difference would be to only use a time picker instead of the date time picker used for `dateTime`.

I'd be happy to work on it if you would point me to where the modifications should be made.

### Examples 🌈

_No response_

### Motivation 🔦

`date` and `dateTime` are already available as native column types, it's been amazing to be able to use them in my projects.  
Since I use a display a lot of time variables in my projects, this would save a lot of custom code.

### Order ID 💳 (optional)

_No response_

Zenoo:
Closes #5421

This PR should add support for the `time` column type.

This is a work in progress, I simply searched the whole repo for places where `dateTime` was mentioned and edited/added the equivalents for `time`.


oliviertassinari:
From what I recall, the time picker has 1/10 of the requests compared to a date picker in material-ui-pickers. Maybe this should only be a demo?

Also, I would expect a batter UX from using an Autocomplete rather than the Time Picker to implement this.

aaarichter:
@Zenoo it would be great to have your change. looking forward to it

Zenoo:
Sure, but it's been quite some time.  
If you can point to what needs to be done, I'd be happy to.

michelengelen:
This PR has been inactive for a while, so we’re going to close it for now.
If there’s any work here that should be preserved, feel free to create a fork. Thanks for your contributions!

PR #5428, opened 2022-07-08
[core] Add technical support link to _redirects

joserodolfofreitas:
The link will be initially used on the store's keymailer.

More on: https://github.com/mui/mui-store/pull/187

### DIFF
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -8,6 +8,7 @@
 /r/x-license-eula https://mui.com/store/legal/mui-x-eula/ 302
 /r/x-license-key-installation https://mui.com/x/introduction/licensing/#license-key-installation 302
 /r/x-data-grid-no-dimensions https://mui.com/x/react-data-grid/layout/ 302
+/r/x-professional-support https://mui.com/x/introduction/support/#professional-support 302
 
 # Legacy redirection
 # 2021
--- a/docs/public/_redirects
+++ b/docs/public/_redirects
@@ -8,7 +8,7 @@
 /r/x-license-eula https://mui.com/store/legal/mui-x-eula/ 302
 /r/x-license-key-installation https://mui.com/x/introduction/licensing/#license-key-installation 302
 /r/x-data-grid-no-dimensions https://mui.com/x/react-data-grid/layout/ 302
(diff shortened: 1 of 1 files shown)


oliviertassinari (approved):

docs/public/_redirects
```diff
@@ -8,6 +8,7 @@
 /r/x-license-eula https://mui.com/store/legal/mui-x-eula/ 302
 /r/x-license-key-installation https://mui.com/x/introduction/licensing/#license-key-installation 302
 /r/x-data-grid-no-dimensions https://mui.com/x/react-data-grid/layout/ 302
+/r/x-professional-support https://mui.com/x/introduction/support/#professional-support 302
```
mbrookes, line 11:
```suggestion
/r/x-technical-support https://mui.com/x/introduction/support/#technical-support 302
```

mbrookes (changes requested):
Will need to update the wording in the docs

docs/public/_redirects
joserodolfofreitas: Hey @mbrookes, I believe the idea of using "Professional" support is to contrast with the "Community" support.
We've been using "Professional" support in a lot of places by now.

oliviertassinari: Ah, now I get the point that Matt wants to make.

+1 for the proposed change:

```suggestion
/r/x-technical-support https://mui.com/x/introduction/support/#technical-support 302
```

I have started to explore this problem a bit more in https://www.notion.so/mui-org/Professional-services-b9548d4fee2a4181a5787d373854603f, it aligns matter with the terminology I could find in the wild.

mbrookes: I guessed that it was a mix up (mash up?) between professional services and technical support. What's (briefly) described at the linked page is technical support, not professional services (consultancy).

oliviertassinari:
Should we rename https://mui.com/store/legal/mui-x-eula/#support to "Technical Support" too?

docs/public/_redirects
mbrookes: > We've been using "Professional" support in a lot of places by now.

Then you might have a lot of places to fix it. 😁

mbrookes:
We could do for consistency (for the section header only), but it isn't incorrect, just less specific.

PR #5429, opened 2022-07-08
[docs] Add automatic vale check

alexfauquette:


alexfauquette:
For now, I set Google and write-good

Maybe interesting to see if there are some rules to keep from them or not (notice that Google encourages "it's" and write-good encourages "it is" 🙈)

To run locally, you need to install vale with `brew install vale` (on linux and mac)

You can use `vale -h` to get the list of available options

Then run `vale sync` to fetch the styles, and `vale [The file or folder to analyse]` wich are probably `vale README.md` or `vale docs/`

alexfauquette:
@samuelsycamore I cleaned the vale configuration.

Basically every thing is in [.vale.ini](https://github.com/mui/mui-x/pull/5429/files#diff-0da89932043a07a361f29194237fa0cb85aba0bebef98a012a03f84989c26e2e)

For starting I only added a few rules from Google styling rules.

The most problematic one is `Google.We` which catch all the `we`, `us` in the doc. Tried to fix most of them but not always sure I'm improving the doc with that.

I did not add `Google.Passive` and `Google.Will` for now, because we are using them quite a lot and I struggled to rephrase them.

samuelsycamore:
Very cool! I think the passive voice rule might lead to too many awkward phrasings if we're too strict about it. Sometimes it makes sense to go with passive, but I think Vale suggests to replace it every time.

This looks like a great place to start, and we can refine the rules as we go from here.

docs/data/data-grid/api-object/api-object.md
```diff
@@ -62,7 +62,7 @@ const CustomDataGrid = (props) => {
 ```
 
 :::warning
-The API object will be populated by the various plugins of the grid during the first render of the component.
+Various plugins of the grid populate the API object during the first render of the component.
```
samuelsycamore, line 65: This is a good example of a grammatical "edge case" where the passive voice makes sense—as I understand in context, we want to emphasize "the API object." I would just replace the future tense "will be" with "is."

```suggestion
The API object is populated by the various plugins of the grid during the first render of the component
```

oliviertassinari:
Is Vale working well enough on X to be rolled out on MUI Core?

alexfauquette:
At least nobody complained about it, so it does not harm.

However only @LukasTy found an interesting insight from it. Plus the discussion about how strongly we should enforce the rule about `we`.

That's probably because we are super light about the rules to enforce

LukasTy:
I wouldn't say that we've integrated vale into our PR process fully, but whenever I see vale comments on text I do spot some valuable insights.
I'd say we need a bit more time and involvement from every1 to fully assert it's usefulness.
But—in any case—as most of us are non-native English speakers—I think such tools should be at least some value. 😃

PR #5432, opened 2022-07-08
[docs] Fix description for "Striped rows" example

lindapaiste:
This PR fixes a mistake introduced in #3882

### Mistake

The description for the "Custom theme" example contains text that is meant for the "Striped rows" example:
image

While the "Striped rows" example description has a typo ("stripped" instead of "striped"):
image

### Source

The mistake happened in [this commit](https://github.com/mui/mui-x/commit/4f2c9a5a4a88fba0ae6091efcbcdf1c556d9ec3b#diff-cdc05bc9f3c8c0f4cff91c069e3f8959221bf1cf18082e28e8990762735adab4).
image

### Fix

- Re-applied the previous (pre-#3882) description for the "Custom theme" example.
- Moved the current "Custom theme" description to "Striped rows".

### DIFF
--- a/docs/data/data-grid/style/style.md
+++ b/docs/data/data-grid/style/style.md
@@ -121,15 +121,15 @@ You must use `headerAlign` to align the content of the header.
 
 ## Striped rows
 
-The following demo illustrates how the rows of the grid can be stripped.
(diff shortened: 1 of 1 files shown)


m4theushw (approved):
Thanks!

PR #5449, opened 2022-07-11
[l10n] Improve Romanian (ro-RO) locale

rolule:
Add missing translation keys to roRO locale


alexfauquette (approved):
Thanks for the improvement, it should be availabe in next release :)

PR #4956, opened 2022-05-21
[docs] Fix navigation links

oliviertassinari:
Fixing navigation issues reported in https://app.ahrefs.com/site-audit/3523498/11/overview

PR #5450, opened 2022-07-11
[docs] Fix internal link to `valueParser`

### LINKED ISSUE #5176: Broken link to valueParser
### Duplicates

### Latest version

### Current behavior 😯

The valueParser link is a 404

Screenshot 2022-06-11 at 14 14 12

https://mui.com/x/react-data-grid/column-definition/#value-formatter

### Expected behavior 🤔

200

### Steps to reproduce 🕹

_No response_

### Context 🔦

I could only find https://mui.com/x/react-data-grid/editing/#value-parser-and-value-setter as a potential link target but this only covers the use of `valueParser` for the editing feature. I'm not sure that it's good enough, the property is also used for the filtering feature.

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #5176

### DIFF
--- a/docs/data/data-grid/column-definition/column-definition.md
+++ b/docs/data/data-grid/column-definition/column-definition.md
@@ -73,7 +73,7 @@ In the following demo, a formatter is used to display the tax rate's decimal val
 
 The value generated is only used for rendering purposes.
 Filtering and sorting do not rely on the formatted value.
-Use the [`valueParser`](/x/react-data-grid/cells/#value-parser) to support filtering.
+Use the [`valueParser`](/x/react-data-grid/editing/#value-parser-and-value-setter) to support filtering.
 
 ## Rendering cells
 


joserodolfofreitas (approved):

docs/data/data-grid/column-definition/column-definition.md
```diff
@@ -73,7 +73,7 @@ In the following demo, a formatter is used to display the tax rate's decimal val
 
 The value generated is only used for rendering purposes.
 Filtering and sorting do not rely on the formatted value.
-Use the [`valueParser`](/x/react-data-grid/cells/#value-parser) to support filtering.
+Use the [`valueParser`](/x/react-data-grid/editing/#value-parser-and-value-setter) to support filtering.
```
oliviertassinari, line 76: I think that it's a bit confusing. We say, use the value parser to implement customer filtering

Screenshot 2022-07-25 at 13 13 11

and then link people to a different feature: edition, which is not filtering. Maybe it would make sense to have the same section in the documentation as https://mui.com/x/react-data-grid/filtering/#parsing-values but for the column filtering.

alexfauquette: I remind that during the first months, I found parser, getter, formatted and setter quite confusing and open #3215

What would you think about a dedicated page for them to provide an overview of how data is transformed like proposed in #3215 ?

oliviertassinari: Flavien had commented a bit on this problem in https://github.com/mui/mui-x/pull/4956#issuecomment-1134565813. I guess the demo in https://mui.com/x/react-data-grid/column-definition/#value-formatter is a good illustration of the need for `valueParser`. When you search for 20 you get no results 🙃. Link it here

The chart https://github.com/mui/mui-x/issues/3215#issuecomment-1000744699 helps understand this works 👍   . It could help to have documentation about it. The tricky part might be, how do we have developers find it when they need it?

Issue #5112, opened 2021-09-24
[pickers] Not possible to provide custom styling to the DesktopDatePicker component


Theofilos-Chamalis:
I have updated to the latest @mui/material v5.0.1, used the DesktopDatePicker component and tried to provide custom styling to it (different background-color, font color, date-selection color and a different input field style).

Using the renderInput to render our custom InputField component was successfull, however I was not able to style the elements from the Modals that show up (a.k.a MuiCalendarPicker).

- [x] The issue is present in the latest release.
- [x] I have searched the [issues](https://github.com/mui-org/material-ui/issues) of this repository and believe that this is not a duplicate.

## Current Behavior 😯

I have used the following methods which were unsuccessful:

1. Wrap the DesktopDatePicker or its LocalizationProvider wrapper component with the ThemeProvider using a custom theme.
2. Using the style prop in the DesktopDatePicker
3. Using the sx prop in the DesktopDatePicker (trying to select the MuiCalendarPicker component with `sx={{'& .MuiCalendarPicker-root':{backgroundColor: 'purple'}}}`
4. Using the PopperProps (this only controls an overlay behind the actual MuiCalendarPicker)
5. Using a custom classname and selecting `'& .MuiCalendarPicker-root'` or `'& .MuiCalendarPicker'` to style.

## Expected Behavior 🤔

The styling of the MuiCalendarPicker or the styling of the Modal contents in general should change

## Your Environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>
  
```

   System:
    OS: macOS 11.6
  Binaries:
    Node: 16.10.0 - ~/.nvm/versions/node/v16.10.0/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v16.10.0/bin/yarn
    npm: 7.24.1 - ~/.nvm/versions/node/v16.10.0/bin/npm
  Browsers:
    Chrome: 94.0.4606.61
    Edge: Not Found
    Firefox: Not Found
    Safari: 15.0
  npmPackages:
    @emotion/react: ^11.4.1 => 11.4.1 
    @emotion/styled:  10.0.27 
    @mui/core:  5.0.0-alpha.48 
    @mui/lab: ^5.0.0-alpha.48 => 5.0.0-alpha.48 
    @mui/material: ^5.0.1 => 5.0.1 
    @mui/private-theming:  5.0.1 
    @mui/styled-engine:  5.0.1 
[...]

PR #5471, opened 2022-07-12
[DatePicker] Fix to pass down `className` prop provided on DatePicker to `renderInput`

### LINKED ISSUE #4942: [DatePicker] `className`s provided via prop are never applied to root element
Not much more to say; the `className` prop is destructured in preparation to apply to the component but never gets used.

https://github.com/mui-org/material-ui/blob/ad2535a8b15f76355f4559dbac02e9cb379269ba/packages/mui-lab/src/internal/pickers/Picker/Picker.tsx#L77

CruseCtrl:


CruseCtrl:
Actually, this doesn't fix #4292 because that issue is mostly concerned with the `StaticDatePicker`, but I've made the `className` work for the `DatePicker`. But maybe it helps #5112

m4theushw (commented):
Thanks for the contribution! The CI is green now.

MUI components forward the `className` prop to the root element. However, the pickers don't have an root element, since they wrap a TextField. With that being said, I don't know if it would be correct to forward the root `className` prop directly to the input. In the other hand, https://github.com/mui/mui-x/issues/4942 is different, the calendar of the StaticDatePicker is the root element and `className` should be spread to it. Do you want to go back and modify the PR so it fixes https://github.com/mui/mui-x/issues/4942?

CruseCtrl:
Thanks @m4theushw. I'm not actually using the `StaticDatePicker`, it's just the `DatePicker` that I'm trying to fix at the moment. It does seem strange that the typescript types currently let you pass in a `className` which never gets used. Would the `TextField` not count as the root element in this case?

alexfauquette (approved):
I do not manage to understand which problem does it solves

You pass the `DatePicker` props `className` to the `renderInput`. That means you can now do

```jsx
<DatePicker className={myClassName} renderInput={params => <TextField {...params}/>} />
```

Which is the same as

```jsx
<DatePicker renderInput={params => <TextField {...params} className={myClassName} />} />
```

From what I understand, both #4942 and #5112 are about customizing the content of the `Modal`/`Popper` which will not be solved with this PR

---

But it seems to be consistent with the core package. The [Autocomplete](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Autocomplete/Autocomplete.js#L534:L534) component also do not have a root component. Or to be precise it has a `AutocompeteRoot` which is a `div` wrapping the `renderInput` but do not contain the `Popper`. So this PR seems to be the closest solution

LukasTy (approved):
LGTM, thank you for the contribution! 🙌 
I've also pushed a tiny change to add same logic to `MobileDatePicker`

alexfauquette (approved):

PR #5448, opened 2022-07-11
[DataGrid] Improve German (de-DE) locale

sebastianfrey:
Add missing translations for DataGrid.

PR #5473, opened 2022-07-12
[test] Wait for flags to load on regression tests

m4theushw:
Fixes https://github.com/mui/mui-x/pull/5448#issuecomment-1181723830

### DIFF
--- a/test/regressions/index.test.js
+++ b/test/regressions/index.test.js
@@ -101,9 +101,17 @@ async function main() {
         // Move cursor offscreen to not trigger unwanted hover effects.
         page.mouse.move(0, 0);
 
+        const pathsToNotWaitForFlagCDN = [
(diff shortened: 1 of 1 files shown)


cherniavskii (approved):

oliviertassinari:
In the project I worked on before MUI, I solved the problem like this:

```jsx
  await page.evaluate(async () => {
    const selectors = Array.from(document.querySelectorAll('img'))
    await Promise.all(
      selectors.map(img => {
        if (img.complete) {
          return null
        }

        return new Promise(resolve => {
          img.addEventListener('load', resolve)
          img.addEventListener('error', resolve)
        })
      })
    )
  })
```

Is this something that could work? It seems a challenge to keep `pathsToNotWaitForFlagCDN` in sync with the codebase.

Issue #5453, opened 2022-07-10
[pickers] AdapterLuxon onChange sends Luxon DateTime instead of js Date


cyrilchapon:
**Expected**
`onChange` event of DatePicker and others should send a JS Date object.
(just like with DateFns => https://codesandbox.io/s/sqoo8j?file=/demo.tsx)

**Actual**
When using AdapterLuxon, it sends a Luxon DateTime
(see console => https://codesandbox.io/s/materialuipickers-demo-mui-x-forked-73ng8p?file=/demo.tsx)

flaviendelangle:
This is the expected behavior
Why do you think it should pass a JS Date object instead ?

cyrilchapon:
@flaviendelangle, hum, interesting.
After a re-read of documentation, and re-test; it turns out it might just be a doc-clarification need.

My process was the following :

1. Read the general doc on the pickers modules : https://mui.com/x/react-date-pickers/getting-started/
2. Read the paragraphs about date library adapters, etc.
  At that point, I instinctively chose Luxon because I like it, and was pleased to see an official adapter
3. Installed stuff, copy-pasted setup and imports, adapted the LocalizationProvider in my app too use it
  ```
  // or for Luxon
  import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';

  function App({ children }) {
    return (
      <LocalizationProvider dateAdapter={AdapterDateFns}>
        {children}
      </LocalizationProvider>
    );
  }
  ```
4. Chose a component , `<DatePicker />` in my case.
5. Searched for the documentation of it (https://mui.com/x/react-date-pickers/date-picker/), and understood it
6. Simply copy-pasted the basic example, and started coding with it
  ```
  import * as React from 'react';
  import TextField from '@mui/material/TextField';
  import { DatePicker } from '@mui/x-date-pickers/DatePicker';

  export default function MyPicker() {
    const [value, setValue] = React.useState<Date | null>(null);

    return (
      <DatePicker
        label="Basic example"
        value={value}
        onChange={(newValue) => {
          setValue(newValue);
        }}
        renderInput={(params) => <TextField {...params} />}
      />
    );
  }
  ```
  At this point, my guess was the abstraction layer was actually abstracting the `onChange` and `value` typings, manipulating **internally** with date libraries through @date-io; but simply exposing JS Dates

Indeed, now I'm realizing :
- that all examples are actually using *date-fns*
[...]

PR #5481, opened 2022-07-13
[docs] Use dayjs instead of date-fns in doc examples

flaviendelangle:
Part of #5453

- [x] Migrate "Getting started" demos
- [x] Migrate "Date picker" demos
- [x] Migrate "Date range picker" demos
- [x] Migrate "Date time picker" demos
- [x] Migrate "Time picker" demos
- [x] Migrate "Localization" demos
- [x] Migrate "Custom components" demos

## Why stop using `date-fns` on demos ?

`date-fns` uses regular `Date` object, which make invisible all our date translation behaviors. 
For instance, it is very hard to see that `minDate` takes a parsed date, not an input date.

Note that this change is only about our doc example.
The packages still support `date-fns` like before.

## Why use `dayjs`

Among the 4 adapter we support out of the box

- Moment is not recommended for new projects
- Date Fns uses regular `Date` object
- Luxon is a lot less used than the others

Luxon would have been my second choice.

image

_____

While passing through all the demos, I noticed some things I will like to improve in a follow up:

- https://mui.com/x/react-date-pickers/date-picker/#landscape-orientation => Only test the orientation in this demo, no need to demonstrate how the `shouldDisableDate` works


docs/data/date-pickers/date-time-picker/CustomDateTimePicker.tsx
```diff
@@ -1,58 +1,59 @@
...
-            setValue(newValue);
+            setDateWithInitialValue(newValue);
           }}
           label="With error handler"
           onError={console.log}
-          minDate={new Date('2018-01-01T00:00')}
-          inputFormat="yyyy/MM/dd hh:mm a"
+          minDate={dayjs('2018-01-01T00:00')}
+          inputFormat="YYYY/MM/DD hh:mm a"
           mask="___/__/__ __:__ _M"
```
alexfauquette, line 51: Just noticed one character is missing
```suggestion
          inputFormat="YYYY/MM/DD hh:mm a"
          mask="____/__/__ __:__ _M"
```

docs/data/date-pickers/localization/LocalizedTimePicker.tsx
```diff
@@ -1,22 +1,22 @@
...
+import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
 import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
 import { TimePicker } from '@mui/x-date-pickers/TimePicker';
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 import FormControl from '@mui/material/FormControl';
 import InputLabel from '@mui/material/InputLabel';
 
 const localeMap = {
-  en: enLocale,
+  en: undefined,
```
alexfauquette, line 19: It's a bit surprising when seeing `undefined` for a locale
```suggestion
  // Do not need to pass `en` since it's the default one
  en: undefined,
```

flaviendelangle: The problem is that the `en` local exported by `dayjs` is incomplete
I don't understand why.
But since it does not have all the needed formats, if you try passing it to the adapter it crashes .

oliviertassinari:
Oh wow, I'm surprised by how many downloads dayjs has. I expected something similar to luxon.

I wish npm was providing a way to get download analytics on **direct** dependencies. For instance, Cypress with 4M downloads/week has a dependency on dayjs which accounts for 33% of days downloads. The same way, concurrently with 3M+ downloads/week has a dependency on date-fns.

SimilarWeb gives a few more information:

- [Dayjs](https://www.similarweb.com/website/day.js.org/): 167k sessions in August, China: 13% of the traffic, US, 7%
- [Date-fns](https://www.similarweb.com/website/date-fns.org/) 203k sessions in August, China: 5% of the traffic, US, 9%

PR #5413, opened 2022-07-07
[docs] Improve the Events doc page

flaviendelangle:

PR #5494, opened 2022-07-14
[docs] Add note for unsubscribe callback on "Data grid - Events" page

marcandreh:
The "Subscribing to events" section on the "Data grid - Events" page does not clearly mention how to actually unsubscribe the event handler afterwards. This may lead to bugs, especially when using `apiRef.current.subscribeEvent()` within a `React.useEffect` hook.

The documentation comment for the return `@returns A function to unsubscribe from this event` can easily be overlooked, so a warning or info should emphasise the use of the unsubscribe callback.

### DIFF
--- a/docs/data/data-grid/events/events.md
+++ b/docs/data/data-grid/events/events.md
@@ -29,6 +29,11 @@ subscribeEvent: (
 ) => () => void;
 ```
 
+:::warning
+The call to `apiRef.current.subscribeEvent` returns a callback that **unsubscribes** the given handler.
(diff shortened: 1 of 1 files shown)


flaviendelangle:
This page is being reworked in https://github.com/mui/mui-x/pull/5413

I'll add a cleaner note about the return function there :+1:

marcandreh:
> This page is being reworked in #5413
> 
> I'll add a cleaner note about the return function there 👍

Sounds good, thank you 👍

flaviendelangle:
Closes in favor of #5413

PR #5498, opened 2022-07-14
[l10n] Add Traditional Chinese (zh-TW) locale

happyincent:
Add Traditional Chinese locale to data grid.
> In this PR, I translate "rows/columns" as "個/欄" to avoid the indefinite translation that could be "列/行" or "行/列" in a different context [1][2].
> [1] [Translations in Localized Microsoft Products](https://www.microsoft.com/en-us/language/Search?&searchTerm=columns&langID=129&Source=true&productid=0)
> [2] [The determinant in linear algebra](https://ccjou.wordpress.com/2012/04/17/)


flaviendelangle (approved):

PR #5504, opened 2022-07-15
[pickers] New unstable field components

flaviendelangle:
**These components are very early stage, please don't use them in production**

[Doc Preview of DateField](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/)
[Doc Preview of DateRangeField](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-range-field/)

Part of #5531

### Wanted interactions

- Ability to add a day / a month / a year with the "ArrowUp" and "ArrowDown" keys
- Ability to switch from day to month to year with the "ArrowLeft" and "ArrowRight" keys

### Missing `@date-io` methods

- `addYear` (similar to `addDay` and `addMonth`) => https://github.com/dmtrKovalenko/date-io/pull/623


joserodolfofreitas:
The interactions with arrow keys feel great! Could it somehow block the implementation for a chain of digits as input?

flaviendelangle:
> Btw, where does this POC leave us in terms of a single input implementation for a date range?

I did not look into it yet
It will definitely make the analysis of the input harder but maybe that's doable.
Let's try to see if it's viable for a the simple use case and then stress test it on harder one.

We also have the date time where we may want to handle it like Retool and open a select, in which case the input must only cover the date part.

flaviendelangle:
> Could it somehow block the implementation for a chain of digits as input?

I did not understand sorry

joserodolfofreitas:
> > Could it somehow block the implementation for a chain of digits as input?
> 
> I did not understand sorry

I raised the question because I noticed you cannot input digits on the current text field POC. It currently only supports the arrow keys, so the question is, will there be any problem with supporting digits as well? 

And by "chain of digits", I meant an input like: 1 5 0 7 2 2 0 2 2 to have an output like 15/07/2022

flaviendelangle:
> Will there be any problem with supporting numeric values as well?

Yes, I started to work on it.
It raised a certain amount of questions so I'm trying to get a first working draft to discuss trade-offs.
For instance on Telerik, when you have a full letter month, you can still type it's number and it sets the month (type 5 => sets "May")

flaviendelangle:
@joserodolfofreitas Concerning the editing, Telerik does not goes from what section of the date to another automatically.

If you are on the day and type "1", "2", "3", it will set the day to "1", then "12", then "3".
To go to the section section you have to use the keyboard arrows.

Spectrum does move automatically to the next section.
For numerical value it is doable. But I'm curious to see how the handle that for full letter months.

___

This behavior could be customizable if we have users asking for both

flaviendelangle:
https://user-images.githubusercontent.com/3309670/183668935-f57ded52-f24e-4da8-93fc-fb01b466f916.mp4

I'm playing with the range version
It add quite a lot of complexity, but it's totally doable with the Telerik approach

LukasTy:
@flaviendelangle There seems to be an issue with preceding zeroes in `days` section.

https://user-images.githubusercontent.com/4941090/184370612-10f6e2d9-3e07-4d69-9d15-5b0c67bee71c.mov

Or am I missing something?

flaviendelangle:
@LukasTy there are probably quite a lot of similar problems yes
If you want to have a look feel free :+1:

oliviertassinari:
Is the support for mobile part in the scope of this initial PR?

alexfauquette:
@oliviertassinari the scop is to identify early blockers that could force us to move for another strategy.

On my mobile device, the behavior is completely broken. I do not manage to enter a date

gerdadesign:
I'm curious for the day field in [this preview](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/) — month seems to truncate to 2 digits and year to 4, but day seems to allow you minimum 3 digits with no maximum? 

Also noting a pattern I've seen of the placeholder labels giving an indication for how many digits the input format is looking for: MM/DD/YYYY or DD/MM/YY

alexfauquette:
The day bug should be fixed. The problem being that with invalid date (partially entered) the maximum value for day was not defined

LukasTy:
The focus management needs looking into.
- Focus the field
- change day with arrows
- press `ArrowRight` -> focus jumps to `year` (and in case of date range field to the end range year)

LukasTy:
Regarding accessibility, this, to me, seems like an implementation very similar to the follow [spin button example](https://www.w3.org/WAI/ARIA/apg/example-index/spinbutton/datepicker-spinbuttons.html).
I'd say we could implement the same a11y behaviour (of course with a following PR), unless you see that there would be any issues with it?

LukasTy:
Arrow (Up & Down) navigation for month field in letters only works after any month value has already been initialised.

https://user-images.githubusercontent.com/4941090/185784949-304c8c6b-f962-4afb-b50c-d9af86ce2dbd.mov

LukasTy:
[Clearable](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-input-props) example does not work correctly. It does not reset the value in the component.

LukasTy:
[Time format/input](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-date-format) does not seem like it wants to cooperate. 🤔 

https://user-images.githubusercontent.com/4941090/185811088-f350b506-824e-4115-8a8a-93df947d0235.mov

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -0,0 +1,415 @@
...
+      case event.key === 'ArrowRight': {
+        event.preventDefault();
+
+        if (state.selectedSectionIndexes == null) {
+          updateSelectedSections(0);
+        } else if (state.selectedSectionIndexes.start !== state.selectedSectionIndexes.end) {
+          updateSelectedSections(state.selectedSectionIndexes.end);
+        } else if (state.selectedSectionIndexes.start < state.sections.length - 1) {
+          updateSelectedSections(state.selectedSectionIndexes.start + 1);
+        }
```
LukasTy, line 135: Switching these cases solves the [focus management issue](https://github.com/mui/mui-x/pull/5504#issuecomment-1221503822).
```suggestion
        } else if (state.selectedSectionIndexes.start < state.sections.length - 1) {
          updateSelectedSections(state.selectedSectionIndexes.start + 1);
        } else if (state.selectedSectionIndexes.start !== state.selectedSectionIndexes.end) {
          updateSelectedSections(state.selectedSectionIndexes.end);
        } 
```

flaviendelangle:
@LukasTy I rewrote the arrow up / down edition on invalid dates, it should be a lot better now

flaviendelangle:
@gerdadesign 
> Also noting a pattern I've seen of the placeholder labels giving an indication for how many digits the input format is looking for: MM/DD/YYYY or DD/MM/YY

For now the placeholder of an empty section is the name of the section (`month` for a month section for instance)
Which is really basic and do not support any kind of localization.
We should probably improve this, I'm open to suggestions for the default behavior and the customization options.

flaviendelangle:
> [Clearable](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-input-props) example does not work correctly. It does not reset the value in the component.

@LukasTy I had a dumb bug where I did not use the controlled value in the component, several of your bugs were probably cause by this error, it is fixed now

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  const handleInputKeyDown = useEventCallback((event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (!inputRef.current || state.sections.length === 0) {
+      return;
+    }
+
+    // eslint-disable-next-line default-case
+    switch (true) {
+      // Select all
+      case event.key === 'a' && event.ctrlKey: {
```
LukasTy, line 120: In order to support both `Ctrl + A` and Mac specific `Command + A`:
```suggestion
      case event.key === 'a' && (event.ctrlKey || event.metaKey): {
```

LukasTy:
> > [Clearable](https://deploy-preview-5504--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-input-props) example does not work correctly. It does not reset the value in the component.
> 
> I had a dumb bug where I did not use the controlled value in the component, several of your bugs were probably cause by this error, it is fixed now

I can confirm that all my mentioned issues have been resolved. 🙌 👍

LukasTy:
@flaviendelangle What do you think about adding the following diff?
It adds management of `Home`, `End`, `PageUp` and `PageDown` keys [mentioned here](https://www.w3.org/WAI/ARIA/apg/example-index/spinbutton/datepicker-spinbuttons.html) as well as some function name alignment.
**Note**: We currently don't have a clear way to know actual `max` or `min` `year`, hence I did not that `Home`/`End` management for it.

```diff
diff --git a/packages/x-date-pickers/src/internals/hooks/useField/useField.ts b/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
index 5470f19c..84c01798 100644
--- a/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
+++ b/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
@@ -11,12 +11,13 @@ import {
   UseFieldState,
 } from './useField.interfaces';
 import {
+  AvailableAdjustKeyCode,
   cleanTrailingZeroInNumericSectionValue,
   getMonthsMatchingQuery,
   getSectionValueNumericBoundaries,
   getSectionVisibleValue,
-  incrementDateSectionValue,
-  incrementOrDecrementInvalidDateSection,
+  adjustDateSectionValue,
+  adjustInvalidDateSectionValue,
   setSectionValue,
 } from './useField.utils';
 
@@ -153,7 +154,7 @@ export const useField = <
       }
 
       // Reset the value of the selected section
-      case event.key === 'Backspace': {
+      case event.key === 'Backspace' || event.key === 'Delete': {
         event.preventDefault();
 
         if (readOnly) {
@@ -181,7 +182,12 @@ export const useField = <
       }
 
       // Increment / decrement the selected section value
-      case event.key === 'ArrowUp' || event.key === 'ArrowDown': {
+      case event.key === 'ArrowUp' ||
+        event.key === 'ArrowDown' ||
+        event.key === 'Home' ||
+        event.key === 'End' ||
+        event.key === 'PageUp' ||
+        event.key === 'PageDown': {
         event.preventDefault();
 
         if (readOnly || state.selectedSectionIndexes == null) {
@@ -196,21 +202,21 @@ export const useField = <
 
         // The date is not valid, we have to increment the section value rather than the date
         if (!utils.isValid(activeDate.value)) {
-          const newSectionValue = incrementOrDecrementInvalidDateSection(
+          const newSectionValue = adjustInvalidDateSectionValue(
             utils,
             activeSection,
-            event.key === 'ArrowUp' ? 'increment' : 'decrement',
+            event.key as AvailableAdjustKeyCode,
           );
 
           updateSections(
             setSectionValue(state.sections, state.selectedSectionIndexes.start, newSectionValue),
           );
         } else {
-          const newDate = incrementDateSectionValue(
+          const newDate = adjustDateSectionValue(
             utils,
             activeDate.value,
             activeSection.dateSectionName,
-            event.key === 'ArrowDown' ? -1 : 1,
+            event.key as AvailableAdjustKeyCode,
           );
 
           const newValue = activeDate.update(newDate);
diff --git a/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts b/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
index 0dcd3f56..64977a0d 100644
--- a/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
+++ b/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
@@ -37,33 +37,89 @@ export const getDateSectionNameFromFormat = (format: string): DateSectionName =>
   throw new Error(`getDatePartNameFromFormat don't understand the format ${format}`);
 };
 
-export const incrementDateSectionValue = <TDate>(
+export type AvailableAdjustKeyCode =
+  | 'ArrowUp'
+  | 'ArrowDown'
+  | 'PageUp'
+  | 'PageDown'
+  | 'Home'
+  | 'End';
+
+const getDeltaFromKeyCode = (keyCode: Omit<AvailableAdjustKeyCode, 'Home' | 'End'>) => {
+  switch (keyCode) {
+    case 'ArrowUp':
+      return 1;
+    case 'ArrowDown':
+      return -1;
+    case 'PageUp':
+      return 5;
+    case 'PageDown':
+      return -5;
+    default:
+      return 0;
+  }
+};
+
+export const adjustDateSectionValue = <TDate>(
   utils: MuiPickersAdapter<TDate>,
   date: TDate,
   datePartName: DateSectionName,
-  datePartValue: 1 | -1,
+  keyCode: AvailableAdjustKeyCode,
 ) => {
+  const delta = getDeltaFromKeyCode(keyCode);
+  const isStart = keyCode === 'Home';
+  const isEnd = keyCode === 'End';
   switch (datePartName) {
     case 'day': {
-      return utils.addDays(date, datePartValue);
+      if (isStart) {
+        return utils.startOfMonth(date);
+      }
+      if (isEnd) {
+        return utils.endOfMonth(date);
+      }
+      return utils.addDays(date, delta);
     }
     case 'month': {
-      return utils.addMonths(date, datePartValue);
+      if (isStart) {
+        return utils.startOfYear(date);
+      }
+      if (isEnd) {
+        return utils.endOfYear(date);
+      }
+      return utils.addMonths(date, delta);
     }
     case 'year': {
-      return utils.addYears(date, datePartValue);
+      return utils.addYears(date, delta);
     }
     case 'am-pm': {
-      return utils.addHours(date, datePartValue * 12);
+      return utils.addHours(date, delta ? 1 : 1 * 12);
     }
     case 'hour': {
-      return utils.addHours(date, datePartValue);
+      if (isStart) {
+        return utils.startOfDay(date);
+      }
+      if (isEnd) {
+        return utils.endOfDay(date);
+      }
+      return utils.addHours(date, delta);
     }
     case 'minute': {
-      return utils.addMinutes(date, datePartValue);
+      if (isStart) {
+        return utils.setMinutes(date, 0);
+      }
+      if (isEnd) {
+        return utils.setMinutes(date, 59);
+      }
+      return utils.addMinutes(date, delta);
     }
     case 'second': {
-      return utils.addSeconds(date, datePartValue);
+      if (isStart) {
+        return utils.setSeconds(date, 0);
+      }
+      if (isEnd) {
+        return utils.setSeconds(date, 59);
+      }
+      return utils.addSeconds(date, delta);
     }
     default: {
       return date;
@@ -277,13 +333,16 @@ export const getSectionValueNumericBoundaries = <TDate>(
   }
 };
 
-export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends FieldSection>(
+export const adjustInvalidDateSectionValue = <TDate, TSection extends FieldSection>(
   utils: MuiPickersAdapter<TDate>,
   section: TSection,
-  type: 'increment' | 'decrement',
+  keyCode: AvailableAdjustKeyCode,
 ) => {
   const today = utils.date()!;
-  const delta = type === 'increment' ? 1 : -1;
+  const delta = getDeltaFromKeyCode(keyCode);
+  const isStart = keyCode === 'Home';
+  const isEnd = keyCode === 'End';
+  const shouldSetAbsolute = section.value === '' || isStart || isEnd;
 
   switch (section.dateSectionName) {
     case 'year': {
@@ -299,8 +358,8 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
 
     case 'month': {
       let newDate: TDate;
-      if (section.value === '') {
-        if (type === 'increment') {
+      if (shouldSetAbsolute) {
+        if (delta > 0 || isEnd) {
           newDate = utils.endOfYear(today);
         } else {
           newDate = utils.startOfYear(today);
@@ -314,8 +373,8 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
 
     case 'day': {
       let newDate: TDate;
-      if (section.value === '') {
-        if (type === 'increment') {
+      if (shouldSetAbsolute) {
+        if (delta > 0 || isEnd) {
           newDate = utils.endOfMonth(today);
         } else {
           newDate = utils.startOfMonth(today);
@@ -332,7 +391,7 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
       const pm = utils.formatByString(utils.endOfDay(today), section.formatValue);
 
       if (section.value === '') {
-        if (type === 'increment') {
+        if (delta > 0 || isEnd) {
           return pm;
         }
         return am;
@@ -347,8 +406,8 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
 
     case 'hour': {
       let newDate: TDate;
-      if (section.value === '') {
-        if (type === 'increment') {
+      if (shouldSetAbsolute) {
+        if (delta > 0 || isEnd) {
           newDate = utils.endOfDay(today);
         } else {
           newDate = utils.startOfDay(today);
@@ -362,9 +421,9 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
 
     case 'minute': {
       let newDate: TDate;
-      if (section.value === '') {
+      if (shouldSetAbsolute) {
         // TODO: Add startOfHour and endOfHours to adapters to avoid hard-coding those values
-        const newNumericValue = type === 'increment' ? 59 : 0;
+        const newNumericValue = delta > 0 || isEnd ? 59 : 0;
         newDate = utils.setMinutes(today, newNumericValue);
       } else {
         newDate = utils.addMinutes(utils.setMinutes(today, Number(section.value)), delta);
@@ -375,9 +434,9 @@ export const incrementOrDecrementInvalidDateSection = <TDate, TSection extends F
 
     case 'second': {
       let newDate: TDate;
-      if (section.value === '') {
+      if (shouldSetAbsolute) {
         // TODO: Add startOfMinute and endOfMinute to adapters to avoid hard-coding those values
-        const newNumericValue = type === 'increment' ? 59 : 0;
+        const newNumericValue = delta > 0 || isEnd ? 59 : 0;
         newDate = utils.setSeconds(today, newNumericValue);
       } else {
         newDate = utils.addSeconds(utils.setSeconds(today, Number(section.value)), delta);

```

flaviendelangle:
I applied @LukasTy suggestion to handle other keys edition

LukasTy (approved):

gerdadesign:
Not trying to reopen the merged PR, but just continuing this convo! LMK if I should move it elsewhere. 

> For now the placeholder of an empty section is the name of the section (`month` for a month section for instance)
Which is really basic and do not support any kind of localization.
We should probably improve this, I'm open to suggestions for the default behavior and the customization options.

Seeing "month" written out gives me the expectation that I will be choosing a month written out (January, February, March, etc.), even though the input is looking for a number. Since dates and time can have so many different formats, it would be beneficial to be as clear as possible for the end user about the expected input. 

I was also thinking about the customization options and put together a [basic prototype](https://www.figma.com/proto/2z9h8rxbPi7rVqChAx8D54/Date-and-Time-pickers?node-id=131%3A4889&scaling=min-zoom&page-id=2%3A10&starting-point-node-id=131%3A4889&show-proto-sidebar=1) to explore customization. There are 2 flows: the standard Material components + a customized version with many style changes that I could imagine someone may want. 

This also brought up the options for how to show information that is expected, but hasn't been keyed in yet. I see we are currently used zeroes to show that, but maybe this is something that people would want to customize as well? 
Options

flaviendelangle:
> Not trying to reopen the merged PR, but just continuing this convo! LMK if I should move it elsewhere.

Maybe move it to the issue https://github.com/mui/mui-x/issues/5531, it seems to be a good place to discuss cross PR topics around the fields.

oliviertassinari (commented):
It looks great! I have left a few comments as code comments so we can benefit from threads (#5531) doesn't provide this.

docs/data/date-pickers/date-field/BasicDateField.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
+import { Unstable_DateField as DateField } from '@mui/x-date-pickers/DateField';
+
+export default function BasicDateField() {
+  return (
+    <LocalizationProvider dateAdapter={AdapterDateFns}>
+      <DateField label="Basic date field" />
```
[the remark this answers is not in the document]
flaviendelangle: This one is a known issue
You have the same in Telerik: https://www.telerik.com/kendo-react-ui/components/dateinputs/datepicker/

The problem is that I select all sections on focus.
And when you click, it fires onFocus before onClick.
So when onClick is fired, the focus the selection caused by the focus is already applied, so it clicks on a fully selected input.

And on a fully selected input, I can't find the cursor position of the click, the cursor just moves to the beginning of the input.


One solution would be to delay the selection caused by the focus and to skip it if a click selection has been done in the meanwhile.

docs/data/date-pickers/date-field/LocalizedDateField.tsx
```diff
@@ -0,0 +1,30 @@
...
+            label="French locale (default format)"
+            value={value}
+            onChange={(newValue) => setValue(newValue)}
+          />
+          <DateField
+            label="French locale (full letter month)"
+            value={value}
+            onChange={(newValue) => setValue(newValue)}
+            format="dd MMMM yyyy"
+          />
```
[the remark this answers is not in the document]
flaviendelangle: The placeholder is clearly temporary.

Telerik uses the section type

image

Spectrum uses something closer from the date tokens

image

docs/data/date-pickers/date-field/BasicDateField.js
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
+import { Unstable_DateField as DateField } from '@mui/x-date-pickers/DateField';
+
+export default function BasicDateField() {
+  return (
+    <LocalizationProvider dateAdapter={AdapterDateFns}>
+      <DateField label="Basic date field" />
```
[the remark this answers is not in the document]
flaviendelangle: I don't know for this one
In Spectrum, if I press `0` it does not write anything, even if the end format will be `01`

Press `0` => still render the placeholder (totally ignores the event I think)
Press `1` => renders `01`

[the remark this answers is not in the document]
oliviertassinari: > Should we allow 00?

To be more precise, I was thinking of what would happen when leaving the selection. A native browser seems to turn 00 to 01 when unselecting.

docs/data/date-pickers/date-field/BasicDateField.tsx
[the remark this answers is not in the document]
flaviendelangle: I have a PR to improve this behavior https://github.com/mui/mui-x/pull/5901

packages/x-date-pickers/src/DateField/DateField.tsx
```diff
@@ -0,0 +1,22 @@
...
+  ref: React.Ref<HTMLInputElement>,
+) {
+  const { inputRef, inputProps } = useDateField<
+    TInputDate,
+    TDate,
+    DateFieldProps<TInputDate, TDate>
+  >(inProps);
+
+  return <TextField ref={ref} inputRef={inputRef} {...inputProps} />;
+}) as DateFieldComponent;
```
oliviertassinari, line 22: Looking at the code, I almost feel like it would be better to expose this userland. Not sure.

flaviendelangle: It is exposed and should be documented soon

Do you think we should not expose the component ?

oliviertassinari: OK great to know.

> Do you think we should not expose the component ?

I can't make my mind on this. I think that it's not impossible that we could be better off without this abstraction, with the rationale that lower level API makes easier customizations.

As a side note, I wonder if developers won't expect to be able to target `MuiDateField` in the theme.

flaviendelangle: https://github.com/mui/mui-x/blob/next/packages/x-date-pickers/src/DateField/DateField.tsx

The component has a little more logic know and supports theme overrides as you mentionned.
And we have multi-input range fields with a lot more logic (those I would clearly not remove them).

Issue #5192, opened 2022-06-14
[data grid] Columns disappear when switch browser to full screen


deepen77:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

When browser is opened as example in the half screen and try to switch to full screen then part of columns on the right hand side are disappeared. When you start to resize  or refres the web browser , everything back to normal.

https://4w7qra--run.stackblitz.io/

### Expected behavior 🤔

After switch to full screen, the columns shoull be visible directly

### Steps to reproduce 🕹

Steps:

1. 
![image](https://user-images.githubusercontent.com/58135902/173524400-22d365e3-959f-4492-a8da-bc440c2bb17e.png)

2. 
![image](https://user-images.githubusercontent.com/58135902/173524866-86413ae5-1472-4b3d-b55a-9e763e46941f.png)

### Context 🔦

This issue affected data loaded before switching to full screen. After switch to full screen, the users usually trying refresh the page to see all columns and if data is paginated, all those features are back to initial states.

### Your environment 🌎

System:
    OS: Windows 10 10.0.19044
  Binaries:
    Node: 16.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 8.5.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.19041.1266.0), Chromium (102.0.1245.39)
  npmPackages:
    @emotion/react: ^11.9.0 => 11.9.0
    @emotion/styled: ^11.8.1 => 11.8.1
    @mui/base:  5.0.0-alpha.83
    @mui/icons-material: ^5.8.2 => 5.8.2
    @mui/material: ^5.8.2 => 5.8.2
    @mui/private-theming:  5.8.0
    @mui/styled-engine:  5.8.0
    @mui/system:  5.8.2
    @mui/types:  7.1.3
    @mui/utils:  5.8.0
    @mui/x-data-grid: ^5.12.0 => 5.12.0
    @types/react:  18.0.10
    react: ^18.1.0 => 18.1.0
    react-dom: ^18.1.0 => 18.1.0
    typescript:  4.7.2

alexfauquette:
The stackblitz linkes take time to open, but this bug is simple to reproduce. You can use a codesandbox link of the documentation demo:

https://569mbm.csb.app/

Probably the autosizer that does not catch this specific resizing of the window

[...]

PR #5566, opened 2022-07-22
[DataGrid] Fix container width change on React 18

### LINKED ISSUE #5085: [DataGrid] tabs switching breaks DataGrid columns.
### Duplicates

### Latest version

### Current behavior 😯

I have 3 datagrid, each one is in it's own tab. Sometimes when i switch to a different tab, only 3 columns show up. When i open the 'Show Columns' menu, all the columns appear to be switched on but only 3 can be seen on the data grid.

### Expected behavior 🤔

show all columns normally.

### Steps to reproduce 🕹

[https://codesandbox.io/s/basictabs-demo-material-ui-forked-w0vq5n?file=/demo.tsx](https://codesandbox.io/s/basictabs-demo-material-ui-forked-w0vq5n?file=/demo.tsx)

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #5085 

Related to #5192

I saw this bug when working in #4155 and one of the tests was failing with React 18 but passing with React 17. By debugging, I noticed that `rootRef.current` was null when the `resize` event is fired. I suppose this is happening because the event is triggered while the ref is unmounted. With React 17, the event is fired after the ref is mounted.

Before: https://codesandbox.io/s/datagriddemo-demo-mui-x-forked-hw2hd1?file=/demo.tsx
After: https://codesandbox.io/s/datagriddemo-demo-mui-x-forked-s2p48r?file=/package.json

How to reproduce:

msedge_bl8ELjzZqH

### DIFF
--- a/packages/grid/x-data-grid/src/tests/layout.DataGrid.test.tsx
+++ b/packages/grid/x-data-grid/src/tests/layout.DataGrid.test.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 // @ts-ignore Remove once the test utils are typed
 import { createRenderer, screen, ErrorBoundary, waitFor } from '@mui/monorepo/test/utils';
-import { SinonStub, stub, spy } from 'sinon';
+import { stub, spy } from 'sinon';
 import { expect } from 'chai';
 import {
   DataGrid,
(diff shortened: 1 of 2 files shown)


packages/grid/x-data-grid/src/tests/layout.DataGrid.test.tsx
```diff
@@ -59,14 +59,10 @@ describe('<DataGrid /> - Layout & Warnings', () => {
     });
 
     it('should resize the width of the columns', async () => {
-      // Using a fake clock also affects `requestAnimationFrame`.
-      // Calling clock.tick() should call the callback passed, but it doesn't work.
-      stub(window, 'requestAnimationFrame').callsFake((fn: any) => fn());
-      stub(window, 'cancelAnimationFrame');
```
m4theushw, line 65: Mocking `requestAnimationFrame` doesn't feel right so I inverted the logic: use real clock by default and fake clock on demand.

PR #5569, opened 2022-07-22
[pickers] Prevent daysweek transformation

### LINKED ISSUE #5330: Week days name in jalali calendar

### LINKED ISSUE #5541: [pickers] Day of the week title (PickersCalendarDayHeader) is forced to truncate to the first char
### Current behavior 😯
The titles of the days of the week in the DayPicker are truncated to the first character, regardless of what the localized helper `getWeekdays` returns.

`children: day.charAt(0).toUpperCase()`

### Expected behavior 🤔
I propose to add an option to disable truncating of days titles or the ability to pass a custom `PickersCalendarDayHeader` component

rajdee:
Fixes #5330
Fixes #5541 

Add a `preventDaysWeekTransformation` prop to DatePickers API to prevent any transformation of calendar weekday titles.
Thus we can change the name of the days of the week using styles or by extending the adapter function.

Before:
Untitled-1

After:
after


alexfauquette:
This address the same issues as #5373 but with a different approach

Both are nonbreaking changes. They only add an option to allow more than one character

- This one is easier to understand for dev, seems it's a boolean
- #5373 is more flexible since it adds a callback to allow custom formating of the day without having to modify the library or the adapter

rajdee:
@alexfauquette Thank you for your comment!
I wanted to simplify the interface as much as possible so that there were more chances that this PR was accepted :)
Ideally, I would like to add the ability to use a custom component with the original array of weekday names as prop

m4theushw:
I'm in favor of going with the approach from #5373. If we only add a new prop, users will still not be able to fully customize the header labels. But adding both (new prop to disable the transformation and the prop to customize the formatter) it may pollute the component's interface.

m4theushw:
Superseded by #5373, but thanks for the contribution.

PR #5651, opened 2022-07-29
[pickers] Fix mobile picker not opening on label click

### LINKED ISSUE #5395: [MobileDatePicker] Picker does not open on first click when value is null and clicking on the label
### Duplicates

### Latest version

### Current behavior 😯

MobileDatePicker does not open on the first TextField click when value is null, only after multiple clicks on TextField.

<MobileDatePicker
      label="my-date"                      
      inputFormat="DD/MM/YYYY"
      value={null}                                                  
      DialogProps={{ className: 'mui-datepicker' }}
      onChange={(date) => this.handleChange('my-date', date)}
      onOpen={this.handleOpen}
      renderInput={(params) => <TextField fullWidth {...params} />}
  />

### Expected behavior 🤔

It's supposed to open on first TextField click.

### Steps to reproduce 🕹

Steps:

1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #5395 

Applied the least intrusive fix to fix the mentioned issue in Mobile Chrome.

Based on the fact that Firefox has no problems with this we could say it's a Chrome bug.
But the fact, that the `label` which is being rendered above the actual Date/Time Picker input [has explicit `z-index: 1` set](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/InputLabel/InputLabel.js#L83) breaks the stacking context logic and puts the `label` on top of `input`, which is the component below the label in the DOM Tree.

There are other solutions which can also solve this issue:
- using `onMouseDown` instead of [`onClick`](https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/internals/components/PureDateInput.tsx#L162)
- overriding `z-index` on the label itself could also work as autocomplete is turned off in these cases anyways
- setting `cursor:pointer` on the `label`, but this cases potentially unwanted behaviour on desktop with `mobile` picker

I'm free to suggestions if you think that any of above or even a different solution is better in this case.

Preview: https://deploy-preview-5651--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#responsiveness


alexfauquette:
It seems to be a good workaround 

@siriwatknp I'm wondering if there is beter way to solve this problem or if the fix should be implemented in mui-core. The problem is the following:

A modal is open when user clicks on the `input` of the `TextField` (I assume it's to avoid listening to click on the `endAdorment`)
On mobile, when the user clicks on the `label`, the event does not reach the `input` component and so the modal is not open

packages/x-date-pickers/src/internals/components/PureDateInput.tsx
```diff
@@ -140,6 +140,10 @@ export const PureDateInput = React.forwardRef(function PureDateInput<TInputDate,
     () => ({
       ...InputProps,
       readOnly: true,
+      // explicit `z-index` on input makes sure the input is above label in the stacking context
+      // `filled` and `outlined` variants of `InputLabel` having explicit `z-index: 1` forces an override here
+      // this solves a bug uncovered only in Mobile (touch input) Chrome
+      sx: { zIndex: 1 },
```
siriwatknp, line 146: From the [preview](https://deploy-preview-5651--material-ui-x.netlify.app/x/react-date-pickers/date-picker/), I don't see that the `zIndex` is applied, do I miss something?

LukasTy: It is being applied to the `div` wrapping this input.
Screenshot 2022-08-23 at 11 54 56

siriwatknp, line 146:
```suggestion
      sx: [{ zIndex: 1 }, ...Array.isArray(InputProps.sx) ? InputProps.sx : [InputProps.sx]],
```
https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop

siriwatknp: I got a double `z-index`🤔 in https://deploy-preview-5651--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#responsiveness
image

LukasTy: Well, that is very peculiar... 🤔 
Which browser is it?

siriwatknp:
@LukasTy I would say that this is an okay workaround specifically for the date picker (because it is a read-only input). I have to find out why the label has `z-index: 1` in the first place.

packages/x-date-pickers/src/internals/components/PureDateInput.tsx
LukasTy: FYI, I did replicate the double application of any `sx` prop passed after syncing with origin (the latest commit merging master into this branch).
Investigating.

siriwatknp:
@oliviertassinari thanks for the info. let's try the `onMouseDown`!

LukasTy:
Thank you @oliviertassinari for providing valuable insight. 🙇 
I haven't thought about breaking such cases.
Did some more testing and `onMouseDown` seems to cover all bases, it correctly works on both emulation and actual mobile device. 👍

flaviendelangle (approved):
The solution makes sense to me :+1:

packages/x-date-pickers/src/MobileDatePicker/MobileDatePicker.test.tsx
```diff
@@ -206,7 +206,7 @@ describe('<MobileDatePicker />', () => {
       />,
     );
     const start = adapterToUse.date();
-    fireEvent.click(screen.getByRole('textbox'));
+    fireEvent.mouseDown(screen.getByRole('textbox'));
```
oliviertassinari, line 209: For context, in the Select component, we have seen a few developer struggle with figuring out how to open the popup once we moved from onClick to onMouseDown. I don't remember how we fixed this in the Select and in the Autocomplete, but anyway, it's for later, once developers open issues about it.

LukasTy: Using a more [abstract API](https://testing-library.com/docs/user-event/convenience#click) (higher-level) for writing tests might help. 🤔

flaviendelangle: Related to https://github.com/mui/mui-x/pull/5920

I believe our higher level API is `userEvent.mousePress`

PR #5668, opened 2022-07-31
[DataGrid] Improve Korean (ko-KR) locale

Einere:
update ko-KR translation for #3211


alexfauquette (approved):
Thanks for your contribution 🙏

Issue #5590, opened 2022-07-25
5.14.0 overlay change adds `pointerEvents: none` which breaks buttons inside no rows overlay


philjones88:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

This change:

https://github.com/mui/mui-x/commit/aef8033fdaa0012a7eb4751cc5604333b46f3f63

Impacts our custom `components` `NoRowsOverlay` component that has a button in it to trigger the filter panel to show.

With `pointer-events: none` and I can't see a way to customize it, it is a breaking change in functionality for us from 5.13.x

### Expected behavior 🤔

Not have `pointer-events: none` or a way to customize the behaviour.

### Steps to reproduce 🕹

Link to live example:

Steps:
1.
2.
3.
4.

### Your environment 🌎

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  System:
    OS: macOS 11.6.6
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Edge: Not Found
    Firefox: 102.0.1
    Safari: 15.5
  npmPackages:
    @emotion/react: ^11.8.2 => 11.9.3
    @emotion/styled: ^11.8.1 => 11.9.3
    @mui/icons-material: ^5.8.4 => 5.8.4
    @mui/lab: ^5.0.0-alpha.91 => 5.0.0-alpha.91
    @mui/material: ^5.9.1 => 5.9.1
    @mui/styled-engine:  5.8.7
    @mui/system: ^5.9.1 => 5.9.1
    @mui/utils:  5.9.1
    @mui/x-data-grid:  5.14.0
    @mui/x-data-grid-pro: ^5.14.0 => 5.14.0
    @mui/x-license-pro: ^5.14.0 => 5.14.0
    @types/react: ^17.0.39 => 17.0.47
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
    typescript: ^4.6.3 => 4.7.4
```
</details>

### Order ID 💳 (optional)

44827

alexfauquette:
I do not see a way to override the props of the wrapper overlay and the style is hardcoded in the element.

@DanailH is the pointer event property necessary for fixing the position?

[...]

PR #5674, opened 2022-08-01
[DataGrid] Fix `GridOverlays` bypassing pointer events

philjones88:
Fixes https://github.com/mui/mui-x/issues/5590

### DIFF
--- a/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
+++ b/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
@@ -45,7 +45,6 @@ function GridOverlayWrapper(props: React.PropsWithChildren<{}>) {
         top: headerHeight,
         bottom: height === 'auto' ? 0 : undefined,
         zIndex: 4, // should be above pinned columns, pinned rows and detail panel
-        pointerEvents: 'none',
       }}
       {...props}
     />
--- a/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
+++ b/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
@@ -44,7 +44,6 @@ function GridOverlayWrapper(props: React.PropsWithChildren<{}>) {
         position: 'absolute',
         top: headerHeight,
         bottom: height === 'auto' ? 0 : undefined,
-        zIndex: 4, // should be above pinned columns, pinned rows and detail panel
       }}
       {...props}
     />
--- a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
@@ -1,14 +1,23 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import CircularProgress from '@mui/material/CircularProgress';
+import { styled } from '@mui/material/styles';
 import { GridOverlay, GridOverlayProps } from './containers/GridOverlay';
+
+const GridLoadingOverlayRoot = styled(GridOverlay)({
+  position: 'absolute',
+  top: 0,
+  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
+  width: '100%',
+  pointerEvents: 'none',
+});
 
 const GridLoadingOverlay = React.forwardRef<HTMLDivElement, GridOverlayProps>(
   function GridLoadingOverlay(props, ref) {
     return (
-      <GridOverlay ref={ref} {...props}>
+      <GridLoadingOverlayRoot ref={ref} {...props}>
         <CircularProgress />
-      </GridOverlay>
+      </GridLoadingOverlayRoot>
     );
   },
 );
--- a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
@@ -1,23 +1,14 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import CircularProgress from '@mui/material/CircularProgress';
-import { styled } from '@mui/material/styles';
 import { GridOverlay, GridOverlayProps } from './containers/GridOverlay';
-
-const GridLoadingOverlayRoot = styled(GridOverlay)({
-  position: 'absolute',
-  top: 0,
-  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
-  width: '100%',
-  pointerEvents: 'none',
-});
 
 const GridLoadingOverlay = React.forwardRef<HTMLDivElement, GridOverlayProps>(
   function GridLoadingOverlay(props, ref) {
     return (
-      <GridLoadingOverlayRoot ref={ref} {...props}>
+      <GridOverlay ref={ref} {...props}>
         <CircularProgress />
-      </GridLoadingOverlayRoot>
+      </GridOverlay>
     );
   },
 );
--- a/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx
@@ -29,8 +29,13 @@ const GridOverlayRoot = styled('div', {
   slot: 'Overlay',
   overridesResolver: (props, styles) => styles.overlay,
 })(({ theme }) => ({
-  display: 'flex',
+  position: 'absolute',
+  top: 0,
+  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
+  width: '100%',
   height: '100%',
+  pointerEvents: 'none',
+  display: 'flex',
   alignSelf: 'center',
   alignItems: 'center',
   justifyContent: 'center',


cherniavskii:
For the context: `pointer-events: none` was added to avoid this regression https://github.com/mui/mui-x/pull/4405 after we changed zIndex of overlay container in https://github.com/mui/mui-x/pull/5558

philjones88:
@cherniavskii I'm not sure what to suggest then, the change to add `pointerEvents` causes other regressions.

cherniavskii:
What we can do to solve both issues is to:
- revert #5558
- apply `zIndex` and `pointerEvents` styles only to default error overlay

This diff should do the job:

```diff
diff --git a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
index 8e7e4ea35..55ef3921f 100644
--- a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
@@ -1,14 +1,23 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import CircularProgress from '@mui/material/CircularProgress';
+import { styled } from '@mui/material/styles';
 import { GridOverlay, GridOverlayProps } from './containers/GridOverlay';
 
+const GridLoadingOverlayRoot = styled(GridOverlay)({
+  position: 'absolute',
+  top: 0,
+  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
+  width: '100%',
+  pointerEvents: 'none',
+});
+
 const GridLoadingOverlay = React.forwardRef<HTMLDivElement, GridOverlayProps>(
   function GridLoadingOverlay(props, ref) {
     return (
-      <GridOverlay ref={ref} {...props}>
+      <GridLoadingOverlayRoot ref={ref} {...props}>
         <CircularProgress />
-      </GridOverlay>
+      </GridLoadingOverlayRoot>
     );
   },
 );
diff --git a/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx b/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
index c4f1fec86..ce3a618e5 100644
--- a/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
+++ b/packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
@@ -44,8 +44,6 @@ function GridOverlayWrapper(props: React.PropsWithChildren<{}>) {
         position: 'absolute',
         top: headerHeight,
         bottom: height === 'auto' ? 0 : undefined,
-        zIndex: 4, // should be above pinned columns, pinned rows and detail panel
-        pointerEvents: 'none',
       }}
       {...props}
     />
```

packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
```diff
@@ -44,8 +44,6 @@ function GridOverlayWrapper(props: React.PropsWithChildren<{}>) {
         position: 'absolute',
         top: headerHeight,
         bottom: height === 'auto' ? 0 : undefined,
-        zIndex: 4, // should be above pinned columns, pinned rows and detail panel
```
m4theushw, line 47: @cherniavskii shouldn't we keep `zIndex: 4` here? It was added in https://github.com/mui/mui-x/pull/4863. Without it, other overlays, e.g. no filter results, don't take the full height.

I added a background on purpose to one overlay. Here it's how it looks like:

image

philjones88: I could add it back, I thought moving it in the git patch proposed would be ok but the naming suggests that zIndex only applies to loading overlays...

How can we test this locally? Is thee a way to produce a custom NPM package? do a file:/ reference in my project's package.json...

LukasTy: > How can we test this locally?

@philjones88 Please refer to the [contribution guide](https://github.com/mui/material-ui/blob/HEAD/CONTRIBUTING.md#how-can-i-use-a-change-that-wasnt-released-yet) for options on how to test an unreleased package.

cherniavskii: @m4theushw nice catch!
I think we should not keep zIndex on the overlay wrapper, because there's no way to customize it.

Instead, we can move those styles to `GridOverlay` instead:
```diff
diff --git a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
index 55ef3921f..8e7e4ea35 100644
--- a/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/GridLoadingOverlay.tsx
@@ -1,23 +1,14 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import CircularProgress from '@mui/material/CircularProgress';
-import { styled } from '@mui/material/styles';
 import { GridOverlay, GridOverlayProps } from './containers/GridOverlay';
 
-const GridLoadingOverlayRoot = styled(GridOverlay)({
-  position: 'absolute',
-  top: 0,
-  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
-  width: '100%',
-  pointerEvents: 'none',
-});
-
 const GridLoadingOverlay = React.forwardRef<HTMLDivElement, GridOverlayProps>(
   function GridLoadingOverlay(props, ref) {
     return (
-      <GridLoadingOverlayRoot ref={ref} {...props}>
+      <GridOverlay ref={ref} {...props}>
         <CircularProgress />
-      </GridLoadingOverlayRoot>
+      </GridOverlay>
     );
   },
 );
diff --git a/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx b/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx
index 789704435..77146b237 100644
--- a/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx
+++ b/packages/grid/x-data-grid/src/components/containers/GridOverlay.tsx
@@ -29,8 +29,13 @@ const GridOverlayRoot = styled('div', {
   slot: 'Overlay',
   overridesResolver: (props, styles) => styles.overlay,
 })(({ theme }) => ({
-  display: 'flex',
+  position: 'absolute',
+  top: 0,
+  zIndex: 4, // should be above pinned columns, pinned rows and detail panel
+  width: '100%',
   height: '100%',
+  pointerEvents: 'none',
+  display: 'flex',
   alignSelf: 'center',
   alignItems: 'center',
   justifyContent: 'center',
```

This would make default overlays taking full height while allowing to scroll horizontally (for example to access column used for filtering that is not in the viewport).
But most importantly, it would allow to implement custom overlays with interactive elements while avoiding `pointerEvents: none` (so it won't cause another issue like #5590).

cherniavskii: I've tried to push the proposed changes to this branch to make sure there's no visual regressions, but turned out I don't have the permission to push to https://github.com/VQComms/mui-x/tree/remove-grid-overlays-pointer-events-style

@philjones88 could you push the changes so we trigger the CI?

philjones88: I've applied the git diff above, updated my fork and rebased this branch onto master latest.

cherniavskii: @philjones88 
Thanks! Can you rebase again to restart the CI? The failure in Argos CI looks like a false negative

philjones88: @cherniavskii I pushed an empty commit to trigger the CI. The branch was already up to date with master. Seems to have all passed!

cherniavskii (approved):

packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
```diff
@@ -44,8 +44,6 @@ function GridOverlayWrapper(props: React.PropsWithChildren<{}>) {
         position: 'absolute',
```
m4theushw, line 44: We removed `pointerEvents: none` but this was not sufficient to fix #5725. Since we didn't add `zIndex`, the overlay stays behind the virtual scroller because it has `position: relative`, while the overlay has `position: absoulte`. We need to put the overlay on top of everything to fix #5725 too

```suggestion
        position: 'relative',
```

In https://codesandbox.io/s/mui-x-norowsoverlay-bug-forked-g6npfv?file=/src/App.tsx I still can't click the button.

m4theushw: One solution that would keep https://github.com/mui/mui-x/blob/master/test/e2e/index.test.ts#L330 passing is to render the overlays inside the virtualization component.

m4theushw (approved):
Looks good. It's what we can do given the constraints we have. In v6 we can rearrange the overlay to render correctly without additional styling.

packages/grid/x-data-grid/src/components/base/GridOverlays.tsx
cherniavskii: Maybe it will be possible to do without breaking changes - I'll take a look

cherniavskii:
Thanks @philjones88

PR #5686, opened 2022-08-02
[DataGrid] Preserve cell mode when entering edit mode while commiting

### LINKED ISSUE #5540: [data grid] Grid cells inconsistent state when trying to edit cell validating it with a server
### Duplicates

### Current behavior 😯

If we go to the documentation example https://mui.com/x/react-data-grid/editing/#persistence and edit a cell, and immediately before the server responds try to edit another one, the second cell will remain in edit mode forever (until the Datagrid component is unmounted and mounted again).

### Expected behavior 🤔

The second cell should not be able to become editable or it should become editable but behave as the first one (don't remain in editable mode forever).

### Steps to reproduce 🕹

Link to live example:

Steps:
1. Go to the https://mui.com/x/react-data-grid/editing/#persistence example
2. Double click in one cell, edit it a lose focus.
3. Immediately after that (and before the server responded the first request) double click in another cell.
4. The second cell remains in edit mode forerever.

### Context 🔦

_No response_

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

m4theushw:
Before: https://codesandbox.io/s/serversidepersistence-demo-mui-x-forked-jeqzyz?file=/demo.tsx
After: https://codesandbox.io/s/serversidepersistence-demo-mui-x-forked-8yxugo?file=/demo.tsx

Fixes #5540 

The bug in #5540 happens because in https://github.com/mui/mui-x/blob/689cbfcaf676fabac5f02fcad46982c6965b4291/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts#L363-L368, when we create the new object for `cellModesModel`, the value that is spread below is outdated, meaning that it's empty, even though another cell entered the edit mode while waiting for `processRowUpdate`.
  
https://github.com/mui/mui-x/blob/689cbfcaf676fabac5f02fcad46982c6965b4291/packages/grid/x-data-grid/src/hooks/features/editRows/useGridCellEditing.new.ts#L271-L273

I fixed this problem by introducing a `cellModesModelRef` ref and also keeping the state. The ref ensures that the value will be updated immediately, not having to wait for the next rerender. The state, in the other hand, is used to trigger the rerender.


packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
```diff
@@ -56,6 +56,15 @@ export const useGridFocus = (
         return;
       }
 
+      if (focusedCell) {
+        // There's a focused cell but another cell was clicked
+        // Publishes an event to notify that the focus was lost
+        apiRef.current.publishEvent(
+          'cellFocusOut',
```
m4theushw, line 63: Adding the ref fixed the main problem but created another one. When the server responds, by default, the grid focus the cell that is below the cell being edited, this means that the new cell being edited (that one that entered edit mode while waiting `processRowUpdate`) lost focus. When a cell loses focus it should commit the value but it doesn't do this because `cellFocusOut` is only published when there's a click, not when the focus is changed programmatically, so I'm also publishing the event here now. 

This may create an unwanted behavior because it commits the next cell when the server responds because of the first. However, this can be solved in two ways:
- 1st option: Use row editing mode
- 2nd option: Cancel the default behavior in `onCellEditStop` and manually call `apiRef.current.stopCellEditMode` without passing which cell to focus after.

packages/grid/x-data-grid-pro/src/tests/editComponents.DataGridPro.old.test.tsx
```diff
@@ -852,14 +852,15 @@ describe('<DataGridPro /> - Edit Components', () => {
...
+      expect(cell).to.have.class('MuiDataGrid-cell--editing');
       await new Promise((resolve) => nativeSetTimeout(resolve)); // Wait for promise
 
       expect(preProcessEditCellProps.callCount).to.equal(1);
       expect(preProcessEditCellProps.lastCall.args[0].props).to.deep.equal({
         value: true,
         isValidating: true,
       });
-      expect(cell).to.have.class('MuiDataGrid-cell--editing');
+      expect(cell).not.to.have.class('MuiDataGrid-cell--editing');
```
m4theushw, line 863: The old assertion was wrong. After double-clicking another cell, the first cell should leave the edit mode. This is how it works outside a test environment. With the changes I did here this problem became apparent.

Issue #5606, opened 2022-07-26
[data grid] Remove the editing experimental features and make them enabled by default


joserodolfofreitas:
# Summary 

Breaking change for v6.
For more context: #3287 

  https://github.com/mui/mui-x/blob/b778028da23e7455c985c93c53366ab63848ee6d/packages/grid/x-data-grid/src/models/props/DataGridProps.ts#L34-L41

PR #5728, opened 2022-08-07
[docs] Use new edit API in the introduction demos

oliviertassinari:
Considering how much we are burying the demo of the legacy editing API, and that the legacy API is buggy, e.g. with the <kbd>Escape</kbd> key:

https://user-images.githubusercontent.com/3165635/183312070-6a7daf69-1e53-40ac-b0b3-22bfa5808033.mov

https://mui.com/x/react-data-grid/#mit-version

I believe it would make more sense to demo it on the introduction demos. I believe these demos are used by developers to judge whether they will use or not the component.

One step in this direction #5606.

Preview: https://deploy-preview-5728--material-ui-x.netlify.app/x/introduction/

### DIFF
--- a/docs/data/data-grid/demo/FullFeaturedDemo.js
+++ b/docs/data/data-grid/demo/FullFeaturedDemo.js
@@ -311,6 +311,7 @@ export default function FullFeaturedDemo() {
         checkboxSelection
         disableSelectionOnClick
         rowThreshold={0}
+        experimentalFeatures={{ newEditingApi: true }}
         initialState={{
           ...data.initialState,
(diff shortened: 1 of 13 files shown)


m4theushw:
Editing the Status column in https://deploy-preview-5728--material-ui-x.netlify.app/x/react-data-grid/demo/ is throwing an error because the API method changed and this is a custom edit component. I'll take care of this...

oliviertassinari:
@m4theushw Nice catch

m4theushw (approved):
I pushed 5d7455c fixing the custom edit component.

A follow up may be to also use the new API in the demo in https://mui.com/x/

PR #5730, opened 2022-08-07
[license] Only log an error type once

oliviertassinari:
The goal is to reduce the spam of duplicated error messages, it's one chunk of #1364. This is equivalent to [what React does](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/packages/react-dom/src/shared/ReactDOMUnknownPropertyHook.js#L29).
Since React 18, opening any commercial demo, leads to a duplicated error message, this is not great, e.g. https://codesandbox.io/s/1cm3xe?file=/demo.tsx, there should only be one.

**Before** (8)

Screenshot 2022-08-08 at 01 12 35

**After** (2)

Screenshot 2022-08-08 at 01 12 54

On http://0.0.0.0:3000/x/, it's insane, 161 errors 🙃.

https://deploy-preview-5730--material-ui-x.netlify.app/x/react-data-grid/#commercial-version, https://codesandbox.io/s/l7tnq4?file=/demo.tsx

PR #5731, opened 2022-08-07
[license] Give more context in the missing license

oliviertassinari:
This is related to #5730. I'm trying to improve the DX from https://codesandbox.io/s/1cm3xe?file=/demo.tsx.

**Before**

Screenshot 2022-08-08 at 01 26 43

**After**

Screenshot 2022-08-08 at 01 26 31

### DIFF
--- a/packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
+++ b/packages/x-license-pro/src/useLicenseVerifier/useLicenseVerifier.ts
@@ -32,6 +32,7 @@ export function useLicenseVerifier(
       ? ['premium']
       : ['pro', 'premium'];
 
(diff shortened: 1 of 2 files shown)


LukasTy (commented):
LGTM, but what do you think about adding similar additions to other two (invalid and expired) license status errors?

LukasTy (approved):

oliviertassinari:
@LukasTy I think that we have room to continue these types of improvements. It's one iteration I had the time to do during the weekend. #1364 is kind of an umbrella issue for this topic.

oliviertassinari:
@LukasTy I have updated #1364 to better keep track of the opportunities that we have

PR #5740, opened 2022-08-09
[pickers] fix to trigger `onChange` when clearing not empty value

### LINKED ISSUE #5690: onChange is not triggered when clearing an invalid date
### Duplicates

### Latest version

### Current behavior 😯

When you enter an invalid date (such as 13/02/2022 for MM/DD/YYYY) and then press the clear button, no callback is triggered.

### Expected behavior 🤔

A callback should be triggered in such a way that a controlled input would have the error state and value cleared.

### Steps to reproduce 🕹

Link to live example:
https://codesandbox.io/s/objective-water-g8s2ln?file=/src/App.js
Steps:
1. Type 13/02/2022 into the DateInput
2. Open the calendar
3. Press the clear button in the calendar

### Context 🔦

I'm trying to control a date input in a form, and users of the form find the current behavior counterintuitive.

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #5690 

Force trigger an `onChange` call when clicking `clear` button with not empty `value`.
Covers case, when date picker has `Invalid date` due to manual entry, but internally the state is saved as `null` already.

Simply adding a check in `setDate` function causes inconsistent behaviour, where simply closing the picker would also reset an invalid value.

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
+++ b/packages/x-date-pickers/src/internals/hooks/usePickerState.ts
@@ -277,7 +277,12 @@ export const usePickerState = <TInputValue, TValue, TDate>(
       },
       onAccept: () => {
         // Set all date in state to equal the current draft value and close picker.
-        setDate({ value: dateState.draft, action: 'acceptAndClose' });
+        setDate({
+          value: dateState.draft,
+          action: 'acceptAndClose',
+          // force `onChange` in cases like input (value) === `Invalid date`
+          forceOnChangeCall: !valueManager.areValuesEqual(utils, value as any, parsedDateValue),
+        });
       },
       onDismiss: () => {
         // Set all dates in state to equal the last committed date.
@@ -294,16 +299,7 @@ export const usePickerState = <TInputValue, TValue, TDate>(
         setDate({ value: valueManager.getTodayValue(utils), action: 'acceptAndClose' });
       },
     }),
-    [
-      isOpen,
-      setDate,
-      valueManager,
-      value,
-      dateState.draft,
-      dateState.committed,
-      dateState.resetFallback,
-      utils,
-    ],
+    [setDate, isOpen, utils, dateState, valueManager, value, parsedDateValue],
   );
 
   // Mobile keyboard view is a special case.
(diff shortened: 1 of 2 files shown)


packages/x-date-pickers/src/internals/hooks/usePickerState.ts
```diff
@@ -257,7 +264,16 @@ export const usePickerState = <TInputValue, TValue, TDate>(
...
-        setDate({ value: valueManager.emptyValue, action: 'acceptAndClose' });
+        setDate({
+          value: valueManager.emptyValue,
+          action: 'acceptAndClose',
+          // force `onChange` in cases like input (value) === `Invalid date`
+          forceOnChangeCall: !valueManager.areValuesEqual(
+            utils,
+            value as any,
+            valueManager.emptyValue,
+          ),
```
alexfauquette, line 275: Should we do the same with `onAccept`?

https://codesandbox.io/s/summer-sound-rn9osw?file=/src/App.js

packages/x-date-pickers/src/internals/hooks/usePickerState.ts
```diff
@@ -278,7 +294,16 @@ export const usePickerState = <TInputValue, TValue, TDate>(
...
+    [
+      isOpen,
+      setDate,
+      valueManager,
+      value,
+      dateState.draft,
+      dateState.committed,
+      dateState.resetFallback,
+      utils,
+    ],
```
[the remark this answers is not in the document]
LukasTy: There should be difference, especially in this case.
I've changed it to this style only because it was the change that the IDE has done as auto suggestion.
WiIl add `value` manually for cleanliness.

packages/x-date-pickers/src/internals/hooks/usePickerState.ts
LukasTy: Do you mean that clicking `ok` (accept) should also revert date to empty/null if it is invalid?

LukasTy: Thanks for noting this case. 🙏 
Updated.
Although, I'm no longer happy with this `memo`ized `wrapperProps` as it is changing on every input value change, this forces the component to re-render even though it essentially **has** to re-render only on static picker open (to update the selected value).

alexfauquette: it there is not value selected, I assume that accepting no value means cleaning the field.

But I'm thinking a bit too much about what is logical instead of what user want. I assume nobody click on accept if they do not have selected a date

alexfauquette (approved):
Sounds like a nice tradeoff 👍

PR #5773, opened 2022-08-12
[DateRangePicker] disable day on date range

### LINKED ISSUE #5769: [DateRangePicker] Separate "shouldDisableDate" for start date and end date
### Duplicates

### Latest version

### Summary 💡

Would be nice if you could specify separate **shouldDisableDate** for start date and end date. 

For example, if you have a "Round trip" in a booking system where you have different timetables that runs on different days for out and home. 

For example. Ship that travels from A to B on every monday, wednesday, friday and B to A on every tuesday, thursday, saturday. So when start is active you want to disable dates that are on tuesday, thursday, saturday. And vice versa when end is active. 

Now we only have one **shouldDisableDate** for both start and end date which make it hard to accomplish this. 

### Examples 🌈

_No response_

### Motivation 🔦

The ability to have separate shouldDisableDate on start and end date.

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #5769

I tried a POC for #5769 to add a parameter `'start'` and `'end'` for now only on `DesktopDateRangePicker`. It will need some cleaning but seems to work.


LukasTy:
The POC looks superb 🙌 
I do believe that this should easily solve the original issue.
But as Flavian mentioned in it, it would be best to also add this option to `shouldDisableMonth` and `shouldDisableYear`. 👎

packages/x-date-pickers/src/internals/hooks/validation/models.ts
```diff
@@ -30,9 +30,10 @@ export interface DayValidationProps<TDate> {
    * Disable specific date. @DateIOType
    * @template TDate
    * @param {TDate} day The date to test.
+   * @param {string} position The date to test, 'start' or 'end' for range pickers.
    * @returns {boolean} Returns `true` if the date should be disabled.
    */
-  shouldDisableDate?: (day: TDate) => boolean;
+  shouldDisableDate?: (day: TDate, position?: 'start' | 'end') => boolean;
```
flaviendelangle, line 36: It would be nice to only add it to the range interfaces
But that may add a lot of complexity :/

LukasTy: Yes, I wanted to comment on this as well. 🤔 
Seems a bit counter intuitive to provide optional field for every picker, when only `range` ones have this field present.  
I'll look into whether it would be feasible to have separate interfaces for range and other pickers.

alexfauquette: Yes, because this is propagated nearly everywhere. But I can give a second try to override it properly

alexfauquette: For a reason I ignore, it seems to work without any complex modification 🤯

I'm happy and suspicious at the same time because docs:API generate the expected output (one parameter for basic input, and two for range inputs) but I did not had to `Omit` `shouldDisableDate` anywhere 🤔

flaviendelangle: You have to omit if you want to make this param non-nullable for range pickers

```diff
diff --git a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
index 3ec1c82fa..f80b83938 100644
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
@@ -40,7 +40,7 @@ export interface ExportedDateRangePickerViewProps<TDate>
   extends ExportedDesktopDateRangeCalendarProps<TDate>,
     Omit<
       ExportedCalendarPickerProps<TDate>,
-      'onYearChange' | 'renderDay' | keyof BaseDateValidationProps<TDate>
+      'onYearChange' | 'renderDay' | keyof BaseDateValidationProps<TDate> | 'shouldDisableDate'
     > {
   /**
    * Overrideable components.
@@ -81,7 +81,7 @@ export interface ExportedDateRangePickerViewProps<TDate>
    * @param {string} position The date to test, 'start' or 'end'.
    * @returns {boolean} Returns `true` if the date should be disabled.
    */
-  shouldDisableDate?: (day: TDate, position?: 'start' | 'end') => boolean;
+  shouldDisableDate?: (day: TDate, position: 'start' | 'end') => boolean;
 }
 
 interface DateRangePickerViewProps<TInputDate, TDate>
diff --git a/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts b/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
index 969eb1f59..c61bbcce3 100644
--- a/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
+++ b/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
@@ -5,14 +5,12 @@ import {
   DateValidationError,
   validateDate,
   BaseDateValidationProps,
-  DayValidationProps,
 } from '@mui/x-date-pickers/internals';
 import { isRangeValid, parseRangeInputValue } from '../../utils/date-utils';
 import { DateRange } from '../../models';
 
 export interface DateRangeValidationProps<TInputDate, TDate>
-  extends DayValidationProps<TDate>,
-    Required<BaseDateValidationProps<TDate>>,
+  extends Required<BaseDateValidationProps<TDate>>,
     ValidationProps<DateRangeValidationError, DateRange<TInputDate>> {
   /**
    * Disable specific date. @DateIOType
@@ -21,7 +19,7 @@ export interface DateRangeValidationProps<TInputDate, TDate>
    * @param {string} position The date to test, 'start' or 'end'.
    * @returns {boolean} Returns `true` if the date should be disabled.
    */
-  shouldDisableDate?: (day: TDate, position?: 'start' | 'end') => boolean;
+  shouldDisableDate?: (day: TDate, position: 'start' | 'end') => boolean;
 }
 
 export const validateDateRange: Validator<any, DateRangeValidationProps<any, any>> = ({
```

_____

I just have one issue which is that the doc generation looses the enum.
Most likely during the proptype generation phase:

```
@param {string} position The date to test, 'start' or 'end'.
```

But that's a different topic that probably impact other parts of our codebase and should be adressed

flaviendelangle: By the way, I would be in favor of having a `DayRangeValidationProps` containing only `shouldDisableDate`
And have the two interface with the range `shouldDisableDate` extend it.

alexfauquette: >I just have one issue which is that the doc generation looses the enum.
Most likely during the proptype generation phase:
`@param {string} position The date to test, 'start' or 'end'.`

The `string` does not come from the typescript definition but from the comment line
```ts
/**
 * @param {string} position The date to test, 'start' or 'end'.
 * /
 ```
 
I did that because using  `@param {'start' | 'end'}` or 
```ts
type RangePosition = 'start' | 'end'
/**
 * @param {RangePosition}
 */
 ```
 ends up with an error in the doc generation saying the Template Literal Types are not supported by the documentation build, so instead of fixing the script I put `string` and listed the values.

LukasTy (approved):
LGTM. Very nice final result! 🚀 💯 ❤️

flaviendelangle (approved):

PR #5784, opened 2022-08-13
[docs] Fix default value of DataGrid's logLevel prop to false

HwangTaehyun:
Docs said that the default value of DataGrid's logLevel prop is "debug", but its default value is false. 
Please refer to https://github.com/mui/mui-x/blob/a55478c44364734517f7517072cbba50a9360131/packages/grid/x-data-grid/src/models/props/DataGridProps.ts#:~:text=logLevel%3A%20keyof%20Logger%20%7C%20false%3B.

So, if logLevel is not set to debug, DataGrid logs are not visible. The docs need to be modified. @oliviertassinari

### DIFF
--- a/docs/pages/x/api/data-grid/data-grid-premium.json
+++ b/docs/pages/x/api/data-grid/data-grid-premium.json
@@ -124,7 +124,7 @@
         "name": "enum",
         "description": "'debug'<br>&#124;&nbsp;'error'<br>&#124;&nbsp;'info'<br>&#124;&nbsp;'warn'<br>&#124;&nbsp;false"
       },
-      "default": "\"debug\""
+      "default": "\"error\" (\"warn\" in dev mode)"
     },
     "nonce": { "type": { "name": "string" } },
     "onAggregationModelChange": { "type": { "name": "func" } },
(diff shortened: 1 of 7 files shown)


oliviertassinari:
I wouldn't expect it to be `false`, so I had a closer look at the codebase, the default value seems to be

https://github.com/mui/mui-x/blob/39a919445dcabc438542896ccb416a8980dbd1c2/packages/grid/x-data-grid/src/DataGrid/useDataGridProps.ts#L66

HwangTaehyun:
Oh, yes. I think so. What do you think the best way to display the logLevel default value in the data grid docs?
It shows default is debug and is wrong.
image

And I have one more question, can you tell me why the test case fails in argo like below?
image

cherniavskii:
Hey @HwangTaehyun 
Maybe put it like this:
```
* @default "error" ("warn" in dev mode)
```

> And I have one more question, can you tell me why the test case fails in argos like below?

That's a false negative regression, just ignore it. It should pass once you push to this branch and trigger the CI pipeline to restart

HwangTaehyun:
Thanks @cherniavskii!

cherniavskii:
Thanks @HwangTaehyun!

Looks like we will need to merge https://github.com/mui/mui-x/pull/5799 first.
I'll make sure to merge this PR as soon as all checks are successful, no action needed on your side

cherniavskii (approved):

PR #5798, opened 2022-08-15
[Pickers] Add `className` prop to `PickersActionBar`

### LINKED ISSUE #5779: [DatePicker] - Overriding `PickersActionBar` with `styled` component not working
### Duplicates

### Latest version

### Current behavior 😯

Using a styled component in the `ActionBar` slot for the DatePicker `components` prop doesn't properly apply styles.

Using `import { styled } from '@mui/material/styles';` the CSS is generated but the classes are wrong. 

In the attached example you can search for the selector `nth-last-child(2)` and see that it's been generated.

The code generated is 
```
<style data-emotion="css" data-s="">.css-1uz2a6j{background-color:red;}</style>
<style data-emotion="css" data-s="">.css-1uz2a6j >*:first-child{margin-right:auto;}</style>
<style data-emotion="css" data-s="">.css-1uz2a6j >*:nth-last-child(2){margin-right:inherit;}</style>
```

But nothing in the document has the class or even prefix of `css-1uz2a6j`. The classes generated are missing the component name and slot too. The ActionBar in this case has a class like `css-hlj6pa-MuiDialogActions-root` which doesn't even match the generated prefix.

### Expected behavior 🤔

I would expect the components generated with `styled` to have their styles properly applied with proper class name.  Additionally I would expect the classes to follow the naming structure of `css-abc123-ComponentName-Slot`

### Steps to reproduce 🕹

Link to live example: https://codesandbox.io/s/actionbarcomponent-demo-mui-x-forked-whd9pu?file=/demo.tsx

Steps:
1.  Create a DatePicker component
2. Create a custom ActionBar Component using `styled` with some styles.
3. Apply the custom ActionBar.
4. See the styles are not applied

### Context 🔦

I want to apply this style to only the DialogActions for the DatePicker. The selectors I've provided work perfectly find as part of the theme but will apply to all DialogActions. Providing a `className` and writing some CSS won't work. I'm unable to provide a `className` prop to `actionBar` VIA `componentsProps` and applying a `className` to `DatePicker` is useless in this scenario since the control in question is part of a `Dialog`.

### Your environment 🌎

<details>

```
Using Chrome: 104.0.5112.79

System:
    OS: macOS 11.6.7
  Binaries:
    Node: 14.17.3 - ~/.nvm/versions/node/v14.17.3/bin/node
    Yarn: Not Found
    npm: 7.24.2 - ~/.nvm/versions/node/v14.17.3/bin/npm
  Browsers:
    Chrome: 104.0.5112.79
    Edge: Not Found
    Firefox: 103.0.2
    Safari: 15.5
  npmPackages:
    react: 17.0.2 => 17.0.2
    typescript: 4.4.3 => 4.4.3
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fix #5779 

The component did not support accepting `className` and was unable to process usages of `styled(PickersActionBar)` correctly.

Enables using `styled(PickersActionBar)` to provide custom styles

### DIFF
--- a/packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
+++ b/packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
@@ -22,10 +22,11 @@ export interface PickersActionBarProps {
   onClear: () => void;
   onCancel: () => void;
   onSetToday: () => void;
+  className?: string;
 }
 
 export const PickersActionBar = (props: PickersActionBarProps) => {
-  const { onAccept, onClear, onCancel, onSetToday, actions } = props;
+  const { onAccept, onClear, onCancel, onSetToday, actions, className } = props;
   const wrapperVariant = React.useContext(WrapperVariantContext);
 
   const localeText = useLocaleText();
@@ -67,5 +68,5 @@ export const PickersActionBar = (props: PickersActionBarProps) => {
     }
   });
 
-  return <DialogActions>{buttons}</DialogActions>;
+  return <DialogActions className={className}>{buttons}</DialogActions>;
 };
--- a/packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
+++ b/packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import Button from '@mui/material/Button';
-import DialogActions from '@mui/material/DialogActions';
+import DialogActions, { DialogActionsProps } from '@mui/material/DialogActions';
 import { useLocaleText } from '../internals/hooks/useUtils';
 import {
   WrapperVariant,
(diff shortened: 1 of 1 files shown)


packages/x-date-pickers/src/PickersActionBar/PickersActionBar.tsx
```diff
@@ -22,10 +22,11 @@ export interface PickersActionBarProps {
   onClear: () => void;
   onCancel: () => void;
   onSetToday: () => void;
+  className?: string;
 }
 
 export const PickersActionBar = (props: PickersActionBarProps) => {
-  const { onAccept, onClear, onCancel, onSetToday, actions } = props;
+  const { onAccept, onClear, onCancel, onSetToday, actions, className } = props;
```
alexfauquette, line 29: To be more generic, you can extend `DialogActionProps` in the definition of `PickersActionBarProps` and then propagate all the other props to the `DialogActions`

```suggestion
  const { onAccept, onClear, onCancel, onSetToday, actions, ...other } = props;
```

LukasTy: Thanks for a great suggestion 👍 🙇

alexfauquette (approved):

PR #5801, opened 2022-08-15
[pickers] Add `PaperContent` component slot

### LINKED ISSUE #5767: [DatePicker] Add custom component to the top of DatePicker
### Duplicates

### Latest version

### Summary 💡

[Like `ActionsBar`](https://github.com/mui/mui-x/issues/4988), I would like to add a component to the top of DatePicker.

### Examples 🌈

```tsx
import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
import { DatePicker } from '@mui/x-date-pickers/DatePicker';
import Divider from '@mui/material/Divider';

const CustomTopBar: React.FC = () => (
  <React.Fragment>
    <em>hi</em>

    <Divider />
  </React.Fragment>
);

const CustomActionBar: React.FC = () => (
  <React.Fragment>
    <Divider />

    <em>hi</em>
  </React.Fragment>
);

export default function MaterialUIPickers() {
  const [value, setValue] = React.useState<Date | null>(
    new Date('2022/08/12')
  );

  const handleChange = (newValue: Date | null) => {
    setValue(newValue);
  };

  return (
      <DatePicker
        label="Date&Time picker"
        value={value}
        onChange={handleChange}
        // components={{TopBar: CustomTopBar}}
        components={{ ActionBar: CustomActionBar }}
      />
    </LocalizationProvider>
  );
}
```

### Motivation 🔦

I want to add Text and Divider to show the tips for users.

```tsx
const Tips: React.ReactElement = (
  <>
    <em>Hi! Please select valid date below.</em>

    <Divider />
  </>
)
```

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #5767 

Add `PaperContent` slot to pickers allowing easy extension of what is rendered.

[Add documentation](https://deploy-preview-5801--material-ui-x.netlify.app/x/react-date-pickers/custom-components/#paper-content) on how to use this slot in to build date range picker shortcuts.

Add test asserting rendering.


alexfauquette:
When you add `orientation="landscape"` the result is a bit strange

image


My main concern is about generalization. The approach of adding a `TopBar` is nice for solving one user use case, but we might end up with also `BottomBar` for sure, and maybe `RightBar` and `LeftBar`.

Would it be possible to customize the root Paper such that dev could wrap the content as they want.

For example 

```jsx
const CustomPopperPaper = ({children, ...props}) => <Paper {...props}>
	<p>Helper thext</p>
	{children}
</Paper>


<DatePicker
	// ...
	component={{PopperPaper: CustomPopperPaper }}
/>
```

LukasTy:
Great suggestion on going for a more generic solution to allow passing a flexible layout @alexfauquette 
Only problem i see is that `Static` picker version does not use a `paper` and hence would need to be refactored or wouldn't bet this option. 🤔

alexfauquette:
No, I proposed `Paper` because it's the first element where it could be set. But it could be another naming or a wrapper which by default do nothing

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -80,3 +80,11 @@ This can be used in combination with `componentsProps`.
 In the example below, the actions are the same as in the section above, but they are rendered inside a menu:
 
 {{"demo": "ActionBarComponent.js"}}
+
+## Popper paper
+
+The popper paper is available on all picker components except `Static`. It adds a flexible way to extend what is rendered in the picker paper.
```
alexfauquette, line 86:
```suggestion
The popper paper is available on all desktop picker components except `Static`.
It adds a flexible way to extend what is rendered in the picker paper.  
```

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
```diff
@@ -284,6 +285,7 @@ export const PickersPopper = (props: PickerPopperProps) => {
   };
 
   const ActionBar = components?.ActionBar ?? PickersActionBar;
+  const PopperPaperContent = components?.PopperPaper || React.Fragment;
```
alexfauquette, line 288: Should we rename it to be explicit on the fact it does not override the `Paper`?
```suggestion
  const PopperPaperContent = components?.PopperPaperContent || React.Fragment;
```

LukasTy: Renamed the component to `PaperContent` and made sure it is propagated to the `Static` pickers as well. 😉

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -80,3 +80,14 @@ This can be used in combination with `componentsProps`.
 In the example below, the actions are the same as in the section above, but they are rendered inside a menu:
 
 {{"demo": "ActionBarComponent.js"}}
+
+## Paper content
+
+The paper content is available on all desktop picker components.
```
alexfauquette, line 86: I don't know what is added at the end of the line, but it creates a `<br/>` in the generated HTML which is a bit weird

image


```suggestion
The paper content is available on all desktop picker components.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -80,3 +80,14 @@ This can be used in combination with `componentsProps`.
 In the example below, the actions are the same as in the section above, but they are rendered inside a menu:
 
 {{"demo": "ActionBarComponent.js"}}
+
+## Paper content
+
+The paper content is available on all desktop picker components.  
+It adds a flexible way to extend what is rendered in the picker paper.  
+You can provide any custom component to this slot as long as it passes `children` down maintaining original picker behavior.  
+In the examples below we provide ways of implementing date range shortcuts using this slot.
```
alexfauquette, line 89:
```suggestion
It adds a flexible way to extend what is rendered in the picker paper.

You can provide any custom component to this slot as long as it passes `children` down maintaining original picker behavior.
In the examples below we provide ways of implementing date range shortcuts using this slot.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -80,3 +80,14 @@ This can be used in combination with `componentsProps`.
...
+
+{{"demo": "PaperContentComponent.js"}}
+
+{{"demo": "StaticPaperContentComponent.js"}}
```
alexfauquette, line 93: Should the two demo be merged in a single one?

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -80,3 +80,14 @@ This can be used in combination with `componentsProps`.
...
 {{"demo": "ActionBarComponent.js"}}
+
+## Paper content
+
+The paper content is available on all desktop picker components.  
+It adds a flexible way to extend what is rendered in the picker paper.  
+You can provide any custom component to this slot as long as it passes `children` down maintaining original picker behavior.  
+In the examples below we provide ways of implementing date range shortcuts using this slot.
+
+{{"demo": "PaperContentComponent.js"}}
```
alexfauquette, line 91: For Pro demo you can disable ad, avoiding the following banner to appear by using `"disableAd": true`
image

And the code is too large to display a useful preview, so you can hide it with `"defaultCodeOpen": false`
```suggestion
{{"demo": "PaperContentComponent.js", "defaultCodeOpen": false, "disableAd": true}}
```

alexfauquette (commented):
The code sounds good, maybe @flaviendelangle will  have a better idea for the naming of the slot

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
```diff
@@ -323,15 +324,17 @@ export const PickersPopper = (props: PickerPopperProps) => {
...
-              {children}
-              <ActionBar
-                onAccept={onAccept}
-                onClear={onClear}
-                onCancel={onCancel}
-                onSetToday={onSetToday}
-                actions={[]}
-                {...componentsProps?.actionBar}
-              />
+              <PaperContent>
```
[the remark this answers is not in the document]
LukasTy: As far as I've checked, I believe that only the `ActionBar` supports `componentProps` as it can accept changes in two ways:
- which actions to show via `componentProps`
- what to render via `ActionBar`

In this case, like you mentioned, we only allow user adding extra content before or after the main calendar content.

docs/data/date-pickers/custom-components/custom-components.md
LukasTy: I've added [two spaces](https://www.w3schools.io/file/markdown-line-break/#:~:text=The%20fourth%20way%2C%20use%20two%20spaces%20with%20entering%20or%20return%20keystrokes.), which creates a line break, but not a separate paragraph.

Do you disagree with the current visual formatting and would like to reduce the amount of new lines?

docs/data/date-pickers/custom-components/custom-components.md
LukasTy: How would you image them being merged?
Render both components in a single `demo`? 🤔

docs/data/date-pickers/custom-components/custom-components.md
alexfauquette: It sounds a bit like bullet points, but without the bullets

I think you can group the two first one because they answer the same question "what is it" and put the two other in a distinct paragraph since they are more about technical details

docs/data/date-pickers/custom-components/custom-components.md
alexfauquette: Yes, in other demo when we display the mobile, desktop, and static components

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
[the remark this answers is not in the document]
alexfauquette: It's a convention used by both the core and the data grid. When I can customize a slot with `components` I can also pass props with `componentsProps`. Most of the other slots are icons, and to be consistent with the data grid they should be supported too in `componentsProps`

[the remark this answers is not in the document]
flaviendelangle: > As far as I've checked, I believe that only the ActionBar supports componentProps as it can accept changes in two ways:

Indeed, and that's something we should improve.
The pickers slots are not very mature yet.
But always having the props is a nice thing.
We don't pass the props to some icons because it's useless with the MUI icons for instance, and we have someone complaining because he uses custom icon components on which it would be usefull.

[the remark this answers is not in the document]
LukasTy: I've added the `paperContent` props as suggested.
Also added a usage [example of it here](https://github.com/mui/mui-x/pull/5801/files#diff-198e8a43ff0e8f34f29cab37e8e3b3f57163ac164f0b9c9a813be213e2ea13b8R201).

LukasTy:
@flaviendelangle Do you have any further insights or suggestions?
If not, I'm ready to merge this. 🤔

docs/data/date-pickers/custom-components/PaperContentComponent.js
```diff
@@ -0,0 +1,203 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
```
flaviendelangle, line 4: We migrated to `dayjs` on our doc examples (see #5481), the diff should be pretty straightforward, you have examples in the adapter in the `addWeeks` etc... methods if you want

LukasTy: Thanks for reminding.
Did sync with master and refactor to use `dayjs` 😉

docs/data/date-pickers/custom-components/PaperContentComponent.js
```diff
@@ -0,0 +1,203 @@
...
+export default function PaperContentComponent() {
+  const [value, setValue] = React.useState([null, null]);
+  const WrappedPaperContent = React.useCallback(
+    ({ children }) => (
+      <RangeShortcutsPanel setValue={setValue}>{children}</RangeShortcutsPanel>
+    ),
+    [],
+  );
+
+  const WrappedStaticPaperContent = React.useCallback(
```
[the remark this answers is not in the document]
LukasTy: Very nice observation, thanks.
Did the refactor.

I also did generate the new docs (with slots info) after merging with master.

flaviendelangle (approved):
Looks great !

The effort to expand / document the slots is far from over but it's improving fast

packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
```diff
@@ -46,6 +46,32 @@ describe('<DesktopDatePicker />', () => {
...
+        components={{
+          PaperContent: CustomPaperContent,
+        }}
+      />,
+    );
+
+    openPicker({ type: 'date', variant: 'desktop' });
+
+    expect(screen.getByText('test custom content')).not.equal(null);
+    expect(screen.getByText('January 1970')).not.equal(null);
```
m4theushw, line 72: This test is failing on Edge: https://app.circleci.com/pipelines/github/mui/mui-x/22602/workflows/82b0a275-340e-426d-88e6-337221d41e49/jobs/130351

Using `screen.getByText` is sometimes problematic because Edge has a different timezone, on BrowserStack, than other browsers. Is there a way to use another text that doesn't contain the full month + year? In a different timezone it might not be January. If not, then you need to first convert the date to a string so it can be asserted.

We had a similar problem in the past about this. Maybe https://github.com/mui/mui-x/pull/5221 helps.

LukasTy: Thanks for pointing this out.
Have we considered ["locking" in the Timezone](https://blog.liprandi.net/2020/10/running-karma-in-specific-timezone.html) to have consistent results across the board?
Or do we want this dependency on various time zones to catch (un)expected issues?

LukasTy: Created [a PR](https://github.com/mui/mui-x/pull/5955) to fix it.

PR #5820, opened 2022-08-17
[DatePicker] Improve focus management

### LINKED ISSUE #4597: [DatePicker] Cannot select date after changing month with keyboard controls

### LINKED ISSUE #4900: StaticDatePicker accessibility issues

### LINKED ISSUE #5534: is there a way to enable day selection/highlighting via keyboard when using the StaticDateRangePicker?
### Order ID 💳

3197

### Duplicates

### Latest version

### The problem in depth 🔍

we are using the StaticDateRangePicker and accessibility requires user ability to navigate through each day via keyboard.

I tried https://mui.com/x/react-date-pickers/date-range-picker/ and was not able to achieve keyboard navigation on calendars to select a date

### Your environment 🌎

<details>

```
npx: installed 2 in 1.984s

  System:
    OS: Windows 10 10.0.22000
  Binaries:
    Node: 14.19.1 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
  Browsers:
    Chrome: Not Found
    Edge: Spartan (44.22000.120.0), Chromium (103.0.1264.62)
  npmPackages:
    @emotion/react: 11.9.0 => 11.9.0
    @emotion/styled: 11.8.1 => 11.8.1
    @mui/base:  5.0.0-alpha.75
    @mui/icons-material: 5.5.1 => 5.5.1
    @mui/lab: 5.0.0-alpha.75 => 5.0.0-alpha.75
    @mui/material: 5.6.0 => 5.6.0
    @mui/private-theming:  5.6.0
    @mui/styled-engine:  5.6.0
    @mui/styles: 5.6.0 => 5.6.0
    @mui/system:  5.6.0
    @mui/types:  7.1.3
    @mui/utils:  5.6.1
    @mui/x-data-grid:  5.8.0
    @mui/x-data-grid-pro: 5.8.0 => 5.8.0
    @mui/icons-material: 5.5.1 => 5.5.1
    @mui/lab: 5.0.0-alpha.75 => 5.0.0-alpha.75
    @mui/material: 5.6.0 => 5.6.0
    @mui/private-theming:  5.6.0
    @mui/styled-engine:  5.6.0
    @mui/styles: 5.6.0 => 5.6.0
    @mui/system:  5.6.0
    @mui/types:  7.1.3
    @mui/utils:  5.6.1
    @mui/x-data-grid:  5.8.0
    @mui/x-data-grid-pro: 5.8.0 => 5.8.0
    @mui/x-date-pickers-pro: 5.0.0-alpha.0 => 5.0.0-alpha.0
    @mui/x-license-pro: 5.7.0 => 5.7.0
    @types/react:  17.0.1
    react: 18.0.0 => 18.0.0
    react-dom: 18.0.0 => 18.0.0
```
</details>

Browser:    Chrome Version 103.0.5060.114 (Official Build) (64-bit)

alexfauquette:
Fix #5534
Fix #4597
Fix #4900

Follow up: the picker should set `autoFocus=true` to `<CalendarOrClockPicker />` when opening, due to a user interaction.

For now, if `autoFocus=false`, the focus is lost which is a problem for popper, because you need to press tap on the entire page to reach it.


LukasTy:
One issue I've discovered:
- [Focus the year element](https://deploy-preview-5820--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#static-mode) by tabbing
- select a different year navigating with arrows and confirming with `Enter` or `Space`
- notice how the opened day picker looses focus and you would need to `tab` -> `shift + tab` to get back to the selected day
I would expect the day to be focused. 🤔 

Also, a side note, shouldn't we open month picker after selecting year instead of day picker?

LukasTy:
Another issue/question:

- Focus the selected day [in landscape orientation](https://deploy-preview-5820--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#landscape-orientation) calendar
- hold `LeftArrow`
- notice how the disabled day blocks the focused day from being changed to the previous enabled date (i.e. jump from 15 to 12) like it would if no days were disabled

LukasTy:
Same issue with not being able to navigate elements having disabled ones applies to all the picker types.
We can go with either of the following approaches (listed in increasing difficulty):
- leave as is as it's still better than what we have and aim to improve it via a separate PR
- look into a recursive solution finding the next/previous possible date to navigate to
- have some sort of mechanism/utility allowing to check next/previous available date given a `step` (by how much we want to advance) and `disabledDates`

LukasTy (changes requested):
LGTM other than [this issue](https://github.com/mui/mui-x/pull/5820#issuecomment-1219378038), that I'd say we ought to solve with this PR.
Other than that, I'd say we could release it in such state with the other notes as future improvements. 🤔

alexfauquette:
> Also, a side note, shouldn't we open month picker after selecting year instead of day picker?

About that aspect, dev can customize `views=['year', 'month', 'day']` if they prefer to ask for month between year and day

About the management of disabled button in keyboard management, It's a subject with a lot of edge cases, such as "should we allow month switching if the previous available date is in the previous month". Better to move this topic to another PR

----

Regarding the issue "When I select a year, the focus does not go directly to the available date"

The problem of the previous approch is that each component is handling it's own `hasFocus` state. To be able to move from one component to another, it should be shared between each of them. So in `CalendarOrClockPicker`.

For now, it works with all the calendar interfaces (year, month, day). Did not looked yet to the clock

packages/x-date-pickers/src/StaticDatePicker/StaticDatePickerKeyboard.test.tsx
```diff
@@ -95,12 +95,111 @@ describe('<StaticDatePicker /> keyboard interactions', () => {
...
+        act(() => year.focus());
+        fireEvent.keyDown(year, { keyCode, key });
 
         expect(document.activeElement).to.have.text(expectFocusedYear);
       });
     });
   });
+  describe('MonthPicker keyboard navigation', () => {
+    [
+      { keyCode: 37, key: 'ArrowLeft', expectFocusedMonth: 'Jul' },
```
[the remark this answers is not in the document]
alexfauquette: Yes, I though it was here for strange bug in testing library so I kept them during copy/past, but seems to work with `key` alone 👍

LukasTy:
@alexfauquette Could you double check the changes?
it seems that some recent commit might have broken the functionality.
It's again no longer possible to navigate static sub-components with keyboard.

alexfauquette:
@LukasTy Which behavior? I tried on the preview for StaticDateTimePicker and di not faced problems with tab and arrows

https://user-images.githubusercontent.com/45398769/185643011-0d228fd6-e44a-4d99-a30e-eb8876d3beb1.mp4

LukasTy:
I meant the [sub-components](https://deploy-preview-5820--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#sub-components), they don't seem to function with keyboard anymore.

alexfauquette:
Yes, that's a nightmare. My last idea will be to only store `hasFocus` in `CalendarOrClock` and let `DayPicker`, `MontPicker` and other simply notify it when their internal state switch from `hasFocus=true` to `hasFocus=false`


image

packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/useFocusManagement.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+
+interface FocusState {
+  focusedPicker: 'year' | 'month' | 'day';
+  hasFocus: boolean;
```
LukasTy, line 5: Nitpick: Wouldn't having `focusedPicker: 'year' | 'month' | 'day' | undefined` without `hasFocus` flag be easier to manage and still achieve the same functionality? 🤔

alexfauquette: Yes, it would work too with `hasFocus = focusedPicker !== undefined` but not sure it's would be much easier.

However, we can have `focusedPicker: 'year' | 'month' | 'day' | null` and instead of using `hasFocus`, use `focusedPicker` for <Calendar />, such that 

```jsx
<YearPicker
	hasFocus={focusedPicker!==undefined ? focusedPicker==='year' : undefined}
	...
/>
<MonthPicker
	hasFocus={focusedPicker!==undefined ? focusedPicker==='month' : undefined}
	...
/>
<DayPicker
	hasFocus={focusedPicker!==undefined ? focusedPicker==='day' : undefined}
	...
/>
```

LukasTy: Yup, or `null`, whichever would be easiest to implement, but I meant that from the code it doesn't seem that separate `hasFocus` flag add much in flexibility 🤔 
```tsx
<YearPicker hasFocus={focusedPicker === 'year'} />
<MonthPicker hasFocus={focusedPicker === 'month'} />
<DayPicker hasFocus={focusedPicker === 'day'} />
```
But in any case, it's just food for thought as eventually we might see that to achieve certain state/behaviour we might need a separate flag. 😄

LukasTy:
After the latest changes we still have a regression with sub-components.
The days in the `CalendarPicker` are not navigable.

https://user-images.githubusercontent.com/4941090/186358938-6e45117e-f3de-4655-bfe2-6ab4ea8ddf86.mov

LukasTy:
I've also discovered a disparity in how month disabled checking works.

https://user-images.githubusercontent.com/4941090/186360178-40817c02-033a-4a7f-a082-4f144083fe6d.mov

For `isDisabled` flag we are checking the start of month.
For the ability to focus date we are checking if `today` is not disabled that month.

Hence, the result is that we can select a month with a pointer, but keyboard does not allow it, because `maxDate` is set to `2023-06-01`.

It would seem that for `MonthPicker` `today` should not be taken into account? If the first day of the month is available - the month is available?

packages/x-date-pickers/src/MonthPicker/MonthPicker.tsx
```diff
@@ -140,24 +150,71 @@ export const MonthPicker = React.forwardRef(function MonthPicker<TDate>(
...
+    (month: number) => {
+      if (!isMonthDisabled(utils.setMonth(selectedDateOrToday, month))) {
+        setFocusedMonth(month);
+        setHasFocus(true);
+      }
+    },
+    [selectedDateOrToday, isMonthDisabled, utils],
+  );
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
```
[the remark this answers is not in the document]
flaviendelangle: I think those kind of function are good candidates for the MUI `useEventCallback` utility hook and the future `useEvent` built-in React hook

alexfauquette:
The `hasFocus` is here for a unique use case, keeping the focus when switching between two views.

We can not set automatically the `hasFocus` to `true` for the open view, because of the static pickers. For this case, it would be equivalent to have `autoFocus={true}` so if the dev wants to set the autofocus on another input it will be a problem

alexfauquette:
I've added test to ensure focus is kept when keyboard moves from one calendar to another, And that clicking on a year leads to a date picker which has the focus

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
```diff
@@ -116,6 +117,8 @@ export interface CalendarPickerProps<TDate>
    * @returns {void|Promise} -
    */
   onMonthChange?: (month: TDate) => void | Promise<void>;
+  focusedPicker?: CalendarPickerView | null;
```
flaviendelangle, line 120: Shouldn't it be ?

```suggestion
  focusedView?: CalendarPickerView | null;
```

alexfauquette: Yes, for now it's confusing because we have the `YearPicker`, `monthPicker`, ...  which are in fact views

flaviendelangle: Same for `onFocusedPickerChange`

And yes I agree that `YearPicker` & co are views not pickers

LukasTy (approved):
Looks good to me overall.
Great job 🚀 
The only gripes I would have are the lacking keyboard navigation support for clock picker and no keyboard support for range picker 😭 
But there are totally different beasts. 🐻 😆

flaviendelangle:
> no keyboard support for range picker 

That one is very related to discussions we had recently
Should the pickers behave like tooltip or like modals
Right now the range pickers are tooltip and the other modals
I would be in favor of migrating the range pickers to a model approach for consistency (like telerik and spectrum) and see later if we want to add an option for a coherent tooltip approach for all (like Retool pickers)

That would be one of the possible major evolutions of the new pickers that are built on top of the new fields

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
```diff
@@ -498,7 +498,7 @@ export const CalendarPicker = React.forwardRef(function CalendarPicker<TDate>(
               date={date}
               onChange={handleDateMonthChange}
               shouldDisableMonth={shouldDisableMonth}
-              onFocusedPickerChange={handleFocusedPickerChange('month')}
+              onFocusedViewChange={handleFocusedPickerChange('month')}
```
flaviendelangle, line 501:
```suggestion
              onFocusedViewChange={handleFocusedViewChange('month')}
```

:angel:

alexfauquette: I will also do 
```diff
-internalFocusedPicker
+internalFocusedView
```
👼

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
```diff
@@ -342,8 +290,9 @@ const PickersDayRaw = React.forwardRef(function PickersDay<TDate>(
       data-mui-test="day"
       disabled={disabled}
       tabIndex={selected ? 0 : -1}
-      onFocus={handleFocus}
-      onKeyDown={handleKeyDown}
+      onKeyDown={(event) => onKeyDown(event, day)}
```
flaviendelangle, line 293: You can avoid the `noop` with `onKeyDown?.(event, day)`

But as you want

packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/useFocusManagement.tsx
```diff
@@ -0,0 +1,60 @@
...
+type FocusModificationPayload = { view: CalendarPickerView };
+
+type ReducerAction<TType, TAdditional = {}> = { type: TType } & TAdditional;
+
+export const createCalendarStateReducer =
+  () =>
+  (
+    state: FocusState,
+    action:
+      | ReducerAction<'focus', FocusModificationPayload>
```
flaviendelangle, line 17: I'm really not a fan of the `ReducerAction` interface that I saw elsewhere on this project
I find it hard to render.

I would prefer to use basic types

```ts
type FocusManagementAction = 
  | { type: 'focus', view: CalendarPickerView } 
  | { type: 'blur', view: CalendarPickerView }
```

Or if you are confident that all actions need the view

```ts
type FocusManagementAction = { action: 'focus' | 'blur', view: CalendarPickerView }
```

flaviendelangle: By the way, can't we just use a `useState` here ?
The reducer feels somehow overengineered

flaviendelangle (commented):
Sorry for the delay in the review
The new behavior is a lot better, I just have a few suggestions

packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/useFocusManagement.tsx
alexfauquette: Yes, it have been simplified with time an so does not need reducer anymore

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
alexfauquette: Already did a lot of modification, I prefer to stick with what was there before

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -155,6 +158,19 @@ export function DayPicker<TDate>(props: DayPickerProps<TDate>) {
     disableFuture,
   });
 
+  const [internalFocusedDay, setInternalFocusedDay] = React.useState<TDate>(
```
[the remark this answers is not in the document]
flaviendelangle: Maybe it's worth doing a `useMemo` for the risky check, to avoid calling it too often

flaviendelangle (approved):

PR #5833, opened 2022-08-18
[pickers] Add Polish (pl-PL) locale

drastus:
I’d like to provide Polish (pl-PL) translations for the date picker component (#3211).

### DIFF
--- a/packages/x-date-pickers/src/locales/plPL.ts
+++ b/packages/x-date-pickers/src/locales/plPL.ts
@@ -2,7 +2,7 @@ import { PickersLocaleText } from './utils/pickersLocaleTextApi';
 import { getPickersLocalization } from './utils/getPickersLocalization';
 import { CalendarPickerView } from '../internals/models';
 
-const plPLPickers: PickersLocaleText<any> = {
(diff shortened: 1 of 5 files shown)


packages/x-date-pickers/src/locales/plPL.ts
```diff
@@ -0,0 +1,52 @@
+import { PickersLocaleText } from './utils/pickersLocaleTextApi';
+import { getPickersLocalization } from './utils/getPickersLocalization';
+import { CalendarPickerView } from '../internals/models';
+
+const plPLPickers: PickersLocaleText<any> = {
```
alexfauquette, line 5: It should be partial because when we modify the default translation, `PickersLocaleText<any>` type will change but we will not be able to maintain other locales by ourself

```suggestion
const plPLPickers: Partial<PickersLocaleText<any>> = {
```

alexfauquette (approved):
The code sounds good. I think andrew have some notion of polish to review the translation themself

Thanks for your contribution 🥳️

cherniavskii (approved):
Thanks @drastus 🎉

Issue #4754, opened 2022-05-04
[data grid] Row reorder visual updates


DanailH:
### Summary 💡

As a follow-up of the row reordering PR https://github.com/mui/mui-x/pull/4034 we can do an integration where the focus will be the visual (UX) part of the feature. Below is the summary of what we can improve.

- [ ] 1. Update the value of what is displayed in the draggable container once the dragging start. Currently, the ID of the row that is being dragged is displayed there. To be changed to show the content of the first visible cell in that row. #11404
- [x] 2. Currently while dragging the rows swap places immediately. An improvement would be to show to the user the place where the row will be moved (check the image below for more details), or to keep the current implementation, and highlight the row being moved. Some implementations use a transition as the moved row swaps places with an adjacent row.

<img width="433" alt="166166263-cebccdab-84ff-4a8b-a401-e286c4ffbae2" src="https://user-images.githubusercontent.com/5858539/166642828-8c46ff8e-305f-4172-9916-e3554ab7ce5a.png">

- [x] 3. Update the icon that illustrates if a row is draggable. Make it smaller, maybe 18px, and use the `theme.palette.action.active` design token like the IconButton: https://mui.com/material-ui/react-button/#icon-button.

<img width="777" alt="166074158-d99c7174-e15f-4141-8446-1a749de9bf7e" src="https://user-images.githubusercontent.com/5858539/166645634-c19ef7ca-7c01-4ff4-9ed1-fce07318f235.png">

For more details refer to these comments: https://github.com/mui/mui-x/pull/4034#discussion_r863184696, https://github.com/mui/mui-x/pull/4034#issuecomment-1113789228

artsiadi:
@DanailH Thank you so much for adding this feature!
I was wondering, is there a way to disable the drag just for certain rows in the data grid?

DanailH:
Hey @artsiadi currently no, but I'm interested in knowing what your use case is?

artsiadi:
Hi @DanailH, we use the grid to display rows with priorities. the user can re-order the rows and change their priority. in addition to that we have a default row that always appears at the bottom of the list and has the lowest priority. the default row cannot be edited or change priority.

DanailH:
@artsiadi we can add the option to disable a row to be reordered but if you can reorder the rows around it that will still change the location of your fixed row, no?

artsiadi:
@DanailH I added a check in the onRowOrderChange callback to make sure that the new index of the row is not the index of the pinned one.
[...]

PR #5883, opened 2022-08-23
[DataGridPro] Add `isRowDraggable` prop

DanailH:
Part of https://github.com/mui/mui-x/issues/4754
Based on this https://github.com/mui/mui-x/issues/4754#issuecomment-1190101937

`isRowDraggable` allows for the user to define which rows are draggable and which aren't.


packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
```diff
@@ -54,15 +59,18 @@ export const useGridRowReorder = (
...
+    },
+    [rowReordering, isRowDraggable, sortModel, treeDepth],
+  );
 
   const handleDragStart = React.useCallback<GridEventListener<'rowDragStart'>>(
     (params, event) => {
       // Call the gridEditRowsStateSelector directly to avoid infnite loop
       const editRowsState = gridEditRowsStateSelector(apiRef.current.state);
-      if (isRowReorderDisabled || Object.keys(editRowsState).length !== 0) {
+      if (isRowReorderDisabled(params) || Object.keys(editRowsState).length !== 0) {
```
m4theushw, line 73: Do we need this check if `rowDragStart` is already only fired if the row is draggable?

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
```diff
@@ -41,11 +42,12 @@ const GridRowReorderCell = (params: GridRenderCellParams) => {
...
   const isDraggable = React.useMemo(
     () =>
-      !!rootProps.rowReordering &&
+      !!rowReordering &&
       !sortModel.length &&
       treeDepth === 1 &&
-      Object.keys(editRowsState).length === 0,
-    [rootProps.rowReordering, sortModel, treeDepth, editRowsState],
+      Object.keys(editRowsState).length === 0 &&
+      isRowDraggable(apiRef.current.getRowParams(params.id)),
```
m4theushw, line 49: Based on the comments in #4754 I don't know what users expect. They may want to keep some rows always at the bottom, like pinned rows but not pinned, or just disable the icon. If we disable the icon, then `isRowDraggable` could be only a prop of `GridRowReorderCell`, not for the grid. I think we should also add a demo of this feature, to clarify https://github.com/mui/mui-x/issues/4754#issuecomment-1187580566

packages/grid/x-data-grid-pro/src/hooks/features/rowReorder/useGridRowReorder.tsx
DanailH: Probably not, but you once might try and reorder rows programatically?

packages/grid/x-data-grid-pro/src/components/GridRowReorderCell.tsx
DanailH: Noted, I'll add a demo. We can also go with the approach of adding a prop on the `GridRowReorderCell`. I went with the current approach as it's the closest in relation to what we have for the columns.
Regarding the first point of having pinned rows - same as the columns. You can reorder around them but can't reorder the one that's disabled.

thuatole:
Hi @DanailH , why did we close this PR ? Is there any another approach to define which rows are draggable and which aren't ? Thanks

DanailH:
@thuatole the main issue with this approach is that even if you mark a row as a nondraggable you can still change its position by reordering rows around it. That was @m4theushw's main concern. It would be interesting to know your use case of use this feature and why the row pinning feature doesn't work in your case.

PR #5888, opened 2022-08-24
[DataGrid] Add `sort` prop to columns panel slot

### LINKED ISSUE #5830: [data grid] Add prop to customize sort order in the columns panel
### Duplicates

### Latest version

### Summary 💡

I'm using Mui DataGrid-Pro/X and have a request to make the column chooser menu appear in alphabetical order. So far, the only way I can see to do this is to change the order of the columns in the grid or to write a whole new column panel element. Is there a more practical / simple way to do this? Or is there a way to access the code for the default component to perform some light customization without reinventing the wheel?

### Examples 🌈

_No response_

### Motivation 🔦

User requests that the column picker show columns in alphabetical order

### Order ID 💳 (optional)

_No response_

gavbrennan:
Fix #5830

Adds ability to specify a sort order for display of columns in column picker, without changing the order of the columns themselves.

### DIFF
--- a/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
@@ -66,6 +66,8 @@ export interface GridColumnsPanelProps extends GridPanelWrapperProps {
   sort?: 'asc' | 'desc';
 }
 
+const collator = new Intl.Collator();
+
 export function GridColumnsPanel(props: GridColumnsPanelProps) {
   const apiRef = useGridApiContext();
   const searchInputRef = React.useRef<HTMLInputElement>(null);
@@ -81,11 +83,11 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
     switch (sort) {
       case 'asc':
         return [...columns].sort((a, b) =>
-          (a.headerName || a.field) > (b.headerName || b.field) ? 1 : -1,
+          collator.compare(a.headerName || a.field, b.headerName || b.field),
         );
       case 'desc':
-        return [...columns].sort((a, b) =>
-          (a.headerName || a.field) < (b.headerName || b.field) ? 1 : -1,
+        return [...columns].sort(
+          (a, b) => -collator.compare(a.headerName || a.field, b.headerName || b.field),
         );
       default:
         return columns;
--- a/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
@@ -80,6 +80,7 @@ function GridColumnsPanel(props: GridColumnsPanelProps) {
   const classes = useUtilityClasses(ownerState);
 
   const { sort } = props;
+
   const sortedColumns = React.useMemo(() => {
     switch (sort) {
       case 'asc':
--- a/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
@@ -79,9 +79,8 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
       return [...columns].sort((a, b) =>
         (a.headerName || a.field) < (b.headerName || b.field) ? 1 : -1,
       );
-    } else {
-      return columns;
-    }
+    } 
+    return columns;
   }, [columns, rootProps.componentsProps?.columnsPanel.sort]);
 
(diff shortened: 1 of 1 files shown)


packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -70,6 +70,20 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
   const ownerState = { classes: rootProps.classes };
   const classes = useUtilityClasses(ownerState);
 
+  const sortedColumns = React.useMemo(() => {
+    if (rootProps.componentsProps?.columnsPanel.sort === 'asc') {
```
alexfauquette, line 74: `rootProps.componentsProps?.columnsPanel` is already passed to the column panel. 

https://github.com/mui/mui-x/blob/33cf86590cd69cdcd94dbaa831c1aeb2e70785c0/packages/grid/x-data-grid/src/hooks/features/columns/useGridColumns.tsx#L390:L391
```jsx
const ColumnsPanel = props.components.ColumnsPanel;
return <ColumnsPanel {...props.componentsProps?.columnsPanel} />;
```

So you can simply get `sort` from the `props`

with 
```jsx
const { sort, ...other } = props
```

and replace the `...props` by `...other` in the `GridPanelWrapper`

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -70,6 +70,20 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
...
+      return [...columns].sort((a, b) =>
+        (a.headerName || a.field) > (b.headerName || b.field) ? 1 : -1,
```
alexfauquette, line 76: To let sorting adapt to different locales, you can use `Intl.Collator();`
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator

This behavior has already be done in the filter panel [here](https://github.com/mui/mui-x/blob/master/packages/grid/x-data-grid/src/components/panel/filterPanel/GridFilterForm.tsx#L228:L243)

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
alexfauquette: You can also add the `sort` property to the type definition to `GridColumnsPanelProps`. For the filter panel it looks like that 

```ts
/**
 * Changes how the options in the columns selector should be ordered.
 * If not specified, the order is derived from the `columns` prop.
 */
columnsSort?: 'asc' | 'desc';
```

This should allow TS autocompletion, and warning is `asc` or `desc` is miss spell

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -70,6 +70,19 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
...
+      );
+    } else if (rootProps.componentsProps?.columnsPanel.sort === 'desc') {
```
alexfauquette, line 78: You can remove the `else` because the previous `if` ended with a `return`. Maybe `switch (sort)` will be easier to read

```suggestion
    }
    if (rootProps.componentsProps?.columnsPanel.sort === 'desc') {
```

gavbrennan:
@m4theushw if you are interested

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
m4theushw: +1 for Alexandre's suggestion. You also need to replace `props` with `other` in

https://github.com/mui/mui-x/blob/412d8db2feb1a9e14ecc9cbe4270cf25749b4678/packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx#L129 to avoid passing `sort` to the DOM element.

image

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -70,6 +79,24 @@ export function GridColumnsPanel(props: GridColumnsPanelProps) {
   const ownerState = { classes: rootProps.classes };
   const classes = useUtilityClasses(ownerState);
 
+  const { sort } = props;
+  const sortedColumns = React.useMemo(() => {
```
m4theushw, line 83:
```suggestion

  const sortedColumns = React.useMemo(() => {
```

alexfauquette (approved):
Sounds good to me 🚀

m4theushw (approved):
Thanks!

Issue #5843, opened 2022-08-19
[pickers][DatePicker/DateTimePicker] How to pass props to custom OpenPickerIcon


nazarukroman:
Hi everyone!
I want to pass custom props to OpenPickerIcon.

```
import {  DatePicker,  LocalizationProvider } from '@mui/x-date-pickers';
import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';

    <LocalizationProvider dateAdapter={AdapterDateFns}>
      <DateTimePicker
        onChange={handleChange}
        value={date}
        renderInput={({ inputProps: muiInputProps, ...params }) => (
          <Input
            iconLeft={iconLeft}
            size={size}
            {...controlProps}
            {...params}
          />
        )}
        label={label}
        components={{
          OpenPickerIcon: CalendarIcon,
        }}
        inputFormat={inputFormat}
        mask={mask}
        {...rest}
      />
    </LocalizationProvider>
```

`componentsProps={{OpenPickerIcon: {some: 'prop' }}}` doesn't work.

How can I pass custom props to my OpenPickerIcon?

LukasTy:
Hello @nazarukroman 
Are you using TypeScript?
There are no types for the component you provided.
What kind of use case do you have?
I believe the following example could help you:
```tsx
import * as React from 'react';
import TextField from '@mui/material/TextField';
import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
import { DatePicker } from '@mui/x-date-pickers/DatePicker';
import CalendarIcon from '@mui/icons-material/CalendarMonth';

const OpenPickerIcon = () => <CalendarIcon fontSize="large" />;

export default function BasicDatePicker() {
  const [value, setValue] = React.useState<Date | null>(null);

  return (
    <LocalizationProvider dateAdapter={AdapterDateFns}>
      <DatePicker
        label="Basic example"
        value={value}
        onChange={(newValue) => {
          setValue(newValue);
        }}
        renderInput={(params) => <TextField {...params} />}
        components={{
          OpenPickerIcon,
        }}
      />
    </LocalizationProvider>
  );
}

```

nazarukroman:
@LukasTy sorry, forgot about this thread..

Yep I have TS but I thought it doesn't matter but ok
Look at this example:

```typescript
[...]

PR #5890, opened 2022-08-24
[pickers] Clean slots on `PickersArrowSwitcher` component

flaviendelangle:
To be merged in the v6 alpha

The goal is to totally clean the slots of the picker components to follow what `@mui/base` is doing

- support callback format of `componentsProps` for instance
- always have a `componentsProps` matching the `components`
- allow to use module augmentation to define new props for the slots
- switch button order when in RTL mode

Part of https://github.com/mui/mui-x/issues/4466

[Doc preview](https://deploy-preview-5890--material-ui-x.netlify.app/x/react-date-pickers/custom-components/)
[RTL example](https://codesandbox.io/s/date-and-time-pickers-forked-wzyrmv)


packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const RightArrowIcon = components?.RightArrowIcon || ArrowRight;
+
+  return (
+    <PickersArrowSwitcherRoot
+      ref={ref}
+      className={clsx(classes.root, className)}
+      ownerState={ownerState}
+      {...other}
+    >
+      <LeftArrowButton {...leftArrowButtonProps}>
```
flaviendelangle, line 111: Shouldn't we also switch the buttons when `isRtl === true` ?

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
```diff
@@ -63,20 +67,9 @@ export interface ExportedClockPickerProps<TDate> extends ExportedTimeValidationP
...
-export interface ClockPickerSlotsComponent {
-  LeftArrowButton: React.ElementType;
-  LeftArrowIcon: React.ElementType;
-  RightArrowButton: React.ElementType;
-  RightArrowIcon: React.ElementType;
-}
-
-// We keep the interface to allow module augmentation
-export interface ClockPickerComponentsPropsOverrides {}
+export interface ClockPickerSlotsComponent extends PickersArrowSwitcherSlotsComponent {}
```
flaviendelangle, line 70: The slots were duplicated, `ClockPicker` should not have been responsible of this abstraction

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
```diff
@@ -0,0 +1,144 @@
...
+      disabled: isLeftDisabled,
+      edge: 'end',
+      onClick: onLeftClick,
+    },
+    ownerState: { ...ownerState, hidden: isLeftHidden },
+    className: classes.button,
+  });
+
+  const LeftArrowIcon = components?.LeftArrowIcon ?? ArrowLeft;
+  const { ownerState: leftArrowIconOwnerState, ...leftArrowIconProps } = useSlotProps({
```
flaviendelangle, line 85: The spread is here to avoid this bug https://github.com/mui/material-ui/issues/34056

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -90,3 +90,15 @@ You can provide any custom component to this slot as long as it passes `children
 In the examples below we provide ways of implementing date range shortcuts using this slot.
 
 {{"demo": "PaperContentComponent.js", "defaultCodeOpen": false, "disableAd": true}}
+
+## Arrow switcher buttons and icons
+
+:::info
+Slots: `LeftArrowIcon` / `RightArrowIcon` / `LeftArrowButton` / `RightArrowButton`
```
flaviendelangle, line 97: @samuelsycamore @joserodolfofreitas I would be interested in your feedback here
I am trying to list the slots that the section is talking about

I would like to apply the same UI on every section of this slots documentation.

LukasTy:
What kind of breaking change does this introduce? 🤔 
Should we maybe also include a fix to pass props to all the possible pickers components solving #5843 at the same time?

flaviendelangle:
> What kind of breaking change does this introduce? thinking

I was wondering if we should change the slots here so I added the label to make sure we don't merge it if we do introduce one.
But yeah for now I don't think there is any breaking change.

> Should we maybe also include a fix to pass props to all the possible pickers components solving https://github.com/mui/mui-x/issues/5843 at the same time?

I did add the components props to the slots handled by `PickersArrowSwitcher`.

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
flaviendelangle: Discussed during yestrerday's meeting, I switched the button

docs/data/date-pickers/custom-components/custom-components.md
samuelsycamore: Interesting. We should definitely give this some thought. My first thought would be to create a list, but after seeing this text on the page I appreciate how concise it is. Maybe we should treat it like a plain sentence and use commas instead?

```suggestion
Slots: `LeftArrowIcon`, `RightArrowIcon`, `LeftArrowButton`, `RightArrowButton`.
```

flaviendelangle: We will have a similar topics for sections concerning several picker components but not all

For instance:

```
Components supporting this behavior: DatePicker, DateTimePicker, DateRangePicker, DateTimeRangePicker
```

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
```diff
@@ -0,0 +1,144 @@
...
+    elementType: LeftArrowIcon,
+    externalSlotProps: componentsProps.leftArrowIcon,
+    ownerState: undefined,
+  });
+
+  const RightArrowButton = components.RightArrowButton ?? PickersArrowSwitcherButton;
+  const rightArrowButtonProps = useSlotProps({
+    elementType: LeftArrowButton,
+    externalSlotProps: componentsProps.leftArrowButton,
```
alexfauquette, line 94:
```suggestion
    elementType: RightArrowButton,
    externalSlotProps: componentsProps.rightArrowButton,
```

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onRightClick: () => void;
+}
+
+export type PickersArrowSwitcherOwnerState = PickersArrowSwitcherProps;
+
+export interface PickersArrowSwitcherComponentsPropsOverrides {}
+
+export interface PickersArrowSwitcherSlotsComponent {
+  /**
+   * Button allowing to switch to the left view.
```
alexfauquette, line 47:
```suggestion
   * Button allowing to switch to the previews view.
```

When left-to-right, is activated, is left the next view? using previous/next remove the confusion

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
alexfauquette: I was trying the following to avoid overflow in the clock view

```jsx
leftArrowButton: {
    color: "primary"
  },
  rightArrowButton: {
    color: "error"
  },
  ```
  
  
image

Comment (flaviendelangle) [reply packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx:94]: Oups, fixed :p

File: packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx (hunk 6 above)
Comment (flaviendelangle) [reply packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx:47]: In this case we should probably rename the slots no ?

Comment (alexfauquette) [reply packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx:47]: If we still accept breaking changes, yes

Comment (flaviendelangle) [reply packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx:47]: I think we will start the alpha quite soon, this PR can wait until then

File: docs/data/date-pickers/custom-components/custom-components.md (hunk 4 above)
Comment (samuelsycamore) [reply docs/data/date-pickers/custom-components/custom-components.md]: Maybe a list would be better after all? It would be easier to quickly scan.

Comment (flaviendelangle) [reply docs/data/date-pickers/custom-components/custom-components.md]: I wonder if we could create some custom style (a bit like the `:::info`) that lets you know that this small block contains information about the scope of the section

**Comment by flaviendelangle** [comment]
@LukasTy @alexfauquette 

Another point
I used the `@mui/base` file naming pattern (`MyComponent.types.tsx` instead of `MyComponent.interfaces.tsx`).
Which one do you prefer ?

**Comment by LukasTy** [comment]
@flaviendelangle 
> I used the @mui/base file naming pattern (MyComponent.types.tsx instead of MyComponent.interfaces.tsx).
Which one do you prefer ?

I don't have a strong preference over one or the other, as long as we maintain consistency.
In JS/TS world `types` seems like a reasonable choice.
One nitpick, that you could probably go with `MyComponent.types.ts` instead of `tsx` as there is usually no JSX code in types. 😃

**Comment by alexfauquette** [comment]
I dont' have preferences on the file naming

**Comment by flaviendelangle** [comment]
> I don't have a strong preference over one or the other, as long as we maintain consistency.

Fully agree that we should be consistent

File: packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx (hunk 6 above)
Comment (flaviendelangle) [reply packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx:37]: I tried to RTL mode and it's totally broken even on master
As far as I understand, we should not switch any icon / button but just the props associated to them.

I pushed a commit changing the behavior (and removing the deprecated locale text, there name were not correct and I don't think it was worth fixing them since we are merging this PR during the alpha phase)

**Review by samuelsycamore — COMMENTED** [review]
Text looks good! Just a couple small grammar and punctuation things.

File: docs/data/date-pickers/custom-components/custom-components.md [hunk 7]
```diff
@@ -90,3 +90,15 @@ You can provide any custom component to this slot as long as it passes `children
...
 
 {{"demo": "PaperContentComponent.js", "defaultCodeOpen": false, "disableAd": true}}
+
+## Arrow switcher buttons and icons
+
+:::info
+Slots: `LeftArrowIcon`, `RightArrowIcon`, `LeftArrowButton`, `RightArrowButton`.
+:::
+
+The following slots allows you to customize how to render the buttons and icons for an arrow switcher component (the component to navigate to the "Previous" and "Next" step of the picker)
```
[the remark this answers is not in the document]
LukasTy: @samuelsycamore Do we have conventions on [which dash/hyphen](https://jakubmarian.com/hyphen-minus-en-dash-and-em-dash-difference-and-usage-in-english/) to use when?
As far as I can see, there are only a few instances of other than the minus dash in codebase... 🤔

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -182,16 +187,16 @@ export function DateRangePickerViewDesktop<TDate>(props: DesktopDateRangeCalenda
...
           <DateRangePickerViewDesktopContainer key={index}>
             <DateRangePickerViewDesktopArrowSwitcher
-              onLeftClick={selectPreviousMonth}
-              onRightClick={selectNextMonth}
-              isLeftHidden={index !== 0}
-              isRightHidden={index !== calendars - 1}
-              isLeftDisabled={isPreviousMonthDisabled}
-              isRightDisabled={isNextMonthDisabled}
-              leftArrowButtonText={leftArrowButtonText}
+              goToPrevious={selectPreviousMonth}
```
LukasTy, line 190: Nitpick: Don't we usually follow convention to prefix `actions` with `on` to make it more clear, what kind of prop is this?
I.e.: `onPreviousClick`, `onNextClick`

flaviendelangle: I liked the idea of focusing on the behavior (go to the previous / next item) rather than focusing on the user action (click on a button).
Even if, for this component, both are probably 100% equivalent.

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
LukasTy: What do you mean by this comment?
I don't see that anything would have been changed here. 🤔

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
LukasTy: I do understand and am pro for this change.
I was simply asking about the prop naming convention in regards to `on` prefix for actions, like `onChange`, `onClick`, `onGoToNext` or like I said `onNextClick`

packages/x-date-pickers/src/ClockPicker/ClockPicker.tsx
flaviendelangle: Oh @alexfauquette already did the fix on #5893 so my comment does not make sense anymore
Basically we add the slots of the `PickersArrowSwitcher` redefined here instead of extending the interface.

packages/x-date-pickers/src/internals/index.ts
```diff
@@ -11,8 +11,12 @@ export type {
...
 } from './components/wrappers/MobileWrapper';
 export { MobileKeyboardInputView } from './components/CalendarOrClockPicker/CalendarOrClockPicker';
-export { PickersArrowSwitcher } from './components/PickersArrowSwitcher';
-export type { ExportedArrowSwitcherProps } from './components/PickersArrowSwitcher';
+export { PickersArrowSwitcher } from './components/PickersArrowSwitcher/PickersArrowSwitcher';
+export type {
+  ExportedPickersArrowSwitcherProps,
+  PickersArrowSwitcherSlotsComponent,
+  PickersArrowSwitcherSlotsComponentsProps,
+} from './components/PickersArrowSwitcher';
```
LukasTy, line 19: Nitpick: Wouldn't just doing this achieve the same here as it already exports everything that is mentioned here in the `index.ts`.
Or do we want to be so explicit/in–control? 🤔 
```suggestion
export * from './components/PickersArrowSwitcher';;
```

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
flaviendelangle: For the convention, maybe `onGoToNext`.

I just want to replace the `click` part by a `go to` part.

packages/x-date-pickers/src/internals/index.ts
flaviendelangle: In `internal` I am always very explicit
But maybe that's not super relevant

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
flaviendelangle: Done :+1:

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx
LukasTy: I'll second the opinion. If we are going with renames in the code as well as `localeText`, could we also rename the slots themselves? 🤔 
`PreviousArrowButton`, `NextArrowIcon` and so on.

LukasTy: Also, couldn't we change the extension of this file to `ts`? 🤔

flaviendelangle: From my testing of the RTL, I feel like the slots are named correctly

When I switch to RTL, I still want the arrow pointing left on the left and the arrow pointing right on the right.
But I want the arrow pointing left to go to the next view instead of the previous.

If we change the slots for "previous" and "next", then on RTL people should switch the icons to have something coherent displayed. Otherwise, they would have arrows pointing to one another like "> <" instead of "< >".

Unless my RTL test on Chrome was not representative of a real RTL setup

LukasTy: Screenshot 2022-09-02 at 12 00 21
Toggling our docs theme setting does seem to act exactly like you said it shouldn't. 😢

flaviendelangle: If you try on this PR, does it work correctly ?

I changed both the them setting and the RTL flag on Chrome to test, and it seems to work now

LukasTy: I've tested on the PR docs version deployed on netlify. 🤔

flaviendelangle: https://codesandbox.io/s/date-and-time-pickers-forked-wzyrmv?file=/package.json

Looks like your right
I followed [our guide](https://mui.com/material-ui/guides/right-to-left/#opting-out-of-rtl-transformation) to test on a real setup.
I must have failed my setup yesterday.

In that case, I'm also in favor of renaming the slots.

flaviendelangle: As you saw I am not very confident on this RTL :laughing: 

But I'm not sure for the name of the slot
Let's take a concrete example: 

On LTR, on my screen I have:
- the 1st icon going to the previous view and looking like "<"
- the 2nd icon going to the next view and looking like ">"

If I switch to RTL, stylis changes the order of appearance on screen.
On the DOM I still have the same rendering order, but on my screen, it goes from right to left.

If we rename the `LeftArrowIcon` slot into `PreviousArrowIcon`.
Then the icon will be on the left on LTR and on the right on RTL => perfect.
But it will still be the "<" icon, to it will look like "> <"

So we would have to switch the icon default value depending on the direction.
AND users overriding it would have to do the same, which is not nice I think.


_____


I would rather keep the behavior existing before this PR.
Which is to switch in the code the order of rendering.

- on LTR, I render `{left} {right}` and it displays `{left} {right}` (or `< >` visually) 
- on RTL, I render `{right} {left}, but stylis invert on screen and displays `{left} {right}

Which is what we want, the left is still on the left with the RTL and if users want to switch the icon, they don't have to care about RTL to have something coherent visually in all circumstances.

But we switch the props applied when on RTL: we put the `next` props on the left and the `previous` props on the right.
That way, the icon on the left is going to next and the icon on the right to previous.

flaviendelangle: Compared to my PR, it would be this diff:

```diff
--- a/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
+++ b/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
@@ -47,6 +47,7 @@ export const PickersArrowSwitcher = React.forwardRef(function PickersArrowSwitch
   ref: React.Ref<HTMLDivElement>,
 ) {
   const theme = useTheme();
+  const isRTL = theme.direction === 'rtl';
 
   const {
     children,
@@ -80,8 +81,7 @@ export const PickersArrowSwitcher = React.forwardRef(function PickersArrowSwitch
     label: previousLabel,
   };
 
-  const [leftProps, rightProps] =
-    theme.direction === 'rtl' ? [nextProps, previousProps] : [previousProps, nextProps];
+  const [leftProps, rightProps] = isRTL ? [nextProps, previousProps] : [previousProps, nextProps];
 
   const LeftArrowButton = components.LeftArrowButton ?? PickersArrowSwitcherButton;
   const leftArrowButtonProps = useSlotProps({
@@ -129,6 +129,18 @@ export const PickersArrowSwitcher = React.forwardRef(function PickersArrowSwitch
     ownerState: undefined,
   });
 
+  const leftPart = (
+    <LeftArrowButton {...leftArrowButtonProps}>
+      <LeftArrowIcon {...leftArrowIconProps} />
+    </LeftArrowButton>
+  );
+
+  const rightPart = (
+    <RightArrowButton {...rightArrowButtonProps}>
+      <RightArrowIcon {...rightArrowIconProps} />
+    </RightArrowButton>
+  );
+
   return (
     <PickersArrowSwitcherRoot
       ref={ref}
@@ -136,9 +148,7 @@ export const PickersArrowSwitcher = React.forwardRef(function PickersArrowSwitch
       ownerState={ownerState}
       {...other}
     >
-      <LeftArrowButton {...leftArrowButtonProps}>
-        <LeftArrowIcon {...leftArrowIconProps} />
-      </LeftArrowButton>
+      {isRTL ? rightPart : leftPart}
       {children ? (
         <Typography variant="subtitle1" component="span">
           {children}
@@ -146,9 +156,7 @@ export const PickersArrowSwitcher = React.forwardRef(function PickersArrowSwitch
       ) : (
         <PickersArrowSwitcherSpacer className={classes.spacer} ownerState={ownerState} />
       )}
-      <RightArrowButton {...rightArrowButtonProps}>
-        <RightArrowIcon {...rightArrowIconProps} />
-      </RightArrowButton>
+      {isRTL ? leftPart : rightPart}
     </PickersArrowSwitcherRoot>
   );
 });

```

flaviendelangle: TLDR: On LTR the icon going to the next view looks like `>`, on RTL it looks like `<`.
So if we use previous/ next in the slot, the icon need to be direction-dependent which is not great for customizability.

flaviendelangle: Another approach is to add a `noflip` like explained [here](https://mui.com/material-ui/guides/right-to-left/#emotion-amp-styled-components)

```diff
--- a/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
+++ b/packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.tsx
@@ -16,9 +16,10 @@ const PickersArrowSwitcherRoot = styled('div', {
   overridesResolver: (props, styles) => styles.root,
 })<{
   ownerState: PickersArrowSwitcherProps;
-}>({
-  display: 'flex',
-});
+}>`
+  /* @noflip */
+  display: flex;
+`;
 
 const PickersArrowSwitcherSpacer = styled('div', {
   name: 'MuiPickersArrowSwitcher',
```

The core has only one `no-flip` in its codebase, maybe that's not something we want to rely on

LukasTy: I've tested your [diff](https://github.com/mui/mui-x/pull/5890#discussion_r961499437) and to me it seems to work ideally 👍 
I've compared behaviour with https://demo.mobiscroll.com/javascript/calendar/rtl-right-to-left as well as Firefox native picker with Arabic language selected.
But after even more exploration I'm leaning 100% towards `Previous` and `Next` slot names instead of `Left` and `Right` to reduce confusion on what the user will actually be changing. 🤔

flaviendelangle: > I'm leaning 100% towards Previous and Next slot names instead of Left and Right to reduce confusion on what the user will actually be changing. 

I'm confused
How do you solve the problems I'm describing above ?

> On LTR the icon going to the next view looks like >, on RTL it looks like <.
So if we use previous/ next in the slot, the icon need to be direction-dependent which is not great for customizability.

LukasTy: @flaviendelangle I've done more looking into the `direction` conundrum and I'd be for not introducing too much "magic".
Basically, handle rendering different `arrows` based on `theme.direction` and if users will want to change icons and will need `rtl` support - they can also use `theme` hook to render whatever they need based on selected direction.
And let `css direction` handle rest - layout.
What do you [think about it](https://gist.github.com/LukasTy/3ad6334c4e69a1251a51b1423cf61d02)?
P.S. The diff also handles renaming `left/right` to `previous/next` to maintain prop "sanity" 😃

flaviendelangle: I don't understand the point of going from left / right to prev / next if it make the override more verbous.

What is the magic introduced if we just call it left / right and assure it is rendered on the left and on the right with both LTR and RTL ?

LukasTy: > assure it is rendered on the left and on the right with both LTR and RTL ?

How do you propose to switch the icons based on `direction`?
Rendering/layouting the buttons in the correct position is in charge of `direction: rtl`, we don't need to do anything regarding that, the issue is that arrows should be swapped based on `direction`.

flaviendelangle: > How do you propose to switch the icons based on direction?

Why should we switch the icons ?
I feel like we are not talking about the same thing :laughing: 

My proposal is to keep everything like today:

- The slots are named `left` and `right`
- The left slot is always visible on the left side and the right slot is always visible on the right side (RTL or not)

The only modification I propose is maybe to use `@noflip` instead of a conditional rendering like we have on `master` to avoid playing with the rendering order.

And with this behavior, a developer wanting to change the icon, just change the icon, he does not have to be careful to do a ternary depending on the direction.

LukasTy: > Why should we switch the icons ?

To maintain behaviour on which icon actually get's the provided props, because with the current approach and your suggested fix of flipping props it could get a bit confusing, where `rightArrowIcon` props would be applied either to the right (next) or left (previous) icon, whereas changing only the icon rendering would maintain behaviour of applying props to the same fundamental icon.
In the screenshots below I applied `title: 'Go to Previous'` to `leftArrowButton` and in RTL mode it would incorrectly apply this prop to icon button going to next month, instead of `previous`.
Screenshot 2022-09-05 at 11 25 58
Screenshot 2022-09-05 at 11 27 08

> My proposal is to keep everything like today:
The slots are named `left` and `right`

I don't have the strongest opinion on this point, slots can maintain the same names if we wish to reduce the amount of breaking changes, I'm simply in favour of them being renamed to align with all the other props and changes moving towards `previous/next` instead of `left/right` 🤔 

> The only modification I propose is maybe to use `@noflip` instead of a conditional rendering

Correct me if I'm wrong, but what which `rule` would we ask to not get `flipped` in `rtl`? We currently don't have any directional styles in pickers arrow switcher component. 🤔

flaviendelangle: OK so basically, either we have a solution where developers can change the icon without caring about the direction, because the slot is bound to the position (left / right)

Or we have a solution where developers can change direction-related prop (like the title in your example) without caring about the direction, because the slot is bound to the action (go to previous / next)

____

I think I prefer to stick with the slot bound to the position.
Because it impacts one of the most basic customization: change the icon.

Whereas for the props, the title comes from the locale and should probably never be changed by prop.
And I don't see which prop you would like to customize that wants to know the current position of the icon (or button).
But maybe I'm missing some.

___

> where rightArrowIcon props would be applied either to the right (next) or left (previous) icon,

No, with my fix `rightArrowIcon` always go to the icon rendered on the right.
In LTR it's the next and on RTL it's the previous.
But it is always the one on the right.
Otherwise the naming would make no sense.
___

> I'm simply in favour of them being renamed to align with all the other props and changes moving towards previous/next instead of left/right 

But those are the props of `PickersArrowSwitcher`, not the props of `LeftArrowIcon`, `RightArrowIcon` and the 2 buttons.

flaviendelangle:
@samuelsycamore We changed the behavior and thus the doc quite a bit :angel:

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx
alexfauquette: Not sure to get all the debate.

As far as I understand, it's about using right/left of prev/next. The problem is that with `rtl` the element order is modified and so the first element is put on the right instead of the left so you have to switch them to be consistent.

I took a look at the [`TablePagination`](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/TablePagination/TablePaginationActions.js#L57-L66) in core which have the same problem. It sound like what they do is a distinction between the icon and the button.

For example, they have a `backIconButtonProps` (or `PreviousButton`) which render the following icon
```jsx
{theme.direction === 'rtl' ? <KeyboardArrowRight /> : <KeyboardArrowLeft />}
```

Having a `PreviouseIconButton`, `NextIconButton`, `ArrowLeftIcon`, and `ArrowRightIcon` could solve the problem because:
- buttons are always at the right place thanks to the browser `rtl` switch.
- `ArrowLeftIcon` defines an arrow pointing to the left instead of `LeftArrowIcon` which defines the arrow at the left of the component

LukasTy:
LGTM as is, but @alexfauquette has made some very insightful comments 👀

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx
flaviendelangle: > ArrowLeftIcon defines an arrow pointing to the left instead of LeftArrowIcon which defines the arrow at the left of the component

So if we have others arrow icons elsewhere in the codebase of the pickers, would you share the slot ?
If not, the name is maybe too generic (and is already too generic today, I'm just thinking about that now).

The behavior you describe is OK for me.

docs/data/date-pickers/custom-components/custom-components.md
samuelsycamore: That's an interesting idea @flaviendelangle. Maybe @danilo-leal or @gerdadesign have thoughts about this?

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -120,3 +120,29 @@ This can be used in combination with `componentsProps`.
...
+
+:::info
+Slots:
+
+- `LeftArrowIcon`
+- `RightArrowIcon`
+- `LeftArrowButton`
+- `RightArrowButton`
+
+:::
```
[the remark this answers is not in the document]
samuelsycamore: Agreed @alexfauquette, it creates a lot of dead space to the right side of the list. Does this need to be a callout at all? What if we just listed them directly in the text? (Suggestion below.)

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -120,3 +120,29 @@ This can be used in combination with `componentsProps`.
...
+
+The following slots let you customize how to render the buttons and icons for an arrow switcher component—the component to navigate to the "Previous" and "Next" steps of the picker.
```
samuelsycamore, line 136:
```suggestion
The following slots let you customize how to render the buttons and icons for an arrow switcher component—the component to navigate to the "Previous" and "Next" steps of the picker: `LeftArrowIcon`, `RightArrowIcon`, `LeftArrowButton`, `RightArrowButton`.
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -120,3 +120,29 @@ This can be used in combination with `componentsProps`.
...
+
+### Component props
+
+You can pass props to the icon and buttons as show below:
```
samuelsycamore, line 140:
```suggestion
You can pass props to the icons and buttons as shown below:
```

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -120,3 +120,29 @@ This can be used in combination with `componentsProps`.
...
+
+{{"demo": "ArrowSwitcherComponentProps.js", "defaultCodeOpen": false, "disableAd": true}}
+
+### Component
+
+You can pass custom components, for example to replace the icons as show below:
```
samuelsycamore, line 146:
```suggestion
You can pass custom components—to replace the icons, for example—as shown below:
```

docs/data/date-pickers/custom-components/custom-components.md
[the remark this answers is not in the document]
flaviendelangle: My goal would be at some point to document the slot props correctly in the API section and to link from here (or from the bottom of the page if we split the page into several pages)

docs/data/date-pickers/custom-components/custom-components.md
gerdadesign: Not sure I'm following — can you provide more context?

flaviendelangle: Sorry if I explains things that are already obvious to you:

In our components, we have *slots* which are parts of UI that users can customize by passing props to them or override with there own component.
For instance `<DataGrid components={{ Toolbar: MyToolbar }} />`, `<DatePicker components={{ Input: JoyTextField }} />` (not implemented yet for the 2nd one).

In the doc, we have:
- a `Slots` section in the API page that lists the slots ([here][https://mui.com/x/api/date-pickers/date-picker/#slots)).

- a `Custom components` ([data grid][https://mui.com/x/react-data-grid/components/), [pickers](https://mui.com/x/react-date-pickers/custom-components/)) that gives detailed example about how to use the slots. Each section of these pages can cover one or several slots (for instance in this PR we have 4 slots that works together so we document them in the same section).

My main goal was to provide a link between the two.
I would like to be able to correctly document the props of the slots on the API section.
And from the `Custom components` pages to be able to access those props by clicking the name of the slot.

packages/x-date-pickers/src/internals/components/PickersArrowSwitcher/PickersArrowSwitcher.types.tsx
alexfauquette: Sorry, I missed your answer

Sharing the slot is fine. For a given component, it sounds normal that the arrows to switch between views (or months) are similar

flaviendelangle: Fixed
I'll check that it works in the full RTL setup

flaviendelangle: Works fine https://codesandbox.io/s/date-and-time-pickers-forked-huxr57

flaviendelangle:
Got an approval of @alexfauquette on Slack
Merging to be ready for next release

PR #5892, opened 2022-08-24
[core] Fix Typescript error on field hooks

flaviendelangle:
#5397 broke the typing of the fields
The fix is not trivial so I just added some type casting to avoid the error and unlock master.

Will automerge


flaviendelangle:
The clean implementation is on https://github.com/mui/mui-x/pull/5899

PR #5895, opened 2022-08-24
[DataGridPremium] Export column grouping in Excel

### LINKED ISSUE #5878: [data grid] Reflect column grouping in excel export
Follow up of #5133

alexfauquette:
Fix #5878
To test the behavior: https://codesandbox.io/s/silent-fast-ogp38t?file=/demo.tsx

### libre office

image

### G-sheet
image


DanailH (approved):
It's great that we can support this!
Should we also add an additional test in here `exportExcel.DataGridPremium.test.tsx` to ensure we don't break it in the future?

PR #5899, opened 2022-08-25
[pickers] Use defaultized hook pattern for field hooks

flaviendelangle:


packages/x-date-pickers-pro/src/DateRangeField/useDateRangeField.ts
```diff
@@ -125,15 +132,32 @@ export const dateRangeFieldValueManager: FieldValueManager<
   hasError: (error) => error[0] != null || error[1] != null,
 };
 
+export const useDefaultizedDateRangeFieldProps = <TInputDate, TDate, AdditionalProps extends {}>(
```
LukasTy, line 135: Especially after this change it would be great to clean up the `...other` props being passed down to `TextField`, which now also include the default validation state props

flaviendelangle: Not sure to understand what you mean

LukasTy (approved):
Other than the fact that we now have even more error messages due to extraneous props being passed down to `TextField` this looks good small alignment with recent changes to usual pickers 👍

packages/x-date-pickers-pro/src/DateRangeField/useDateRangeField.ts
flaviendelangle: We can't do that inside `useField` because the hook is not aware of the props needed to validate.
It must be done inside `useDateField` and `useDateRangeField`.

But returning and `inputProps` containing stuff that must not be passed to the input is weird.
So maybe we should either 

1. Rename `inputProps` in `useField` into `props` or `enrichedProps`
2. Change the params of `useField` to take an object with the props to forward and the other. An inspiration could be the `useSlotProps` in `@mui/base` (`forwardedProps` and `externalProps` for instance)

PR #5901, opened 2022-08-25
[pickers] Improve DOM event management on `useField`

flaviendelangle:
- Focus the clicked section even when the input is not currently focused
- Call the DOM events provided to the hook

https://deploy-preview-5901--material-ui-x.netlify.app/x/react-date-pickers/date-field/


oliviertassinari:
I have quickly played with this problem:

```diff
diff --git a/packages/x-date-pickers/src/internals/hooks/useField/useField.ts b/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
index 0f5642aa0..3447f0a51 100644
--- a/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
+++ b/packages/x-date-pickers/src/internals/hooks/useField/useField.ts
@@ -317,8 +317,13 @@ export const useField = <
   });

   const handleInputFocus = useEventCallback(() => {
-    // TODO: Avoid applying focus when focus is caused by a click
-    updateSelectedSections(0, state.sections.length - 1);
+    setTimeout(() => {
+      if (inputRef.current.selectionEnd - inputRef.current.selectionStart === 0) {
+        handleInputClick();
+      } else {
+        updateSelectedSections(0, state.sections.length - 1);
+      }
+    })
   });

   const handleInputBlur = useEventCallback(() => updateSelectedSections());
```

applied on HEAD seems to solve the problem too.

LukasTy:
The current implementation introduces the following regression.
Click on section -> blur -> focus with tab -> arrow navigation does not work.
The same issue is not happening with the solution suggested by Olivier, which also looks more simple. 🤔 
What do you think @flaviendelangle ?

https://user-images.githubusercontent.com/4941090/186891938-7de26764-4155-4ce7-ae20-3f394a9d33cc.mov

flaviendelangle:
@LukasTy I can implement it of course

flaviendelangle:
Done

flaviendelangle:
The approach we went for do not work on Android
I'll try to find a workaround later

PR #5912, opened 2022-08-26
[DataGrid] Fix `ownerState` undefined in theme style overrides

### LINKED ISSUE #4413: [data grid] Undefined ownerState when customizing theme.
### Duplicates

### Latest version

### Current behavior 😯

When trying to customize themes using [overrides based props](https://mui.com/customization/theme-components/#overrides-based-on-props), the ownerState passed as parameter is undefined.

### Expected behavior 🤔

ownerState to contain props.

### Steps to reproduce 🕹

Steps:

1. Open https://codesandbox.io/s/datagriddemo-theme-override-hwj2y8?file=/demo.js
2. Debug message is displayed on the console.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

iigrik:
Fixes #4413 .

### DIFF
--- a/packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
+++ b/packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
@@ -9,6 +9,8 @@ import Paper from '@mui/material/Paper';
 import Popper, { PopperProps } from '@mui/material/Popper';
 import { useGridApiContext } from '../../hooks/utils/useGridApiContext';
 import { isEscapeKey } from '../../utils/keyboardUtils';
+import { useGridRootProps } from '../../hooks/utils/useGridRootProps';
+import { DataGridProcessedProps } from '../../models/props/DataGridProps';
 
 export interface GridPanelClasses {
   /** Styles applied to the root element. */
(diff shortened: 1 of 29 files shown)


packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
```diff
@@ -54,6 +58,8 @@ const GridPaperRoot = styled(Paper, {
 const GridPanel = React.forwardRef<HTMLDivElement, GridPanelProps>((props, ref) => {
   const { children, className, classes: classesProp, ...other } = props;
   const apiRef = useGridApiContext();
+  const rootProps = useGridRootProps();
+  const ownerState = rootProps;
```
m4theushw, line 62: For `GridRoot` it's correct to pass the root props to the `ownerState`, but for any other component I think we should keep only the props received directly as owner state. We may in the future allow `GridPanel` to be used outside the grid, so the root props wouldn't be available. Could you make this change to all components, only keeping `GridRoot`?

```suggestion
  const ownerState = props;
```

m4theushw (commented):
Thanks for the contribution. Sorry for the late review. 😳

packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
iigrik: @m4theushw  Thanks for the response. I believe passing only internal props as `ownerState` is incorrect - in that case we are loosing public props of component.

As mentioned in  [`styleOverrides` docs](https://mui.com/material-ui/customization/theme-components/#overrides-based-on-props):

> The ownerState prop is a combination of public props that you pass to the component + internal state of the component.

vntech-engineering53:
Hi, any timeline on when this fix will be shipped?

PR #5938, opened 2022-08-28
[core] Fix scroll restoration in the docs

oliviertassinari:
https://deploy-preview-5938--material-ui-x.netlify.app/x/introduction/

- Fix the bug reported in https://github.com/mui/material-ui/pull/34037
- Remove duplication of Next.js configuration. We should be able to abstract most of the docs infrastructure in the mono-repo. We are one step closer to having our own MUI https://github.com/facebook/docusaurus.


flaviendelangle (approved):

oliviertassinari:
I have found an issue with this approach, it leads to a warning on the core's side. We need to change the import a bit.

flaviendelangle (approved):

PR #5941, opened 2022-08-29
[pickers] Do not require `date-fns` in `@mui/x-date-pickers-pro`

### LINKED ISSUE #5916: [pickers] Export of parseNonNullablePickerDate is missing
### Duplicates

### Latest version

### Current behavior 😯

When I used [5.0.0-beta.7] version , I got the import **parseNonNullablePickerDate** error , because it was called from /node_modules/@mui/x-date-pickers-pro/DateRangePicker/shared.js , but not exported in the index.js

### Expected behavior 🤔

it should work 

### Steps to reproduce 🕹

import { LocalizationProvider } from '@mui/x-date-pickers-pro';
import { DateRangePicker } from '@mui/x-date-pickers-pro/DateRangePicker';

and use it with **5.0.0-beta.7** version

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #5916

### DIFF
--- a/packages/x-date-pickers/src/internals-fields/index.ts
+++ b/packages/x-date-pickers/src/internals-fields/index.ts
@@ -1,0 +1,15 @@
+// We don't add those exports to the regular `@mui/x-date-pickers/internals`,
+// Because they rely on date-fns, which is not used by all applications.
+export {
+  useField,
+  createDateStrFromSections,
+  addPositionPropertiesToSections,
+  splitFormatIntoSections,
+} from '../internals/hooks/useField';
+export type {
+  UseFieldProps,
+  UseFieldParams,
+  UseFieldResponse,
+  FieldValueManager,
+  FieldSection,
+} from '../internals/hooks/useField';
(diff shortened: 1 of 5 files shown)


alexfauquette (approved):
That's an unexpected consequence of the experiment 🤯

I confirm the regression is fixed by this PR
issue reproduction: https://codesandbox.io/s/admiring-bird-djug44?file=/index.tsx
issue reproduction with the fix: https://codesandbox.io/s/bold-germain-jt0dsl?file=/demo.tsx:133-170

packages/x-date-pickers/src/internals-fields/index.ts
```diff
@@ -0,0 +1,15 @@
+// We don't add those exports to the regular `@mui/x-date-pickers/internals`,
+// Because they rely on date-fns, which is not used by all applications.
+export {
+  useField,
```
LukasTy, line 4: I'd say that we should still look into removing the direct dependency of `date-fns` from `useField.utls` going forward 🤔

flaviendelangle: Yes I think we all agree on that
I requires new methods in the adapter so it was not feasable for a simple POC
But the components can't be stable with hardcoded date-fns stuff

PR #5943, opened 2022-08-29
[docs] Label api.updateRows as a Pro feature

oliviertassinari:
From what I understand of https://mui.com/x/react-data-grid/row-updates/#the-updaterows-method, this is a Pro-only feature. I have noticed this while working a bit on https://www.notion.so/mui-org/apiRef-MIT-vs-Commercial-c81b2e6fee4d404880fad5bd25836739.


LukasTy (approved):

flaviendelangle:
@oliviertassinari this file is auto-generated

Right now we only add the label to methods that are only added to the pro api
And this method is also added to the community api, even though it will not be used very often since you can't access the api object from everywhere

PR #5947, opened 2022-08-29
[DataGrid] Allows keyboard navigation in group header

alexfauquette:


packages/grid/x-data-grid/src/hooks/features/columnGrouping/useGridColumnGrouping.ts
```diff
@@ -149,24 +85,68 @@ export const useGridColumnGrouping = (
...
+        ...state,
+        columnGrouping: {
+          ...state.columnGrouping,
+          headerStructure: columnGroupsHeaderStructure,
+        },
+      };
+    });
+  }, [apiRef, props.columnGroupingModel]);
+
+  useGridApiEventHandler(apiRef, 'columnOrderChange', handleColumnReorderChange);
```
DanailH, line 110: Shouldn't that be handled in the `useGridColumnReorder` hook?

packages/grid/x-data-grid/src/hooks/features/columnGrouping/useGridColumnGrouping.ts
```diff
@@ -103,15 +48,6 @@ const createGroupLookup = (columnGroupingModel: GridColumnNode[]): GridColumnGro
 
   return { ...groupLookup };
 };
-export const columnGroupsStateInitializer: GridStateInitializer<
```
DanailH, line 106: I may be lost but does this mean that the groups are no longer persisted in the state? Shouldn't we initialize an empty `columnGrouping` state?

packages/grid/x-data-grid/src/hooks/features/density/densitySelector.ts
```diff
@@ -18,17 +18,7 @@ export const gridDensityHeaderHeightSelector = createSelector(
   (density) => density.headerHeight,
 );
 
-export const gridDensityHeaderGroupingMaxDepthSelector = createSelector(
```
DanailH, line 21: Nice for removing these - it doesn't make much sense for them to be in the density selectors file

DanailH (commented):
Looks great! The only thing I would consider adding is a check if a column header doesn't have content (it's empty) not to be able to navigate in it. Idk how difficult lit would be tho.

alexfauquette:
@DanailH It was my first it do not navigate to empty cell but then comes the question about where does the focus go. And their is no obvious answer to that, so I looked at other libraries, and they just let navigate in empty column group. For example [AG grid column group](https://www.ag-grid.com/javascript-data-grid/column-groups/)

packages/grid/x-data-grid/src/hooks/features/columnGrouping/useGridColumnGrouping.ts
alexfauquette: It's unclear for me when an initializer is necessary. I thaught it was used to initialise the state before pipe processors are called.

Since column grouping is not used by any processors, I removed the initialization. But if it's needed for other purpose, I can bring initializer back

alexfauquette: Ok, since the state is reused by `columnReorder` and `columnHeader` I put back the initializer

PR #5958, opened 2022-08-30
[pickers] Support mobile edition

flaviendelangle:


packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -133,6 +133,125 @@ export const useField = <
...
+    );
+
+    if (keyPressed.length > 1) {
+      // TODO: Might be able to support it in some scenario
+      return;
+    }
+
+    const isNumericValue = !Number.isNaN(Number(keyPressed));
+
+    if (isNumericValue) {
```
flaviendelangle, line 186: The content below was just moved from the `handleInputKeyDown` method, no logic change

flaviendelangle:
The behavior is not perfect yet but I think it's ok to start the review

LukasTy:
A few notes:
- have you looked into the reason why on android the initial focus always jumps to the whole input instead of a clicked section?
- what do you think about adding `type="tel"` to the `input` to achieve the same behaviour as with masked pickers (number keyboard)
- delete/backspace situation on mobile is "not the best" 🤔 
- honestly, after all this testing and going into an [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) to test the native input experience is such a relief. I'm starting to doubt the usefulness of out `DateField` implementation for mobile usage. Simply put, I doubt anyone would want to use the `DateField` on mobile instead of picker / native date input.

flaviendelangle:
> have you looked into the reason why on android the initial focus always jumps to the whole input instead of a clicked section?

I just saw that the check Olivier made is not working but I don't know why yet. I'll investigate (or if you want to spend some time on it)

> what do you think about adding type="tel" to the input to achieve the same behaviour as with masked pickers (number keyboard)

Does it impact what can be rendered ? (for the separator like `/`).
We can have a look, it would be nice to even only open in tel more only when you focus a number section but that might not be doable :laughing: 

> delete/backspace situation on mobile is "not the best" thinking

Yes, that's something I did not work on yet

> honestly, after all this testing and going into an [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) to test the native input experience is such a relief. I'm starting to doubt the usefulness of out DateField implementation for mobile usage. Simply put, I doubt anyone would want to use the DateField on mobile instead of picker / native date input.

The way I see it, if we can provide an okayish behavior on mobile, it's better than having it totally broken like Telerik.
But clearly our doc should say that `DateField` & co and intended for a desktop usage and that we don't recommend using them for a mobile / responsive application.
That for a responsive application, people should use the `DatePicker` & co and disable the view part on desktop with a prop if they want.

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -130,6 +141,128 @@ export const useField = <
...
+        startOfDiffIndex = i;
+      }
+
+      if (
+        endOfDiffIndex === -1 &&
+        prevValueStr[prevValueStr.length - i - 1] !== valueStr[valueStr.length - i - 1]
+      ) {
+        endOfDiffIndex = i;
+      }
+    }
```
alexfauquette, line 165: I'm not super confident with this method. I'm surprised it does not bugg when a character of the previous value is included in the new one.

For example if it was 12 and you copy-past 22, this algo will detect that a 1 has been added.

I would feel safer by using `getValueStrFromSections` to generate `beforeSections` and `afterSections` which would be the string generated before the active section and after the active section.

For that you can get the active section: with `valueStr.slice(beforeSections.lenght, valueStr.length - afterSection.length)`

And you can check their is not modification outside with `valueStr.slice(0, beforeSections.lenght) === beforeSections`

flaviendelangle: > I would feel safer by using getValueStrFromSections to generate beforeSections and afterSections which would be the string generated before the active section and after the active section.

I think it does not work on Android since on the 2nd call the value is the value with the selected section cleaned but the sections in state is not cleaned.
Maybe we can keep the section up to date with this weird behavior.

I'd prefer to merge #6141 which touches a lot those behaviors.

flaviendelangle:
Rebased upon #6141 
Clean diff added to the PR description

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -21,8 +21,17 @@ import {
...
 } from './useField.utils';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const debug = (...messages: (string | number | boolean | null | undefined)[]) => {
+  const node = document.querySelector('#debug');
+  if (node) {
+    node.innerHTML += `${messages.join(' ')}<br />`;
+  }
+};
+
```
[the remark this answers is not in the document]
flaviendelangle: It will not be released yes
It's super convenient to debug on mobile for now

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -101,6 +109,201 @@ export const useField = <
     setSelectedSections(null);
   });
 
+  const handleInputPaste = useEventCallback((event: React.ClipboardEvent<HTMLInputElement>) => {
```
flaviendelangle, line 112: I'm adding an error because letting `onChange` handle it would not work.
Implementing it will be an interesting follow up.

joserodolfofreitas:
Considering the numeric keyboard doesn't have arrows,
I think we need to automatically trigger "next" if a string of numbers is valid.
For example: If the user types '12' or '05' to enter the month, the component immediately moves forward to input the day
in a MM/dd/yyy format.

Is that something feasible to achieve?

flaviendelangle:
Didn't we already discuss it several time ?
As mentioned on Monday, Telerik dropped this behavior on their last major because it was to much pain to maintain.
So that's not something I would implement lightly.

alexfauquette:
About the automatic switch to the next section, it could be a topic for another PR. Switching when pressing either '/', ' ', or '-' plus the ability to touche the next section sounds good enough for a first attempt

I don't know if it's related to this PR. but most of the fields ([control/uncontrolled demo](https://deploy-preview-5958--material-ui-x.netlify.app/x/react-date-pickers/date-field/#uncontrolled-vs-controlled) for example) are buggy. When I press a digit (on desktop) for day or month, it retruns `Invalid Date/Invalid Date/Invalid Date`

flaviendelangle:
@alexfauquette Same bug on `next`, but I'll have a look on this PR since the impacted code moves a lot

flaviendelangle:
Should be fixed, I'll add tests for the digit editing (which is basically not tested for now)

joserodolfofreitas:
To be honest, I missed the fact that users can navigate forward with ' ' and '-'.
If we could use the '/'  (or whichever character used by the user on the date format), that'd be a nice improvement.

flaviendelangle:
> If we could use the '/' (or whichever character used by the user on the date format), that'd be a nice improvement.

You can 
It could be improved to match the date separator as you said, for now I'm using the 3 most common which cover most of the use cases.

docs/data/date-pickers/fields/fields.md
```diff
@@ -13,3 +13,4 @@ They should not be used in a production setup.
 
 - [DateField](/x/react-date-pickers/date-field/)
 - [DateRangeField](/x/react-date-pickers/date-range-field/)
+  B
```
alexfauquette, line 16:
```suggestion
```

alexfauquette (approved):

PR #5972, opened 2022-08-31
[DateTimePicker] Extract `tabs` component from `toolbar`

### LINKED ISSUE #5841: [StaticDateTimePicker] Show DateTimeToolbar when showToolbar is true and wrapperVariant is 'desktop'
https://github.com/mui/mui-x/blob/b5111dd5bc8677b1e0135f14c91124586106b0ec/packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/CalendarOrClockPicker.tsx#L114

Due to the line of code linked above, it doesn't seem possible to get a `StaticDateTimePicker` to show it's toolbar when `displayStaticWrapperAs="desktop"`. I'd like to do exactly that, but I can't see a way to do that via the props API.

This feels like an oversight since `showToolbar` works as expected on `StaticDatePicker` and `StaticTimePicker` while in desktop mode. Since there is a `DateTimePicker` `toolbarComponent` that's set on the prop interface of `StaticDateTimePicker` [here](https://github.com/mui/mui-x/blob/master/packages/x-date-pickers/src/StaticDateTimePicker/StaticDateTimePicker.tsx#L77) I'm not sure why it's only allowed when the `wrapperVariant !== 'desktop'`.

Could that additional condition be removed or modified in some way to allow showing the toolbar regardless of `wrapperVariant`?

### LINKED ISSUE #5959: [MobileDateTimePicker] showToolbar prop not working correctly
### Duplicates

### Latest version

### Current behavior 😯

I'm trying to hide only Toolbar with `showToolbar={false}` but this prop hides also tabs even if I add `hideTabs={false}`

### Expected behavior 🤔

`showToolbar={false}` hides only Toolbar

### Steps to reproduce 🕹

Link to live example:

Steps:
1.
2.
3.
4.

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #5959
Fixes #5841  

DateTimePicker tabs component is currently inside of `toolbar` and its control is dependant of toolbar visibility.

Extracting tabs out of `Toolbar` and putting it in `CalendarOrClockPicker` enables more correct control of visibility using `showToolbar` and `hideTabs` props.

Add `DateTimePicker` case in `releaseChangelog` script.

Refactor `TabsComponent` customisation to be done via component (`tabs`) slots.

[Documentation/customisation example](https://deploy-preview-5972--material-ui-x.netlify.app/x/react-date-pickers/custom-components/#tabs)


packages/x-date-pickers/src/DateTimePicker/DateTimePickerToolbar.tsx
```diff
@@ -198,14 +190,6 @@ export const DateTimePickerToolbar = <TDate extends unknown>(
...
       )}
-      {showTabs && (
-        <DateTimePickerTabs
-          dateRangeIcon={dateRangeIcon}
-          timeIcon={timeIcon}
-          view={openView}
-          onChange={setOpenView}
-        />
-      )}
     </React.Fragment>
```
alexfauquette, line 193: Seems that `React.Fragment` is not used anymore.

What about removing this `wrapperVariant !== 'desktop'` on top and use it as a default value for `showToolbar`

alexfauquette: Related to #5841

packages/x-date-pickers/src/DateTimePicker/shared.ts
```diff
@@ -59,6 +61,11 @@ export interface BaseDateTimePickerProps<TInputDate, TDate>
    * @default DateTimePickerToolbar
    */
   ToolbarComponent?: React.JSXElementConstructor<BaseToolbarProps<TDate, TDate | null>>;
+  /**
+   * Component that will replace default tabs renderer.
+   * @default DateTimePickerTabs
+   */
+  TabsComponent?: React.JSXElementConstructor<DateTimePickerTabsProps>;
   /**
```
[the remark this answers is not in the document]
LukasTy: I've left it the same as is for the `ToolbarComponent` at the moment as they are/were closely related.
What does @flaviendelangle think about it?
Should we move these (ToolbarComponent & TabsComponent) props to `components` while we are at it?
IMHO, we should leave this part for breaking changes in `v6` 🤔

packages/x-date-pickers/src/DateTimePicker/DateTimePickerToolbar.tsx
LukasTy: Thanks for pointing this out. 🙇 
By the way, we already default to your mentioned condition via 
```js
const toShowToolbar = showToolbar ?? wrapperVariant !== 'desktop'
```
So, removing this hard-coded logic in Toolbar will easily solve 5841 🎉 👍

alexfauquette: 😂
- issue solved: 1
- line of code added: -1

packages/x-date-pickers/src/DateTimePicker/shared.ts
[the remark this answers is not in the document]
flaviendelangle: I also think `ToolbarComponent` should become a component slot in v6 
But it would not bother me to put the tabs in the slots right away
I would be in favor to avoid future migrations whenever we can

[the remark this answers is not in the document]
LukasTy: I've created `PickerTabs` component slot instead. 😉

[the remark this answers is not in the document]
flaviendelangle: Do you think we could skip the `picker` prefix ?
We are inside a picker component to I guess it's obvious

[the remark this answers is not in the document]
LukasTy: Fair point. 🤔 
Looking at `ActionBar`.
It simply seemed strange to have so short and generic name. 😆 
Initially I tried `TabsComponent`, but `Component` part seemed unnecessary when it's already part of `components` slots.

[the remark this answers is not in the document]
flaviendelangle: I guess `Tabs` is enough :laughing:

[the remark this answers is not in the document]
LukasTy: Done 🙈

docs/data/date-pickers/custom-components/Tabs.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import dayjs, { Dayjs } from 'dayjs';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
+import { TextField, Box } from '@mui/material';
+import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+import {
+  DateTimePickerTabs,
+  DateTimePickerTabsProps,
+} from '@mui/x-date-pickers/DateTimePicker/DateTimePickerTabs';
```
flaviendelangle, line 10:
```suggestion
} from '@mui/x-date-pickers/DateTimePicker';
```
An import of this depth will not work with some building process because we don't provide a `package.json` in the folder.
If we are using this component in a demo, we should export it from `@mui/x-date-pickers/DateTimePicker` and `@mui/x-date-pickers`

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -90,3 +90,33 @@ You can provide any custom component to this slot as long as it passes `children
 In the examples below we provide ways of implementing date range shortcuts using this slot.
 
 {{"demo": "PaperContentComponent.js", "defaultCodeOpen": false, "disableAd": true}}
+
+## Tabs
+
+The tabs are available on all date time picker components.
+By default, it is `hidden` on desktop, and `visible` on mobile.
+This behavior can be overridden by setting `hideTabs` prop.
```
flaviendelangle, line 98: Maybe for v6: Should we drop the `hideTabs` prop in favor of a `hide` prop to the `Tabs` slot that makes the component return null ?

LukasTy: That could be quite a logical change. 🤔 
Decoupling the props from higher level/order components and changing them directly on the components in question. 👍

docs/data/date-pickers/custom-components/Tabs.tsx
LukasTy: Done. Re-exported `DateTimePickerTabs` in `DateTimePicker/index`

docs/data/date-pickers/custom-components/Tabs.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+import dayjs, { Dayjs } from 'dayjs';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
+import { TextField, Box } from '@mui/material';
```
flaviendelangle, line 5:
```suggestion
import TextField from '@mui/materia/TextField';
import Box from '@mui/materia/Box';
```

We avoid root import from `@mui/material` 
We have some for utilities that are only exported from the root and it's not great because it bundles the whole package on dev (which is quite heavy).
But I think it's worth doing the deep imports on the doc whenever possible to give good habits to our users.

packages/x-date-pickers/src/internals/components/CalendarOrClockPicker/CalendarOrClockPicker.tsx
```diff
@@ -158,6 +173,15 @@ export function CalendarOrClockPicker<TDate, View extends CalendarOrClockPickerV
           toggleMobileKeyboardView={toggleMobileKeyboardView}
         />
       )}
+      {showTabs && (
```
flaviendelangle, line 176: We can avoid creating an empty component above by adding a check here

```suggestion
      {showTabs && !!TabsComponent && (
```

flaviendelangle (approved):
Two small suggestions but other than that it is great !

flaviendelangle (approved):

PR #6003, opened 2022-09-02
[DataGrid] Remove `GridEvents` enum

### LINKED ISSUE #4684: [data grid] Remove `GridEvents` enum

flaviendelangle:
Closes #4684


DanailH:
I've no objections to this change. If we are doing this then it probably makes sense to also remove, as part of a separate PR (I can take care of it), the `GridFeatureModeConstant` and just keep the type.

Also, we can point this to `next` on Friday and merge it.

flaviendelangle:
@DanailH if you can approve this one

DanailH (approved):

Issue #5844, opened 2022-08-19
[pickers] Allow to override some local texts directly on a picker component


oliviernt:
### Order ID 💳

46586

### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### The problem in depth 🔍

Hello,

We're using the new DateRangePicker with LocalizationProvider but are now encountering some unexpected behavior regarding labeling.

We want to be able to label different pickers with different labels depending on their form context. For example one will have a start label "Absence from" while another might only have "From" or "Start at".

From what I understand, this currently required multiple LocalizationProviders as it's where the labels currently are defined. See https://mui.com/x/react-date-pickers/localization/

Is there a way to do this without defining multiple LocalizationProviders or it that the way to go? It feels counter intuitive to us at the moment.

Our preferred way of doing this would be to use the `startText` and `endText` props but those are deprecated. See https://mui.com/x/api/date-pickers/date-range-picker/#props

Thank you for your help on this issue.

Best regards
Olivier

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```  System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.0/bin/yarn
    npm: 8.7.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Chrome: 104.0.5112.101
    Firefox: 103.0.2
  npmPackages:
    @emotion/react: ^11.7.1 => 11.9.3 
    @emotion/styled: ^11.6.0 => 11.9.3 
    @mui/base:  5.0.0-alpha.86 
    @mui/icons-material: ^5.2.5 => 5.8.4 
    @mui/lab: ^5.0.0-alpha.54 => 5.0.0-alpha.87 
    @mui/material: ^5.2.7 => 5.8.5 
    @mui/private-theming:  5.8.4 
    @mui/styled-engine:  5.8.0 
    @mui/styles: ^5.2.3 => 5.8.4 
    @mui/system: ^5.2.6 => 5.8.5 
    @mui/types:  7.1.4 
    @mui/utils:  5.8.4 
    @mui/x-date-pickers:  5.0.0-beta.1 
[...]

PR #6011, opened 2022-09-02
[pickers] Allow nested `LocalizationProvider`

flaviendelangle:
Part of #5844

This PR focuses on supporting nested `LocalizationProvider` so that we can add a small wrapper on all pickers component calling `LocalizationProvider` with `props.localeText`.
This will also allow users to create one top level `LocalizationProvider` with the adapter (or even in the theme) and create `LocalizationProvider` deeper in the tree if they want to adapter the `adapterLocale` or the `localeText`.

I added le `localeText` on `DesktopDatePickers` for demonstration purpose.

____

- [x] Remove `LocalizationProvider['locale']`


alexfauquette (approved):
Sounds good 👍

flaviendelangle:
I'm removing the `localeText` of `DesktopDatePicker`
I'll add it to all components in a follow up PR, it was just for demonstration purpose.

PR #6016, opened 2022-09-02
[DataGrid] Remove legacy editing API

m4theushw:


DanailH (approved):

DanailH:
@m4theushw I guess this solves that issue https://github.com/mui/mui-x/issues/5606

The question is does it also save the rest that is listed under the "Editing" section here https://github.com/mui/mui-x/issues/3287

Issue #5856, opened 2022-08-20
[pickers] Hydration mismatch


oliviertassinari:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

<img width="600" alt="Screenshot 2022-08-20 at 20 22 04" src="https://user-images.githubusercontent.com/3165635/185761180-f12116fc-8a92-4c82-884b-e59dbabced92.png">

https://reactjs.org/docs/error-decoder.html/?invariant=425

### Expected behavior 🤔

No hydration error, like in v5.15.1.

### Steps to reproduce 🕹

Steps:
1. Open the [v5.15.2](https://app.netlify.com/sites/material-ui-x/deploys/62f50b1265a7ea0008f60dad) permalink release https://62f50b1265a7ea0008f60dad--material-ui-x.netlify.app/x/react-date-pickers/date-picker/
2. Open the dev tools

### Context 🔦

Looking at the diff https://github.com/mui/mui-x/compare/v5.15.1...v5.15.2, I suspect it's related to React 18.

### Your environment 🌎

v5.15.2

oliviertassinari:
I had a quick look at what's going on. It seems that React 18 is more strict when it comes to hydration mismatches. As of today, in https://62f50b1265a7ea0008f60dad--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#custom-input-component, I see the following

Before hydration:

<img width="541" alt="Screenshot 2022-08-20 at 21 33 15" src="https://user-images.githubusercontent.com/3165635/185763438-c1ca5f10-b2ef-44e8-ad45-0e3ae957d553.png">

After hydration:

<img width="498" alt="Screenshot 2022-08-20 at 21 32 56" src="https://user-images.githubusercontent.com/3165635/185763432-256e6ac4-bda7-4212-bc6a-3974062e89d0.png">

Changing my macOS clock to the August 11th, 2022 resolves the hydration mismatch.

flaviendelangle:
Closing it
We no longer have this demo and we try to never rely on the current date-time to avoid this kind of problems

oliviertassinari:
> we try to never rely on the current date-time

Ok, so option 2. from the list above. Fair enough, simple.

PR #6032, opened 2022-09-05
[docs] Fix demonstration date to avoid hydratation errors

alexfauquette:
Fix partially #5856

I replaced the components values of `dayjs()` to a fix date.

The only one exception is the `DateFieldValidation` demo, because to test `disablePast` you need to know when is today. It could be done by using a `useEffect` but it would add complexity to the demo.


docs/data/date-pickers/date-field/DebouncedDateField.tsx
```diff
@@ -8,10 +8,10 @@ import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 
-const today = dayjs();
+const today = dayjs('2022-04-07');
 
 export default function DebouncedDateField() {
-  const [value, setValue] = React.useState<Dayjs | null>(() => today.add(1, 'week'));
+  const [value, setValue] = React.useState<Dayjs | null>(dayjs('2022-04-07'));
```
LukasTy, line 14: Was there any issues leaving the previous logic intact in here? 🤔 
This changes how the demo behaves.
```suggestion
  const [value, setValue] = React.useState<Dayjs | null>(today.add(1, 'week'));
```

alexfauquette: Why does it change the demo behavior? The goal was to set the date to something which is not today. Such that you can click on the "Set to today" button with a delay of 500ms

The problem is the same as others. If you put the default to `today.add(1, 'week')` nextjs will prepare the html for 15 of September. But tomorrow hydratation will override it with the 16 of September

LukasTy: > Why does it change the demo behavior?

I meant that now once you open this demo, the `set to today` is enabled, because the selected day differs from `today`, but in your proposal it is disabled as `today` === `selected date`.
Screenshot 2022-09-08 at 12 52 40


> The problem is the same as others. If you put the default to today.add(1, 'week') nextjs will prepare the html for 15 of September. But tomorrow hydratation will override it with the 16 of September

`today` is set to `2022-04-07`, should it really ever hydrate to anything than `2022-04-14`? 🤔

docs/data/date-pickers/date-field/DebouncedDateField.tsx
```diff
@@ -8,10 +8,10 @@ import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
 import Button from '@mui/material/Button';
 
-const today = dayjs();
+const today = dayjs('2022-04-07');
```
alexfauquette, line 11:
```suggestion
const today = dayjs();
```

docs/data/date-pickers/date-field/DebouncedDateField.tsx
LukasTy: But if we use dynamic `today` , won't this demo remain with the hydration issue? 🤔 
Maybe using a different CTA would be better?
Like `Reset date` or something along these lines?

alexfauquette: Is `today` the problem of is it to display an HTML element based on `today` value?

But yes of course fixing all the date with another cal to action does the job

LukasTy: The problem is a difference in the rendered UI at least based on errors being thrown locally as well, hence, if we will make sure the button `disabled` flag always renders as `false` - it should be all good. 👌

PR #6040, opened 2022-09-06
[core] Remove Storybook

flaviendelangle:
We discussed the Storybook yesterday: https://mui-org.notion.site/xGrid-Weekly-meeting-2022-09-05-931b28fff95b43d4905f328a884076b1.

It has received almost no new story in a year and is therefore totally outdated.
Right now, it is only an overhead that we have to maintain.

- [x] Move the `useData` and `getData` utilities that are being used in tests from the storybook to `@mui/x-data-grid-generator`
- [x] Remove the `storybook` internal package
- [x] Remove the Storybook from regression tests
- [x] Remove the Storybook from the CI and from our architecture files


oliviertassinari (commented):
It sounds good.

(*I have linked the meeting notes that talk about the change and made them public for future/community scrutiny.*)

test/regressions/webpack.config.js
```diff
@@ -24,6 +24,10 @@ module.exports = {
     new webpack.DefinePlugin({
       DISABLE_CHANCE_RANDOM: JSON.stringify(true),
     }),
+    new webpack.ProvidePlugin({
```
flaviendelangle, line 27: Copied from the core

https://github.com/mui/material-ui/blob/c8feb29320a22cc4461fb5ef5f4f7de7bd96a59b/test/regressions/webpack.config.js#L26
I think the Storybook was injecting a default config somehow.

packages/grid/x-data-grid-generator/src/services/basic-data-service.ts
```diff
@@ -0,0 +1,106 @@
...
+  currencyPair: string;
+  [price: string]: number | string;
+}
+
+export interface GridBasicData {
+  columns: GridColDef[];
+  rows: GridBasicRowModel[];
+}
+
+export const getBasicGridData = (rowLength: number, colLength: number): GridBasicData => {
```
DanailH, line 81: Nice, I guess this was part of the storybook package. I wanter if we can remove it altogether and just use either the `Commodity` or `Employee` data sets. If there are no assertions in the tests related to the value of the cells then replacing the data set should be more or less easy, what do you think?

flaviendelangle: Yep, it's the old `useData` and `getData`.

For the tests, it can be useful to have those dummy data tests without any intelligence behind them.
For instance in the pagination, it makes very clear what you are testing.

I would personally be in favor of an even dumber data set.
Something like:

```
{ id: 0, name: `name-0` },
{ id: 1, name: `name-1\ },
...
```

And to use it everywhere we have some "Nike" / "Puma" dataset.
To reduce as much as possible the complexity of understanding the dataset and focusing on the actual behavior we are testing.

packages/storybook/src/stories/grid-style.stories.tsx
```diff
@@ -1,219 +0,0 @@
...
-          {
-            id: 1,
-            text: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam a accumsan lacus. Etiam faucibus fermentum ipsum. Integer bibendum diam quis est sagittis laoreet. Pellentesque finibus lacus vitae erat aliquet dignissim.',
-          },
-        ]}
-      />
-    </div>
-  );
-};
-export const MultilineCustomCellContentSnap = () => {
```
oliviertassinari, line 195: Since the Argos CI step is broken, we don't have visibility on the scale of this, but here is one visual test that is gone. Will we miss these visual regression tests?

flaviendelangle: image

Here are all the stories with the `Snap` suffix
If the team wants to migrate some to a hidden doc example I can do it
I personally don't see any good candidate.

flaviendelangle: For instance the most recent `PinnedColumnWithCheckboxSelection` has an equivalent in the doc:
https://mui.com/x/react-data-grid/column-pinning/#pinning-the-checkbox-selection-column

oliviertassinari: > I personally don't see any good candidate.

Then I think that we are good

> to a hidden doc example

In the main repo, we store them here https://github.com/mui/material-ui/tree/master/test/regressions/fixtures (outside of the docs)

flaviendelangle: Oh I think we had pickers fixtures that I did not migrate :smile: 
I'll have a look to maybe bring back this logic later

DanailH (approved):

PR #5260, opened 2022-06-20
[license] Add new license status 'Out of scope'

flaviendelangle:
Fixes #5246

PR #6043, opened 2022-09-06
[core] Simplify testing architecture

flaviendelangle:
- [x] Use the `@mui/monorepo` `setupJSDOM` file to avoid duplicating code (equivalent of our current `setup.js` file which is badly named by the way).
  This has a side effect. The core adds the `url: 'http://localhost'` property to its JSDOM instance, which breaks our print tests because `printWindow.contentDocument.body` is now null (don't know why).
   If we don't find a fix, I can either copy / paste the `setupJSDOM` file to remove the `url` property or make it optional on the core, or skip the printing test when using JSDOM (what I currently did).

- [x] Use the same file naming strategy as in the core

- [x] Set the license key before each test to prepare https://github.com/mui/mui-x/pull/5260#discussion_r951593312

- [x] Avoid mocha hook duplication (for example we had a `sinon.restore` on both `setup.js` for `test:unit` and on `karma.test.js` for `test.karma`.


package.json
```diff
@@ -33,7 +33,6 @@
     "test:coverage:html": "cross-env NODE_ENV=test BABEL_ENV=coverage nyc mocha 'packages/**/*.test.{js,ts,tsx}' --exclude '**/node_modules/**' && nyc report --reporter=html",
     "test:karma": "cross-env NODE_ENV=test karma start test/karma.conf.js",
     "test:unit": "cross-env NODE_ENV=test node --expose_gc ./node_modules/.bin/mocha 'packages/**/*.test.{js,ts,tsx}' --exclude '**/node_modules/**' --timeout 3000",
-    "test:watch": "yarn test:unit --watch",
```
flaviendelangle, line 36: The core does not have that one and I don't think it's worth keeping
A simple `yarn test:unit --watch` works fine.

test/utils/setupBabel.js
```diff
@@ -1,3 +1,4 @@
 require('@babel/register')({
   extensions: ['.js', '.ts', '.tsx'],
+  ignore: [/node_modules\/(?!@mui\/monorepo)/],
```
flaviendelangle, line 3: We had two babel files with almost the same config.
I removed the one in the old `setup.js` file and used this one everywhere.

test/utils/setupXJSDOM.js
```diff
@@ -0,0 +1,16 @@
+require('./licenseRelease');
+const { createXMochaHooks } = require('./mochaHooks');
+
+const mochaHooks = createXMochaHooks();
+
+// The JSDOM implementation is too slow
+// https://github.com/jsdom/jsdom/issues/3234
+window.getComputedStyle = function getComputedStyleMock() {
```
flaviendelangle, line 8: Copied from the old `createDOM.js` file

test/utils/setupJSDOM.js
```diff
@@ -0,0 +1,17 @@
+require('@mui/monorepo/test/utils/setupJSDOM');
+require('./licenseRelease');
+const { createXMochaHooks } = require('./mochaHooks');
+
+const mochaHooks = createXMochaHooks();
+
+// The JSDOM implementation is too slow
+// https://github.com/jsdom/jsdom/issues/3234
+window.getComputedStyle = function getComputedStyleMock() {
```
flaviendelangle, line 9: Copied from the old `createDOM.js` file

cherniavskii (commented):
Looks great!

test/utils/setupKarma.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env mocha */
```
m4theushw, line 1: This file could be merged into `test/karma.test.js`, no? It only contains a dozen of lines. The core version is much bigger and makes sense the split.

flaviendelangle: It was to follow the core file structure but yes our version could have a single file

packages/grid/x-data-grid-pro/src/tests/printExport.DataGrid.test.tsx
```diff
@@ -54,6 +56,13 @@ describe('<DataGridPro /> - Print export', () => {
     },
   ];
 
+  before(function beforeHook() {
+    if (isJSDOM) {
+      // Can't access the document.body to print
+      this.skip();
```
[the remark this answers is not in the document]
m4theushw: If adding an option to pass the URL takes too much time I don't mind just duplicating `createDOM.js`.

package.json
oliviertassinari: I suspect that the main monorepo used to have `yarn test:watch` but it was replaced with `yarn t xxx`. `yarn t xxx` is superior to `yarn test:unit --watch` when you are working on a single test file, and the codebase has a lot of unit tests that all needs to be loaded before a single test can run, even with `only`.

packages/grid/x-data-grid-pro/src/tests/printExport.DataGrid.test.tsx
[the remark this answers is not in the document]
oliviertassinari: I have tried https://github.com/mui/mui-x/pull/6273 to solve the dependency on the jsdom configuration.

I have opened https://github.com/mui/material-ui/pull/34449 because I suspect that the logic was added to fix a symptom but not the root problem (see the linked PR inside it for what the root was).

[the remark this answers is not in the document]
oliviertassinari: > I don't mind just duplicating createDOM.js.

@m4theushw It's already duplicated. In https://github.com/mui/mui-x/pull/5714/files#diff-5915ab32081ceded46f473b03d334fbf63caf0633659b742eff53bab92ddf203R19, I think that we saw one of the costs we pay for it. Is it the best option? I don't know, but I personally struggle to envision fundamental reasons for having different JSDOM environments.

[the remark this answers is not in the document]
flaviendelangle: If we can have both a single `createDOM` AND the print tests working that's of course better :ok_hand:

test/utils/setupKarma.js
flaviendelangle: Done

packages/grid/x-data-grid-pro/src/tests/printExport.DataGrid.test.tsx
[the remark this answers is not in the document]
flaviendelangle: Seems to work :ok_hand:

cherniavskii (approved):

oliviertassinari:
Nice cleanup :)

PR #6048, opened 2022-09-06
[pickers] Support slots on the `DateField` component (WIP)

flaviendelangle:


alexfauquette:
More in favor of option 2, because I'm not a big fan of adding useless `div`. And it makes sense that the "multi input fields" has the same slot plus a wrapper (`Root`) and a `Separator`.

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
```diff
@@ -13,17 +35,100 @@ export const MultiInputDateRangeField = React.forwardRef(function MultiInputDate
...
+  const startInputProps = useSlotProps({
+    elementType: Input,
+    externalSlotProps: componentsProps?.input,
+    ownerState: { ...ownerState, position: 'start' },
+  });
+  const endInputProps = useSlotProps({
+    elementType: Input,
+    externalSlotProps: componentsProps?.input,
+    ownerState: { ...ownerState, position: 'end' },
+  });
```
alexfauquette, line 83: If I understand correctly, overriding `placeholder` could be done as follow:

```js
componentsProps: {
  input: ({ position }) => ({ placeholder: position==='start' ? 'Depature' : 'Arrival' })
}

File: packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx [hunk 2]
```diff
@@ -13,17 +35,100 @@ export const MultiInputDateRangeField = React.forwardRef(function MultiInputDate
...
-    </Stack>
+    <Root {...rootProps}>
+      <Input
+        {...startDateProps}
+        inputProps={{
+          // ...startDateProps.inputProps,
+          ref: startInputRef,
+          onKeyDown: onStartInputKeyDown,
+        }}
+      />
```
alexfauquette, line 122: I don't get why the `startDateProps` is commented and `startInputProps` does not appears
```suggestion
      <Input
        {...startDateProps}
        {...startInputProps}
        inputProps={{
          ...startDateProps.inputProps,
          ...startInputProps.inputProps,
          ref: startInputRef,
          onKeyDown: onStartInputKeyDown,
        }}
      />
```

flaviendelangle: I was not supporting correctly the props passed to each input
It should be better now.

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
flaviendelangle: The placeholder is probably not a good example since the field is never empty.
The label would probably still be provided with `{ localText: { start: 'Departure', end: 'Arrival' } }` but that could be discussed.

flaviendelangle:
We discussed it with @joserodolfofreitas, it feel like we all agree that option 2 (see PR description) is the best here

alexfauquette (approved):

PR #6049, opened 2022-09-06
[DataGrid] Fix blank space when changing page with dynamic row height

### LINKED ISSUE #6029: Data grid bug: with Dynamic row height and rowsPerPageOptions - lost rows
### Duplicates

### Latest version

### Current behavior 😯

I use demo from  documentation. If default  pageSize: 25,
 and  rowsPerPageOptions={[10, 25, 50, 100]}, then I click 10 and on second page I see that 4 rows are lost .

https://user-images.githubusercontent.com/56801823/188325061-bdbbfcaf-b423-4414-b2f4-80504cc4cbb9.mp4

### Expected behavior 🤔

expect not to lose rows and  see 10 rows on every page

### Steps to reproduce 🕹

Link to live example:
https://codesandbox.io/s/recursing-noether-zk2c6r?file=/demo.tsx:1775-1813

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #6029

### DIFF
--- a/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
+++ b/packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
@@ -118,17 +118,19 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
   const getNearestIndexToRender = React.useCallback(
     (offset: number) => {
       const lastMeasuredIndexRelativeToAllRows = apiRef.current.unstable_getLastMeasuredRowIndex();
-      const lastMeasuredIndexRelativeToCurrentPage =
-        lastMeasuredIndexRelativeToAllRows - (currentPage.range?.firstRowIndex || 0);
-      const lastMeasuredIndex = Math.max(0, lastMeasuredIndexRelativeToCurrentPage);
-
-      let allRowsMeasured = lastMeasuredIndex === Infinity;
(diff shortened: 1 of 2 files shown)


packages/grid/x-data-grid/src/hooks/features/virtualization/useGridVirtualScroller.tsx
```diff
@@ -118,17 +118,19 @@ export const useGridVirtualScroller = (props: UseGridVirtualScrollerProps) => {
...
       }
 
-      if (allRowsMeasured || rowsMeta.positions[lastMeasuredIndex] >= offset) {
+      const lastMeasuredIndexRelativeToCurrentPage = clamp(
+        lastMeasuredIndexRelativeToAllRows - (currentPage.range?.firstRowIndex || 0),
+        0,
+        rowsMeta.positions.length,
+      );
+
+      if (allRowsMeasured || rowsMeta.positions[lastMeasuredIndexRelativeToCurrentPage] >= offset) {
```
m4theushw, line 133: The bug in #6029 was caused because the index used to check with `rowsMeta.positions` was outside the bounds, so it would satisfy this condition and call `exponentialSearch` below. My fix was to adjust so the index used to check `rowsMeta.positions` is relative to the current page only.

DanailH (approved):
Do we need to port this one to `master`? Looks like something that is important and not that difficult to port. That logic will most likely stay the same in v6.

Issue #5565, opened 2022-07-21
[discussion] New doc structure for the pickers


flaviendelangle:
Related to #5539

The goal is to really start improving the pickers documentation in depth, both for the basic and the advanced use cases.

## Naming convention

I think this new doc structure should come with some components renaming in v6.

I tried to apply the new naming from https://github.com/mui/mui-x/issues/4484 for the `Static` components.
They are a good start, but I still struggle to clarify between: 

- a static picker like `StaticDatePicker`, which has a toolbar and a header to switch views
- a composite view like `CalendarPicker`, which has no toolbar but a header to switch views
- a granular view like `MonthPicker`, which has no toolbar and no header

I would like your opinion on this, for now I named all of them `XXXView` below.
I will try to do a full summary on #4484

## Proposal

- *Date and Time pickers*
  - *Introduction*
    - Overview (part of the current *Getting started*)
    - Installation (part of the current *Getting started*)
    - Migrating from the lab (current *Migration from the lab*)

  - *Pickers* (a field + a view inside a popper)
    - Date picker
    - Date range picker
    - Date time picker
    - Date time range picker
    - Time picker
    - Time range picker
  
  - *Fields* 
    - Date field (see #5531)
    - Date range field

  - *Standalone views*
    - Date picker view (current `StaticDatePicker`)
    - Date time picker view (current `StaticDateTimePicker`)
    - Time picker view (current `StaticTimePicker`)
    - Calendar picker view (current `CalendarPicker`)
    - Day picker view (current `DayPicker`)
    - Month picker view (current `MonthPicker`)
    - Year picker view (current `YearPicker`)

  - *Custom behaviors*
    - *Validation* (see #5539)
    - *Localization* (not sure where to put that one, maybe at the root instead of in a group of pages)

  - *Visual customization*
    - *Custom components* (will probably be splitted in the future) 

  - *Typescript*

  - *API Reference*
    - ...

alexfauquette:
For clarification, I assume ùitalic* titles are pages/subpages and normal text sections in the page

- *Date and Time pickers*
  - *Introduction*
    - Overview (part of the current *Getting started*)
    - Installation (part of the current *Getting started*)
    - Migrating from the lab (current *Migration from the lab*)
  
	> Maybe this page can be very small, just two demos. One for community, the other for pro
	> It could also contain section "Structure overview" to share some vocabulary about view, fields, static
  - *Pickers* (a field + a view inside a popper)
  	- Single picker 
[...]

PR #6064, opened 2022-09-07
[docs] New page for the pickers: Validation

### LINKED ISSUE #5539: [docs] Properly document how to disable some dates in the pickers components
## The problem

The current structure of the picker doc is very similar to what can be found on the core with one page per component.
But those components have a lot in common and this structure prevents us from doing a more advanced documentation because a lot of the time, the documented topic relates to several components.

This issue focuses on the date disabling but the same problem can probably be raised for other topics.

### Concrete example

Imagine I am a developer and I want to disable some dates on my `DateTimePicker` component.

If I do to the *Date time picker* documentation page, I have a *Date time validation* sections that feels related to my topic and only shows how to use `minTime` / `maxTime` / `minDate` / `maxDate`.
But this section says nothing about `minDate` / `maxDate` / `shouldDisableDate`.

I can then go to the *Date picker* documentation page, thinking that maybe it's a topic shared across components (and it's a problem that I have to think that much where the content lives).
And there, no section called *Date validation* or related. If I try to look for *disable* on the page, I see one example of `shouldDisableDate` inside the *Landscape orientation* section (why there ? no idea).

### Problem summary

1. The documentation is not complete (no documentation of `minDate` / `maxDate` / `disablePast` / `disableFuture` or maybe in some more obscure page)
2. The documentation is scattered across several pages and several sections

## The solution

All of the props are used across several components.
I tried to do a summary table of all our validation props below.

|                     | `DatePicker` | `DateRangePicker` | `DateTimeRangePicker` | `DateTimePicker` | `TimePicker` | `TimeRangePicker` |
|---------------------|--------------|-------------------|-----------------------|------------------|--------------|-------------------|
| `disablePast`       | ✅            | ✅                 | ✅                     | ✅                |              |                   |
| `disableFuture`     | ✅            | ✅                 | ✅                     | ✅                |              |                   |
| `shouldDisableDate` | ✅            | ✅                 | ✅                     | ✅                |              |                   |
| `minDate`           | ✅            | ✅                 | ✅                     | ✅                |              |                   |
| `maxDate`           | ✅            | ✅                 | ✅                     | ✅                |              |                   |
| `shouldDisableTime` |              |                   | ✅                     | ✅                | ✅            | ✅                 |
| `minTime`           |              |                   | ✅                     | ✅                | ✅            | ✅                 |
| `maxTime`           |              |                   | ✅                     | ✅                | ✅            | ✅                 |
| `minDateTime`       |              |                   | ✅                     | ✅                |              |                   |
| `maxDateTime`       |              |                   | ✅                     | ✅                |              |                   |

We can see several things

1. The `Date` and `DateRange` components seems to always have the same props (one exception would be if we introduce a `isRangeValid` prop).
2. We have 3 families of props, the one impacting the date, the one impacting the time and the one impacting the date and the time

I propose to create a stand-alone page containing all the doc related to those props and only keep one *Date validation* section on each component page linking to this page and if needed adding some precision only applicable to the current component.

cc @joserodolfofreitas @samuelsycamore @alexfauquette

flaviendelangle:
Closes #5539
Part of #5565

[Doc preview](https://deploy-preview-6064--material-ui-x.netlify.app/x/react-date-pickers/validation/)

______

- [x] Create a new page explaining the global behavior of the validation and how each prop works
- [x] Remove validation examples scattered across the doc
- [x] Add a back link section on every component page 

______

A few points I would like feedback on:

1. Do you like UI to display the same behavior on several components ?

image

I did not use the `label` prop since `DateRangePicker` does not have a top level label.
Once we have something clean, I would like to bring the same spacing / sizing to the other date range picker demos where we need a top level label

image

And we will have a lot of other demos with the same logic.
For me it's a lot smoother to just make an example with each component rather than adding a sentence like "this prop is supported by X, Y and Z" (when the example is simple of course).

2. I did one section per prop instead of grouping them. It causes some redundancy (`minTime` / `maxTime` for instance).
But I liked the idea of having a page very easy to understand. On each big section "Date validation" / "Time validation" / "Date and time validation" props as ordered from the less customizable to the most customizable (boolean => date => callback).

3. I would like to add a very light one line table on each section with the signature of the prop. That's something I would like us to discuss at some point, how to bring the typing information related to the current page to our readers ? If we just add links to the API page of each component at the bottom, it will not be easy to use (you have to pick the right component and find back your prop among dozens of others).

4. I added notes for quirks users might fall in. For instance the fact that `shouldDisableDate` will not disable fully disabled months. Do you see others ? I would like feedback on the warnings in `minDateTime` / `maxDateTime`, are they clear enough to explain what's the difference with `minTime` / `maxTime` ?


alexfauquette:
1. About the UI components, I like it. It allows to directly experiment the different components.

Just a problem with the clock picker in "Rendering of invalid values" because it's a 12h clock so the invalid hours do no appear. I assume it works the morning when it's on AM by default

2. About having one section per prop, it may be a bit too much. When I see too many examples that are simples, I tend to say "okay, this page is about trivial stuff I can skip". Maybe the props could be grouped 2 by 2, such that each section bring a new information.

About demo it could be interesting to have somewhere displayed wat it the min/max date. Otherwise we have to open the code to know what's the value of `minDate`

3. Why not, might be a bit tricky to do the matching. Let me think that it could be interesting to have anchor in the table such that we could link to  specific line in the API pages

4. Time examples seems clear. Is their a reason why using `dayjs().set('hour', 18).startOf('hour')` instead of `dayjs('2022-01-01T18:00')`?

flaviendelangle:
> Just a problem with the clock picker in "Rendering of invalid values" because it's a 12h clock so the invalid hours do no appear. I assume it works the morning when it's on AM by default

I think I just have to set a date instead of `null` to force to the afternoon.

> 4. Time examples seems clear. Is their a reason why using dayjs().set('hour', 18).startOf('hour') instead of dayjs('2022-01-01T18:00')?

In the doc, I say that the simplest approach is to take today's date an totally ignore the date.
So in the example I uses today.
But we could say "pick an arbitrary date" and use an example like yours.
I just want to be coherent between the two.

flaviendelangle:
> I tend to say "okay, this page is about trivial stuff I can skip"

Do you think most of our doc users really read our doc linearly and skip some pages ?
For me the main use case was more "I need to disable my component for all non-working day, it's not obvious with me IDE autocompletion, let's look at the doc".
I never read a doc upfront.

And in this use case, I like docs with small sections because it makes it very easy to just go to the one I need.

alexfauquette:
> Grouping together minTime / maxTime, minDate / maxDate, minDateTime / maxDateTime and disableFuture / disablePast would be fine to me because I guess the sections remain small enough.
But I would personnally only do it for very highly related props (I would not group shouldDisableMonth and shouldDisableDate / shouldDisableYear for instance because they have different warnings).

We agree on that point

alexfauquette:
About missing elements, it could be good to remind how to display a custom error message

flaviendelangle:
@alexfauquette a few grouping proposal

1. One section with each prop in its own demo. The warning / info notes are shared at the bottom of the section

image

2. One section with each prop in its own sub-section and own demo. The warning / info notes are shared at the bottom of the section

image


3. One section with a single demo

I really don't like this one because if we want to add a static picker on some sections (when it helps understanding)
Then the demo will become very hard to read.

And we can't do 2 columns on the demo because of the range pickers who take to much space.

image

flaviendelangle:
> About missing elements, it could be good to remind how to display a custom error message

Done

alexfauquette:
> And we can't do 2 columns on the demo because of the range pickers who take too much space.

I would be in favor of two columns even if it implies not putting the date range picker. But I'm curious about the opinion of others on this topic

flaviendelangle:
How would you demonstrate the behavior on the date range picker / fields then ?

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,188 @@
+---
+title: Date and Time pickers - Validation
+---
+
+# Date and Time pickers - Validation
+
+<p class="description">Add validation props to customize the values users can select.</p>
+
```
joserodolfofreitas, line 8: Should we add an mini introduction?

```suggestion
All the date and time pickers have an API for adding validation constraints, and by default, provide visual feedback if the component value doesn't meet the validation criteria.

```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,188 @@
...
+:::info
+On this page, the validation props are showcased on each type of picker component (`TimePicker`, `DatePicker`, `DateRangePicker`, ...).
+But they are also available on the other component variants.
```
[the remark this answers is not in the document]
flaviendelangle: Yes, I don't see any scenario where a variant would not have some validation prop

flaviendelangle:
> For range picker case - if we need two inputs for the demo, then sure, put them in different rows, but if we could have a single input - wouldn't it fit two date range picker examples?

Not sure to understand this one
The date range picker has two inputs.

alexfauquette:
To give some context before the retro we discussed the extreme case of [TextField](https://mui.com/material-ui/react-text-field/#form-props) demonstrations

image

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
+---
+title: Date and Time pickers - Validation
+---
+
+# Date and Time pickers - Validation
+
+<p class="description">Add custom validation to user inputs.</p>
+
+All the date and time pickers have an API for adding validation constraints,
+and by default, provide visual feedback if the component value doesn't meet the validation criteria.
```
LukasTy, line 10:
```suggestion
All the date and time pickers have an API for adding validation constraints.
By default they provide visual feedback if the component value doesn't meet the validation criteria.
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
...
+
+:::info
+The validation props are showcased for each type of picker component (`TimePicker`, `DatePicker`, `DateRangePicker`, etc).
+But the same props are available on all component variants.
+
+For example, the validation props available on the `DatePicker` component are also available on:
```
LukasTy, line 16:
```suggestion
For example—the validation props available on the `DatePicker` component are also available on:
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
...
+
+- `DesktopDatePicker`
+- `MobileDatePicker`
+- `StaticDatePicker`
+
+:::
+
+## Invalid values feedback
+
+On the input, it enables its error state.
```
LukasTy, line 26:
```suggestion
On the input—it enables its error state.
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
...
+
+{{"demo": "ValidationBehaviorInput.js", "defaultCodeOpen": false}}
+
+On the calendar and clock views, the invalid values are displayed as disabled to prevent their selection.
```
LukasTy, line 30:
```suggestion
On the calendar and clock views—the invalid values are displayed as disabled to prevent their selection.
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Date validation
+
+All the props described below are available on all the components supporting date edition.
+
+### Disable past or future dates
+
+The `disablePast` prop prevents the selection of all values before today.
+
+- On the `day` view, all the days before today won't be selectable.
```
LukasTy, line 42:
```suggestion
- On the `day` view—all the days before today won't be selectable.
```

flaviendelangle:
> To give some context before the retro we discussed the extreme case of [TextField](https://mui.com/material-ui/react-text-field/#form-props) demonstrations
> 
> image

I personally find those issues with tons of examples very messy and hard to use.

docs/data/date-pickers/validation/DateValidationDisablePast.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [dateTimePickerValue, setDateTimePickerValue] = React.useState<Dayjs | null>(
+    yesterday,
+  );
+  const [dateRangePickerValue, setDateRangePickerValue] = React.useState<
+    DateRange<Dayjs>
+  >([yesterday, today]);
+
+  return (
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <Stack spacing={4}>
```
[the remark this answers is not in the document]
flaviendelangle: image

Here is the result with this diff
I'll push it for the `DateValidationDisablePast` demo.

We clearly gain a lot of space
But on the other hand I find it harder to distinguish the various pickers (see that one line 1 it's 2 pickers with one input each but on line 2 it's 1 picker with two inputs).

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
+---
```
LukasTy, line 1: Here is a diff for the remaining punctuation suggestions 🙈 
```diff
diff --git a/docs/data/date-pickers/validation/validation.md b/docs/data/date-pickers/validation/validation.md
index dad257c9..06a72ae0 100644
--- a/docs/data/date-pickers/validation/validation.md
+++ b/docs/data/date-pickers/validation/validation.md
@@ -13,7 +13,7 @@ By default they provide visual feedback if the component value doesn't meet the
 The validation props are showcased for each type of picker component (`TimePicker`, `DatePicker`, `DateRangePicker`, etc).
 But the same props are available on all component variants.
 
-For example, the validation props available on the `DatePicker` component are also available on:
+For example—the validation props available on the `DatePicker` component are also available on:
 
 - `DesktopDatePicker`
 - `MobileDatePicker`
@@ -71,7 +71,7 @@ The `maxDate` prop prevents the selection of all values after `props.maxDate`.
 
 The `shouldDisableDate` prop prevents the selection of all dates for which it returns `true`.
 
-In the example below, the weekends are not selectable:
+In the example below—the weekends are not selectable:
 
 {{"demo": "DateValidationShouldDisableDate.js", "defaultCodeOpen": false}}
 
@@ -79,15 +79,15 @@ In the example below, the weekends are not selectable:
 `shouldDisableDate` only prevents the selection of disabled dates on the `day` view.
 For performance reasons—when rendering the `month` view—we are not calling the callback for every day of each month to see which one should be disabled (same for the `year` view).
 
-If you know that all days of some months are disabled, you can provide the [`shouldDisableMonth`](#disable-specific-months) prop to disable them in the `month` view.
+If you know that all days of some months are disabled—you can provide the [`shouldDisableMonth`](#disable-specific-months) prop to disable them in the `month` view.
 Same with the [`shouldDisableYear`](#disable-specific-years) prop for the `year` view.
 :::
 
 #### Disable specific dates in range components [<span class="pro-premium"></span>](/x/introduction/licensing/#pro-plan)
 
-For components supporting date range edition (`DateRangePicker`, `DateTimeRangePicker` 🚧), the `shouldDisableDate` prop receives a second argument to differentiate the start and the 
end date.
+For components supporting date range edition (`DateRangePicker`, `DateTimeRangePicker` 🚧)—the `shouldDisableDate` prop receives a second argument to differentiate the start and the e
nd date.
 
-In the example below, the start date can't be in the weekend but the end date can.
+In the example below—the start date can't be in the weekend but the end date can.
 
 {{"demo": "DateRangeValidationShouldDisableDate.js", "defaultCodeOpen": false}}
 
@@ -101,7 +101,7 @@ The `shouldDisableMonth` prop prevents the selection of all dates in months for
 `shouldDisableMonth` only prevents the selection of disabled months on the `day` and `month` views.
 For performance reasons—when rendering the `year` view—we are not calling the callback for every month of each year to see which one should be disabled.
 
-If you know that all months of some years are disabled, you can provide the [`shouldDisableYear`](#disable-specific-years) prop to disable them in the `year` view.
+If you know that all months of some years are disabled—you can provide the [`shouldDisableYear`](#disable-specific-years) prop to disable them in the `year` view.
 :::
 
 ### Disable specific years
@@ -146,7 +146,7 @@ shouldDisableTime={(timeValue, view) => view === 'minutes' && timeValue >= 45};
 shouldDisableTime={(timeValue, view) => view === 'seconds' && timeValue >= 30};
 ```
 
-In the example below, the last quarter of each hour is not selectable.
+In the example below—the last quarter of each hour is not selectable.
 
 {{"demo": "TimeValidationShouldDisableTime.js", "defaultCodeOpen": false}}
 
@@ -163,7 +163,7 @@ The `maxDateTime` prop prevents the selection of all values after `props.maxDate
 {{"demo": "DateTimeValidationMaxDateTime.js", "defaultCodeOpen": false}}
 
 :::warning
-If you want to put time boundaries independent of the date, use the [`time boundaries`](#set-time-boundaries) instead.
+If you want to put time boundaries independent of the date—use the [`time boundaries`](#set-time-boundaries) instead.
 
 ```tsx
 // Disable the values between 6 PM and midnight for every day (tomorrow 5 PM is not disabled).
@@ -183,9 +183,9 @@ If you want to put time boundaries independent of the date, use the [`time bound
 
 ## Show the error
 
-To render the current error, you can subscribe to the `onError` callback which is called every time the error changes.
+To render the current error—you can subscribe to the `onError` callback which is called every time the error changes.
 You can then use the `helperText` prop of the `TextField` to pass your error message to your input as shown below.
 
-Try to type a date that is inside the first quarter of 2022, the error will go away.
+Try to type a date that is inside the first quarter of 2022—the error will go away.
 
 {{"demo": "RenderErrorUnderField.js"}}
```

docs/data/date-pickers/validation/DateValidationDisablePast.tsx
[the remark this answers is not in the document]
LukasTy: I'm in favour of the `disablePast` variant and am for replacing layout to such for all the demos containing 2 or 3 inputs. 🤔 
What do you think @joserodolfofreitas @alexfauquette @samuelsycamore ?

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,191 @@
...
+## Date validation
+
+All the props described below are available on all the components supporting date edition.
+
+### Disable past or future dates
+
+The `disablePast` prop prevents the selection of all values before today.
+
+- On the `day` view—all the days before today won't be selectable.
+- On the `month` and `year` views—all the values ending before today won't be selectable.
```
alexfauquette, line 43: PR #6226 is adding the support of time for `disablePast` and `disableFuture`. So the behavior is becoming more generic.

If accepted, this part should be a section (instead of a subsection)

alexfauquette (approved):

docs/data/date-pickers/validation/DateValidationShouldDisableMonth.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [dateTimePickerValue, setDateTimePickerValue] = React.useState<Dayjs | null>(
+    today,
+  );
+
+  return (
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <Stack spacing={4}>
+        <GridItem label="DatePicker">
+          <DatePicker
+            shouldDisableDate={isInCurrentMonth}
```
alexfauquette, line 46:
```suggestion
            shouldDisableMonth={isInCurrentMonth}
```

Same on Year demo. It seems that `DateTimePicker` does not support the `shouldDisableYear`

flaviendelangle: Indeed, I passed the wrong prop :grimacing: 

> It seems that DateTimePicker does not support the shouldDisableYear

What makes you think that
I just tried it and for me it works

docs/data/date-pickers/validation/validation.md
flaviendelangle: Indeed, I will move the section once it is merged

docs/data/date-pickers/validation/DateValidationShouldDisableMonth.tsx
alexfauquette: The days are disabled, but the TextField is not in error mode

image

flaviendelangle: Oooohhhh

flaviendelangle: https://github.com/mui/mui-x/blob/44eb781b15bb3c2c04c4413830f42fc7df27b7f9/packages/x-date-pickers/src/internals/hooks/validation/useDateTimeValidation.ts#L32

Do you see a reason not to pass every prop to both date and time validation ?
That's the cause of the bug.

By the way, even if this page was never published, we found lots of bugs writing and reviewing it :laughing:

alexfauquette: Sounds good to pass everything and let hooks peak what they need.

flaviendelangle: > On the same topic, I was surprised that DatePickers do not support maxDateTime/minDateTime

Right now `minDateTime` / `maxDateTime` management is quite ugly, it's just replaced in `DateTimePicker` which mean even it can't use both `minDateTime` and `minDate` at the same time for instance.

For `DatePicker` what would be the behavior for you ?
Would it validate the time when passed `maxDateTime` / `minDateTime` or just the date ?

docs/data/date-pickers/validation/DateValidationDisablePast.tsx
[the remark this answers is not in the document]
samuelsycamore: I like the look of the `disableFuture` examples better. I would much prefer to see each example on its own line. Makes it easier to quickly skim and understand what you're looking at.

It's also important to consider the difference in UX between light and dark mode in the docs - what looks like decent spacing in light mode can appear too cluttered in dark mode, so I think it's important to err on the side of too much space rather than not enough.

docs/data/date-pickers/validation/DateValidationShouldDisableMonth.tsx
alexfauquette: Intuitively I would have expected it to validate the date. But it does not make a lot of sense since it would have the same behavior as `maxDate`

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,193 @@
...
+## Date validation
+
+All the props described below are available on all the components supporting date edition.
+
+### Set date boundaries
+
+The `minDate` prop prevents the selection of all values before `props.minDate`.
+
+- On the `day` view—all the days before the `minDate` won't be selectable.
+- On the `month` and `year` views—all the values ending before the `minDate` won't be selectable.
```
LukasTy, line 61: A few points:
- we are talking about `year, month, day` views, but by default date pickers don't include `month` view and that won't be seen in the demo, but is mentioned. I'd say we should add explicit views for these demos, like in this case `views={['year', 'month', 'day']}`
- demos include a `DateTimePicker`, but text does not mention `hour, minutes, seconds` views. Should we?

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,193 @@
...
+
+On the calendar and clock views—the invalid values are displayed as disabled to prevent their selection.
+
+{{"demo": "ValidationBehaviorView.js", "defaultCodeOpen": false}}
+
+## Past and future validation
+
+All pickers support the past and future validation.
+
+The `disablePast` prop prevents the selection of all values before today.
```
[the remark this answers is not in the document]
flaviendelangle: Maybe we can detail more ?

```suggestion
The `disablePast` prop prevents the selection of all values before the current date and time.
For date pickers it will prevent selections of all values before today.
To time pickers, it will prevent selections of all values before the current time.
For date time pickers, it will combine both.
```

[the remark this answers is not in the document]
LukasTy: That is a proper wall of text, but I reckon it's better to be as precise as we can—sounds good 🤔 👍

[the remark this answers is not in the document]
flaviendelangle: A shorter version:

```suggestion
The `disablePast` prop prevents the selection of all values before today for date pickers and the selection of all values before the current time for time pickers.
For date time pickers, it will combine both.
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,193 @@
...
+The `minDateTime` prop prevents the selection of all values after `props.minDateTime`.
+
+{{"demo": "DateTimeValidationMinDateTime.js", "defaultCodeOpen": false}}
+
+The `maxDateTime` prop prevents the selection of all values after `props.maxDateTime`.
+
+{{"demo": "DateTimeValidationMaxDateTime.js", "defaultCodeOpen": false}}
+
+:::warning
+If you want to put time boundaries independent of the date—use the [`time boundaries`](#set-time-boundaries) instead.
```
alexfauquette, line 168: I don't know if we should warn about this edge case

```suggestion
If you want to put time boundaries independent of the date—use the [`time boundaries`](#set-time-boundaries) instead.

For now, you can not use `maxDateTime` and `maxTime` together.
`maxDateTime` will override the `maxTime` behavior—Same for `minDateTime`/`minTime`
```

alexfauquette (approved):
Sounds super clear 👌

docs/data/date-pickers/validation/validation.md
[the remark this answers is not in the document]
joserodolfofreitas: What do you think about using "now"?

```suggestion
The `disablePast` prop prevents the selection of all values before now. That means before today on date pickers, or before the current time, on time pickers.
```

docs/data/date-pickers/validation/validation.md
flaviendelangle: I added the views and some details for the time views

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,206 @@
...
+shouldDisableTime={(timeValue, view) => view === 'seconds' && timeValue >= 30};
+```
+
+In the example below—the last quarter of each hour is not selectable.
+
+{{"demo": "TimeValidationShouldDisableTime.js", "defaultCodeOpen": false}}
+
+## Date and time validation
+
+### Set date and time boundaries
```
joserodolfofreitas, line 167:
```suggestion
### Boundaries - min and MaxDateTime
```

joserodolfofreitas (approved):
I like a lot how it's grouped.
I think grouping per component makes a lot of sense when they each have variants of an API.

Users are usually not trying to understand all the components at the same time, so it makes sense that they can directly refer to the props for the component in question.

When we have more "cross-pickers-props" like the case for **past and future**, I'd be even in favor of having some redundancy and considering rendering those inside the component section too.
For now, I think it works fine in its own section.

On another note, I'm not entirely sure about the term "boundaries".
I'm thinking there might be other more common terms users might use on searches like "limits", or even simply min and max date.
https://trends.google.com/trends/explore?q=date%20boundaries,date%20limits,date%20constraints,date%20validation

docs/data/date-pickers/validation/validation.md
LukasTy: I was thinking about something along these lines: add every possible view to each of the component.
What do you think? 🤔 
Also, in the text we mention `seconds` view, but it is nowhere to be seen, maybe including it in the views would make more sense here? 🤔 
But again, if we add `seconds` view it might make sense to have an option to reset 'now'.
```diff
diff --git a/docs/data/date-pickers/validation/DateValidationDisablePast.tsx b/docs/data/date-pickers/validation/DateValidationDisablePast.tsx
index cc0db38c6..b373671b6 100644
--- a/docs/data/date-pickers/validation/DateValidationDisablePast.tsx
+++ b/docs/data/date-pickers/validation/DateValidationDisablePast.tsx
@@ -10,6 +10,7 @@ import { DatePicker } from '@mui/x-date-pickers/DatePicker';
 import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
 import { DateRange, DateRangePicker } from '@mui/x-date-pickers-pro/DateRangePicker';
 import { TimePicker } from '@mui/x-date-pickers/TimePicker';
+import Button from '@mui/material/Button';
 
 const today = dayjs();
 const yesterday = dayjs().subtract(1, 'day');
@@ -47,10 +48,12 @@ export default function DateValidationDisablePast() {
   const [dateRangePickerValue, setDateRangePickerValue] = React.useState<
     DateRange<Dayjs>
   >([yesterday, today]);
+  const [key, setKey] = React.useState(dayjs().millisecond());
 
   return (
     <LocalizationProvider dateAdapter={AdapterDayjs}>
-      <Stack spacing={4}>
+      <Stack spacing={4} key={key}>
+        <Button onClick={() => setKey(dayjs().millisecond())}>Re-render</Button>
         <GridItem label="DatePicker">
           <DatePicker
             disablePast
@@ -66,6 +69,7 @@ export default function DateValidationDisablePast() {
             value={dateTimePickerValue}
             onChange={(newValue) => setDateTimePickerValue(newValue)}
             renderInput={(params) => <TextField {...params} />}
+            views={['year', 'month', 'day', 'hours', 'minutes', 'seconds']}
           />
         </GridItem>
         <GridItem label="TimePicker">
@@ -74,6 +78,7 @@ export default function DateValidationDisablePast() {
             value={timePickerValue}
             onChange={(newValue) => setTimePickerValue(newValue)}
             renderInput={(params) => <TextField {...params} />}
+            views={['hours', 'minutes', 'seconds']}
           />
         </GridItem>
         <GridItem label="DateRangePicker" spacing={2}>
```

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,206 @@
...
+
+:::info
+The current time is computed during the first render of the `LocalizationProvider`, it will not change during the lifetime of the component.
+:::
+
+## Date validation
+
+All the props described below are available on all the components supporting date edition.
+
+### Set date boundaries
```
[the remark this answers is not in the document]
flaviendelangle: Probably more like that no ?

```suggestion
### Date boundaries - min and max date
```

Or depending on your other comment

```suggestion
### Date limits - min and max date
```

flaviendelangle:
> On another note, I'm not entirely sure about the term "boundaries".
I'm thinking there might be other more common terms users might use on searches like "limits", or even simply min and max date.
https://trends.google.com/trends/explore?q=date%20boundaries,date%20limits,date%20constraints,date%20validation

No strong opinion, `limit` is fine for me

docs/data/date-pickers/validation/validation.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Disable specific years
+
+The `shouldDisableYear` prop prevents the selection of all dates in years for which it returns `true`.
+
+{{"demo": "DateValidationShouldDisableYear.js", "defaultCodeOpen": false}}
+
+## Time validation
+
+### Set time boundaries
```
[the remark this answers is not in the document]
joserodolfofreitas:
```suggestion
### Minimum and maximum time
```

docs/data/date-pickers/validation/validation.md
flaviendelangle: > I was thinking about something along these lines: add every possible view to each of the component.
What do you think? 

OK, I was lazy but you are right

flaviendelangle: For the seconds, how about we just don't talk about the seconds since it's a very weird use case ?
With the `reset` button I feel it impacts a lot the whole demo for an edge case.

Or maybe add a small demo inside my `info` block with only a `TimePicker` with only the `seconds` view and we log the current second in a div about the picker.

flaviendelangle: I added the `month` view to the maxDate / minDate / shouldDisableDate demos as well

LukasTy: > For the seconds, how about we just don't talk about the seconds since it's a very weird use case ?
With the reset button I feel it impacts a lot the whole demo for an edge case.
Or maybe add a small demo inside my info block with only a TimePicker with only the seconds view and we log the current second in a div about the picker.

Yeah, sure, that was just a suggestion, but I do agree, that these are possibly quite a niche edge-cases that might not deserve this much attention. 👌

LukasTy (approved):

PR #6073, opened 2022-09-07
[README] Add link to the security page

oliviertassinari:
It reproduces https://github.com/mui/material-ui#security. My assumption is that there isn't any real reason for having a different tradeoff on how we structure the main Readme, so using the same strategy would be simpler.

I noticed the opportunity while working on https://github.com/mui/material-ui/pull/34219.


oliviertassinari:
Since I did a plain copy & paste of the README template that is on http://github.com/mui/material-ui, I'm merging without reviews.

PR #6075, opened 2022-09-08
[core] Remove the `test_browsers` CI step

DanailH:
The fix of removing `Firefox` didn't work. I reverted my initial change and opted to comment on the entire step based on our initial discussion. I'm not sure if this is the correct way to do it because that way it won't be run on the PRs as well.


oliviertassinari (changes requested):
This is removing the ChromeHeadless steps from the CI.

Also, could we run other platfroms inside CircleCI rather than BrowserStack?

PR #6082, opened 2022-09-08
[DatePicker] Allow keyboard navigation to ignore disabled date for left/right arrow

### LINKED ISSUE #6076: Cannot navigate through or skip the disabled date in Calendar view on DatePicker
### Duplicates

### Latest version

### Current behavior 😯

We can pass the `shouldDisableDate` prop to disable specific date. which will result showing the calendar with some dates disabled. - that is expected behaviour
The problem happens when we want to navigate through the dates using keyboard only. Tab moves us between toolbar, dates grid and action bar. up/right/bottom/left buttons navigates us through the dates. EXCEPT when date is disabled, it does not go through it. if `8th` is disabled, and i want to navigate from `6 -> 7 -> 8 -> 9`. it will stop on `7th`, instead of skipping it and go directly to the `9th`. 
Disaster happens when only `today` is active, then it is selected by default and you cannot navigate anywhere and got kinda stuck

### Expected behavior 🤔

When focus to the Calendar bar with dates, it should focus to the first enabled date, and every further navigation should jump between only enabled dates. 
if `8th` and `11th` are disabled then it should go `6 -> 7 -> 9 -> 10 -> 12 -> 13`

### Steps to reproduce 🕹

Steps:

1. Call DatePicker with some disabled dates using `shouldDisableDate` prop
2. Click the TextField of Datepicker 
3. Press Tab to select the `InputAdornment` open calendar. Press `enter`
4. Press Tab to navigate between the days. 
5. When you reach the disabled date, it will stop and further navigation is impossible

### Context 🔦

We have been told that MUI components fully support a11y, while this example is not acceeible using keybpard only

### Your environment 🌎

<details>

```
```
</details>

alexfauquette:
Fix #6076

If the next date to focus is disabled, it looks in the current month, and the next one if there is an available date to focus.

### DIFF
--- a/packages/x-date-pickers/src/CalendarPicker/CalendarPickerKeyboard.test.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/CalendarPickerKeyboard.test.tsx
@@ -97,5 +97,46 @@ describe('<CalendarPicker /> keyboard interactions', () => {
         expect(document.activeElement).toHaveAccessibleName(expectFocusedDay);
       });
     });
+
+    describe('navigation with disabled dates', () => {
+      const disabledDates = [
+        adapterToUse.date(new Date(2020, 0, 10)),
+        // month extremities
+        adapterToUse.date(new Date(2019, 11, 31)),
+        adapterToUse.date(new Date(2020, 0, 1)),
+        adapterToUse.date(new Date(2020, 0, 2)),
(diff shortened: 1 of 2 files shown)


LukasTy (approved):
LGTM 🙌 🚀 
The only issue that remains is how to handle a case, where the whole current month is disabled, but I'd agree that it is quite a big edge-case.

PR #6089, opened 2022-09-08
[YearPicker] Scroll to the current year even with `autoFocus=false`

### LINKED ISSUE #4601: [StaticDatePicker] The year view on `StaticDatePicker` does not auto scroll to current year
### Duplicates

### Latest version

### Current behavior 😯

When opening the year view of a Standalone Date Picker, the currently selected year is not revealed. I have to scroll down to see the currently selected year. This is inconsistent with the Date Picker component where when opening the year view, the current year is automatically scrolled to.

Date Picker:
image

Standalone Date Picker:
image

### Expected behavior 🤔

The currently selected year should be revealed in the StandaloneDatePicker.

### Steps to reproduce 🕹

Steps:

1. Open the StandaloneDatePicker demo from the docs: https://codesandbox.io/s/staticdatepickerdemo-material-demo-forked-805v23?file=/package.json
2. Observe that the current year is not automatically revealed/scrolled into view
3. Compare to the DatePicker demo from the docs where when switching to the year view, the current year is revealed: https://codesandbox.io/s/basicdatepicker-material-demo-forked-kt9zjh?file=/demo.tsx

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
   System:
    OS: Windows 10 10.0.17134
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 17.07 GB / 31.86 GB
  Binaries:
    Node: 14.19.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
  Managers:
    Maven: 3.6.0 - C:\Program Files\Apache\maven\bin\mvn.CMD
    pip3: 20.1.1 - ~\AppData\Local\Programs\Python\Python38-32\Scripts\pip3.EXE
  Utilities:
    Git: 2.32.0.
  Virtualization:
    Docker: 19.03.8 - C:\Program Files\Docker\Docker\resources\bin\docker.EXE
  IDEs:
    VSCode: 1.66.1 - C:\Users\jspear\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 4.4.19 - C:\Windows\system32\bash.EXE
    Java: 17.0.2
    Python: 3.8.5
  Browsers:
    Chrome: 100.0.4896.127
    Edge: Spartan (42.17134.2208.0)
    Internet Explorer: 11.0.17134.1

```
</details>

### Order ID 💳 (optional)

#27566

alexfauquette:
Fix #4601


flaviendelangle (commented):
Is there a way to achieve the same behavior without adding a new div ?

And what do you think about hiding the content of the div before the scroll is completed ?
Because in it's current shape, we have an ugly flickering.

flaviendelangle:
> I do not see the flickering. (light mode the PR deploy, dark mode the current doc)

Use `openTo="year"` to see it

alexfauquette:
For the blink, it's not a problem with the ref that is responsible. The autofocus has the same behavior

https://user-images.githubusercontent.com/45398769/189872754-cca7677b-0367-4b42-b6a5-6accf9b7071d.mp4

You would like to add a

--- 

About avoiding to add a new div, I realized I ended up using `display: contents` which is breaking the accessibility.

We have the following structure:

```jsx
<CalendarRoot> // a display flex column
	<CalendarHeader /> // fix height
	<TransitionGroup> // extends as much as possible (height: 100%)
		<div>
			<RenderedView/>
```

The problem is that with `height: 100%` the div height matches it's content, not its parent. And so I cannot put the `overflow` to this element.

I can not put the `ref` on the `TransitionGroup` (exact name is `CalendarPickerViewTransitionContainer`) because if `reduceAnimations=true` it simply render its children

The current solution I implemented is to manage the scroll in the `YearPicker` because it's the only component with a scroll bar.

flaviendelangle:
> For the blink, it's not a problem with the ref that is responsible. The autofocus has the same behavior

OK, it's not very sexy, but if the `autoFocus` has it I guess we can release it like that

> The current solution I implemented is to manage the scroll in the YearPicker because it's the only component with a scroll bar.

Makes sense to me

flaviendelangle (approved):

PR #6090, opened 2022-09-08
[Datagrid] Fix error when focused cell removed and lost focus

### LINKED ISSUE #5983: [data grid] Editing a row after deleting another causes uncaught error
### Duplicates

### Latest version

### Current behavior 😯

When one tries to edit a row after deleting a different row, there is an uncaught error where the deleted row is not found.
This will only cause a console error and the functionality works, however, when trying the same and focusing a field programmatically the site crashes instead.

Maybe there is a stale state in the rowModesModel?

### Expected behavior 🤔

One should be able to edit rows after deleting some in the grid

### Steps to reproduce 🕹

Link to live example: https://codesandbox.io/s/uqtpzu?file=/demo.tsx

Steps:
Version 1: Only console error
1. Delete a row
2. Try to edit a cell of another row

Version 2: Site crashes
1. Change line 112 to ->   setRowModesModel({ ...rowModesModel, [id]: { mode: GridRowModes.Edit, fieldToFocus: 'age' } });
2. Delete a row
3. Try to edit a row

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

chaiwattsw:
Fixes #5983

### DIFF
--- a/packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
+++ b/packages/grid/x-data-grid/src/hooks/features/focus/useGridFocus.ts
@@ -56,7 +56,7 @@ export const useGridFocus = (
         return;
       }
 
-      if (focusedCell) {
+      if (focusedCell && apiRef.current.getRow(focusedCell.id)) {
         // There's a focused cell but another cell was clicked
         // Publishes an event to notify that the focus was lost
         apiRef.current.publishEvent(


m4theushw:
Thanks for the contribution but https://github.com/mui/mui-x/pull/6009 is fixing the same thing. If you want to contribute you can check the other issues labeled with "good first issue".

PR #6101, opened 2022-09-09
[docs] Fix broken links

alexfauquette:
Some anchors have been modified. Found during #6102

PR #6102, opened 2022-09-09
[core] Parse markdown to detect broken links

alexfauquette:
Sometimes, we change header names or the entire structure of a doc page. But we do not realize it's breaking all the external anchor links.

I propose a script that checks all the `.md` files and verifies that internal links are ok. Used to create #6101

It’s not perfect because it has:

1. false positives, the #slot because API page structure are made with `json` and not `md`
2. does not support links across projects
3. only about internal links

Linked to: https://www.notion.so/mui-org/Detect-301-404-in-the-CI-a8a2404d62a24fccb0f10b8318150ca3

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "prettier": "node ./scripts/prettier.js --branch master",
     "prettier:all": "node ./scripts/prettier.js write",
     "proptypes": "cross-env BABEL_ENV=development babel-node -i \"/node_modules/(?!@mui)/\" -x .ts,.tsx,.js ./docs/scripts/generateProptypes.ts",
+    "check-links": "node ./docs/scripts/checkLinks.js",
     "size:snapshot": "node --max-old-space-size=2048 ./scripts/sizeSnapshot/create",
     "size:why": "yarn size:snapshot --analyze --accurateBundles",
     "test": "lerna run test --parallel",
--- a/docs/package.json
+++ b/docs/package.json
@@ -8,6 +8,7 @@
     "build": "cross-env NODE_ENV=production next build --profile",
     "build:clean": "rimraf .next && yarn build",
     "build-sw": "node ./scripts/buildServiceWorker.js",
+    "check-links": "node ./scripts/checkLinks.js",
     "dev": "rimraf ./node_modules/.cache/babel-loader && next dev --port 3001",
     "deploy": "git push upstream master:docs-v5",
     "export": "rimraf docs/export && next export --threads=3 -o export && yarn build-sw",
--- a/docs/scripts/checkLinks.js
+++ b/docs/scripts/checkLinks.js
@@ -1,0 +1,146 @@
+/* eslint-disable no-console */
+const path = require('path');
+const fse = require('fs-extra');
+const { createRender } = require('@mui/monorepo/docs/packages/markdown');
+const { marked } = require('marked');
+
+// Use renderer to extract all links into a markdown document
+const getPageLinks = (markdown) => {
+  const hrefs = [];
+
+  const renderer = new marked.Renderer();
+  renderer.link = (href) => {
+    if (href[0] === '/') {
+      hrefs.push(href);
+    }
+  };
+  marked(markdown, { renderer });
+  return hrefs;
+};
+
+// List all .js files in a folder
+const getJsFilesInFolder = (folderPath) => {
+  const files = fse.readdirSync(folderPath, { withFileTypes: true });
+  return files.reduce((acc, file) => {
+    if (file.isDirectory()) {
+      const filesInFolder = getJsFilesInFolder(path.join(folderPath, file.name));
+      return [...acc, ...filesInFolder];
+    }
+    if (file.name.endsWith('.js')) {
+      return [...acc, path.join(folderPath, file.name)];
+    }
+    return acc;
+  }, []);
+};
+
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "prettier": "node ./scripts/prettier.js --branch master",
     "prettier:all": "node ./scripts/prettier.js write",
     "proptypes": "cross-env BABEL_ENV=development babel-node -i \"/node_modules/(?!@mui)/\" -x .ts,.tsx,.js ./docs/scripts/generateProptypes.ts",
+    "check-links": "node ./docs/scripts/checkLinks.js",
```
LukasTy, line 29:
```suggestion
    "check-links": "yarn workspace docs check-links",
```

oliviertassinari (commented):
I would be curious to see if this works on http://github.com/mui/material-ui.

I think that it would be great to add this to the CI and have it exit 1 when there is an issue. Without it, I don't think that the script will really bring value (it will get forgotten about) hence, I think that it would be better to delay merging until this point.

Regarding the limitations of this script, covering external MUI projects, I think that either we are ok with it (we use ahrefs's crawl once a week) or we use a different tradeoff. By "different tradeoff" I would expect that we can find good third-party projects for this problem because doesn't sound like an easy problem and I would expect something many people need.

alexfauquette:
On core it needs minor modifications (mostly renaming folders) and ignoring exceptions (careers pages, root pages seems a bit buggy. Some errors remain, but it allows to detect the mistakes accumulated with time

About the necessity to add it into the CI I'm not sure. It's not that sure. We are used to scripts that take between 30s to 1min and must be run for each PR. 

This one takes 1s for MUI-X repo and 4 seconds for MUI-core repo. and running them once a week is enough

Another solution to avoid binary result (returning 1 if broken links are detected) which do not support well script with false positives, we could generate a file with all the broken links (like we do with exported elements). 

Then a PR breaking a link would lead to a modification of this file that should be either resolved by fixing links, or run the script and add the file modification to the PR (equivalent of accepting the warning as a false positive)

## Results for material-ui repo

https://mui.com/blog/2019-developer-survey-results/#4-how-can-we-improve-material-ui-for-you
https://mui.com/blog/2020-developer-survey-results/#5-how-can-we-improve-material-ui-for-you
https://mui.com/blog/material-ui-v4-is-out/#premium-themes-store-✨ (wrong probably due to the emoji)

https://mui.com/joy-ui/main-features/perfect-dark-mode/#current-problem-the-flickering

https://mui.com/material-ui/customization/default-theme/#material-ui-core-styles-vs-material-ui-styles
https://mui.com/material-ui/customization/how-to-customize/#5-global-css-override
https://mui.com/material-ui/customization/palette/#user-preference
https://mui.com/material-ui/customization/theme-components/#adding-new-component-variants
https://mui.com/material-ui/customization/theming/#createtheme-options-theme
https://mui.com/material-ui/customization/typography/#typography-html-font-size

https://mui.com/material-ui/discover-more/roadmap/#our-priorities

https://mui.com/material-ui/getting-started/faq/#material-ui-is-awesome-how-can-i-support-the-project
https://mui.com/material-ui/getting-started/support/#professional-support-premium

https://mui.com/material-ui/guides/interoperability/#themeprovider
https://mui.com/material-ui/guides/minimizing-bundle-size/#option-2

https://mui.com/material-ui/icons/#font-icons

https://mui.com/material-ui/react-button/#grouped-buttons
https://mui.com/material-ui/react-button/#loading-buttons
https://mui.com/material-ui/react-button/#unstyled (wrong, because 'unstyled' section is added outside of the md)

https://mui.com/material-ui/react-modal/#unstyled (wrong, because 'unstyled' section is added outside of the md)

https://mui.com/material-ui/react-slider/#unstyled (wrong, because 'unstyled' section is added outside of the md)

https://mui.com/material-ui/react-switch/#unstyled (wrong, because 'unstyled' section is added outside of the md)

https://mui.com/material-ui/react-tabs/#unstyled (wrong, because 'unstyled' section is added outside of the md)

https://mui.com/material-ui/react-use-media-query/#usemediaquery

https://mui.com/system/getting-started/overview/#api-tradeoff
https://mui.com/system/getting-started/the-sx-prop/#performance-tradeoffs

https://mui.com/system/styles/advanced/#with-material-ui-core

oliviertassinari:
> This one takes 1s for MUI-X repo and 4 seconds for MUI-core repo. and running them once a week is enough

Ok interesting, my assumption is that it will eventually be forgotten. We can always try, simply that I would be surprised if it's not the final outcome 😁. 

> Results for material-ui repo

Awesome, it sounds like we have some broken links to fix! Bad for the DX.

> https://mui.com/material-ui/react-slider/#unstyled (wrong, because 'unstyled' section is added outside of the md)

This can be an argument to look at the HTML output rather than markdown.

alexfauquette:
Links will be fixed by https://github.com/mui/material-ui/pull/34320

Most of the broken links are due to:
- renaming from material-ui to mui
- section renaming (probably the rewriting of doc pages)

Then I built a distinct script to run locally that compares both mui-core and mui-x pages. I will run it from time to time when checking the SEO of mui-x

oliviertassinari:
It sounds great, once it becomes reliable 👍 to include it inside the docs-infra, though @mui/monorepo.

PR #6105, opened 2022-09-09
[docs] Fix link issues 404 & 301

oliviertassinari:
Based on https://app.ahrefs.com/site-audit/3992793/1/issues, a continuation of #6101

Screenshot 2022-09-12 at 17 12 07


docs/data/data-grid/api-object/api-object.md
```diff
@@ -1,8 +1,4 @@
----
-title: Data Grid - API object
----
```
alexfauquette, line 3: Why removing this meta data which is present in all the other files?

oliviertassinari: It duplicates with the h1 now. It's the same value.

alexfauquette: Sounds like it's always the same value except for pro/premium features which have the icon the `<h1>` and not in the title

alexfauquette (approved):

docs/data/data-grid/api-object/api-object.md
oliviertassinari: The capitalization is different in most of the places we're we do this, e.g. https://github.com/mui/mui-x/blob/f1294f981c78143b4670ef20276416f6d40946ed/docs/data/data-grid/column-definition/column-definition.md?plain=1#L1-L5

There is not duplication. However, based on @samuelsycamore https://www.notion.so/mui-org/Style-conventions-for-MUI-components-and-products-dff7d2d7d1ba4a4abd47f48cf345b800#e1ef8fa1b88d4f27819604f6ffd6f537, we are replacing the `<h1>` to match the `<title>`, to use the Title Case convention too.

oliviertassinari: To be very clear, on the line that you comment on, I did

```diff
-# Data grid - API object
+# Data Grid - API object
```

oliviertassinari:
@alexfauquette Note that if you want to own the SEO for MUI X, I would be happy to offload it 😁

PR #6144, opened 2022-09-13
[DatePicker] Display week number

### LINKED ISSUE #4606: [DatePicker] Display week numbers
I would like to have the possibility to display week numbers in the picker. Maybe as the first column before the first day of the week. Something like this would fit my needs:

Capture d’écran 2020-04-12 à 22 24 36

https://jqueryui.com/datepicker/#show-week

Capture d’écran 2020-04-12 à 22 21 14

http://react-day-picker.js.org/examples/customization-week-numbers

Capture d’écran 2020-04-12 à 22 21 54

https://reactdatepicker.com/#example-display-week-numbers

## Your Environment 🌎

| Tech        | Version |
| ----------- | ------- |
| @material-ui/lab | v5.0.0-alpha.16  |

alexfauquette:
Fix #4606

[doc preview](https://deploy-preview-6144--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#add-week-number)

image

After agreeing on the main pipeline:
- [x] Test
- [x] Refine design
- [x] Update adapter for native support
	- [x] date-fns
	- [x] dayjs
	- [x] luxon
	- [x] moment


alexfauquette:
@flaviendelangle I agree. I just wanted to see before with you and @LukasTy if we agree on the DX.

For now, there is no way to get the week number from date-io. So I propose a two steps strategy:

- We release a version with a callback and let to developer the responsibility of computing the week number.
- We extend adapters to support week number

This should avoid breaking changes

flaviendelangle:
If we were to add a method in the adapter and release it under a minor version, it would not be breaking right ?

alexfauquette:
Yes. For developpers using `getWeekNumber` the only difference will be that they can remove this prop when we add the adapter method. 

It's like introducing the customization callback before the default behavior 😅

alexfauquette:
I do not understand those typescript errors
```
error TS2345: Argument of type 'MuiPickersAdapter<TDate>' is not assignable to parameter of type 'MuiPickersAdapter<TDate | null>'.
```

Yes `MuiPickersAdapter<TDate>` and `MuiPickersAdapter<TDate | null>` are incompatible, but I do not find where does this `null` comes from

flaviendelangle:
`CalculateRangeChangeOptions` expects` `newDate: TDate`
And you are passing `rangePreviewDay` which has the type `TDate | null` in `DateRangePickerViewDesktop`.
So TS consider that the `TDate` of `CalculateRangeChangeOptions` equals `TDate | null` in `DateRangePickerViewDesktop`.

Which causes your problem: the utils of type `MuiPickersAdapter<TDate>` in `DateRangePickerViewDesktop` is considered to equal `MuiPickersAdapter<TDate | null>` by `CalculateRangeChangeOptions`.

____

This kind of issue often comes from similar problem, the one prop TS uses to infer the generic is nullable when it should not.

alexfauquette:
Thanks for the explanation. I fixed most of them. The remaining ones are about localization with `TInputDate`. I'm waiting for the PR removing this type to do the final clean-up

Sounds to be an interesting PR for detecting typing errors

flaviendelangle:
> Yes. For developpers using `getWeekNumber` the only difference will be that they can remove this prop when we add the adapter method.
> 
> It's like introducing the customization callback before the default behavior sweat_smile

My only issue with this is that people can customize the adapter to override methods.
So once we release the adapters with this new method, their will be 2 ways of doing this customization

- Pass a `getWeekNumber` prop
- Override the `getWeekNumber` (naming tbd) method of the adapter

If we consider that adapter override is never something we advertise but more an escape gate when people have no other solution.
Then I'm fine with introduction the `getWeekNumber` prop.
But if we recommend the override of the adapter for other topics, then I would be in favor of doing the same here.

alexfauquette:
For me, the `getWeekNumber` is the encouraged way to customize week number. 

The adapter problem is mention for usage with an extra adapter such as jalali which we do not export

gerdadesign:
> I feel like the week number is to close from the 1st column of days

Agree! Is there any reason the width of the week number can't match the width of the date number (36px)? It seems like this gif would demonstrate the expected behaviour as it scales, so that it can accommodate added columns while maintaining alignment. 


weeknumber

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -365,6 +416,16 @@ export function DayPicker<TDate>(inProps: DayPickerProps<TDate>) {
   return (
     <div role="grid" aria-labelledby={gridLabelId}>
       <PickersCalendarDayHeader role="row" className={classes.header}>
+        {displayWeekNumber && getWeekNumber && (
+          <PickersCalendarWeekNumberLabel
+            variant="caption"
+            role="columnheader"
+            aria-label={localeText.calendarWeekNumberHeaderLabel}
```
LukasTy, line 423: Have you tested a11y behaviour with any tool?
I've tried this implementation with NVDA on Windows and it does not seem to pick up this relationship between this `columnheader` and the `rowheader` on L469.
Also, with `week number` the screen reader assumes `8` columns, whereas it's actually still 7 with a `describing/header` week column.

FYI. I'm indifferent on what to do with `aria-label` on L423 as of right now it's value does not seem to be used, but in theory it makes sense to have it... 🤷 🤔 
Maybe @m4theushw has encountered such a11y case of having a non-interactive describing grid/table column and knows how to best implement it?

I'm proposing the following solution to allow for correct `week` and column count announcement when navigating the days.

```diff
diff --git a/packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx b/packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
index 961bebaa..6e241d92 100644
--- a/packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
@@ -459,18 +459,26 @@ export function DayPicker<TDate>(inProps: DayPickerProps<TDate>) {
             role="rowgroup"
             className={classes.monthContainer}
           >
-            {utils.getWeekArray(currentMonth).map((week) => (
+            {utils.getWeekArray(currentMonth).map((week, index) => (
               <PickersCalendarWeek
                 role="row"
                 key={`week-${week[0]}`}
+                // fix existing issue of announcing first row as row 2
+                aria-rowindex={index + 1}
                 className={classes.weekContainer}
               >
                 {displayWeekNumber && getWeekNumber && (
-                  <PickersCalendarWeekNumber className={classes.weekNumber} role="rowheader">
+                  <PickersCalendarWeekNumber
+                    className={classes.weekNumber}
+                    role="rowheader"
+                    aria-label={localeText.calendarWeekNumberColumnHeaderText(
+                      getWeekNumber(week[0]),
+                    )}
+                  >
                     {getWeekNumber(week[0])}
                   </PickersCalendarWeekNumber>
                 )}
-                {week.map((day) => {
+                {week.map((day, index) => {
                   const isFocusableDay =
                     focusableDay !== null && utils.isSameDay(day, focusableDay);
                   const isSelected = validSelectedDays.some((selectedDay) =>
@@ -495,6 +503,7 @@ export function DayPicker<TDate>(inProps: DayPickerProps<TDate>) {
                     tabIndex: isFocusableDay ? 0 : -1,
                     role: 'gridcell',
                     'aria-selected': isSelected,
+                    'aria-colindex': index + 1,
                   };
                   if (isToday) {
                     pickersDayProps['aria-current'] = 'date';
diff --git a/packages/x-date-pickers/src/locales/enUS.ts b/packages/x-date-pickers/src/locales/enUS.ts
index 23e58351..948fccf7 100644
--- a/packages/x-date-pickers/src/locales/enUS.ts
+++ b/packages/x-date-pickers/src/locales/enUS.ts
@@ -49,6 +49,7 @@ const enUSPickers: PickersLocaleText<any> = {
   // Calendar labels
   calendarWeekNumberHeaderLabel: 'Week number',
   calendarWeekNumberHeaderText: '#',
+  calendarWeekNumberColumnHeaderText: (weekNumber) => `Week ${weekNumber}`,
 
   // Open picker labels
   openDatePickerDialogue: (value, utils) =>
diff --git a/packages/x-date-pickers/src/locales/utils/pickersLocaleTextApi.ts b/packages/x-date-pickers/src/locales/utils/pickersLocaleTextApi.ts
index 2dc29945..0b988a2b 100644
--- a/packages/x-date-pickers/src/locales/utils/pickersLocaleTextApi.ts
+++ b/packages/x-date-pickers/src/locales/utils/pickersLocaleTextApi.ts
@@ -15,6 +15,7 @@ export interface PickersLocaleText<TDate> {
   end: string;
   calendarWeekNumberHeaderLabel: string;
   calendarWeekNumberHeaderText: string;
+  calendarWeekNumberColumnHeaderText: (weekNumber: string) => string;
   calendarViewSwitchingButtonAriaLabel: (currentView: CalendarPickerView) => string;
   inputModeToggleButtonAriaLabel: (
     isKeyboardInputOpen: boolean,
```

packages/x-date-pickers/src/locales/deDE.ts
```diff
@@ -59,6 +59,10 @@ const deDEPickers: Partial<PickersLocaleText<any>> = {
   minutesClockNumberText: (minutes) => `${minutes} ${clockViews.minutes}`,
   secondsClockNumberText: (seconds) => `${seconds}  ${clockViews.seconds}`,
 
+  // Calendar labels
```
LukasTy, line 62: Could you also add these to the newly introduced Japanese translations? 🙈

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
```diff
@@ -128,6 +141,38 @@ const PickersCalendarWeekDayLabel = styled(Typography, {
...
+  justifyContent: 'center',
+  alignItems: 'center',
+  color: theme.palette.text.disabled,
+}));
+
+const PickersCalendarWeekNumber = styled(Typography, {
+  name: 'MuiDayPicker',
+  slot: 'WeekNumber',
+  overridesResolver: (_, styles) => styles.weekNumber,
+})(({ theme }: { theme: Theme }) => ({
```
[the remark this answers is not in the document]
alexfauquette: No, it's a from when I tried to add an `ownerState`

packages/x-date-pickers/src/CalendarPicker/DayPicker.tsx
alexfauquette: I tried to respect sementic meaning but did not tested.

I searched a bit about how other libraries do. A solution is to use `aria-hidden` on the header and the week number to have the correct number of row/columns. But then you have to adapt the `aria-label` of each button such that they provide the day

LukasTy: Yeah, I also thought/tried `aria-hidden` approach to the non-interactive row/column, but it seemed like a really not cool hack. 🙈 
IMO, `aria-colindex` and `aria-rowindex` are way less intrusive way to achieve what we want.
FYI, possibly using a `table` would be more accessible by screen readers, haven't looked into it, but even MDN specifies that if possible, table layout should be preferred instead of replicating same behaviour with necessary `role` elements.

alexfauquette: I agree but not sure it's good to do a feature and an big accessibility update in the same PR

alexfauquette: Done 👍

alexfauquette:
@gerdadesign Yes, we can move the month name and arrows by modifying the left and right padding

Is that well aligned?
image

gerdadesign (approved):
the alignment looks better to me now, thanks!

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -110,7 +110,7 @@ This value is usually set to `6` for Gregorian calendars, because months display
 ### Add week number
 
 To display week number, use the `displayWeekNumber`.
-If you use a `dateAdapter` imported from outside of our library, you will have to provide the method `getWeekNumber`.
+If you do not import your `dateAdapter` from `@mui/x-date-pickers` or `@mui/x-date-pickers-pro`, you will have to provide the method `getWeekNumber`.
```
flaviendelangle, line 113: Now that we know we will force usage of adapters imported from our packages, is this `getWeekNumber` method relevent ?

docs/data/date-pickers/date-picker/JalaliDatePicker.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { AdapterDateFnsJalali } from '@mui/x-date-pickers/AdapterDateFnsJalali';
 import { DatePicker } from '@mui/x-date-pickers/DatePicker';
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 
 export default function JalaliDatePicker() {
   const [value, setValue] = React.useState<Date | null>(new Date(2022, 3, 7));
 
   return (
-    <LocalizationProvider dateAdapter={AdapterJalali}>
+    <LocalizationProvider dateAdapter={AdapterDateFnsJalali}>
```
flaviendelangle, line 11: I guess this is not related to this PR
Maybe we could extract it into a standalone PR

But this section will most likely go away after #6625 anyway

Same for the change in the `.md` file

packages/x-date-pickers-pro/src/DateRangePicker/date-range-manager.ts
```diff
@@ -1,13 +1,15 @@
 import { MuiPickersAdapter } from '@mui/x-date-pickers/internals';
 import { DateRange } from '../internal/models';
 
-interface CalculateRangeChangeOptions<TDate> {
+interface CalculateRangePreviewOptions<TDate> {
```
flaviendelangle, line 4: Is this change related to the current PR ?
Make sure it is compatible with #6763 since it touches the same parts of the code

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -98,13 +98,22 @@ You can take advantage of the [PickersDay](/x/api/date-pickers/pickers-day/) com
...
+### Show additional days
+
 To shows all days of displayed weeks, included those outside of the current month, use `showDaysOutsideCurrentMonth`.
 
 By default, only weeks of the current month are displayed, but you can provide a total number of week to display with `fixedWeekNumber` prop.
 This value is usually set to `6` for Gregorian calendars, because months display can vary between 4 and 6 weeks.
 
 {{"demo": "CustomMonthLayout.js"}}
 
+### Add week number
```
flaviendelangle, line 110: This is also available on the date time pickers and the date range pickers
How can we make that clear ? (same for `showDaysOutsideCurrentMonth` by the way).

Should we add a small section on those pages saying that the calendar customization of the date picker also apply to them ?
Or move this content in a `DateCalendar` / `DayCalendar` page ?

LukasTy (approved):
I feel that the layout of both the picker popup as well as date range picker does seem somewhat claustrophobic/squashed with `displayWeekNumbers` prop added, but I take it that we will address it with the linked issue regarding views width/height customisability? 🤔 

Screenshot 2022-11-17 at 08 47 50 1
Screenshot 2022-11-17 at 08 48 22

docs/data/date-pickers/date-picker/date-picker.md
flaviendelangle: I would say on the Date Calendar page ultimately with a small note on the Date Picker and Date Time Picker page to make sure people find it.

But for now maybe on the Date Picker page with a small note on the Date Time Picker page.

If @joserodolfofreitas or @samuelsycamore have an opinion here, it's a topic that will happen more and more.

docs/data/date-pickers/date-picker/JalaliDatePicker.tsx
alexfauquette: Ok, I first taught I mixed two PR, but in fact it's not the case.

Now that we have all the adapter exported internally (except js-joda) I added the `getWeekNumber` method to all the adapters. THen I realized the doc was still encouraging to import them from date-io so I made a correction

packages/x-date-pickers/src/AdapterDateFnsJalali/index.ts
```diff
@@ -1,5 +1,7 @@
 import BaseAdapterDateFnsJalali from '@date-io/date-fns-jalali';
 import defaultLocale from 'date-fns-jalali/locale/fa-IR';
+import getWeek from 'date-fns-jalali/getWeek';
+// @ts-ignore
```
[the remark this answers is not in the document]
alexfauquette: We do not expect an error, we just ignore ts here because the internal date-fns file is not typed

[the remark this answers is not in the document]
flaviendelangle: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

Here is the official recommendation

packages/x-date-pickers/src/DateCalendar/tests/DateCalendar.test.tsx
```diff
@@ -163,6 +163,20 @@ describe('<DateCalendar />', () => {
...
+      <LocalizationProvider dateAdapter={AdapterClassToUse}>
+        <DateCalendar
+          value={adapterToUse.date(new Date(2019, 0, 1))}
+          onChange={() => {}}
+          displayWeekNumber
+        />
+      </LocalizationProvider>,
+    );
+
+    expect(screen.getAllByRole('rowheader').length).to.equal(5);
```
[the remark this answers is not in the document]
alexfauquette: I tend to prefer selectors based on roles because it forces to respect HTML semantic

packages/x-date-pickers/src/AdapterDateFnsJalali/index.ts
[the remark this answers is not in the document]
alexfauquette: Nice, explanations.
Let stay with ignore, in case date-fns export types for its `'date-fns/_lib/format/longFormatters'`

flaviendelangle:
@joserodolfofreitas this translation key is a callback.
It is the one used to render 45 / 46 etc... on your screen, so it needs to by dynamic.
For the header you are looking for `calendarWeekNumberHeaderText`

joserodolfofreitas:
> For the header you are looking for `calendarWeekNumberHeaderText`

Thanks! It'll probably be covered by any community translation, but should it work like this?

localeText={{ calendarWeekNumberHeaderText: "KW" }} to replace the "#" ?

https://codesandbox.io/s/custom-week-numbers-es327n?file=/demo.tsx

flaviendelangle:
Oh indeed we have a big bug with the nested `LocalizationProvider`

The following work: https://codesandbox.io/s/custom-week-numbers-forked-2d033y?file=/demo.tsx
But if you put your locale on the `LocalizationProvider`, it gets lost

docs/data/date-pickers/date-picker/date-picker.md
joserodolfofreitas: Based on the mentioned examples, I think the Date Calendar sounds like a good page to host these props.
But I feel we're adding a layer of complexity and it can maybe start getting confusing if we spread these high-level properties in different subcomponent pages.

We briefly discussed a slightly different variation of this idea, but maybe we could consider a shared properties file and then re-print the documentation on every main component page (with the properties applicable to that component). We may not need to have an "original link" to host them.

joserodolfofreitas, line 110:
```suggestion
### Display week number
```

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -98,13 +98,22 @@ You can take advantage of the [PickersDay](/x/api/date-pickers/pickers-day/) com
 
 You can customize the month layout with some props.
 
+### Show additional days
+
 To shows all days of displayed weeks, included those outside of the current month, use `showDaysOutsideCurrentMonth`.
```
joserodolfofreitas, line 103: Perhaps we can review the props names to be more consistent between each other (in another PR).
For instance, would it make sense to rename this property to `displayDaysOutsideCurrentMonth`?

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -98,13 +98,22 @@ You can take advantage of the [PickersDay](/x/api/date-pickers/pickers-day/) com
...
 
 By default, only weeks of the current month are displayed, but you can provide a total number of week to display with `fixedWeekNumber` prop.
 This value is usually set to `6` for Gregorian calendars, because months display can vary between 4 and 6 weeks.
 
 {{"demo": "CustomMonthLayout.js"}}
 
+### Add week number
+
+To display week number, use the `displayWeekNumber`.
+You can customize the string displayed, by using the localization key`localeText.calendarWeekNumberText`.
```
joserodolfofreitas, line 113: I'm unsure if there are use cases where users need to render something different for the week numbers, I'd think there's more chance that they want to update the header.

```suggestion
You can customize the calendar week header by using the localization key `localeText.calendarWeekNumberHeaderText`.
As well you can also customize what's rendered as a calendar Week Number, using a callback for the localization key `localeText.calendarWeekNumberText`
```

Should we also add an example for the callback usage?

docs/data/date-pickers/date-picker/date-picker.md
alexfauquette: Why not. Can you create an issue for it?

docs/data/date-pickers/date-picker/date-picker.md
alexfauquette: Not super inspired by what could be a customized week number, so I simply added a point at the end of week numbers, and I override the header to set the default `"#"`. It does not change any thing but it highlight the localization key.

flaviendelangle (approved):
Great work !
Sorry for the delay reviewing that one

packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.tsx
```diff
@@ -36,7 +42,7 @@ const StaticNextDateRangePicker = React.forwardRef(function StaticNextDateRangeP
     showToolbar: defaultizedProps.showToolbar ?? displayStaticWrapperAs === 'mobile',
   };
 
-  const { renderPicker } = useStaticPicker({
+  const { renderPicker } = useStaticPicker<DateRange<TDate>, TDate, DateOrTimeView, typeof props>({
```
flaviendelangle, line 45: I would pass `'day'` instead of `DateOrTimeView` here like for the other date range pickers
And remove the added views to the lookup

`DateOrTimeView` is wrong since we will never have time view here
And `DateView` would feel misleading since we don't really have month / year view

flaviendelangle (commented):
Nice catch !

Eshakheni123:
why we can not click weekNumber.
I need functionality like this, if I click weekNumber then should be the render method.
something like this:
https://codesandbox.io/s/react-calendar-show-week-numbers-and-fixed-number-of-weeks-forked-hi4y1t?file=/src/index.js

Issue #5346, opened 2022-01-17
[StaticDateRangePicker] The month does not change when the value is changed programmatically


JOEAV:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Current behavior 😯

Hey Guys.
It seems that setting up a new Date value to the StaticDateRangePicker programmatically (as with changing the value prop dynamically  without touching the picker ui by hands) 
not refocusing the picker to the new value. 
The actual date does being selected, but you have to go to the targeted month by hand.

My Case Scenario :
<img width="484" alt="Screen Shot 2022-01-17 at 15 51 09" src="https://user-images.githubusercontent.com/11891207/149781654-2f924aaa-9c43-4687-8f77-0b0ed6841f44.png">

I would like to render a "Popular Occasions" by the calendar , so  a user wouldn't have to travel  a lot by ui gestures related to changing the pickers month.

### Expected behavior 🤔

The StaticDateRangePicker should switch rendered month to the targeted dynamically changed Date, even if not changed by the input textbox / ui Gesture (left/right month arrows)

### Steps to reproduce 🕹

Steps:

1.Add a new StaticDateRangePicker
2. Change the value prop dynamically by any method you like (button, function etc), to a date which happens in a month other than the current rendered month
3. See that the other month you set not being rendered as should
4.

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```

System:
    OS: macOS 11.6
  Binaries:
    Node: 15.14.0 - ~/.nvm/versions/node/v15.14.0/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 8.1.3 - ~/.nvm/versions/node/v15.14.0/bin/npm
  Browsers:
    Chrome: 97.0.4692.71
    Edge: Not Found
    Firefox: Not Found
    Safari: 15.0
  npmPackages:
    @emotion/react: ^11.6.0 => 11.7.1 
    @emotion/styled: ^11.6.0 => 11.6.0 
    @mui/base:  5.0.0-alpha.63 
    @mui/icons-material: ^5.1.1 => 5.2.5 
    @mui/lab: ^5.0.0-alpha.55 => 5.0.0-alpha.63 
    @mui/material: ^5.1.0 => 5.2.7 
    @mui/private-theming:  5.2.3 
[...]

PR #6166, opened 2022-09-15
[DateRangePicker] Switch to new month when changing the value from the outside

flaviendelangle:
Closing #5346

## Behavior changes

- Do not check if `open === true` (I don't understand this check, on pickers and it always `true` - otherwise this component is not rendered - and on static pickers it is always `false`, which caused #5346
- When `disableAutoMonthSwitching`, allow to switch month when it is caused by full value change (makez #5346 works even with this prop)
- Do not restrict the month change when the target date is disabled, `CalendarPicker` is not doing it.


LukasTy (approved):
LGTM 🙌 👍

PR #6168, opened 2022-09-15
v5.17.3

m4theushw:


CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 5.17.3
+<!-- generated comparing v5.17.2..master -->
```
LukasTy, line 7:
```suggestion

```

CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
...
+
+- [DataGrid] Only update input with value prop if debounce is off (#5646) @m4theushw
+
+### `@mui/x-date-pickers@v5.0.2` / `@mui/x-date-pickers-pro@v5.0.2`
+
+#### Changes
+
+- [pickers] Add Icelandic (is-IS) locale (#6137) @flaviendelangle
+- [pickers] Fix `themeAugmentation` and style overriding (#6156) @LukasTy
+- [pickers] Fix `x-date-pickers-pro` theme augmentation (#6096) @LukasTy
```
LukasTy, line 28:
```suggestion
- [pickers] Fix community package `themeAugmentation` and style overriding (#6156) @LukasTy
- [pickers] Fix pro package `themeAugmentation` and style overriding (#6096) @LukasTy
```

CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 5.17.3
+<!-- generated comparing v5.17.2..master -->
+_Sep 15, 2022_
+
+We'd like to offer a big thanks to the 6 contributors who made this release possible. Here are some highlights ✨:
+
+- 📝 Fix lost characters when typing into edit fields (#5646) @m4theushw
```
flaviendelangle, line 12: For this kind of description, I would mention explicitly the datagrid because the exact same sentence could make sense for the pickers

CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
...
+
+#### Changes
+
+- [DataGrid] Only update input with value prop if debounce is off (#5646) @m4theushw
+
+### `@mui/x-date-pickers@v5.0.2` / `@mui/x-date-pickers-pro@v5.0.2`
+
+#### Changes
+
+- [pickers] Add Icelandic (is-IS) locale (#6137) @flaviendelangle
```
flaviendelangle, line 26:
```suggestion
- [pickers] Add Icelandic (is-IS) locale (#6137) @elvatli
```

I did the picking on master, but I would be in favor of mentioning the original contributor.

CHANGELOG.md
flaviendelangle, line 28:
```suggestion
- [pickers] Fix `@mui/x-date-pickers-pro` theme augmentation (#6096) @LukasTy
```

CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
...
+- [pickers] Fix `x-date-pickers-pro` theme augmentation (#6096) @LukasTy
+- [pickers] Improve German (de-DE) locale (#6138) @alexfauquette
+
+### Docs
+
+- [docs] Improve main demo to show new functionalities (#5292) @joserodolfofreitas
+
+### Core
+
+- [core] Remove Storybook (#6040, #6098) @flaviendelangle
```
flaviendelangle, line 37: Can be removed since I reverted the change (I merged it in `master` by mistake)

CHANGELOG.md
```diff
@@ -3,6 +3,42 @@
...
+- [core] Update to typescript 4.8.3 (#6136) @flaviendelangle
+- [core] Update RFC template (#6100) @bytasv
+- [core] v5.17.2 (#6086) @flaviendelangle
```
flaviendelangle, line 40: Can be removed

CHANGELOG.md
m4theushw:
```suggestion
- 📝 Fix lost characters when typing into fields in the DataGrid(#5646) @m4theushw
```

CHANGELOG.md
m4theushw:
```suggestion
```

CHANGELOG.md
m4theushw:
```suggestion
```

CHANGELOG.md
```diff
@@ -3,6 +3,39 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 5.17.3
+_Sep 15, 2022_
+
+We'd like to offer a big thanks to the 6 contributors who made this release possible. Here are some highlights ✨:
+
+- 📝 Fix lost characters when typing into fields in the DataGrid(#5646) @m4theushw
```
m4theushw, line 11:
```suggestion
- 📝 Fix lost characters when typing into fields in the DataGrid (#5646) @m4theushw
```

CHANGELOG.md
```diff
@@ -3,6 +3,39 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 5.17.3
```
m4theushw, line 6:
```suggestion
## 5.17.3

```

CHANGELOG.md
```diff
@@ -3,6 +3,40 @@
...
+
+_Sep 15, 2022_
```
m4theushw, line 8:
```suggestion
_Sep 16, 2022_
```

LukasTy (approved):

alexfauquette (approved):

joserodolfofreitas (approved):

packages/x-date-pickers/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mui/x-date-pickers",
-  "version": "5.0.1",
+  "version": "5.0.2",
```
oliviertassinari, line 3: Time to do the shift?

```suggestion
  "version": "5.17.3",
```

For example:

Screenshot 2022-09-16 at 13 40 16

https://bestpractices.coreinfrastructure.org/en/projects/6293

LukasTy: IMHO, I think that we already missed the possible version sync when moving from `beta` to `stable`.
Where
- `5.0.0-beta.7` -> `5.17.1` would have seemed somewhat logical
- `5.0.1` -> `5.17.3` seems like a strange unwarranted version jump

Correct me if I'm wrong, but based on your screenshot, I don't see anything forcing to have synced versions number – we must have unique version numbers for every release..? 🤔

m4theushw: I'll keep as it is but we can discuss about unifying the versions and do it in the next release.

oliviertassinari: > 5.0.0-beta.7 -> 5.17.1 would have seemed somewhat logical

Agree, I don't think that we should have had v5.0.0, v5.0.1 and v5.0.2

> I don't see anything forcing to have synced versions number – we must have unique version numbers for every release

True, the rationale would be: simplicity. On the same topic, I think that the only reason why we don't have https://lerna.js.org/docs/features/version-and-publish#fixedlocked-mode-default is because we have these unstable npm repository for new components.

LukasTy: > I think that it would have been ideal, but I also think that it's never too late to align the versions.

We could aim to have synced versions going forward from v6. Although, that sort of also means, that we would always need to release all the packages, even though they don't have any changes in them.
What about the `x-license-pro` package then – it doesn't receive frequent changes. 🤔 

> I wished we could use https://lerna.js.org/docs/features/version-and-publish#fixedlocked-mode-default but because of the unstable npm repository for new components, it seems that we can't.

Isn't [our `lerna.json` config](https://github.com/mui/mui-x/blob/next/lerna.json#L4) a bit incorrect?
Lerna docs state, that if we want to run `lerna` in `independent` mode, the version field should have this string for version instead..? 🤔 
Well, if want to have all the packages in same repo and over time add new packages with their own (canary) lifecycle, it forces us to run in `independent` mode.

oliviertassinari: > that we would always need to release all the packages, even though they don't have any changes in them.
What about the x-license-pro package then – it doesn't receive frequent changes.

@LukasTy Oops, sorry for the confusion. What I was referring to with "ideal" or "same version"s is https://lerna.js.org/docs/features/version-and-publish#fixedlocked-mode-default 's behavior, which is different.

> Isn't our lerna.json config a bit incorrect?

Interesting. I guess it depends. In MUI Core, we do https://github.com/mui/material-ui/blame/2be49496773b599a539e815f2e4e45c8bd9a484c/lerna.json#L4, it's a bit painful to have to set each version manually but there are quite some alpha packages, so it might still be less pain, than having to fix:

Screenshot 2022-09-19 at 16 10 39

On MUI X, we have fewer alpha packages, so I think a yes to all of these is simple enough:

Screenshot 2022-09-19 at 16 11 50

LukasTy: For the near future we will have cycles of having stable and/or canary releases, switching release strategy because of that wouldn't make much sense IMO.

I see only one thing in question which we should decide on—should we sync the stable version of a package coming out of canary towards stable channel with the version of another package in the repo already in stable channel.
With this question—it would implicitly mean the version sync of other packages as well.

IMHO, I would be in favour of sticking with the `mui` repository approach of different packages having their own version cycles and in turn not having empty releases.
One of the arguments for going with this—the developers should already be familiar with this approach of having different versions from the `mui` repo packages. 🤔 
Even industry "giants" like [babel](https://github.com/babel/babel/tree/main/packages) are also going with a similar approach of having packages with their own release lifecycles.

One thing I think we should fix/adjust with the next release is the `version` field in `lerna.json`.
Should we align it with our release approach and change to `independent` with the next release @flaviendelangle @m4theushw ?

PR #6170, opened 2022-09-15
[pickers] Do not support unparsed date formats anymore

### LINKED ISSUE #5759: [RFC] Should we keep the different between `TInputDate` and `TDate` in v6
## Context

Currently, all the picker components have two different formats:

- `TDate`: the date representation of the adapter (`Date` for `date-fns`, `DayJs` for `dayjs`, etc...)

- `TInputDate`: the date format that the parsing method of the adapter can understand (usually it is something like `string | null | Date | TDate`)

Depending on what prop you are looking at, you might pass / receive one or the other.

Here are a few example:

- `props.value` takes a `TInputDate`
- `props.minDate` takes a `TDate`
- `props.onChange` receives a `TDate`
- `props.getOpenDialogAriaText` receives a `TInputDate` (no idea why)
- `props.isDateDisabled` receives a `TDate`
- etc...

`TDate` and `TInputDate` exact format depends on the adapter being used, which is why I am using the generic name and not an actual interface name.

## Why do we have this behavior

The clear benefit of this behavior is that developers don't have to parse there date before passing it to the pickers.
For instance, when using `dayjs`, any of the following is valid:

```tsx
<DatePicker value={dayjs()} />
<DatePicker value={new Date()} />
<DatePicker value="2022-08-11T09:49:30.347Z" />
```

### It is hard to understand when to use one format or the other

Some people passes a `string` in `props.value` so they think they must convert the date received in `onChange` to a `string` (#4976)

Some people passes a `Date` in `props.minDate` / `props.maxDate`, thinking it works just like `props.value` (#4474)

### It does not follow the classic controlled input format

Usually, when we have a controlled input, it has the following format:

```ts
interface MyComponentProps {
  value: A,
  onChange: (newValue: A) => void, // Or some event containing the value in the `A` format`
}
```

Right now we basically have:

```ts
interface MyComponentProps {
  value: A,
  onChange: (newValue: B) => void,
}
```

### Benchmark

Alternatives are not doing it:

- [Spectrum](https://react-spectrum.adobe.com/react-spectrum/DateField.html#value): both `value` and `onChange` uses a custom `CalendarDate` format (equivalent to our `TDate` with an adapter for there `@internationalized/date` package) 

- [Telerik](https://www.telerik.com/kendo-react-ui/components/dateinputs/api/DatePickerChangeEvent/): both `value` and `onChange` uses `Date` (equivalent to our `TDate` for the `date-fns` adapter)

- [Retool](https://retool.com/components/date): `value` uses `Date`, I did not find information about the format used by `onChange`

flaviendelangle:
Closes #5759


docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -73,7 +73,7 @@ For ease of use, the date picker will automatically change the layout between po
 
 ## Sub-components
 
-Some lower-level sub-components (`CalendarPicker`, `MonthPicker`, and `YearPicker`) are also exported. These are rendered without a wrapper or outer logic (masked input, date values parsing and validation, etc.).
+Some lower-level sub-components (`CalendarPicker`, `MonthPicker`, and `YearPicker`) are also exported.
```
alexfauquette, line 76: I don't see the link with this PR

docs/data/date-pickers/time-picker/time-picker.md
```diff
@@ -55,7 +55,7 @@ The time picker component can be disabled or read-only.
 
 ## Sub-components
 
-Some lower-level sub-components (`ClockPicker`) are also exported. These are rendered without a wrapper or outer logic (masked input, date values parsing and validation, etc.).
+Some lower-level sub-components (`ClockPicker`) are also exported.
```
alexfauquette, line 58: same

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
```diff
@@ -143,10 +141,10 @@ function DateRangePickerViewRaw<TInputDate, TDate>(
     shouldDisableDate &&
     ((dayToTest: TDate) => shouldDisableDate?.(dayToTest, currentlySelectingRangeEnd));
 
-  const [start, end] = parsedValue;
+  const [start, end] = value;
   const { changeMonth, calendarState, onMonthSwitchingAnimationEnd, changeFocusedDay } =
-    useCalendarState({
-      value: start || end,
+    useCalendarState<TDate>({
+      value: start ?? end,
```
alexfauquette, line 147: This would change the behavior if  `start` is `null`?

alexfauquette (commented):
I read some files, but it would take an eternity to open each file modification to check if modification are just type renaming or logic modification. Are there some specific lines on which you have a doubt?

LukasTy:
> I read some files, but it would take an eternity to open each file modification to check if modification are just type renaming or logic modification.

```js
[...document.querySelectorAll('button.load-diff-button')].forEach(button => button.click());
```
🙈 😆

packages/x-date-pickers/src/locales/deDE.ts
```diff
@@ -60,13 +60,13 @@ const deDEPickers: Partial<PickersLocaleText<any>> = {
   secondsClockNumberText: (seconds) => `${seconds}  ${clockViews.seconds}`,
 
   // Open picker labels
-  openDatePickerDialogue: (rawValue, utils) =>
-    rawValue && utils.isValid(utils.date(rawValue))
-      ? `Datum auswählen, gewähltes Datum ist ${utils.format(utils.date(rawValue)!, 'fullDate')}`
+  openDatePickerDialogue: (value, utils) =>
+    value != null && utils.isValid(value)
```
LukasTy, line 64: Nitpick: Shouldn't it be `value !== null` instead? 🤔

LukasTy (approved):

docs/data/date-pickers/date-picker/date-picker.md
flaviendelangle: > date values parsing

There is no parsing any more, that's the part related to this PR

> validation

Not related to this PR but this is false, they do have validation

> masked input

Indeed, but I don't think it's worth specifying the the `MonthPicker` does not have a masked input, it's obvious by looking at the example below

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
flaviendelangle: > This would change the behavior if start is null?

No, `null` also triggers the fallback here
The diff would be if `start === 0` or `start === ''`
But I agree it's not purely related to this PR.
I tend to remove the `date || xxx` because if we have a date library storing dates as timestamp it can blow in our face.
With that being said, that's probably very unlikely

Removed

packages/x-date-pickers/src/locales/deDE.ts
flaviendelangle: It should not change anything here since the value can't be undefined
But yes we should have the strict test

Fixed

flaviendelangle:
> I read some files, but it would take an eternity to open each file modification to check if modification are just type renaming or logic modification. Are there some specific lines on which you have a doubt?

The internals mostly, the main point of risk for me is in `usePickerState` and the `shared.ts` (for each picker family)

packages/x-date-pickers/src/internals/components/wrappers/WrapperProps.ts
```diff
@@ -1,8 +1,9 @@
+import * as React from 'react';
```
alexfauquette, line 1:
```suggestion
```

packages/x-date-pickers-pro/src/MultiInputDateRangeField/useMultiInputDateRangeField.ts
```diff
@@ -36,16 +34,10 @@ export const useMultiInputDateRangeField = <
...
-        newDateRange = dateRangePickerValueManager.parseInput(utils, valueProp);
-      } else if (firstDefaultValue.current !== undefined) {
-        newDateRange = dateRangePickerValueManager.parseInput(utils, firstDefaultValue.current);
-      } else {
-        newDateRange = dateRangePickerValueManager.emptyValue;
-      }
-
-      newDateRange[index] = newDate;
+      const currentDateRange =
+        valueProp ?? firstDefaultValue.current ?? dateRangePickerValueManager.emptyValue;
```
alexfauquette, line 38: Don't you need to clean this value?

```suggestion
        dateRangePickerValueManager.cleanValue(valueProp) ?? firstDefaultValue.current ?? dateRangePickerValueManager.emptyValue;
```

alexfauquette (approved):

packages/x-date-pickers/src/internals/components/wrappers/WrapperProps.ts
flaviendelangle: My IDE complains when you use JSX without importing React which is not recommended) :angel: 

image

packages/x-date-pickers-pro/src/MultiInputDateRangeField/useMultiInputDateRangeField.ts
flaviendelangle: Indeed
I will improve it in a follow up to use `useControlled` and avoid these fallback.

packages/x-date-pickers/src/internals/components/wrappers/WrapperProps.ts
alexfauquette: Ho yes, I missed `React.ElementType` in the interface

PR #6171, opened 2022-09-15
[docs] Link the OpenSSF Best Practices card

oliviertassinari:
This is part of improving our score in https://deps.dev/npm/%40mui%2Fx-date-pickers

Screenshot 2022-09-15 at 19 21 45

See https://bestpractices.coreinfrastructure.org/en for why it matters, it also seems to be confirmed with https://mui.zendesk.com/agent/tickets/4926. An example of a project having this: https://github.com/tensorflow/tensorflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ They feature advanced functionality for data-rich applications and a wide range 
 MUI X is **open core**—base components are MIT-licensed, while more advanced features require a Pro or Premium commercial license.
 See the [Licensing page](https://mui.com/x/introduction/licensing/) for details.
 
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/6293/badge)](https://bestpractices.coreinfrastructure.org/projects/6293)
+
 ## Components
 
 - [Data Grid](https://mui.com/x/react-data-grid/)


README.md
```diff
@@ -13,6 +13,8 @@ They feature advanced functionality for data-rich applications and a wide range
 MUI X is **open core**—base components are MIT-licensed, while more advanced features require a Pro or Premium commercial license.
 See the [Licensing page](https://mui.com/x/introduction/licensing/) for details.
 
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/6293/badge)](https://bestpractices.coreinfrastructure.org/projects/6293)
```
LukasTy, line 16: 🙌 
What do you think about putting this badge just after the image/name, before any text - after line 6 in our case.
From what I can see, it's a more common approach by other libraries, like the one you linked or:
- [React](https://github.com/facebook/react)
- [Vite](https://github.com/vitejs/vite)
- [freeCodeCamp](https://github.com/freeCodeCamp/freeCodeCamp)
- [Flutter](https://github.com/flutter/flutter)

FYI, have you considered adding [scorecard generation](https://securityscorecards.dev/) as well? 🤔

oliviertassinari: I have copied https://github.com/mui/material-ui#readme when it comes to where to place the badges, no real preferences.

> FYI, have you considered adding [scorecard generation](https://securityscorecards.dev/) as well?

Yeah, it could make sense, for the badge as well.

oliviertassinari:
I'm merging as the goal is to have it present. The two comments left could be follow-ups:

1. we could add more badges: agree.
2. we could change the position of these badges: agree but I'm following MUI Core and MUI Toolpad conventions.

PR #6180, opened 2022-09-16
[core] Update monorepo

flaviendelangle:
I need https://github.com/mui/material-ui/pull/34311 for https://github.com/mui/mui-x/pull/6070

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@mnajdova/enzyme-adapter-react-18": "^0.2.0",
     "@mui/icons-material": "^5.10.3",
     "@mui/material": "^5.10.4",
-    "@mui/monorepo": "https://github.com/m4theushw/material-ui.git#master",
+    "@mui/monorepo": "https://github.com/flaviendelangle/material-ui.git#master",
     "@mui/utils": "^5.10.3",
     "@octokit/plugin-retry": "^3.0.9",
     "@octokit/rest": "^18.12.0",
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "@mnajdova/enzyme-adapter-react-18": "^0.2.0",
     "@mui/icons-material": "^5.10.3",
     "@mui/material": "^5.10.4",
-    "@mui/monorepo": "https://github.com/m4theushw/material-ui.git#master",
+    "@mui/monorepo": "https://github.com/flaviendelangle/material-ui.git#master",
```
oliviertassinari, line 85:
```suggestion
    "@mui/monorepo": "https://github.com/mui/material-ui.git#master",
```

flaviendelangle: The problem is that, unless it has been fixed and I don't know, the codesandbox cache of the `mui` repo is still broken.
It's the problem when migrating Argos.

m4theushw: I created https://github.com/codesandbox/codesandbox-ci-ui/issues/49 but I don't hope they will implement this. There're issues from last year without any comment in this repo.

m4theushw (approved):

package.json
oliviertassinari: Oh, I see. I think that I could solve the issue. See the build history in https://ci.codesandbox.io/status/mui/mui-x/pr/5938.

- 294382 failed, I started to look at a fix
- 294385 failed again with the first attempt I did
- 294386 passed with the second attempt I did https://github.com/mui/mui-x/pull/5938/commits/0d204c98c1fb527156cb3983a6acb570a650c01a
- 294388 still passed without the second attempt

PR #6191, opened 2022-09-17
[docs] Fix hidden popper in restore state example

### LINKED ISSUE #6190: [DataGrid/Docs] In the Restore the state example, CustomToolbar was hidden behind the pinned-column.
### Duplicates

### Latest version

### Current behavior 😯

When you `Save current view` then you pin the column on the example, the custom config dropdown was in behind the pinned column.

data-grid-state-apiref-bug

### Expected behavior 🤔

expected-data-grid

### Steps to reproduce 🕹

Link to live example: https://mui.com/x/react-data-grid/state/#restore-the-state-with-apiref

Steps:
1. Click on `Save current view` and named it "custom view name" (if it's not long enough, it will completed hide by pinned column)
2. Save the view
3. Click Dropdown on **Desk** column and choose **Pin to left**
4. Click **Custom View** to open dropdown and you will see it

### Context 🔦

This caused because the pinned column has `z-index:1` but Popper itself has no `z-index`. I personally don't think this is big issue because developer can figure it out themselves as well but having the document that works with all edge cases would be good 😄 

I can open the PR for fix this really small stuff. 🙇🏻 

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

heyfirst:
Closes #6190 

By fixing the Popper, I explicitly add `sx={{ zIndex: 'modal' }}` to Popper which will apply this z-index on to `MuiPopperUnstyled-root` class directly as `style` attribute does not work with Popper component.

Expected behavior you can see this screen record below. More information provided on [the issue](#6190)

expected-data-grid

### DIFF
--- a/docs/data/data-grid/state/RestoreStateApiRef.js
+++ b/docs/data/data-grid/state/RestoreStateApiRef.js
@@ -307,7 +307,7 @@ function CustomToolbar() {
           role={undefined}
           transition
           placement="bottom-start"
-          sx={{ zIndex: 1 }}
+          sx={{ zIndex: 'modal' }}
         >
           {({ TransitionProps }) => (
             <Fade {...TransitionProps} timeout={350}>
--- a/docs/data/data-grid/state/RestoreStateApiRef.js
+++ b/docs/data/data-grid/state/RestoreStateApiRef.js
@@ -307,7 +307,7 @@ function CustomToolbar() {
           role={undefined}
           transition
           placement="bottom-start"
-          style={{ zIndex: 1 }}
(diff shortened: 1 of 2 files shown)


alexfauquette (approved):
Nice, I ignored this `sx` shortcut :)

heyfirst:
Tested and it works as expected.

Screen Shot 2565-09-19 at 14 31 44

alexfauquette:
Thanks for the fix 👍

heyfirst:
@alexfauquette My pleasure! Warmup for Hacktoberfest 🥳

PR #6219, opened 2022-09-19
[I10n] Add Finnish (fi-FI) locale to Datepicker

PetroSilenius:
I noticed that the DataGrid had support for Finnish locale but datepicker was lacking it

- Adds Finnish locale to Datepicker
- Adds missing Finnish translations to DataGrid

Translations were formed by myself a native Finnish speaker with the assist of [Microsoft language portal](https://www.microsoft.com/en-us/language/Search) and [Google Translate](https://translate.google.com/)

~~Couldn't find where to change the default placeholder for the datepicker. It should be `pp.kk.vvvv` in Finnish instead of `dd.mm.yyyy`~~


flaviendelangle:
> Couldn't find where to change the default placeholder for the datepicker. It should be pp.kk.vvvv in Finnish instead of dd.mm.yyyy

For now the placeholder is not customizable with the locale.
That's something we are discussing for the next field component.

You can manually override the placeholder passed to your input like so:

```tsx
<DatePicker
    label="Basic example"
    value={value}
    onChange={(newValue) => {
        setValue(newValue);
    }}
    renderInput={(params) => <TextField {...params} inputProps={{ ...params.inputProps, placeholder: 'pp.kk.vvvv' }} />}
/>
```

Note that it won't react to any change in the format and you have to manually take care of it.

PetroSilenius:
> For now the placeholder is not customizable with the locale. That's something we are discussing for the next field component.
> 
> You can manually override the placeholder passed to your input like so:

Right, I assumed so since I wasn't able to find anything related to it. Sounds like a good addition as other translations are handled by mui that the placeholder would be handled also👍🏻

flaviendelangle (commented):

It currently targets the `next` branch which is for the v6 alpha releases.
I will port this PR to the `master` branch to have it available on the v5 release as well.

PR #6226, opened 2022-09-20
[TimePicker] Add support for `disablePast` and `disableFuture` validation props

### LINKED ISSUE #4868: [DateTimePicker] `disablePast` is not disabling or validating time

## Current Behavior 😯

When adding disablePast to DateTimePicker, previous dates are disabled and validations kicks in when typing past date. But the time picker are not disabling past time, and there is no validation to it either.

## Expected Behavior 🤔

If current date is selected, past time should also be disabled in picker, and validation/onError should happen when typing past time.

## Steps to Reproduce 🕹

https://codesandbox.io/s/datetimepicker-disablepast-qr8re?file=/src/App.tsx

Steps:

1. The first DateTimePicker in the sandbox is displaying the expected validation error on past date
2. The second DateTimePicker is not displaying the expected validation error on past time (except for when visiting the sandbox < 5 minutes past midnight 😅)

## Context 🔦

I want to validate that the datetime is not in the past and is not before some other date (so I've already used the minDateTime for that)

## Your Environment 🌎

<details>
  
```
```
</details>

LukasTy:
Fixes #4868 

Add `disablePast` and `disableFuture` prop handling to `timeValidation` and clock picker disabled time calculation.


alexfauquette (approved):
Not a big fan of replacing everywhere
```diff
-If `true` future days are disabled.
+If `true` future selection is disabled.
```

Because it feels like you are disabling all the selections that will come next. Maybe 

```
If `true` disable values after the current time
```


By the way, I am not fully convinced it's a useful feature except for hours. because disabled minutes move fast (around one modification per minute 😉)

alexfauquette:
@flaviendelangle If you have an option on this PR?

Otherwise, I will merge after doing

```diff
-If `true` future selection is disabled.
+If `true` disable values after the current time.

-If `true` past selection is disabled.
+If `true` disable values before the current time
```

flaviendelangle:
I'll try to review it today or tomorrow, if it is not blocking you for something else :+1:

flaviendelangle:
> ```diff
> +If `true` disable values after the current time.
> ```

I agree with this diff :+1:

packages/x-date-pickers/src/internals/hooks/validation/models.ts
```diff
@@ -3,12 +3,12 @@
  */
 export interface BaseDateValidationProps<TDate> {
```
flaviendelangle, line 4: I'll create a `FutureAndPastValidationProps` to avoid using a date related interface on time components

flaviendelangle:
I added the default value to `disableFuture` and `disablePast` for time pickers and reworked the typing of the validation quite a bit because it was becoming very hard to play with.

flaviendelangle (approved):

flaviendelangle:
I will merge it after today's release

PR #6242, opened 2022-09-21
[core] Remove outdated docsearch.js dependency

oliviertassinari:
In the future, we should look into how we can have the docs-infra control it's own npm dependencies so https://github.com/mui/material-ui/pull/34421 is enough.

### DIFF
--- a/benchmark/browser/webpack.config.js
+++ b/benchmark/browser/webpack.config.js
@@ -43,11 +43,12 @@ module.exports = {
   resolve: {
     extensions: ['.js', '.ts', '.tsx'],
   },
-  target: 'web',
   plugins: [
     new HtmlWebpackPlugin({
       hash: true, // Avoid doing benchmark with cached files
       template: path.resolve(__dirname, './index.html'),
     }),
   ],
+  // TODO: 'browserslist:modern'
+  target: 'web',
 };
--- a/docs/data/data-grid/events/CatalogOfEventsNoSnap.js
+++ b/docs/data/data-grid/events/CatalogOfEventsNoSnap.js
@@ -42,16 +42,16 @@ const EventRow = ({ event }) => {
           .join(', ')})
 <DataGrid
   ${event.componentProp}={handleEvent}
-  {...other} 
+  {...other}
 />`
       : '';
 
     return `
 const handleEvent: GridEventListener<'${event.name}'> = (
   ${args.join('\n  ')}
-) => {...}    
-  
-// Imperative subscription    
+) => {...}
+
+// Imperative subscription
 apiRef.current.subscribeEvent(
   '${event.name}',
   handleEvent,
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "format-util": "^1.0.5",
     "glob-gitignore": "^1.0.14",
+    "html-webpack-plugin": "^5.5.0",
     "jsdom": "^20.0.1",
     "jss": "^10.9.2",
     "jss-plugin-template": "^10.9.2",
@@ -149,6 +150,7 @@
     "nyc": "^15.1.0",
     "playwright": "^1.21.1",
     "prettier": "^2.7.1",
+    "process": "^0.11.10",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "serve": "^14.0.1",
@@ -177,9 +179,7 @@
       "**/@mui/monorepo"
     ]
   },
-  "dependencies": {
-    "html-webpack-plugin": "^5.5.0"
-  },
+  "dependencies": {},
   "resolutions": {
     "**/react-is": "^18.2.0"
   }
--- a/test/regressions/index.js
+++ b/test/regressions/index.js
@@ -74,6 +74,12 @@ const tests = requireDocs.keys().reduce((res, path) => {
     return res;
   }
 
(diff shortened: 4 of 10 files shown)


flaviendelangle:
image

It seems like it has removed some used dependencies as a side effect

oliviertassinari:
Wow, so strange, why would remove npm dependencies changes how webpack match aliases? As far as I could push the exploration, the issue is that

```jsx
const requireDocs = require.context('docsx/data', true, /js$/);
```

Gets converted by webpack alias. As far as I know https://github.com/tleunen/babel-plugin-module-resolver isn't reading it. So it turns into:

```jsx
const requireDocs = require.context('./docs/data', true, /js$/);
```

Which is then turned to:

```jsx
const requireDocs = require.context('./node_modules/@mui/monorepo/docs', true, /js$/);
```

by the second alias. So the fix I'm proposing is to have webpack also try to resolve with:

```jsx
const requireDocs = require.context('./docs/data', true, /js$/);
```

flaviendelangle:
The webpack fail is indeed weird
I have no idea if the fix proposed risk to break something else. It seems weird to have an alias target to totally different folders, one being the target of another alias.
But if @oliviertassinari is confident enough.

benchmark/browser/webpack.config.js
```diff
@@ -43,11 +43,12 @@ module.exports = {
...
   },
-  target: 'web',
   plugins: [
     new HtmlWebpackPlugin({
       hash: true, // Avoid doing benchmark with cached files
       template: path.resolve(__dirname, './index.html'),
     }),
   ],
+  // TODO: 'browserslist:modern'
+  target: 'web',
```
oliviertassinari, line 53: Copy the mono repo

docs/data/data-grid/events/CatalogOfEventsNoSnap.js
```diff
@@ -42,16 +42,16 @@ const EventRow = ({ event }) => {
...
 />`
       : '';
 
     return `
 const handleEvent: GridEventListener<'${event.name}'> = (
   ${args.join('\n  ')}
-) => {...}    
-  
-// Imperative subscription    
+) => {...}
```
oliviertassinari, line 52: Something I found by chance, it's not ideal to have these trailing spaces when you use https://next.mui.com/x/react-data-grid/events/#catalog-of-events

Screenshot 2022-10-08 at 22 08 42

package.json
```diff
@@ -177,9 +179,7 @@
       "**/@mui/monorepo"
     ]
   },
-  "dependencies": {
-    "html-webpack-plugin": "^5.5.0"
```
oliviertassinari, line 181: Wrong location, this is a dev dependency.

package.json
```diff
@@ -149,6 +150,7 @@
     "nyc": "^15.1.0",
     "playwright": "^1.21.1",
     "prettier": "^2.7.1",
+    "process": "^0.11.10",
```
oliviertassinari, line 153: Required as we were relying on it, with a transitive dependency of docsearch.js.

test/regressions/webpack.config.js
```diff
@@ -42,4 +42,14 @@ module.exports = {
       },
     ]),
   },
+  resolve: {
+    ...webpackBaseConfig.resolve,
+    alias: {
+      ...webpackBaseConfig.resolve.alias,
+      docs: false, // Disable this alias as it creates a circual resolution loop with the docsx alias
```
oliviertassinari, line 49: Since we are not supposed to import from the docs-infra or mono-repo docs in the visual regression tests, this works as well. It's cleaner than the first attempt:

```diff
-      docs: path.resolve(__dirname, './node_modules/@mui/monorepo/docs'),
+      docs: [
+        path.resolve(__dirname, './node_modules/@mui/monorepo/docs'),
+        path.resolve(__dirname, './docs'),
+      ],
```

oliviertassinari:
I have updated the approach, see if you are happy with it.

test/regressions/index.js
```diff
@@ -74,6 +74,12 @@ const tests = requireDocs.keys().reduce((res, path) => {
     return res;
   }
 
+  // TODO: Why does webpack include a key for the absolute and relative path?
+  // We just want the relative path
+  if (!path.startsWith('./')) {
+    return res;
+  }
+
```
oliviertassinari, line 82: I have copied https://github.com/mui/material-ui/pull/28248.

test/regressions/webpack.config.js
oliviertassinari: Actually, I might not have needed to fix this, I see that it's already all broken on HEAD, e.g. https://app.circleci.com/pipelines/github/mui/mui-x/25566/workflows/5b4b75f8-3d75-4d0f-8c6e-90812dd0563a/jobs/147039

Screenshot 2022-10-09 at 02 35 11

docs/data/data-grid/events/CatalogOfEventsNoSnap.js
oliviertassinari: I think that we could enforce a rule in the CI with https://github.com/mui/material-ui/pull/34686#discussion_r991169741 to have the PRs fail if they have trailing spaces.

oliviertassinari:
The impact of the fix is now a lot more scoped than before, I have made the configuration closer to the one used in https://github.com/mui/material-ui/. Since it's not really important to get peer feedback on this YOLO, I'm moving forward 😁.

PR #6263, opened 2022-09-23
[test] Remove BrowserStack

DanailH:
Based on the previous discussion - if we remove all the browsers from the BrowserStack list should we just remove the BrowserStack condition all together?


m4theushw:
Do we need to remove BrowserStack? The latest builds are most green again, the red ones are becoming rare and they are not because a test failed.

image

DanailH:
> Do we need to remove BrowserStack? The latest builds are most green again, the red ones are becoming rare and they are not because a test failed.
> 
> image

In that case, we don't. The current setup has `firefox` disabled - the question is should we bring it back?

oliviertassinari (approved):
It sounds good, IMHO this code was equivalent to being removed the moment we decided to no longer run it on each PR.

test/karma.conf.js
```diff
@@ -136,63 +101,5 @@ module.exports = function setKarmaConfig(config) {
...
   };
 
-  let newConfig = baseConfig;
-
-  if (browserStack.enabled && browserStack.accessKey) {
-    newConfig = {
-      ...baseConfig,
-      browserStack,
-      browsers: baseConfig.browsers.concat(['chrome', 'safari', 'edge']),
-      plugins: baseConfig.plugins.concat(['karma-browserstack-launcher']),
```
oliviertassinari, line 146: A dependency to remove `karma-browserstack-launcher`

DanailH:
@m4theushw @flaviendelangle If you can have a look once you have the time it would be great.

packages/grid/x-data-grid/src/tests/cells.DataGrid.test.tsx
```diff
@@ -157,11 +157,6 @@ describe('<DataGrid /> - Cells', () => {
   });
 
   it('should throw when focusing cell without updating the state', function test() {
```
oliviertassinari, line 159:
```suggestion
  it('should throw when focusing cell without updating the state', () => {
```

packages/grid/x-data-grid/src/tests/export.DataGrid.test.tsx
```diff
@@ -38,10 +38,6 @@ describe('<DataGrid /> - Export', () => {
 
   describe('component: GridToolbar', () => {
     it('should export with the default csvOptions', async function test() {
-      // Safari 13 doesn't have Blob.text().
```
oliviertassinari, line 41: So a bug? We are supposed to support Safari 13 https://mui.com/material-ui/getting-started/supported-platforms/#browser

DanailH: I couldn't find cases where `Blob.text()` is being used.

packages/grid/x-data-grid/src/tests/cells.DataGrid.test.tsx
oliviertassinari: if we could fix the other instances too, it would be great.

Issue #6065, opened 2022-09-07
[docs] DataGrid/DataGridPro row styling problem


dmitryk-carne:
### Order ID 💳

41229

### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### The problem in depth 🔍

It looks like row styling in DataGrid/DataGridPro is now broken. We are using code from https://mui.com/x/react-data-grid/style/#styling-rows, 'row styling' section. However hovering over styled row resets the style to default grey colour, in spite of the hover style setting. The example on the https://mui.com/x/react-data-grid/style/#styling-rows, appears to be broken as well, as hovering over produces grey colour instead of lighter/darker main colour. See the screenshots. Styling example at https://mui.com/x/react-data-grid/style/#styling-rows used to work properly - showing a lighter shade of blue in the first screenshot and lighter shade of green in the second just a couple of days ago. Tried in both Firefox and Chrome.
![hoverover_error1](https://user-images.githubusercontent.com/60099087/188877125-6dc86b4c-fb61-4f16-89e4-e53e3dd909f5.png)
![hoverover_error2](https://user-images.githubusercontent.com/60099087/188877132-15111355-1919-431f-8807-b9233412d20e.png)

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  System:
    OS: Windows 10 10.0.19044
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.3.1 - C:\Program Files\nodejs\npm.CMD                                                                   
[...]

PR #6264, opened 2022-09-23
[docs] Fixes DataGrid/DataGridPro row styling issue

MBilalShafi:
Preview: https://deploy-preview-6264--material-ui-x.netlify.app/x/react-data-grid/style/#styling-rows

Issue: #6065 

Tried to see if we can do some fix by changing the order of how the classes are being used inside DataGrid, it didn't work probably due to how emotion bundles css assets. Solution proposed by @m4theushw seemed most appropriate.

**Update:** The catch to fix specificity was to use sx on Grid itself rather than Box wrapping the Grid.

### DIFF
--- a/docs/data/data-grid/style/StylingRowsGrid.js
+++ b/docs/data/data-grid/style/StylingRowsGrid.js
@@ -2,7 +2,7 @@ import * as React from 'react';
 import Box from '@mui/material/Box';
 import { DataGrid } from '@mui/x-data-grid';
 import { useDemoData } from '@mui/x-data-grid-generator';
-import { darken, lighten } from '@mui/material/styles';
+import { darken, lighten, styled } from '@mui/material/styles';
 
 const getBgColor = (color, mode) =>
   mode === 'dark' ? darken(color, 0.7) : lighten(color, 0.7);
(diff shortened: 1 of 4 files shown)


m4theushw (commented):
I noticed another difference between https://v4.mui.com/components/data-grid/style/#styling-rows. A selected row can't be distinguished from an unselected in dark mode. It's the same problem of specificity. IMO we could go further and provide a custom background color for selected rows that also match the background of the unselected state, e.g. unselected red row when selected changes to dark red and so on.

msedge_HNueiNIitZ

MBilalShafi:
Here's the latest look of demo:
gridUpdated

docs/data/data-grid/style/StylingRowsGrid.js
```diff
@@ -2,66 +2,110 @@ import * as React from 'react';
...
 
-const getBackgroundColor = (color, mode) =>
+const getBgColor = (color, mode) =>
+  mode === 'dark' ? darken(color, 0.7) : lighten(color, 0.7);
+
+const getHoverBgColor = (color, mode) =>
   mode === 'dark' ? darken(color, 0.6) : lighten(color, 0.6);
 
-const getHoverBackgroundColor = (color, mode) =>
+const getSelectedBgColor = (color, mode) =>
```
DanailH, line 13:
```suggestion
const getSelectedBackgroundColor = (color, mode) =>
```

docs/data/data-grid/style/StylingRowsGrid.js
```diff
@@ -2,66 +2,110 @@ import * as React from 'react';
...
   mode === 'dark' ? darken(color, 0.5) : lighten(color, 0.5);
 
+const getSelectedHoverBgColor = (color, mode) =>
```
DanailH, line 16:
```suggestion
const getSelectedHoverBackgroundColor = (color, mode) =>
```

docs/data/data-grid/style/StylingRowsGrid.js
```diff
@@ -2,66 +2,110 @@ import * as React from 'react';
 import Box from '@mui/material/Box';
 import { DataGrid } from '@mui/x-data-grid';
 import { useDemoData } from '@mui/x-data-grid-generator';
-import { darken, lighten } from '@mui/material/styles';
+import { darken, lighten, styled } from '@mui/material/styles';
 
-const getBackgroundColor = (color, mode) =>
+const getBgColor = (color, mode) =>
```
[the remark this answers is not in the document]
MBilalShafi: Variable name `getSelectedHoverBackgroundColor` seemed very long that is why I tried to shorten a bit, anyhow, confusion with locale makes sense. Renamed the functions.

DanailH (approved):
Awsome 🚀

Issue #5090, opened 2022-06-02
[pickers] Add tests around date disabled state


flaviendelangle:
Follow up on #5081 

We should test in depth for all components the `disablePast`, `disableFuture`, `minDate`, `maxDate`, ... props

flaviendelangle:
Should probably be splitted in several tasks

- Test them on the fields
- Test them on each view (already did some tests for some)
- Test them on the pickers (how to avoid duplicating 10 times each test ? Maybe we need something like `describeConformance`)

alexfauquette:
Closing this one in favor of smaller issues about remaining aspect to test: #6552 #6551

PR #6265, opened 2022-09-23
[core] Test validation on textfield and date views

alexfauquette:
Part of #5090
Interface tested for validation

- [x] text field
- [x] day picker
- [x] month picker
- [x] year picker
- [ ] hours
- [ ] minutes


flaviendelangle:
Should we also use the same `describe` on pickers ?
Or should it be another one to check both the field and the views ?

alexfauquette:
I was looking for testing hours/minutes. But the behavior is already tested in text field, And if we do deep modification on the clock component, maybe better to add the tests latter

packages/x-date-pickers/src/CalendarPicker/CalendarPicker.test.tsx
```diff
@@ -13,7 +13,7 @@ import {
   wrapPickerMount,
   createPickerRenderer,
 } from 'test/utils/pickers-utils';
-import describeValidation from '@mui/x-date-pickers/internals/utils/describeValidation';
+import describeValidation from '@mui/x-date-pickers/tests/describeValidation';
```
flaviendelangle, line 16: Should it be outside of the package ? 
In the `test/utils` folder ?

alexfauquette: Don't think so, because it's tests related to the package. testValidation will probably neither be used elsewhere

flaviendelangle: The core had the pickers test in `test/utils` but that's of course something we can reconsider.
If we want to put `describeValidation` where you put it, mayber the `pickerUtils` should go there two i the future

alexfauquette:
@flavien I added tests to the `TimeField` and `DateTimeField`. I copied the behavior of `maxDateTime` and `minDateTime` from the `DateTimePicker`

packages/x-date-pickers/src/DateTimeField/useDateTimeField.ts
```diff
@@ -36,8 +36,11 @@ const useDefaultizedDateTimeField = <TDate, AdditionalProps extends {}>(
     disablePast: props.disablePast ?? false,
     disableFuture: props.disableFuture ?? false,
     format: props.format ?? utils.formats.keyboardDateTime,
-    minDate: applyDefaultDate(utils, props.minDate, defaultDates.minDate),
-    maxDate: applyDefaultDate(utils, props.maxDate, defaultDates.maxDate),
+    disableIgnoringDatePartForTimeValidation: Boolean(props.minDateTime || props.maxDateTime),
```
flaviendelangle, line 39: For v6 I wanted to remove this prop from the public signature and move the `minTime: props.minDateTime ?? props.minTime` logic inside the validation.

But maybe we can temporarily add it here

packages/x-date-pickers/src/TimeField/tests/TimeField.validation.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+describe('<TimeField /> validation', () => {
+  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+
+  describeValidation(TimeField, () => ({
+    render,
+    clock,
+    views: ['hours', 'minutes'],
+    skip: ['day', 'month', 'year'],
+    ignoredProps: ['minDate', 'maxDate', 'shouldDisableDate'],
+    isFieldComponent: true,
```
flaviendelangle, line 14: Maybe do the opposite with a `isLegacyPicker` :angel:

packages/x-date-pickers/src/YearPicker/YearPicker.test.tsx
```diff
@@ -4,9 +4,21 @@ import { expect } from 'chai';
...
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+  });
+
+  describeValidation(YearPicker, () => ({
+    render,
+    clock,
+    views: ['year'],
+    skip: ['textField'],
+    overrideDefault: { renderInput: undefined },
```
flaviendelangle, line 19: Unused property

If you could type the params

flaviendelangle (approved):

PR #6299, opened 2022-09-27
[DatePicker] Allows to fix the number of week displayed

### LINKED ISSUE #5387: [DatePicker] Add prop showFixedNumberOfWeeks

alexfauquette:
Fix #5387

For simplicity, I encourage a review commit per commit for this PR 😉

I used a number, such that if other calendars require a different number of weeks they could customize it. Of course, switching to a boolean is not a big deal if you prefer.

I think the documentation of week number (#6144) could be added in the same section


flaviendelangle (commented):
@joserodolfofreitas if you have a preference between a number and a boolean
Maybe we could start with the number, and if people ask for something more automatic, add an `auto` value which takes the maximum amount of weeks per month in the year.

flaviendelangle:
@joserodolfofreitas if you have a minute to check my comment above

joserodolfofreitas:
Yeah, I agree that using numbers makes sense here. 
"auto" also sounds like a good idea if there's a demand for the implementation.

PR #6304, opened 2022-09-27
[CalendarPicker] Prevent getting focus when `autoFocus=false`

alexfauquette:
In [DatePicker doc page](https://mui.com/x/react-date-pickers/date-picker/) the focus is caught by the year picker.

This PR should fix this bug

### DIFF
--- a/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.tsx
@@ -437,8 +437,13 @@ export const CalendarPicker = React.forwardRef(function CalendarPicker<TDate>(
     },
   );
 
+  const prevOpenViewRef = React.useRef(openView);
   React.useEffect(() => {
     // Set focus to the button when switching from a view to another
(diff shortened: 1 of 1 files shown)


flaviendelangle (approved):
Nice fix

I did not test it, but what is preventing us from doing this update in the render like we do in `usePickerState` ?
React tends to recommend that approach over an effect: https://fr.reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

alexfauquette:
@flaviendelangle I don't get to which part of `usePickerState` you are referring to

alexfauquette:
It's not possible because it updates multiple components, and apparently, you should not update your parents during the render phase. Only inside effect. 

I assume the proposed technic to derive state from props only works if the state is limited to the current component

flaviendelangle:
> It's not possible because it updates multiple components, and apparently, you should not update your parents during the render phase. Only inside effect.

Make sense :+1:

flaviendelangle (commented):
OK to merge it as is :+1:

PR #6309, opened 2022-09-27
[DatePicker] Open in a valid month

### LINKED ISSUE #5091: [pickers] Default calendar month does not take `minDate` and `maxDate` into account when `value = null`
### Duplicates

### Latest version

### Current behavior 😯

When `minDate` is set to a future date (E.g newDate('2033')) and `value` is set as `null`, when you **open the date picker** the calendar will **render the today date**, instead of moving the calendar to the `minDate`.

### Expected behavior 🤔

Calendar should render the `minDate` instead of the **today date**.

### Steps to reproduce 🕹

Steps:

1. Render DatePicker
2. Set value as null
3. Set minDate as future date (E.g new Date('2033')
4. Open the DatePicker
5. The calendar will render today date as initial view

You can find a codesanbox example [here](https://codesandbox.io/s/basicdatepicker-material-demo-forked-f6tf97?file=/demo.js) 

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
  System:
    OS: Windows 10 10.0.22000
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - ~\AppData\Roaming\npm\yarn.CMD
  Browsers:
    Chrome: 101.0.4951.67
    Edge: Spartan (44.22000.120.0), Chromium (102.0.1245.30)
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #5091

### DIFF
--- a/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.test.tsx
+++ b/packages/x-date-pickers/src/CalendarPicker/CalendarPicker.test.tsx
@@ -211,6 +211,32 @@ describe('<CalendarPicker />', () => {
         adapterToUse.date(new Date(2018, 0, 2, 11, 11, 11)),
       );
     });
+
+    it('should open after `minDate` if now is outside', () => {
+      render(
+        <CalendarPicker
+          value={null}
+          onChange={() => {}}
+          view="day"
+          minDate={adapterToUse.date(new Date(2031, 2, 3))}
+        />,
+      );
+
+      expect(screen.getByText('March 2031')).not.to.equal(null);
+    });
+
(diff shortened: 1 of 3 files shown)


flaviendelangle (commented):
This approach would not support the other validation prop no ?
For `disableFuture` and `disablePast`, it's useless since we are using `now`.
For the others, that might be more complicated, maybe in a follow up ?

alexfauquette:
Not sure it should take into account the `disableDate`. For example you are booking a camping for this summer. `minDate` is June, so you expect to see the calendar open to June, and if all the dates are disabled you know the camping is full for this month.

Issue #4547, opened 2021-12-13
[pickers] New component: DateTimeRangePicker


rodrigofariow:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

Not sure how this should be implemented according to the material design specs exactly but I assume you guys have a better and more conceivable idea than me. 😁

### Motivation 🔦

My company has recently bought the MUI X PRO license and despite it's clear benefit (specially `DataGridPro`) we were expecting a more complete `DateRangePicker`.
~~**Note that we are aware that it's still in the lab** but~~ I think there is a real need to have the option to select the time along with dates to formulate a date range.

A common use case is to define a period of time where some type of user events have occured. Be it activity reports from a user, be it the amount of money the user has spend during that exact period. There are endless possibilities.

Is this something that is on your roadmap? I haven't found any info regarding this question.

Thank you! :) 

### Benchmark

- Ant Design, for instance, has support for this -> 
![image](https://user-images.githubusercontent.com/22502084/145814399-7fc1de01-1f45-4851-a64a-7eb27b413f4c.png)
- More: https://mui-org.notion.site/Date-Time-Range-Picker-component-7a7cc24a3ba34ad0a04ebf6458eb5c44 (public) / https://www.notion.site/mui-org/Date-Time-Range-Picker-component-7a7cc24a3ba34ad0a04ebf6458eb5c44 (private)

### Other requests

- https://www.notion.so/mui-org/Sharechat-Prateek-Amar-2aee32660fb044df8bf0337f4f357feb#e09c842ed3da47409bfe7257e5b7dac5
- https://mui.zendesk.com/agent/tickets/7124

fdcolongonzalez:
Is there an ETA for this new feature?

hoa00:
upvote for this feature 👍

lawcia:
👍 would also like this

ltphuoc:
👍 upvote

trunglvse151138:
would like too see this soon 🔥🔥

CMeyer19:
Would love to have this feature

adarsh-drishya:
please have this feature

purefun:
👍 would also like this

kingkong0828:
would like too see this soon

Jfish-mastt:
👍  We would like this also

piyushroongta:
would like  have this

WinmezzZ:
+1

[...]

Issue #4460, opened 2021-11-17
[pickers] New component: TimeRangePicker


ErnestBrandi:
## Benchmarks

- https://demo.mobiscroll.com/jquery/range/time-range#

am-vishal:
we need time range picker in @mui something like this.
I found many people facing similar issues.
https://stackoverflow.com/questions/72811784/from-time-and-to-time-validation-in-react-dropdown-react-datepicker-hour

simon-guider:
+1

tfletch5:
+1

Abdelrahman-Rabi:
+1

a-ahandani:
+1

WinmezzZ:
+1

filipstojakovic:
+1

davidsl4:
+1

submetu:
+1

bulatkhayrullin:
+1

gcardacci:
+1

NonozgYtb:
+1

itseasy21:
+1

flaviendelangle:
Small update here.
We know that a lot of people are waiting for this feature and it remains one of our top priority.

The team is currently rebuilding the input part of the pickers from scratch to provide a better experience.
We should be able to soon release a first documented (but still unstable) version of those components soon.
You can find a basic internal documentation of those components [here](https://next.mui.com/x/react-date-pickers/date-field) if you are curious.

The plan is then to create new keyboard-only components from *time range* and *date time range* in the coming weeks.
And then to build new pickers around them, just like we plan to migrate the already existing pickers to those new components.

In summary the `TimeRangePicker` will come in two steps:

1. It's keyboard-only component
2. It's full picker with the calendar views

flaviendelangle:
As mentioned a few days ago, we now have a new component for date time range editing called `MultiInputTimeRangeField` (very long name, but a `SingleInputTimeRangeField` is in the pipeline).

The documentation is available [here](https://next.mui.com/x/react-date-pickers/time-range-field/)
It is still very early-stage but will serve as the baseline for the future `TimeRangePicker`.
So the more feedback we have to improve it, the easier it will be to build this new picker.

mahouha:
+1

DvirMalka:
+1

dreemorra:
+1

FlorianCassayre:
Guys, be aware that this feature will only be made available in the paid plan, not the "community" version.

[...]

PR #6312, opened 2022-09-27
[pickers] New components `TimeField` and `DateTimeField`

### LINKED ISSUE #6062: [pickers] Support time validation on the `DateField` component
For now, the `useDateField` hook uses the `validateDate` method.
We need to support the time validation.
But if we just use the `validateDateTime`, the date will be validated even when users use `DateField` with a time-only format and the time will be validated even when users use `DateField` with a date-only format.

I see two options:

1. We expose 3 components `DateField` / `DateTimeField` / `TimeField` with some analysis of the format to prevent users from passing a format that contains time in `DateField` or date in `TimeField`.
2. We analyze the format and pass a dynamic validator based on the section the date is currently containing

flaviendelangle:
Closes #6062
[Doc preview](https://deploy-preview-6312--material-ui-x.netlify.app/x/react-date-pickers/fields/) (not really a doc, just an hidden example to help testing)

Most of the logic is duplicated from `DateField` but I don't think it's worth creating additional attractions yet.

- [x] Add a `supportedDateSections` param to `useField` and warn when a section is not contained in this list
- [x] New component `TimeField`
- [x] New component `DateTimeField`
- [x] Test the time sections with `TimeField` (I'm not testing `DateTimeField` now, we could add basic tests but I would not re-test every behavior from the other two) 
- [x] Fix a few typos I saw during the duplication of `DateField`

## Follow up

- Create one doc page per field (probably keep each pair of single and multi range fields in the same page, for instance a page *Date range field* for `SingleInputDateRangeField` and `MultiInputDateRangeField`)
- Create a home page in the doc for the field to display some common content. 
- Create `MultiInputDateTimeRangeField` and `MultiInputTimeRangeField` to show that we are building the bricks to support #4547 and #4460.
- Create `SingleInputDateTimeRangeField` and `SingleInputTimeRangeField` (maybe less of a priority).

## Testing

@alexfauquette it would be great if #6265 was compatible with the fields at some point.
If they are, we could do almost test only `DateField` / `SingleInputDateRangeField` and `MultiInputDateRangeField`.
Since most of the difference between those three and the others is validation.


docs/data/date-pickers/date-field/CustomDateFormat.tsx
```diff
@@ -29,18 +29,6 @@ export default function CustomDateFormat() {
           onChange={(newValue) => setValue(newValue)}
           format="DD MMMM YYYY"
         />
-        <DateField
```
flaviendelangle, line 32: I will add the new doc pages in a follow up PR to avoid making this one to big

packages/x-date-pickers/src/DateField/useDateField.ts
```diff
@@ -47,9 +47,10 @@ const useDefaultizedDateField = <TDate, AdditionalProps extends {}>(
   const defaultDates = useDefaultDates<TDate>();
 
   return {
-    disablePast: false,
-    disableFuture: false,
     ...props,
+    disablePast: props.disablePast ?? false,
```
flaviendelangle, line 51: The default value was not applied when the hook received `disablePast: undefined` (same for the others).
Which for some fields is the case depending on the component implementation.

packages/x-date-pickers/src/internals/hooks/validation/useDateValidation.ts
```diff
@@ -16,12 +16,6 @@ export interface DateComponentValidationProps<TDate>
     YearValidationProps<TDate>,
     Required<BaseDateValidationProps<TDate>> {}
 
-export interface DateComponentDefaultizedValidationProps<TDate>
```
flaviendelangle, line 19: This interface was 100% equivalent to `DateComponentValidationProps` above
Maybe it will be interesting to have the defaultized and non defaultized interfaces in the future but for now it was just confusing

alexfauquette:
For the DateField validation tests are passing 🎉

I have to write the range pickers first to test single and multi-input field components. But it should not be a problem

flaviendelangle:
> For the DateField validation tests are passing tada
> 
> I have to write the range pickers first to test single and multi-input field components. But it should not be a problem

:partying_face: 
We will be able to test `DateTimeField` and `TimeField` as well
That will cover most of the logic.

packages/x-date-pickers/src/DateField/useDateField.ts
alexfauquette: Should we add test in validation to check the default values are applied?

alexfauquette (commented):
Looks good, their is just an edge case about hours and am/pm.

If you press 2 multiple times, it switches between 2AM and 10PM because the component does not take into account that 'am-pm' section implies that hours are between 0 and 12

https://user-images.githubusercontent.com/45398769/193023736-d0e7bd3f-4019-4369-9703-13dc2031b5b5.mp4

flaviendelangle:
> If you press 2 multiple times, it switches between 2AM and 10PM because the component does not take into account that 'am-pm' section implies that hours are between 0 and 12

That was already the case when the time editing was in `DateField` so I think we can merge and fix later.
But indeed in `getSectionValueNumericBoundaries` we should return 12 for some hour format.
This could be an information stored in the adapter like the content type in #5958

packages/x-date-pickers/src/DateField/useDateField.ts
flaviendelangle: Could it be done in your `describeValidation` ?

alexfauquette (approved):

PR #6323, opened 2022-09-29
[docs] Create first publishable version of the field doc

flaviendelangle:
[Doc preview](https://deploy-preview-6323--material-ui-x.netlify.app/x/react-date-pickers/fields/)


docs/data/date-pickers/fields/fields.md
```diff
@@ -7,11 +7,82 @@ title: React Fields components
...
+
+New pickers using them will be released in a few months.
+
+## Form props
+
+The standard form attributes (`disabled`, `readOnly`) are supported by all the field components:
+
+{{"demo": "FormProps.js", "defaultCodeOpen": false}}
+
+## Advanced
```
flaviendelangle, line 40: The sections below should probably move to pages not specific to the fields to make them easier to find when the field are stable

But for now I propose to keep the field content inside the scope of the fields, since they are unstable.

docs/data/date-pickers/fields/DateRangeFieldExamples.tsx
```diff
@@ -35,19 +32,10 @@ const GridItem = ({
...
-          <DateField defaultValue={date1} />
-        </GridItem>
-        <GridItem label="TimeField">
-          <TimeField defaultValue={date1} />
-        </GridItem>
-        <GridItem label="DateTimeField">
-          <DateTimeField defaultValue={date1} />
-        </GridItem>
         <GridItem label="SingleInputDateRangeField">
           <SingleInputDateRangeField defaultValue={[date1, date2]} />
```
LukasTy, line 40: Offtopic: Did we think about how are we going to handle `localeText.start` and `localeText.end` application to this and `MultiInput` components?
Currently both of these are not using those fields. 🤔

docs/data/date-pickers/fields/SingleDateFieldExamples.tsx
```diff
@@ -0,0 +1,51 @@
...
+  );
+};
+
+const date1 = dayjs('2022-04-07T14:30:22');
+
+export default function SingleDateFieldExamples() {
+  return (
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <Grid container spacing={4} width={550}>
+        <GridItem label="DateField">
```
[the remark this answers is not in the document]
flaviendelangle: I aligned this demo with the `DateRangeFieldExamples` just below.
Do you think we should only use this external label system when one of the components in the demo has 2 inputs ?

docs/data/date-pickers/fields/DateRangeFieldExamples.tsx
flaviendelangle: Yes that is something we must support in the future indeed.
It would require wrapping the component in `LocalizationProvider` to forward the `localeText` prop to allow the override.
And then use the locale text to set the label.

But I propose to do it after #6392 to avoid conflicts.

docs/data/pages.ts
```diff
@@ -169,6 +169,25 @@ const pages: MuiPage[] = [
...
+          '/x/react-date-pickers/date-time-field',
+          '/x/react-date-pickers/time-field',
+        ],
+        newFeature: true,
+        children: [
+          { pathname: '/x/react-date-pickers/fields', title: 'Getting started' },
+          { pathname: '/x/react-date-pickers/date-field' },
+          { pathname: '/x/react-date-pickers/date-range-field', plan: 'pro' },
+          { pathname: '/x/react-date-pickers/date-time-field' },
+          { pathname: '/x/react-date-pickers/time-field' },
```
alexfauquette, line 188: For me it makes more sense to group the docs by community first, pro second such that it becomes easier to distinguish free and paid components.

And sort than by date/time/dateTime because it's like two atoms and their combination
```suggestion
          { pathname: '/x/react-date-pickers/time-field' },
          { pathname: '/x/react-date-pickers/date-time-field' },
          { pathname: '/x/react-date-pickers/date-range-field', plan: 'pro' },
```

docs/data/date-pickers/fields/fields.md
alexfauquette: Yes, and it's the kind of stuf you want people to be aware of when they start using the component

alexfauquette (approved):

docs/data/date-pickers/fields/fields.md
```diff
@@ -7,11 +7,82 @@ title: React Fields components
...
+These components are unstable.
+We might do some breaking change on their props to have the best components possible by the time of the stable release.
 :::
 
-{{"demo": "FieldExamples.js", "bg": "inline", "defaultCodeOpen": false}}
+## Introduction
 
-- [DateField](/x/react-date-pickers/date-field/)
-- [DateRangeField](/x/react-date-pickers/date-range-field/)
+The fields are React components that lets you enter a date with the keyboard, without any modal or dropdown UI.
```
joserodolfofreitas, line 16:
```suggestion
The fields are React components that let you enter a date or time with the keyboard, without using any modal or dropdown UI.
They provide refined navigation through arrow keys and support advanced behaviors like localization and validation.
```

docs/data/date-pickers/fields/fields.md
```diff
@@ -7,11 +7,82 @@ title: React Fields components
...
+
+:::warning
+You need to make sure the input is focused before imperatively updating the selected sections.
+:::
+
+{{"demo": "ControlledSelectedSections.js", "defaultCodeOpen": false }}
+
+:::warning
+For range fields, you will not be able to use the string format since each section is present twice.
+We will add new apis in the future to better support this use case.
```
joserodolfofreitas, line 87:
```suggestion
We'll add new APIs in the future to better support this use case.
```

joserodolfofreitas (commented):
Great work!
I really reviewed only the getting started page, but it's looking solid overall.
And the `new` badge is indeed very charming 👌.
I missed a section about the headless approach, though. Something like we had on the first [preview page](https://next.mui.com/x/react-date-pickers/date-field/#headless-usage).
Curious if there's a reason on why it was left out.

flaviendelangle:
> I missed a section about the headless approach, though. Something like we had on the first [preview page](https://next.mui.com/x/react-date-pickers/date-field/#headless-usage).

I removed it to avoid poorly documenting a very broad an important topic on this first version.
That's something that would, I think, be worth a dedicated page with when to use headless fields, when to use headless input, examples with Joy, examples with Browser input / Tailwind inputs, etc...

And this page should probably be outside of the scope of the field, because it would also be applicable for pickers (that would even be the main use case for most users).
So probably in *Date & Time Pickers* => *Customization* => *Usage without Material UI* or something similar.

joserodolfofreitas:
Yeah, I agree it must be well documented, But I thought it would be useful to have something to link in the Blog post. Maybe a mini section advising to wait for the full documentation if you intend to use it in an application.

flaviendelangle:
@joserodolfofreitas 

Would a live example inside the blog post with a `DateField` using `Joy` work for you ?
With a disclaimer saying that this part of the doc will come in the following weeks.

joserodolfofreitas:
> Would a live example inside the blog post with a `DateField` using `Joy` work for you ? With a disclaimer saying that this part of the doc will come in the following weeks.

That'd be great! I'm considering publishing the post today with the current basic section link, and patching it later when we have the links for the migration guide and the headless approach.

docs/data/date-pickers/fields/fields.md
oliviertassinari: I think that it would be better without the use of "we" entirely on the page.

```suggestion
New APIs will be added in the future to better support this use case.
```

docs/data/date-pickers/fields/fields.md
```diff
@@ -7,11 +7,82 @@ title: React Fields components
...
+4. If `null` is provided, no section will be selected
+
+:::warning
+You need to make sure the input is focused before imperatively updating the selected sections.
+:::
+
+{{"demo": "ControlledSelectedSections.js", "defaultCodeOpen": false }}
+
+:::warning
+For range fields, you will not be able to use the string format since each section is present twice.
```
oliviertassinari, line 86: I believe we discouraging the use of "will".

```suggestion
For range fields, you can't use the string format since each section is present twice.
```

docs/data/date-pickers/fields/fields.md
```diff
@@ -7,11 +7,82 @@ title: React Fields components
 <p class="description">The fields components let the user select a date / date range with the keyboard.</p>
 
 :::warning
-These components are in a very early stage.
-They should not be used in a production setup.
+These components are unstable.
+We might do some breaking change on their props to have the best components possible by the time of the stable release.
```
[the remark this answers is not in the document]
oliviertassinari:
```suggestion
They might receive breaking changes on their props to have the best components possible by the time of the stable release.
```

docs/data/date-pickers/fields/fields.md
joserodolfofreitas: Regarding the "We" rule, It seems there's often a paradox in the writing style.
By not using "we", the chances are that we'll end up in passive voice, which is also discouraged.

In this case, I think the passive voice sounds preferred indeed.

oliviertassinari: Very true

@samuelsycamore what's the least worse option between using the passive voice and we?

docs/data/date-pickers/fields/SingleDateFieldExamples.tsx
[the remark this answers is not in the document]
LukasTy: IMHO, I'd say we should go with the style we are going to use eventually.
For DateField components it's possible already—so, why not? 😃 
For DateRangeFields—yeah, we could temporarily stick with label-on-top.

docs/data/pages.ts
LukasTy: Great observation, I would agree with this suggestion. 👍 
DateField -> TimeField -> DateTimeField -> DateRangeField -> ...

flaviendelangle: Fine for me
As long as we keep the same order everywhere

LukasTy:
Nitpick: I'd probably go with `fullWidth` on both components to avoid data clipping on this demo.
Screenshot 2022-10-11 at 09 38 36

LukasTy (approved):

docs/data/date-pickers/fields/fields.md
flaviendelangle: Waiting for Sam feedback for this one

docs/data/pages.ts
flaviendelangle: Done

flaviendelangle:
I added the time range fields and date time range fields doc
The PR has also the diff from those components, you can ignore it, it should be merged soon.

docs/data/date-pickers/fields/SingleDateFieldExamples.tsx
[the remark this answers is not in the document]
LukasTy: Oh, sorry, I missed the fact that we applied similar styling to Validation page as well.
It might have slipped through my eyes, but was there any specific reason to use labels on top of input elements instead of `label` attribute on the picker components like in most other pages?

This is the difference I'm talking about and the bigger portion of our doc uses the latter (like disabled and readonly) style.
Screenshot 2022-10-12 at 09 23 30

flaviendelangle:
@LukasTy 

> Nitpick: I'd probably go with fullWidth on both components to avoid data clipping on this demo.

Full width on the field ? It does not seem to change anything on my side.
Can we switch to a `direction="column" on small screens ?

LukasTy:
> Full width on the field ? It does not seem to change anything on my side. Can we switch to a `direction="column" on small screens ?

Sorry, forgot to mention the need for `flexGrow={1}` on the wrapping `Stack` component.

Indeed, having some sort of responsive support would go a long way, but we would still need a combination of solution to not clip data on wider screens as well as fit gracefully on mobile. 🤔 

On another note, I remember us discussing that documentation support on mobile is not our priority as we currently also have quite a few problems with it, but majority of developers reading docs will probably not be using mobile for it. 🙈

LukasTy:
There also seems to be sort of an issue with side menu configuration for the last two menu entries as seen in the video.

https://user-images.githubusercontent.com/4941090/195523556-3ae3008e-f66b-4f87-914d-09f97c6cf77b.mov

flaviendelangle:
> Indeed, having some sort of responsive support would go a long way, but we would still need a combination of solution to not clip data on wider screens as well as fit gracefully on mobile. thinking

I added some breakpoints to clean this, can you check if it's ok for you ? (CI in progress)
Maybe we could create a centralized layout component for the pickers as our use case are often very similar.

flaviendelangle:
> There also seems to be sort of an issue with side menu configuration for the last two menu entries as seen in the video.

That one is easy to fix, I forgot to add them to the scope of the folder

flaviendelangle:
For `DateTimeRangeField`, I need the multi input range fields to be able to switch to column mode, which is not handled right now.
It would be a nice issue for a follow up.

flaviendelangle:
I'm merging it, we can do a follow up for the opened comment

Issue #6315, opened 2022-09-28
[data grid] Allow control default focus on `GridColumnsPanel`


scott-avery:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

provide an option to decide whether default focus on search input field inside `GridColumnsPanel`

### Motivation 🔦

We are using mui x datagrid in our projects. And some of our users will use tablets like surface go.

If the user try to control column visibility, it will trigger the input panel on touch devices which is annoying.

PR #6336, opened 2022-09-30
[DataGrid] Allow to disable autofocusing search field in columns panel

scott-avery:
Fixes mui/mui-x#6315


scott-avery:
closing the PR as it violates the company policy.  Will do it again with another account.

PR #6339, opened 2022-09-30
[pickers] Support meridiem edition (WIP)

flaviendelangle:


packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -232,8 +232,23 @@ export const adjustInvalidDateSectionValue = <TDate, TSection extends FieldSecti
...
 
-export const getSectionVisibleValue = (section: Omit<FieldSection, 'start' | 'end'>) =>
-  section.value || section.emptyValue;
+export const getSectionVisibleValue = (
+  section: Omit<FieldSection, 'start' | 'end'>,
+  willBeRenderedInInput: boolean,
+) => {
+  const value = section.value || section.emptyValue;
+
+  // In the input, we add an empty character at the end of each section without trailing zeros.
```
flaviendelangle, line 241: Thread if you want to discuss this behavior

LukasTy:
> Do not parse the day value in the setSectionValueOnDate of digit editing.


https://user-images.githubusercontent.com/4941090/194046244-d6ea4dbc-daea-4840-8cbd-6d220abd7b90.mov

I find this behaviour somewhat strange.
Firstly, there is no error indication, that the currently entered value is erroneous (there is in current implementation).
Secondly, after we finally enter a valid date - we can no longer make it invalid even manually.

At minimum, I'd expect the error state to be triggered while the date is invalid. 🤔

flaviendelangle:
@LukasTy indeed we probably lack an error when the date becomes completed but invalid.

LukasTy:
> Could you expand on this, I did not understand

I meant that date selection behaviour differs depending on the fact if the date is valid or not.
I'm not saying that we should allow incorrect date entry—was just pointing out, that based on how the field currently behaves—it would seem very inconsistent.

On another note—I did test how Telerik handles it—it seems that they are parsing incomplete date (possibly setting todays year/month) to disallow invalid day selection in the first place. Have you had any thoughts about such behaviour as it would make the component easier to use and mostly correct (excluding leap years).

flaviendelangle:
>  (possibly setting todays year/month) to disallow invalid day selection in the first place

It was what I did before
But when today's date is in a 30-day month, if the user types 31 it crashes.
This is the bug that caused the reworked and fixing it seems to have introduced regressions.

For me it is clear that we need to fix the bug.
But if we could fix it and keep the same level on validation on the other scenarios, that's of course better.


> I meant that date selection behaviour differs depending on the fact if the date is valid or not.

Could you please give me concrete reproduction case because I struggle to understand what you really mean.

I noticed one inconsistency which is if you have a filled field and you empty it, you still can't type 31 if the previous date was in a 30-day month (probably because of the reference date, I'll have a look).

LukasTy:
> Could you please give me concrete reproduction case because I struggle to understand what you really mean.
What is an invalid day ? Something above 31 ? 31 on a 30-day month (or February).

Yes, that's exactly what I meant. While you are entering a new date—any date seems to be fine (visually), but after a valid date is accepted—only another valid date might be entered both manually (numbers) and by incrementing/decrementing with arrows.
I wouldn't get too hang up on this point I brought up, it wouldn't matter if the component would give a better visual feedback of when input is invalid. 👌

LukasTy (commented):
Currently both single and multi input range fields have a bit weird error state behaviour.
- Multi input fields never display error state (both when the date is invalid or when end date exceeds start date)
- Single input field displays error only when both start and end values are not empty.

Are these known issues we are keeping in for now?

https://user-images.githubusercontent.com/4941090/194277890-bc22251a-8647-4a8a-a1f4-e33548c35a5d.mov

flaviendelangle:
> Are these known issues we are keeping in for now?

Yes, I will create an issue.
I agree that the validation of the range fields is very bad right now.

PR #6364, opened 2022-10-03
[pickers] Support pasting on fields

flaviendelangle:


packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -111,23 +112,40 @@ export const useField = <
...
+    if (readOnly) {
       return;
     }
 
-    const prevValueStr = fieldValueManager.getValueStrFromSections(state.sections);
     const valueStr = event.target.value;
 
+    // If no section is selected, we just try to parse the new value
+    // This line is mostly triggered by imperative code / application tests.
+    if (selectedSectionIndexes == null) {
```
flaviendelangle, line 142: Little bonus not directly related to pasting but using the same logic

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -110,23 +111,40 @@ export const useField = <
...
+      return;
+    }
+
+    // TODO: Implement single section paste
+    if (
+      selectedSectionIndexes &&
+      selectedSectionIndexes.startIndex === selectedSectionIndexes.endIndex
+    ) {
       return;
     }
```
alexfauquette, line 126: By the way not using `event.preventDefault()` can do the job. But that's for an other PR

packages/x-date-pickers/src/DateField/tests/editing.DateField.test.tsx
```diff
@@ -510,6 +510,65 @@ describe('<DateField /> - Editing', () => {
...
+      render(<DateField onChange={onChange} readOnly />);
+      const input = screen.getByRole('textbox');
+      clickOnInput(input, 1);
+
+      // Select all sections
+      userEvent.keyPress(input, { key: 'a', ctrlKey: true });
+
+      firePasteEvent(input, '09/16/2022');
+      expect(onChange.callCount).to.equal(0);
+    });
```
alexfauquette, line 569: Would be nice to also add a test about what append when the past value is wrong

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
flaviendelangle: For a valid date we would still have to update the parsed date and fire `onChange` no ?

packages/x-date-pickers/src/DateField/tests/editing.DateField.test.tsx
flaviendelangle: Done :+1:

alexfauquette (approved):

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
alexfauquette: I will open a PR to test that when this one is merged

PR #6369, opened 2022-10-03
[docs] Group picker pages

flaviendelangle:
Part of #5565
Doc preview: https://deploy-preview-6369--material-ui-x.netlify.app/x/react-date-pickers/date-picker/

## Follow up

- Create one page per "Calendar" component


samuelsycamore:
> I went for `Picker components` as the title of the group instead of `Pickers` to be more explicit.

Why not simply "Components" for this section? This would match the pattern of the Core docs.

Also, given our recent discussions about formatting component names, I would expect these to be capitalized like proper nouns: Date Picker, Date Range Picker, etc.

flaviendelangle:
@samuelsycamore another question if we do that: how do we avoid confusion between "Components" and "Custom components" ?

samuelsycamore:
> @samuelsycamore another question if we do that: how do we avoid confusion between "Components" and "Custom components" ?

I think the "Custom components" page should be renamed 😛 to make it more clear what's in that doc. I'm actually planning to create docs on this same info for the Core libraries, so I'll add that X page to that project.

As for the main topic, I see your point about going too deep with the groupings. I think the way you have it arranged here works well. 👍

joserodolfofreitas (commented):
I mean, users will click and know what's inside the menu, but I was wondering if  "Picker components" doesn't lead to thinking that it's about the components used to build the pickers.

Another consideration: to establish a clear difference from the group of pickers that are coming soon (based on the new fields), would it make sense to rename this group and the next to something like:

=> Legacy Pickers
=> Next Pickers
=> Date and time fields
=> Custom sub-components

=> Masked Pickers
=> Pickers
=> Date and time fields
=> Custom sub-components

=> Masked Pickers
=> Rich keyboard Pickers
=> Rich keyboard Fields
=> Custom sub-components

joserodolfofreitas:
> I would avoid to use `Masked Pickers` because I think most people won't know what "masked" mean.
> 
> I would prefer Legacy Pickers / New Pickers, or event Legacy Pickers (v5) / New Pickers (v6) to be very explicit.
> 
> Renaming "Custom components" into "Custom sub-components" is fine for me, it think it makes it clear enough.

Right. 
One of my reasonings for trying to establish a clear difference based on how they work, is that we still need more validation to "ditch" the pickers based on the masked approach.

oliviertassinari:
The proposed grouping feels a bit backward. I think that the first thing that developers care about is the type of field that they need: time, date, date time, and only then they will consider the type of UX they need, standalone, keyboard only, integrated, mobile only. So, I suspect that reversing the grouping could work better.

flaviendelangle:
You could go for the following ?

- Date
  - Date Field
  - Date Picker

- Time
  - Time Field
  - Time Picker

- Date Time
  - Date Time Field
  - Date Time Picker

- Date Range
  - Date Range Field
  - Date Range Picker

- Time Range
  - Time Range Field
  - Time Range Picker

- Date Time Range
  - Date Time Range Field
  - Date Time Range Picker


If so, I feel like it's a lot of folders and in the case I would indeed go with headers like @samuelsycamore proposed.

oliviertassinari:
@flaviendelangle I was thinking about something like this yes, I like the idea of Sam to use the menu item subheaders to avoid too much nesting.

flaviendelangle:
@oliviertassinari if we do that, where would you put the "Getting started" page of the field ?

docs/data/pages.ts
```diff
@@ -148,62 +148,98 @@ const pages: MuiPage[] = [
...
             pathname: '/x/react-date-pickers/date-time-range-field',
             title: 'Date Time Range Field',
-            plan: 'pro',
+            newFeature: true,
           },
         ],
       },
+      { pathname: '/x/react-date-pickers/validation' },
+      { pathname: '/x/react-date-pickers/localization' },
+      { pathname: '/x/react-date-pickers/fields', title: 'Keyboard editing' },
```
alexfauquette, line 241: I'm wondering if those should not be on top of the components, because by priority of reading, I think it's:

1. Common usage (validation, localization)
2. Specific usage (Date picker special props)
3. Customization (slots)
4. components API

LukasTy: IMHO—the current order makes a tiny bit more sense.
From DX PoV I'd firstly be looking at which component to (can I) use. And only after that consider localisation and validation. 🤔 

The only section to me that sounds and looks off is the `Keyboard editing`... 🙈 

I think that we should either update the page title/contents a bit or use a different name.

Maybe going with `Field components` would make more sense?
And, IMO, this section could use a `newFeature: true` prop 🤔

LukasTy (approved):

oliviertassinari (commented):
This more clearly group pro and MIT features, nice 👍 

---

**Off-topic**

A few opportunities that I could find browsing the pages:

1. The static mode isn't static, it forces the focus: https://deploy-preview-6369--material-ui-x.netlify.app/x/react-date-pickers/date-picker/. It's why the scroll bar is broken on this page when loading it. This happens on 4-5 of the pages. It makes it hard to read the docs.
2. We could disable the ads on some of the pages, e.g. https://deploy-preview-6369--material-ui-x.netlify.app/x/react-date-pickers/date-range-field/.
3. The live edit demo https://deploy-preview-6369--material-ui-x.netlify.app/x/react-date-pickers/localization/#date-library-locale is broken. 

Screenshot 2022-11-15 at 19 08 12

We need to update the logic in the docs-infra of how we extract imports. This https://github.com/mui/material-ui/blob/04d2a26389c8ca299d5d5ab18da3e95a506f7332/docs/packages/markdown/extractImports.js#L1 is too simple. It should be like https://github.com/mui/material-ui/blob/04d2a26389c8ca299d5d5ab18da3e95a506f7332/docs/src/modules/sandbox/Dependencies.ts#L117.
Now, it uses the v1 of the live edit, the v2 makes the demo work when no edits are done. So, the bug will be less visible. TL:DR updating the @mui/monorepo will fix a large chunk of the issue. cc @bharatkashyap if it's a problem you would enjoy working on.

docs/data/pages.ts
flaviendelangle: With the new grouping by type, I feel like we are pushing the user to first find its component page and use it as the entry-point for the rest.

I don't think you come to the doc looking for information about the Localization / Validation before first checking how to use the component you need.

On nuance, the Localization pages will contain very basic topic on how to install and use the adapters after #6625 (which is probably even more early-stage in the user needs), but the needed sections will be linked from the getting started.

flaviendelangle: The only section to me that sounds and looks off is the Keyboard editing... see_no_evil

I think that we should either update the page title/contents a bit or use a different name.

Maybe going with Field components would make more sense?
And, IMO, this section could use a newFeature: true prop thinking

____

@LukasTy 

This page contains two quite different kind of content

1. Basic topics about what fields are. This could probably be moved to the *Getting Started* page on a new sections explaining that we have 3 type of components: the *Pickers*, the *Fields* and the *Calendar* / *Clock*. It could also be in a *Components* page that would be a summary of the available components if we feel that it's to much for the *Getting Started*/

2. Advanced topics (the current *Advanced* section)? This could be moved in an *Advanced* folder like on the grid with a *Field behaviors* page. Or if we don't want to create a folder, directly at the root.

I had to put the current page somewhere but I do agree that it is not great.
I can name it *Field components* (as the alpha only have power users who can take the time to look what "Field" mean) and work on a better split on a follow up PR.

LukasTy: Yes, I agree that we can simply name it `Field components` for now and re-work the whole structure and content in a follow-up PR leaving this one with single responsibility. 😉

joserodolfofreitas (approved):

PR #6387, opened 2022-10-05
[core] Add CodeQL workflow

DanailH:
Enable GitHub free code scanning.

Based on this https://mui-org.slack.com/archives/C03D1SRSUAH/p1664964827326369

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,74 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches:
+      - master
+      - next
+  schedule:
+    - cron: '0 2 * * 1-5'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'typescript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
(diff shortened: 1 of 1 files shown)


oliviertassinari (commented):
I had a quick look at what the other most popular repositories (GitHub stars) are doing. I have found a couple with this can:

- https://github.com/twbs/bootstrap/blob/main/.github/workflows/codeql.yml
- https://github.com/axios/axios/blob/v1.x/.github/workflows/codeql-analysis.yml
- https://github.com/microsoft/vscode/blob/main/.github/workflows/codeql.yml
- https://github.com/microsoft/TypeScript/blob/main/.github/workflows/codeql.yml

Some that find it too noisy

- https://github.com/nodejs/node/issues/36957#issuecomment-761565509
- https://github.com/ant-design/ant-design/commit/d5097a88e9af641ca637cc30cd1b0efe2525b80f

We could consider using https://securityscorecards.dev/#get-involved like

- https://github.com/angular/angular/blob/main/.github/workflows/scorecard.yml
- https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/scorecards-analysis.yml

joserodolfofreitas (approved):
This sounds great! 👌
I've used SonarQube in the past as well, I remember it being a bit of a pain to setup and customize correctly, so if CodeQL integration brings a quick win on this, I think it makes a lot of sense to at least experiment with it.

flaviendelangle (approved):
As discussed during our meeting, sounds like a good first step to iterate on :+1:

oliviertassinari:
There are some noise, but it's kind on OK https://github.com/mui/mui-x/security/code-scanning. It's likely worth to apply to our other repositories.

oliviertassinari:
I have flag all the reports in https://github.com/mui/mui-x/security/code-scanning that seems completely off. We are left with one report that might be interesting for us to account for. @flaviendelangle in https://github.com/mui/mui-x/security/code-scanning/8 the range used in the RegEx also match with these 5 characters `[\]^_` and with `. I guess these could be used a separators so shouldn't match?

flaviendelangle:
You are probably right, if you have a better selector in mind, I'm interested, this part is clearly not optimal.

oliviertassinari:
%aybe 

```diff
diff --git a/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts b/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
index c06e6d714..0fb541cf8 100644
--- a/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
+++ b/packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
@@ -291,7 +291,7 @@ export const splitFormatIntoSections = <TDate>(

   for (let i = 0; i < expandedFormat.length; i += 1) {
     const char = expandedFormat[i];
-    if (!char.match(/([A-zÀ-ú]+)/g)) {
+    if (!char.match(/([a-z]+)/ig)) {
       if (currentTokenValue === '') {
         sections[sections.length - 1].separator += char;
       } else {
```

PR #6392, opened 2022-10-05
New components: `MultiInputDateTimeRangePicker` and `MultiInputTimeRangePicker`

### LINKED ISSUE #6386: New components: `MultiInputDateTimeRangeField` and `MultiInputTimeRangeField`
See #6385 for the single input versions

Those field components are the first step for #4547 and #4460

You can have a look at https://github.com/mui/mui-x/pull/6312, the task here is pretty similar.
It is probably worth creating a new shared hook called `useMultiInputField` to avoid copy pasting to much logic.

alexfauquette:
Fix #6386

Having a single `useMultiInputField` is a bit hard because most of the functions called are dependent on the type of the input for example: `useDefaultizedTimeRangeFieldProps`, `useTimeField`, `timeRangeFieldValueManager`.

The most significant part that could be shared is the `buildChangeHandler`. But the function depends on its context; if we move it elsewhere, I assume we will need something like that

```jsx
const [startInputProps, endInputProps] = useInputProps({
  inStartInputProps,
  inEndInputProps,
  valueProp, 
  defaultValue, 
  format, 
  onChange,
  firstDefaultValue 
})
```

Not sure it will improve a lot the clarity of the code

I moved them to the internals for two reasons:
1. It's 3 time the similar code with import differences for now, so better to keep an eye on the 3 at the same place
2. There is no reason why it would be in `SingleInput...RandPicker` or `MultiInput...RandPicker` folder. So I prefer to put them in the internal such that I know what is specific to a picker and what is shared between them

---

On the same idea of shared parts, I reorganized the `models/dateRange.ts` as follows: 

- Generic types that are common to all range pickers are now in `range.ts`. 
- Specific types are in `dateRange.ts`, `timeRange.ts`, and `dateTimeRange.ts`

I spend a lot of time hacking id types with `Date` in their naming can be used in DateTime/Time components. Now it's obvious from the import path.

On the same idea, I created `UseDateRangeFieldProps<TDate>` by copying `UseSingleInputDateRangeFieldProps<TDate>`. The idea being that `UseDateRangeFieldProps` is the common type, and `UseSingleInputDateRangeFieldProps` and `UseMultiInputDateRangeFieldProps` are extending it. Instead of having `UseMultiInputDateRangeFieldProps` extending `UseSingleInputDateRangeFieldProps`

---

The objects  `[date|time|dateTime]RangeFieldValueManager` are now all in the same folder with a helper `rangeFieldValueManager` which does most of the job. 

Only `isSameError` is varying between them and mostly for typing reasons. We could replace `DateTimeRangeValidationError` by a generic `TRangeValidationError` and `isSameError: isSameDateTimeRangeError,` by 
```jsx
isSameError: (
  a: TRangeValidationError,
  b: TRangeValidationError | null,
) => b !== null && a[1] === b[1] && a[0] === b[0];
```

Because the `isSame[Date|Time|DateTime]RangeError` are the same at the exception of the typing

```tsx
export const isSameDateRangeError = (
  a: DateRangeValidationError,
  b: DateRangeValidationError | null,
) => b !== null && a[1] === b[1] && a[0] === b[0];

export const isSameDateTimeRangeError = (
  a: DateTimeRangeValidationError,
  b: DateTimeRangeValidationError | null,
) => b !== null && a[1] === b[1] && a[0] === b[0];
```

---

What changes between components? Basically the defaultized props and the passing of validation props. I will start a distinct PR to create `describeValidationRange` which will test the validation of range pickers


docs/data/date-pickers/date-range-field/BasicDateRangeField.js
```diff
@@ -4,13 +4,18 @@ import { LocalizationProvider } from '@mui/x-date-pickers';
...
+import { Unstable_MultiInputTimeRangeField as MultiInputTimeRangeField } from '@mui/x-date-pickers-pro/MultiInputTimeRangeField';
 
 export default function BasicDateRangeField() {
   return (
     <LocalizationProvider dateAdapter={AdapterDayjs}>
       <Stack spacing={4}>
         <MultiInputDateRangeField />
         <SingleInputDateRangeField />
+        {/* For experiment reasons (will be removed) */}
+        <MultiInputDateTimeRangeField />
```
flaviendelangle, line 17: We will create a quick follow up to #6323 to create two new pages

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
```diff
@@ -13,11 +13,8 @@ import {
   BaseDateValidationProps,
   DayValidationProps,
 } from '@mui/x-date-pickers/internals';
-import {
-  DateRange,
-  CurrentlySelectingRangeEndProps,
-  DayRangeValidationProps,
-} from '../internal/models/dateRange';
+import { DateRange, CurrentlySelectingRangeEndProps } from '../internal/models/range';
```
flaviendelangle, line 16: Off topic: I guess the `DateRange` name is becoming confusing
But that's not something easy to change given that it is used everywhere, including by users :/

packages/x-date-pickers-pro/src/DateRangePicker/shared.ts
```diff
@@ -67,6 +67,7 @@ export function useDateRangePickerDefaultizedProps<
   };
 }
 
+// What about renaming it `rangePickerValueManager` such that it's clear this manager is common to date, time and dateTime?
```
flaviendelangle, line 70: Are you sure that it should be the same ?
`timePickerValueManager` does have some differences with `datePickerValueManager` for instance (to avoid loosing the date).

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.tsx
```diff
@@ -0,0 +1,147 @@
+import * as React from 'react';
+import Stack, { StackProps } from '@mui/material/Stack';
+import TextField, { TextFieldProps } from '@mui/material/TextField';
+import Typography, { TypographyProps } from '@mui/material/Typography';
+import { styled, useThemeProps } from '@mui/material/styles';
+import { useSlotProps } from '@mui/base/utils';
+import { MultiInputDateTimeRangeFieldProps } from './MultiInputDateTimeRangeField.types';
+import { useMultiInputDateTimeRangeField } from '../internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField';
+
+const MultiInputDateTimeRangeFieldRoot = styled(
```
flaviendelangle, line 10: By the way, I find this part very ugly (including on the existing component)

If you have a better idea to create a slot with props passed only to the default slot.

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
```
flaviendelangle, line 1: Can we merge the `.interface.ts` file into this one for all multi input pickers ?
It's a mistake on my side for the existing one

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputTimeRangeField.ts
```diff
@@ -0,0 +1,120 @@
...
+} from '../../../MultiInputTimeRangeField/MultiInputTimeRangeField.types';
+import { dateRangePickerValueManager } from '../../../DateRangePicker/shared';
+import { timeRangeFieldValueManager } from '../valueManager/timeRangeValueManager';
+
+export const useDefaultizedTimeRangeFieldProps = <TDate, AdditionalProps extends {}>(
+  props: UseMultiInputTimeRangeFieldProps<TDate>,
+): UseMultiInputTimeRangeFieldDefaultizedProps<TDate> & AdditionalProps => {
+  const utils = useUtils<TDate>();
+
+  const ampm = props.ampm ?? utils.is12HourCycleInCurrentLocale();
```
flaviendelangle, line 27: I did not add the `ampm` prop to the `DateTimeField` and `TimeField`
If we are adding them today on the range fields we should add them there on another PR before the official release as well for consistency, and adapt the doc.

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputTimeRangeField.ts
```diff
@@ -0,0 +1,120 @@
...
+  const value = valueProp ?? firstDefaultValue.current ?? dateRangePickerValueManager.emptyValue;
+
+  const validationError = useValidation<
+    DateRange<TDate>,
+    TDate,
+    TimeRangeValidationError,
+    TimeRangeComponentValidationProps
+  >({ ...sharedProps, value }, validateTimeRange, () => true);
+  const inputError = React.useMemo(
+    () => timeRangeFieldValueManager.hasError(validationError),
```
flaviendelangle, line 105: Can be a follow up: On the pickers we are differentiating the start error and the end error
We could probably do the same here by calling the `hasError` of a date field on each input.

image

flaviendelangle (commented):
Great work !
I'm only focusing on some details

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerView.tsx
alexfauquette: Yes that's why I kept it, and used the import path to indicate that it's not only for date

packages/x-date-pickers-pro/src/DateRangePicker/shared.ts
alexfauquette: Effectively, I missed the `valueReducer`. Would make more sens if all the value manager implemented the same interface.

DateTime pickers could have `valueReducer = (utils, lastValidDateValue, newDate) => newDate`

For Date Picker, I'm wondering, why don't they need a value reducer to ensure they do not modify time part?

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
alexfauquette: I though you had a really precise definition of what should be in type and what in the interface 😂

packages/x-date-pickers-pro/src/DateRangePicker/shared.ts
flaviendelangle: To be fair the `valueReducer` is quite a dirty fix and maybe it will not be needed for the new pickers.
I feel like the generic approach of the new fields that never looses the rest of the date is better.

alexfauquette: I propose we keep it as it is for now and keep the comment. We will see at the end of alpha if a specific method has been needed. Fo now `valueReducer` is useless for range fields because they only modify the selected section of the date.

I tested it with a Date DateTime, and Time range picker fields controlled by the same value, and they stay in sync.

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputTimeRangeField.ts
LukasTy: Shouldn't that be left for locale/format to handle?
Major date libraries support different format for 12 or 24 hour clocks.
- [dayjs](https://day.js.org/docs/en/display/format#list-of-all-available-formats:~:text=0%2D23-,The%20hour,-HH)
- [moment](https://momentjs.com/docs/#/displaying/format/:~:text=am%20pm-,Hour,-H)
- [datefns](https://date-fns.org/v2.29.3/docs/format#:~:text=in%20the%20morning%2C%20...-,Hour%20%5B1%2D12%5D,-h)

Or was there a specific reason why we added separate `ampm` control? 🤔

flaviendelangle: On the existing pickers you have a default which is based on the local.
The behavior is as follow:

```ts
const ampm = props.ampm ?? utils.is12HourCycleInCurrentLocale();
const inputFormat: ampm ? utils.formats.fullTime12h : utils.formats.fullTime24h,
```

So the `ampm` is just a shortcut to help user switch from 12h to 24h format without caring about the actual format.

Is this shortcut worth the overhead ? I'm not sure.

packages/x-date-pickers-pro/src/DateRangePicker/shared.ts
flaviendelangle: Yes the  `valueReducer` will probably be needed for the `TimeRangePicker`.

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  Separator?: React.ElementType;
+}
+
+export interface MultiInputDateTimeRangeFieldSlotsComponentsProps<TDate> {
+  root?: SlotComponentProps<typeof Stack, {}, MultiInputDateTimeRangeFieldOwnerState<TDate>>;
+  input?: SlotComponentProps<
+    typeof TextField,
+    {},
+    MultiInputDateTimeRangeFieldOwnerState<TDate> & { position: 'start' | 'end' }
```
LukasTy, line 69: Nitpick: I see we are using `'start' | 'end'` type a lot in our code in-lined.
Maybe it would be worth to create a type for it? 🤔

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
```diff
@@ -0,0 +1,132 @@
...
+    ...props,
+    disablePast: props.disablePast ?? false,
+    disableFuture: props.disableFuture ?? false,
+    format: props.format ?? defaultFormat,
+    minDate: applyDefaultDate(utils, props.minDateTime ?? props.minDate, defaultDates.minDate),
+    maxDate: applyDefaultDate(utils, props.maxDateTime ?? props.maxDate, defaultDates.maxDate),
+    minTime: props.minDateTime ?? props.minTime,
+    maxTime: props.maxDateTime ?? props.maxTime,
+    disableIgnoringDatePartForTimeValidation: Boolean(props.minDateTime || props.maxDateTime),
+  } as any;
```
LukasTy, line 48: Maybe I'm missing something, but is `AdditionalProps` and this `as any` casting is necessary here?
Wouldn't removing those break something?

```diff
diff --git a/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts b/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
index ab5c640c2..bc93340ee 100644
--- a/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
+++ b/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
@@ -24,9 +24,9 @@ import {
 import { dateRangePickerValueManager } from '../../../DateRangePicker/shared';
 import { dateTimeRangeFieldValueManager } from '../valueManager/dateTimeRangeValueManager';
 
-export const useDefaultizedDateTimeRangeFieldProps = <TDate, AdditionalProps extends {}>(
+export const useDefaultizedDateTimeRangeFieldProps = <TDate>(
   props: UseMultiInputDateTimeRangeFieldProps<TDate>,
-): UseMultiInputDateTimeRangeFieldDefaultizedProps<TDate> & AdditionalProps => {
+): UseMultiInputDateTimeRangeFieldDefaultizedProps<TDate> => {
   const utils = useUtils<TDate>();
   const defaultDates = useDefaultDates<TDate>();
 
@@ -45,7 +45,7 @@ export const useDefaultizedDateTimeRangeFieldProps = <TDate, AdditionalProps ext
     minTime: props.minDateTime ?? props.minTime,
     maxTime: props.maxDateTime ?? props.maxTime,
     disableIgnoringDatePartForTimeValidation: Boolean(props.minDateTime || props.maxDateTime),
-  } as any;
+  };
 };
 
 export const useMultiInputDateTimeRangeField = <TDate, TChildProps extends {}>({
@@ -55,7 +55,7 @@ export const useMultiInputDateTimeRangeField = <TDate, TChildProps extends {}>({
   startInputRef,
   endInputRef,
 }: UseMultiInputDateTimeRangeFieldParams<TDate, TChildProps>) => {
-  const sharedProps = useDefaultizedDateTimeRangeFieldProps<TDate, TChildProps>(inSharedProps);
+  const sharedProps = useDefaultizedDateTimeRangeFieldProps<TDate>(inSharedProps);
 
   const { value: valueProp, defaultValue, format, onChange } = sharedProps;
```

packages/x-date-pickers-pro/src/internal/hooks/validation/useDateTimeRangeValidation.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  if (!isRangeValid(adapter.utils, value)) {
+    return ['invalidRange', 'invalidRange'];
+  }
+
+  return [null, null];
+};
+
+type DateTimeRangeValidationErrorValue = DateTimeValidationError | 'invalidRange' | null;
```
LukasTy, line 67: 1. Should we export these types as we with various validation error types in `x-date-pickers/src/index`? Users might need them to type methods handling `onError` or typing the variables storing error itself.
2. `| null` is redundant here as it is already included in `DateTimeValidationError` as one of the common error types.

LukasTy:
A bit too late to the 🎉 🤦

docs/data/date-pickers/date-range-field/BasicDateRangeField.js
```diff
@@ -6,11 +6,18 @@ import { Unstable_MultiInputDateRangeField as MultiInputDateRangeField } from '@
...
 
 export default function BasicDateRangeField() {
+  const [value, setValue] = React.useState([null, null]);
   return (
     <LocalizationProvider dateAdapter={AdapterDayjs}>
       <Stack spacing={4}>
-        <MultiInputDateRangeField />
-        <SingleInputDateRangeField />
+        <MultiInputDateRangeField
+          value={value}
```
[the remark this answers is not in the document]
flaviendelangle: This example goes away on the new doc anyway :+1: , we can discuss the actual organization in https://github.com/mui/mui-x/pull/6323

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
flaviendelangle: Agree, we start to use it in a lot of places

packages/x-date-pickers-pro/src/internal/models/range.ts
```diff
@@ -0,0 +1,13 @@
+import { FieldSection } from '@mui/x-date-pickers/internals-fields';
+
+export type DateRange<TDate> = [TDate | null, TDate | null];
+export type NonEmptyDateRange<TDate> = [TDate, TDate];
+
+export interface CurrentlySelectingRangeEndProps {
```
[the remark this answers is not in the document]
flaviendelangle: In the new `DateRangeCalendar` component I'm calling it `currentDatePosition`
This interface is not used on the fields or on the new pickers so it will just go away once we drop the old pickers :+1:

Issue #6391, opened 2022-10-05
[pickers] New component: `DateRangeCalendar`


flaviendelangle:
### Picking a date in a single date picker

For single date selection, we have a `CalendarPicker` component (soon to be renamed `Calendar` or `DateCalendar` in #6389).
This component is used to manage the various views that can be used to pick a full date.
This includes:

- Adding a header to switch between the views
- Render the year view 
- Render the month view
- Render the day view

This component can be used either as a standalone component or inside a date picker / date time picker.

### Picking a date in a date range picker 

In `DateRangePicker`, the closest equivalent is `DateRangePickerView` / `DateRangePickerViewMobile` and `DateRangePickerViewDesktop`. But it is very tightly linked to `DateRangePicker`.

This new `DateRangeCalendar` component would lay the base ground for several majors works:

- Allow to reuse the date range picking logic inside another range picker (#4547)
- Easily add other views such as `month` and `year` inside the `DateRangePicker` (#4546)
- Allow to use the date range picking behavior on a standalone component (like it can be done for `CalendarPicker` today)

PR #6403, opened 2022-10-06
[DateRangePicker] Fix the shape of the first selected day when the start date has an hour set

flaviendelangle:
Noticed it while working on #6391
Not sure what caused this regression

[Demo on `master`](https://codesandbox.io/s/great-sinoussi-bfzhp0)
[Demo on `next`](https://codesandbox.io/s/competent-liskov-hufkye)
[Demo on this branch](https://codesandbox.io/s/friendly-hellman-wqy7z8?file=/package.json)

With error:

image

Without error:

image


alexfauquette (approved):

packages/x-date-pickers-pro/src/internal/utils/date-utils.ts
```diff
@@ -13,12 +13,20 @@ export const isRangeValid = <TDate>(
...
   utils: MuiPickersAdapter<TDate>,
   day: TDate,
   range: DateRange<TDate> | null,
 ) => {
-  return isRangeValid(utils, range) && utils.isWithinRange(day, range);
+  if (!isRangeValid(utils, range)) {
+    return false;
+  }
+
+  const cleanRange = [
```
LukasTy, line 25: It seems strange that we'd need such data manipulation on every day comparison. 🤔 
I've tried debugging this method both on master and next branches with the following breakpoint:
`!!range && range[0] && range[0].hour() !== 0` and discovered that it is being triggered only on `next`, whereas on `master` this method receives day without hours, even though the defined value has hours (same demo used).
It seems that the root cause of the regression is somewhere up the chain. 🤔 

Are we sure that we want to fix it with such somewhat expensive (day manipulation) operation?

flaviendelangle: The change likely come from `parseRangeInputValue` which was moving to the start of the day.
I think we do have to support date with hours deeper in the tree now.
Because of #6416 (which has no parsing) and because of the future `DateTimeRangePicker` where we will have to keep the time.

flaviendelangle: But I could probably do this transformation inside the component to avoid doing it for every day

LukasTy: Gotcha.
Well, in such case it would still be nicer to user utility methods to overcome this issue instead of date manipulation.
Something along these lines:
```diff
- day.isBetween(range[0], range[1], null, '[]');
+ day.isBetween(range[0], range[1], 'day', '[]');
```
But that is not supported with current date-io implementation and I haven't checked if all date-io supported libraries do support `granularity` option.

LukasTy (approved):

packages/x-date-pickers-pro/src/internal/utils/date-utils.ts
LukasTy: > But I could probably do this transformation inside the component to avoid doing it for every day

Yes, my concern is the amount of additional date modifications we are going to do after this change 🙈

flaviendelangle: I pushed a commit to do the transform in the render.

> Well, in such case it would still be nicer to use utility methods to overcome this issue instead of date manipulation.

I agree that it would be a cleaner long term solution.

LukasTy: > I haven't checked if all date-io supported libraries do support granularity option.

I've checked date-io and can conclude that this feature would be tricky to implement as only a handful of date-util libraries do support comparison granularity. 😢

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerViewDesktop.tsx
```diff
@@ -176,9 +176,19 @@ export function DateRangePickerViewDesktop<TDate>(inProps: DateRangePickerViewDe
   const isNextMonthDisabled = useNextMonthDisabled(currentMonth, { disableFuture, maxDate });
   const isPreviousMonthDisabled = usePreviousMonthDisabled(currentMonth, { disablePast, minDate });
 
+  // Range going for the start of the start day to the end of the end day.
+  // This makes sure that `isWithinRange` works with any time in the start and end day.
+  const valueDayRange = React.useMemo<DateRange<TDate>>(
```
LukasTy, line 181: It seems to solve the root issue.
Should we still proceed with including this change in `alpha.2`?
If so, it should be replicated for Mobile version as well. 🤔

flaviendelangle: Can't wait for #6416 to stop having this duplication...

flaviendelangle: Done

Issue #4564, opened 2020-12-11
[DateRangePicker] In static mode, we can't choose to edit the `end` date then the `start` date


Krasnopir:
Hello. Maybe somebody know, is it possible for a static mode to choose which date is selected first "Check-in"(start) or "Check-out"(end)?

https://next--material-ui.netlify.app/ru/components/date-range-picker/#static-mode

Thank for  answer. sorry for issue

MohanadTafeal:
I have a question if anyone please can help me.
how can I add the ability -if the date in different year - to let the user click in the year to change the year first then choose the day. 
in DateRangePicker. please if anyone one know. thanks and sorry for my poor English.

flaviendelangle:
@gerdadesign on the [Static Date Range Picker](https://mui.com/x/react-date-pickers/date-range-picker/#static-mode), if we allow  the app to say "I want to first select the end date and then the start date".
Do you think we need to add a visual feedback somewhere to show that it's the end date that will be selected on click.
Or would the context in the app be enough ?

gerdadesign:
Hoping I understand this correctly! I would imagine the context should prep the user for selecting an end date first. 

For visual feedback and looking at the current behaviour, I think we are reliant on the hover indicator here. In static mode, if the date first selected is the end date, I would expect the dashed border to appear when I hover behind the selected date. Much like how it does when you select "Check out" first in non-static mode.

flaviendelangle:
On the new pickers, we would just have to add a `defaultRangePosition` prop on static range pickers (jinterface `StaticRangeOnlyPickerProps`) and initialize `rangePosition` with it in `useStaticRangePicker`.

If people need to fully control this state, we could expose a `rangePosition` and `onRangePositionChange` props as well and set a `useControlled` instead of the `useState` used for `rangePosition` in `useStaticRangePicker`.

PR #6416, opened 2022-10-07
[pickers] New component: `DateRangeCalendar`

### LINKED ISSUE #6391: [pickers] New component: `DateRangeCalendar`

flaviendelangle:
Closes #6391

Small side effect, this component allow to control the current date selected, so it will help for https://github.com/mui/mui-x/issues/4564 if we export the props at the level of the static component.

- [x] Create the new `DateRangeCalendar` component
- [x] Make sure it has all the props of `CalendarPicker` except the one which don't make sense
- [x] Copy the tests from `DesktopDateRangePicker.test.tsx` that only targets the view part (should create others in the future)

I would be in favor of not documenting it straightaway because it has the new apis (`defaultValue` for instance) which can be confusing since the other components are still with the old api.

Most of the logic is copy pasted for the current `DateRangePickerView` / `DateRangePickerViewDesktop` / `DateRangePickerViewMobile` components. But if you have improvements, that can be a good moment to do them.

## Follow up

- Support keyboard navigation on this component (to prepare for a date range pickers that supports it as well)


packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -0,0 +1,553 @@
...
+
+  const commonViewProps = {
+    disableHighlightToday,
+    readOnly,
+    disabled,
+  };
+
+  // When disabled, limit the view to the selected date
+  const minDateWithDisabled = (disabled && value[0]) || minDate;
+  const maxDateWithDisabled = (disabled && value[1]) || maxDate;
```
alexfauquette, line 289: Should the `minDate`/`maxDate` also take into account the `disablePast`/`disableFuture`?

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -0,0 +1,553 @@
...
+
+  const CalendarTransitionProps = React.useMemo(
+    () => ({
+      onMouseLeave: () => setRangePreviewDay(null),
+    }),
+    [],
+  );
+
+  // Range going for the start of the start day to the end of the end day.
+  // This makes sure that `isWithinRange` works with any time in the start and end day.
```
alexfauquette, line 301: Would it make sense to remove that part and replace `isWithinRange` with a helper `isWithinDayRange` which does the check by ignoring the time

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
alexfauquette: After reading the all file, I better understand its purpose. Maybe renaming it `defaultizedMinDate`.

alexfauquette (commented):
Why not use this new `DateRangeCalendar` in the `DateRangePicker`?

flaviendelangle:
> Why not use this new DateRangeCalendar in the DateRangePicker?

Because it does not support the toolbar and the mobile keyboard edition (equivalent of what `CalendarOrClockPicker` is doing)
And as the component is going to die, I did not want to spend time evolving it.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
flaviendelangle: You named those variables in `CalendarPicker` no ? :laughing: 
I can rename them in the two components

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
flaviendelangle: It is what I did at first in https://github.com/mui/mui-x/pull/6403
But we discussed with @LukasTy and that first draft was doing the date manipulation for everyday instead of once in the render.
So that's for performance reasons.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
alexfauquette: It's not the first time (and not the last) that I introduce stupid names for internal variables 😸

alexfauquette (approved):
Just some strange behavior on hover, but it's not from this PR. The fix is available in #6448

PR #6422, opened 2022-10-07
[fields] Support paste in single section

alexfauquette:


packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -111,21 +111,28 @@ export const useField = <
...
+      const activeSection = state.sections[selectedSectionIndexes.startIndex];
+      const isValidPastedValue =
+        (activeSection.contentType === 'letter' && /^[a-zA-Z]+$/.test(pastedValue)) ||
+        (activeSection.contentType === 'digit' && /^[0-9]+$/.test(pastedValue));
+      if (isValidPastedValue) {
+        // Early return to let the paste update section, value
+        return;
+      }
+      event.preventDefault();
       return;
```
alexfauquette, line 133: If we remove the `return` we could copy a valid date string without having to select all sections. Which can be interesting, because  <kbd>Ctrl</kbd>+<kbd>A</kbd> shortcut is not super famous

```suggestion
      event.preventDefault();
```

LukasTy: Honestly, I'm kind of in favour of this suggestion, especially if we don't encounter any clear issues with it. 🤔 
Based on my initial testing it seems to work without issues. 👌 

Maybe we could leave a comment for future selves, that if we want to disable full date pasting we should add `return` here? 🤔

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -111,21 +111,28 @@ export const useField = <
...
 
-    // TODO: Implement single section paste
+    const pastedValue = event.clipboardData.getData('text');
     if (
       selectedSectionIndexes &&
       selectedSectionIndexes.startIndex === selectedSectionIndexes.endIndex
     ) {
+      const activeSection = state.sections[selectedSectionIndexes.startIndex];
+      const isValidPastedValue =
+        (activeSection.contentType === 'letter' && /^[a-zA-Z]+$/.test(pastedValue)) ||
```
LukasTy, line 126: There seems to be issues with letter month pasting.
As an example pasting [in the last input here](https://deploy-preview-6422--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-date-format):
-  `10 February 2022` works fine
- `February` when month is selected has no effect 🤔

LukasTy:
Maybe we should add `DateField` to the `releaseChangelog` for changes related to new fields?
Or let's keep using `pickers`?
@flaviendelangle 
Because `[field]` would end up in `other` changes section. 🤔

flaviendelangle:
> Because [field] would end up in other changes section. thinking

We could update the script that generates the changelog.
If it impacts several fields, I'd say that `fields` is better than `DateField`

alexfauquette:
I updated the changelog script. `[fields]` should be placed correctly.

@LukasTy I fixed your two points by removing early returns :)

LukasTy (approved):

flaviendelangle (approved):

PR #6424, opened 2022-10-07
[core] Send explicit warning when dayjs locale is not found

alexfauquette:
I was playing with #6337 codesanbox when I tried to use `fr` locale, if failed saying that `"L"` was not a valid token. It took me some time to realize `dayjs` needs to have `'fr'` and not `'frFR'`. And the I have to import `'dayjs/locale/fr'` to let it work.

This PR add a warning and a fallback on `en`

### DIFF
--- a/packages/x-date-pickers/src/AdapterDayjs/index.ts
+++ b/packages/x-date-pickers/src/AdapterDayjs/index.ts
@@ -5,7 +5,7 @@ import { buildWarning } from '../internals/utils/warning';
 
 const localeNotFoundWarning = buildWarning([
   'Your locale has not been found.',
-  'Either the locale key is nt a support one. Locales supported by dayjs are available here: https://github.com/iamkun/dayjs/tree/dev/src/locale',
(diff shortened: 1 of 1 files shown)


packages/x-date-pickers/src/AdapterDayjs/index.ts
```diff
@@ -1,6 +1,14 @@
 import { Dayjs } from 'dayjs';
 import BaseAdapterDayjs from '@date-io/dayjs';
 import { MuiFormatTokenMap, MuiPickerFieldAdapter } from '../internals/models';
+import { buildWarning } from '../internals/utils/warning';
+
+const localeNotFoundWarning = buildWarning([
+  'Your locale has not been found.',
+  'Either the locale key is nt a support one. Locales supported by dayjs are available here: https://github.com/iamkun/dayjs/tree/dev/src/locale',
```
flaviendelangle, line 8:
```suggestion
  'Either the locale key is not a supported one. Locales supported by dayjs are available here: https://github.com/iamkun/dayjs/tree/dev/src/locale',
```

flaviendelangle (approved):

PR #5874, opened 2022-08-23
[core] Remove Firefox from the BrowserStack list

DanailH:
After discussing it with @m4theushw and @flaviendelangle we have decided that due to the flakiness of the tests run on Firefox that we will remove Firefox from the BrowserStack list (for not). This should not impact the overall stability of the code as we haven't had a case yet where this helped us catch a regression or bug.
We might add it back in the future if the problem is resolved.

Note: If removing Firefox doesn't solve the problem with the away as failed build on master we also discussed disabling the BrowserStack step altogether. For now, we will see if this PR solves the problem.

PR #6442, opened 2022-10-09
[test] Run the intergration tests in Firefox

oliviertassinari:
#5874 had removed Firefox because BrowserStack was unreliable. This PR is adding it back this time using CircleCI as the execution environment.

If we want a different version of Firefox, we can use https://circleci.com/developer/orbs/orb/circleci/browser-tools#commands-install-firefox.

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -30,6 +30,7 @@ const browserStack = {
 };
 
 process.env.CHROME_BIN = playwright.chromium.executablePath();
+process.env.FIREFOX_BIN = playwright.firefox.executablePath();
 
 // BrowserStack rate limit after 1600 calls every 5 minutes.
 // Per second, https://www.browserstack.com/docs/automate/api-reference/selenium/introduction#rest-api-projects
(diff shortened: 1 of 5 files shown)


test/karma.conf.js
```diff
@@ -44,7 +45,7 @@ const MAX_CIRCLE_CI_CONCURRENCY = 83;
 module.exports = function setKarmaConfig(config) {
   const baseConfig = {
     basePath: '../',
-    browsers: ['chromeHeadless'],
+    browsers: ['chromeHeadless', 'Firefox'],
```
oliviertassinari, line 48:
```suggestion
    browsers: ['chromeHeadless', 'FirefoxHeadless'],
```

oliviertassinari:
@DanailH @flaviendelangle I could make Firefox run without BrowserStack, however, one of the date picker tests seems to leak (isolation) and then break a bunch of others. It's not clear to me if it's worth fixing or not. The data grid on the other hands works without any issues, so we could always skip Firefox for the problematic tests to at least have the data grid covered 🤷

LukasTy:
> one of the date picker tests seems to leak (isolation) and then break a bunch of others. It's not clear to me if it's worth fixing or not.

Have you looked into how headless Firefox behaves regarding resolution and pointer type?
These two factors are important for date picker tests to be "desktop like", for this reason we have two arguments setting up chrome environment to:
- [have decent size window](https://github.com/mui/mui-x/blob/next/test/karma.conf.js#L132) to not be treated as mobile
- [have a "precise pointer"](https://github.com/mui/mui-x/blob/next/test/karma.conf.js#L129)—needed to render desktop version based on `@media (pointer: fine)`

DanailH:
Thanks for taking a swing at it @oliviertassinari. If we can have both ChromeHeadless and FirefoxHeadless it would be great. We can try skipping the problematic test for now and see how it will go.

oliviertassinari:
Ok, so I'm closing for now, it will require some time to make pass, we can wait to see the value.

PR #6443, opened 2022-10-10
[pickers] Refactor `@mui/material` imports to `@mui/utils`

### LINKED ISSUE #6394: Avoid importing generateUtilityClass from @mui/material to reduce module weight
### Duplicates

### Latest version

### Summary 💡

When I import date picker from `@mui/x-date-pickers/DatePicker`, my bundle contains all of `@mui/material`, due to the import of `generateUtilityClass` and `generateUtilityClasses` from `@mui/material`. I'm not actually sure how to resolve that, but I see that there have been some preliminary stabs at it (#3094). 

It would be great for my use case if I didn't have the full weight of @mui/material when I use a date picker.

### Examples 🌈

_No response_

### Motivation 🔦

I'm building a NextJS 12 application to be deployed on AWS Lambda. In doing that, I'm using `next build` with the `standalone` target.  This does some magic to trace what code is actually used, and only include that in the final build. So I've got a finite size of the built application due to Lambda limitations, so it would be great not to carry around this extra weight. In practice, we haven't hit our limit yet, but we're close. I've tried a bunch of things to do improve tree-shaking, but always end up back at the limitation that this module brings in all of `@mui/material`. 

So in summary, removing that import of `@mui/material` would help me sleep better knowing I'm not about to lose the ability to deploy code :).

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #6394 

Update imports avoiding the usage of root `@mui/material` to ease tree-shaking.

Tested changes with vite/rollup project (`rollup-plugin-visualizer`) -> zero effect on the tree-shaken final packages


flaviendelangle:
Looks good
I'll approve once the conflicts are solved :+1:

LukasTy:
> I'll approve once the conflicts are solved 👍

I wanted the issue author to test if it resolves his problem.
Because for usual web bundlers (webpack, rollup) it does not seem to matter.

If it does solve the author problem, I also have changes ready for data-grid package, although there would be one issue (type usage) needing special attention.

LukasTy:
> Looks good I'll approve once the conflicts are solved 👍

It seems that the issue author is not responding—I do still think that changes in this PR make sense regardless whether it solves the root problem or not.
But I think that data-grid changes would be best to introduce with a separate PR, hence I only updated this one.

flaviendelangle (approved):

PR #6448, opened 2022-10-10
[DateRangePicker] Fix calendar day outside of month layout shifting on hover

alexfauquette:
When moving to the slot, the DateRangeDay is adding styles to empty grid cells which leads to some inconsistencies

Here is a record of the problem when moving the mouse hover the cell

https://user-images.githubusercontent.com/45398769/194884995-50e17774-e6fd-4bfa-86cb-187e119775c5.mp4

The reason is that on hover the cell adds a 1px border. We do not see it because component are hidden, but it modifies the layout


flaviendelangle:
```diff
diff --git a/packages/x-date-pickers/src/PickersDay/PickersDay.tsx b/packages/x-date-pickers/src/PickersDay/PickersDay.tsx
index ea562295f..9191d61f4 100644
--- a/packages/x-date-pickers/src/PickersDay/PickersDay.tsx
+++ b/packages/x-date-pickers/src/PickersDay/PickersDay.tsx
@@ -184,6 +184,7 @@ const PickersDayFiller = styled('div', {
   ...styleArg({ theme, ownerState }),
   // visibility: 'hidden' does not work here as it hides the element from screen readers as well
   opacity: 0,
+  pointerEvents: 'none',
 }));
```

The following also fixes the problem, if you don't want the `DateRangePickersDay` to know why a day is hidden or not.

LukasTy:
Is it still a problem?
I struggle to reproduce the issue on latest `next` 🤷

alexfauquette:
I do not manage to reproduce it either. But Flavien's proposal makes sense even if the bug has been solved by an unknown commit.
Hidden element should not react to the pointer

flaviendelangle:
I can still reproduce it with a `DateRangeCalendar`

alexfauquette:
I confirm I can reproduce it (and the PR solves the bug) with 

```jsx
import { DateRangeCalendar } from '@mui/x-date-pickers-pro/DateRangeCalendar/index';
import { LocalizationProvider } from '@mui/x-date-pickers-pro'
import { AdapterDayjs } from '@mui/x-date-pickers-pro/AdapterDayjs'

export default function PlaygroundExample() {
  return (
      <DateRangeCalendar />
    </LocalizationProvider>
  );
}
```

Can not be done in codesandbox because `DateRangeCalendar` is not currently exported

LukasTy (approved):

LukasTy:
Proposal for a PR title change:
`[DateRangePicker] Fix calendar day outside of month layout shifting on hover`

PR #6463, opened 2022-10-12
[pickers] Do not use `Partial` for `components` and `componentsProps` props

flaviendelangle:
## Why ?

On the new pickers, some internal components have mandatory slots (`components.Field` for the mobile and desktop wrappers).
Which is not compatible with a `Partial` that makes every slot optional.

This PR replaces all `Partial` with an `?` on each of its slots.


flaviendelangle:
Merging this one, don't think it's worth spending time reviewing it given the amount of other important PRs opened

PR #6466, opened 2022-10-12
[core] Fix playground structure

LukasTy:
The existing git structure was not as expected for `playground` usage.

- Remove `playground.example.tsx`
- Create `playground/example.tsx` explaining how to use it
- Remove redundant `.gitignore` entry


.gitignore
```diff
@@ -11,7 +11,6 @@ __diff_output__
 /.nyc_output
 /coverage
 /docs/.next
-/docs/pages/playground.tsx
 /docs/pages/playground
```
flaviendelangle, line 14: Does `/docs/pages/playground` also catch `/docs/pages/playground.tsx` if it exists ?

LukasTy: Did you mean `docs/pages/playground/playground.tsx`?
If no, then no, it doesn't.
Should we?
Are most devs used to experiment by editing `docs/pages/playground.tsx`?

flaviendelangle: Sorry it was not clear
Does this new `.gitignore` correctly ignores  the old `/docs/pages/playground.tsx`

I still use it for my daily playground in which I just copy paste reproduction case, and `playground/new-pickers.tsx` (or others) for playground I want to keep for longer.

I could migrate to `playground/index.tsx` of course.
But we must be careful not to have someone push his old `playground.tsx` file by mistake.

LukasTy: Ok, gotcha, fair point, I'll update the ignore to include a `*` then. 👍

flaviendelangle (approved):

oliviertassinari:
I think that this change defeats the purpose of the introduction of the playground. The problem I tried to solve was:

1. I need to test a change in isolation
2. I open `docs/pages/x/introduction/index.js`
3. I copy & paste the reproduction on the GitHub issue
4. I make a fix
5. I start to make a commit
6. 💥 pain. I need to git reset the changes done in `docs/pages/x/introduction/index.js`

We are introducing pain 6 back with this PR:

Screenshot 2022-10-12 at 16 54 26

A counter proposal: https://github.com/mui/material-ui/pull/34739

LukasTy:
@oliviertassinari The original implementation had the same "issue" you are referring to—if you were to modify `playground.example.tsx`—the changes appear in git.
The idea is that developers would create their own playgrounds in `playground` folder and `example.tsx` just serves as that—an example.

I was initially unaware of this workflow to test components in isolation and changes were made to adapt to the way others (Siriwat, Flavian) got used to using it.

By no means—I do agree that this process could be improved and it's discoverability made better (like what you are suggesting in your PR 👍 ).

oliviertassinari:
> The original implementation had the same "issue" you are referring to—if you were to modify playground.example.tsx—the changes appear in git.

@LukasTy Ah right, makes sense 👍

flaviendelangle:
My current usage is as followed:

I have one playground which is cleaned several times a day (`pages/playground.tsx` right now, could be `pages/playground/temp.tsx` for instance).
On this one I do basically what @oliviertassinari describes.
I copy paste reproduction cases, doc examples etc... anything I want to play with right now.

But I also have other playgrounds (currently `new-pickers.tsx`) which I can keep for weeks and that I don't want to loose between two sessions of work on a given PR.

PR #6472, opened 2022-10-12
[docs] New page to migrate pickers from v5 to v6

flaviendelangle:
My main inspiration is https://mui.com/material-ui/migration/v5-component-changes/

I'm not sure we should group by component like they do, otherwise we could have a ton of duplication.


docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -0,0 +1,266 @@
...
+  componentsProps={{
+-   rightArrowButton: {},
++   nextIconButton: {},
+  }}
+/>
+```
+
+### Replace the `DialogProps` prop
+
+The `DialogProps` prop has been replaced by a `dialog` component props slot on all responsive and mobile pickers:
```
LukasTy, line 87: What do you think about removing `slot` word for these cases to avoid confusion between `component slot`? 🤔 
```suggestion
The `DialogProps` prop has been replaced by a `dialog` component props on all responsive and mobile pickers:
```

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+### Replace the `TransitionComponent` prop
+
+The `TransitionComponent` prop has been replaced by a `desktopTransition` component slot on all responsive and desktop pickers:
+
+```diff
+<DatePicker
+-  TransitionComponent={Fade}
++  components={{ desktopTransition: Fade }}
```
LukasTy, line 125:
```suggestion
+  components={{ DesktopTransition: Fade }}
```

LukasTy (commented):
I do agree that grouping by components would not work in pickers scenario. 👌

docs/data/pages.ts
```diff
@@ -278,6 +278,10 @@ const pages: MuiPage[] = [
         subheader: 'Upgrade to v6',
         children: [
           { pathname: '/x/migration/migration-data-grid-v5', title: 'Breaking changes: Data Grid' },
+          {
+            pathname: '/x/migration/migration-pickers-v5',
+            title: 'Breaking changes: Date & Time Pickers',
```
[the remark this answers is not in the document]
flaviendelangle: We creating those packages, we decided to go with the `Date and time pickers` wording to describe the new components.
The capitalization was not clearly defined (and with the recent decision we could unify everywhere).

I wanted to avoid `Pickers` because if one day we have a color picker, or an emoji picker, they would not be in the scope of those packages.

For the NPM packages, we favored concision by dropping the `-and-time` part.
I agree that this is inconsistent. I would not rename the NPM package even if we all thought that another name would be better, because it's a very heavy change that will bother our community.

If we do want to change the naming used in the doc and in the blog posts, I would be in favor of thinking it thoroughly and have @samuelsycamore in the loop.


_____

For the smaller change (`Breaking changes: Date and Time pickers` insteand of `Date & Time pickers`), I fully agree, it's a typo on my side.

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Update the format of the `value` prop
+
+The `value` prop of the pickers now expects a parsed value.
+Until now, it was possible to provide any format that your date management library was able to parse.
+For instance, you could pass `value={new Date()}` when using `AdapterDayjs`.
+
+This behavior brought a lot of confusion and was removed.
+The format expected by the `value` prop is now the same as for any other prop holding a date.
+Here is the syntax to initialize a date picker at the current date for each adapter:
```
alexfauquette, line 30: I tied to rewrite it to clearly distinguish what is before and after.

I took inspiration from a random [migration guideline from Google](https://developer.chrome.com/docs/workbox/migration/migrate-from-v4/#default-precache-manifest-replacement-point) where I like the "Previously ....., In v5 ...." structure

```suggestion
Previously, it was possible to provide any format that your date management library was able to parse.
For instance, you could pass `value={new Date()}` when using `AdapterDayjs`. 
This behavior brought a lot of confusion.

In v6, the format expected by the `value` prop is the same as for any other prop holding a date.
Here is the syntax to initialize a date picker at the current date for each adapter:
```

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -0,0 +1,266 @@
...
+-MuiCalendarPickerSkeleton: {
++MuiDayCalendarSkeleton: {
+```
+
+```diff
+-MuiMonthPicker: {
++MuiMonthCalendar: {
+```
+
+```diff
```
alexfauquette, line 245:
```suggestion

```

Could put all the examples in the same diff to be consistent with the last section

image

LukasTy (approved):

PR #6475, opened 2022-10-12
[Experiment] use grid layout

alexfauquette:
The idea is to put all the components into the following grid layout.

Basic customization could be done by setting for components in which row/column they should appear.

image


flaviendelangle:
Since I drop `CalendarOrClockPicker` in https://github.com/mui/mui-x/pull/6070, that PR has a lot of duplicated work.
Not sure how to handle the two in parallel.
I created a [https://github.com/mui/mui-x/blob/5179b3933cbbd24707a00488c54e62bf639ee692/packages/x-date-pickers/src/internals/components/PickerViewLayout.tsx](PickerViewLayout) file that does the layouting part of your new component but not the view management (that has been moved upward to avoid bundling useless views)

alexfauquette:
@flaviendelangle Yes, I will rebase when #6445 will be merged. It's mostly to validate the CSS approach

PR #6497, opened 2022-10-13
[docs] Check link validity in PR

alexfauquette:
Reuse helper proposed in https://github.com/mui/material-ui/pull/34588

Remaining: wait for core release, to update monorepo


oliviertassinari:
Does the logic finds this 404 on HEAD?
https://github.com/mui/mui-x/blob/eae809f2e9a46c00cc94af2403da11080be01983/docs/data/date-pickers/date-picker/date-picker.md?plain=1#L53

alexfauquette:
It's weird. The import was fine with `yarn link` but it does not work now with the monorepo

From what I understand, the file `/node_modules/@mui/monorepo/docs/scripts/reportBrokenLinks.js` is not able to resolve `@mui/markdown` 

Doing the following modification in the `node_module` file solves the problem
```diff
-const { createRender } = require('@mui/markdown');
+const { createRender } = require('../packages/markdown');
```

So a solution could be to this modification in the material-ui repository. 

Maybe there is another solution, but it's out of my knowledge about package alias

oliviertassinari:
@alexfauquette I have pushed a fix for it, the tricky aspect is that there is a lot happening beyond the scene that is hard to pinpoint the origin of where it starts and where it ends. I hope https://github.com/mui/material-ui/pull/34841#discussion_r1001921449 and my commit bring a bit more clarity on how this fits together.

PR #6503, opened 2022-10-14
[fields] Distinguish start and end input error on multi input fields

flaviendelangle:
### Goal

When only one of the two date does not pass the validation, we should only mark its input as invalid.
That is the behavior on the current `DateRangePicker` component.

For the single input range fields, we have to set the whole input as invalid of course.

### Reproduction case

On both examples below, only the start date is invalid (it is before the `minDate` prop)

[Before](https://codesandbox.io/s/date-and-time-pickers-forked-9pqyfg?file=/src/App.tsx)
[After](https://codesandbox.io/s/date-and-time-pickers-forked-czkwue?file=/src/App.tsx)


alexfauquette:
Nice, I took a similar approach in #6504

Do you know why the were memorized?

flaviendelangle:
> Do you know why the were memorized?

I memoized but given the complexity of the method, it clearly make no sense

alexfauquette (approved):

PR #6504, opened 2022-10-14
[test] Add validation test on range pickers

alexfauquette:


packages/x-date-pickers-pro/src/DateRangePicker/DateRangePicker.test.tsx
```diff
@@ -5,6 +5,7 @@ import { DateRangePicker } from '@mui/x-date-pickers-pro/DateRangePicker';
 import { fireEvent, screen } from '@mui/monorepo/test/utils/createRenderer';
 import { expect } from 'chai';
 import { createPickerRenderer, stubMatchMedia, wrapPickerMount } from 'test/utils/pickers-utils';
+import describeValidation from '../tests/describeValidation';
```
flaviendelangle, line 8: Should it be an absolute path like on the community package ?

packages/x-date-pickers-pro/src/tests/describeValidation.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const testsToRun = {
+  // day: testDayViewValidation,
+  textField: testTextFieldValidation,
+};
+
+interface ConformanceOptions {
+  render: any;
+  clock: any;
+  after?: () => void;
```
flaviendelangle, line 39: I don't seea ny usage of this callback, including in the community version, what would be the use case for it ?

alexfauquette: No idea, I assume if the core team added such option it was for good reason but Yes for now we don't need it

flaviendelangle: Oh you just took it from `describeConformance` ? OK to keep the same options for now even  if not used

packages/x-date-pickers-pro/src/tests/describeValidation.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Tests various aspects of picker validation
+ * components.
+ * @param {React.ReactElement} minimalElement - the picker to test
+ * @param {() => ConformanceOptions} getOptions
+ */
+export default function describeValidation(minimalElement, getOptions: () => ConformanceOptions) {
+  describe('RangePickers validation API', () => {
```
flaviendelangle, line 56:
```suggestion
  describe('Range pickers validation API', () => {
```

flaviendelangle (approved):

PR #6539, opened 2022-10-17
[fields] Add a `validationError` property to the `onChange` callback

### LINKED ISSUE #6421: [pickers] New Date field components - Why doesn't onChange receive the error string?
### Order ID 💳

33240

### Duplicates

### Latest version

### The problem in depth 🔍

Hey there!

I have been playing with the new field components, mainly `DateField`, and I would like to understand the technical decision behind having the validation `error` string only available `onError`.

**Why do I find it difficult to work with?**
In some cases, mostly when using date pickers/fields to do searching on a table or list, I would like to only set the value `onChange` in case it is valid ("valid" meaning it respects validation of `maxDate`, `minDate`, etc).

Right now, I'm kind of forced to either:
- Always set the value `onChange`, regardless of it being valid (and thus I have to validate the changed value before, for instance, calling an endpoint)
   - Although this is achievable, it's not great because I have already passed the `maxDate`, `minDate`, etc props. Having to do that again through some other way is redundant.
- Try to do a mix of `onError` and `onChange` to try to get what I want which is not a reliable way to do it.

If `onChange` signature could become:
```tsx
onChange?: (value: TValue, error: TError) => void;
```
It would allow developers for all kinds of use cases.

What do you think about this?

Cheers!

### Your environment 🌎

<details>

```
  System:
    OS: macOS 12.6
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.15.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 106.0.5249.103
    Edge: Not Found
    Firefox: Not Found
    Safari: 16.0
  npmPackages:
    @emotion/react: 11.10.4 => 11.10.4 
    @emotion/styled: 11.10.4 => 11.10.4 
    @mui/icons-material: 5.10.6 => 5.10.6 
    @mui/lab: 5.0.0-alpha.102 => 5.0.0-alpha.102 
    @mui/material: 5.10.8 => 5.10.8 
    @mui/system: 5.10.8 => 5.10.8 
    @mui/x-data-grid: 6.0.0-alpha.2 => 6.0.0-alpha.2 
    @mui/x-data-grid-generator: 6.0.0-alpha.2 => 6.0.0-alpha.2 
    @mui/x-data-grid-pro: 6.0.0-alpha.2 => 6.0.0-alpha.2 
    @mui/x-date-pickers: 6.0.0-alpha.2 => 6.0.0-alpha.2 
    @mui/x-date-pickers-pro: 6.0.0-alpha.2 => 6.0.0-alpha.2 
    @types/react: 18.0.21 => 18.0.21 
    react: 18.2.0 => 18.2.0 
    react-dom: 18.2.0 => 18.2.0 
    styled-components: 5.3.5 => 5.3.5 
    typescript: 4.9.0-dev.20221007 => 4.9.0-dev.20221007 
```
</details>

flaviendelangle:
Fixes #6421

- [x] Pass the validation error in the `onChange` callback

I went for an object instead of just passing the validation error as the 2nd parameter to be future proof.
I think that the validation is light enough to compute it on every call but we could change for a lazy approach:

```tsx
interface FieldChangeHandlerContext<TError> {
  validationError: TError;
}

onChange: (value: TValue,  getContext: () => FieldChangeHandlerContext<TError>) => void;
```

Or 

```tsx
interface FieldChangeHandlerContext<TError> {
  getValidationError: () => TError;
}

onChange: (value: TValue,  context: FieldChangeHandlerContext<TError>) => void;
```

### DIFF
--- a/docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
+++ b/docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
@@ -19,8 +19,8 @@ export default function LifeCycleIgnoreInvalidValue() {
               setValue(newValue);
             }
           }}
-          minDate={dayjs('2022-04-01')}
-          maxDate={dayjs('2022-04-30')}
+          minDate={dayjs('2022-01-01')}
+          maxDate={dayjs('2022-12-31')}
         />
       </LocalizationProvider>
       <Typography>Value: {value == null ? 'null' : value.format('L')}</Typography>
--- a/docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
+++ b/docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
@@ -1,0 +1,29 @@
+import * as React from 'react';
+import dayjs from 'dayjs';
+import Stack from '@mui/material/Stack';
+import Typography from '@mui/material/Typography';
+import { LocalizationProvider } from '@mui/x-date-pickers';
+import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
+import { Unstable_DateField as DateField } from '@mui/x-date-pickers/DateField';
+
+export default function LifeCycleIgnoreInvalidValue() {
+  const [value, setValue] = React.useState(null);
+
+  return (
+    <Stack spacing={2}>
+      <LocalizationProvider dateAdapter={AdapterDayjs}>
+        <DateField
+          value={value}
+          onChange={(newValue, context) => {
+            if (context.validationError == null) {
+              setValue(newValue);
+            }
+          }}
+          minDate={dayjs('2022-04-01')}
+          maxDate={dayjs('2022-04-30')}
+        />
+      </LocalizationProvider>
+      <Typography>Value: {value == null ? 'null' : value.format('L')}</Typography>
+    </Stack>
+  );
+}
--- a/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
+++ b/packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
@@ -6,10 +6,16 @@ import {
 } from '@mui/x-date-pickers/DateTimeField';
 import {
   applyDefaultDate,
+  DateTimeValidationError,
   useDefaultDates,
+  useLocalizationContext,
   useUtils,
   useValidation,
 } from '@mui/x-date-pickers/internals';
+import {
+  FieldChangeHandler,
+  FieldChangeHandlerContext,
+} from '@mui/x-date-pickers/internals-fields';
 import { DateRange } from '../../models/range';
 import type {
   UseMultiInputDateTimeRangeFieldDefaultizedProps,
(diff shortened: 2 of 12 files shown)


packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
```diff
@@ -59,24 +65,36 @@ export const useMultiInputDateTimeRangeField = <TDate, TChildProps extends {}>({
...
   const sharedProps = useDefaultizedDateTimeRangeFieldProps<TDate, TChildProps>(inSharedProps);
+  const adapter = useLocalizationContext<TDate>();
 
   const { value: valueProp, defaultValue, format, onChange } = sharedProps;
 
   const firstDefaultValue = React.useRef(defaultValue);
 
   // TODO: Maybe export utility from `useField` instead of copy/pasting the logic
-  const buildChangeHandler = (index: 0 | 1) => {
+  const buildChangeHandler = (
```
flaviendelangle, line 75: That part could be centralized between the 3 multi input fields.
It would be a first part of shared logic.

docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
```diff
@@ -0,0 +1,29 @@
...
+      <LocalizationProvider dateAdapter={AdapterDayjs}>
+        <DateField
+          value={value}
+          onChange={(newValue, context) => {
+            if (context.validationError == null) {
+              setValue(newValue);
+            }
+          }}
+          minDate={dayjs('2022-04-01')}
+          maxDate={dayjs('2022-04-30')}
```
alexfauquette, line 23: April 2022 is maybe too restrictive for a fast reader. A full year might be easier to find (I know it's described but people tends to play with the demo before finishing the text)
```suggestion
          minDate={dayjs('2022-01-01')}
          maxDate={dayjs('2022-12-31')}
```

alexfauquette (approved):
We already call the validation 31 times to know which date should be disable so running it one time on date change should not hurt

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateTimeRangeField.ts
alexfauquette: I agree, it look duplicated, but this function use its context. If you extract it, you will have to pass `firstDefaultValue`, `dateRangePickerValueManager`, ... as parameters and it will be harder to read

flaviendelangle: True, maybe we can try to refacto later to see if we can reach a better trade-off
But starting with a duplication avoid to build bad abstractions

docs/data/date-pickers/fields/LifeCycleIgnoreInvalidValue.js
flaviendelangle: You are probably right, updated :+1:

PR #6547, opened 2022-10-18
[adapters] Fix localisation of the placeholder

### LINKED ISSUE #6540: [pickers] When using DatePicker and LuxonAdapter, changing prop adapterLocale does not influence the keyboard mask.
## Duplicates
- [x] I have searched the existing issues

## Latest version
- [x]  I have tested the latest version (alpha-3)

## Current behavior 😯
When DatePicker is used together with LuxonThere there is a problem with using adapterLocale to influence the placeholder. It does not work. Additionally warning is shown:
`Mask does not support numbers with variable length such as 'M'.
Either use numbers with fix length or disable mask feature with 'disableMaskedInput' prop
Falling down to uncontrolled no-mask input. `

Context - https://github.com/mui/mui-x/issues/4755#issuecomment-1280636612

## Expected behavior 🤔
Placeholder text should react to changing locale. Additionally no warnings in console would be nice.

## Steps to reproduce 🕹
https://codesandbox.io/s/date-and-time-pickers-forked-fdfsxx?file=/src/App.tsx

Sorry for formatting, I created an issue from existing comment and there was template, my bad.

alexfauquette:
Fix #6540

Luxon has specific behavior about localization. You should always specify the locale when calling a method from `DateTime`

Now the `adapterLocale` is applied to both the placeholder and the formatted for the 4 libraries

image

### DIFF
--- a/packages/x-date-pickers/src/AdapterMoment/index.ts
+++ b/packages/x-date-pickers/src/AdapterMoment/index.ts
@@ -62,7 +62,9 @@ export class AdapterMoment
       .map((token) => {
         const firstCharacter = token[0];
         if (firstCharacter === 'L' || firstCharacter === ';') {
-          return this.moment.localeData().longDateFormat(token as LongDateFormatKey);
+          return this.moment
+            .localeData(this.getCurrentLocaleCode())
+            .longDateFormat(token as LongDateFormatKey);
         }
 
         return token;
--- a/test/utils/pickers-utils.tsx
+++ b/test/utils/pickers-utils.tsx
@@ -6,11 +6,12 @@ import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
 import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
 import { AdapterMoment } from '@mui/x-date-pickers/AdapterMoment';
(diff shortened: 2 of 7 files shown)


packages/x-date-pickers/src/AdapterMoment/index.ts
```diff
@@ -62,7 +62,9 @@ export class AdapterMoment
       .map((token) => {
         const firstCharacter = token[0];
         if (firstCharacter === 'L' || firstCharacter === ';') {
-          return this.moment.localeData().longDateFormat(token as LongDateFormatKey);
+          return this.moment
+            .localeData(this.getCurrentLocaleCode())
+            .longDateFormat(token as LongDateFormatKey);
```
alexfauquette, line 67: This is needed because otherwise the default locale of moment will be applied (the last imported one)

It's a part of code copypasted from date-io. after this PR I will add the modification to date-io

test/utils/pickers-utils.tsx
```diff
@@ -70,18 +72,25 @@ export function wrapPickerMount(mount: (node: React.ReactNode) => import('enzyme
 
 export function createPickerRenderer({
   locale,
+  adapterName,
```
alexfauquette, line 75: We can run the entire test codebase with any of the adapters. But only some of them are relevant to be run on various adapters. That's why I add this parameter. By priority, tests run with the adapter provided by:

- adapterName
- test:unit paramter
- date-fns (default one)

flaviendelangle (approved):

Issue #4662, opened 2022-04-26
[Date Picker] Customized month/year rendering


Semigradsky:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Summary 💡

There is the solution for customizing days - https://mui.com/x/react-date-pickers/date-picker/#customized-day-rendering

Need to have a solution for customizing months and years.

### Examples 🌈

`renderMonth` and `renderYear` properties?

flaviendelangle:
I would indeed be nice to be able to customize the year / month

@alexfauquette for day / month / year custom rendering, should be also use component slots ? (so replace `props.renderDay` by a `props.components.Day`

alexfauquette:
Yes, the slot seems to work well on datagrid, so we could reuse it here.

@Semigradsky What do you want to perform with custom month/year? Most of the modifications can be done with CSS class names

Semigradsky:
@alexfauquette My current need - specially display 'today' month/year. For example, today (2022-04-26) 'today month' is April and 'today year' is 2022.

alexfauquette:
Showing today does not feel like an isolated use-case. We could add a `.MuiPickersYear-today` and `.MuiPickersMonth-today` like we already have `.MuiPickersDay-today`

Semigradsky:
> We could add a .MuiPickersYear-today and .MuiPickersMonth-today like we already have .MuiPickersDay-today

And `renderMonth`/`renderYear`?

flaviendelangle:
We plan to create those two :+1:

LukasTy:
The ability to override views has been added via `viewRenderers` in v6.
The issue is that it has not yet been documented well and the only single example of its usage is given in the [migration guide](https://mui.com/x/migration/migration-pickers-v5/#stop-rendering-a-clock-on-desktop).

If you do have the time and need, could you please investigate if this option covers your use case?

As for the issue. I'm renaming it to cover the implementation of the `viewRenderers` documentation with examples.

LukasTy:
@sarink `viewRenderers` are for providing a totally custom view.
If you need to highlight specific days, there are definitely better ways to do it. Consider using `slotProps.day`.
Try checking [this example](https://mui.com/x/react-date-pickers/date-calendar/#custom-day-rendering), maybe it will help steer you in the right direction. 🤔

MonstraG:
[...]

PR #6550, opened 2022-10-18
[pickers] Rename remaining `private` components

LukasTy:
A follow-up to `themeAugmentation` work done a while back.
A few components have had their names left intact to avoid breaking changes.
Renaming those remaining components to align with other `pickers` components names.

- Fix `YearPicker` `yearButton` style overriding
- Simplify `PickersSlideTransition` classes merging/passing

#### Breaking changes

Previously we had 4 component names with `Private` prefix in order to avoid breaking changes in v5.
These components were renamed:

- `PrivatePickersMonth` -> `MuiPickersMonth`
- `PrivatePickersSlideTransition` -> `MuiPickersSlideTransition`
- `PrivatePickersToolbarText` -> `MuiPickersToolbarText`
- `PrivatePickersYear` -> `MuiPickersYear`

Manual style overriding will need to use updated classes:

```diff
-.PrivatePickersMonth-root {
+.MuiPickersMonth-root {

-.PrivatePickersSlideTransition-root {
+.MuiPickersSlideTransition-root {

-.PrivatePickersToolbarText-root {
+.MuiPickersToolbarText-root {

-.PrivatePickersYear-root {
+.MuiPickersYear-root {
```

Component name changes are also reflected in `themeAugmentation`:

```diff
 const theme = createTheme({
   components: {
-    PrivatePickersMonth: {
+    MuiPickersMonth: {
       // overrides
     },
-    PrivatePickersSlideTransition: {
+    MuiPickersSlideTransition: {
       // overrides
     },
-    PrivatePickersToolbarText: {
+    MuiPickersToolbarText: {
       // overrides
     },
-    PrivatePickersYear: {
+    MuiPickersYear: {
       // overrides
     },
   },
 });
```


packages/x-date-pickers/src/MonthCalendar/PickersMonth.tsx
```diff
@@ -12,11 +12,14 @@ import {
   PickersMonthClasses,
 } from './pickersMonthClasses';
 
-interface PickersMonthProps {
+export interface ExportedPickersMonthProps {
+  classes?: Partial<PickersMonthClasses>;
+}
```
alexfauquette, line 17: Why export only this prop? I must admit that I do not fully understand how to choose between exported and not exported interfaces

Also, should it have a `className` like for `ExportedPickersYearProps`?

packages/x-date-pickers/src/themeAugmentation/props.d.ts
```diff
@@ -61,24 +66,20 @@ export interface PickersComponentsPropsList {
...
   MuiYearCalendar: YearCalendarProps<unknown>;
-  // TODO v6: add `PrivatePickersMonth` once component is renamed to `MuiPickersMonth`
-  PrivatePickersMonth: never;
-  // TODO v6: add `PrivatePickersSlideTransition` once component is renamed to `MuiPickersSlideTransition`
-  PrivatePickersSlideTransition: never;
-  // TODO v6: add `PickersToolbarTextProps` once component is renamed to `MuiPickersToolbarText`
-  PrivatePickersToolbarText: never;
-  // TODO v6: add `PickersYearProps` once component is renamed to `MuiPickersYear`
-  PrivatePickersYear: never;
 }
```
alexfauquette, line 83: Love those TODO removed 😍

alexfauquette (approved):
Sounds super clean 👍. Just added two minor questions for stuff I'm not sure about what should I don't know

packages/x-date-pickers/src/MonthCalendar/PickersMonth.tsx
LukasTy: > Why export only this prop?

This is an internal component—I'd imagine no one would want to use it standalone.
Hence, I thought, that it doesn't make sense to expose the possibility of setting other props as `defaultProps` when they are going to be overwritten by `MonthCalendar`.
Do you think that someone would want to use this component standalone? 🤔 

> Also, should it have a className like for ExportedPickersYearProps?

I'll look into addressing it. Wasn't looking into aligning everything 100% with this PR. 🙈

alexfauquette: In standalone probably not, But we have #4662 creating slots for month and year buttons, so dev will need to know which props they can use on those slots

flaviendelangle: > Also, should it have a className like for ExportedPickersYearProps?

That one is tricky.
In some situations we are using the `className` prop internally so if we expose it and people override it using the theme, it will break our component.

I'm not very clear on that topic tbh.

flaviendelangle: So I would be in favor of not exposing `className` for instance

LukasTy: > In some situations we are using the className prop internally so if we expose it and people override it using the theme, it will break our component.

Are you sure it would break anything? `defaultProps` in theme definition only sets the props to defined values in case they are undefined 🤔 
If for you or another reason a higher order component is passing a `className` down - customized `defaultProps` should take no effect.

P.S. I did test the changes—both `PickersMonth` and `PickersYear` work fine with either `className` or `classes.root` defined in `defaultProps`. 🤷

flaviendelangle: > Are you sure it would break anything? defaultProps in theme definition only sets the props to defined values in case they are undefined thinking

You are right, it would not break our code
But what would be the point of this override ?
If we take the example of the `className`, in which scenario would someone use it if it can be overridden and they don't know when ?

LukasTy: Fair points.
If I follow your suggestion correctly—you are suggesting to only expose `className` to usable standalone components and not the internal/undocumented (currently) ones?

flaviendelangle: Something like that yes
In the new pickers I have the `PickerViewLayout` component which is wrapped in a `styled` for the static pickers.
So for this component, I think that exposing `className` in the theme would make no sense for instance.

LukasTy: - Removed `className` from exposed props in `PickersYear`
- Remove `className` in `PickersMonth` (it was not present—no point in adding it as long as it's not needed)

flaviendelangle (approved):

PR #6561, opened 2022-10-19
[pickers] Improve consistency between single and multiple range pickers

### LINKED ISSUE #6404: [pickers] Correctly support validation for range fields (both single and multi input)
See https://github.com/mui/mui-x/pull/6339#pullrequestreview-1132660722

alexfauquette:
Fix #6404 

Before, we had an early return if one of the fields was incomplete.

Now if one of the fields has an error, it displays the error immediately (even if the second one is not set). 

I had in mind the following use case: If you set the start date to 2021 and `disablePast` is set to `true` you don't need to wait for the second field to be filled to warn the user that the start date is wrong.

If you're ok with the behavior Iw ill add tests

### DIFF
--- a/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
+++ b/packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
@@ -21,11 +21,6 @@ export const validateDateRange: Validator<
 > = ({ props, value, adapter }) => {
   const [start, end] = value;
 
-  // for partial input
-  if (start === null || end === null) {
-    return [null, null];
-  }
-
   const { shouldDisableDate, ...otherProps } = props;
 
   const dateValidations: [DateRangeValidationErrorValue, DateRangeValidationErrorValue] = [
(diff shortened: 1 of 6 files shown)


LukasTy (approved):
LGTM 🙌 
Nitpick: This PR prefix does not exist in our `releaseChangelog` 🙈

packages/x-date-pickers-pro/src/internal/hooks/validation/useDateRangeValidation.ts
```diff
@@ -51,6 +46,11 @@ export const validateDateRange: Validator<
     return dateValidations;
   }
 
+  // for partial input
+  if (start === null || end === null) {
```
flaviendelangle, line 50: From what I see, `isRangeValid` will return false anyway if one of the two date is null.
Are you sure this change works how you want it to work ?

alexfauquette: `isRangeValid` is not applied if one of the date is `null` because of the early return.

flaviendelangle (commented):
:ok_hand: way better

Like Lukas said, the PR title will not be analyzed correctly
I think we can stick with `[pickers]` / `[field]` / `[{component name}]`.

flaviendelangle (approved):

PR #6563, opened 2022-10-19
[test] Test Arrow up/down on every token

### LINKED ISSUE #6210: [pickers] Add a test looping upon the `formatTokenMap` of each adapter to test the field behavior with it
## Goal

Test the basic interactions for every token format of every adapter (Arrow Up => increment, Arrow Down => decrement, Backspace => Reset, ...)

## Basic behavior

I did not try it, but it would work something like:

```tsx
it('should allow to edit all the date format token', () => {
  import { adapterToUse } from 'test/utils/pickers-utils';
  
  const formatToken = (formatToken: string, sectionName: MuiDateSectionName) => {
    const { unmount } = render(<TestCase format={formatToken} />)
  
    const input = // ...
    fireEvent.keyDown(input, { key: 'ArrowUp' });
    assert(); // assert that value has increased by one of the `sectionName` part.
  
    fireEvent.keyDown(input, { key: 'ArrowDown' });
    assert(); // assert that value has decreased by one of the `sectionName` part.
  
    // ... other interactions to test 
  
    unmount();
  }
  
  Object.entries(adapterToUse.formatTokenMap)
    .map(([formatToken, sectionName]) => formatToken(formatToken, sectionName));
});
```

```

### PR DESCRIPTION
**Pull request by alexfauquette**
Fix #6210

@flaviendelangle I'm not fully convinced it's useful to test every token, because the parsing/formatting is made (and tested by the parser) the remaining logic is similar for all tokens. One test per section name could be enough.

For now, I did the UP/Down keys.

I'm thinking about adding tests for:

- [ ] Pressing Home/End
- [ ] Pressing PageUp/PageDown
- [ ] Pressing a 1 digit number
- [ ] Pressing a 2 digit number
- [ ] Pressing month name in different locales

Do you see other edge cases?

### CODE REVIEW

**Comment by LukasTy** [comment]
> Do you see other edge cases?

Did you mean actual possible edge-cases or other possible interactions?
If we are talking about interactions—there are <kbd>Home</kbd>/<kbd>End</kbd> and <kbd>PageUp</kbd>/<kbd>PageDown</kbd>.

There is also am/pm entry. 🤔

**Comment by alexfauquette** [comment]
> If we are talking about interactions—there are Home/End and PageUp/PageDown.

Nice that's the kind of stuff I was going to forget

**Comment by flaviendelangle** [comment]
> @flaviendelangle I'm not fully convinced it's useful to test every token, because the parsing/formatting is made (and tested by the parser) the remaining logic is similar for all tokens.

You still have differences that we would not catch with one test by section.
For instance, if you test a format `1-31` for the day, you are not testing the format `01-31`, and I had bugs specific to the 2nd one. Same for 
Same, if you test the format `1-12` for the month, you are not testing `Jan-Dec` or `January-December`, or `01-12`.

I agree that most of the logic is shared.
And we probably don't need 20 tests per token (for instance testing the PageUp / PageDown for all of them seems redundant if we already test ArrowUp / ArrowDown).
But I would at least to 2-3 tests to check the basic behaviors.

File: packages/x-date-pickers/src/tests/fieldKeyboardInteraction.test.tsx [hunk 1]
```diff
@@ -43,12 +36,12 @@ function updatedDate<TDate>(
 const expectInputValue = (input: HTMLInputElement, expectedValue: string) =>
   expect(input.value.replace(/‎/g, '')).to.equal(expectedValue);
 
-const adapterToTest: ("luxon" | "date-fns" | "dayjs" | "moment")[] = ['luxon', 'date-fns', 'dayjs', 'moment']
-
-// const adapterToTest: ("luxon" | "date-fns" | "dayjs" | "moment")[] = ['luxon'];
-// const adapterToTest: ('luxon' | 'date-fns' | 'dayjs' | 'moment')[] = ['date-fns'];
-// const adapterToTest: ("luxon" | "date-fns" | "dayjs" | "moment")[] = ['dayjs'];
-// const adapterToTest: ("luxon" | "date-fns" | "dayjs" | "moment")[] = ['moment'];
+const adapterToTest: ('luxon' | 'date-fns' | 'dayjs' | 'moment')[] = [
```
flaviendelangle, line 39:
```suggestion
const adapterToTest =['luxon', 'date-fns', 'dayjs', 'moment'] as const
```

alexfauquette:
I propose to merge this PR and when #6590 is done add digit and letters tests

flaviendelangle (approved):

PR #6567, opened 2022-10-20
[DataGridPremium] Add support for cell selection

### LINKED ISSUE #208: [data grid] Implement cell selection with range

m4theushw:
Preview: https://deploy-preview-6567--material-ui-x.netlify.app/x/react-data-grid/selection/#cell-selection

Fixes #208

This PR contains the initial implementation of the cell selection feature. The new feature is opt-in, meaning that users need to enable it with the `cellSelection` prop. Both row selection and cell selection can work in parallel, however, using keyboard for selection will lead to different selections. I can't see how this feature will be used in real applications, besides to copy and paste cells, so I didn't implement all equivalent props from row selection here, e.g. `isRowSelectable`. This can be done later as we receive feedback from users.


joserodolfofreitas:
At first glance, the selection of multiple cells feels great!
And it's really nice that we can already use the controlled model to update multiple selected cells as well.
As I understand we need a new issue to implement "copy" from selection, right?

joserodolfofreitas:
It's fine. It's reasonable to have it as a separate issue and PR for it, it also reduces the review scope. 
Maybe what could be in the scope of the current PR is perhaps an apiRef method to get the selected cells (and values). Would that make sense? Or is it most of the work for copying already?

MatthijsKok:
The "click and drag" in the demo of the Preview does not work for me in Firefox (106.0.5), but does work in Edge.  
The other shortcuts do work.  

This is what happens in Firefox when I try to "click and drag" from "Wheat" to "Clayton McCoy"
image

docs/data/data-grid/selection/CellSelectionGrid.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export default function CellSelectionGrid() {
+  const [rowSelection, setRowSelection] = React.useState(false);
+
+  const { data } = useDemoData({
+    dataSet: 'Commodity',
+    rowLength: 10,
+    maxColumns: 6,
+  });
+
```
cherniavskii, line 14: Is it possible to scroll automatically when selecting cells range?
Currently, using a mouse you can only select cell range within visible cells


https://user-images.githubusercontent.com/13808724/200921355-9957b82a-47f0-450f-baf6-78b8b2cb2a0a.mov

docs/data/data-grid/selection/CellSelectionGrid.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
```
cherniavskii, line 15: For the Shift + Click range selection, is it possible to "remember" the first cell that started the range? 
Like in spreadsheet:

https://user-images.githubusercontent.com/13808724/200922650-ee9c591d-92ff-480c-833a-967570bcfed2.mov

Currently, the grid creates new range selection started from the last click cell:

https://user-images.githubusercontent.com/13808724/200922830-9a70d55b-afd6-4098-a8b5-5381f2e4e98c.mov

packages/grid/x-data-grid-premium/src/hooks/features/cellSelection/useGridCellSelection.ts
```diff
@@ -0,0 +1,363 @@
...
+        finalStartColumnIndex = endColumnIndex;
+        finalEndColumnIndex = startColumnIndex;
+      }
+
+      const visibleColumns = apiRef.current.getVisibleColumns();
+      const rowsInRange = visibleRows.rows.slice(finalStartRowIndex, finalEndRowIndex + 1);
+      const columnsInRange = visibleColumns.slice(finalStartColumnIndex, finalEndColumnIndex + 1);
+
+      const newModel = rowsInRange.reduce<GridCellSelectionModel>((acc, row) => {
+        acc[row.id] = columnsInRange.reduce<Record<GridColDef['field'], boolean>>(
```
cherniavskii, line 118: If I understand correctly, the cell selection model keeps track of every selected cell by id + field.
I've noticed a few things related to this:
- you can create a range selection with gaps
- you end up with gaps in range selection after sorting

Maybe we should consider a selection model that has only start and end cell coordinates, where coordinates are row and column indices rather than rowId and column field?

docs/data/data-grid/selection/CellSelectionFormulaField.js
```diff
@@ -0,0 +1,87 @@
...
+          Update selected cells
+        </Button>
+      </Stack>
+      <div style={{ height: 400 }}>
+        <DataGridPremium
+          apiRef={apiRef}
+          rowSelection={false}
+          cellSelectionModel={cellSelectionModel}
+          onCellSelectionModelChange={handleCellSelectionModelChange}
+          cellSelection
```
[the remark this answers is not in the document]
oliviertassinari: Or 😁

```suggestion
          unstable_cellSelection
```

docs/data/data-grid/selection/CellSelectionGrid.tsx
oliviertassinari: In terms of UX, from the sources that I could benchmark:

- Airtable: OK but it feels that it starts scrolling too far away from the bottom of the viewport. It's distracting.
- Google Sheet: OK but it feels that it starts scrolling too late, only one you are beyond the bottom of the viewport. You don't see what you are going to select next.
- Notion: Pretty strong, the threshold feels good, and they also have an incremental scrolling speed based on far you are from the bottom of the viewport 👌.

docs/data/data-grid/selection/selection.md
```diff
@@ -84,25 +84,89 @@ The following demo shows the prop in action:
...
+To select a single cell, click on it or, with the cell focused, press <kbd><kbd class="key">Shift</kbd>+<kbd class="key">Space</kbd></kbd>.
+Multiple cells can be selected by holding <kbd class="key">Ctrl</kbd> while clicking the cells.
+A cell can be deselected by also clicking it while <kbd class="key">Ctrl</kbd> is pressed.
+
+To select all cells within a range, you can use one of the shortcuts available:
+
+- Click on a cell, then drag the mouse over another cell and release it
+
+- Click on a cell and, while holding <kbd class="key">Shift</kbd>, click on another cell—if a third cell is clicked the selection will restart from the last clicked cell
+
```
oliviertassinari, line 105:
```suggestion
```

docs/data/data-grid/selection/CellSelectionGrid.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <div style={{ width: '100%' }}>
+      <Button sx={{ mb: 2 }} onClick={() => setRowSelection(!rowSelection)}>
```
oliviertassinari, line 17: This feels enough spacing:

```suggestion
      <Button sx={{ mb: 1 }} onClick={() => setRowSelection(!rowSelection)}>
```

docs/data/data-grid/selection/selection.md
```diff
@@ -84,25 +84,89 @@ The following demo shows the prop in action:
...
+// Selects the cell with field=name from row with id=1
+<DataGrid
+  cellSelectionModel={{ 1: { name: true } }}
+/>
+
+// Unselects the cell with field=name from row with id=1
+<DataGrid
+  cellSelectionModel={{ 1: { name: false } }}
+/>
+```
```
oliviertassinari, line 130: I assume?

````suggestion
```tsx
// Selects the cell with field=name from row with id=1
<DataGridPremium
  cellSelectionModel={{ 1: { name: true } }}
/>

// Unselects the cell with field=name from row with id=1
<DataGridPremium
  cellSelectionModel={{ 1: { name: false } }}
/>
```
````

File: docs/data/data-grid/selection/selection.md [hunk 8]
```diff
@@ -84,25 +84,89 @@ The following demo shows the prop in action:
...
+When multiple selected cells make a range, specific class names are applied to the cells at the edges of this range.
+The class names applied are the following:
+
+- `MuiDataGrid-cell--rangeTop`: applied to all cells of the first row of the range
+
+- `MuiDataGrid-cell--rangeBottom`: applied to all cells of the last row of the range
+
+- `MuiDataGrid-cell--rangeLeft`: applied to all cells of the first column of the range
+
+- `MuiDataGrid-cell--rangeRight`: applied to all cells of the last column of the range
```
oliviertassinari, line 150: Same feedback as above?

```suggestion
- `MuiDataGrid-cell--rangeTop`: applied to all cells of the first row of the range
- `MuiDataGrid-cell--rangeBottom`: applied to all cells of the last row of the range
- `MuiDataGrid-cell--rangeLeft`: applied to all cells of the first column of the range
- `MuiDataGrid-cell--rangeRight`: applied to all cells of the last column of the range
```

---

**Off-topic.** I would be in favor of going to http://github.com/mui/material-ui and adding a bit of margin-bottom for `ul > li`. e.g. 2px so it feels better. Added the idea to https://www.notion.so/mui-org/Opportunities-408a9771556a4715830d18f3749af753#66fe1b01bb3b495388492d2121674f00

packages/grid/x-data-grid-premium/src/tests/cellSelection.DataGridPremium.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+import {
+  DataGridPremium,
+  DataGridPremiumProps,
+  GridApi,
+  useGridApiRef,
+  gridClasses,
+} from '@mui/x-data-grid-premium';
+import { getBasicGridData } from '@mui/x-data-grid-generator';
+
+describe('<DataGridPremium /> - Cell Selection', () => {
```
oliviertassinari, line 16: Sentence case, I think in this case.

```suggestion
describe('<DataGridPremium /> - Cell selection', () => {
```

packages/grid/x-data-grid-premium/src/hooks/features/cellSelection/useGridCellSelection.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    if (event.shiftKey && lastClickedCell.current) {
+      apiRef.current.selectCellRange(lastClickedCell.current, { id, field });
+      lastClickedCell.current = { id, field };
+      return;
+    }
+
+    lastClickedCell.current = { id, field };
+
+    if (event.ctrlKey) {
```
oliviertassinari, line 223: For macOS?

```suggestion
    if (event.ctrlKey || event.metaKey) {
```

I tested it, and can now use Cmd like on Google Sheet.

docs/data/data-grid/selection/selection.md
```diff
@@ -84,25 +84,89 @@ The following demo shows the prop in action:
 
 {{"demo": "ControlledSelectionServerPaginationGrid.js", "bg": "inline"}}
 
-## apiRef [<span class="plan-pro"></span>](/x/introduction/licensing/#pro-plan)
+## Cell selection [<span class="plan-premium"></span>](/x/introduction/licensing/#premium-plan)
```
oliviertassinari, line 87: **Outside of the scope of this PR (maybe for a follow-up)**

It might be better to systematically have premium/pro features on a dedicated page if the section is larger than the viewport. My fear is that a developer might land in the middle of the section using the search without being able to easily see that it's a Pro/Premium feature. With a dedicated page, we get a UX benefit, even in the middle of the page, we see:

Screenshot 2022-11-10 at 00 27 59

MatthijsKok:
Wow! Thanks for the quick response @m4theushw !

docs/data/data-grid/selection/CellSelectionFormulaField.js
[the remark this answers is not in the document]
joserodolfofreitas: While I agree that using `unstable_` is preferred,  I'd say that what has the biggest risk to change is the `cellSelectionModel`, not this API particularly.

joserodolfofreitas (commented):
Without the support for command (ctrl) on Mac, I couldn't test this use case. But can we select "multiple" ranges? On AgGrid this operation is possible with **Ctrl + drag**.

image

Without Ctrl, I could also not test Range "deselection".

<hr />

A tangential point for another issue: 
I feel the "empty column" left at the end of the grid can be misleading. Jan had already mentioned a pain point with the expectation that it should be clickable to select a row, with cell selections this pain point might become more prominent.

image

fabioel:
Hello,

I could not see it in the demo, but how does this feature work in combination with dropdown fields, that are limited to certain values? How will this operate?
e.g. here (https://mui.com/x/react-data-grid/#commercial-version) status field. How could the value selector work combined with the dropdown?

docs/data/data-grid/selection/CellSelectionFormulaField.js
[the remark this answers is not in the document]
DanailH: I think the logic behind marking it as `unstable` is to signify that the entire feature with its underlying API is unstable.

DanailH:
Looks great 🙌.
I found one issue related to selecting cells using the Shift key. It's hard to explain so I made a screen recording.

https://user-images.githubusercontent.com/5858539/202136427-49db7c58-9dd3-495b-ae74-55efb0366577.mov

docs/data/data-grid/selection/CellSelectionFormulaField.js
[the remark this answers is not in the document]
oliviertassinari: @joserodolfofreitas Agree, 👍 for `unstable_cellSelectionModel` too.

docs/data/data-grid/selection/CellSelectionGrid.tsx
cherniavskii: > It's similar to what we have for the column headers (although it doesn't work).

Yes, there's an open issue for it https://github.com/mui/mui-x/issues/6236

packages/grid/x-data-grid-premium/src/tests/cellSelection.DataGridPremium.test.tsx
oliviertassinari: My thought is that it seems that can apply https://www.notion.so/mui-org/Style-guide-2a957a4168a54d47b14bae026d06a24b#d8e9cfb2423c4f83844a9f29d2032e3f in these cases.

docs/data/data-grid/selection/CellSelectionGrid.tsx
cherniavskii: We agreed on keeping the current behavior for now, as it requires some changes in how we manage the focused/outlined cell state.
@m4theushw will submit a follow-up PR to change the cell focus management so it's possible to always start a new range from a first selected cell of a previous range.
Did I capture it correctly?

packages/grid/x-data-grid-premium/src/hooks/features/cellSelection/useGridCellSelection.ts
cherniavskii: Since the majority of use cases for cell range selection don't involve sorting and/or filtering, we decided to keep the existing behavior and reconsider it in the future if needed.

cherniavskii:
> Hello,
> 
> I could not see it in the demo, but how does this feature work in combination with dropdown fields, that are limited to certain values? How will this operate? e.g. here ([mui.com/x/react-data-grid/#commercial-version](https://mui.com/x/react-data-grid/#commercial-version)) status field. How could the value selector work combined with the dropdown?

Hey @fabioel 
Could you clarify your question, please? I don't see how dropdown fields are related to the cell range selection feature.

fabioel:
Hello @cherniavskii,
For me there are two use-cases in regards to having `singleSelect`-type columns. The overarching question for me is: is there a way to propagate the value-options of the selected cells?

1. Multi-Select of `singleSelect`-type columns
You have 1 column with type=`singleSelect` with options Apple, Banana and you select multiple cells with different values. Now you want to overwrite all of the selected values based on the valueOptions (Apple or Banana) bound to the column. How can the valueOptions be accessed?

2. Mutli-Select of differently typed columns
if you have two columns: 1 with type='number', 1 with type='singleSelect'.
The `singleSelect` column has restrictions in regards to the allowed input values (e.g., Apple, Banana).
If you now select cells that have both types, how will the value setting work?

packages/grid/x-data-grid-premium/src/tests/cellSelection.DataGridPremium.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+  describe('Ctrl + click', () => {
+    it('should add the clicked cells to the selection', () => {
+      render(<TestDataGridSelection />);
+      expect(document.querySelector('.Mui-selected')).to.equal(null);
+      const cell11 = getCell(1, 1);
+      fireEvent.click(cell11);
+      expect(cell11).to.have.class('Mui-selected');
+      const cell21 = getCell(2, 1);
+      fireEvent.click(cell21, { ctrlKey: true });
+      expect(cell21).to.have.class('Mui-selected');
```
cherniavskii, line 72:
```suggestion
      expect(cell21).to.have.class('Mui-selected');
      expect(cell11).to.have.class('Mui-selected');
```

Should check if first selected cell is still selected

packages/grid/x-data-grid-premium/src/tests/cellSelection.DataGridPremium.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+  it('should unselect already selected cells when selecting a cell', () => {
+    render(<TestDataGridSelection />);
+    const cell01 = getCell(0, 1);
+    fireEvent.click(cell01);
+    expect(cell01).to.have.class('Mui-selected');
+    const cell11 = getCell(1, 1);
+    fireEvent.click(cell11);
+    expect(cell01).not.to.have.class('Mui-selected');
+    expect(cell11).to.have.class('Mui-selected');
```
cherniavskii, line 60: Maybe change the assertion to avoid [ambiguous errors](https://github.com/mui/mui-x/pull/5851#discussion_r952413615) when the test fails?
```suggestion
    expect(cell11.classList.contains('Mui-selected')).to.equal(true);
```

cherniavskii (approved):
Looks great!

DanailH (approved):
As we discussed the behavior when you try and make a new selection when there is already one active can be addressed in a separate PR.

docs/data/data-grid/selection/selection.md
oliviertassinari: Sounds great

oliviertassinari (commented):
Maybe it would be great to have a list of the next steps once this PR is merged, maybe for new GitHub issues 🤷‍♂️. Things I'm aware of:

- [ ] Move the docs to a separate page to better isolate Pro vs. MIT vs. Premium features https://github.com/mui/mui-x/pull/6567#discussion_r1018509333
- [ ] Support viewport auto scroll when reaching the edge https://github.com/mui/mui-x/pull/6567#discussion_r1018326298

packages/grid/x-data-grid-premium/src/tests/cellSelection.DataGridPremium.test.tsx
cherniavskii: https://github.com/mui/mui-x/pull/7054 should fix it, so if we're fine with the approach I've taken there - we can keep these assertions

cherniavskii: https://github.com/mui/mui-x/pull/7054 was merged

cherniavskii (approved):

MBilalShafi (approved):
🎉

oliviertassinari:
@m4theushw Awesome 👌, great to see this first iteration!

Design-wise, the customization might look better than the default https://next.mui.com/x/react-data-grid/cell-selection/#customize-range-styling, maybe it should be the default 😁

PR #6569, opened 2022-10-20
[DataGrid] Refactor `@mui/material` imports to `@mui/utils`

LukasTy:
Continuation of https://github.com/mui/mui-x/pull/6443 for DataGrid module(s).


packages/grid/x-data-grid/src/components/panel/GridPanel.tsx
```diff
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import { MUIStyledCommonProps } from '@mui/system';
 import { styled, Theme } from '@mui/material/styles';
-import { generateUtilityClasses, InternalStandardProps as StandardProps } from '@mui/material';
+import { unstable_generateUtilityClasses as generateUtilityClasses } from '@mui/utils';
+import { InternalStandardProps as StandardProps } from '@mui/material';
```
LukasTy, line 7: @DanailH Could you look into this case and possible ways to refactor it? 🤔 
As it is the single import blocking us from introducing the eslint rule discouraging usage of root `@mui/material` imports.

flaviendelangle: To be tested, but I think that `import type {}` should avoid bundling even in dev mode

LukasTy: That might be a possible workaround.
But IMHO looking at JSDoc—a more correct approach not using this export would probably be best.
Screenshot 2022-10-20 at 11 23 29

LukasTy: > To be tested, but I think that import type {} should avoid bundling even in dev mode
Tested and confirming that this change has no effect as bundler already understands what is a type import and it is removed from js bundle, but we still get it in type definition file.

DanailH: Ah, I see the problem. Maybe we can drop it and type it ourselves?

@mnajdova maybe there is a better way?

LukasTy: I've extracted the mentioned type [into `data-grid` package](https://github.com/mui/mui-x/pull/6569/files#diff-2cb999d376cde6ad0066b36709f8b42a1d48cd3a93f6b7d1fd2cb41f59f51023).
In order to do that—I've also added [explicit `@mui/types` dependency](https://github.com/mui/mui-x/pull/6569/files#diff-b631d036207bbd8389cbfbcde31151f39b96d3a07285af5772b73f52bd5d3a40R50).
Let me know what you think about it. 😉

cherniavskii: Maybe it's a dumb question, but why do we need `InternalStandardProps` in the first place?
It seems like we can use `PopperProps`:

```diff
- export interface GridPanelProps
-   extends StandardProps<MUIStyledCommonProps<Theme> & PopperProps, 'children'> {
+ export interface GridPanelProps extends PopperProps {
```

LukasTy: I've done some testing and it seems that the only things that will be different between existing and Andrew's suggestions are the `theme` and `as` properties.
I don't think I'm in the best position to answer if those are needed in this case as I'm not very knowledgable of all possible styling approaches in Data Grid packages.

cherniavskii: > it seems that the only things that will be different between existing and Andrew's suggestions are the theme and as properties

Indeed, thanks @LukasTy 
Here's a better solution that should cover everything:
```diff
- export interface GridPanelProps
-   extends StandardProps<MUIStyledCommonProps<Theme> & PopperProps, 'children'> {
+ export interface GridPanelProps extends React.ComponentProps<typeof GridPanelRoot> {
```

LukasTy: I've tried experimenting with both `as` and `theme` props and I fail to understand their role.
What would be the possible use-cases for providing them?
- I've tried passing in `theme` with `palette.background.paper` hoping that it would be used by the `styled` component—but it failed to provide any result...
- I've tried passing in various HTML elements to `as` prop—but all produced some wanky results

Based on this experiment I'm leaning towards:
```diff
- export interface GridPanelProps
-   extends StandardProps<MUIStyledCommonProps<Theme> & PopperProps, 'children'> {
+ export interface GridPanelProps extends Omit<PopperProps, 'children'> {
```

LukasTy: @cherniavskii @DanailH could any of you look into ☝️ above comment and decide on the most correct approach for the prop typing of this component? 🤔

cherniavskii: @LukasTy 
Let's use [this one](https://github.com/mui/mui-x/pull/6569#discussion_r1006814898) as it's 1:1 replacement to what we have had before (meaning that the same props are supported).
I'm not sure what's going on with `as` and `theme` props, we can look into that later

LukasTy: 🙇 Updated the solution with your suggestion. 👍

cherniavskii: Thanks @LukasTy !

PR #6571, opened 2022-10-20
[pickers] Export other adapters derived from moment or date-fns

alexfauquette:
This PR is for the 3 adapters derived from moment or date-fns

- AdapterDateFnsJalali
- AdapterMomentHijri
- AdapterMomentJalaali

The js-joda is hard to understand because there locales are build by a script from [`cldr-data`](https://www.npmjs.com/package/cldr-data)and after half a day on it I have no clue about where in the exported package could I get a way to extend the format


flaviendelangle:
> The js-joda is hard to understand because there locales are build by a script from [cldr-data](https://www.npmjs.com/package/cldr-data)and after half a day on it I have no clue about where in the exported package could I get a way to extend the format

https://github.com/mui/mui-x/issues/4703
It seems that this adapter is buggy anyway
Maybe we can consider dropping it for now :grimacing:

packages/x-date-pickers/src/AdapterDateFnsJalali/index.ts
```diff
@@ -0,0 +1,61 @@
+import BaseAdapterDateFnsJalali from '@date-io/date-fns-jalali';
+import defaultLocale from 'date-fns-jalali/locale/en-US';
+// @ts-ignore
+import longFormatters from 'date-fns-jalali/_lib/format/longFormatters';
+import { MuiFormatTokenMap, MuiPickerFieldAdapter } from '../internals/models';
+
+const formatTokenMap: MuiFormatTokenMap = {
+  y: 'year',
```
flaviendelangle, line 8: Should we just import this from the date-fns file ? Same for the others derived adapters whenever le token are the same.

For the methods it would also be nice not to duplicate, just like on date-io, but it would probably be more complicated.
That's a small additional argument in favor of not keeping the migration phase with some methods here and some on date-io forever.

test/utils/pickers-utils.tsx
```diff
@@ -6,16 +6,34 @@ import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns';
...
+  | 'jalaali'
+  // | 'js-joda'
+  | 'date-fns-jalali';
+
 const availableAdapters: { [key: string]: new (...args: any) => MuiPickersAdapter<any> } = {
   'date-fns': AdapterDateFns,
   dayjs: AdapterDayjs,
   luxon: AdapterLuxon,
   moment: AdapterMoment,
+  hijri: AdapterMomentHijri,
```
flaviendelangle, line 33: Could we always call them `moment-hijri` and `moment-jalaali` ?

LukasTy (commented):
Leaving comments about these adapters behaviour with `fields` components.

If I might have missed something in my demoes, which stopped me from correctly testing it—please point it out. 🙈 

If covering their functionality was not the intention and we're gonna be making follow-up PRs for that—feel free to ignore those comments.

packages/x-date-pickers/src/AdapterDateFnsJalali/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  yy: 'year',
+  yyy: 'year',
+  yyyy: 'year',
+  M: 'month',
+  MM: 'month',
+  MMMM: { sectionName: 'month', contentType: 'letter' },
```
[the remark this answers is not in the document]
alexfauquette: Yes, the problem comes from the `rtl` in string wich seems to be applied as soon as their is arabic characters.

We can fore the order with Unicode characters to fore either `rtl` or `ltr` (`\u200e` and `\u200f`) but before that I will need to find what is the correct display of such dates

alexfauquette:
I investigated this problem of moving elements, be ready, it's a bit complex.

The default format for date time input is `yyyy/MM/dd h:mm a`

If you fill it with data it can be `1452/10/20 16:54 ب.ظ.`
But it's not made to be displayed with RTL interface so `aa hh:mm yyyy/mm/dd`

The problem is nubers which are tricky because the computer can not distinguish when numbers are to link with the previous word (for example: building 5) or not

alexfauquette:
I start having something working correctly in rtl

https://codesandbox.io/s/lively-pond-gxr7vq?file=/demo.tsx

The strategy is to group elements by types according to spaces, assuming all the characters without spaces between them are of the same type (either digit of number)

Basically, on date time keyboard format, it's the following:

- raw format: `yyyy/mm/dd hh:mm aa`
- grouped format: `<>yyyy/mm/dd</> <>hh:mm</> <>aa</>`
- rendered in rtl:  `<>aa</> <>hh:mm</> <>yyyy/mm/dd</>`


Remaining points: 
- If the placeholder are rtl characters, it does not work, because in `hh:mm` with `hh` replaced by arabic characters, when you fill minutes, you get `<rtl characters>:12` which will be displayed  `12:<rtl characters>` in rtl context. (you can test in the codesandbox)
- the left/right keyboards need to be adapted, but that sounds like a minor issue compared to the other ones

flaviendelangle:
Great start !
I'm learning stuff about RTL

____

> The strategy is to group elements by types according to spaces, assuming all the characters without spaces between them are of the same type (either digit of number)

Does it work with format like `dd / mm` (where the `/` has spaces around it in the format itself, not just added by our pickers) ?

____

We need to document what `positionInInput` / `startInInput` and `endInInput` mean because from the outside it's not obvious.
By the way, maybe we can drop the `LRM` space I added (and the cleaning associated) if the new hidden characters allow to avoid the onChange not triggering when pressing `1` on a section which equals `1`.

alexfauquette:
> Does it work with a format like `dd / mm`

Not with the current technic, because it would split the character in the following groups `<>dd</> <>/</> <>mm</>` and having punctuation in an isolated group does not make sense. I tested and here is the result:

- Without the grouping `dd / mm / yyyy` is rendered `yyyy / mm / dd` in rtl
- With the grouping, it is rendered `/ yyyy / mm dd` in rtl. (I assume it considers `/` as a part of `mm`.

For context: isolation is made to indicate if weak characters like punctuation or digits should refer to the text before or after them. As an example, if you translate in Arabic:

- `I like CSS!`, you expect to get `!CSS ["I like" in arabic]` and not `CSS! ["I like" in arabic]` So you isolate the "CSS"
- `I like C++`, you expect to get `C++ ["I like" in arabic]` and not `++C ["I like" in arabic]` So you isolate the "C++"

---

For now, I'm using 
- U+2068 FIRST STRONG ISOLATE
- U+2066 POP ISOLATE

Which isolates a part of the text, and to know if it should be displayed rtl or ltr, it searches for the first strong character (a character known to be only in ltr or rtl). Basically, it ignores digits and punctuation. 

This is problematic when the value and the placeholder are not of the same type:
- Arabic placeholder with digit values
- Latine placeholder and value in arabic text

For example: `yyyy/MMM/dd` where months are indicated by a word is problematic because you will have the following states:
- month not set: `1405/MMM/02`
- month set: `02/[arabic month]/1405`

Seems we can force the rendering direction with 

U+2066 LEFT-TO-RIGHT ISOLATE
U+2067 RIGHT-TO-LEFT ISOLATE

It's a new opportunity I've to test

test/utils/pickers-utils.tsx
```diff
@@ -77,14 +96,21 @@ export function createPickerRenderer({
 }: CreatePickerRendererOptions = {}) {
   const { clock, render: clientRender } = createRenderer(createRendererOptions);
 
-  let adapterLocale = (adapterName ?? adapterToUse.lib) === 'date-fns' ? locale : locale?.code;
+  let adapterLocale = [
```
flaviendelangle, line 99:
```suggestion
  let adapterLocale = (adapterName ?? adapterToUse.lib).includes('date-fns')
```

Just a suggestion

LukasTy (approved):

test/utils/pickers-utils.tsx
alexfauquette: It took me some time to figure out I need to include `js-joda` here so I prefer to keep the comment for latter or for the other dev who will add it

yarn.lock
```diff
@@ -9846,6 +9870,11 @@ lru-cache@^6.0.0:
   resolved "https://registry.yarnpkg.com/luxon/-/luxon-3.0.3.tgz#573e65531efd3d92265feb640f02ba7a192e2388"
   integrity sha512-+EfHWnF+UT7GgTnq5zXg3ldnTKL2zdv7QJgsU5bjjpbH17E3qi/puMhQyJVYuCq+FRkogvB5WB6iVvUr+E4a7w==
 
+luxon@^3.0.2:
```
LukasTy, line 9873: This is a strange yarn decision to add a new version of luxon when the above version matcher would fall in line with it 🤔 
```
"luxon@^1.28.0 || ^2.0.0 || ^3.0.2":
```
Maybe yarn can't add `, luxon@^3.0.2` on it because it's version matcher with `||` clauses. 🤷

PR #6618, opened 2022-10-24
[pickers] Ignore milliseconds in mask logic

### LINKED ISSUE #6587: [TimePicker] Backspace doesn't work in the TimePicker when a default date value is provided
### Duplicates

### Latest version

### Current behavior 😯

When using the `ampm={false}` prop and passing a `Date` object instead of null, the timepicker doesn't allow erasing the minutes by pressing backspace, unless you select both digits and then.

[bug-timepicker.webm](https://user-images.githubusercontent.com/30949004/197204919-028a736e-93ff-4c97-b251-9956a2b3fb2d.webm)

Selecting both digits and then erasing works, but as soon as you type one digit, it fills in two digits, like shown in the video.

This doesn't work if the first value to be passed is `null`.

### Expected behavior 🤔

The user should be able to erase the time with the backspace key, digit by digit.

### Steps to reproduce 🕹

Link to live example: [demo](https://codesandbox.io/s/sleepy-brook-j3olef?file=/demo.tsx)

Steps:
1. Try to use backspace to erase the time in the provided text field.
2. Notice it doesn't allow you to.
3. Replace `React.useState<Date | null>(new Date());` with `React.useState<Date | null>(null);`
4. Notice you can freely type in a time and erase it, without bugs.

### Context 🔦

I'm trying to make a timepicker that receives a default value of `00:00` or a value provided by the server.

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #6587

Maybe a better approach would be that date parsing ignore the milliseconds such that we can always use `isEqual` without having to worry about those miliseconds

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/useMaskedInput.tsx
+++ b/packages/x-date-pickers/src/internals/hooks/useMaskedInput.tsx
@@ -95,7 +95,13 @@ export const useMaskedInput = <TInputDate, TDate>({
     const newParsedValue = rawValue === null ? null : utils.date(rawValue);
     const isAcceptedValue = rawValue === null || utils.isValid(newParsedValue);
 
-    if (!localeHasChanged && (!isAcceptedValue || utils.isEqual(innerInputValue, newParsedValue))) {
+    const innerEqualsParsed =
+      innerInputValue === null
+        ? newParsedValue === null
+        : newParsedValue !== null &&
+          Math.abs(utils.getDiff(innerInputValue, newParsedValue, 'seconds')) === 0;
+
+    if (!localeHasChanged && (!isAcceptedValue || innerEqualsParsed)) {
       return;
     }
 


oliviertassinari:
Would it make sense to close #6587 with the assumption that https://next.mui.com/x/react-date-pickers/time-field/ fixes the issue?

alexfauquette:
Not being able to do backspace is a big bug, that's why I investigated. But yes, mask logic will be removed. That's why I did a quick fix without adding tests

flaviendelangle (approved):

oliviertassinari:
The context for my comment wasn't really about this PR specifically, but to use it as a case study for a larger topic: the exploit vs. explore ressource allocation tradeoff. I think that the opposite of the points made could be argued:

1. The users that we are losing because of low quality are users that are willing to spend time to get a better implementation. Therefore I would assume that breaking changes we do is not a concern for them. To have a chance of them picking our new implementation we need to have it better than the alternatives, hence focus on it.
2. For existing bugs, I think that there a significant lag between changes we do, and when they benefits most (>80%) of the developers/end-users. If I had to estimate this lag, I would expect it to be 12+ months. We can a look at how long it takes for versions to be adopted to get some sense of this lag. So it might not get the impact we hope. 

Personally 👍 to fix regressions, and 👍 to focus more on the path forward ("focus" as saying no to the legacy APIs). For the pickers, specifically, I think that it would be great to iterate enough one it to reach the point where we are confident that once we make it stable, we can resist breaking changes long enough to reach >80% of adoption of v6.x. I don't think that we had this for a very long time, the versions has been spread.

stem-ui:
This update introduced a breaking change when using Luxon.... Luxon equality takes into consideration the zone on its DateTime object. If I change the zone (e.g. through an accompanying zone selector component), this new equality check passes instead of fails, resulting in the rendered input string not updating.... any suggested workarounds?

PR #6621, opened 2022-10-24
[core] Change reproduction position

oliviertassinari:
We have tried the change https://github.com/mui/material-ui/pull/34279 in core and toolpad, it seems to be a step forward, a proposal to do the same here.


oliviertassinari:
We have used this in Core and Toolpad for some time now, it seems to have gone smoothly. It's more enjoyable.

PR #6074, opened 2022-09-08
[pickers] Re-fix mobile picker not opening on label click

LukasTy:
Fixes #6063 

Revert the change introduced in #5651 as it introduced a regression due to `onMouseUp` event registered on `Clock` component.

With this change we introduce an additional `onClick` listener on the root element to overcome the issue of `click` event not being registered when clicking on `label` element due to DOM structure and the fact, that `TextField` has a `label` with `pointer-events: none`

PR #6645, opened 2022-10-27
[DateRangePicker] Fix input focused style and mobile behaviour

### LINKED ISSUE #6627: [DateRangePicker] on mobile opens keyboard after selecting date
### Duplicates

### Latest version

### Current behavior 😯

Selecting the date range from the calendar and pressing "OK" automatically focuses to the input and opens keyboard on my Android mobile device. 

### Expected behavior 🤔

Keyboard should not automatically open up. This is the behavior with DatePicker component for example. 

### Steps to reproduce 🕹

Link to live example:
Any example from the documentation: https://mui.com/x/react-date-pickers/date-range-picker/

Steps:
1. Open date selection from the input
2. Select dates and press OK

### Context 🔦

Automatically opening the keyboard after selecting date range is bad UX. 

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

51665

LukasTy:
Fixes #6627 

- fix `focused` prop declaration to allow for `focused` styles application both on tab-focus and when picker is open
- add `readonly` prop to `inputProps`  in mobile date range picker to align the behaviour with regular Date Picker and avoiding opening keyboard on mobile when value is selected

### DIFF
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerInput.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerInput.tsx
@@ -144,7 +144,10 @@ export const DateRangePickerInput = React.forwardRef(function DateRangePickerInp
     lazyHandleChangeCallback([start, date], inputString);
   };
 
-  const openRangeStartSelection = () => {
+  const openRangeStartSelection = (
+    event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
+  ) => {
+    event.stopPropagation();
     if (setCurrentlySelectingRangeEnd) {
       setCurrentlySelectingRangeEnd('start');
     }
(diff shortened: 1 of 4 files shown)


alexfauquette (commented):
I don't know why, but it seems to bring a regression. On mobile when clicking on the label, the view does not open (The bug you fixed  in #5651)

I appears in the preview but not in the doc

LukasTy:
> I don't know why, but it seems to bring a regression. On mobile when clicking on the label, the view does not open (The bug you fixed in #5651)

Actually, the correct fix is in https://github.com/mui/mui-x/pull/6074. Will need to replicate something similar. 🤔

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerInput.tsx
```diff
@@ -144,7 +144,10 @@ export const DateRangePickerInput = React.forwardRef(function DateRangePickerInp
     lazyHandleChangeCallback([start, date], inputString);
   };
 
-  const openRangeStartSelection = () => {
+  const openRangeStartSelection = (
+    event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
```
flaviendelangle, line 148: The following is enough if you just want to stop the propagation.
But if you think in the future we might use the event for other things, the more detailed typing can make sense

```suggestion
    event: React.UIEvent
```

LukasTy: Yeah, I was debating about using a more common type, but IMHO—it seems a bit more logical to be explicit. Especially when we know what kind of events might be passed in. 🤔

flaviendelangle (approved):

flaviendelangle:
I'll make sure that the new range pickers also have this behavior :+1:

PR #6682, opened 2022-10-30
[docs] Update material.io links to the right version

oliviertassinari:
Material Design has changed the structure of its docs to have clear versions.

### DIFF
--- a/docs/data/date-pickers/getting-started/getting-started.md
+++ b/docs/data/date-pickers/getting-started/getting-started.md
@@ -3,7 +3,7 @@ product: date-pickers
 title: Date picker, Time picker React components
 components: DatePicker,DateTimePicker,TimePicker
 githubLabel: 'component: DatePicker'
-materialDesign: https://material.io/components/date-pickers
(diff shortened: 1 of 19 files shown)


docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -3,7 +3,7 @@ product: date-pickers
 title: Date picker, Time picker React components
 components: DatePicker,DateTimePicker,TimePicker
 githubLabel: 'component: DatePicker'
```
LukasTy, line 5: Nitpick, but a `component: pickers` would probably be a better label for such page.
We use it as a common pickers label.

In any case, none of these values are actually used in any way as of now.
We are missing the following section from these `md` files 🤷 

```md
```

LukasTy (approved):

docs/data/date-pickers/getting-started/getting-started.md
oliviertassinari: @LukasTy Agree, 👍 for a quick follow-up to match with https://next.mui.com/x/react-data-grid/.

PR #6684, opened 2022-10-30
[docs] Unify a few of the capitalizations

oliviertassinari:
I was browsing the docs for something else and saw it wasn't matching https://www.notion.so/mui-org/Style-conventions-for-MUI-components-and-products-dff7d2d7d1ba4a4abd47f48cf345b800 (in place since about a month ago) so I did a PR to fix a few of the instances. I didn't try to cover everything.


samuelsycamore (approved):
Much better!

PR #6697, opened 2022-10-31
[datagrid] Opt-out for column jump back on re-order

gavbrennan:
Relates to https://github.com/mui/mui-x/issues/6644


gavbrennan:
> Thanks! Could you run `yarn proptypes` and commit the result? It will make the CI green.

Done.  Now stuck on the docs generation?

PR #6715, opened 2022-11-02
[fields] Support RTL out of the box

alexfauquette:
Follow up of #6571


alexfauquette:
The current strategy is to split the format when there is a space, and consider that as a group of sections we force it to be left to right. In those groups, each section s isolated with direction automatically deduced from its content

For example:

`yyyy/MM hh:mm` becomes
```
<ltr>
	<auto>yyyy</auto>/<auto>MM</auto>
</lt>{" "}
<ltr>
	</auto>hh</auto>:<auto>mm</auto>
</lt>`

(I added some spaces and line breaks to simplify the reading)


This forces sections to stay in the same position, whatever the content (digit, ltr, or rtl text) is. Moreover, it ensures we can compute the order in which elements are visible.


The management of the right/left arrows probably needs to be redesigned. For now it's a bit nasty because I did not found where to put it

Here is a code sandbox to experiment behavior

date-fns-jalali : [full screen](https://8xs2xr.csb.app/) | [with code](https://codesandbox.io/s/gallant-mcclintock-8xs2xr?file=/demo.tsx)

**Comment by flaviendelangle** [comment]
@alexfauquette is this ready for review ?
It is not marked as draft but has no review aksed

File: packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts [hunk 1]
```diff
@@ -656,3 +700,55 @@ export const clampDaySection = <TDate, TSection extends FieldSection>(
     };
   });
 };
+
+export const getSectionOrder = (
```
alexfauquette, line 704:
```suggestion
/**
 * Return a description of sections order from left to right.
 **/
export const getSectionOrder = (
```

LukasTy:
Tremendous work on this part. 🚀 
RTL was an oversight initially and, I take, a very painful aspect to implement for fields. 🙈 💯 

With `AdapterDateFnsJalali` it seems to work really nicely. 👌 

A few general notes:
1. Is this randomly appearing vertical marker an oversight? Or an unavoidable side-effect of using the special characters?

https://user-images.githubusercontent.com/4941090/204653866-f18124e3-c182-459b-8e04-883f4a2af3bc.mov

2. Is only `AdapterDateFnsJalali` supported for now or am I doing something wrong?
[In this example](https://codesandbox.io/s/mui-x-fields-fix-rtl-hcmkus?file=/demo.tsx) both moment libraries are "not working in many ways".

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
```diff
@@ -249,6 +251,12 @@ export interface FieldValueManager<TValue, TDate, TSection extends FieldSection,
    * @returns {boolean} `true` if the current error is not empty.
    */
   hasError: (error: TError) => boolean;
+  getSectionOrder: (
```
flaviendelangle, line 254: Can we add some JSDoc here ?

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
```diff
@@ -99,6 +99,8 @@ export type UseFieldResponse<TForwardedProps extends UseFieldForwardedProps> = O
 export interface FieldSection {
   start: number;
```
flaviendelangle, line 100: Can we add some JSDoc on the 4 position attribute ?

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -364,6 +392,13 @@ export const splitFormatIntoSections = <TDate>(
 
   return sections.map((section) => {
     if (section.separator !== '/') {
+      if (section.separator !== null && section.separator.includes(' ')) {
+        return {
+          ...section,
+          separator: section.separator.replace(' ', '\u2069 \u2066'),
```
flaviendelangle, line 398: If we have several spaces, should it wrap each of them with `\u2069` and `\u2066` ? Or all of them at once ? Or only the first one (which is the current behavior)

flaviendelangle (commented):
Looks great !

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
alexfauquette: I did not consider this case. Maybe the more generic solution is to put one ate the beginning and one at the end of the separator

alexfauquette:
> Is this randomly appearing vertical marker an oversight? Or an unavoidable side-effect of using special characters?

Probably an error when computing the start/end indexes of the selected characters. I will have a look

> Is only `AdapterDateFnsJalali` supported for now or am I doing something wrong?

Yes, I've faced problems with moment-hijri because there is inconsistencies between parse and format. Not sure if it's a bug or misuse. I opened an issue, but no answer: https://github.com/xsoh/moment-hijri/issues/83

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: Maybe

I think two spaces is something that can occur, especially if people use formats like `DD - MM - YYYY``

LukasTy:
> > Is this randomly appearing vertical marker an oversight? Or an unavoidable side-effect of using special characters?
> 
> Probably an error when computing the start/end indexes of the selected characters. I will have a look

I can confirm that this issue is fixed with latest changes. 🙌 👍 

> > Is only `AdapterDateFnsJalali` supported for now or am I doing something wrong?
> 
> Yes, I've faced problems with moment-hijri because there is inconsistencies between parse and format.

Gotcha. And what about `moment-jalaali`? Same issue?

alexfauquette:
> Gotcha. And what about moment-jalaali? Same issue?

Yes, it has the same issue. For example key <kbd>Arrow Up</kbd> moved from `MM` to `01` but not from `01` to `02`

And this does not occur with `moment` 🤷

flaviendelangle:
> For example key Arrow Up moved from MM to 01 but not from 01 to 02

We have several problem to support the editing of moment plugins

1. The date-io `addMonths` and `addYear` adds Gregorian years instead of Jalaalian one (easy to fix)

2. `moment-jalaali` consider an incomplete date like `1401/MM/DD` to be valid, which breaks everything :grimacing:

flaviendelangle:
@alexfauquette I opened a PR on your repo (https://github.com/alexfauquette/mui-x/pull/7) to try fixing `moment-jalaali`

alexfauquette:
I confirm it fixes the problem

LukasTy:
> For simplicity, the current solution is adding Unicode character in both `ltr` and `rtl`. Do you think it worths the effort to avoid adding those invisible characters in `ltr` mode?

IMHO—fields implementation is already very complex as-is. I'd lean towards leaving this behaviour in both modes. Besides, it might allow us to spot any possible future issues, which would possibly be missed  (because we wouldn't be working with `rtl` often). 🙈 🤔

flaviendelangle:
> For simplicity, the current solution is adding Unicode character in both ltr and rtl. Do you think it worths the effort to avoid adding those invisible characters in ltr mode?

We already have invisible characters in LTR for the numeric editing anyway.
I would be in favor to keep all the invisible characters and add a small section in "Testing caveats" on how to clean the input value (maybe we could export a testing utility).

alexfauquette:
On `moment-hijri` it's a distinct problem.

The adapter also have the problem of adding gregorian months and not hijri ones. But their is a bigger problem linked to the library itself.

Let assume you have a field with `YYYY/02/DD`

When you press the key UP on months it roughly does the following:

```js
const initialMonth = utils.parse('02', 'iMM');
const newMonth =utils.addMonths(initialMonth, 1);
const new SectionValue = utils.format(newMOnth, 'iMM' );
```

The problem, with `moment-hijri` when you do `format(parse( '02', 'iMM', isStrict))` You can either get:
- `12` if `isStrict` is `false`
- `Invalid date` if `isStrict` is `true`

The example is in [the issue I opened](https://github.com/xsoh/moment-hijri/issues/83)

alexfauquette:
Good for me

flaviendelangle (approved):

PR #6722, opened 2022-11-03
[pickers] Rename `inputFormat` prop `format` on the new pickers

flaviendelangle:


packages/x-date-pickers/src/internals/models/props/basePickerProps.tsx
```diff
@@ -72,7 +72,7 @@ export interface BaseNextNonStaticPickerProps {
   /**
    * Format of the date when rendered in the input(s).
```
LukasTy, line 73:
```suggestion
   * Format of the date when rendered in the input(s).
   * Defaults to localized format based on the used `views`.
```

LukasTy (approved):

alexfauquette (approved):

PR #6724, opened 2022-11-03
[docs] Improve demo about customizing pagination

### LINKED ISSUE #6721: [data grid] Mui Datagrid does not have numbered pagination?
### Duplicates

### Latest version

### Steps to reproduce 🕹

https://codesandbox.io/s/f7t5uf?file=/demo.tsx

### Current behavior 😯

In DataGrid, for pagination there is only left and right arrow. I needed complete numbered pagination with arrows, but it seems like there is no option for it.

### Expected behavior 🤔

https://codesandbox.io/s/2slrb5?file=/demo.tsx
In this demo, there is way to have numbered pagination, but here total number of pages and records count information is missing.
What I want is usual information like total number of pages and records count with numbered pagination.
Is there any way for this?

### Context 🔦

_No response_

### Your environment 🌎

_No response_

m4theushw:
Closes #6721 

Since the demo in https://mui.com/x/react-data-grid/components/#pagination replaces completely the pagination component, it also removes the page size control and the number of rows. I changed to reuse `GridPagination` and only replace the buttons to navigate between pages. I think this better aligns with what users expect.


MBilalShafi (approved):
Nice addition to docs 👍

PR #6731, opened 2022-11-04
[pickers] Support the `onError` prop and add context on the `onChange` prop

flaviendelangle:
I entered a small rabbit hole trying to fix the `onError` missing calls so the PR is bigger than expected :grimacing: 

- [x] Call `onError` on the new pickers (handled by `usePickerValue`)
- [x] Add the `validationError` to the `onChange` context on the new pickers like on the fields
- [x] Move `isSameError` from the field value manager to the general value manager
- [x] Refacto the field value managers to avoid duplication, we now have one for single item values and one for range values
- [x] Refacto the general value managers to avoid duplication, we now have one for single item values and one for range values (and temporary one for the legacy time pickers but we should be able to drop it soon).


packages/x-date-pickers/src/internals/utils/valueManagers.ts
```diff
@@ -0,0 +1,53 @@
...
+export const singleItemValueManager: PickerStateValueManager<
+  any,
+  any,
+  DateValidationError | TimeValidationError | DateTimeValidationError
+> = {
+  emptyValue: null,
+  getTodayValue: (utils) => utils.date()!,
+  cleanValue: replaceInvalidDateByNull,
+  areValuesEqual: (utils, a, b) => utils.isEqual(a, b),
+  isSameError: (a, b) => a === b,
```
alexfauquette, line 22: Maybe you can remove the following line (same for time and date-time)
```ts
export const isSameDateError = (a: DateValidationError, b: DateValidationError) => a === b;
```

alexfauquette (approved):
Sounds good

Now we will be able to add an expect on `onError` in the two `testTextFieldValidation.tsx` files

packages/x-date-pickers/src/internals/utils/valueManagers.ts
flaviendelangle: It is still used in the `useDateValidation` (that will be dropped once we remove the legacy pickers I think)

PR #6763, opened 2022-11-07
[DateRangePicker] Add dragging support to edit range

### LINKED ISSUE #5311: [DateRangePicker] Drag date marker to edit a range
### Duplicates

### Latest version

### Summary 💡

If you can pick the right dates in the first three clicks everything is perfect. 

The pain starts when you want to edit the `To` date.
On the current behavior, the 4th click is considered a new `From` date, and now you have to fix `From` and `To`, which would require again 2 more clicks. Basically to edit the `To` date, you need to reset the full range.

https://mui.com/x/react-date-pickers/date-range-picker/#static-mode

### Examples 🌈

https://react-spectrum.adobe.com/react-spectrum/RangeCalendar.html

### Motivation 🔦

Smoothier user experience.

### Order ID 💳 (optional)

_No response_

LukasTy:
Resolves #5311 

Add option to edit selected date range by dragging range `start` or `end` dates.

Allows:
- extending range
- reducing range
- dragging `start` position past and vice versa (range flipping)
- drag editing on mobile (uses touch events)

Uses same styling as selected range.

Available only on `DateRangeCalendar`—which is going to be used for the new `DateRangePicker` in v6.

Dragging is cancelled when:
- <kbd>Esc</kbd> is clicked
- `drop` is executed on disabled date
- `drop` is executed outside of the calendar

Outstanding questions:
- Does range preview functionality still make sense?
- Is using same styling as for selection/highlighting good enough?

An example of the implementation [is available here](https://codesandbox.io/s/admiring-currying-gsfxxc?file=/demo.tsx) as well as in the [new component docs](https://deploy-preview-6763--material-ui-x.netlify.app/x/react-date-pickers/date-range-picker/).

Also [made sure that react-dnd](https://codesandbox.io/s/nervous-architecture-9hmxhq?file=/src/Demo.tsx) wouldn't cause issues (we initially had problems with data grid column reordering).


joserodolfofreitas (commented):
`DragEditing` feels great!

And nice touch with the "fade out" applied to the "origin day" of the dragging marker. I noticed, though, that it needs a further step to finish the effect:

https://user-images.githubusercontent.com/550141/201469613-32e2c059-e76f-4496-9a9d-f6cbd5972c7a.mov

I also like the combination of the grey and the blue circle for the "origin day" style (pic below), but it doesn't seem consistently applied on every drag interaction (sometimes it seems the grey circle is rendered below/before the blue, so the effect is different).

I understand that working with both dates simultaneously was quite challenging. Are there other implications to get this effect working consistently?

image

flaviendelangle:
Not sure why but on my Computer (latest Chrome, latest Ubuntu) I have an ugly white square behind the dragging day

https://user-images.githubusercontent.com/3309670/202145086-9b52ed42-d1cc-4887-b51d-7f1ea0ad5ec7.mp4

React Spectrum has a different approach which feels a lot more polished on my computer

https://user-images.githubusercontent.com/3309670/202145352-7ab3ea27-dce9-4427-9bd5-99b9fc901b94.mp4

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -101,6 +103,7 @@ function useDateRangeCalendarDefaultizedProps<TDate>(
     minDate: applyDefaultDate(utils, themeProps.minDate, defaultDates.minDate),
     maxDate: applyDefaultDate(utils, themeProps.maxDate, defaultDates.maxDate),
     calendars: themeProps.calendars ?? 2,
+    disableDragEditing: themeProps.disableDragEditing ?? false,
```
flaviendelangle, line 106: You can add this prop to `DateRangeCalendarDefaultizedProps`

packages/x-date-pickers-pro/src/DateRangePicker/date-range-manager.ts
```diff
@@ -6,27 +6,37 @@ interface CalculateRangeChangeOptions<TDate> {
   range: DateRange<TDate>;
   newDate: TDate;
   currentlySelectingRangeEnd: 'start' | 'end';
+  allowRangeFlip?: boolean;
```
flaviendelangle, line 9: Could you document this param ? It's goal is not obvious from first read even if I think I get why it has been added

packages/x-date-pickers/src/DateCalendar/DayCalendar.tsx
```diff
@@ -223,6 +223,7 @@ function WrappedDay<TDate extends unknown>({
       role: 'gridcell',
       isAnimating: isMonthSwitchingAnimating,
       selectedDays,
+      'data-timestamp': utils.toJsDate(day).valueOf(),
```
flaviendelangle, line 226: Seems like you are not using it anywhere.

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
```diff
@@ -0,0 +1,197 @@
...
+    if (isSameAsDraggingDate(event)) {
+      return;
+    }
+    const newDate = resolveDateFromTarget(event.target, utils);
+    if (newDate) {
+      onDrop(newDate);
+    }
+  });
+
+  return React.useMemo(
```
flaviendelangle, line 167: You are spreading the response from this hook in `DateRangeCalendar` so the `useMemo` has no effect.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -322,6 +350,35 @@ const DateRangeCalendar = React.forwardRef(function DateRangeCalendar<TDate>(
...
+    handleSelectedDayChange(newDate, undefined, true);
+  });
+
+  const handleRangeDragDayChange = useEventCallback((val: TDate | null) => {
+    if (!utils.isEqual(val, rangeDragDay)) {
+      setRangeDragDay(val);
+    }
+  });
+
+  const shouldDisableDragEditing = React.useMemo(
```
flaviendelangle, line 369: I don't think this `useMemo` is usefull, the computation is super light

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -331,15 +388,39 @@ const DateRangeCalendar = React.forwardRef(function DateRangeCalendar<TDate>(
     ...componentsProps,
     day: (dayOwnerState) => {
       const { day } = dayOwnerState;
+      const isSelectedStartDate = isStartOfRange(utils, day, valueDayRange);
```
flaviendelangle, line 391: This part becomes heavy and hard to read, maybe we could improve it later.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -289,6 +300,7 @@ const DateRangeCalendar = React.forwardRef(function DateRangeCalendar<TDate>(
   const maxDateWithDisabled = (disabled && value[1]) || maxDate;
 
   const [rangePreviewDay, setRangePreviewDay] = React.useState<TDate | null>(null);
+  const [rangeDragDay, setRangeDragDay] = React.useState<TDate | null>(null);
```
flaviendelangle, line 303: Maybe I'm missing something, but personally I would move this start inside `useDragRange` and return the value from the hook instead of passing the setter to the hook.

That way we encapsulate all of the logic

flaviendelangle (commented):
I like the overall structure

Concerning the preview, when is it visible now that we have the drag ?
When the drag is disabled with `disableDragEditing` ?
It might become used again once this component supports keyboard editing.
But maybe we could merge its state with the drag one in some way (I did not dive deep enough into your code to know if it would work).

alexfauquette:
About the white background of the dragged image, I see it too. It comes from how Chrome compute the preview image. 

Since seeing the day you are dragging is not usefull, you could keep the code structure and replace the preview with an empty image thanks to [setDragImage](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setDragImage#examples).

https://stackoverflow.com/questions/7680285/how-do-you-turn-off-setdragimage

alexfauquette:
A bit out of context, but it's more visible when drag/drop is activated: We should not disable the start/end days because it leads to strange behavior when the disabled days are not the same for start/end

https://codesandbox.io/s/zealous-dirac-hku5uc?file=/demo.tsx

LukasTy:
> Concerning the preview, when is it visible now that we have the drag ?
> When the drag is disabled with `disableDragEditing` ?

The preview behaviour was not changed one bit.
It is working in desktop mode and allows extending or shrinking (no visual indication) date range for the `currentDatePosition`.

> But maybe we could merge its state with the drag one in some way (I did not dive deep enough into your code to know if it would work).

It's not trivial. Preview range styling uses a different function to calculate the range. Basically, it wouldn't reduce the computation count doing this.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
LukasTy: Could you expand on this point?
It's added to `DateRangeCalendarProps` and is hence propagated to `DateRangeCalendarDefaultizedProps` and initialized here in `useDateRangeCalendarDefaultizedProps` 🤔

flaviendelangle: Unless I missread, `DateRangeCalendarDefaultizedProps['disableDragEditing']` has a type of `boolean | undefined` when in fact it is `boolean`.

We just have to add it to the list of defaultized props in this type (like `calendars` for instance)

It's clearly a detail, but it will avoid having future developers add a `props.disableDragEditing!` before passing it to a function that expects a boolean for instance.

I had several of those hacks for validation props in the past before cleaning the typing.

LukasTy: Thanks for explaining. Will do. 😉

packages/x-date-pickers/src/DateCalendar/DayCalendar.tsx
LukasTy: But you make a good point. I'll add a comment about how and why it is used. 😉

flaviendelangle: Oh I forgot that the imperative API was named `dataset.*` :see_no_evil:

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
LukasTy: I've structured it like this to have the least amount of frequently changing variables being shared among component and hook in order to facilitate least amount of hook result re-creations.
But I've experimented and I think splitting the hook into a sub-hook only for event handling helps in this case. 🤔

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
LukasTy: Changed it. Now it's no longer being spread in the wrapping hook and stops the hook from re-creating on each render. 👌

packages/x-date-pickers-pro/src/DateRangePicker/date-range-manager.ts
LukasTy: Added JsDoc

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
LukasTy: Updated: https://github.com/mui/mui-x/pull/6763/commits/d388bcd293d7d32f82605ea4c0d7c71892d05e4e

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
```diff
@@ -169,6 +169,8 @@ const useDragRangeEvents = <TDate>({
     event.stopPropagation();
     setIsDragging(false);
     setRangeDragDay(null);
+    // make sure the focused element is the element where drop ended
+    event.currentTarget.focus();
```
LukasTy, line 173: @joserodolfofreitas This should fix your mentioned issue of focus remaining on the previously selected date—where the drag started.

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
```diff
@@ -226,7 +226,7 @@ const DateRangePickerDayDay = styled(PickersDay, {
   ...(!ownerState.selected && {
     backgroundColor: 'transparent',
   }),
-  ...(!ownerState.isHighlighting && {
+  ...(!ownerState.isHighlighting && !ownerState.isDragging && {
```
LukasTy, line 227: @joserodolfofreitas This should have fixed the issue you noticed with grey border being applied to the drag start day when you drag away from it.

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
```diff
@@ -136,12 +145,10 @@ const DateRangePickerDayRoot = styled('div', {
   ...((ownerState.isStartOfHighlighting || ownerState.isStartOfMonth) && {
     ...startBorderStyle,
     paddingLeft: 0,
-    marginLeft: DAY_MARGIN / 2,
```
LukasTy, line 139: I've removed these styles to fix the following inconsistency (currently available, but less noticeable while dragging is not supported) when the highlighting range changes.
Do you @alexfauquette @flaviendelangle  see any potential issues with it? 🤔 

https://user-images.githubusercontent.com/4941090/203561310-62940c7d-c30e-4766-8622-0b1028991017.mov


https://user-images.githubusercontent.com/4941090/203561336-0480cc31-4b59-469c-8722-29ff2d51d910.mov

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -132,7 +133,11 @@ const DateRangeCalendar = React.forwardRef(function DateRangeCalendar<TDate>(
 
   const ownerState = props;
   const classes = useUtilityClasses(ownerState);
-  const isMobile = React.useContext(WrapperVariantContext) === 'mobile';
+  const wrapperContext = React.useContext(WrapperVariantContext);
+  // defaults to `true` in environments where `window.matchMedia` would not be available (i.e. test/jsdom)
+  // needed as a fallback when this component is used standalone and there is no `WrapperContext`
+  const isDesktop = useMediaQuery('@media (pointer: fine)', { defaultMatches: true });
+  const isMobile = wrapperContext ? wrapperContext === 'mobile' : !isDesktop;
```
LukasTy, line 140: What do you guys @alexfauquette @flaviendelangle  think? Should we do it?
Currently when `DateRangeCalendar` is used standalone this always resolves to `false` and even in mobile would produce a buggy range previewing behavior... 😢

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
```diff
@@ -228,8 +228,10 @@ const DateRangePickerDayDay = styled(PickersDay, {
   }),
   ...(!ownerState.isHighlighting &&
     !ownerState.isDragging && {
-      '&:hover': {
-        border: `1px solid ${(theme.vars || theme).palette.grey[500]}`,
+      '@media (pointer: fine)': {
```
LukasTy, line 231: Found this with as a solution to avoid unwanted hover styles on mobile being stuck on previously clicked date (where touch did not start).

LukasTy:
> A bit out of context, but it's more visible when drag/drop is activated: We should not disable the start/end days because it leads to strange behavior when the disabled days are not the same for start/end

Are you suggesting to `disableDragEditing` when we have `shouldDisableDate` prop provided? 🤔 
This is actually a really great issue you discovered. I really did not think about it...
Will look into the option of updating the `currentDatePosition` when dates are dragged over to facilitate correct disabled days resolution. 🤔

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
LukasTy: Did you not see the weekdays jumping in width in the first video when we are dragging in them? 🤔

alexfauquette: Amazing, I don't know how I made to miss it 👀

flaviendelangle: It's great to fix these jumping
Did you see any down side of removing this margin ?

LukasTy: I'll do one more testing round with different variations (mobile, 1 and several calendars). 😉

flaviendelangle:
I no longer have the dragging image, that's great !

We are pretty close from React Spectrum has.
Just our performances are catastrophic in comparison, but discussed it and we can work it on a follow up PR (it's not purely related to the dragging).

I just have one behavior that I prefer on Spectrum: they don't keep the initial date highlighted when dragging it. 

https://user-images.githubusercontent.com/3309670/203743814-997dccad-1321-461a-8b0e-0b54102fbccb.mp4

vs

https://user-images.githubusercontent.com/3309670/203744059-d73e0af7-e954-41ea-95c1-8f6f3d878874.mp4


___

Also the start highlighted date is not vertically centered

image

But the bug is already on next so we can fix it after

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
LukasTy: Seems like this is a fix we needed anyways.
All instances of calendars seems to be missing this alignment adjustment.

**Mobile before:**
mobile-before
**Mobile after:**
mobile-after
**Desktop before:**
desktop-before
**Mobile after:**
desktop-after

LukasTy:
> Also the start highlighted date is not vertically centered
But the bug is already on next so we can fix it after

We might have had this issue for quite some time. 🤷 
It is only happening when using components without explicit theme in playground (which does have certain theme providers already).

LukasTy:
> I just have one behavior that I prefer on Spectrum: they don't keep the initial date highlighted when dragging it.

Maybe I'm biased, but I do prefer our behavior—where the currently still present selection is visible. I do prefer it just because while dragging—nothing actually happens, we do not get any change events and only after a successful drop or touch end does the date range actually change.
But that is of course an open question and I'd really like to hear others' opinions. 🤔

flaviendelangle:
Maybe @gerdadesign will have an opinion here

It's of course very subjective

joserodolfofreitas (commented):
Nice work polishing the feature!

Can we keep the latest day when the drop doesn't end in a specific day? 
I think it's fine to rollback the value if the drop ends outside of the calendar, but users might miss the drop for only a few pixels, and it will rollback.
I think the video below can better explain the problem.

https://user-images.githubusercontent.com/550141/203953786-7028905c-3964-4119-a71f-b15f2700009d.mp4

Also, I noticed there's an inconsistency with the mouse pointers; 
Sometimes it shows the dragging hand, and sometimes it remains as an arrow. Would this be something quick to fix?

-----
>Maybe I'm biased, but I do prefer our behavior—where the currently still present selection is visible. I do prefer it just because while dragging—nothing actually happens, we do not get any change events and only after a successful drop or touch end does the date range actually change. 
But that is of course an open question and I'd really like to hear others' opinions

I think the original day is a piece of additional information that adds to usability because it makes it easier for users to trace back their steps. But it can also be seen negatively when we consider that every piece of information competes for the user's attention. Considering that, perhaps it shouldn't hold the same accent color as the two primary markers.

@LukasTy mentioned that having two separate styles may not be feasible. So I was wondering, if we could simply omit the original range value, and let only the "click" style (the gray circle that remains after you click) signal the original day. 

Additionally, could we consider this to become a prop for behavior customization? 
Something like `displayOriginalDayWhenEditDragging`? (suggestions for other names are very welcomed)

gerdadesign (commented):
This is looking really good! 

I agree that it would be nice to have a secondary styling for the origin (perhaps a lower opacity background could be an option — so that as you're moving the market it leaves a "ghost impression" of where it has been?). Although if it's very difficult to do, I think this is already in a great position!  

The only other thing I see is on desktop, the hover style seems to stay on the origin date until I move my mouse. 


https://user-images.githubusercontent.com/106991024/204165318-59dca147-00f7-4d79-a1e3-79816f5f8f8e.mov

LukasTy:
> A bit out of context, but it's more visible when drag/drop is activated: We should not disable the start/end days because it leads to strange behavior when the disabled days are not the same for start/end

@alexfauquette [This commit](https://github.com/mui/mui-x/pull/6763/commits/0185ff94c45b8be30d9d32db4b8337599b4a5ec4) should have addressed the issue you have discovered with the `shouldDisableDate` not reacting correctly to the range flipping.
Let me know if you still think it needs further improvement. 👌

flaviendelangle:
> I think this is already in a great position!

Maybe we can improve it, but I do agree that in it's current shape it is already very good for a first version :+1: 
I'm totally fine with going with this design and iterate based on the user feedbacks.

LukasTy:
> Can we keep the latest day when the drop doesn't end in a specific day?
I think it's fine to rollback the value if the drop ends outside of the calendar, but users might miss the drop for only a few pixels, and it will rollback.

It's not easily achievable with current implementation. We are getting a `dragEnd` event whenever a drag ends, but it has no info on where it actually landed. We could only try computing it from `clientX` and `clientY` coordinates, but even then—I don't see a way to differentiate a case when the drag ended by releasing the mouse or by cancelling it (i.e. Esc) 🙈 

After discovering an seeing all this trouble and edge cases—I understand why Spectrum has just chosen to go with pointer events.

___

> I think the original day is a piece of additional information that adds to usability because it makes it easier for users to trace back their steps. But it can also be seen negatively when we consider that every piece of information competes for the user's attention. Considering that, perhaps it shouldn't hold the same accent color as the two primary markers.

I've added `opacity` to the origin date of `'.6'` to reduce it's significance.

LukasTy:
> Also, I noticed there's an inconsistency with the mouse pointers;
Sometimes it shows the dragging hand, and sometimes it remains as an arrow. Would this be something quick to fix?

I have noticed it a few times as well, but honestly—I can't understand the reason behind it. The style (`cursor: 'grabbing'`) is always applied, but sometimes it simply doesn't take effect. 🤷 
I think that it might also be an issue with browsers implementation as well. I could hardly reproduce it on Safari. 🤔

LukasTy:
I have experimented with using `pointer` events to have a single point of logic handling both desktop and mobile, but it has its own limitations.
For example: no way to correlate the origin and the target—thus having no way to check if the drag is ending on the origin—no need to update.
It also does not solve the performance issue in any way.

___

I've stashed the experiment if we'll discover a reason to use different implementation in the future, but in essence it ends up being basically the same implementation as `touch` events version.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
flaviendelangle: We must just make sure that SSR does not break.
But unless I'm missing a scenario, we are good with your approach.
On the server`rangePreviewDay === null` so `previewingRange === [null, null]`.
And if `previewingRange` is empty, then `isWithinRange`, `isStartOfRange` and `isEndOfRange` returns `false`.
So all the ternary are equal to `false` no-matter what `isMobile` equals on the 1st render.

___

Just, if we are doing that, I would change the ternaries to be `isDesktop ? isXXXX() : false`, to avoid having both `isMobile` and `isDesktop` variables defined.

flaviendelangle (approved):
I propose to merge it, to start to play with it an come back to it later if needs be

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
LukasTy: Yeah, it should not impact SSR or tests as it's the similar approach we use for many components.
How would you write this condition differently? 🤔 Am I missing something, but I don't seem to come-up with a more simplistic approach. 🙈 

`isDesktop` here will be used on when this component is used standalone and no `context` is available.
On the other hand—all our internal components having `React.useContext(WrapperVariantContext)` are behaving like this and maybe it's a bit pointless introducing this somewhat pointless optimisation, which does not improve performance in any perceivable way? 🤔

flaviendelangle: > How would you write this condition differently? thinking Am I missing something, but I don't seem to come-up with a more simplistic approach.

Maybe

```tsx
const isDesktopMediaQuery = useMediaQuery('@media (pointer: fine)', { defaultMatches: true });
const isDesktop = wrapperContext ? wrapperContext === 'desktop' : isDesktopMediaQuery
```

My goal is just to avoid using `isDesktop` by mistake (since it would lead to buggy behavior)

LukasTy: Fair point. Initially ` : false` (the else) case confused me.

But my question still stands—do you think it's worth introducing this fix/optimisation only to this component, when we have quite a few others, which could also potentially be used without `wrapperContext`.
After thinking about this—I' m leaning towards all or none solution instead of just here. 🙈 🤷

flaviendelangle: > I' m leaning towards all or none solution instead of just here. see_no_evil shrug

I agree, and I'd prefer the "all" solution :angel: 
But that can be in a follow up PR if you want

LukasTy: Updated with your suggestion, although, after thinking about it a bit more and checking other cases, where only the `variant` is being passed instead of being resolved into a boolean.
I'm leaning towards removing this edge-case handling—even if just for reducing the computations count. If we were to proceed with changing every case to have this edge-case handling—each component would additionally compute the media query on it's own, event though, it would probably be pointless in 99% of the cases. 🙈

LukasTy: I've [created an issue](https://github.com/mui/mui-x/issues/7064) to potentially address this concern in a more appropriate manner.

PR #6771, opened 2022-11-07
[test] Verrify `onError` call

alexfauquette:
Their is an inconsistency in range pickers. If the value is valid at first render, the following occurs:

- The following are calling `onError` with `[null, null]`
  - `<DateRangePicker />`
  - `<DesktopDateRangePicker />`
  - `<MobileDateRangePicker />`
  - `<SingleInputDateRangeField />`
- The following are not calling the `onError`
  - `<MultiInputDateRangeField />`
  - `<MultiInputDateTimeRangeField />`
  
  
@flaviendelangle I'm more in favor of the second option, not calling it, but in case this behavior is for a specific reason I prefer to ask before debugging


flaviendelangle:
I guess we don't have a call on single item pickers (`DatePicker` for instance).
So I would also go for option 2 as it is coherent.

packages/x-date-pickers-pro/src/internal/hooks/useMultiInputRangeField/useMultiInputDateRangeField.ts
```diff
@@ -103,7 +103,12 @@ export const useMultiInputDateRangeField = <TDate, TChildProps extends {}>({
     TDate,
     DateRangeValidationError,
     DateRangeComponentValidationProps<TDate>
-  >({ ...sharedProps, value }, validateDateRange, () => true);
+  >(
+    { ...sharedProps, value },
+    validateDateRange,
+    rangeValueManager.isSameError,
```
flaviendelangle, line 109: I totally forgot to pass `isSameError` :facepalm:

flaviendelangle (commented):
Nice catch !

So if I understand correctly, now if the value is valid (error `null` on single item and `[null, null]` on range), then no picker or field should call `onError` on mount ?

alexfauquette:
@flaviendelangle Yes, all components have a default error value (`null` or `[null, null]`) the `onError` is called if the error value is modified. In the special case of the mount, modified means different from default value

flaviendelangle (approved):

PR #6777, opened 2022-11-08
[pickers] Improve validation JSDoc descriptions

flaviendelangle:


packages/x-date-pickers/src/internals/hooks/validation/models.ts
```diff
@@ -24,13 +24,13 @@ export interface BaseTimeValidationProps extends FutureAndPastValidationProps {}
  */
 export interface TimeValidationProps<TDate> {
   /**
-   * Min time acceptable time.
-   * For input validation date part of passed object will be ignored if `disableIgnoringDatePartForTimeValidation` not specified.
+   * Minimal time acceptable time.
```
LukasTy, line 27:
```suggestion
   * Minimal selectable time.
```

packages/x-date-pickers/src/internals/hooks/validation/models.ts
```diff
@@ -24,13 +24,13 @@ export interface BaseTimeValidationProps extends FutureAndPastValidationProps {}
...
+   * The date part of the object will be ignored unless `props.disableIgnoringDatePartForTimeValidation === true`.
    */
   minTime?: TDate;
   /**
    * Max time acceptable time.
```
LukasTy, line 32:
```suggestion
   * Maximal selectable time.
```
Just to align with picker descriptions

LukasTy (approved):

flaviendelangle:
:man_facepalming: I think I found for the generic, the regression was introduced in #6731
That makes me think that I did not copy the `.spec` files which would have easily caught it.
I will rework them and add them back on all the public components (or if someone else wants to have a go at it)

PR #6778, opened 2022-11-08
[pickers] Support `theme.vars`

### LINKED ISSUE #6711: [pickers] Support CSS theme variables
Part of  #6702 for the pickers

Marking it as high since it's blocking @siriwatknp

alexfauquette:
Fix #6711


packages/x-date-pickers-pro/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import type {} from '@mui/material/themeCssVarsAugmentation';
```
siriwatknp, line 1:
```suggestion
```

This **should not** be enabled by the library, it should be done by the consumer instead.

Can you try importing it into the MUI X global project instead?

alexfauquette: what do you mean by
> importing it into the MUI X global project

cherniavskii: Without it, TypeScript complains about `theme.vars` as it's not available on the standard `Theme` interface.
Can we solve it somehow without importing `themeCssVarsAugmentation`?

flaviendelangle: Even if we fix it on our side, make sure to test a bundled version of our package with `skipLIbCheck: false` because I suspect it will fail without this import.

Doesn't it make sense for `@mui/x-date-pickers` to import the module augmentation that it uses internally ?

siriwatknp: > Doesn't it make sense for @mui/x-date-pickers to import the module augmentation that it uses internally

I am open to any option but the developers have to opt-in to get the type. Meaning when they install `@mui/x-date-pickers` and `@mui/material`, they should not see `theme.vars`.

If they want to have `theme.vars`, they have to opt-in (either import from `@mui/material` or `@mui/x-date-pickers`).

alexfauquette: I struggle to understand how we could use `theme.vars` in our files, and hide the `theme.vars` type from users

packages/x-date-pickers/src/internals/utils/themeVar.ts
```diff
@@ -0,0 +1,11 @@
+import { Theme } from '@mui/material/styles';
+import type { CssVarsTheme, ThemeVars } from '@mui/material/styles/experimental_extendTheme';
+
+export type ExtendedTheme = Theme | CssVarsTheme;
+
+export const getThemePalette = (theme: ExtendedTheme): Theme['palette'] | ThemeVars['palette'] => {
```
flaviendelangle, line 6: I do like this solution a lot

alexfauquette: @siriwatknp If you're good with this solution I will do the same in the data-grid

siriwatknp: Screen Shot 2565-11-16 at 11 23 48

Seems to work.

siriwatknp: Confirmed, you need to add it to `tsconfig.build.json` too, otherwise, `yarn build` fails.

alexfauquette: clearly less hacky, and I confirm `yarn release:build` succeed

flaviendelangle: Seems like apps using `@mui/x-date-pickers` with this approach also works and don't have access to `theme.vars`

flaviendelangle (approved):

packages/x-date-pickers-pro/src/DateRangePickerDay/DateRangePickerDay.tsx
```diff
@@ -218,12 +220,16 @@ const DateRangePickerDayDay = styled(PickersDay, {
...
     },
   }),
   ...(!ownerState.selected &&
     ownerState.isHighlighting && {
-      color: theme.palette.getContrastText(alpha(theme.palette.primary.light, 0.6)),
+      color: theme.palette.getContrastText(
+        theme.vars
+          ? `rgba(${theme.vars.palette.primary.lightChannel} / 0.6)`
+          : alpha(theme.palette.primary.light, 0.6),
+      ),
```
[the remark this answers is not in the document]
siriwatknp: Even though it does not use CSS variables but if the value is the same, then the perfect dark mode works on SSR.

siriwatknp (changes requested):
Left one last comment, the rest looks good.

siriwatknp (approved):
🤩 Can't wait for this!

PR #6781, opened 2022-11-08
[docs] Remove default prop and improve format

oliviertassinari:
Based on my frustration exploring the docs for row grouping in https://mui.zendesk.com/agent/tickets/5620. I propose the following changes:

- Remove default prop (`rowGroupingColumnMode="single"`)
- Replace the future tense with the present tense, from the perspective of a reader, I think that it happens almost instantly. 
- Replace "we" with "it". When a developer reads the docs, I think that it's clearer for him to describe how the software behaves rather than how we intended for the software to behave.

Preview https://deploy-preview-6781--material-ui-x.netlify.app/x/react-data-grid/row-grouping/


flaviendelangle (commented):
Make sense to remove the default value of `rowGroupingColumnMode` in the doc
For the workings, they seem ok, but I know that we also try to avoid passive form usually

cherniavskii (approved):

docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -7,22 +7,24 @@ title: Data Grid - Row grouping
 <p class="description">Group your rows according to some column values.</p>
 
 For when you need to group rows based on repeated column values, and/or custom functions.
-In the following example groups all movies based on their production `company`
+In the following example groups all movies based on their production `company`:
```
samuelsycamore, line 10:
```suggestion
In the following example, movies are grouped based on their production `company`:
```

docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -83,13 +85,13 @@ Use the `hideDescendantCount` property of the `groupingColDef` to hide the numbe
 ### Hide the grouped columns
 
 By default, the columns used to group the rows remains visible.
-For instance if you group by `"director"`, you will have two columns titled _Director_:
+For instance if you group by `"director"`, you have two columns titled _Director_:
```
samuelsycamore, line 88: UI element labels should be in bold rather than italic.

```suggestion
For instance if you group by `"director"`, you have two columns titled **Director**:
```

docs/data/data-grid/row-grouping/row-grouping.md
```diff
@@ -83,13 +85,13 @@ Use the `hideDescendantCount` property of the `groupingColDef` to hide the numbe
...
 
 - The grouped column (the column from which you grouped the rows)
 - The grouping column on which you can toggle the groups
 
 To automatically hide the grouped columns, use the `useKeepGroupedColumnsHidden` utility hook.
-The hook will automatically hide the columns when added to the model and show them back when removed from it.
+The hook automatically hides the columns when added to the model and show them back when removed from it.
```
samuelsycamore, line 94:
```suggestion
The hook automatically hides the columns when added to the model, and displays them when removed from it.
```

docs/translations/api-docs/data-grid/data-grid-premium.json
```diff
@@ -124,7 +124,7 @@
     "processRowUpdate": "Callback called before updating a row with new values in the row and cell editing.<br><br><strong>Signature:</strong><br><code>function(newRow: R, oldRow: R) =&gt; Promise&lt;R&gt; | R</code><br><em>newRow:</em> Row object with the new values.<br><em>oldRow:</em> Row object with the old values.<br> <em>returns</em> (Promise<R> | R): The final values to update the row.",
     "rowBuffer": "Number of extra rows to be rendered before/after the visible slice.",
     "rowCount": "Set the total number of rows, if it is different from the length of the value <code>rows</code> prop. If some rows have children (for instance in the tree data), this number represents the amount of top level rows.",
-    "rowGroupingColumnMode": "If <code>single</code>, all column we are grouping by will be represented in the same grouping the same column. If <code>multiple</code>, each column we are grouping by will be represented in its own column.",
+    "rowGroupingColumnMode": "If <code>single</code>, all columns that are grouped by are represented in the same column. If <code>multiple</code>, each column that is grouped by is represented in its own column.",
```
samuelsycamore, line 127: I don't understand this. What does it mean for the columns to be grouped "by"? Grouped by what, exactly? Is there another way to say this?

```suggestion
    "rowGroupingColumnMode": "If <code>single</code>, all columns that are grouped by are represented in the same column. If <code>multiple</code>, each column that is grouped by is represented in its own column.",
```

packages/grid/x-data-grid-premium/src/DataGridPremium/DataGridPremium.tsx
```diff
@@ -278,7 +278,7 @@ DataGridPremiumRaw.propTypes = {
   error: PropTypes.any,
   /**
    * Features under development.
-   * For each feature, if the flag is not explicitly set to `true`, the feature will be fully disabled and any property / method call will not have any effect.
+   * For each feature, if the flag is not explicitly set to `true`, the feature is fully disabled and any property / method call don't have any effect.
```
samuelsycamore, line 281:
```suggestion
   * For each feature, if the flag is not explicitly set to <code>true</code>, then the feature is fully disabled, and neither property nor method calls will have any effect.",
```

packages/grid/x-data-grid-premium/src/DataGridPremium/DataGridPremium.tsx
```diff
@@ -823,8 +823,8 @@ DataGridPremiumRaw.propTypes = {
    */
   rowCount: PropTypes.number,
   /**
-   * If `single`, all column we are grouping by will be represented in the same grouping the same column.
-   * If `multiple`, each column we are grouping by will be represented in its own column.
+   * If `single`, all columns that are grouped by are represented in the same column.
+   * If `multiple`, each column that is grouped by is represented in its own column.
```
samuelsycamore, line 827: I don't understand this. What does it mean for the columns to be grouped "by"? Grouped by what, exactly? Is there another way to say this? Should this be "grouped _together_" instead?

samuelsycamore (commented):
I'm not sure which of these is the source of truth and which are autogenerated, so apologies if I left suggestions in the wrong places!

samuelsycamore:
> In other words: Active > Passive > We. cc @samuelsycamore for context.

I think this is a pretty good way to think about it. The key thing to keep in mind is that _the most important part_ of the sentence should be in the front. That's usually the action that you're describing (active voice), but sometimes it's more important to emphasize the _object_ that's being acted on (passive voice). Either way, it's always preferable to avoid injecting "we" into the docs. "You" the reader should always be the center of the action.

oliviertassinari:
@samuelsycamore Thanks

> I'm not sure which of these is the source of truth and which are autogenerated, so apologies if I left suggestions in the wrong places!

It flows like this: type (model.ts) -> prop-types (component.tsx) -> docs (json). So `packages/grid/x-data-grid-premium/src/models/dataGridPremiumProps.ts` is the source of truth.

PR #6789, opened 2022-11-09
Convert scripts to ESM

LukasTy:
Preparing for incoming monorepo update https://github.com/mui/material-ui/pull/35036

### To do before merge:
- [ ] reset monorepo dependency


scripts/build.mjs
```diff
@@ -0,0 +1 @@
+import '@mui/monorepo/scripts/build.mjs';
```
LukasTy, line 1: Removed the custom code in our codebase based on the [`TODO` comment](https://github.com/mui/mui-x/pull/6789/files#diff-a992dd64a8c626fedc95d613f43602b91f6dc6339893afaf405dc05fa13c9ceaL1)
Checked the scripts and couldn't find any difference between codebases. 🤷

package.json
```diff
@@ -85,7 +85,7 @@
     "@mnajdova/enzyme-adapter-react-18": "^0.2.0",
     "@mui/icons-material": "^5.10.9",
     "@mui/material": "^5.10.12",
-    "@mui/monorepo": "https://github.com/mui/material-ui.git#master",
+    "@mui/monorepo": "https://github.com/michaldudak/material-ui.git#esm-scripts",
```
flaviendelangle, line 88: We should replace that one before merging I guess

scripts/build.mjs
flaviendelangle: So now it is our `babel` file that is loaded ? What made that change possible ?

package.json
LukasTy: Of course, it's mentioned in the description. Just waiting for the original PR to be merged in material-ui repo. 😉

scripts/build.mjs
LukasTy: 🤦 Great spot. Somehow I spotted deduplicate issue, but not this one...
It's the same problem—`getWorkspaceRoot` will return `monorepo` root and we'd need `mui-x` root.
Will revert to custom implementation and this is just another note to @michaldudak 😉

LukasTy: But none the less, does not seem that we have too much needed custom code for build to happen as even not using our repo `babel.confg` it does complete 🙈

flaviendelangle: Our babel config adds the release dates into the bundle, the core one does not
That's the main diff I have in mind

PR #6838, opened 2022-11-14
[DataGrid] Keyboard arrow should scroll the body

### LINKED ISSUE #4282: [data grid] Keyboard arrow doesn't scroll
### Duplicates

### Latest version

### Current behavior 😯

When using the <kbd>Arrow</kbd> keys, the new active cell is not visible

actual

https://master--material-ui-x.netlify.app/components/data-grid/#commercial-version

### Expected behavior 🤔

When using the <kbd>Arrow</kbd> keys, the new active cell should be visible (scroll if necessary).

expected

http://0.0.0.0:3001/components/data-grid/#commercial-version

with

```diff
diff --git a/packages/grid/x-data-grid/src/components/cell/GridCell.tsx b/packages/grid/x-data-grid/src/components/cell/GridCell.tsx
index 710c9577..bc983146 100644
--- a/packages/grid/x-data-grid/src/components/cell/GridCell.tsx
+++ b/packages/grid/x-data-grid/src/components/cell/GridCell.tsx
@@ -172,7 +172,7 @@ function GridCell(props: GridCellProps) {
       const elementToFocus = focusableElement || cellRef.current;

       if (doesSupportPreventScroll()) {
-        elementToFocus.focus({ preventScroll: true });
+        elementToFocus.focus();
       } else {
         const scrollPosition = apiRef.current.getScrollPosition();
         elementToFocus.focus();
```

We can also see it behave correctly on the first demo of https://www.ag-grid.com/.

### Steps to reproduce 🕹

Steps:

1. Open https://master--material-ui-x.netlify.app/components/data-grid/#commercial-version
2. Follow the previous videos

### Context 🔦

It's a regression introduced in #2999 with the objective to fix #2508. I have found this by looking into #4280 in more detail. cc @m4theushw

- First, in #2999 I have never seen a valid use case for using `preventScroll` so I'm always dubious when I see it. 
- Secondly, in #2508 I would read the root problem as `.focus` is called when it shouldn't be called. Why is it called too often? Maybe because we use `React.useLayoutEffect` or because its dependencies change when they shouldn't.

I have tested it again https://github.com/mui/mui-x/issues/2508#issuecomment-910005959. It seems that the root problem in #2508 was solved closer to the root later on (I suspect in #3718). Therefore we could revert #2999 to fix this GitHub issue with no downsides.

### Your environment 🌎

MUI X v5.7.0.

### Order ID 💳 (optional)

_No response_

yaredtsy:
Fixes #4282

**#Problem**

when using the `arrow` or clicking on the cell that is out of the viewport from the browser. it will not scroll the browser body.

160253962-d9117972-7804-4b86-8da1-2ed37d27b445

**#Soultion**

check if the focused cell is out of the viewport. and if it is out of the viewport call `scrollIntoView` and scroll to the nearest position that will make it visible in the viewport.


oliviertassinari (commented):
Could you rebase on `next`. We are focusing on this branch, thanks

yaredtsy:
> This solution bypasses the focus state. The focus state could point to cell A, but cell B is clicked and it will scroll to cell B.

fixed, I moved it to the `useEffect` which handles the focus.

PR #6840, opened 2022-11-14
[core] Small changes on new pickers internals

flaviendelangle:


packages/x-date-pickers/src/themeAugmentation/overrides.d.ts
```diff
@@ -73,6 +74,7 @@ export interface PickersComponentNameToClassKey {
   MuiPickersToolbarButton: PickersToolbarButtonClassKey;
   MuiPickersToolbarText: PickersToolbarTextClassKey;
   MuiPickersYear: PickersYearClassKey;
+  MuiPickerViewLayout: PickerViewLayoutClassKey;
```
flaviendelangle, line 77: Maybe I should rename it `PickersViewLayout` to be consistent with the other components

LukasTy: I'd agree with the rename—would match all the over components starting with `MuiPickers` 👍

LukasTy (approved):

packages/x-date-pickers/src/themeAugmentation/overrides.d.ts
flaviendelangle: I'll do it on this PR to avoid exposing something and then breaking it :+1:

PR #6851, opened 2022-11-15
[pickers] Rename `ClockPicker` into `TimeClock`

flaviendelangle:
Closes #6709

#### Breaking changes

The `ClockPicker` view component has been renamed  `TimeClock` to better fit its usage:

```diff
-<ClockPicker {...props} />
+<TimeClock {...props} />
```

Component name in the theme have changed as well:

```diff
-MuiClockPicker: {
+MuiTimeClock: {
```

PR #6389, opened 2022-10-05
[pickers] Replace the `Picker` prefix in the view component by `Calendar` (eg: `MonthPicker` => `MonthCalendar`)

flaviendelangle:
Part of #4484

I'll add the detailed breaking changes later.
But here is the list of the renaming:

- `CalendarPicker` => `DateCalendar` (we can discuss naming it simply  `Calendar` if you want, in which case the range version would be `RangeCalendar` and not `DateRangeCalendar` which would match [Spectrum's naming](https://react-spectrum.adobe.com/react-spectrum/RangeCalendar.html))
- `CalendarPickerSkeleton` => `DayCalendarSkeleton`
- `MonthPicker` => `MonthCalendar`
- `YearPicker` => `YearCalendar`

## Follow up:

- Rename `ClockPicker` => `Clock` 

#### Breaking changes

- The view components allowing to pick a date or parts of a date without an input have been renamed to better fit their usage:

  ```diff
  -<CalendarPicker {...props} />
  +<DateCalendar  {...props} />
  ```

  ```diff
  -<DayPicker {...props} />
  +<DayCalendar  {...props} />
  ```

  ```diff
  -<CalendarPickerSkeleton {...props} />
  +<DayCalendarSkeleton  {...props} />
  ```

  ```diff
  -<MonthPicker {...props} />
  +<MonthCalendar  {...props} />
  ```

  ```diff
  -<YearPicker {...props} />
  +<YearCalendar  {...props} />
  ```

PR #6855, opened 2022-11-15
[pickers] Rename `CalendarPickerView`, `ClockPickerView` and `CalendarOrClockPickerView`

flaviendelangle:
Follow up on #6389 and #6851

I am not documenting the breaking change on `inputModeToggleButtonAriaLabel` because this translation method will go away with the removal of the legacy pickers


LukasTy (approved):
LGTM 🙌

PR #6864, opened 2022-11-16
[internal] Upgrade monorepo

m4theushw:
Preview: https://deploy-preview-6864--material-ui-x.netlify.app/x/introduction/

Fixes the broken logo in the sidebar of https://mui.com/x/introduction/

image

It was changed in https://github.com/mui/material-ui/pull/35030

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,14 +6,4 @@
 CHANGELOG.md
 dist
 node_modules
-<<<<<<< HEAD
 build
-=======
-.nyc_output
-
-# These come from crowdin.
-# If we would commit changes crowdin would immediately try to revert.
-# If we want to format these files we'd need to do it in crowdin
-docs/**/*-pt.md
-docs/**/*-zh.md
->>>>>>> 8e98c5fea (Remove prettier scripts (#6815))


flaviendelangle:
Not the first time we have a broken image like this
@michaldudak if you have an idea on how to avoid finding the problem on prod

michaldudak:
@m4theushw, there's also doit-dark.svg that I believe you'll have to include.

@flaviendelangle honestly, I don't. Having the whole docs infrastructure inside the Core monorepo gives us a false sense of security. Frankly, I don't always remember to verify if a given change could affect other projects.

michaldudak:
We're using a pretty uncommon (I think) setup, where we have multiple monorepos with dependencies between them. I suppose going either to one monorepo or extracting the common parts into a separate repo/package may help solve some of the issues we have (cc @Janpot).

flaviendelangle:
And on our repo, sometime we have duplicated code, sometime we import from the monorepo, sometime a little bit of both
Which makes it very hard for you to know what we use

flaviendelangle:
> If we create a separate package we'll still have the same problem as we have with the monorepo. We'll need to bump the docs-commons package version and the changes will only go live in the next release.

The separator package / repo mostly solves the false sense of security.
Everything that is on this repo would need to evolve in a non-breaking way whenever possible and when there is a breaking change it would need to be checked with the other teams.

But that's a very broad topic, as this repo could also serve for all the rest of the shared infra (the proptypes generation, the doc generation, the test setup files etc...)
So not something to do lightly.

I just think the current architecture will not be viable for the years to come, as the team grows.

Janpot:
How I understand the root of the problem is that we're installing a workspace root (`@mui/monorepo`) as a npm dependency. Only the dependencies of the workspace root are traversed and installed. From he perspective of MUI X, the individual workspaces (e.g. `@mui/monorepo/docs`, `@mui/monorepo/packages/mui-material`) in the monorepo are just subfolders. Yet we treat them as if they are installed node modules themselves. None of the dependencies of these workspaces are installed when `@mui/monorepo` is installed. In MUIX and Toolpad, we work around this by also installing the same dependencies in our docs workspaces. It's error-prone to keep those in sync with the monorepo and the loose nature by which node.js resolves dependencies makes it non-deterministic which version of a package will be loaded by files in `@mui/monorepo/docs`.

The problem is compounded with the aliasing we do in babel and webpack, which makes it more confusing to find out where modules are exactly loaded from. On top of that, most package managers have buggy implementations of github dependencies.

My recommendation would always be to avoid github dependencies and aliasing at any cost.

In our specific case, with my limited knowledge of the setup, my long term recommendations would be:

1. To avoid the monorepo dependency, either bring toolpad/X/core together in a big monorepo, or alternatively create and publish two packages:

    * `@mui/docs`: a documentation template in the style of docosaurus, nextra,... but for MUI docs specifically.
    * `@mui/scripts`: containing all the build/publish scripts required by the other repositories.

1. To avoid the aliasing, I'd look into continuously building each workspace and just rely on the yarn linking to resolve local packages to each other.

ofir5300:
Hey guys, Im a developer from [DoiT International](doit.com) .
I have made the recent changes to our logos inside MUI. https://github.com/mui/material-ui/pull/35030
Im just wondering if any change is a required from our side, or its an internal issue (as I understand from discussion).
Please let me know if I should assist or maybe supply different logo SVGs.
@m4theushw @michaldudak @flaviendelangle @Janpot

DanailH (approved):

PR #6868, opened 2022-11-16
[pickers] Pass the ampm prop from the new pickers to their field

flaviendelangle:
Currently, the format in the field does not react to the `ampm` props when used inside a picker.

### DIFF
--- a/packages/x-date-pickers/src/MobileNextTimePicker/tests/field.MobileNextTimePicker.test.tsx
+++ b/packages/x-date-pickers/src/MobileNextTimePicker/tests/field.MobileNextTimePicker.test.tsx
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import { screen } from '@mui/monorepo/test/utils';
+import { createPickerRenderer, expectInputValue } from 'test/utils/pickers-utils';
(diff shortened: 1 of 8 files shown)


alexfauquette (approved):
Since we had a lot of issues with props propagation. What about adding test to ensure props are correctly transmitted from next pickers to fields?

packages/x-date-pickers/src/MobileNextTimePicker/tests/field.MobileNextTimePicker.test.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { screen } from '@mui/monorepo/test/utils';
+import { createPickerRenderer, expectInputValue } from 'test/utils/pickers-utils';
+import { Unstable_MobileNextTimePicker as MobileNextTimePicker } from '@mui/x-date-pickers/MobileNextTimePicker';
+
+describe('<MobileNextTimePicker /> - Field', () => {
```
flaviendelangle, line 6: @alexfauquette I added it in a new file to start organize the new tests

We would have the legacy tests needing cleaning and refacto in `src/{PickerName}{PickerName}.test.tsx`
And the cleaned tests in `src/{PickerName}/tests/{testFamily}.{PickerName}.test.tsx` just like on the fields.

PR #6876, opened 2022-11-16
[core] Introduce `x-codemod` package

LukasTy:
Add the initial "infra" for `@mui/x-codemod` package.

- [x] Copy relevant `mui-codemod` code
- [x] Add one simple codemod/transform
- [x] Create `preset-safe`
- [x] Refactor code to TypeScript (involved some file moving and the final build package structure became more simple)


.codesandbox/ci.json
```diff
@@ -3,6 +3,7 @@
   "installCommand": "install:codesandbox",
   "node": "14",
   "packages": [
+    "packages/x-codemod",
```
LukasTy, line 6: Added it here initially just for testing. I don't think we need to keep releasing it on every pipeline. Would you agree?

packages/x-codemod/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "homepage": "https://github.com/mui/mui-x/tree/next/packages/x-codemod",
+  "funding": {
+    "type": "opencollective",
+    "url": "https://opencollective.com/mui"
+  },
+  "dependencies": {
+    "@babel/core": "^7.20.2",
+    "@babel/runtime": "^7.20.1",
+    "@babel/traverse": "^7.20.1",
+    "jscodeshift": "0.13.1",
```
LukasTy, line 37: Locking version to the same as in `material-ui` as 0.14.0 produces unstable transformations...
I.e.:
```tsx
return (
  <LocalizationProvider>
    stuff
  </LocalizationProvider>
);
```
becomes
```tsx
return (
  (<LocalizationProvider>
    stuff
  </LocalizationProvider>)
);
```

packages/x-codemod/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "copy-files": "cpy README.md build && cpy package.json build",
+    "build": "node ../../scripts/build.mjs node --ignore 'src/**/*.test' 'src/types.ts' && yarn copy-files"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mui/mui-x.git",
+    "directory": "packages/x-codemod"
+  },
+  "license": "MIT",
+  "homepage": "https://github.com/mui/mui-x/tree/next/packages/x-codemod",
```
LukasTy, line 28: Ideally this should be updated once we do a stable release. 🙈

packages/x-codemod/src/codemod.ts
```diff
@@ -0,0 +1,149 @@
...
+    { dry, jscodeshift, parser, print },
+    other as string[] || [],
+  );
+}
+
+yargs
+  .command({
+    command: '$0 <codemod> <paths...>',
+    describe: 'Applies a `@mui/x-codemod` to the specified paths',
+    // @ts-ignore-next-line
```
LukasTy, line 104: If any1 want's to tackle this bizarre TS error—be my guest 🙈

siriwatknp:
2 things that could be improved from the core repo:
- DX for writing tests. My pain point from the core repo is that the codemod logic files are separated from the test data files. It is quite frustrating when the codemods grow at some point. Ideally, the logic and test data files should live in the same folder:
   ```js
   codemod-a
     - codemod-a.js
     - codemod-a.test.js
     - ...actual and expected files
   codemod-b
     - ...similar structure
   ```
- I remember that the community would like to pass some options to jscodeshift but it is not possible because `@mui/codemod` does not allow them to do that.

flaviendelangle:
Fully agree with Jun 1st point !

LukasTy:
@siriwatknp Thank you for the suggestion on a better file structure.
I've made changes to follow the following pattern. What do you think?
```
codemod-a
  - index.ts
  - codemod-a.test.ts
  - actual.test.js
  - expect.test.js
codemod-b
  - ...similar structure
```

siriwatknp:
> @siriwatknp Thank you for the suggestion on a better file structure. I've made changes to follow the following pattern. What do you think?
> 
> ```
> codemod-a
>   - index.ts
>   - codemod-a.test.ts
>   - actual.spec.js
>   - expect.spec.js
> codemod-b
>   - ...similar structure
> ```

🤩 Definitely better than the core repo. One note about the spec files because it is possible to have multiple test cases in a single codemod, it could look like this:

```
codemod-a
  - index.ts
  - codemod-a.test.ts
  - {meaningful name1}-actual.spec.js
  - {meaningful name1}-expect.spec.js
  - {meaningful name2}-actual.spec.js
  - {meaningful name2}-expect.spec.js
```

LukasTy:
@siriwatknp 
Question regarding:
> I remember that the community would like to pass some options to jscodeshift but it is not possible because @mui/codemod does not allow them to do that.

Are you sure that it does not allow? There is even a mention of it [in readme](https://github.com/mui/material-ui/tree/master/packages/mui-codemod#jscodeshift-options). 🤔 
But the problem I've noticed after testing this behaviour is that it seems that only 1 option is supported at a time. 🤔 
Passing more than 1 seems to ignore all the options.
I'm still investigating this weird behaviour.... 🤷

LukasTy:
@siriwatknp 
I've done quite a bit of testing and I think I've narrowed down the issue.
I've updated [the code and examples/readme](https://github.com/mui/mui-x/pull/6876/commits/768e72857f589404e6f677396eac9360769aca3c) to make it more clear for users.
Let me know if you still see room for improvement. 😉

packages/x-codemod/README.md
```diff
@@ -0,0 +1,94 @@
...
+  --version     Show version number                                 [boolean]
+  --help        Show help                                           [boolean]
+  --parser      which parser for jscodeshift to use.
+                                                    [string] [default: 'tsx']
+  --jscodeshift Pass options directly to jscodeshift                  [array]
+
+Examples:
+  npx @mui/x-codemod v6.0.0/preset-safe src
+  npx @mui/x-codemod v6.0.0/component-rename-prop src --
+  --component=Grid --from=prop --to=newProp
```
flaviendelangle, line 34: Maybe use `localization-provider-rename-locale` here instead, unless you want yo show how to pass params

flaviendelangle (approved):
Great work !

packages/x-codemod/README.md
LukasTy: This is the same example mui-codemod gives. It's a utility transform we expose to allow users making quick custom prop switching.
IMHO, it doesn't make much sense adding example of a generic transform, which will eventually end up in the migration doc anyways. 🤔

flaviendelangle: Oh ok, then maybe change the name of the component to  `DatePicker`
I just saw `Grid` which was a bit stranged

LukasTy: 🤦 Damn, good spot, the fact that it was at least missing `Data` prefix slip through my eyes. 😆

LukasTy: Updated it to `DataGrid` just because we now that it's a way more widely used component with way more props. 🙈

PR #6881, opened 2022-11-16
[license] Polish error messages

oliviertassinari:
- We mention "scope" in the docs but I believe that we only used this terminology in the code to be more flexible in the future. As of now, I believe that it would be clearer to reference a "plan".
- I have used "license key" over "license" to reinforce "Please note that the license key does not cover every scenario, so possession of a valid license key does not by itself confer the right to use the Software." [source](https://mui.com/legal/mui-x-eula/#license-key).

It's a side fix while I was looking at how to fix the root problem of https://mail.google.com/mail/u/0/#inbox/FMfcgzGqRZhTNhvHTVzRCHKXcnSDlnfm. I think that #1364 could be it, point 5. and 7.

### DIFF
--- a/docs/data/introduction/licensing/licensing.md
+++ b/docs/data/introduction/licensing/licensing.md
@@ -216,7 +216,7 @@ however you'll be required to renew your license if you need to continue develop
 #### License key plan mismatch
 
 This error indicates that the component you are trying to use is not included in the plan of your license key.
-For example, this happen if you try to use `DataGridPremium` with a license key for the Pro plan.
+This happens if you try to use `DataGridPremium` with a license key for the Pro plan.
 
 #### Invalid license key
 
--- a/docs/data/introduction/licensing/licensing.md
+++ b/docs/data/introduction/licensing/licensing.md
@@ -213,14 +213,15 @@ For example, if you purchase a one-year license today, you will be able to updat
 Those versions will always be available for use in a deployed application,
 however you'll be required to renew your license if you need to continue development with a version released after twelve months.
 
-#### Out of scope license key
+#### License key plan mismatch
 
-This error indicates that the component you are trying to use is not included in your license.
-This can happen if you try to use `DataGridPremium` with a Pro license.
+This error indicates that the component you are trying to use is not included in the plan of your license key.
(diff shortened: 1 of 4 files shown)


docs/data/introduction/licensing/licensing.md
```diff
@@ -213,14 +213,15 @@ For example, if you purchase a one-year license today, you will be able to updat
 Those versions will always be available for use in a deployed application,
 however you'll be required to renew your license if you need to continue development with a version released after twelve months.
 
-#### Out of scope license key
+#### License key plan mismatch
 
-This error indicates that the component you are trying to use is not included in your license.
-This can happen if you try to use `DataGridPremium` with a Pro license.
+This error indicates that the component you are trying to use is not included in the plan of your license key.
+For example, this happen if you try to use `DataGridPremium` with a license key for the Pro plan.
```
flaviendelangle, line 219: The usage of DataGridPremium with a pro license is the only scenario I see today
Not sure the "For example" make sense.

oliviertassinari: Ok, sold

flaviendelangle (approved):

joserodolfofreitas (commented):
I had a draft with some suggestions for this review.

packages/x-license-pro/src/utils/licenseErrorMessageUtils.ts
```diff
@@ -22,12 +22,13 @@ export function showInvalidLicenseError() {
 
 export function showOutOfScopeLicenseError() {
   showError([
-    'MUI: Out of scope license key.',
+    'MUI: License key plan mismatch.',
     '',
-    'Your MUI X license key isn\'t valid. You are rendering a DataGridPremium component that requires a license key with the "premium" feature scope but your license key has the "pro" feature scope.',
+    'Your use of MUI X is not compatible with the plan of your license key.',
```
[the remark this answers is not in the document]
oliviertassinari: My goal was to match https://www.notion.so/mui-org/Technical-writing-7e55b517ac2e489a9ddb6d0f6dd765de#85219822c3194b6f8a49ee08ea82b90a. Start with What, then Why, and finally give a solution.

PR #6884, opened 2022-11-16
correct filterOperatorBefore

HassanGhazy:
Signed-off-by: Hassan Ghazy Sammour <46076264+HassanGhazy@users.noreply.github.com>


HassanGhazy (commented):
The previous word "بعد" is wrong, I added the right one which is "قبل"

PR #6895, opened 2022-11-17
[pickers] Do not loose the translations when using nested `LocalizationProvider` with each a `localeText` prop

flaviendelangle:


packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
```diff
@@ -56,10 +56,15 @@ export function LocalizationProvider<TDate>(inProps: LocalizationProviderProps<T
     MuiPickersAdapterContext,
   ) ?? { utils: undefined, localeText: undefined };
 
+  const tempLocaleText = React.useMemo(
```
flaviendelangle, line 59: Otherwise when `inProps.localeText` is not undefined, we loose the parent one.

alexfauquette (approved):

packages/x-date-pickers/src/LocalizationProvider/LocalizationProvider.tsx
```diff
@@ -52,31 +52,25 @@ export interface LocalizationProviderProps<TDate> {
...
-    dateAdapter: DateAdapter,
-    dateFormats,
-    dateLibInstance,
-    adapterLocale,
-    localeText,
-  } = props;
+  const localeText = React.useMemo(
+    () => ({ ...props.localeText, ...parentLocaleText, ...inLocaleText }),
+    [props.localeText, parentLocaleText, inLocaleText],
+  );
```
alexfauquette, line 71: `props.localeText` is not obvious to get

What about moving this after `props` destructuring  to be able to write
```
() => ({ ...defaultizedLocaleText, ...parentLocaleText, ...inLocaleText }),
```

flaviendelangle: OK for me, I'll just rename `defaultizedLocaleText` into `themeLocaleText` to make clear that it can only contain the theme value and never the value from the `inProps`

PR #6900, opened 2022-11-17
[pickers] Use grid for modifying the layout

alexfauquette:
Documentation of the new layout is available at: https://deploy-preview-6900--material-ui-x.netlify.app/x/react-date-pickers/custom-layout/

Text is not polished but allows to have an overview

About adding elements in the layout, we could export sub-components, to allows a full DOM modification. But for most use-case, the original order make sense, and we add only add helping/description elements on top/bottom


docs/data/date-pickers/custom-layout/LayoutBlocks.js
```diff
@@ -0,0 +1,112 @@
...
+          fullWidth
+          color="primary"
+          value={currentComponent}
+          onChange={(event, value) => setCurrentComponent(value)}
+          exclusive
+          sx={{ mb: 2 }}
+        >
+          <ToggleButton value={'date'}>date picker</ToggleButton>
+          <ToggleButton value={'time'}>time picker</ToggleButton>
+          <ToggleButton value={'date-time'}>date range picker</ToggleButton>
```
flaviendelangle, line 39: Is it date time or date range ?

flaviendelangle (commented):
Sorry if we already answered this question, but don't you think that it makes the keyboard navigation weird ?

We update the visual order, but not the rendering order.

https://user-images.githubusercontent.com/3309670/202702745-5b327398-ceb3-4f92-8b44-20c1d7a05e23.mp4

alexfauquette:
@flaviendelangle Not already answered :)

For this specific demo, the idea would be to replace the action bar by the shortcuts latter. But it brings a larger topic: element order

For toolbar, views, and actions sounds obvious to have them in the given order.
For the shortcuts, it's a bit more complicated, because they could either be placed at the left, right or bottom of the view


# Potential solutions

## Solution 1

Add props to the `LayoutRoot` such as `tabsPosition='bottom' | 'top'` and `shortcutPosition='before' | 'after'` to modify the render order of the elements (not a big fan)

Not a big fan because it might ends into an large list of props

## Solution 2

Add prop `elementsOrder` to the `LayoutRoot` such as `elementsOrder={['toolbar', 'views', 'shortcuts', 'actionbar']}`

Why not but it dos not cover the case of dev willing to add elements in the middle

## Solution 3

Give more responsibility to `LayoutRoot`. If the default `LayoutRoot` is defined as follow

```jsx
const LayoutRoot = () => 
	<div>
		<Toolbar />
		<ContentWrapper>
			<Tabs />
			{children}
		</ ContentWrapper>
		<ActionBar />
	</ div>
```

Such that when Overriding the `LayoutRoot` developers would have to define all the elements around the view.

Maybe the better solution but sounds too heavy for a dev who just want to move the shortcut from the left to the bottom. Or we should provide a really nice example reusing the `components` prop such that they can create their own Layout and reuse it every where

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+
+Here is a demonstration with the 3 components highligted with color full borders
+
+{{"demo": "LayoutBlocks.js"}}
+
+## Customize layout
+
+### Understanding the structure
+
+The layout is structured as follow:
```
LukasTy, line 33:
```suggestion
The layout is structured as follows:
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+
+A `<LayoutRoot />` component wrapp all the sub components to provide the structure.
```
LukasTy, line 35:
```suggestion
A `<LayoutRoot />` component wraps all the sub-components to provide the structure.
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+By default it renders a `div` with `display: grid`.
+Such that all sub-components take place into a 3 by 3 [CSS grid](https://developer.mozilla.org/fr/docs/Web/CSS/CSS_Grid_Layout).
```
LukasTy, line 37:
```suggestion
Such that all sub-components take place inside a 3 by 3 [CSS grid](https://developer.mozilla.org/fr/docs/Web/CSS/CSS_Grid_Layout).
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+    <View />
+  </LayoutContent>
+  // Action bar
+  <LayoutActionBar>
+    <ActionBar />
+  </LayoutActionBar>
+</LayoutRoot>
+```
+
+### Moving components
```
LukasTy, line 60:
```suggestion
### Moving elements
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+
+To move an element, the easiest way is to override it's wrapper position with properties [`gridColumn`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column) and [`gridRow`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row).
```
LukasTy, line 62:
```suggestion
To move an element—the easiest way is to override it's wrapper position with [`gridColumn`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column) and [`gridRow`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row) properties.
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+
+In the next example, the action bar is replaced by a list, and placed at the right of the component.
```
LukasTy, line 64:
```suggestion
In the next example—the action bar is replaced by a list and placed at the right of the component.
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,73 @@
...
+To place it at the right, we use style `{ gridColumn: '3', gridRow: '2' }`.
```
LukasTy, line 65:
```suggestion
Placing at the right is achieved by applying `{ gridColumn: '3', gridRow: '2' }` style.
```

LukasTy (commented):
Overall I do like the grid approach to layout the pickers elements. 👍 
The question of shortcuts integration would still remain and I'd say we need to think about it. Meaning, how well/easy would that option integrate into this approach? 🤔

LukasTy (changes requested):
After the recent changes there are quite a few issues.
- the demos are broken (props are not being propagated to the correct components)
- quite a few issues in dev tools console because of that
- eslint issues

docs/data/date-pickers/custom-components/custom-components.md
```diff
@@ -5,7 +5,7 @@ title: Date and Time pickers - Custom components
 
 # Custom components
 
-<p class="description">The date picker lets users select a date from a menu.</p>
+<p class="description">The date picker lets you customize sub-components.</p>
```
flaviendelangle, line 8: :see_no_evil:

alexfauquette:
> the demos are broken (props are not being propagated to the correct components)

Yes, I was naively propagating props to children which included `children` 🫢 and props here to be used but not propagated 🙈

docs/data/date-pickers/custom-layout/AddComponent.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <StaticNextDatePicker
+        componentsProps={{
+          layout: ({ toolbar, actionBar, content }) => ({
+            sx: {
+              // Keep the structure with display grid
+              display: 'grid',
+              gridAutoColumns: 'max-content auto max-content',
+              gridAutoRows: 'max-content auto max-content',
+
+              '& .MuiPickersViewLayout-actionbar': {
```
alexfauquette, line 64:
```suggestion
              // Modify CSS positions
              '& .MuiPickersViewLayout-actionbar': {
```

docs/data/date-pickers/custom-layout/AddComponent.tsx
```diff
@@ -0,0 +1,82 @@
...
+                gridColumn: '1',
+                gridRow: '2',
+              },
+              '& .MuiPickersViewLayout-toolbar': {
+                gridColumn: '2',
+                gridRow: '1',
+              },
+            },
+            children: [<RestaurantHeader />, toolbar, actionBar, content],
```
alexfauquette, line 73:
```suggestion
	    // Modify DOM structure
            children: [<RestaurantHeader />, toolbar, actionBar, content],
```

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
```diff
@@ -15,27 +15,51 @@ export const PickersViewLayoutRoot = styled('div', {
...
+  gridRow: '3',
 });
 
 const useUtilityClasses = (ownerState: PickersViewLayoutProps<any, any>) => {
   const { classes, isLandscape } = ownerState;
   const slots = {
     root: ['root'],
     content: ['content', isLandscape && 'content--landscape'],
+    toolbar: ['toolbar', isLandscape && 'toolbar--landscape'],
+    actionbar: ['actionbar', isLandscape && 'actionbar--landscape'],
```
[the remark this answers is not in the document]
alexfauquette: For now, sub-components are responsible for their position without knowing the parent state. For example, the toolbar modifies its position with CSS selector `[`&.${pickersViewLayoutClasses['toolbar--landscape']}`]`

```jsx
export const PickersViewLayoutToolbar = styled('div', {
  name: 'MuiPickersViewLayout',
  slot: 'Toolbar',
  overridesResolver: (props, styles) => styles.toolbar,
})(({ theme }) => ({
  gridColumn: '1 / 4',
  gridRow: '1',
  [`&.${pickersViewLayoutClasses['toolbar--landscape']}`]: {
    gridColumn: theme.direction === 'rtl' ? '3' : '1',
    gridRow: '1 / 3',
  },
}));
```


Another option would be to let the `root` component do the entire positioning job.
 
 Not fully sure which is the best, but I agree instinctively I would go for the second option

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
```diff
@@ -80,14 +112,10 @@ export const PickersViewLayout = React.forwardRef(function PickersViewLayout<
...
 
-  if (view == null) {
-    return null;
-  }
-
-  return (
-    <PickersViewLayoutRoot className={clsx(className, classes.root)} ref={ref}>
-      <PickersViewLayoutContent className={classes.content}>
-        {shouldRenderToolbar && !!Toolbar && (
+  const subComponents = {
```
alexfauquette, line 115: Not sure it's a good way of doing

docs/data/date-pickers/custom-layout/LayoutBlocks.js
```diff
@@ -0,0 +1,112 @@
...
+import { Unstable_StaticNextTimePicker as StaticNextTimePicker } from '@mui/x-date-pickers/StaticNextTimePicker';
+import { Unstable_StaticNextDateTimePicker as StaticNextDateTimePicker } from '@mui/x-date-pickers/StaticNextDateTimePicker';
+
+export default function LayoutBlocks() {
+  const [currentComponent, setCurrentComponent] = React.useState('date');
+  const [orientation, setOrientation] = React.useState('portrait');
+
+  return (
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <Stack sx={{ width: '100%' }} alignItems="center">
```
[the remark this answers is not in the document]
alexfauquette: I took some inspiration from joy, but as you can see, I'm not a designer 🙈🥸

image

image

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
```diff
@@ -14,33 +15,67 @@ export const PickersViewLayoutRoot = styled('div', {
...
   overridesResolver: (props, styles) => styles.root,
-})({
-  display: 'flex',
-  flexDirection: 'column',
-});
+})(({ theme }) => ({
+  display: 'grid',
+  gridAutoColumns: 'max-content auto max-content',
+  gridAutoRows: 'max-content auto max-content',
+  [`&.${pickersViewLayoutClasses.landscape} .${pickersViewLayoutClasses.toolbar}`]: {
```
LukasTy, line 22: 👍 I do feel that, even though it makes even the default styles a bit more terse, in the end it provides a less cumbersome way for end users to override the layout. 🤔 
```tsx
componentsProps={{
  layout: {
    sx: {
      [`&.${pickersViewLayoutClasses.landscape}`]: {
        [`.${pickersViewLayoutClasses.content}`]: {
          gridColumn: 1,
        },
        [`.${pickersViewLayoutClasses.toolbar}`]: {
          gridColumn: 2,
        },
      }
    }
  }
}}
```

docs/data/date-pickers/custom-layout/AddComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+            sx: {
+              // Modify DOM structure
+              '& .MuiPickersViewLayout-actionbar': {
+                gridColumn: '1',
+                gridRow: '2',
+              },
+              '& .MuiPickersViewLayout-toolbar': {
+                gridColumn: '2',
+                gridRow: '1',
+              },
```
LukasTy, line 67: I'd suggest encouraging our users to use the class constants whenever possible in our demos, what do you think @alexfauquette @flaviendelangle?
It also uncovers an issue of not correctly re-exporting the `pickersViewLayoutClasses` 🙈 
By the way, I'm pretty sure that `& ` is not necessary.

```suggestion
              [`.${pickersViewLayoutClasses.actionbar}`]: {
                gridColumn: '1',
                gridRow: '2',
              },
              [`.${pickersViewLayoutClasses.toolbar}`]: {
                gridColumn: '2',
                gridRow: '1',
              },
```

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
```diff
@@ -14,33 +15,67 @@ export const PickersViewLayoutRoot = styled('div', {
...
 
 const useUtilityClasses = (ownerState: PickersViewLayoutProps<any, any>) => {
   const { classes, isLandscape } = ownerState;
   const slots = {
-    root: ['root'],
-    content: ['content', isLandscape && 'content--landscape'],
+    root: ['root', isLandscape && 'landscape'],
+    content: ['content'],
+    toolbar: ['toolbar'],
+    actionbar: ['actionbar'],
```
LukasTy, line 63: Nitpick: maybe it should be `actionBar` instead? 🤔

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
```diff
@@ -99,18 +127,49 @@ export const PickersViewLayout = React.forwardRef(function PickersViewLayout<
...
+  >({
+    elementType: Layout,
+    externalSlotProps: componentsProps?.layout,
+    additionalProps: {
+      ref,
+      className: clsx(className, classes.root),
+      ...subComponents,
+    },
+    externalForwardedProps: other,
+    ownerState: { ...props, ...subComponents },
```
LukasTy, line 167: I don't have the strongest opinion on this approach, but I tend to lean towards maybe having a render function instead of this yet another approach to customize rendering. 🙈 🤷 
Or am I mistaken and we do already have similar approach in other place(s)? 🤔

LukasTy (changes requested):
Looks really close to a finished solution, great job! 👍 🙏

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
alexfauquette: I replace it with 
```jsx
...(ownerState.isLandscape && {
    [`& .${pickersViewLayoutClasses.toolbar}`]: {
      gridColumn: theme.direction === 'rtl' ? '3' : '1',
      gridRow: '1 / 3',
    },
  }),
```

Such that the CSS Selector only use `&` and `.toolbar` classes, so developpers can usn `&.isLandscape .toolbar` to be more specific and override it.

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,96 @@
...
+  layout: ({toolbar, content, actionBar}) =>({
+    sx: {
+      /* CSS reordering */
+    },
+    children: [toolbar, actionBar, content ]
+  }),
+}}
+```
+
+Here is the previous demonstration with a fix for the tabulation order.
```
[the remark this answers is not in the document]
alexfauquette: Yes, I forget that I moved action list from left to right to hidde the problem ;)

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
LukasTy: Sure, that seems to also work. 🤔 
Just don't forget the "It also uncovers an issue of not correctly re-exporting the `pickersViewLayoutClasses`" point to avoid following import:
```ts
import { pickersViewLayoutClasses } from '@mui/x-date-pickers/internals/components/PickersViewLayout/pickersViewLayoutClasses';
```

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
flaviendelangle: We already use the `ownerState` of the slot to pass information (the `previousIconButton` slot for instance has a `hidden` property added on top of the props).

But I never saw React nodes passed their.
My idea was to re-use an already existing customization feature (the callback version of the slot props).
But maybe that's too hacky.
Maybe it would be interesting writing an example with both approaches to pick the one we prefer.

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
LukasTy: Are you sure moving this component to the first level export is the best course of action? 🤔 
I.e.: `PickersPopper` is somewhat on the same level of importance but is exported and placed in `internals` 🤔 
Your reasoning is probably the fact that its classes might be used quite frequently and thus exporting them from `/internals` is not a nice DX?

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.tsx
alexfauquette: I'm not sure about what would be a "render function" interface.

```jsx
const renderLayout = (props, Toolbar, Content, ActionBar, Shortcuts) => <>
    <Toolbar />
    <Content />
    <ActionBar />
    <Shortcuts />
  </>

<DatePicker
  renderLayoutContent={renderLayout}
  // Or
  componentProps={{ layout: { renderLayout } }}
/>
```

It does not seems to be less hacky


----

A last option I see could be to do not fully respect usual slot behavior. Instead of replacing the `PickersViewLayout` when `components.Layout` is provided, we replace it's content.

```jsx
const PickersViewLayout = (props) => {
	const { components, componentsProps, ...other} = props;

	// We still create the default components based on the props
	const actionBarProps = {onAccept: other.onAccept, onClear: other.onClear, ..., ...componentsProps?.actionBar}

	// ...
	
	if(components?.Layout) {
		// If the layout is overridden, we provide components and components props to writ every thing simply
		const Layout = components.Layout;
		return <Layout {...props} components={{ActionBar, Toolbar, ...}} componentsProps={{ actionBar: actionBarProps, ...}}/>
	}

	// Otherwise we do as usual
	return <LayoutRoot >
		{/* Defautl content */ }
	</LayoutRoot>

}
```


From a DX point of view I tend to prefer the initial solution with children. It's hacky but we are at the limit where a headless approach could make sense. So hacky solution for dev who don't have time and headless for clean solution sounds like a good trade off

flaviendelangle: I agree that your solution is less hacky

packages/x-date-pickers/src/internals/components/PickersToolbar.tsx
```diff
@@ -51,7 +51,7 @@ const PickersToolbarRoot = styled('div', {
   justifyContent: 'space-between',
   padding: theme.spacing(2, 3),
   ...(ownerState.isLandscape && {
-    height: 'auto',
+    height: '100%',
```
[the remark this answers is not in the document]
alexfauquette: That modification was a mistake. I thought it would fix Argos diff. But those diffs were due to a modification of the layout: Now the action bar goes below the toolbar in landscape mode. Before the toolbar was covering the entire height

image

I reverted this `height` to `auto`, and I modified the behavior such that in landscape mode, it's a `flex-start` instead of a `space-between` which sounds more natural

image

docs/data/date-pickers/custom-layout/LayoutBlocks.js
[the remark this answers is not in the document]
flaviendelangle: image

Not a design either :laughing: , and we have to keep the responsive aspect of the doc in mind
I prefer this orientation, but I would remove the shadow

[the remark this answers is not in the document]
alexfauquette: image

I mobile buttons move on top. I could reduce spacing for this case

image

alexfauquette:
@flaviendelangle @LukasTy I finished to update the docs with the new behavior

I'm thinking about the following diff
```diff
-PickersViewLayout
+PickersLayout
```

It's not a breaking change since `PickersViewLayout` is an internal component for now

LukasTy:
> It's not a breaking change since `PickersViewLayout` is an internal component for now

It seems that it would be a breaking change, because we have it in `themeAugmentation`. 😞

LukasTy:
> It a breaking change between one alpha version and another, so not a big deal right ?
> This component was created for the new pickers.

Yes, indeed. Sorry, my bad.
Then by no means, `PickersLayout` sounds a bit more correct. 👌

packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const props = useThemeProps({ props: inProps, name: 'MuiPickersLayout' });
+
+  const { toolbar, content, tabs, actionBar } = usePickerLayout(props);
+  const { sx, className, isLandscape } = props;
+
+  return (
+    <PickersLayoutRoot
+      ref={ref}
+      sx={sx}
+      className={clsx(className, pickersLayoutClasses.root)}
```
LukasTy, line 50: Have you considered the following change aligning the component implementation with others and also applying `MuiPickersLayout-landscape` class in landscape orientation:

```diff
diff --git a/packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx b/packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
index 4d2a8ac11..c4ee96b3c 100644
--- a/packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
+++ b/packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
@@ -1,10 +1,20 @@
 import * as React from 'react';
 import clsx from 'clsx';
 import { styled, useThemeProps } from '@mui/material/styles';
+import { unstable_composeClasses as composeClasses } from '@mui/utils';
 import { PickersLayoutProps } from './PickersLayout.types';
-import { pickersLayoutClasses } from './pickersLayoutClasses';
+import { getPickersLayoutUtilityClass, pickersLayoutClasses } from './pickersLayoutClasses';
 import usePickerLayout from './usePickerLayout';
 
+const useUtilityClasses = (ownerState: { isLandscape: boolean }) => {
+  const slots = {
+    root: ['root', ownerState.isLandscape && 'landscape'],
+    contentWrapper: ['contentWrapper'],
+  };
+
+  return composeClasses(slots, getPickersLayoutUtilityClass, {});
+};
+
 export const PickersLayoutRoot = styled('div', {
   name: 'MuiPickersLayout',
   slot: 'Root',
@@ -25,7 +35,7 @@ export const PickersLayoutRoot = styled('div', {
 
 export const PickersLayoutContentWrapper = styled('div', {
   name: 'MuiPickersLayout',
-  slot: 'Content',
+  slot: 'ContentWrapper',
   overridesResolver: (props, styles) => styles.contentWrapper,
 })({
   gridColumn: 2,
@@ -43,15 +53,17 @@ export const PickersLayout = React.forwardRef(function PickersLayout(
   const { toolbar, content, tabs, actionBar } = usePickerLayout(props);
   const { sx, className, isLandscape } = props;
 
+  const classes = useUtilityClasses({ isLandscape });
+
   return (
     <PickersLayoutRoot
       ref={ref}
       sx={sx}
-      className={clsx(className, pickersLayoutClasses.root)}
+      className={clsx(className, classes.root)}
       ownerState={{ isLandscape }}
     >
       {toolbar}
-      <PickersLayoutContentWrapper className={pickersLayoutClasses.contentWrapper}>
+      <PickersLayoutContentWrapper className={classes.contentWrapper}>
         {tabs}
         {content}
       </PickersLayoutContentWrapper>

```

packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
```diff
@@ -0,0 +1,61 @@
...
+        gridColumn: theme.direction === 'rtl' ? 3 : 1,
+        gridRow: '1 / 3',
+      }
+    : { gridColumn: '1 / 4', gridRow: 1 },
+  [`& .${pickersLayoutClasses.actionBar}`]: { gridColumn: '1 / 4', gridRow: 3 },
+}));
+
+export const PickersLayoutContentWrapper = styled('div', {
+  name: 'MuiPickersLayout',
+  slot: 'Content',
```
LukasTy, line 28:
```suggestion
  slot: 'ContentWrapper',
```

packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export interface ExportedPickersLayoutSlotsComponent<TValue, TView extends DateOrTimeView> {
+  /**
+   * Custom component for the action bar, it is placed bellow the picker views.
+   * @default PickersActionBar
+   */
+  ActionBar?: React.ElementType<PickersActionBarProps>;
+  /**
+   * Custom component for wrapping the layout.
+   * It wrapps the toolbar, views, and action bar
```
LukasTy, line 20:
```suggestion
   * It wraps the toolbar, views, and action bar.
```

LukasTy (commented):
The final solution is looking really great! 🎉 🚀 ❤️ 

Maybe we could increase the `maxWidth` (to 165 instead of 160) of the toolbar in landscape to at least allow our english translation to fit in one line?

It seems that we have an issue with `ampmInClock` prop. Am/pm selection does not look good in the provided screenshot, but if `ampmInClock={false}` is passed, there is no other way to change meridiem. Shouldn't it be added to the `DateTImePickerToolbar` at that point? 🤔 @flaviendelangle 
Screenshot 2022-12-21 at 09 23 00

P.S. Shall we proceed with removing `PaperContent` slot after this is merged? 🤔

docs/data/date-pickers/custom-layout/MovingActionsReorder.tsx
```diff
@@ -0,0 +1,78 @@
+export default () => null;
```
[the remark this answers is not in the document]
alexfauquette: Of course not, it's just in case you asked for a distinct demo between adding at reordering components :)

flaviendelangle:
> P.S. Shall we proceed with removing PaperContent slot after this is merged?

We should indeed remove it from all the next gen pickers

alexfauquette:
@lukas About the AM/PM miss placed, it's an old issue, 
I started to work on it #5149 but we had a strange bug, and it fall into the lands of "not urgent, will do it later"

About the next steps we can:
- Implement the shortcuts (#7154)
- Remove PaperContent in new pickers
- Refine the sub-component styling. I plan to take some time in the first week of January to have a look at all the toolbars and check if the styling is good, and if they are responsive or not

LukasTy:
> About the AM/PM miss placed, it's an old issue.

Yeah, I also thought about it. The button placement is not the biggest issue, I think that the `am/pm` disappearing if `ampmInClock` is defined is a bigger issue.
But in any case, it of course can and should be tackled with a separate PR. 🆗

docs/scripts/api/buildComponentsDocumentation.ts
```diff
@@ -224,6 +224,9 @@ const buildComponentDocumentation = async (options: {
   const { filename, project, apiPagesFolder, documentedInterfaces, projects, pagesMarkdown } =
     options;
 
+  if (path.basename(filename).startsWith('use')) {
```
flaviendelangle, line 227: Shouldn't this condition be added in `getComponentsWithApiDoc` ? To only have actual components when we reach `buildComponentDocumentation`

scripts/x-date-pickers-pro.exports.json
```diff
@@ -106,6 +106,8 @@
   { "name": "enUS", "kind": "Variable" },
   { "name": "esES", "kind": "Variable" },
   { "name": "ExportedDateRangeCalendarProps", "kind": "Interface" },
+  { "name": "ExportedPickersLayoutSlotsComponent", "kind": "Interface" },
```
flaviendelangle, line 109: We should probably not export `ExportedPickersLayoutSlotsComponent` and `ExportedPickersLayoutSlotsComponentsProps` (we can remove them from `PickersLayout/index` and adapt our internal imports).

docs/scripts/api/buildComponentsDocumentation.ts
alexfauquette: Moved to the following solution in `getComponentsWithApiDoc`
```diff
       const isExported = !!project.exports[componentName];
-      if (isExported) {
+      const isHook = path.basename(file).startsWith('use');
+      if (isExported && !isHook) {
         paths.push(path.join(project.rootPath, file));
       }

```

docs/pages/x/api/date-pickers/pickers-action-bar.js
```diff
@@ -0,0 +1,23 @@
...
+export default function Page(props) {
+  const { descriptions, pageContent } = props;
+  return <ApiPage descriptions={descriptions} pageContent={pageContent} />;
+}
+
+Page.getInitialProps = () => {
+  const req = require.context(
+    'docsx/translations/api-docs/date-pickers',
+    false,
+    /\/pickers-action-bar(-[a-z]{2})?\.json$/,
```
LukasTy, line 15: Should we maybe add this API entry in the submenu as well as maybe the "index" page, even though it's link has already been removed, so, it's probably no longer that relevant. 🤔 
Screenshot 2022-12-22 at 09 37 08

docs/pages/x/api/date-pickers/pickers-layout.js
```diff
@@ -0,0 +1,23 @@
...
+export default function Page(props) {
+  const { descriptions, pageContent } = props;
+  return <ApiPage descriptions={descriptions} pageContent={pageContent} />;
+}
+
+Page.getInitialProps = () => {
+  const req = require.context(
+    'docsx/translations/api-docs/date-pickers',
+    false,
+    /\/pickers-layout(-[a-z]{2})?\.json$/,
```
LukasTy, line 15: Same for this one? Also, it does not seem to be linked anywhere (has no demos), maybe it's worth adding it in `custom-layout`?

packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export interface ExportedPickersLayoutSlotsComponent<TValue, TView extends DateOrTimeView> {
+  /**
+   * Custom component for the action bar, it is placed bellow the picker views.
+   * @default PickersActionBar
+   */
+  ActionBar?: React.ElementType<PickersActionBarProps>;
+  /**
+   * Custom component for wrapping the layout.
+   * It wraps the toolbar, views, and action bar
```
LukasTy, line 20:
```suggestion
   * It wraps the toolbar, views, and action bar.
```

LukasTy (approved):
LGTM! 🚀 💯 
Just a few final nitpicks... 🙈

LukasTy:
Maybe we should add `'docs-date-pickers-date-calendar/DateCalendarServerRequest.png'` to `test/regressions/index.js` blacklist?
Or change the demo to render same data for same date. 🤔

flaviendelangle:
@LukasTy yes, that's my bad
I did it in https://github.com/mui/mui-x/pull/7280/files#diff-26370639d5cddf1b1016c2de2ceb142f773f358cf916cbc637bdd327be854495R19, but you can extract it if you want.

docs/pages/x/api/date-pickers/pickers-action-bar.js
alexfauquette: I first did not add them because the tables were nearly empty, but it was a bug from my side

oliviertassinari:
Maybe we could override `Math.random()` to always return 0.5? We did this for chance https://github.com/mui/mui-x/blob/3e7bfb2a3b11d3728892b245e70928c7669a0bd6/packages/grid/x-data-grid-generator/src/services/random-generator.ts#L20-L24 but we probably don't want to have this leak into a developer's demo.

packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
```diff
@@ -0,0 +1,134 @@
...
+// TODO: Remove this functions. It get introduced to mark `value` prop in PickersLayoutProps as not required.
+// The true type should be
+// - For pickers value: TDate | null
+// - For rangepickers value: [TDate | null, TDate | null]
+function toolbarHasValue<TValue, TView extends DateOrTimeView>(
+  toolbarProps: BaseToolbarProps<TValue | undefined, TView> | any,
+): toolbarProps is BaseToolbarProps<TValue, TView> {
+  return toolbarProps.value !== undefined;
+}
+
```
alexfauquette, line 25: Since the `value: TValue` of layout props is interpreted as `Proptypes.any.require`, it returns an error when we pass `null`, which is a valid value for pickers (not for range pickers)

Since I did not find any solution to this problem, I did the following trick: override it with

```
value?: TValue
```

To be compliant with props types expected by the toolbar, I added this `toolbarHasValue` to ensure value is not undefined and set the correct type

@LukasTy @flaviendelangle  Do you agree with this nasty TS trick?

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+## CSS customization
+
+To move an element—the easiest way is to override it's wrapper position with [`gridColumn`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column) and [`gridRow`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row) properties.
+
+In the next example, the action bar is replaced by a list, and placed at the left of the component.
+Placing at the right is achieved by applying `{ gridColumn: 1, gridRow: 2 }` style to it.
+
+{{"demo": "MovingActions.js"}}
```
joserodolfofreitas, line 63: Changing the `gridColumn` and `gridRow` of the action bar on these demos has no effect.

We need to add the respective class to the custom ActionList to make the selector works. 
This is worth mentioning in the docs as well.

`<List className={pickersLayoutClasses.actionBar}>`

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,95 @@
...
+  // Put the action bar before the content
+  return <div>
+    {toolbar}
+    {actionBar}
+    {content}
+  </div>
+}
+```
+
+This slot can also be used to add additional information in the layout.
```
joserodolfofreitas, line 90: I feel this is quite important information.
To consider:

```suggestion
:::info
This slot can also be used to add additional information in the layout.
:::
```

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,95 @@
...
+
+Here is the previous demonstration with a fix for the tabulation order and logo added into the layout.
+Notice the use of `pickersLayoutClasses`, `PickersLayoutRoot`, and `PickersLayoutContentWrapper` to avoid having to rewrite the default CSS.
```
joserodolfofreitas, line 93:
```suggestion
Notice the use of `pickersLayoutClasses`, `PickersLayoutRoot`, and `PickersLayoutContentWrapper` to avoid rewriting the default CSS.
```

joserodolfofreitas (commented):
Amazing work, @alexfauquette!
The new hook is elegant, and the abilities to customize the layout are simply great!
The exploration in collaboration with the team was also outstanding.

I left a few suggestions to trim the docs a bit and spotted an issue with the demos using the custom `ActionList`.

docs/data/date-pickers/custom-layout/custom-layout.md
joserodolfofreitas: There's an open question on using the class to get the gridLayout working. In this case, it had no effect, but it could potentially break something. 
Perhaps we should also handle this case on the docs.

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,95 @@
...
+
+Here is a demonstration with the 3 main blocks outlined with color borders.
+
+{{"demo": "LayoutBlocks.js", "hideToolbar": true, "bg": "inline"}}
+
+## Layout structure
+
+The layout is structured as follows:
+
+A `<PickersLayoutRoot />` wraps all the sub-components to provide the structure.
```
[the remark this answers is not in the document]
LukasTy: We are using **sub-components** in the side menu as well 🤔 
Screenshot 2022-12-23 at 08 59 27

packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
LukasTy: I've got a question from different PoV: do we need `value: TValue` in toolbar instead of `value?: TValue`?
Pickers are fine working without any initial value defined. Do we need to force our selves so that internal components have to have this field defined marked as required if it causes us issues?
```diff
diff --git a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
index e12140b14..5fe81985e 100644
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
@@ -70,7 +70,7 @@ export const DateRangePickerToolbar = React.forwardRef(function DateRangePickerT
   const props = useThemeProps({ props: inProps, name: 'MuiDateRangePickerToolbar' });
 
   const {
-    value: [start, end],
+    value,
     isMobileKeyboardViewOpen,
     toggleMobileKeyboardView,
     rangePosition,
@@ -78,6 +78,7 @@ export const DateRangePickerToolbar = React.forwardRef(function DateRangePickerT
     toolbarFormat,
     className,
   } = props;
+  const [start, end] = value ?? [null, null];
 
   const localeText = useLocaleText<TDate>();
 
diff --git a/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx b/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
index bd0a75b6a..5008ddd57 100644
--- a/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
+++ b/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
@@ -13,16 +13,6 @@ function toolbarHasView<TValue, TView extends DateOrTimeView>(
   return toolbarProps.view !== null;
 }
 
-// TODO: Remove this functions. It get introduced to mark `value` prop in PickersLayoutProps as not required.
-// The true type should be
-// - For pickers value: TDate | null
-// - For rangepickers value: [TDate | null, TDate | null]
-function toolbarHasValue<TValue, TView extends DateOrTimeView>(
-  toolbarProps: BaseToolbarProps<TValue | undefined, TView> | any,
-): toolbarProps is BaseToolbarProps<TValue, TView> {
-  return toolbarProps.value !== undefined;
-}
-
 const useUtilityClasses = (ownerState: PickersLayoutProps<any, any>) => {
   const { classes, isLandscape } = ownerState;
   const slots = {
@@ -104,10 +94,7 @@ const usePickerLayout = <TValue, TView extends DateOrTimeView>(
     ownerState: { ...props, wrapperVariant },
   });
   const toolbar =
-    toolbarHasView(toolbarProps) &&
-    toolbarHasValue<TValue, TView>(toolbarProps) &&
-    shouldRenderToolbar &&
-    !!Toolbar ? (
+    toolbarHasView(toolbarProps) && shouldRenderToolbar && !!Toolbar ? (
       <Toolbar {...toolbarProps} />
     ) : null;
 
diff --git a/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx b/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
index 9cdda9ae5..b0f829dfb 100644
--- a/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
+++ b/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
@@ -125,7 +125,7 @@ export function TimePickerToolbar<TDate extends unknown>(inProps: TimePickerTool
   const {
     ampm,
     ampmInClock,
-    value,
+    value = null,
     isLandscape,
     isMobileKeyboardViewOpen,
     onChange,
diff --git a/packages/x-date-pickers/src/internals/models/props/toolbar.ts b/packages/x-date-pickers/src/internals/models/props/toolbar.ts
index 5acd32d80..f721169ab 100644
--- a/packages/x-date-pickers/src/internals/models/props/toolbar.ts
+++ b/packages/x-date-pickers/src/internals/models/props/toolbar.ts
@@ -5,7 +5,7 @@ export interface BaseToolbarProps<TValue, TView extends DateOrTimeView>
   extends ExportedBaseToolbarProps {
   isLandscape: boolean;
   onChange: (newValue: TValue) => void;
-  value: TValue;
+  value?: TValue;
   /**
    * Currently visible picker view.
    */

```
Or maybe even making the value not required in the original type?
```diff
diff --git a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
index e12140b14..5fe81985e 100644
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
@@ -70,7 +70,7 @@ export const DateRangePickerToolbar = React.forwardRef(function DateRangePickerT
   const props = useThemeProps({ props: inProps, name: 'MuiDateRangePickerToolbar' });
 
   const {
-    value: [start, end],
+    value,
     isMobileKeyboardViewOpen,
     toggleMobileKeyboardView,
     rangePosition,
@@ -78,6 +78,7 @@ export const DateRangePickerToolbar = React.forwardRef(function DateRangePickerT
     toolbarFormat,
     className,
   } = props;
+  const [start, end] = value ?? [null, null];
 
   const localeText = useLocaleText<TDate>();
 
diff --git a/packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts b/packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts
index 21dc3ad5a..67b2f7e5d 100644
--- a/packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts
+++ b/packages/x-date-pickers/src/PickersLayout/PickersLayout.types.ts
@@ -70,8 +70,7 @@ export interface PickersLayoutSlotsComponentsProps<TValue, TView extends DateOrT
 }
 
 export interface PickersLayoutProps<TValue, TView extends DateOrTimeView>
-  extends Omit<UsePickerLayoutPropsResponseLayoutProps<TValue, TView>, 'value'> {
-  value?: TValue;
+  extends UsePickerLayoutPropsResponseLayoutProps<TValue, TView> {
   className?: string;
   children?: React.ReactNode;
   sx?: SxProps<Theme>;
diff --git a/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx b/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
index bd0a75b6a..5008ddd57 100644
--- a/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
+++ b/packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
@@ -13,16 +13,6 @@ function toolbarHasView<TValue, TView extends DateOrTimeView>(
   return toolbarProps.view !== null;
 }
 
-// TODO: Remove this functions. It get introduced to mark `value` prop in PickersLayoutProps as not required.
-// The true type should be
-// - For pickers value: TDate | null
-// - For rangepickers value: [TDate | null, TDate | null]
-function toolbarHasValue<TValue, TView extends DateOrTimeView>(
-  toolbarProps: BaseToolbarProps<TValue | undefined, TView> | any,
-): toolbarProps is BaseToolbarProps<TValue, TView> {
-  return toolbarProps.value !== undefined;
-}
-
 const useUtilityClasses = (ownerState: PickersLayoutProps<any, any>) => {
   const { classes, isLandscape } = ownerState;
   const slots = {
@@ -104,10 +94,7 @@ const usePickerLayout = <TValue, TView extends DateOrTimeView>(
     ownerState: { ...props, wrapperVariant },
   });
   const toolbar =
-    toolbarHasView(toolbarProps) &&
-    toolbarHasValue<TValue, TView>(toolbarProps) &&
-    shouldRenderToolbar &&
-    !!Toolbar ? (
+    toolbarHasView(toolbarProps) && shouldRenderToolbar && !!Toolbar ? (
       <Toolbar {...toolbarProps} />
     ) : null;
 
diff --git a/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx b/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
index 9cdda9ae5..b0f829dfb 100644
--- a/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
+++ b/packages/x-date-pickers/src/TimePicker/TimePickerToolbar.tsx
@@ -125,7 +125,7 @@ export function TimePickerToolbar<TDate extends unknown>(inProps: TimePickerTool
   const {
     ampm,
     ampmInClock,
-    value,
+    value = null,
     isLandscape,
     isMobileKeyboardViewOpen,
     onChange,
diff --git a/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts b/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
index 077cabcea..564eb8496 100644
--- a/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
+++ b/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
@@ -203,7 +203,7 @@ export interface UsePickerValueViewsResponse<TValue> {
  * Props passed to `usePickerLayoutProps`.
  */
 export interface UsePickerValueLayoutResponse<TValue> extends UsePickerValueActions {
-  value: TValue;
+  value?: TValue;
   onChange: (newValue: TValue) => void;
 }
 
diff --git a/packages/x-date-pickers/src/internals/models/props/toolbar.ts b/packages/x-date-pickers/src/internals/models/props/toolbar.ts
index 5acd32d80..f721169ab 100644
--- a/packages/x-date-pickers/src/internals/models/props/toolbar.ts
+++ b/packages/x-date-pickers/src/internals/models/props/toolbar.ts
@@ -5,7 +5,7 @@ export interface BaseToolbarProps<TValue, TView extends DateOrTimeView>
   extends ExportedBaseToolbarProps {
   isLandscape: boolean;
   onChange: (newValue: TValue) => void;
-  value: TValue;
+  value?: TValue;
   /**
    * Currently visible picker view.
    */

```

docs/data/date-pickers/custom-layout/custom-layout.md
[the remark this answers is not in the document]
joserodolfofreitas: I saw both in use on the documentation page, so I did a quick research on it yesterday, and "subcomponent" (without hyphen) seems way more commonly used. But they are both correct. 
In our docs, only MUI Base is consistently using it without, but it's also where Sam made his biggest contribution.

Generally speaking, I prefer the readability without the hyphen, but I'm fine with keeping the hyphen as well.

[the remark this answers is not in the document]
LukasTy: Gotcha, I also don't have a strong preference, as long as we are more or less consistent. 👌

packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
alexfauquette: Thanks for the different PoV, I may have spent too much time on Layout typing that I forget I could modify others 👍

Since toolbar value is not controlled, it could be `undefined`, and that's already a better solution than what I proposed

But the cleanest solution could be about the props type generator. I don't get why it set `Required` to a type `any`. If it is of type `any`, it can contain `null` or `undefined` and so it's not required

LukasTy: > But the cleanest solution could be about the props type generator. I don't get why it set `Required` to a type `any`. If it is of type `any`, it can contain `null` or `undefined` and so it's not required

You could look into [this solution](https://github.com/mui/mui-x/pull/7225#pullrequestreview-1226827528) suggested by Olivier. 🤔

docs/data/date-pickers/custom-layout/custom-layout.md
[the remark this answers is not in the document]
alexfauquette: Ok, then I will do the modification an enforce the consistency :)

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -0,0 +1,95 @@
...
+By default it renders a `div` with `display: grid`.
+Such that all sub-components are placed in a 3 by 3 [CSS grid](https://developer.mozilla.org/fr/docs/Web/CSS/CSS_Grid_Layout).
+
+Subcomponents are wrapped inside layout helper, responsible for placing it into the grid.
```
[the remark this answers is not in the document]
alexfauquette: Totally false since I modified the behavior 🙈

docs/data/date-pickers/custom-layout/custom-layout.md
alexfauquette: It's a broader topic, about how to propagate correctly elements

I updated the demo as follow such that users can to customization directly in the docs, and added a warning such that they know about it

```diff
--- a/docs/data/date-pickers/custom-layout/MovingActions.tsx
+++ b/docs/data/date-pickers/custom-layout/MovingActions.tsx
@@ -10,7 +10,7 @@ import { Unstable_StaticNextDatePicker as StaticNextDatePicker } from '@mui/x-da
 import { PickersActionBarProps } from '@mui/x-date-pickers/PickersActionBar';
 
 function ActionList(props: PickersActionBarProps) {
-  const { onAccept, onClear, onCancel, onSetToday } = props;
+  const { onAccept, onClear, onCancel, onSetToday, className } = props;
   const actions = [
     { text: 'Accept', method: onAccept },
     { text: 'Clear', method: onClear },
@@ -18,7 +18,7 @@ function ActionList(props: PickersActionBarProps) {
     { text: 'Today', method: onSetToday },
   ];
   return (
-    <List>
+    <List className={className}>
       {actions.map(({ text, method }) => (
         <ListItem key={text} disablePadding>
```

packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
flaviendelangle: I like Olivier's solutio here

oliviertassinari:
I think that it's great to allow developers to assemble the different parts of the data grid as they see fit!

The approach with `usePickerLayout` is interesting, first time I see this. What was the design constraint for now exposing React components? For example, in https://next--material-ui-x.netlify.app/x/react-date-pickers/custom-layout/#dom-customization, a developer can't add props to the `{toolbar}`, e.g. a test-id attribute. I'm asking coming from this context: https://twitter.com/adamwathan/status/1275140854810841091.

flaviendelangle:
The rational is that each of these component is a slot of it own
So we did not want to introduce yet another way to pass props to it.

oliviertassinari:
This sounds fair.

PR #6869, opened 2022-11-16
[docs] Remove `LocalizationProvider` from previews

flaviendelangle:
## Goal

Have smaller preview on the picker doc
We discussed it with @joserodolfofreitas yesterday.

## Changes

- [x] Remove `LocalizationProvider` wrapper

```tsx
<LocalizationProvider dateAdapter={AdapterDayjs}>
  <DateField label="Disabled" defaultValue={defaultValue} disabled />
</LocalizationProvider>
```

Becomes

```tsx
<DateField label="Disabled" defaultValue={defaultValue} disabled />
```

- [x] Remove nested wrappers inside `LocalizationProvider`

```tsx
<LocalizationProvider dateAdapter={AdapterDayjs}>
  <Stack spacing={2} sx={(theme) => ({ width: theme.spacing(48) })}>
    <DateField label="Disabled" defaultValue={defaultValue} disabled />
    <DateField label="Read only" defaultValue={defaultValue} readOnly />
  </Stack>
</LocalizationProvider>
```

Becomes

```tsx
<DateField label="Disabled" defaultValue={defaultValue} disabled />
<DateField label="Read only" defaultValue={defaultValue} readOnly />
```

I am not applying this logic to all wrappers because some DataGrid examples (and Grid examples on the core) do need their wrappers ([this one](https://next.mui.com/x/react-data-grid/layout/#predefined-dimensions) for instance).
But it would be nice to be able to add a comment on the example to disable the wrapper instead of disabling the nesting wrapper removal  like today.

PR #6902, opened 2022-11-17
[docs] Create the doc of the new date pickers

flaviendelangle:
- [x] Copy the current state of `docs/date/date-pickers/date-picker` in `docs/date/date-pickers/legacy-date-picker` as make this new page accessible
- [x] Adapt the examples of  `docs/date/date-pickers/date-picker` to use the new pickers
- [x] Create a new *Controlled vs uncontrolled value* demo for the new date picker (to match the approach of the other input-like component both on X and on the Core)
- [x] Remove useless props in the demos to keep them as focused as possible (do not control when not needed, don't set custom views if it has nothing to do with the feature demonstrated etc...)

I will do a follow up with all the other pickers.
You can ignore all the `/legacy-pickers` files, it's just a copy paste.
The `/pickers` files are interesting to see the actual diff between the old and the new pickers.


docs/data/date-pickers/date-picker/HelperText.tsx
```diff
@@ -1,24 +1,18 @@
...
     <LocalizationProvider dateAdapter={AdapterDayjs}>
-      <DatePicker
+      <NextDatePicker
         label="Helper text example"
-        value={value}
-        onChange={(newValue) => {
-          setValue(newValue);
+        componentsProps={{
+          input: {
+            helperText: 'MM / DD / YYYY',
```
flaviendelangle, line 13: Here we have a small problem

The format is determined inside `useDateField`
But `useDateField` is called with the props resolved from the slots.

The only solution I see to keep this feature possible is to duplicate the format logic (extract https://github.com/mui/mui-x/blob/next/packages/x-date-pickers/src/DateField/useDateField.ts#L25 inside a tiny hook) and call it on `DateField` as well)
Same for the other fields

If you have another solution, I am interested.

docs/data/date-pickers/date-picker/CustomInput.tsx
```diff
@@ -1,27 +1,35 @@
...
-import { DatePicker } from '@mui/x-date-pickers/DatePicker';
+import { Unstable_NextDatePicker as NextDatePicker } from '@mui/x-date-pickers/NextDatePicker';
 
-export default function CustomInput() {
-  const [value, setValue] = React.useState<Dayjs | null>(dayjs('2022-04-07'));
+type BrowserInputProps = TextFieldProps & {
+  ownerState?: any;
+};
+
+const BrowserInput = (props: BrowserInputProps) => {
```
flaviendelangle, line 13: I'm in favor of removing this demo in the future, in favor of detailed sections in the *Custom sub-components* page.

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -22,26 +30,33 @@ The date picker is rendered as a modal dialog on mobile, and a textbox with a po
 
 {{"demo": "BasicDatePicker.js"}}
 
-## Static mode
+## Uncontrolled vs. Controlled
 
-It's possible to render any date picker without the modal/popover and text field. This can be helpful when building custom popover/modal containers.
+The component can be controlled or uncontrolled
```
LukasTy, line 35: Nitpick, but I'd suggest aligning the heading with the content, i.e.:
```diff
-Uncontrolled vs. Controlled
+Controlled vs. Uncontrolled
```
or 
```diff
- The component can be controlled or controlled
+ The component can be uncontrolled or controlled
```

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
```diff
@@ -100,7 +100,7 @@ export interface UsePickerValueBaseProps<TValue, TError> {
    * @param {TValue} value The new value.
    * @param {FieldChangeHandlerContext<TError>} The context containing the validation result of the current value.
    */
-  onChange?: PickerChangeHandler;
+  onChange?: PickerChangeHandler<TValue, TError>;
```
LukasTy, line 103: 🙏

LukasTy (commented):
Looks great overall. 👍 
I've left a few minor comments.

One question stopping from approving it: How are we planning to approach `API` pages?
To me it would seem that we should basically do the same thing—keep adding new api pages with every component swap PR and possibly try to adopt similar naming strategy to them 🤔

flaviendelangle:
@LukasTy 
> One question stopping from approving it: How are we planning to approach API pages?

We just have to turn the `includeUnstableComponents` flag to `true` on `getTypescriptProject`
I created it to have the proptypes working but I wait for the doc to be ready before generating those pages :+1:

docs/data/date-pickers/date-picker/CustomField.tsx.preview
```diff
@@ -0,0 +1,9 @@
+<LocalizationProvider dateAdapter={AdapterDayjs}>
```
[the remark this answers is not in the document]
flaviendelangle: Your vision seems to be working great :laughing: 
I renamed the demo and forgot to remove the preview, fixed

docs/data/date-pickers/date-picker/SubComponentsCalendars.tsx
```diff
@@ -1,39 +1,24 @@
...
+          <MonthCalendar defaultValue={dayjs('2022-04-07')} />
         </Grid>
         <Grid item xs={12} md={6}>
-          <YearCalendar
-            value={value}
-            minDate={minDate}
-            maxDate={maxDate}
-            onChange={(newValue) => setValue(newValue)}
-          />
+          <YearCalendar defaultValue={dayjs('2022-04-07')} />
```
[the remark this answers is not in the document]
flaviendelangle: > but also not having the min and max dates on this calendar causes the calendar to render A LOT of years and also shows a bug—the view is not scrolled into the selected year

For me the minDate / maxDate were defined here because in the past they were mandatory (see #4814)
But if we think that they are useful in the demo, I can of course add them back.

For the bug, I thought @alexfauquette fixed it some time ago, but maybe not in this situation.

docs/data/date-pickers/date-picker/date-picker.md
flaviendelangle: I copied the `TextField` wordings
https://mui.com/material-ui/react-text-field/#uncontrolled-vs-controlled

But I do agree it's not 100% coherent
I prefer your 2nd proposal (to match the order in the demo)

docs/data/date-pickers/date-picker/DatePickerValue.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <LocalizationProvider dateAdapter={AdapterDayjs}>
+      <Stack spacing={2} direction="row">
+        <NextDatePicker
+          label="Uncontrolled picker"
+          defaultValue={dayjs('2022-04-07')}
+        />
+        <NextDatePicker
+          label="Controlled picker"
+          value={value}
+          onChange={(newValue) => setValue(newValue)}
```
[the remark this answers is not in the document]
flaviendelangle: To be honest, we very often recreate a method (same on the grid model setters) and I don't know why
Maybe we are just doing it because we always did it :laughing: 

No problem using the shorter version for me

flaviendelangle:
> To me it would seem that we should basically do the same thing—keep adding new api pages with every component swap PR and possibly try to adopt similar naming strategy to them thinking

Just, those pages being generated, they will always have the name of their component.
So it will be `NextDatePicker` etc... for the new and `DatePicker` etc... for the legacy.
And then when we rename  the components they will change.

But we should make sure that the demos (see #6909) are sent to the right page.

docs/data/date-pickers/date-picker/HelperText.tsx
flaviendelangle: We already callbacks on the slot (when they are built using `useSlotProps`).
But in the params of your function, you don't have the format yet, because the function is called before `useDateField` which computes the format.

LukasTy:
> So it will be `NextDatePicker` etc... for the new and `DatePicker` etc... for the legacy.
> And then when we rename the components they will change.

Yeah, I think that in the alpha/beta phase such naming inconsistency shouldn't be too much of a trouble. 👌

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
flaviendelangle: I thought I did a PR when we found the issue weeks ago, I was surprised to have the same problem again :see_no_evil:

docs/data/date-pickers/date-picker/HelperText.tsx
LukasTy: Thinking about this case a bit more—I don't think that we have any sort of problem here.
Static props can be used to pass static props, but in case something more advanced is needed with access to real/current props—the custom component can be passed down. 👍

docs/data/date-pickers/date-picker/DatePickerValue.tsx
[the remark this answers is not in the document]
alexfauquette: Maybe for pedagogical reasons. The long version explicitly says to newbies that the `onChange` is expecting a function which has the new value in argument. Moreover not sure every one is super familiar with setter accepting a function

docs/data/date-pickers/date-picker/HelperText.tsx
flaviendelangle: I feel like being able to put the format in the helper test is something quite a lot of people want.
And not being able to easily do it (without loosing the automatic format of the pickers / field) is not great.

But we can see if people complain about it, we have solutions if needed

docs/data/date-pickers/date-picker/SubComponentsCalendars.tsx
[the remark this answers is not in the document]
alexfauquette: Yes, it sounds like a bug in the computation of the scroll, because the scroll is done (scrollbar is not at the top) but a bit bellow the expected time (2022) Need to investigate in another PR

image

alexfauquette (approved):
That's a solid starting point 👍

docs/data/date-pickers/date-picker/HelperText.tsx
flaviendelangle: > olution could be to let dev compute the format helper themself:
> 
> ```
> adapter.getFormatHelperText(adapter.formats.keyboardDate)
> ```

Soor @alexfauquette I missed your answer
Sure the way it was done was hacky, and if we want to support it, I'm in favor of passing the format in a `format` property, not a `placeholder` property.

The only issue with your proposal is that the user need to know the format.
He would loose the ability to have the picker decide the format based on the locale / the views / the meridiem etc...

Maybe that's not that big of a deal, we can wait for feedbacks.

docs/data/date-pickers/date-picker/DatePickerValue.tsx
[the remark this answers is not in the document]
flaviendelangle: > Moreover not sure every one is super familiar with setter accepting a function

But we are not using this feature here right ?

docs/data/date-pickers/date-picker/JalaliDatePicker.tsx
```diff
@@ -1,20 +1,12 @@
...
   return (
-    <LocalizationProvider dateAdapter={AdapterJalali}>
-      <DatePicker
-        mask="____/__/__"
-        value={value}
-        onChange={(newValue) => setValue(newValue)}
-        renderInput={(params) => <TextField {...params} />}
-      />
+    <LocalizationProvider dateAdapter={AdapterDateFnsJalali}>
+      <NextDatePicker defaultValue={new Date(2022, 3, 7)} />
```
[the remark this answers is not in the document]
flaviendelangle: I would be in favor of removing it once we merge https://github.com/mui/mui-x/pull/6869 (and improve the demos their with RTL of course)

docs/data/date-pickers/date-picker/HelperText.tsx
LukasTy: > Maybe that's not that big of a deal, we can wait for feedbacks.

I'd say that having the option to access current localized `format` in one way or another for different use cases might be the expectation by users—especially ones used to using previously available `placeholder` prop.
Based on this assumption—it would be worthwhile to think of the easiest way to expose it.
Maybe it could be initialized in the components `format` props? 🤔  Now it seems to be empty if user did not provide any format...

flaviendelangle: > Maybe it could be initialized in the components format props? 

Right now the format is initialized in `useDateField` (or the equivalent for other fields)
Which is called after the resolution of the slots because it needs the result from this resolution
But we could either

1. Make the `format` param mandatory in `useDateField` and apply its default value inside the component instead of the hook

2. Apply the default value both in the component and in the hook (more duplication but people using the hook directly could still benefit from it)

PR #6909, opened 2022-11-18
[docs] Add Demos section on pickers api pages

flaviendelangle:
I reproduced roughly the logic of the core to find the demos linked to each api page.
We have to list the component in the `components` header of the doc markdown pages for them to be found when building the api pages.

A few examples of the demos registered:

- `DateRangePicker`

image

- `MobileDateRangePicker`

image

- `LocalizationProvider`

image

- `TimeClock`

image


oliviertassinari:
It could be good to enforce this like in https://github.com/mui/material-ui/blob/8bd75c8414334cf377d499f4a6e07e2cdd99d4e3/docs/scripts/ApiBuilders/ComponentApiBuilder.ts#L627-L633

flaviendelangle:
> Maybe it would make sense to enforce having at least one demo for all API pages like in MUI Core https://github.com/mui/material-ui/blob/8bd75c8414334cf377d499f4a6e07e2cdd99d4e3/docs/scripts/ApiBuilders/ComponentApiBuilder.ts#L627-L633

Right now he would have 3 errors

```
[ 'DateTimePickerTabs', 'PickerStaticWrapper', 'DateRangeCalendar' ]
```

`PickerStaticWrapper` should probably not be documented and the two others should be referenced.
So I'm indeed in favor of fixing and adding the error :+1:

docs/data/pages.ts
```diff
@@ -264,68 +264,160 @@ const pages: MuiPage[] = [
         pathname: '/x/api/date-pickers',
         title: 'API Reference',
         children: [
-          { pathname: '/x/api/date-pickers', title: 'Index' },
-          { pathname: '/x/api/date-pickers/date-calendar', title: 'DateCalendar' },
-          { pathname: '/x/api/date-pickers/date-picker', title: 'DatePicker' },
+          { pathname: '/x/api/date-pickers/date-calendar/', title: 'DateCalendar' },
```
flaviendelangle, line 267: @LukasTy I added the missing pages and sorted them alphabetically

We only miss `PickerStaticWrapper` which is going away on v6 :+1:

LukasTy: There is some sort of an issue with routing/path setup as going to any of the specific API doc pages does not mark that page as active.

alexfauquette (commented):
Did not understand why, but most of the links are broken

[Here](https://deploy-preview-6909--material-ui-x.netlify.app/x/api/date-pickers/month-calendar/#demos) the demos are linking to 
https://deploy-preview-6909--material-ui-x.netlify.app/date-pickers/date-picker/date-picker/

On the other side, [validation](https://deploy-preview-6909--material-ui-x.netlify.app/x/react-date-pickers/validation/#api) demo page links to 404 


By the way, I'm wondering if it makes sens for pages such as localization or validation to have API section

docs/data/date-pickers/date-picker/date-picker.md
```diff
@@ -1,7 +1,7 @@
 ---
 product: date-pickers
 title: React Date Picker component
-components: DateCalendar, DatePicker, DayCalendarSkeleton, DesktopDatePicker, MobileDatePicker, MonthCalendar, PickersDay, StaticDatePicker, YearCalendar
+components: NestDatePicker, DesktopNextDatePicker, MobileNextDatePicker, StaticNextDatePicker, DateCalendar, MonthCalendar, YearCalendar, PickersDay, DayCalendarSkeleton
```
alexfauquette, line 4:
```suggestion
components: NextDatePicker, DesktopNextDatePicker, MobileNextDatePicker, StaticNextDatePicker, DateCalendar, MonthCalendar, YearCalendar, PickersDay, DayCalendarSkeleton
```

flaviendelangle:
> By the way, I'm wondering if it makes sens for pages such as localization or validation to have API section

My assumption was that the goal of the `Demo` section is to help people find the demos linked to the component the are looking at.
And a lot of those demos are in the Validation & co pages.

With that being said, I am not a fan of the whole current logic (you reach the API page and then you have a list of links without context).
I would prefer to link some of it at the prop level somehow.

docs/scripts/api/buildComponentsDocumentation.ts
```diff
@@ -183,19 +177,57 @@ function parseComponentSource(src: string, componentObject: { filename: string }
...
+    return demos;
+  }
+
+  return pagesMarkdown
+    .filter((page) => {
+      if (!page.pathname.includes('/')) {
+        return false;
+      }
+
+      return page.pathname.indexOf('/x') && page.components.includes(componentName);
```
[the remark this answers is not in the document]
flaviendelangle: Copy pasted from the core :laughing: 

Their, they check if the link is a `base` or a `joy` or a `material`. To avoid having the `/joy/button` page matching the Material Button API page.
But for X it's probably useless indeed...

alexfauquette:
We might have a miss understanding. I'm not talking about links from API to Demo, but about the opposite. Especially this page:
https://deploy-preview-6909--material-ui-x.netlify.app/x/react-date-pickers/localization/#api

image

flaviendelangle:
Oh, I totally forgot that the link was added in both direction.
Well, we can't have one without the other.
If this section is problematic, I can remove the components from the localization page, but the localization page will no longer appear on those components API pages.

docs/scripts/api/buildComponentsDocumentation.ts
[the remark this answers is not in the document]
flaviendelangle: Done

docs/data/pages.ts
flaviendelangle: Fixed :+1:

flaviendelangle:
> On the other side, [validation](https://deploy-preview-6909--material-ui-x.netlify.app/x/react-date-pickers/validation/#api) demo page links to 404

Found out why, we need to define the `product` in the page headers as well
Could be nice to deduce it from the routing in the future.

docs/data/pages.ts
```diff
@@ -301,11 +296,18 @@ const pages: MuiPage[] = [
...
           {
-            pathname: '/x/api/date-pickers/day-calendar-skeleton',
-            title: 'DayCalendarSkeleton',
+            pathname: '/x/api/date-pickers/date-range-picker',
+            title: 'DateRangePicker',
+            plan: 'pro',
+          },
+          { pathname: '/x/api/date-pickers/date-time-field', title: 'DateTimeField' },
+          {
+            pathname: '/x/api/date-pickers/static-date-time-picker-tabs',
```
LukasTy, line 306:
```suggestion
            pathname: '/x/api/date-pickers/date-time-picker-tabs',
```

docs/data/pages.ts
```diff
@@ -328,9 +347,64 @@ const pages: MuiPage[] = [
...
+          {
+            pathname: '/x/api/date-pickers/single-input-date-time-range-field',
+            title: 'SingleInputDateTimeRangeField',
+            plan: 'pro',
+          },
+          {
+            pathname: '/x/api/date-pickers/single-input-time-range-field',
+            title: 'SingleInputTimeRangeField',
+            plan: 'pro',
+          },
```
LukasTy, line 407: Both of these do not exist yet.
Are they here because this PR is going to be merged only after they are implemented? 🤔

LukasTy (changes requested):
As Alex mentioned, we have one main issue—none of the links for demos in API pages are actually correct.
They have an incorrect pathname start and a duplication of the component name.
```diff
-/date-pickers/date-range-picker/date-range-picker/
+/x/react-date-pickers/date-range-picker/
```

docs/data/pages.ts
flaviendelangle: No it's a mistake, fixed

flaviendelangle:
> As Alex mentioned, we have one main issue—none of the links for demos in API pages are actually correct.
They have an incorrect pathname start and a duplication of the component name.


It should have been fixed on a commit I did earlier this morning
If you still have the issue, can you link me a page where it fails ?

alexfauquette (approved):
Links sound to be fixed 👍

LukasTy:
Am I missing something and they are all broken just for me? 🙈 🤷 

https://user-images.githubusercontent.com/4941090/204789362-14e05f20-872d-4675-b448-6f069bd6f549.mov

flaviendelangle:
OK we weren't talking about the same links
It should work now :+1: 
I struggled to understand how the core goes from the file path to the route.

docs/data/pages.ts
LukasTy: This is a problem once again 🙈 
Is this list auto-generated? Or it was a merge conflict issue?

LukasTy (approved):
LGTM. 🙌 
Great work. 👍

docs/data/pages.ts
LukasTy: You probably missed this comment? 😢

flaviendelangle: I fixed in the index but not here, my bad...

PR #6914, opened 2022-11-18
[pickers] Fix focus management

alexfauquette:


packages/x-date-pickers/src/internals/utils/viewRenderers.tsx
```diff
@@ -21,7 +21,6 @@ interface DateViewRendererProps<TDate>
 export const renderDateView = <TDate extends unknown>(props: DateViewRendererProps<TDate>) => (
   <DateCalendar
     {...props}
-    autoFocus
```
alexfauquette, line 24: `autoFocus` has no effect if `focusedView` is controlled

packages/x-date-pickers/src/DateCalendar/DateCalendar.tsx
```diff
@@ -449,17 +449,17 @@ export const DateCalendar = React.forwardRef(function DateCalendar<TDate>(
...
-    }
-
-    // Set focus to the button when switching from a view to another
+    // If the view change and the focus was on the previouse view
+    // Then we update the focus.
     if (prevOpenViewRef.current === openView) {
       return;
     }
+
+    if (focusedView === prevOpenViewRef.current && openView !== focusedView) {
```
flaviendelangle, line 458: We can remove the 2nd check on this line.

If `focusedView === prevOpenViewRef.current`, then `openView !== focusedView`
Otherwise, we would have `openView === focusedView === prevOpenViewRef.current` 
And that's not possible given the check above (line 454)

flaviendelangle (approved):
The behavior make sense
If you could add a test case for the fix it does, that would be great

PR #6942, opened 2022-11-21
[core] Cleanup `describeValidation`

flaviendelangle:
- [x] Type the parameters of `describeConformance`
- [x] Move all `getOptions` outside of the `it` (we had some outside and some inside)
- [x] Remove the `ignoredProps` (was used on 2 components and it had no effect) => remove the `propsToTest` checks
- [x] Replace `isLegacyPicker` by `componentFamily`
- [x] Replace `skip` by checks inside the test suites based on `componentFamily`
- [x] Replace `name` by checks inside the test suites based on `componentFamily`

The main idea is two reduce the amount of ways to disable some tests and how to customize the default props.
We now have 2 properties:

- `views` => describes which part of dates can be edited (through the UI view or the field)
- `componentFamily` => describes the type of component being used (to set custom default props, or disable UI view tests on field for instance)

Plus `isSingleInput` on range validation which remains unchanged.


packages/x-date-pickers/src/tests/describeValidation/describeValidation.types.ts
```diff
@@ -0,0 +1,32 @@
...
+  render: (node: React.ReactElement) => MuiRenderResult;
+  // TODO: Export `Clock` from monorepo
+  clock: ReturnType<typeof createRenderer>['clock'];
+  after?: () => void;
+  componentFamily: DescribeValidationComponentFamily;
+  views: DateOrTimeView[];
+}
+
+export interface DescribeValidationOptions
+  extends DefaultizedProps<DescribeValidationInputOptions, 'views'> {
```
alexfauquette, line 24: Sounds you removed the optional aspect of `view`
```suggestion
  extends DescribeValidationInputOptions {
```

flaviendelangle: Good catch

PR #6947, opened 2022-11-21
[DataGrid] Allow disabling of buttons in column panel

### LINKED ISSUE #4297: [data-grid] Disable the `Hide All` button in the `GridColumnsPanel`
I'm looking for a way to disable just the "HIDE ALL" button inside the GridToolbarColumnsButton dropdown menu.

Any help to implement this is appreciated.

column menu

MBilalShafi:
Fixes #4297 

https://deploy-preview-6947--material-ui-x.netlify.app/x/react-data-grid/column-visibility/#column-visibility-panel


packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -224,12 +236,14 @@ function GridColumnsPanel(props: GridColumnsPanelProps) {
         <rootProps.components.BaseButton
           onClick={() => toggleAllColumns(false)}
           {...rootProps.componentsProps?.baseButton}
+          disabled={disableHideAllButton}
```
MBilalShafi, line 239: These new props (disableHideAllButton, disableShowAllButton) should override `componentsProps.baseButton.disabled` as it's generic for base button, but the new props are explicit to column panel buttons.

MBilalShafi: It seems better to hide both if both props are passed, I'll update it.

packages/grid/x-data-grid/src/components/panel/GridColumnsPanel.tsx
```diff
@@ -220,20 +232,31 @@ function GridColumnsPanel(props: GridColumnsPanelProps) {
...
+          {!disableHideAllButton ? (
+            <rootProps.components.BaseButton
+              onClick={() => toggleAllColumns(false)}
+              {...rootProps.componentsProps?.baseButton}
+              disabled={disableHideAllButton}
+            >
+              {apiRef.current.getLocaleText('columnsPanelHideAllButton')}
+            </rootProps.components.BaseButton>
+          ) : (
+            <span />
```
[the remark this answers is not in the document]
MBilalShafi: We may have to change `display: block` as float doesn't work great with [flexbox](https://stackoverflow.com/a/39194800), probably this approach is more suitable with flexbox?

PR #6954, opened 2022-11-22
[pickers] Support `onMonthChange` / `onYearChange`  in `day` view

### LINKED ISSUE #6953: [pickers] `onMonthChange` doesn't call after click to days outside the current month
### Duplicates

### Latest version

### Steps to reproduce 🕹

Link to live example: https://codesandbox.io/s/zen-hodgkin-8ijm8m?file=/demo.tsx

Steps:
1. Add `showDaysOutsideCurrentMonth`
2. Click to day outside current month
3. Only `onChange` callback was called. `onMonthChange` wasn't called.

### Current behavior 😯

`onMonthChange` wasn't called.

### Expected behavior 🤔

`onMonthChange` called.

### Context 🔦

By documentation:
> `onMonthChange` - Callback firing on month change

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #6953


packages/x-date-pickers/src/DateCalendar/DateCalendar.tsx
```diff
@@ -473,7 +476,7 @@ export const DateCalendar = React.forwardRef(function DateCalendar<TDate>(
         openView={openView}
         currentMonth={calendarState.currentMonth}
         onViewChange={setOpenView}
-        onMonthChange={(newMonth, direction) => handleChangeMonth({ newMonth, direction })}
+        onMonthChange={slideToMonth}
```
flaviendelangle, line 482: I now compute the direction inside `useCalendarState` to have a single month change method exposed

packages/x-date-pickers/src/DateCalendar/useCalendarState.tsx
```diff
@@ -71,18 +72,27 @@ export const createCalendarStateReducer =
           !disableSwitchToMonthOnDayFocus &&
           !utils.isSameMonth(state.currentMonth, action.focusedDay);
 
+        let slideDirection: SlideDirection;
+        if (!needMonthSwitch) {
+          slideDirection = state.slideDirection;
```
flaviendelangle, line 77: Before, the slide to the new month was caused by the `useEffect` tracking the external value.
So the focus of the clicked day was already done
But now it is the click itself  which caused the slide (like when you click on the array left / right).

Without this check, the focus was resetting the slide direction, causing it to go on the wrong direction in some scenario.

packages/x-date-pickers/src/DateCalendar/useCalendarState.tsx
```diff
@@ -139,36 +149,25 @@ export const useCalendarState = <TDate extends unknown>({
     slideDirection: 'left',
   });
 
-  const handleChangeMonth = React.useCallback(
-    (payload: ChangeMonthPayload<TDate>) => {
-      dispatch({
-        type: 'changeMonth',
-        ...payload,
-      });
+  const slideToMonth = useEventCallback((newDate: TDate) => {
```
flaviendelangle, line 152: Renamed to make clear that we are not updating any value, just "meta data" used to handle the visually displayed month

packages/x-date-pickers/src/DateCalendar/DateCalendar.tsx
```diff
@@ -329,6 +313,25 @@ export const DateCalendar = React.forwardRef(function DateCalendar<TDate>(
     disableFuture,
   });
 
+  const handleValueChange = useEventCallback(
+    (newValue: TDate | null, selectionState?: PickerSelectionState) => {
+      setValue(newValue);
+      onChange?.(newValue, selectionState);
+
+      if (newValue != null) {
```
flaviendelangle, line 321: We now do the check here to make sure that any value change causes a call to `onMonthChange` if needed

alexfauquette (commented):
The implementation sounds good to be a fix on `master`. For `next` I'm wondering if we should rename `onMonthChange`.

When I see `onMonthChange` I expect it to get called when the selected value moves from one month to an other. But it's also called when I press Next/Previous month buttons.

So maybe `onVisibleMonthChange` would be more appropriate

flaviendelangle:
> When I see `onMonthChange` I expect it to get called when the selected value moves from one month to an other. But it's also called when I press Next/Previous month buttons.

Good point
@LukasTy if you have an opinion
I don't really know to be honest

packages/x-date-pickers/src/DateCalendar/useCalendarState.tsx
LukasTy: If we are going this route, then maybe `onYearChange` should also be called when year changes when navigating months (not only when choosing a date in a different month—which is also in a different year)? 🤔

LukasTy (commented):
IMO—this implementation seems incorrect.
Even the documentation states that `onYearChange` and `onMonthChange` are callbacks called when the year or month changes, but now `onMonthChange` is called when navigating months—which contradicts the docs.

I'm with Alex on this one.
If we want to emit `year` or `month` navigation changes—they ought to be separate callbacks in order to avoid possible head-aches.

Issue #5492, opened 2022-07-14
How to use coretctly filter with multiselect


michGreg:
### Order ID 💳

44899

### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### The problem in depth 🔍

I set valueOptions wich is array of objects and operator isAnyOf, but when i read filterModel it is changed to array of values. The issue is that info tooltip of filtered values shows values instead of labels. Is there any way how to read to tooltip of labels instead of values or disable tooltip info?

Column setting
```
{
    disableColumnMenu: true,
    field: 'signalStrength',
    filterOperators: [{
      value: 'isAnyOf',
      InputComponent: GridFilterInputMultipleSingleSelect,
    }],
    headerName: formatMessage(messages.DevicesDataGrid.columns.signalStrength),
    hideable: true,
    renderCell: (device) => (
      <DeviceSignalStrengthStatus
        signalStrength={device.value.signalStrength}
      />
    ),
    sortable: true,
    valueOptions: [
      { label: 'Very strong', value: '1' },
      { label: 'Strong', value: '2' },
      { label: 'Good', value: '3' },
      { label: 'Poor', value: '4' },
      { label: 'Unknown', value: '5' },
    ],
  }
```

Tooltip info
![image](https://user-images.githubusercontent.com/106229368/178961590-0d646a15-7778-4d59-9734-ea8128f3a265.png)

Filter model
![image](https://user-images.githubusercontent.com/106229368/178961846-ca4708ca-ce34-4b30-8886-95b6a0959193.png)

### Your environment 🌎

<details>
  <summary>`npx @mui/envinfo`</summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

m4theushw:
We have this same problem with every operator. In the tooltip, we display the value as it is, without any formatting. Another example is the date operator which doesn't format the value to the user's locale. The date should be displayed as dd/mm/YYYY to me.

<img width="300" src="https://user-images.githubusercontent.com/42154031/179295532-950a6438-a861-4d9f-9e26-fecf1cd50823.png" alt="image" style="max-width: 100%;">

I propose to add a function to `GridFilterOperator` to format the value of the filter item to another string which it's safe to be displayed in the UI.

```diff
diff --git a/packages/grid/x-data-grid/src/models/gridFilterOperator.ts b/packages/grid/x-data-grid/src/models/gridFilterOperator.ts
[...]

PR #6956, opened 2022-11-22
[DataGrid] Allow to customize the value displayed in the filter button tooltip

ithrforu:
Related to https://github.com/mui/mui-x/issues/5492 (a follow-up is to fix the built-in operators)

I faced a problem with current filter values in GridToolbarFilterButton tooltip title (use DataGridPremium) and found [issue](https://github.com/mui/mui-x/issues/5492#issuecomment-1185835877) with @m4theushw comment. This problem looks like:
image
image

In this PR i added the ability (from @m4theushw proposal) to specify a function to format the filter value in GridToolbarFilterButton tooltip title.


packages/grid/x-data-grid/src/components/toolbar/GridToolbarFilterButton.tsx
```diff
@@ -78,6 +78,14 @@ const GridToolbarFilterButton = React.forwardRef<HTMLButtonElement, GridToolbarF
           .getLocaleText(`filterOperator${capitalize(item.operatorValue!)}` as GridTranslationKeys)!
           .toString();
 
+      const getFilterItemValue = (item: GridFilterItem): string => {
+        const { getValueAsString } = lookup[item.columnField!].filterOperators!.find(
+          (operator) => operator.value === item.operatorValue,
+        )!;
+
+        return getValueAsString ? getValueAsString(item.value) : item.value;
+      };
```
MBilalShafi, line 87: Filter item got [updated](https://github.com/mui/mui-x/pull/6985), can you rebase with latest `next` ?
```suggestion
      const getFilterItemValue = (item: GridFilterItem): string => {
        const { getValueAsString } = lookup[item.field!].filterOperators!.find(
          (operator) => operator.value === item.operator,
        )!;

        return getValueAsString ? getValueAsString(item.value) : item.value;
      };
```

ithrforu: All done.

MBilalShafi:
@ithrforu Thanks for the contribution, it seems good, can we test the newly added function `getValueAsString` and add some example usage in demos to improve usability?

ithrforu:
@MBilalShafi. Thank you for attention. Of course. Should I create new pull request or continue here?

MBilalShafi:
> Should I create new pull request or continue here

It will be better if we have it as part of this pull request so that we push this new addition tested and more usable.

ithrforu:
@MBilalShafi, I added an example and rebased with latest next, but got error in [docs](https://deploy-preview-6956--material-ui-x.netlify.app/x/react-data-grid/filtering/) page.
image

How can i test `getValueAsString`?

MBilalShafi:
> How can I test `getValueAsString`?

You can create a new test case in https://github.com/mui/mui-x/blob/next/packages/grid/x-data-grid/src/tests/filtering.DataGrid.test.tsx covering this use-case and test if it works as expected.

ithrforu:
@MBilalShafi, all done.

docs/data/data-grid/filtering/filtering.md
```diff
@@ -203,6 +203,10 @@ The [`valueFormatter`](/x/react-data-grid/column-definition/#value-formatter) is
 If the column has a [`valueGetter`](/x/react-data-grid/column-definition/#value-getter), then `params.value` will be the resolved value.
 :::
 
+:::info
+The `getValueAsString` function allows you to convert the value of a filter item to a human-readable form before displayed in the UI of filter button tooltip.
```
MBilalShafi, line 207:
```suggestion
The filter button displays a tooltip on hover if there are active filters. Pass [`getValueAsString`](/x/api/data-grid/grid-filter-operator/) in the filter operator to customize or convert the value to a more human-readable form.
```

ithrforu: Thanks for your edits and remarks. Text is fixed.

MBilalShafi (approved):

ithrforu:
Is it bug?
MUI

ithrforu:
@m4theushw, I mean I hovered filter button but "Show the code" tooltip displayed.

LukasTy:
Indeed, this PR might have some sort of regression in hover event bubbling or something along these lines. The current versions do not display `Show the full source` tooltip on demo hover. 🙈

LukasTy:
> I can also reproduce this bug in https://material-ui.netlify.app/material-ui/react-button/ so certainly is not a regression introduced by this PR but from the core.

Hm. Good point. I tried testing a local dev run and did not see this behaviour, that's why I thought that it might have something to do with this PR, but I do not see any `@mui/monorepo` bumps to warrant that. 🙈

ithrforu:
Can i do something?

MBilalShafi:
It's all good @ithrforu, I'll merge this now.
Regarding the above-mentioned regression, I'll create a ticket in core if there's not one already.

LukasTy:
> It's all good @ithrforu, I'll merge this now. Regarding the above-mentioned regression, I'll create a ticket in core if there's not one already.

It might have also been a desired change from MUI Core, we just weren't aware of it. But I couldn't find a recently merged PR related to it. 🤔

PR #6625, opened 2022-10-25
[docs] Rework localization doc pages (WIP)

flaviendelangle:

PR #6972, opened 2022-11-23
[pickers] Throw error when using adapter from `@date-io`

flaviendelangle:
Need #6625


flaviendelangle:
Yes, I would prefer to merge #6625 before to make sure the URL is final for v6

flaviendelangle:
Link updated

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -34,7 +34,7 @@ export const useField = <
 >(
   params: UseFieldParams<TValue, TDate, TSection, TForwardedProps, TInternalProps>,
 ): UseFieldResponse<TForwardedProps> => {
-  const utils = useUtils<TDate>() as MuiPickerFieldAdapter<TDate>;
+  const utils = useUtils<TDate>() as MuiPickersAdapter<TDate>;
```
LukasTy, line 37: It does not seem that this casting is needed anymore 🤔

packages/x-date-pickers/src/internals/hooks/useField/useFieldState.ts
```diff
@@ -39,7 +39,7 @@ export const useFieldState = <
 >(
   params: UseFieldParams<TValue, TDate, TSection, TForwardedProps, TInternalProps>,
 ) => {
-  const utils = useUtils<TDate>() as MuiPickerFieldAdapter<TDate>;
+  const utils = useUtils<TDate>() as MuiPickersAdapter<TDate>;
```
LukasTy, line 42: Same with this one

flaviendelangle:
@LukasTy too bad :laughing: 
Will do a quick automerged follow up

PR #6975, opened 2022-11-23
[docs] Fix live edit

oliviertassinari:
Revert #6681. Same bug as in https://github.com/mui/material-ui/pull/35118#issuecomment-1324911398.

HEAD, broken: https://next--material-ui-x.netlify.app/x/react-data-grid/#mit-version
PR, works: https://deploy-preview-6975--material-ui-x.netlify.app/x/react-data-grid/#mit-version

I have no idea what's wrong but it would be great to see if the problem is on our side or Next.js's side

Same bug on the master branch: https://mui.com/x/react-data-grid/.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -61,7 +61,7 @@
     "lz-string": "^1.4.4",
     "markdown-to-jsx": "^7.1.7",
     "marked": "^4.2.2",
-    "next": "^13.0.4",
+    "next": "^13.0.2",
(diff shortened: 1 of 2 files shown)


LukasTy (approved):

oliviertassinari:
cc @flaviendelangle in case you are on `dependencies` update duty later on. We will need to figure out what's wrong. This fix is only on the surface, I won't have the time to dive deeper.

PR #6958, opened 2022-11-22
[docs] Create the doc of the new time, date time and date range pickers

flaviendelangle:
Follow up on #6902

- [x] Apply the same copy / adaptation logic as on #6902 
- [x] Unify the section order and some of the wordings across the new pages
- [x] Use the grid introduced in the validation page to have clean label on the date range picker doc

PR #6989, opened 2022-11-24
[pickers] Fix toolbar on the new range pickers

flaviendelangle:
Fixes https://github.com/mui/mui-x/pull/6958#discussion_r1031359737

We did not pass the props to switch between the start and the end date to the toolbar

#### Breaking changes

- The `currentlySelectingRangeEnd` / `setCurrentlySelectingRangeEnd` props on the Date Range Picker toolbar have been renamed to `rangePosition` / `onRangePositionChange`.

  ```diff
   const CustomToolbarComponent = props => (
     <div>
  -    <button onChange={() => props.setCurrentlySelectingRangeEnd('end')}>Edit end date</button>
  +    <button onClick={() => props.onRangePositionChange('end')}>Edit end date</button>
  -    <div>Is editing end date: {props.currentlySelectingRangeEnd === 'end'}</div>
  +    <div>Is editing end date: {props.rangePosition === 'end'}</div>
     </div>
   )
   <DateRangePicker
     components={{
       Toolbar: CustomToolbarComponent
     }}
   />
  ```


packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.tsx
```diff
@@ -37,7 +36,7 @@ const StaticNextDateRangePicker = React.forwardRef(function StaticNextDateRangeP
     showToolbar: defaultizedProps.showToolbar ?? displayStaticWrapperAs === 'mobile',
   };
 
-  const { renderPicker } = useStaticPicker<DateRange<TDate>, TDate, 'day', typeof props>({
+  const { renderPicker } = useStaticRangePicker<TDate, 'day', typeof props>({
```
flaviendelangle, line 39: We no longer need the `TValue` generic since we have one hook for single item editing and one hook for range editing

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
```diff
@@ -173,6 +174,15 @@ export const useDesktopRangePicker = <
     },
   };
 
+  const componentsPropsForLayout: PickersViewLayoutSlotsComponentsProps = {
+    ...componentsProps,
+    toolbar: {
+      ...componentsProps?.toolbar,
+      currentlySelectingRangeEnd: currentDatePosition,
```
flaviendelangle, line 181: Should we update the name of those props now or later ?
They are based on the v5 naming (`currentlySelectingRangeEnd` should become `currentDatePosition`, same for the setter)

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.types.ts
```diff
@@ -43,7 +43,7 @@ interface PickersViewLayoutSlotsComponentsProps
   /**
    * Props passed down to the toolbar component.
    */
-  toolbar?: ExportedBaseToolbarProps;
+  toolbar?: ExportedBaseToolbarProps & Record<string, any>;
```
flaviendelangle, line 46: Otherwise, the typing in the range hooks complains that we try to pass additional props
This interface is private only so I don't think it's an issue to add the record

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.types.ts
```diff
@@ -0,0 +1,60 @@
...
+  extends ExportedPickersViewLayoutSlotsComponent,
+    Pick<PickersSlotsComponent, 'PaperContent'> {}
+
+export interface UseStaticRangePickerSlotsComponentsProps
+  extends ExportedPickersViewLayoutSlotsComponentsProps,
+    Pick<PickersSlotsComponentsProps, 'paperContent'> {
+  toolbar?: ExportedBaseToolbarProps;
+}
+
+export interface StaticRangeOnlyPickerProps {
```
LukasTy, line 24: We already have this type defined in `StaticOnlyPickerProps`.
Maybe it would make sense to `Pick` it from there instead of re-declaring?

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from 'react';
+import { styled } from '@mui/material/styles';
```
LukasTy, line 2: Nitpick: Maybe it would be worth trying to avoid `@mui/material` imports for new code whenever possible if we have plans to support Joy or other UI in the future?
In this case the following should work as well 🤔 
```suggestion
import { styled } from '@mui/system';
```

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.types.ts
LukasTy: Technically this is actually public. It's exposed via `themeAugmentation` and would allow code like this:
```ts
MuiPickersViewLayout: {
  defaultProps: {
    componentsProps: {
      toolbar: {
        // should complain that prop does not exist
        something: 'something'
      }
    }
  },
},
```

Honestly, I would probably be in favour maybe casting whatever is passed in or coming up with a different solution  🤷 🙈

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
flaviendelangle: My main complain with the `currentlySelectingRangeEnd` naming is that I would expect a boolean, not a position.
But my new name is maybe not great.
It is never used on public API so if we want to change it, now is the time.
`currentDateRangePosition` or `currentPositionInDateRange` would probably be more explicit.

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.tsx
flaviendelangle: We should check with the core team, but for me both are not equivalent.
They don't have the same default `theme` and `shouldForwardProp`.

That would be an interesting topic to discuss when we will want to stop relying on `@mui/material`.

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.types.ts
flaviendelangle: Indeed, and we have small inconsistencies between desktop and range desktop pickers as well, extending would prevent those

flaviendelangle: Done

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.tsx
LukasTy: Mhh, I didn't check all the code. Sorry for confusing on this one.
We will have to bring this topic to them someday. 👍

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
flaviendelangle: Or you turn it into an actual boolean and name it `isEditingEndOfRange`

LukasTy: Hmm... Boolean is always less flexible.
What if we will want/need additional positions along the line?
Like `startDate`, `startTime`, `endDate` and `endTime` 🤔 
Just extrapolating, that would probably not be the case, but if "enum" is working for now, I'd stick with it.

flaviendelangle: True

packages/x-date-pickers/src/internals/components/PickersViewLayout/PickersViewLayout.types.ts
flaviendelangle: You are right, I replace it with type casting.

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
flaviendelangle: I am proposing simplfy `rangePosition` and to only have `start` and `end` in it (even for time ranges and date time ranges).
And to use the view to distinguish the start date and start time for instance.

I'm not using `dateRangePosition` because it is available on internals that will be used by `TimeRangePicker`, which is a little weird.

LukasTy: Yeah, `rangePosition` should work as well. 👍 
The only small issue is that this component has it's props exposed via `themeAugmentation -> defaultProps` although, it might not be an issue for anyone to warrant a codemod, because they are required props and are always defined in our components.
The only issue could be if someone used the toolbar component for something externally. 🤔 
However small this chance could be, maybe it's worth noting this change in migration doc?

flaviendelangle: I is also breaking it they provide a custom toolbar, which is a valid use case.
I will list it in the breaking changes :+1:

LukasTy (approved):

PR #7002, opened 2022-11-25
[docs] Keep track of localization completion

alexfauquette:
The idea is to dynamically render the table with available locales, to encourage user to update translations

## Before

image

## After

image

https://deploy-preview-7002--material-ui-x.netlify.app/x/react-data-grid/localization/#supported-locales


alexfauquette:
- data grid: https://deploy-preview-7002--material-ui-x.netlify.app/x/react-data-grid/localization/#supported-locales

- date pickers: https://deploy-preview-7002--material-ui-x.netlify.app/x/react-date-pickers/localization/#supported-locales

MBilalShafi:
@alexfauquette In some languages it doesn't makes sense to translate some specific values or it makes more sense to keep them in English.

Do you think in those cases, it will be better to uncomment the English placeholders to mark it as completed, so the community doesn't bother about it anymore?

alexfauquette:
I think yes. Because when you uncomment a line, it means "The English word is understood in this localization". whereas a commented line could be "Nobody took time to translate this line".

scripts/l10n.ts
```diff
@@ -284,6 +287,63 @@ async function generateReport(missingTranslations: MissingTranslations) {
...
+            'If this tag is correct, add its name to the file `scripts/localeNames.js`',
+          ].join('\n'),
+        );
+      }
+      documentationReport.push({
+        languageTag,
+        importName,
+        localeName,
+        missingKeysNb: infoPerPackage[packageKey].missingKeys.length,
+        totalKeysNb: baseTranslationsNumber[packageKey],
```
LukasTy, line 331: Nitpick, but usually `number` is shortened to `no`, isn't it?
Or maybe we could ho with `Count` instead? 🤔 
```suggestion
        missingKeysCount: infoPerPackage[packageKey].missingKeys.length,
        totalKeysCount: baseTranslationsNumber[packageKey],
```

scripts/l10n.ts
```diff
@@ -284,6 +287,63 @@ async function generateReport(missingTranslations: MissingTranslations) {
...
+  const workspaceRoot = path.resolve(__dirname, '../');
+
+  packagesWithL10n.forEach(async ({ key: packageKey, documentationReportPath }) => {
+    const documentationReport: DocumentationReportItem[] = [];
+    Object.entries(missingTranslations).forEach(([importName, infoPerPackage]) => {
+      const info = infoPerPackage[packageKey];
+      if (info == null) {
+        return;
+      }
+      const githubLink = `${SOURCE_CODE_REPO}/blob/master/${info.path}/`;
```
LukasTy, line 311: Maybe we should link to `next` in the next branch?
Or does the script only process master branch translations? 🤔 
Maybe we could just rely on `SOURCE_CODE_ROOT_URL` instead? 🤔

scripts/l10n.ts
```diff
@@ -284,6 +287,63 @@ async function generateReport(missingTranslations: MissingTranslations) {
...
+
+      const languageTag = `${importName.slice(0, 2).toLowerCase()}-${importName
+        .slice(2)
+        .toUpperCase()}`;
+      const localeName = localeNames[languageTag];
+
+      if (localeName === undefined) {
+        throw new Error(
+          [
+            `locale tage ${languageTag} is not associated to a locale name.`,
```
LukasTy, line 321:
```suggestion
            `locale tag ${languageTag} is not associated to a locale name.`,
```

LukasTy (approved):
This is an insanely nice change, which should the community to help with localisation more. 🤞 💯 🚀 

We should set some other progress bar color, because the contrast is really not great. 🙈
And it's even worse with the yellow progress bar... 
Screenshot 2022-11-30 at 14 20 46
Screenshot 2022-11-30 at 14 20 56

```diff
diff --git a/docs/src/modules/components/LocalizationTable.js b/docs/src/modules/components/LocalizationTable.js
index 0121887c7..23de5073c 100644
--- a/docs/src/modules/components/LocalizationTable.js
+++ b/docs/src/modules/components/LocalizationTable.js
@@ -19,6 +19,19 @@ const Root = styled('div')(({ theme }) => ({
   width: '100%',
   height: 26,
   borderRadius: 2,
+  color: (theme.vars || theme).palette.text.primary,
+  '&.low .progress-bar': {
+    backgroundColor: (theme.vars || theme).palette.error.dark,
+  },
+  '&.medium .progress-bar': {
+    backgroundColor: (theme.vars || theme).palette.warning.dark,
+  },
+  '&.high': {
+    color: (theme.vars || theme).palette.success.contrastText,
+    '.progress-bar': {
+      backgroundColor: (theme.vars || theme).palette.success.dark,
+    }
+  },
 }));
 
 const Value = styled('div')({
@@ -31,15 +44,6 @@ const Value = styled('div')({
 
 const Bar = styled('div')({
   height: '100%',
-  '&.low': {
-    backgroundColor: '#f44336',
-  },
-  '&.medium': {
-    backgroundColor: '#efbb5aa3',
-  },
-  '&.high': {
-    backgroundColor: '#088208a3',
-  },
 });
 
 const ProgressBar = React.memo(function ProgressBar(props) {
@@ -47,17 +51,17 @@ const ProgressBar = React.memo(function ProgressBar(props) {
   const valueInPercent = Math.floor((numerator / denumerator) * 100);
 
   return (
-    <Root>
+    <Root className={clsx({
+      low: valueInPercent < 30,
+      medium: valueInPercent >= 30 && valueInPercent <= 70,
+      high: valueInPercent > 70,
+    })}>
       <Value>
         {`${numerator}/${denumerator}`}
         {numerator === denumerator ? '🎉' : ''}
       </Value>
       <Bar
-        className={clsx({
-          low: valueInPercent < 30,
-          medium: valueInPercent >= 30 && valueInPercent <= 70,
-          high: valueInPercent > 70,
-        })}
+        className='progress-bar'
         style={{ maxWidth: `${valueInPercent}%` }}
       />
     </Root>
```

scripts/l10n.ts
alexfauquette: Indeed, I'm using French abbreviations hoping they are the same since "nombre" (FR) is close to "number" (EN)

But of course, they use an "o" 😄

scripts/l10n.ts
LukasTy: Oh, gotcha. Sounds fair. 👌

alexfauquette:
After Lukas proposal

image

image

LukasTy:
> After Lukas proposal

This is an insanely nice iteration over my suggestion. 👍 ❤️

docs/src/modules/components/LocalizationTable.js
```diff
@@ -0,0 +1,153 @@
...
+    },
+  },
+}));
+
+const Value = styled('div')({
+  position: 'absolute',
+  lineHeight: '24px',
+  width: '100%',
+  display: 'flex',
+  justifyContent: 'center',
```
LukasTy, line 49: One last minor nitpick: I think this element could use the following addition. 🤔 

```suggestion
  justifyContent: 'center',
  alignItems: 'center',
```

docs/src/modules/components/LocalizationTable.js
```diff
@@ -0,0 +1,155 @@
+/* eslint-disable material-ui/no-hardcoded-labels */
+import * as React from 'react';
+import PropTypes from 'prop-types';
```
oliviertassinari, line 3: Would it work with the component in TypeScript?

```suggestion
```

oliviertassinari (commented):
Nice 👍

UI wise, I think that the completion progress could be made 1/2 less visible to be balanced. When I open the page, the progress bar is the first thing that my eye is drawn to, but it's not that important.

Maybe?

Screenshot 2022-11-30 at 18 20 30

docs/src/modules/components/LocalizationTable.js
```diff
@@ -0,0 +1,155 @@
...
+        overflow: 'auto',
+        '& td.progress': {
+          padding: 0,
+        },
+        '& p': {
+          margin: 0,
+        },
+      }}
+    >
+      <Table size="small">
```
oliviertassinari, line 99: It feels a bit backward to use both the markdown for table and the table component. Could we pick only one of the two options?

alexfauquette:
> 1/2 less visible

You mean reducing the width?

docs/src/modules/components/LocalizationTable.js
alexfauquette: I tried with TypeScript but docs element were complaining that they were not able to find it. Did not investigated further.

oliviertassinari:
> You mean reducing the width?

@alexfauquette Yes, I was thinking of reducing the surface area to draw less visually attention. 

I was also wondering about using % to make it easier to process the information, but we would indeed lose the information oh how many keys there is missing. So maybe not better.

The only feedback I would have is that it's not easy to make the difference between what's fully translated, and what has a few keys missing. Developers might miss that they could complete the translations. Maybe green should be reserved for 100% not sure.

In any case, it's a great step forward.

alexfauquette:
> I was also wondering about using % to make it easier to process the information

It was my first idea, but I dropped it because it does not include how many elements are to translation. Another solution could be to rename the column "missing keys"

> it's not easy to make the difference between what's fully translated, and what has a few keys missing

Yes, at the beginning I added "🎉" to indicate 100% completion. But it was taking to much space and was distracting.

The proposal of showing only the number of missing translations could also solve this point (easy to see when a digit is not 0)

On the visual aspect I also thaught about the following aproach

image

or either with lighter colors

image

PR #7003, opened 2022-11-25
[DataGridPremium] Pass `groupId` to aggregation function

### LINKED ISSUE #6995: [question]
### Order ID 💳

54867

### Duplicates

### Latest version

### The problem in depth 🔍

There is a Data Grid with grouping and aggregation enabled, and the aggregation is calculated server-side (we do use not simple math operations like sum or count, that is why the calculation is done on the server). The task is to show aggregated values in the Grand Total and in the Group totals (there are multi-level groups). 
For custom aggregation MUI provides `aggregationFunctions` that can be extended with their own implementation, `GridAggregationFunction.apply({values})` method provides the only argument with the row values that need to be aggregated. There is no other information that can help understand the context: `column`, 'group node` etc.  Without it's it impossible to find the correct summary from the server-side calculated values. 
Solution: This info exists on a callee method, it can be passed down to GridAggregationFunction.apply.

Workaround: implementing `renderCell` method gives all the possibility to accomplish the task... but export to Excel uses incorrect numbers then since the correct ones are on the render level.

### Your environment 🌎

<details>

```
  System:
    OS: macOS 13.0
  Binaries:
    Node: 16.13.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.1.2 - /usr/local/bin/npm
  Browsers:
    Chrome: 107.0.5304.12

  npmPackages:
    @emotion/react: ~11.10.0 => 11.10.4 
    @emotion/styled: ~11.10.0 => 11.10.4 
    @mui/base:  5.0.0-alpha.79 
    @mui/icons-material: ~5.8.4 => 5.8.4 
    @mui/lab: 5.0.0-alpha.80 => 5.0.0-alpha.80 
    @mui/material: ~5.6.4 => 5.6.4 
    @mui/private-theming:  5.10.6 
    @mui/styled-engine:  5.10.7 
    @mui/styles: ~5.6.2 => 5.6.2 
    @mui/system:  5.10.7 
    @mui/types:  7.2.0 
    @mui/utils:  5.10.6 
    @mui/x-data-grid: ~5.17.12 => 5.17.12 
    @mui/x-data-grid-premium: ~5.17.12 => 5.17.12 
    @mui/x-data-grid-pro: ~5.17.12 => 5.17.12 
    @mui/x-date-pickers: ~5.0.0-beta.4 => 5.0.3 
    @mui/x-license-pro:  5.17.12 
    @types/react: ~18.0.21 => 18.0.21 
    react: ~17.0.2 => 17.0.2 
    react-dom: ~17.0.2 => 17.0.2 
    typescript: ~4.6.4 => 4.6.4 
```
</details>

m4theushw:
Closes #6995

I said in https://github.com/mui/mui-x/issues/6995#issuecomment-1327554590 that the API methods are not available when the aggregation function is called, but I was wrong. They are available but not via the public `apiRef`. If we pass the same `apiRef` received in https://github.com/mui/mui-x/blob/79e7ce28ef010331a8d32536f2143ee72fe4e83e/packages/grid/x-data-grid-premium/src/hooks/features/aggregation/createAggregationLookup.ts#L22-L28 to the params it works. That's something to investigate.


flaviendelangle (commented):
No problem to add those two keys

Maybe add some JSDoc

flaviendelangle (approved):

Issue #6997, opened 2022-11-25
[pickers] Add codemods for breaking changes


LukasTy:
The v6 for pickers is bringing quite a few breaking changes.
To ease users' migration we ought to try and cover as most changes with codemods as possible.

The codemod package can be [found here](https://github.com/mui/mui-x/tree/next/packages/x-codemod). It already contains all the core setup needed in order to add new codemods to it.

This is intended as an umbrella issue to group all the codemods needed for `v.6.x` migration.
Ideally as much migrations as possible should be provided with `preset-safe` codemod.

## The process of adding a new codemod
- One of the [existing codemods](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0) can be used as an example (i.e.: [localization-provider-rename-locale](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/localization-provider-rename-locale))
- copy or create a new folder for your codemod
- all the relevant code that is eventually shipped to the user is in `index.ts`
- update/provide tests for your newly created codemod (test fixtures are authored as js files to allow as minimal syntax as you want—without triggering ts issues)
- if your codemod is safe to add to [`preset-safe`](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/preset-safe)—update this codemod with your newly added codemod making sure the [codemods order is correct](https://github.com/mui/mui-x/blob/next/packages/x-codemod/src/v6.0.0/preset-safe/index.ts) if codemod is sensitive to that (i.e. separate props should firstly be migrated to `componentProps` and only later be passed through a codemod renaming `componentProps` to `slotProps` if we'd go with such approach)
- update [README](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md) file with your newly added codemod
You can check [@mui/codemod](https://github.com/mui/material-ui/tree/master/packages/mui-codemod/src) for inspiration on more advanced migration cases.

## How to test your changes locally (on a different project)

P.S. `yarn link` does not create a `bin` command from a linked package
- run `build` with either `yarn workspace @mui/x-codemod build` or `yarn build` if you are in `src/packages/x-codemod`
- run `cd build && yarn pack`
- copy the built `7.tgz` file to the root of your project you want to run (test) the codemod in
- add `"@mui/x-codemod": "file:./<your_package_name>"` into your project dependencies
- run `yarn install`
[...]

PR #7027, opened 2022-11-28
[core] Add codemod to move l10n translation

alexfauquette:
Part of #6997


packages/x-codemod/src/v6.0.0/text-props-to-localeText/expected-expression-values.spec.js
```diff
@@ -0,0 +1,271 @@
...
+      secondsClockNumberText: custom_getSecondsClockNumberText,
+      calendarViewSwitchingButtonAriaLabel: custom_getViewSwitchingButtonText,
+      openPreviousView: custom_leftArrowButtonText,
+      previousMonth: custom_leftArrowButtonText,
+      openNextView: custom_rightArrowButtonText,
+      nextMonth: custom_rightArrowButtonText
+    }} />
+
+  {/* Stand alone */}
+  <CalendarPicker
```
flaviendelangle, line 243: I went to fast...
Those two components don't have a `localeText` prop.
We would need to add a `LocalizationProvider` around them.
Or to add support for `localeText`

packages/x-codemod/src/v6.0.0/text-props-to-localeText/index.js
```diff
@@ -0,0 +1,123 @@
...
+  'MobileTimePicker',
+  'StaticDatePicker',
+  'StaticDateRangePicker',
+  'StaticDateTimePicker',
+  'StaticTimePicker',
+  'TimePicker',
+];
+
+/**
+ * @param {import('jscodeshift').FileInfo} file
```
LukasTy, line 69: Did you experience problems authoring a ts file? 🤔

packages/x-codemod/src/v6.0.0/text-props-to-localeText/text-props-to-localeText.test.js
```diff
@@ -0,0 +1,71 @@
...
+import jscodeshift from 'jscodeshift';
+import transform from './index';
+import readFile from '../../util/readFile';
+
+function read(fileName) {
+  return readFile(path.join(__dirname, fileName));
+}
+
+describe('@mui/codemod', () => {
+  describe('v5.0.0', () => {
```
LukasTy, line 12: Could you take a look at how the existing tests are described? 🤔

LukasTy (commented):
Have you looked at "The process of adding a new codemod" described in https://github.com/mui/mui-x/issues/6997

It seems that this codemod should be pretty safe, isn't it?
To follow the process of mui-material I've suggested to also add such codemods to the `preset-safe`, ideally update it's tests and also update the readme. 🤔 
What do you think?

packages/x-codemod/README.md
```diff
@@ -91,4 +92,37 @@ Renames `locale` into `adapterLocale` (or `LocalizationProvider`)
...
+ <DatePicker
+-  cancelText="Cancel"
++  localeText={{
++    cancelButtonLabel: "Cancel"
++  }}
+ />
+```
+
+```sh
+npx @mui/x-codemod v6.0.0/localization-provider-rename-locale <path>
```
LukasTy, line 109:
```suggestion
npx @mui/x-codemod v6.0.0/text-props-to-localeText <path>
```

packages/x-codemod/README.md
```diff
@@ -91,4 +92,37 @@ Renames `locale` into `adapterLocale` (or `LocalizationProvider`)
 npx @mui/x-codemod v6.0.0/localization-provider-rename-locale <path>
 ```
 
+#### `text-props-to-localeText`
+
+Replace props used for translation such as `cancelText` to their corresponding `localeText` key.
```
LukasTy, line 97:
```suggestion
Replace props used for localization such as `cancelText` to their corresponding `localeText` key.
```

packages/x-codemod/README.md
```diff
@@ -91,4 +92,37 @@ Renames `locale` into `adapterLocale` (or `LocalizationProvider`)
...
++    cancelButtonLabel: "Cancel"
++  }}
+ />
+```
+
+```sh
+npx @mui/x-codemod v6.0.0/localization-provider-rename-locale <path>
+```
+
+If you were using always the same text value in all your components, consider moving those translation from the component to the `LocalizationProvider` by hand.
```
LukasTy, line 112:
```suggestion
If you were always using the same text value in all your components, consider moving those translation from the component to the `LocalizationProvider` by hand.
```

packages/x-codemod/README.md
```diff
@@ -91,4 +92,37 @@ Renames `locale` into `adapterLocale` (or `LocalizationProvider`)
...
+```diff
+<LocalizationProvider
+  dateAdapter={AdapterDayjs}
++ localeText={{ cancelButtonLabel: "Cancel" }}
+>
+ <DatePicker
+-  localeText={{ cancelButtonLabel: "Cancel" }}
+ />
+ <DateTimePicker
+-  localeText={{ cancelButtonLabel: "Cancel" }}
```
LukasTy, line 123: Maybe having something different than the default makes a bit more sense in this case? 🤔 
```suggestion
+ localeText={{ cancelButtonLabel: "Cancelar" }}
>
 <DatePicker
-  localeText={{ cancelButtonLabel: "Cancelar" }}
 />
 <DateTimePicker
-  localeText={{ cancelButtonLabel: "Cancelar" }}
```

LukasTy (approved):
Great job! 🙌 💯

flaviendelangle (changes requested):
`ClockPicker` and `CalendarPicker` don't have a `localeText` prop

alexfauquette:
> ClockPicker and CalendarPicker don't have a localeText prop

Let me realize we could have aproblem in terms of order to apply codemod: component renaming

`ClockPicker` is renamed `TimeClock`. So this codemod should be applied before renaming components

LukasTy:
> `ClockPicker` is renamed `TimeClock`. So this codemod should be applied before renaming components

Good point. 👌 
I'd encourage adding this case to `preset-safe.spec` so that we can adjust once the follow-up codemod for renaming is added 😉

packages/x-codemod/tsconfig.json
```diff
@@ -3,5 +3,10 @@
   "compilerOptions": {
     "types": ["react", "mocha", "node"]
   },
-  "include": ["src/**/*", "../../node_modules/@mui/monorepo/test/utils/initMatchers.ts"]
+  "include": [
+    "src/**/*",
+    "../../node_modules/@mui/monorepo/test/utils/initMatchers.ts",
+    "../../node_modules/@mui/material/themeCssVarsAugmentation",
+    "../../node_modules/date-fns"
```
alexfauquette, line 10: I don't know why the typescript on `@mui/x-codemod` complains that `Locale` is not defined in `useMaskedInput` whereas typescript does not complain on `@mui/x-date-icker` itself

So I added is in the tsconfig, but I assume their is a smarter wai of doing that
```
    "../../node_modules/date-fns"
```

LukasTy: Hmm... Have you tried adding `"skipLibCheck": true,`? That's one thing that is different between these projects. 🤔

alexfauquette: Yes, but does not solve the problem

flaviendelangle (approved):
Great job !

PR #7038, opened 2022-11-29
[pickers] Remove the callback version of the `action` prop in the `actionBar` component slot

flaviendelangle:
### Changelog

- Remove the callback version of the `action` prop on the `actionBar` slot (#7038) @flaviendelangle 

  The `action` prop of the `actionBar` slot is no longer supporting a callback.
  Instead, you can pass a callback at the slot level

  ```diff
   <DatePicker
     componentsProps={{
  -     actionBar: {
  -       actions: (variant) => (variant === 'desktop' ? [] : ['clear']),
  -     },
  +     actionBar: ({ wrapperVariant }) => ({
  +       actions: wrapperVariant === 'desktop' ? [] : ['clear'],
  +     }),
     }}
   />
  ```


packages/x-date-pickers/src/PickersActionBar/PickersActionBar.test.tsx
```diff
@@ -28,24 +28,6 @@ describe('<PickersActionBar />', () => {
     expect(screen.queryByRole('button')).to.equal(null);
   });
 
-  it('should not render buttons if actions function returns an empty array', () => {
```
flaviendelangle, line 31: It is now `useSlotProps` that is responsible for having the same behavior with the callback and the object formats.

packages/x-date-pickers/src/internals/components/PickersModalDialog.tsx
```diff
@@ -87,6 +88,19 @@ export function PickersModalDialog(props: React.PropsWithChildren<PickersModalDi
   } = props;
 
   const ActionBar = components?.ActionBar ?? PickersActionBar;
+  const actionBarProps = useSlotProps({
```
flaviendelangle, line 91: I am also doing the change on the legacy pickers because it was more complicated to support the typing different than actually developing the feature :laughing:

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
```diff
@@ -337,6 +337,19 @@ export function PickersPopper(inProps: PickerPopperProps) {
...
+    elementType: ActionBar,
+    externalSlotProps: componentsProps?.actionBar,
+    additionalProps: {
+      onAccept,
+      onClear,
+      onCancel,
+      onSetToday,
+      actions: [],
+    },
+    ownerState: { wrapperVariant: 'desktop' },
```
flaviendelangle, line 350: On the legacy picker, I'm only passing the variant to the ownerState, that's really a tiny detail since those components are going away

packages/x-date-pickers/src/internals/hooks/useStaticPicker/useStaticPicker.tsx
```diff
@@ -48,7 +48,10 @@ export const useStaticPicker = <
         <PickerStaticViewLayout
           {...layoutProps}
           components={components}
-          componentsProps={componentsProps}
+          componentsProps={{
+            ...componentsProps,
+            actionBar: undefined,
+          }}
```
alexfauquette, line 54: I don't get why you need to override the `actionBar` props with `undefined`.

alexfauquette (commented):
I updated PR description to add the breaking change

packages/x-date-pickers/src/internals/hooks/useStaticPicker/useStaticPicker.tsx
flaviendelangle: Otherwise TS complains that the types of the props of the action bar used by `PickerStaticViewLayout` and `PickersViewLayout` are note the same.

And since the action bar is not handled by `PickerStaticViewLayout` on the new picker, I'm passing `undefined` to bypass the problem.
I could also `@ts-ignore` but I did not want to risk forgetting to remove it once we drop the `actionBar` from `PickerStaticViewLayout` altogether.

Same for the mobile / desktop wrappers.

flaviendelangle: Oh I went too fast... this `actionBar: undefined` only applies on mobile and desktop pickers, on static we don't have any "legacy" wrapper..

Thanks for the catch !

alexfauquette (approved):

PR #7047, opened 2022-11-30
[docs] Use new pickers on the validation page

flaviendelangle:
[Doc preview](https://deploy-preview-7047--material-ui-x.netlify.app/x/react-date-pickers/validation/)

- [x] Use new pickers in the demos
- [x] Improve the note at the top of the page to say that the validation also works on the legacy pickers
- [x] Use field and clock on the `Invalid values feedback` section

### DIFF
--- a/docs/data/date-pickers/validation/DateRangeValidationShouldDisableDate.js
+++ b/docs/data/date-pickers/validation/DateRangeValidationShouldDisableDate.js
@@ -1,10 +1,8 @@
 import * as React from 'react';
 import dayjs from 'dayjs';
-import TextField from '@mui/material/TextField';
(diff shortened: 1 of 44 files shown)


alexfauquette (approved):
Sounds good.

Small note about demos in [Date Time validation](https://deploy-preview-7047--material-ui-x.netlify.app/x/react-date-pickers/validation/#date-and-time-validation) section. They only have one element, but use the `GridItem`. Maybe the label can be removed to simplify the demo

flaviendelangle:
> Maybe the label can be removed to simplify the demo

I kept it to have the exact same UI everywhere, so that a user can look at the demo without even reading the code an know it's a date time picker.

PR #7057, opened 2022-12-01
[docs] Create a `DemoGrid` component to unify demos with several components

flaviendelangle:
## Goals

- [x] Never have components overflowing the demo container even on smaller screen (should always move from `direction="row"` to `"direction="column"` before it does)

- [x] Adapt the breakpoints and the spacing automatically based on the type of component inside the grid (a calendar needs more space that a single input field for instance). The RegExp to categorize the children can evolve if we find broken use cases.

I'm doing everything automatically instead of passing props to the grid because we have almost 100 demos with several pickers / fields / calendars in our documentation. So maintaining a human-written set of responsive rules on each demos feels very complication and not very future proof.

## Usage

1. If all the components have a `label` prop, you can just pass those pickers are child to the grid:

```tsx
<PickersGrid>
  <DatePicker label="Date Picker" />
  <TimePicker label="Time Picker" />
</PickersGrid>
```

2. If some components don't have a `label` prop (a calendar, a clock, a range field, a range picker), you can pass `PickersGridItem` to the grid with the label on it:

```tsx
<PickersGrid>
  <PickersGridItem label="Date Picker">
    <DatePicker />
  </PickersGridItem>
  <PickersGridItem label="Date Range Picker">
    <DateRangePicker />
  </PickersGridItem>
</PickersGrid>
```

We could prevent `1.` and always use `PickersGridItem` but I think this shortcut is good since it makes shorter demo previews.

## Review

I propose to review the internals before migrating all the other demos.
Do not trust only the Argos diffs, since they are done on a fixed width and most of the advantages of the new grid is to have a good responsive behavior.

Here is the list of the modified demos:

- **Date Picker - Uncontrolled vs. Controlled** ([before](https://next.mui.com/x/react-date-pickers/date-picker/#uncontrolled-vs-controlled), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/date-picker/#uncontrolled-vs-controlled))

- **Date Field - Customize the date format** ([before](https://next.mui.com/x/react-date-pickers/date-field/#customize-the-date-format), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/date-field/#customize-the-date-format))

- **Date Time Picker - Uncontrolled vs. Controlled** ([before](https://next.mui.com/x/react-date-pickers/date-time-picker/#uncontrolled-vs-controlled), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/date-time-picker/#uncontrolled-vs-controlled))

- **Date Range Field - Basic usage** ([before](https://next.mui.com/x/react-date-pickers/date-range-field/#basic-usage), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/date-range-field/#basic-usage))

- **Date Range Field - Uncontrolled vs. Controlled** ([before](https://next.mui.com/x/react-date-pickers/date-range-field/#uncontrolled-vs-controlled), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/date-range-field/#uncontrolled-vs-controlled))

- **Validation - Minimum and maximum date** (`maxDate` demo) ([before](https://next.mui.com/x/react-date-pickers/validation/#minimum-and-maximum-date), [after](https://deploy-preview-7057--material-ui-x.netlify.app/x/react-date-pickers/validation/#minimum-and-maximum-date))


packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
```diff
@@ -17,7 +17,9 @@ const MultiInputDateRangeFieldRoot = styled(
     slot: 'Root',
     overridesResolver: (props, styles) => styles.root,
   },
-)({});
+)({
+  alignItems: 'baseline',
```
flaviendelangle, line 21: Not related, the v5 pickers had this alignment and I find it slightly better

tsconfig.json
```diff
@@ -23,7 +23,9 @@
       "@mui/x-date-pickers-pro/*": ["./packages/x-date-pickers-pro/src/*"],
       "@mui/x-license-pro": ["./packages/x-license-pro/src"],
       "@mui/x-license-pro/*": ["./packages/x-license-pro/src/*"],
-      "test/*": ["./test/*"]
+      "test/*": ["./test/*"],
+      "docs/*": ["./node_modules/@mui/monorepo/docs"],
+      "docsx/*": ["./docs/*"]
```
flaviendelangle, line 28: Removes a warning on my IDE, it was not finding the components like `PickersGrid` because the `docsx` alias did not exist in TypeScript, only in `babel`

alexfauquette (approved):
The internal logic sounds good. Maybe you can rename `PickersGrid` by `DemoGrid` or ` DemoPickersGrid` (same for `PickersGridItem`)

Such that the reader immediately get that it's for demonstration purpose and not a component to consider

flaviendelangle:
I would avoid `DemoPickersGrid` because this component is often displayed in demo previews and I try to keep those as short as possible.
But on the other hand, `DemoGrid` can be confusing if we think about the data grid.
How about `DemoLayout` ? Or `DemoContainer` ?

alexfauquette:
`DemoContainer` sounds better. It avoids conflicts with the `Layout` slot

flaviendelangle:
`DemoContainer` / `DemoItem` then

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
LukasTy: Was it an oversight to apply this change only to this multi input field?
Also, I should probably be done via the `stack` prop not to have overridden style rules. 🤔 
Screenshot 2022-12-16 at 15 59 12

flaviendelangle: You are probably right on both points yes

LukasTy: Will address it with Multi Input fields PR. 😉

PR #7066, opened 2022-12-02
[pickers] Improve performances of `DayCalendar`

### LINKED ISSUE #7065: [DateRangePicker] Improve rendering performance
Created from https://github.com/mui/mui-x/pull/6763#issuecomment-1326162341

## Problem

Performance of our Date Range Calendar was not great initially, especially on mobile, but after implementing date range dragging—the issue has exacerbated even more.
We are doing way too many computations for each of the day slot and also re-rendering every day even though it might not be necessary (i.e. it was not selected, nor de-selected).

## Solution

Investigate possible solutions to reduce the amount of computations needed. Solutions to investigate:
- memoization
- pre-computing date ranges (dates) and only checking if the day is one of the pre-computed days (instead of checking if **this** day falls within a given range)

flaviendelangle:
Fixes #7065

- [x] No prop of a given day should change across renders on `DateCalendar` when we select another day
- [x] No prop of a given day should change across renders on `DateCalendar` when we select another day (a few exceptions are explained below)
- [x] Add tests
- [x] Benchmark more precisely the gains

These performance improvements are a trade off, to achieve them, I had to remove the `selectedDays` prop that was by definition causing a re-render of all the days whenever the value change. I updated the doc demo to show how users can do without it.

## Benchmarks

All the benchmarks below have been realized without any custom props `<DateCalendar />` / `<DateRangeCalendar />`, on my desktop computer with a 6x CPU throttling.

### Date Calendar - click on a date (no prev value)

Before: 347m (2 render: 183ms + 164ms)
After: 128ms (2 render: 67ms + 61ms)

### Date Range Calendar - hover a date

Before: 494ms
After: 105ms 

Do we really need this render ? We have none on the `DateCalendar` and their is not preview to render anyway, we should be able to skip it.

### Date Range Calendar - click on a start date (no prev value)

Before: 904ms (2 render: 460ms + 444ms)
After: 239ms (2 render: 130ms + 109ms)

### Date Range Calendar - start dragging

Before: 429ms
After: 121ms

### Date Range Calendar - move dragging from a date to the next one

Before: 837ms (409ms + 418ms)
After: 104ms


flaviendelangle:
> Even before the last commit the date range selection and dragging had near flawless performance on my mobile.

The last commit is mostly desktop oriented
I did some desktop tests with x6 throttling, and the render when start dragging went from 278ms to about 60ms, it's worth testing more but I think the additional complexity is worth it.
For the bugs, I'll dive deeper, I have a few failing tests and I need to find where I made mistakes.

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
```diff
@@ -95,87 +98,136 @@ const useDragRangeEvents = <TDate>({
...
+    setIsDragging(true);
+    const button = event.target as HTMLButtonElement;
+    const buttonDataset = button.dataset;
+    if (buttonDataset.position) {
+      onDatePositionChange(buttonDataset.position as DateRangePosition);
     }
   });
+
   const handleDragEnter = useEventCallback((event: React.DragEvent<HTMLButtonElement>) => {
+    if (!shouldInitDragging) {
```
LukasTy, line 155: Were these needed? 🤔 
If I'm not mistaken—this event shouldn't ever be triggered if a drag has not been started yet 🤔

docs/data/date-pickers/date-picker/CustomDay.tsx
```diff
@@ -63,12 +63,20 @@ function Day(props: PickersDayProps<Dayjs>) {
...
       <StaticNextDatePicker
         displayStaticWrapperAs="desktop"
-        defaultValue={dayjs('2022-04-07')}
+        value={value}
+        onChange={(newValue) => setValue(newValue)}
         components={{ Day }}
+        componentsProps={{
+          day: {
+            selectedDay: value,
+          } as any,
```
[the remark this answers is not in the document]
flaviendelangle: To do avoid the case, users would have to use the slot props override interfaces, 
But inside our doc I don't know how it would work exactly. 
I will create the override interface and extend it in this doc examples to see how it behaves.

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
flaviendelangle: OK the change was more complicated than expected.
The logic I propose is to do the `isElementDraggable` on start events (`onDragStart` and `onTouchStart`), and the other one to just check if `isDragging` is true (to avoid applying logic if the drag event comes from outside of the pickers.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
```diff
@@ -449,7 +449,7 @@ describe('<DateRangeCalendar />', () => {
         }
         const initialValue: [any, any] = [
           adapterToUse.date(new Date(2018, 0, 1)),
-          adapterToUse.date(new Date(2018, 0, 7)),
+          adapterToUse.date(new Date(2018, 0, 2)),
```
LukasTy, line 452: Was this needed? And if yes, why only for touch (mobile) dragging? 🤔

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
```diff
@@ -95,87 +100,138 @@ const useDragRangeEvents = <TDate>({
...
+    setIsDragging(true);
+    const button = event.target as HTMLButtonElement;
+    const buttonDataset = button.dataset;
+    if (buttonDataset.position) {
+      onDatePositionChange(buttonDataset.position as DateRangePosition);
     }
   });
+
   const handleDragEnter = useEventCallback((event: React.DragEvent<HTMLButtonElement>) => {
+    if (!isDragging) {
```
LukasTy, line 158: I have specifically avoided using `isDragging` in these functions to avoid having the need to re-create them whenever that flag changes. 🤔 

But if you think it's the only way to achieve performance gains we need—it probably doesn't matter that much as compared to the reduction in computations. 🙈 👍

LukasTy (commented):
Was it your intention to change how touch drag handles dropping outside of days?
Now the last highlighted/selected day is accepted wherever drop ends.

By the way: We still have the issue with `cursor: grabbing` not being applied to dragging day.

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
flaviendelangle: This test is weird
In a real browser, when we have the `touchStart` event, shouldn't it be on the element we try to drag ?
In this test, we click on `1` but the 1st target is `2`.

So I changed to try to make it pass but I feel like it's either

-  `executeDateTouchDragWithoutEnd` that should execute  `touchStart` with the coordinates of`target` as the touch target (which is what we do in every other test)
- `handTouchStart` that should use `event.target` instead of `resolveElementFromTouch`

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
LukasTy: That should be a good enough solution and shouldn't make much difference in performance. The biggest issues were in other places, but I initially tried at least avoiding these methods recreation.
But now that the root performance issues are no more—we can go with as safe a solution as we need. 👌

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
LukasTy: > In a real browser, when we have the `touchStart` event, shouldn't it be on the element we try to drag ? In this test, we click on `1` but the 1st target is `2`.

Yes—it should and it is, the element we start the drag on is selected with `screen.getByRole('gridcell', { name: '1' })`. Touch events are always fired on the same/origin target, just with different coordinates, which co-respond to specific dates.

packages/x-date-pickers-pro/src/DateRangeCalendar/useDragRange.ts
LukasTy: I think that this question is [answered here](https://github.com/mui/mui-x/pull/7066#discussion_r1039251650) 👌

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
flaviendelangle: OK so do you agree with the fix I did on the test ?

LukasTy: I still don't think it was needed 🤔 
Essentially—it doesn't change anything.
We still start the drag on Jan 1st and assert that the validation has successfully been applied to the flipped day when we touch drag over Jan 9th.

LukasTy: Sorry, we ended up in a misunderstanding because of mis-alignment of what the original question was on and what change you have landed on.
After looking at the current change—yes, I agree with it 100%. It was my mistake in the first place. 🙈

flaviendelangle: I misunderstood my own response when reading it this morning :laughing: 
OK so we can update the text :+1:

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
```diff
@@ -570,4 +572,40 @@ describe('<DateRangeCalendar />', () => {
 
     expect(screen.getAllByMuiTest('pickers-calendar')).to.have.length(3);
   });
+
+  describe('Performances', () => {
```
LukasTy, line 576: It's interesting putting rendering performance tests here, but oh well, it might help us stop and think more when doing future changes. 💪 
Nitpick, I don't think that such plural form of `Performances` exist. Maybe stick to just `Performance`?

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -76,10 +80,27 @@ const DateRangeCalendarArrowSwitcher = styled(PickersArrowSwitcher)({
...
-const DayCalendarForRange = styled(DayCalendar)({
+const DayCalendarForRange = styled(DayCalendar)(({ theme }) => ({
   minWidth: 312,
   minHeight: weeksContainerHeight,
-}) as typeof DayCalendar;
+  [`&.${dateRangeCalendarClasses.dayDragging}`]: {
+    [`& .${dayClasses.dayOutsideRangeInterval}`]: {
+      '@media (pointer: fine)': {
+        '&:hover': {
+          border: `1px solid ${(theme.vars || theme).palette.grey[500]}`,
```
LukasTy, line 90: I'm pretty sure that this style has been incorrectly migrated—it introduces a regression of not providing a grey border around the currently hovered day.
It should be applied inversely—in case the `dayDragging` is not active to avoid styles fighting beinging highlighting and hover. 😉

LukasTy (changes requested):

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.test.tsx
flaviendelangle: I am not a huge fan of those tests
But I don't know how to make better tests without spending a ton of time on them
I think testing the amount of renders is fine here (even though most of the time we prefer to test the render time),
because the regression we want to avoid is the "render all days".

But clearly, they could be improved.

LukasTy: Yeah, by no means—they are a way better precaution of re-introducing previous performance issues than nothing. 👍 
And if we get the time or think of better way to implement those—that can be done in future.

flaviendelangle: Named fixed

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
flaviendelangle: The selector is wrong indeed, it is applied when dragging but it should be the opposite
Migrating from ownerState to CSS classes in a mental effort I don't like :laughing:

LukasTy (approved):
Great job! 👏 
This is an immense improvement, which should definitely be felt by end-users. 🚀

PR #5133, opened 2022-06-07
[DataGrid] Add `columnGroupingModel`

alexfauquette:
Fix #195

- [x] Add `columnGroupingModel` to manage the tree of column groups
- [x] Compute header height according to the max depth of the column groups tree
- [x] Support column resize
- [x] Render the column group
- [x] Improve accessibility with `role="columnheader"` and correct column spanning
- [x] Support virtualization engine when only some columns of the group are rendered
- [x] Limit column drag&drop to the same column group
- [x] Add doc page
- [ ] Add test
	- [x] convert model to component
	- [ ] throw error for inconsistency in the model (struggling)
	- [x] reorder behavior
- [x] Mark it as an experimental feature

docs preview: https://deploy-preview-5133--material-ui-x.netlify.app/x/react-data-grid/column-groups/

with default style: https://codesandbox.io/s/basicgroupingdemo-demo-mui-x-forked-6vhr5j?file=/demo.tsx
with show right border style (it seems buggy when the total width is smaller than the container width): https://codesandbox.io/s/basicgroupingdemo-demo-mui-x-forked-7si9wc?file=/demo.tsx
performance demo: https://codesandbox.io/s/basicgroupingdemo-demo-mui-x-forked-fj4qcs?file=/demo.tsx

## Changelog

You can now group columns by using the `columnGroupingModel` prop. This allows you to display more structured data.

image

To enable this feature, add `experimentalFeatures={{ columnGrouping: true }}`.
The grouping header can be fully customized.
See the [documentation](https://mui.com/x/react-data-grid/column-groups/) to explore everything it has to offer.

PR #7067, opened 2022-12-02
[DataGrid] Remove column separator to match table styles

### LINKED ISSUE #1623: [DataGrid] Match <Table> styles, hide column separator

## Current Behavior 😯

A column header separator is visible.

## Expected Behavior 🤔

No column header separator.

## Steps to Reproduce 🕹

Steps:

2. 

Screenshot 2021-05-11 at 11 20 58

## Context 🔦

It doesn't match https://mui.com/material-ui/react-table/ nor https://material.io/components/data-tables. This was spotted looking into adding coverage for the data grid in the Figma kit.

Regarding the affordance of the column resize feature, I would encourage us to show a separator on hover, so it's clear for the end-users that they can resize with `DataGridPro` but can't in `DataGrid`, it's misleading otherwise (or maybe it was done to encourage developers to upgrade to the pro plan 🙈).

## Your Environment 🌎

v4.0.0-alpha.28

MBilalShafi:
Fixes #1623

Removed column separator, it will only be shown when there's resize support available.

## Changelog

- [x] Remove column separator
- [x] Remove unused `disableHeaderSeparator`

## Design
- [Figma Process](https://www.figma.com/file/W6TItYkiGJZjL5zLnfeSjy/Data-grid?node-id=1227%3A62449&t=MAO4kO2qvUssZFTG-0) 
- Figma Final

### DIFF
--- a/packages/grid/x-data-grid/src/components/columnHeaders/GridGenericColumnHeaderItem.tsx
+++ b/packages/grid/x-data-grid/src/components/columnHeaders/GridGenericColumnHeaderItem.tsx
@@ -39,7 +39,6 @@ interface GridGenericColumnHeaderItemProps
   label: string;
   draggableContainerProps?: Partial<React.HTMLProps<HTMLDivElement>>;
   columnHeaderSeparatorProps?: Partial<GridColumnHeaderSeparatorProps>;
-  disableHeaderSeparator?: boolean;
 }
 
 const GridGenericColumnHeaderItem = React.forwardRef(function GridGenericColumnHeaderItem(
@@ -69,7 +68,6 @@ const GridGenericColumnHeaderItem = React.forwardRef(function GridGenericColumnH
     resizable,
     draggableContainerProps,
     columnHeaderSeparatorProps,
-    disableHeaderSeparator,
     ...other
   } = props;
 
@@ -135,15 +133,13 @@ const GridGenericColumnHeaderItem = React.forwardRef(function GridGenericColumnH
         </div>
         {columnMenuIconButton}
       </div>
-      {!disableHeaderSeparator && (
-        <GridColumnHeaderSeparator
-          resizable={!rootProps.disableColumnResize && !!resizable}
-          resizing={isResizing}
-          height={height}
-          side={separatorSide}
-          {...columnHeaderSeparatorProps}
-        />
-      )}
+      <GridColumnHeaderSeparator
+        resizable={!rootProps.disableColumnResize && !!resizable}
+        resizing={isResizing}
+        height={height}
+        side={separatorSide}
+        {...columnHeaderSeparatorProps}
+      />
       {columnMenu}
     </div>
   );
--- a/packages/grid/x-data-grid/src/components/columnHeaders/GridColumnGroupHeader.tsx
+++ b/packages/grid/x-data-grid/src/components/columnHeaders/GridColumnGroupHeader.tsx
@@ -191,7 +191,6 @@ function GridColumnGroupHeader(props: GridColumnGroupHeaderProps) {
       aria-colspan={fields.length}
       // The fields are wrapped between |-...-| to avoid confusion between fields "id" and "id2" when using selector data-fields~=
       data-fields={`|-${fields.join('-|-')}-|`}
-      disableHeaderSeparator
       {...mouseEventsHandlers}
     />
   );
--- a/packages/grid/x-data-grid/src/components/containers/GridRootStyles.ts
+++ b/packages/grid/x-data-grid/src/components/containers/GridRootStyles.ts
@@ -217,6 +217,7 @@ export const GridRootStyles = styled('div', {
       backgroundColor: (theme.vars || theme).palette.action.hover,
     },
     [`& .${gridClasses.columnSeparator}`]: {
+      visibility: 'hidden',
       position: 'absolute',
       zIndex: 100,
       display: 'flex',
@@ -224,6 +225,11 @@ export const GridRootStyles = styled('div', {
       justifyContent: 'center',
       color: borderColor,
     },
+    [`& .${gridClasses.columnHeaders}:hover .${gridClasses.columnSeparator}, .${gridClasses['columnSeparator--resizing']}`]:
+      {
+        visibility: 'visible',
+        width: 'auto',
+      },
     [`& .${gridClasses['columnSeparator--sideLeft']}`]: {
       left: -12,
     },


packages/grid/x-data-grid/src/components/columnHeaders/GridGenericColumnHeaderItem.tsx
```diff
@@ -39,7 +39,6 @@ interface GridGenericColumnHeaderItemProps
   label: string;
   draggableContainerProps?: Partial<React.HTMLProps<HTMLDivElement>>;
   columnHeaderSeparatorProps?: Partial<GridColumnHeaderSeparatorProps>;
-  disableHeaderSeparator?: boolean;
```
MBilalShafi, line 42: @mui/xgrid Team, I couldn't find this prop being referenced or used anywhere, is it already deprecated and is waiting to be removed or there is something else to it?

cherniavskii: It was added in https://github.com/mui/mui-x/pull/5133 for column grouping. It's used internally since the component is not exported:
https://github.com/mui/mui-x/blob/d8b76d392976d8c1c9eeeb37d13a6d2368751f9a/packages/grid/x-data-grid/src/components/columnHeaders/GridColumnGroupHeader.tsx#L193

cherniavskii (approved):
Looks good to me!

joserodolfofreitas (commented):
Last Friday, I think I saw the separator appear on hover, but now the separator seems to be gone entirely. Is that intentional?

I think the ideal, would be to keep the separators on hover and display a separator on both sides of the header. Would that be possible? (Initially, the separator is only on the right border).

cherniavskii:
@joserodolfofreitas the separator appears on hover, but only if the column is resizable. Do you think showing separators for non-resizable columns would be useful? It won't be possible to resize the column anyway.

joserodolfofreitas:
> the separator appears on hover, but only if the column is resizable.
Do you think showing separators for non-resizable columns would be useful?

Ah ok. In this case, I think It's fine to show the separator only when it's resizable.

joserodolfofreitas:
> I think It's fine to show the separator only when it's resizable.

Independently of that, I think we should explore showing the separator on both sides when we do show them.

MBilalShafi:
> Maybe the menu icon should be rendered immediately after the sort/filter icons.

Thanks for raising this pain point, yes I also feel this could cause a bit of confusion for the user. Rendering it immediately after sort/filter could be a solution, but should this behavior be the same when resize support is available? Not sure how will it look like then.

MBilalShafi:
This is how it looks when everything is enabled:

image

MBilalShafi:
> I think the ideal, would be to keep the separators on hover and display a separator on both sides of the header.

That's a good idea but IMO the user might have difficulty understanding the boundaries of columns especially when numeric and non-numeric columns are mixed up.

Another possibility is to show all separators on hover, I feel it looks better as well as it doesn't show separators when not needed.

Here are both the variants compared:
With one separator: https://codesandbox.io/s/column-separators-variant-1-ujn042?file=/demo.tsx
With all spearators: https://codesandbox.io/s/column-separators-variant-2-ly0i1k?file=/demo.tsx

CC @gerdadesign

cherniavskii:
@MBilalShafi Thanks for the examples!
I really like the demo with all separators on hover!
I've noticed a few things related to the menu icon position change:
1. The menu icon button is cut:
Screenshot 2022-12-13 at 11 16 23
2. Menu icon jump looks a bit odd:

https://user-images.githubusercontent.com/13808724/207292610-2e0bf53f-5173-4fba-8dfd-d97715007fab.mov

3. In MIT DataGrid it's sometimes unclear which column the menu button belongs to.
In the example below, one might think the menu button would open the column menu for the Trader Email column:
Screenshot 2022-12-13 at 11 17 26

Maybe we should consider showing all separators on hover on MIT DataGrid as well while keeping the original position of the menu icon?

gerdadesign:
+1 for displaying all separators on hover. 
I also see how it can be confusing to know which column the actions are attached to on hover. 

If we stuck with keeping the icons where they are now and showing all separators on hover, we might be able to use the separator type to distinguish between resizeable and not resizable. Perhaps we could think about using a vertical divider that matches the row divider in style for not resizing and adding the small separator as a "handle"? visual illustration below


<a href="https://www.figma.com/file/W6TItYkiGJZjL5zLnfeSjy/Data-grid?node-id=1147%3A63394&t=r7JT5Gs1vAc0ZZFU-1"></a>

MBilalShafi:
> Perhaps we could think about using a vertical divider that matches the row divider in style for not resizing and adding

Using a vertical divider is a good idea and it looks great too
image

The only issue is, it doesn't fit great with the toolbar layout
image

Here's a live version: https://codesandbox.io/s/column-separators-variant-3-u2o9w5?file=/demo.tsx

Though adding a border on top may make some sense, I feel it ends up being too cluttered.
image

Maybe showing the disabled resize handle on the MIT data grid too could be the best way forward?

gerdadesign:
>Here's a live version: https://codesandbox.io/s/column-separators-variant-3-u2o9w5?file=/demo.tsx

I like this direction! It does feel incomplete with the toolbar, but without a top border. Is it possible to add that to the hover style of the header row? May feel cluttered if it's on all the time, but if only on hover, it's temporary and provides a clear separation.

joserodolfofreitas:
I like the explorations going on here, but I think it's worth doing on its [issue](https://github.com/mui/mui-x/issues/7268) as it proposes to solve a different problem related to column resizing UI/UX.

Overall I think the solution we reached to show all separators on hover feels like a good output for this issue.

joserodolfofreitas (approved):

oliviertassinari:
Great to see movement on this, it feels like a step forward.

---

On https://deploy-preview-7067--material-ui-x.netlify.app/x/react-data-grid/. I think that hiding the separator when the column isn't resizable would be another step forward. Personally, when I see:

Screenshot 2022-12-27 at 01 17 30

I interpret the separator as: I can resize this column. https://github.com/mui/mui-x/pull/7067#issuecomment-1350200327 might feel even better.

cherniavskii:
@oliviertassinari Showing separators on hover when the column isn't resizable solves (3) in https://github.com/mui/mui-x/pull/7067#issuecomment-1348147798. Without separators, it's pretty confusing.

oliviertassinari:
@cherniavskii I think that the experience with the column menu in (3) would be less confusing than the experience on HEAD when it comes to resizing, but it's only based on one user (me) that tried the change with CSS and didn't experience point (3).

In any cases, assuming we move forward with something in the direction of https://github.com/mui/mui-x/pull/7067#issuecomment-1350200327: no separator by default, a drag and drop handle on hover, a subtle separator between columns on hover, then end-users might have a great UX. Is there a follow up planned?

MBilalShafi:
> Is there a follow up planned?

Yes, we'll try to address it with this [issue](https://github.com/mui/mui-x/issues/7268)

PR #7097, opened 2022-12-05
[pickers] Rework view internals

flaviendelangle:
## In this PR

- [x] Rename `setOpenView` => `setView` in `useViews` (and in any component passing this prop down we name it `onViewChange`) 
- [x] Rename `openView` => `view` in `useViews` (and in any component passing this prop down)
- [x] Move `useFocusManagement` inside `useViews` and avoid having `setFocusedView` return a function (we were not memoizing anything anyway)
- [x] Rename `handleChangeAndOpenNext` => `setValueAndGoToNextView` in `useViews` (avoid mixing `set`, `handle` and `on` for callbacks returned by a hook and make clear what we are changing and opening)
- [x] Rename `openNext` => `goToNextView` (make clear that it is the view we are changing)
- [x] Use the params from `useViews` to create the props of every component using it (was already the case for the new pickers, did it for `DateCalendar` which causes some bug on the old picker because of type inconsistencies. For instance I had to redefine the `sx` prop on non-static pickers. This change helps us migrate away for `ExportedDateCalendarProps = Omit<DateCalendarProps, ...>` which is a very error-prone pattern.

## Questions

- Should we rename `openTo` => `defaultView` ? On static pickers and calendars, this name make no sense, and naming is `defaultView` would make very clear that it behaves like a `value` / `defaultValue` pattern.

=> We are keeping `openTo` for now

- Should we apply a smarter fallback to `openTo` at the component level (`useDatePickerDefaultizedProps` for instance) in case the current fallback is not in the views (for the date picker, it would be something like  `views.includes('day') ? 'day' : views[0]`. Because right now if we apply a custom set of views that do not contain `day`, we have to manually define `openTo`, otherwise it breaks.

=> Yes we should apply smart defaults


docs/pages/x/api/date-pickers/date-calendar.json
```diff
@@ -14,19 +15,25 @@
...
+    "onFocusedViewChange": { "type": { "name": "func" } },
     "onMonthChange": { "type": { "name": "func" } },
     "onViewChange": { "type": { "name": "func" } },
     "onYearChange": { "type": { "name": "func" } },
     "openTo": {
       "type": {
         "name": "enum",
         "description": "'day'<br>&#124;&nbsp;'month'<br>&#124;&nbsp;'year'"
-      },
-      "default": "'day'"
```
flaviendelangle, line 29: We already lost this precision on the new pickers due to the shared interface.

It would be nice to be able to do something like:

```tsx
/**
 * @propDefaultValue view "day"
**/
interface MyComponentProps extends SomeShareInterface {}
```

Not sure about the exact synthax
This would allow to define doc default values without redefining the type itself.

If you know about any existing convention for this use case I'm very interested

docs/pages/x/api/date-pickers/date-range-picker.json
```diff
@@ -55,7 +55,6 @@
     "onError": { "type": { "name": "func" } },
     "onMonthChange": { "type": { "name": "func" } },
     "onOpen": { "type": { "name": "func" } },
-    "onViewChange": { "type": { "name": "func" } },
```
flaviendelangle, line 58: This prop is never used (and drop on the new pickers)
I can add it back to the typing manually but I don't think it's worth it

Same on other legacy range pickers

docs/pages/x/api/date-pickers/date-calendar.json
```diff
@@ -1,5 +1,6 @@
 {
   "props": {
+    "autoFocus": { "type": { "name": "bool" } },
```
flaviendelangle, line 3: `DateCalendar` already had this prop
No idea why the doc gen was not getting it

Same for `DateRangeCalendar` and all the pickers that use them

docs/pages/x/api/date-pickers/date-calendar.json
```diff
@@ -14,19 +15,25 @@
     "disablePast": { "type": { "name": "bool" } },
     "displayWeekNumber": { "type": { "name": "bool" } },
     "fixedWeekNumber": { "type": { "name": "number" }, "default": "undefined" },
+    "focusedView": {
```
flaviendelangle, line 18: `DateCalendar` already had this prop
No idea why the doc gen was not getting it
Same for `onFocusedViewChange`

docs/pages/x/api/date-pickers/next-date-picker.json
```diff
@@ -81,6 +82,12 @@
       }
     },
     "value": { "type": { "name": "any" } },
+    "view": {
```
flaviendelangle, line 85: I have no idea how I missed it, but all the new pickers were missing the `view` prop

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.tsx
```diff
@@ -496,7 +496,7 @@ const DateRangeCalendar = React.forwardRef(function DateRangeCalendar<TDate>(
           {calendars === 1 ? (
             <PickersCalendarHeader
               views={['day']}
-              openView={'day'}
+              view={'day'}
```
flaviendelangle, line 499: `PickersCalendarHeader` is exposed via the theme
But the typing only contains `classes` so I would not count this one as a breaking change

packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.types.ts
```diff
@@ -44,45 +45,12 @@ export interface DateRangeCalendarSlotsComponentsProps<TDate>
   >;
 }
 
-export interface DateRangeCalendarProps<TDate>
+export interface ExportedDateRangeCalendarProps<TDate>
```
flaviendelangle, line 48: The logic before:

```tsx
interface DateRangeCalendarProps = { ... }

interface ExportedDateRangeCalendarProps extends Omit<DateRangeCalendarProps, ...> {}
```

The logic after:

```tsx
interface ExportedDateRangeCalendarProps { ... }

interface DateRangeCalendarProps extends ExportedDateRangeCalendarProps { ... }
```

packages/x-date-pickers/src/DatePicker/shared.ts
```diff
@@ -36,7 +36,7 @@ export interface BaseDatePickerSlotsComponentsProps<TDate>
 }
 
 export interface BaseDatePickerProps<TDate>
-  extends ExportedDateCalendarProps<TDate>,
+  extends Omit<ExportedDateCalendarProps<TDate>, 'value' | 'onChange' | 'defaultValue'>,
```
flaviendelangle, line 39: The calendars have the same `value` / `onChange` / `defaultValue` as the new pickers
But for the legacy pickers, I have to omit here and re-define to avoid type conflicts

packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
```diff
@@ -774,14 +774,6 @@ describe('<DesktopDatePicker />', () => {
     });
   });
 
-  it('should throw console warning when invalid `openTo` prop is provided', () => {
```
flaviendelangle, line 777: That one is intersting?

The `warnOnlyOnce` variables are not resetting between tests.
So the 2nd test failed.

I removed it since it's a legacy pickers.
But I don't know how we could cleanly test those warnings.
Maybe just export the variables from the `internals` and reset them to `false` after the test.

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
```diff
@@ -28,26 +27,13 @@ export type PickerViewRendererLookup<
 /**
  * Props used to handle the views that are common to all pickers.
  */
-export interface UsePickerViewsBaseProps<TView extends DateOrTimeView> {
+export interface UsePickerViewsBaseProps<TView extends DateOrTimeView>
+  extends Omit<UseViewsOptions<any, TView>, 'onChange' | 'onFocusedViewChange' | 'focusedView'> {
```
flaviendelangle, line 31: I did not wanted to add controllable focused view on the picker on this PR.
But that would be very easy to do if we wanted to

docs/pages/x/api/date-pickers/date-calendar.json
alexfauquette: This specific use case does not seem to be supported by jsdocs.

You have the list of "supported tags" [here](https://jsdoc.app/index.html).

We could add a plugin to support an additional tag `@overrideDefault`. But it sounds to be an interesting project, but will require some time to navigate int jsdocs documentation

packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
```diff
@@ -61,7 +61,7 @@ export interface MonthCalendarProps<TDate>
   disableHighlightToday?: boolean;
   onMonthFocus?: (month: number) => void;
   hasFocus?: boolean;
-  onFocusedViewChange?: (newHasFocus: boolean) => void;
+  onFocusedViewChange?: (hasFocus: boolean) => void;
```
alexfauquette, line 64: With the "new" it sounds more obvious that the function has to be call when the focus is modified. Not a strong opinion

alexfauquette (approved):
Nice that focus management get moved to the `useView`

packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
flaviendelangle: My main rational here was that on `onChange` we are always calling it `value` not `newValue` in the interface

(example from the `base` package)
```tsx
onChange?: (event: React.SyntheticEvent, value: number | string) => void;
```

packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.test.tsx
alexfauquette: We could remove the `warnOnce` logic in test env

```js
let warnedOnce = false;
if (!warnedOnce || process.env.NODE_ENV === 'test') {
  console.warn(...)
}
```

flaviendelangle: I'm afraid it would be logged several times and would make the test crash even with `toWarnDev` (which would only catch the first one)
To be tested

Issue #7096, opened 2022-12-05
[docs] Module `@mui/x-date-pickers` has no exported member `bgBG`


johnflux:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

I'm doing:

I have: "@mui/x-date-pickers": "5.0.9" installed.

`import { bgBG } from '@mui/x-date-pickers';`

But I get: 

`Module "@mui/x-date-pickers" has no exported member 'bgBG'`

### Your environment 🌎

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

johnflux:
OH!

The docs are very confusing!

They say to do:

`import bgLocale from 'date-fns/locale/bg';`

But if you do, you'll get Bulgarian!   If you want english you actually have to do:

`import enUSLocale from 'date-fns/locale/en-US';
     <LocalizationProvider
          dateAdapter={AdapterDateFns}
          adapterLocale={enUSLocale} // use locale for date parser/formatter
        >
`

And instead of `bgBG` you have to do:

`import { enUS } from '@mui/x-date-pickers';
export const theme = createTheme({...}, enUS)
`

LukasTy:
Hello @johnflux 
Thank you for opening this issue.
Do I understand correctly that the doc suggesting to use `bg` locale got you confused, because `x-date-pickers` does not actually have this locale yet? Or was it something else?
We will address this inconsistency. 👌

flaviendelangle:
Closing this issue, the linked PR resolves the problem :+1:

johnflux:
But all you did was change it from Bulgarian to German!

Can you please state in the docs that that is for German (just saying de isn't enough. I thought that was the variable name).  And state to replace de with the language that you want. 

And can you show how to use the default American English locale please. I couldn't figure out how and the docs don't say how to do it at all. Typescript makes the parameter required, is that correct?

LukasTy:
@johnflux 
Are we in some sort of misunderstanding?
![Screenshot 2022-12-07 at 17 17 50](https://user-images.githubusercontent.com/4941090/206218063-1485eb6c-7394-4e04-a195-b32b7cf3636b.png)
It's stated in the docs that if you are fine with English (US) locale—you don't need to make any changes.

All the following documentation is for those, who need a different locale.
[...]

PR #7102, opened 2022-12-06
[docs] Update localization doc to use existing locale

LukasTy:
Part of #7096 

Update documentation to use a locale, which exists among all the products (data grid, material ui, pickers).
I've chosen `deDE` as we have already used it in other demos.
Feel free to suggest a better alternative. 🙈 

[Preview](https://deploy-preview-7102--material-ui-x.netlify.app/x/react-date-pickers/localization/)

### DIFF
--- a/docs/data/date-pickers/localization/localization.md
+++ b/docs/data/date-pickers/localization/localization.md
@@ -30,9 +30,9 @@ you just have to import the locale from `@mui/x-date-pikers` (see the [list of s
 
 ```jsx
 import { createTheme, ThemeProvider } from '@mui/material/styles';
-import { DateCalendar, LocalizationProvider, bgBG } from '@mui/x-date-pickers';
+import { DateCalendar, LocalizationProvider, deDE } from '@mui/x-date-pickers';
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
-import bgLocale from 'date-fns/locale/bg';
+import 'date-fns/locale/de';
 
 const theme = createTheme(
   {
@@ -40,7 +40,7 @@ const theme = createTheme(
       primary: { main: '#1976d2' },
     },
   },
-  bgBG, // use 'bg' locale for UI texts (start, next month, ...)
+  deDE, // use 'de' locale for UI texts (start, next month, ...)
 );
 
 function App({ children }) {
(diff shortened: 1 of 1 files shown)


docs/data/date-pickers/localization/localization.md
```diff
@@ -108,7 +108,9 @@ This method can be combined with the ones shown above.
 If you pass a locale through `LocalizationProvider` or the theme, and you provide translation keys through the `localeText` prop of a picker at the same time, then only the latter translation keys will be overridden.
 
 ```tsx
-<LocalizationProvider localeText={frFR}>
+<LocalizationProvider
+  localeText={frFR.components.MuiLocalizationProvider.defaultProps.localeText}
```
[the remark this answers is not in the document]
flaviendelangle: It depends if we want to document how to pass a locale from the theme or pass a custom locale they wrote (often a partial locale then).

To be honest, I don't really understand why a user would not want to use the theme to pass our locales.
So for me the main use case for the `localeText` prop of `LocalizationProvider` is to pass some custom locales to a sub-set of the app.

But I don't know how users actually use it.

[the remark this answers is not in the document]
LukasTy: I'd reckon that the least common case is for users to have custom locales.
IMHO—we should stick with existing implementation documentation in regular pages and only opt for providing custom implementation suggestions in recipes or similar sections.
Like you just mentioned—I'd say that providing custom localisation could be a good candidate for a "recipe". 💡

alexfauquette, line 112: Since the purpose is about priority between the different way of customizing, we could pass an object and after the code example explaining what will be the result:

```suggestion
  localeText={{ clearButtonLabel: "Reset", todayButtonLabel: "Now" }}
```

- today button with text "Now" taken from the Localization Provider's localeText prop
- clear button with text "Vider" overridden by the Date Picker's localeText prop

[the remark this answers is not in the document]
flaviendelangle: Custom translations can be things as simple as custom `start` and `end` label on a range picker (`Check in` / `Check out`) for example.

LukasTy: Thank you. That's a very good suggestion. 👌

LukasTy: [Updated](https://github.com/mui/mui-x/pull/7102/commits/7585ec149ebc19a1a47851245eb4c9d9f51672e7) as per your suggestion. 😉

docs/data/date-pickers/localization/localization.md
```diff
@@ -30,9 +30,7 @@ you just have to import the locale from `@mui/x-date-pikers` (see the [list of s
 
 ```jsx
 import { createTheme, ThemeProvider } from '@mui/material/styles';
-import { DateCalendar, LocalizationProvider, deDE } from '@mui/x-date-pickers';
-import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
-import 'date-fns/locale/de';
+import { deDE } from '@mui/x-date-pickers';
```
LukasTy, line 33: @flaviendelangle Should we go this route on this doc?
It had all the previous imports left after refactoring simplifying examples.
I've removed every import that is not used in the code example.

docs/data/date-pickers/localization/localization.md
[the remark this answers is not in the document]
LukasTy: Handled [with](https://github.com/mui/mui-x/pull/7102/commits/7585ec149ebc19a1a47851245eb4c9d9f51672e7) as per Alex's suggestion. 😉

alexfauquette (approved):
Cleaner docs 👌

docs/data/date-pickers/localization/localization.md
flaviendelangle: That's fine for me

flaviendelangle (approved):

PR #7105, opened 2022-12-06
[docs] Add warning alert for components requiring `WrapperVariantContext`

### LINKED ISSUE #7064: [docs] Document the need for `WrapperVariantContext`
### Problem

Certain pickers components expect `React.useContext(WrapperVariantContext)` to exist in the parent scope.
Some of these components might end up being used standalone and would not behave as expected if the mentioned context is not provided.

### Suggestion
Add a warning/info section (preferably in API docs page) mentioning that this component expects `WrapperVariantContext` to be provided if used standalone for components having this dependency.

LukasTy:
Fixes #7064 

Add a warning alert in component api page to components using `React.useContext(WrapperVariantContext)`.

[Example](https://deploy-preview-7105--material-ui-x.netlify.app/x/api/date-pickers/date-range-calendar/)

### DIFF
--- a/docs/pages/x/api/date-pickers/desktop-time-picker.json
+++ b/docs/pages/x/api/date-pickers/desktop-time-picker.json
@@ -95,5 +95,6 @@
   "packages": [
     { "packageName": "@mui/x-date-pickers-pro", "componentName": "DesktopTimePicker" },
     { "packageName": "@mui/x-date-pickers", "componentName": "DesktopTimePicker" }
-  ]
+  ],
+  "requiresWrapperVariantContext": false
 }
--- a/docs/src/modules/components/ApiPage.js
+++ b/docs/src/modules/components/ApiPage.js
@@ -353,16 +353,17 @@ export default function ApiPage(props) {
             />
           </React.Fragment>
         ) : null}
-        {requiresWrapperVariantContext && (<div className="MuiCallout-root MuiCallout-warning">
-          <p
-            dangerouslySetInnerHTML={{
-              __html:
-                `This component expects <code>WrapperVariantContext</code> to exist above in the React component tree.
+        {requiresWrapperVariantContext && (
+          <div className="MuiCallout-root MuiCallout-warning">
+            <p
+              dangerouslySetInnerHTML={{
+                __html: `This component expects <code>WrapperVariantContext</code> to exist above in the React component tree.
                 If this component is used standalone and the context is not provided—the component might have unexpected behavior.<br>
                 Make sure that the component is used inside any of the Date and Time Picker components or provide the context yourself.`,
-            }}
-          />
-        </div>)}
+              }}
+            />
+          </div>
+        )}
         {componentStyles.name && (
           <React.Fragment>
             <Heading hash="component-name" />
--- a/docs/src/modules/components/ApiPage.js
+++ b/docs/src/modules/components/ApiPage.js
@@ -230,6 +230,7 @@ export default function ApiPage(props) {
     slots,
     styles: componentStyles,
     packages,
+    requiresWrapperVariantContext,
   } = pageContent;
 
   const {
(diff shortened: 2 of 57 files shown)


docs/pages/x/api/date-pickers/desktop-time-picker.json
```diff
@@ -95,5 +95,6 @@
   "packages": [
     { "packageName": "@mui/x-date-pickers-pro", "componentName": "DesktopTimePicker" },
     { "packageName": "@mui/x-date-pickers", "componentName": "DesktopTimePicker" }
-  ]
+  ],
+  "requiresWrapperVariantContext": false
```
flaviendelangle, line 99: We could skip the property when its value is `false`

docs/src/modules/components/ApiPage.js
```diff
@@ -352,6 +353,17 @@ export default function ApiPage(props) {
             />
           </React.Fragment>
         ) : null}
+        {requiresWrapperVariantContext && (
+          <div className="MuiCallout-root MuiCallout-warning">
```
flaviendelangle, line 357: But the context is private right ?
Shouldn't we add a prop to those components like on static pickers instead of having people use the context ?

flaviendelangle: Not sure to understand what you mean

flaviendelangle: > And we do expect users to import and use it if they need it.

OK for me that's a major problem.
It should at least be exported from the root if we intend people to use it.

____

For `WrapperVariant` specifically, I think that it should remain a private component people never touch.
I feel like, if I was a user and I had to do `<WrapperVariant value="mobile"><DateCalendar /></WrapperVariant>`, then I would consider  `DateCalendar` not to be a good standalone component.

We have a comment in the `WrapperVariant` component saying to consider removing it, and I think it should be the end goal.
If we have a look at why `WrapperVariant` is being used:

- In `PickersMonth` we set a custom class with the variant => I think it could be removed, it was probably added for consistency with `PickersYear` (maybe by myself when I unified)

- In `DateRangeCalendar` we enable previewing only on desktop => for `StaticDateRangePicker` I think the previewing should be enabled as well and `displayStaticWrapperAs="desktop"` should only impact the layouting parts, not this kind of behavior. I would be in favor of adding a media query actually matching when the previewing can work or not.

- In `DateTimePickerTabs` we pass it to the `ownerState` and apply a box shadow on desktop => it could be a prop passed to the slot by the pickers

- In `YearCalendar` we render 3 years per row on desktop and 4 on mobile => I think it could be a prop like we have `calendars` on `DateRangeCalendar`, with a default of `3` on `MobileXXXPicker`, `4` on `DesktopXXXPicker` and based on `displayStaticWrapperAs` on `StaticXXXPicker`

- In `PickersYear`  we pass it to the `ownerState` to respect the prop set in `YearCalendar` => `YearCalendar` could just pass a prop `itemsPerRow`

- `Clock` it is used to apply a default value to `ampmInClock` => We could apply the default on the pickers instead of relying on this prop. And when using `TimeClock` alone, I don't think people expect this value to change based on the variant anyway, that's way to magical.


___

Basically, replace the `wrapperVariant` following these rules

- If it applies a default value, then set the default value in the picker instead (like we have for `calendars` for instance) and the calendar / clock is used as standalone, don't put any magical default behavior

- If it applies a style on a private sub-component, add a prop that actually tells which style to apply based on the decision made by the parent (i.e: the amount of years per row)

LukasTy: 🙌 Those are really great points, which we should try applying to identify the feasibility of dropping this context.
One thing I'm vary of is if we'll need some other feature difference between desktop and mobile. At that point we'd need to introduce yet another prop. 🤔 🙈 

I'm just not a huge fan of this suggestion:
> I would be in favor of adding a media query actually matching when the previewing can work or not.

I did explore it in dragging PR and was not a fan of the outcome. Maybe just keeping the default behaviour while `wrapperVariant` is unclear is a decent choice, given we'd introduce a prop, which can be passed in if needed. 🤔

flaviendelangle: > I'm just not a huge fan of this suggestion:

It could also be a `disablePreviewing` prop, set by the pickers (to match the current `wrapperVariant` behavior) and let people choose for standalone usage.

> One thing I'm vary of is if we'll need some other feature difference between desktop and mobile. At that point we'd need to introduce yet another prop.

For me it make sense, on a standalone usage to have several props if the behaviors are not tightly linked together.
People using `DateCalendar` can want 3 years by row on the year view and have previewing (for instance if they are rendering the component in a small container).
And that right now is not possible.

LukasTy:
Super-seeded by https://github.com/mui/mui-x/issues/7372

PR #6619, opened 2022-10-24
[DataGrid] Improved column menu design and api

MBilalShafi:
Fixes #7025
Fixes #4929 

Preview: https://deploy-preview-6619--material-ui-x.netlify.app/x/react-data-grid/column-menu/#column-menu-with-pro-premium-options

## Changelog

- New column menu design with improved look and feel
- Ability to add new menu items and place them on any order
- Replaced all the items to be injected via pipe processor for better control through respective hooks
- Added ability to override/reorder column menu items which may help add more items in any order e.g. for https://github.com/mui/mui-x/issues/5307
- Ability to customize the menu by passing a component in column menu slot
- New l10n labels added
- Added dedicated component and icon slots for Pro and Premium packages for shipping only required components per package
- Fix accessibility bug of focus shifting to body when an item gets hidden

### Breaking changes

- `column` and `currentColumn` is now renamed to `colDef`
- `DATA_GRID_DEFAULT_SLOTS_COMPONENTS` export has been removed

- Some of the components have been renamed for consistency.

  **Community Package:**
  ```diff
  - <GridFilterMenuItem />
  + <GridColumnMenuFilterItem />
  ```

  ```diff
  - <HideGridColMenuItem />
  + <GridColumnMenuHideItem />
  ```

  ```diff
  - <GridColumnsMenuItem />
  + <GridColumnMenuColumnsItem />
  ```

  ```diff
  - <SortGridMenuItems />
  + <GridColumnMenuSortItem />
  ```

  ```diff
  - interface GridFilterItemProps
  + interface GridColumnMenuItemProps
  ```
  
  **Pro package:**
  ```diff
  - <GridColumnPinningMenuItems />
  + <GridColumnMenuPinningItem />
  ```
  
  **Premium package:**
  ```diff
  - <GridAggregationColumnMenuItem />
  + <GridColumnMenuAggregationItem />
  ```
  
  ```diff
  - <GridRowGroupingColumnMenuItems />
  - <GridRowGroupableColumnMenuItems />
  + <GridColumnMenuGroupingItem />
  ```

## Design 
- [Figma Process](https://www.figma.com/file/W6TItYkiGJZjL5zLnfeSjy/Data-grid?node-id=679%3A38262&t=aFUzX2phxDdIDwFL-0)
- [Final](https://www.figma.com/proto/W6TItYkiGJZjL5zLnfeSjy/Data-grid?node-id=1239%3A64017&scaling=min-zoom&page-id=679%3A38262)

PR #7108, opened 2022-12-06
[DataGrid] Render SortGridMenuItems with sortingOrder props

hanbin9775:
Fixes: #7103 

SortGridMenuItems were being rendered regardless of the sortingOrder props.


MBilalShafi:
@hanbin9775 Can you set the target branch as `master`? For `next` it'll be fixed along with https://github.com/mui/mui-x/pull/6619

PR #7119, opened 2022-12-07
[core] Remove internals-fields imports

flaviendelangle:
I created this endpoint to avoid importing `date-fns` on `@mui/x-date-pickers`, but not that the field use the adapters, we don't need this endpoint anymore.


alexfauquette (approved):
Effectively, it's not needed anymore 👍

PR #7120, opened 2022-12-07
[pickers] Initialize date without time when selecting year or month

### LINKED ISSUE #7114: [pickers] Time taken from the current date instead of being zeroed
### Duplicates

### Latest version

### Steps to reproduce 🕹

Link to live example:

Steps:
1. Enter a random invalid date into the DesktopPickerField (possibly other ones, I have tested on this) - ie. 89/89
2. Click calendar icon
3. Change year
4. Select data
5. OnChange callback returns date with time set to the current time

### Current behavior 😯

As per the steps above, the date returned has time set to the current time in the local timezone.

I have found this [issue](https://github.com/mui/mui-x/issues/4449) which has been fixed a few months ago, but it seems some cases still slipped through the cracks.

### Expected behavior 🤔

The date returned will be set to the midnight (zero time).

### Your environment 🌎

<details>

```
   System:
    OS: Windows 10 10.0.19044
  Binaries:
    Node: 18.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 3.2.4 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.18.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 108.0.5359.95
    Edge: Spartan (44.19041.1266.0), Chromium (108.0.1462.42)
  npmPackages:
    @emotion/react: ^11.10.5 => 11.10.5
    @emotion/styled: ^11.10.5 => 11.10.5
    @mui/base:  5.0.0-alpha.109
    @mui/core-downloads-tracker:  5.10.17
    @mui/icons-material: ^5.10.16 => 5.10.16
    @mui/lab: ^5.0.0-alpha.111 => 5.0.0-alpha.111
    @mui/material: ^5.10.17 => 5.10.17
    @mui/private-theming:  5.10.16
    @mui/styled-engine:  5.10.16
    @mui/styles: ^5.10.16 => 5.10.16
    @mui/system:  5.10.17
    @mui/types:  7.2.2
    @mui/utils:  5.10.16
    @mui/x-data-grid:  5.17.14
    @mui/x-data-grid-premium: ^5.17.14 => 5.17.14
    @mui/x-data-grid-pro:  5.17.14
    @mui/x-date-pickers:  5.0.9
    @mui/x-date-pickers-pro: ^5.0.9 => 5.0.9
    @mui/x-license-pro:  5.17.12
    @types/react: 18.0.26 => 18.0.26
    react: ^18.2.0 => 18.2.0
    react-dom: ^18.2.0 => 18.2.0
```
</details>

### Order ID 💳 (optional)

ZD362901

LukasTy:
Fixes #7114 

Instead of current approach, when `now` is used and has a side-effect of initialising date with current time, change to:
- Fix to init date to `startOfYear` when selecting a year
- Fix to init date to `startOfMonth` when selecting a month

### DIFF
--- a/packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
+++ b/packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
@@ -159,6 +159,7 @@ export const MonthCalendar = React.forwardRef(function MonthCalendar<TDate>(
     () => value ?? utils.startOfMonth(now),
     [now, utils, value],
   );
+
   const selectedMonth = React.useMemo(() => {
     if (value != null) {
       return utils.getMonth(value);
--- a/packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
+++ b/packages/x-date-pickers/src/MonthCalendar/MonthCalendar.tsx
@@ -155,7 +155,10 @@ export const MonthCalendar = React.forwardRef(function MonthCalendar<TDate>(
 
   const todayMonth = React.useMemo(() => utils.getMonth(now), [utils, now]);
 
-  const selectedDateOrToday = value ?? now;
+  const selectedDateOrStartOfMonth = React.useMemo(
+    () => value ?? utils.startOfMonth(now),
+    [now, utils, value],
+  );
   const selectedMonth = React.useMemo(() => {
     if (value != null) {
       return utils.getMonth(value);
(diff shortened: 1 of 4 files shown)


LukasTy:
@joserodolfofreitas I would be curious to hear your opinion on this.
Would you agree that going this route and not only `startOfDay` for both cases is the more correct approach?

flaviendelangle (approved):
I am always a bit worried by this type of change because we had several regressions in the past caused by similar fallback changes.

But I don't find any actual regression this one could introduce, finger crossed :crossed_fingers:

joserodolfofreitas:
> Would you agree that going this route and not only startOfDay for both cases is the more correct approach?

100%. I can't think of a better alternative for initializing the date-time value.

PR #7133, opened 2022-12-08
[core] Add notes to remove the legacy pickers internals

flaviendelangle:
Various small fixes and unification of the `TODO` to help us find all of them when we will drop the legacy pickers


packages/x-date-pickers/src/internals/hooks/useStaticPicker/useStaticPicker.types.ts
```diff
@@ -6,18 +6,11 @@ import {
...
 import { UsePickerParams } from '../usePicker';
-import {
-  PickersSlotsComponent,
-  PickersSlotsComponentsProps,
-} from '../../components/wrappers/WrapperProps';
 
-export interface UseStaticPickerSlotsComponent
-  extends ExportedPickersViewLayoutSlotsComponent,
-    Pick<PickersSlotsComponent, 'PaperContent'> {}
+export interface UseStaticPickerSlotsComponent extends ExportedPickersViewLayoutSlotsComponent {}
```
flaviendelangle, line 10: The `PaperContent` slot was never used on the new pickers
The goal of this slot was to be able to add custom content (like the shortcuts) in the layout. 
This will be handled by #6900 on the new pickers

alexfauquette (approved):

PR #7135, opened 2022-12-08
[DateRangePicker] Fix to propagate `disabled` and `readOnly` on multi input picker

### LINKED ISSUE #7122: [DateRangePickers] `disabled` is not applied on new component
### Duplicates

### Latest version

### Steps to reproduce 🕹

Visible on the documentation page

https://next.mui.com/x/react-date-pickers/date-range-picker/#form-props

image

### Current behavior 😯

_No response_

### Expected behavior 🤔

_No response_

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #7122 

- Fix `next mobile pickers` fields to be readonly—disabling data editing via input (same as legacy mobile pickers)
- Introduce `date-range` describe test coverage
- Fix `disabled` and `readOnly` propagation on field components (including Single input)
- Fix to propagate `inputMode` to input props (instead of root `TextField`) in multi input fields
- Fix `readOnly` to stop mobile pickers from opening dialog with <kbd>Space</kbd> or <kbd>Enter</kbd>

P.S. Argos change is expected and detects the fix.


packages/x-date-pickers-pro/src/tests/describeRangeValidation/testTextFieldRangeValidation.tsx
```diff
@@ -27,7 +27,7 @@ export const testTextFieldRangeValidation: DescribeRangeValidationTestSuite = (
 ) => {
   const { componentFamily, render, isSingleInput, withDate, withTime } = getOptions();
 
-  if (!['legacy-picker', 'picker'].includes(componentFamily)) {
+  if (!['legacy-picker', 'new-picker'].includes(componentFamily)) {
```
LukasTy, line 30: This seems to have been an oversight typo, which prevented quite a lot of tests from running 🙈

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
```diff
@@ -40,8 +41,13 @@ export const testControlledUnControlled: DescribeValueTestSuite<any, any> = (
       assertRenderedValue(newValue);
       // TODO: Clean this exception or change the clock behavior
       expect(onChange.callCount).to.equal(componentFamily === 'clock' ? 2 : 1);
-      // TODO: Support range
-      expect(onChange.lastCall.args[0]).toEqualDateTime(newValue as any);
+      if (Array.isArray(newValue)) {
```
LukasTy, line 44: `type` is not available here, hence I've used another `date-range` differentiator. 🤔

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
```diff
@@ -78,12 +79,20 @@ const MultiInputDateRangeField = React.forwardRef(function MultiInputDateRangeFi
     elementType: Input,
     externalSlotProps: componentsProps?.input,
     ownerState: { ...ownerState, position: 'start' },
+    additionalProps: {
+      disabled,
+      readOnly,
+    },
```
LukasTy, line 85: These props need to be propagated to the individual `start` and `end` inputs.

packages/x-date-pickers-pro/src/internal/hooks/useMobileRangePicker/useMobileRangePicker.tsx
```diff
@@ -87,7 +87,7 @@ export const useMobileRangePicker = <
     externalSlotProps: componentsProps.field,
     additionalProps: {
       ...pickerFieldProps,
-      readOnly,
+      readOnly: readOnly ?? true,
```
LukasTy, line 90: Fixes mobile range picker field editing (should not be possible—readOnly)

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
```diff
@@ -82,13 +83,11 @@ export interface UseFieldForwardedProps {
...
+  onPaste?: React.ClipboardEventHandler;
+  inputProps?: Partial<InputBaseProps['inputProps']>;
 }
 
-export type UseFieldResponse<TForwardedProps extends UseFieldForwardedProps> = Omit<
-  TForwardedProps,
-  keyof UseFieldForwardedProps
-> &
-  Required<UseFieldForwardedProps> &
+export type UseFieldResponse = Required<UseFieldForwardedProps> &
```
LukasTy, line 90: This is the most interesting change.
Could you take a look at it @flaviendelangle ?
I've tried to untangle logic in here, but I couldn't understand  the reason why the previous type def was needed. It started erroring out for me when I introduced `inputProps` in `UseFieldForwardedProps `

LukasTy: [This](https://github.com/mui/mui-x/pull/7135/files#diff-968c508c9db1c5aa01a8484bd478e639c7ea15f455614de20f612a50cf7b8a84R91) seems to have solved the typing issue. The reason why it did help is still unclear to me when computing the resulting types in my brain 🙈 🤷

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
alexfauquette: Why `type`  is not available? It should be returned by the `getOptions()`

But `Array.isArray()` has the advantage of not being sensible to typos 🤭

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
```diff
@@ -51,14 +64,43 @@ export const testControlledUnControlled: DescribeValueTestSuite<any, any> = (
...
+
+    ['readOnly', 'disabled'].forEach((prop) => {
+      it(`should apply ${prop}="true" prop`, () => {
+        if (!['field', 'new-picker'].includes(componentFamily)) {
+          return;
+        }
+        const handleChange = spy();
+        render(<ElementToTest value={values[0]} onChange={handleChange} {...{ [prop]: true }} />);
+
+        // assert that editing in "readOnly" mode is not allowed
```
alexfauquette, line 90: I assume the reason to add such a test is that event with the `readOnly` props it's feasible to edit dates with arrowUp/arrowDown

If that's the reason it would be nice to explain it in the comment such that a new dev in 1 year can understand why this test is here

packages/x-date-pickers-pro/src/NextDateRangePicker/describes.NextDateRangePicker.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }));
+
+  describeRangeValidation(NextDateRangePicker, () => ({
+    render,
+    clock,
+    componentFamily: 'new-picker',
+    views: ['day'],
+  }));
+
+  describeValue(NextDateRangePicker, () => ({
```
flaviendelangle, line 44: In my PR, I only added the `describeValue` on `MobileXXX` and `DesktopXXX`, not on the responsive one

I think it's worth adding it at least to those two.
And for the responsive one I don't think it's worth it 

Did you have an issue writing the mobile version of the range `describeValue` ?

packages/x-date-pickers/src/tests/describeValue/testPickerOpenCloseLifeCycle.tsx
```diff
@@ -245,11 +283,21 @@ export const testPickerOpenCloseLifeCycle: DescribeValueTestSuite<any, 'new-pick
       const newValue = setNewValue(values[0], { isOpened: true });
 
       // Dismiss the picker
-      userEvent.mousePress(document.body);
+      if (pickerParams.type === 'date-range') {
+        // for range picker a mouse press does not seem to trigger onClose (at least in "jsdom" env)
```
flaviendelangle, line 287: If pressing `Escape` works for all pickers, I'm fine just replacing.
The way we close the picker is an implementation detail here anyway.

If no solution works for all, then I'm fine with this `if`

alexfauquette (commented):
Did some tests, sounds to work well event with native <input/> 👌

But propagation should also be added to:
- MultiInputDateTimeRangeField
- MultiInputTimeRangeField

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
LukasTy: > Why `type` is not available? It should be returned by the `getOptions()`

Because these tests are used in all pickers and we can't safely provide `new-pickers` type from this test, because it might be used in legacy pickers.

Maybe it would be worth experimenting type propagation from the consuming `describes.xxx.test.tsx` file 🤔 


> But `Array.isArray()` has the advantage of not being sensible to typos 🤭

Indeed, we have that benefit... 🙈 😆

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
LukasTy: I was debating adding it in, but the currently existing issue pushed me towards adding it in. 🤔 

This case would probably be caught by the missing `readonly` attribute and might be redundant... 🤔 
Would you agree that removing it is for the best?

packages/x-date-pickers/src/tests/describeValue/testPickerOpenCloseLifeCycle.tsx
LukasTy: The issue is that we have two separate tests, one asserting that `Esc` click closes and the other asserting that click outside closes pickers. 🤔 
Identifying why range picker does not seem to close in test suite might be worthwhile. 🤔

flaviendelangle: Oh sorry I did not check what the test was actually doing
If the range pickers does not work with click outside, I would bail out with an early return and add a todo to check why later.

flaviendelangle: I struggled on similar problems in the past :grimacing:

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
```diff
@@ -87,7 +87,7 @@ export const useDesktopRangePicker = <
   };
 
   const fieldSlotsProps = useRangePickerInputProps({
-    wrapperVariant: 'mobile',
+    wrapperVariant: 'desktop',
```
LukasTy, line 90: This was a fun oversight to debug/track 🙈 😆

packages/x-date-pickers-pro/src/internal/hooks/useMobileRangePicker/useMobileRangePicker.tsx
```diff
@@ -88,7 +88,7 @@ export const useMobileRangePicker = <
     externalSlotProps: componentsProps.field,
     additionalProps: {
       ...pickerFieldProps,
-      readOnly,
+      readOnly: fieldSlotsProps.startInput.readOnly,
```
LukasTy, line 91: If you have better suggestions of how to not loose the props from being correctly propagated—I am open to suggestions.
The issue is that without this the `readOnly` flag is lost, while it is actually being [correctly initialized in here](https://github.com/mui/mui-x/pull/7135/files#diff-63d527e5271e05c7345f4c6abb8068bd30729006acdf288c8c66390b08b7aef3R94).

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -513,5 +513,6 @@ export const useField = <
     onMouseUp: handleInputMouseUp,
     error: inputError,
     ref: handleRef,
+    inputProps: { ...otherForwardedProps.inputProps, readOnly, disabled },
```
flaviendelangle, line 516: Not sure to understand why we need to forward `disabled` through `inputProps`

`<TextField disabled />` is prefered over `<TextField inputProps={{ disabled }} />` so that the `FormControl` is also aware of it.

Am I missing something ?

___

For the `readOnly` I agree that the change is needed if we want the input to be aware of this prop

packages/x-date-pickers/src/tests/describeValue/testPickerOpenCloseLifeCycle.tsx
```diff
@@ -187,14 +215,19 @@ export const testPickerOpenCloseLifeCycle: DescribeValueTestSuite<any, 'new-pick
...
+        expect(onChange.lastCall.args[0]).toEqualDateTime(newValue as any);
+      }
       expect(onClose.callCount).to.equal(1);
     });
 
     it('should call onClose when clicking outside of the picker without prior change', () => {
-      // TODO: Fix this test and enable it on mobile
-      if (pickerParams.variant === 'mobile') {
+      // TODO: Fix this test and enable it on mobile and date-range
+      if (pickerParams.variant === 'mobile' || pickerParams.type === 'date-range') {
```
LukasTy, line 230: As discussed with @flaviendelangle —it's time consuming and not trivial to track down why clicking on body in `jsdom` does not correctly close date range picker 🤷

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
LukasTy: The behaviour of disabled field and disabled field + input is different and I think a bit more correct.
Here is a video comparing the two. Let me know what you think 🤔 

https://user-images.githubusercontent.com/4941090/207817947-40854523-77d5-4a1b-89e6-05f966086f97.mov

flaviendelangle: > The behaviour of disabled field and disabled field + input is different and I think a bit more correct.

When you are talking about "field", are you talking about `DateField` or `TextField` ?
I do agree that we need to forward the props to the `TextField`, I just don't understand why it should be through its `inputProps` prop.

LukasTy: `field` -> `DateField`.
I'll try looking at the issue more closely and identify, why `disabled` attribute is not applied to `input`, once `TextField` is disabled. 😉

packages/x-date-pickers-pro/src/internal/hooks/useMobileRangePicker/useMobileRangePicker.tsx
LukasTy: Resolved it in a possibly [more correct way](https://github.com/mui/mui-x/pull/7135/files#r1049367534) using the result of your snippet.

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.types.ts
```diff
@@ -24,7 +24,7 @@ export type UseMultiInputDateRangeFieldComponentProps<TDate, TChildProps extends
   UseMultiInputDateRangeFieldProps<TDate>;
 
 export interface MultiInputDateRangeFieldProps<TDate>
-  extends UseMultiInputDateRangeFieldComponentProps<TDate, {}> {
+  extends UseMultiInputDateRangeFieldComponentProps<TDate, TextFieldProps> {
```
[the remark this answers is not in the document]
LukasTy: I have only aligned it with other `MultiInput**RangeFieldProps` types. This is the only one that was missing this type.

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
LukasTy: Tested it and it looks like that this initial try & fail solution is no longer relevant after fixing the correct prop propagation chain. 💪 
Reverted. 😉

LukasTy: Ahh, after latest test failures I can see that removing it introduces a regression of not applying `readonly` prop to the input element. 🤔 🙈

LukasTy: The issue is that `TextField` does not support `readOnly` prop.
It is "officially" supported [via `InputProps`](https://mui.com/material-ui/react-text-field/#form-props). 🤔 
This requires some interesting solution, or getting back to my initial solution 🙈 😆

packages/x-date-pickers/src/tests/describeValue/testControlledUnControlled.tsx
LukasTy: I've [extracted the test](https://github.com/mui/mui-x/pull/7135/files/#diff-5cdf0facdbadac7fca4b02cd2f21e4e274e2911f215483ca1b9fe0906faf3bebR98) for mobile/readonly keyboard editing. 😉

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
flaviendelangle: :see_no_evil:

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
```diff
@@ -28,7 +28,7 @@ export type UseMultiInputDateTimeRangeFieldComponentProps<TDate, TChildProps ext
   UseMultiInputDateTimeRangeFieldProps<TDate>;
 
 export interface MultiInputDateTimeRangeFieldProps<TDate>
-  extends UseMultiInputDateTimeRangeFieldComponentProps<TDate, TextFieldProps> {
+  extends UseMultiInputDateTimeRangeFieldComponentProps<TDate, {}> {
```
LukasTy, line 31: @alexfauquette I'm not sure why this component had `TextFieldProps` in here, but it was the one, which inspired me to initially go with adding it to other multi input range components and confuse me at the end.
Did end up removing this from all components and this is the only one, which had it in the first place.

packages/x-date-pickers-pro/src/SingleInputDateRangeField/SingleInputDateRangeField.tsx
```diff
@@ -45,7 +46,7 @@ const SingleInputDateRangeField = React.forwardRef(function SingleInputDateRange
     <Input
       ref={ref}
       {...fieldProps}
-      inputProps={{ ...fieldProps.inputProps, ref: inputRef, onPaste, inputMode }}
+      inputProps={{ ...fieldProps.inputProps, ref: inputRef, onPaste, inputMode, readOnly }}
```
LukasTy, line 49: Single input range fields also do not propagate `readonly` prop correctly, because of `TextField` API. 😢

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.types.ts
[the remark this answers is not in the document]
LukasTy: Ended up removing it 🙈

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
alexfauquette: I probably forget that props were passed to the Stack and not the TextField

`{}` could be replaced by `StackProps`

It let me realise that the default Root components does not allow to override the `spacing`, `direction`, and `alignItems`


https://github.com/mui/mui-x/blob/21afe09ac1cc2388fce45762597bda0d2110a32c/packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.tsx#L11-L20

flaviendelangle: > It let me realise that the default Root components does not allow to override the spacing, direction, and alignItems

The spread should probably be after indeed

LukasTy: Will create a separate PR to address these issues. 😉

packages/x-date-pickers-pro/src/NextDateRangePicker/describes.NextDateRangePicker.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+      'componentsProp',
+      'themeDefaultProps',
+      'themeStyleOverrides',
+      'themeVariants',
+      'mergeClassName',
+      'propsSpread',
+      'rootClass',
+      'reactTestRenderer',
+    ],
+  }));
```
alexfauquette, line 23: Is there a reason why not copying the following part from [NextDateRangePicker.test.tsx](https://github.com/mui/mui-x/pull/7135/files#diff-a55031bffcc8e2fd75307dca229fa68208b089c2b4e97fef04471bf1b0ac6a78)?

```js
describeRangeValidation(NextDateRangePicker, () => ({
    render,
    clock,
    componentFamily: 'new-picker',
    views: ['day'],
  }));
```

packages/x-date-pickers-pro/src/internal/hooks/useMobileRangePicker/useMobileRangePicker.tsx
alexfauquette: Since you already know it's a mobile component, you could use

```suggestion
      readOnly: readOnly ?? true;
```

packages/x-date-pickers-pro/src/NextDateRangePicker/describes.NextDateRangePicker.test.tsx
LukasTy: Good spot. Yes, there is. As per Flaviens suggestion, all other tests where respective Mobile and Desktop describe suites. 😉 
But this also raises a good question. Do we need a `describeConformance` test on the wrapping component (i.e. NextDateRangePicker), or should we just put two separate conformance tests in both `Mobile` and `Desktop` versions? 🤔 @flaviendelangle 
From the looks of what the `describeConformance` is doing I'd say it would probable end up being redundant 🤔

flaviendelangle: `describeConformance` would test the theme default props applied on the responsive picker.
That can be worth testing.

packages/x-date-pickers-pro/src/internal/hooks/useMobileRangePicker/useMobileRangePicker.tsx
LukasTy: Thanks for a suggestion. Aligned it with regular `useMobilePicker` by updating to `readOnly: true,`. This disables provided `readOnly` prop from being applied as we decided to not support it on mobile pickers for now.

packages/x-date-pickers/src/internals/hooks/useField/useField.types.ts
```diff
@@ -59,6 +59,11 @@ export interface UseFieldInternalProps<TValue, TError> {
    * @default false
    */
   readOnly?: boolean;
+  /**
+   * If `true`, the field is disabled.
+   * @default false
+   */
+  disabled?: boolean;
```
flaviendelangle, line 66: Right now `useField` is not using `disabled`

Maybe I'm missing something, but I think we should either;

1. Handle `disabled` in `useField` (add it to `UseFieldInternalProps` and return it in `UseFieldResponse`), which should allow to remove it from the `forwardedProps` and avoid the `other.disabled` you added.

2. Handle `disabled` in the components

Is there something preventing us from doing `2.` ? Given that `disabled` is at the root level of the `TextField`, shouldn't it behave like any other prop `useField` do not use ? (`variant` for example)

flaviendelangle (commented):
Great work !

I still have a few questions around how we pass the props from the pickers or the field components to `TextField`

packages/x-date-pickers-pro/src/internal/hooks/useRangePickerInputProps.ts
```diff
@@ -81,6 +81,8 @@ export const useRangePickerInputProps = <TDate, TView extends DateOrTimeView>({
     }
   };
 
+  const readOnlyInput = readOnly ?? wrapperVariant === 'mobile';
```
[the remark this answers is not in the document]
LukasTy: Good spot! 🙌 
This is probably a left-over from the initial experiments. Removing it does not seem to produce any regressions. 🎉

packages/x-date-pickers-pro/src/internal/hooks/useRangePickerInputProps.ts
```diff
@@ -104,8 +111,13 @@ export const useRangePickerInputProps = <TDate, TView extends DateOrTimeView>({
...
     // which has `pointer-events: none` and due to DOM structure the `input` does not catch the click event
     ...(!readOnly && !disabled && { onClick: openRangeEndSelection }),
+    // required to disable field editing
+    readOnly: readOnlyInput,
+    // passing same props down to both fields and input props
+    disabled,
     inputProps: {
-      readOnly: wrapperVariant === 'mobile',
+      readOnly: readOnlyInput,
+      disabled,
```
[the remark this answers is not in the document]
LukasTy: [Same answer](https://github.com/mui/mui-x/pull/7135/files#r1054363704) applies 😉

packages/x-date-pickers/src/internals/hooks/useField/useField.types.ts
LukasTy: > Handle disabled in the components

Isn't it already handled in components? For single input and base (Date, Time, and DateTime) fields this prop is transitively passed down.
Did you mean to remove it from `useField` and put it somewhere in `MultiInput` fields? As they are the ones who need this prop explicitly defined. 🤔

flaviendelangle: But if is handled by the component, why do we need to define its type here ?

flaviendelangle: > Did you mean to remove it from useField and put it somewhere in MultiInput fields? As they are the ones who need this prop explicitly defined. 

Yes
To avoid having the prop defined in an interface of `useField` when `useField` never care about it

LukasTy: > Unless we want to define it here to have a better JSDoc

Well... I was also debating on the JSDoc for it... Changing from __If `true`, the component is disabled.__ to __If `true`, the field is disabled.__ makes sense for Field components or single input Range fields, but for multi input ones it is a bit misleading... :/ 
It would probably make more sense staying with the original `TextFieldProps.disabled` JSDoc 🙈

LukasTy: [Moved the disabled type declaration](https://github.com/mui/mui-x/pull/7135/commits/9d2f3f42610e0bd266b49eef08bc80748697c761) to be explicit only for `range` fields.
Seems to simplify the code quite a lot.
Thanks for very insightful comments! 🙏

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
```diff
@@ -151,7 +151,6 @@ export const useDesktopRangePicker = <
         inputProps: {
           ...externalInputProps?.inputProps,
           ...inputPropsPassedByField?.inputProps,
-          ...inputPropsPassedByPicker?.inputProps,
```
LukasTy, line 154: This was previously only available because of untyped explicit `inputProps: { readOnly }` coming from `useRangePickerInputProps`. It had no other props.
That is no longer relevant.

packages/x-date-pickers/src/internals/hooks/useMobilePicker/useMobilePicker.tsx
```diff
@@ -76,9 +76,11 @@ export const useMobilePicker = <
       return {
         ...inputPropsPassedByField,
         ...externalInputProps,
-        disabled: props.disabled,
-        onClick: props.readOnly || props.disabled ? undefined : actions.onOpen,
-        onKeyDown: onSpaceOrEnter(actions.onOpen),
```
LukasTy, line 81: This was the oversight still allowing interactivity in mobile pickers with `readOnly` flag.

flaviendelangle (approved):
Great work :partying_face:

Issue #7126, opened 2022-12-07
[pickers] The placeholder of the masked-based pickers are not localized (shows `yyyy` for year, even in Portuguese)


Asghwor:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

Link to live example:
[demo](https://codesandbox.io/s/small-bird-msf0qt)

Steps:
1. Create a React project and setup Material UI and `@mui/x-date-pickers`
2. Use the pt-br (Brazilian Portuguese) locale for both the `LocalizationProvider` and the `ThemeProvider`
3. Notice the placeholder of the date picker is `dd/mm/yyyy` instead of `dd/mm/aaaa`

### Current behavior 😯

The letter used for the date picker's placeholder for the year is "y" instead of "a" in Brazilian Portuguese.
E.g.: `dd/mm/yyyy` instead of `dd/mm/aaaa`, `mm/yyyy` instead of `mm/aaaa`, etc.

### Expected behavior 🤔

The letter used for the year should be "a".
E.g.:  `dd/mm/aaaa` instead of `dd/mm/yyyy`, `mm/aaaa` instead of `mm/yyyy`, etc.

### Context 🔦

The word for "year" in Portuguese is "ano", starting with an "a", so the placeholders showcasing the mask to be used in a given date picker (day/month/year, month/year, day/year, year, etc) needs to use this letter to be readily understood by Portuguese speakers.

### Your environment 🌎

- @mui/material 5.10.17
- @mui/x-date-picker 5.0.9

flaviendelangle:
On the new pickers that will be released on v6, we are applying localization on the placeholders.
But for the v5 pickers, the placeholders is pretty-much the format from the date library, and the date libraries don't apply any localization on the format itself.

LukasTy:
> On the new pickers that will be released on v6, we are applying localization on the placeholders.

But the result for v6 will be the same? Or am I missing something?
![Screenshot 2022-12-08 at 09 46 19](https://user-images.githubusercontent.com/4941090/206388828-f8146075-ae78-433f-b612-38ea463a71e0.png)

flaviendelangle:
On v6 pickers, the placeholders are generated using the following locale keys:

```ts
  // Field section placeholders
  fieldYearPlaceholder: (params) => 'Y'.repeat(params.digitAmount),
[...]

PR #7139, opened 2022-12-08
[docs] Add doc section about field placeholders localization

flaviendelangle:
Part of #7126


packages/x-date-pickers/src/locales/deDE.ts
```diff
@@ -80,13 +80,13 @@ const deDEPickers: Partial<PickersLocaleText<any>> = {
...
 
   // Field section placeholders
-  // fieldYearPlaceholder: params => 'Y'.repeat(params.digitAmount),
-  // fieldMonthPlaceholder: params => params.contentType === 'letter' ? 'MMMM' : 'MM',
-  // fieldDayPlaceholder: () => 'DD',
-  // fieldHoursPlaceholder: () => 'hh',
-  // fieldMinutesPlaceholder: () => 'mm',
-  // fieldSecondsPlaceholder: () => 'ss',
-  // fieldMeridiemPlaceholder: () => 'aa',
+  fieldYearPlaceholder: (params) => 'J'.repeat(params.digitAmount),
```
flaviendelangle, line 83: @joserodolfofreitas I used the first letter of each word just like in english
If you can check that these placeholder make sense.

I'm using german in the doc because we use it in the example above.

docs/data/date-pickers/adapters-locale/adapters-locale.md
```diff
@@ -137,6 +137,17 @@ This prop is available on all fields and pickers.
 
 {{"demo": "CustomFieldFormat.js"}}
 
+### Custom field placeholder
```
flaviendelangle, line 140: It is not per-say a `format` example, but I'm putting it here because all the other places I thought off are worse.
And most of the time, I guess people will want to add localization to their placeholder just after changing the format, so I think it's the most natural place for them to easily find it.

packages/x-date-pickers/src/locales/deDE.ts
LukasTy: Honestly, I'm still not sure about this format localisation based on localised section name.
Did we have any benchmarks for introducing it?
I imagine some languages might have issues with it.
For example in Lithuanian:
- `Year` -> `Metai`
- `Month` -> `Mėnuo`
- `Day` -> `Diena`

I take it that based on other locales this would translate to `MMMM-MM-DD` (we're mostly using ISO format).
IMHO, I would find it utterly confusing. 🙈

flaviendelangle: The common usage may depend on the language.
In French, it is very common to have the `AAAA` for the year.
Nothing forces the locale to use the language value, if in Lithuanian, people are used to the english values, nothing prevents from keeping them in the Lithuanian locale.

LukasTy: Fair points. It will allow control over what a different locale users would expect.

But it still leaves the question of German locale and `ss:mm:ss` (hh:mm:ss)
I did not find a way to trigger our components into having the seconds in the field format, other than providing `format` (which overrides this behaviour).

That's also a question and maybe a feature we should support?
If `views` array contains `seconds`—maybe it should be included in the format? Especially given the fact, that time `views` currently have no effect on `Next` pickers 🤔

flaviendelangle: > But it still leaves the question of German locale and ss:mm:ss (hh:mm:ss)

The best answer here would have to get a German native speaker (or at least better than me) :laughing: 

> I did not find a way to trigger our components into having the seconds in the field format, other than providing format (which overrides this behaviour).

There is no other way, we could indeed have a smart format based on the year on `TimeField` just like on `DateField`

flaviendelangle: If you want, we can add a note to say that all locale decide to translate or not there placeholders.
If the doc I did felt confusing

LukasTy: > If you want, we can add a note to say that all locale decide to translate or not there placeholders.
> If the doc I did felt confusing

Yup, I'd say that might end being useful. Basically mention that if field placeholder translations are not available for the selected locale—English ones will be used.

I'd also say that it might be useful to have a link to this section from `Field components` page.
Maybe add a section there, or just a `info` alert somewhere in the content?

flaviendelangle: > I'd also say that it might be useful to have a link to this section from Field components page.

I am not sure how the `Field components` page should look like, its content should probably move in part to `Getting Started` and in part to a more advanced field page.
But until we have a clear vision, sure we can add a note there.

flaviendelangle: @LukasTy Added

docs/data/date-pickers/adapters-locale/adapters-locale.md
```diff
@@ -137,6 +137,22 @@ This prop is available on all fields and pickers.
...
+
+These placeholders are based on your current component localization, not on your date localization.
+
+{{"demo": "FieldPlaceholder.js"}}
+
+For more information on how to define your component localization, check out the [Translated components](/x/react-date-pickers/localization/) page.
+
+:::warning
+Not all locale are translating the placeholders.
+Some are keeping the English letters because it is what is usually done for the dates in those languages.
```
LukasTy, line 153: Placeholders translations depend on locale.
Some locales might keep using English placeholders, because that format is commonly used in a given locale.

LukasTy (approved):

PR #7140, opened 2022-12-08
[docs] Rework the pickers Getting Started page

flaviendelangle:
[Doc preview](https://deploy-preview-7140--material-ui-x.netlify.app/x/react-date-pickers/getting-started/)

I have a few goals here

### Help people setup their application

- Explain what the differences are between the pro and the community package (like on the Data Grid  Overview page)
- Add a note about `@mui/material` and its peer deps like on the Data Grid Getting Started page
- Make clear that `LocalizationProvider` should, most of the time, be rendered at the root of the app, like a redux or a react-router provider.

### Help people solve their most basic problems

- Add sections about things like controlling a value

### Help people find which component they need

- Add a list with the commonly used components, it should feet 80% of the needs
- Add a detailed explanation of all the type of components we are exporting and why
- Add a interactive section to let people find the component they need


docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+Each _Picker_ is the combination of one _Field_ and one or several _Calendar / Clock_ components.
+For example, the `DatePicker` is the combination of the `DateField` and the `DateCalendar`.
+
+The _Calendar / Clock_ components are rendered inside a _Popover_ on desktop and inside a _Modal_ on mobile.
+:::
+
+### Date or time editing ?
+
+The Date and Time Pickers are divided into six families of components.
+The demo below shows each one of them using their field component:
```
flaviendelangle, line 211: I'm using the field to use the same type of component on each family but we can replace with the pickers when they are all available.

I don't think it's a problem because it's an advanced section, nobody should reach this section when just looking for a `DatePicker`.

LukasTy: Fair point. I don't see problems with it. 🤔

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+{{"demo": "ComponentFamilies.js"}}
+
+### Responsiveness
```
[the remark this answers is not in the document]
LukasTy: Hm...
This would create an inconsistency between our new approach of naming such section in individual docs as `Available components`.

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+Each _Picker_ is available in three responsive variants:
```
LukasTy, line 217: Technically, they all have only one _responsive_ variant. 🤔 
They can also be used as `desktop` or `mobile` variants as well as `static` variant.
Maybe @samuelsycamore or @joserodolfofreitas would have better ideas on how to "translate" this behaviour/setup of our components into words. 🙈 
For reference: https://next.mui.com/x/react-date-pickers/date-picker/#available-components

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -121,16 +262,6 @@ You don't have to import the theme augmentation from both `@mui/x-date-pickers`
 Importing it from `@mui/x-date-pickers-pro` is enough.
 :::
 
-## Native pickers
```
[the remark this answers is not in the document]
LukasTy: Agreed, IMHO this section made little to no sense in this location...
Maybe it would have usefulness at the beginning of the page, but that is also debatable.
If people are in search of a date library, they probably already know that native solutions lack what they need. 🤷 🙈 💡

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+- The responsive component (e.g. `NextDatePicker`) which renders the desktop component or the mobile one depending on the device it runs on.
+
+{{"demo": "ResponsivePickers.js"}}
+
+### Find your component
+
+Use the form below to find component you need:
+
+{{"demo": "ComponentExplorerNoSnap.js", "hideToolbar": true}}
```
LukasTy, line 233: This is a really nice demo!!! 🚀 ❤️

docs/data/date-pickers/getting-started/ComponentExplorerNoSnap.tsx
```diff
@@ -0,0 +1,239 @@
...
+            <MenuItem value="view">Calendar / Clock</MenuItem>
+          </Select>
+        </FormControl>
+      </Stack>
+      {exportedNames.length > 0 && (
+        <React.Fragment>
+          <Stack>
+            {docPages.map((docPage) => (
+              <Link href={docPage.path} rel="noopener" target="_blank">
+                Documentation {docPage.name}
```
LukasTy, line 218: What do you think about putting the `Documentation` word after the component?
Also it would be nice to change the styles/layout to make anchor not take the whole container width, because now you can mistakenly open it by clicking in empty space. 🙈  
Screenshot 2023-01-12 at 17 17 25

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
 The results may vary in your application depending on the version of each library, the locale, and the bundling strategy used.
 :::
 
+### Other peer dependencies
+
+The Date and Time Pickers package has a peer dependency on `@mui/material`.
+If you are not already using it in your project, you can install it with:
```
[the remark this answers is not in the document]
flaviendelangle: I think `you can` means `your are able to` here not `you have the possibility to`.
It's @samuelsycamore wording on the grid equivalent page, if we change on this one we should change there to.

docs/data/date-pickers/getting-started/getting-started.md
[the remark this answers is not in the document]
flaviendelangle: True, but if we use `Available components` it's super unclear in the context of this page.

On the `date-picker.md`, `Available components` list all the components in the scope of the page.
That's not the case here, we are just talking about the component dedicated to the various type of device.

If at one point we migration to the modal vs popover naming, then we could have a name describing this behavior.

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -7,29 +7,29 @@ materialDesign: https://m2.material.io/components/date-pickers
...
+<p class="description">The Date and Time pickers let the user select date and time values.</p>
 
 {{"component": "modules/components/ComponentLinkHeader.js"}}
 
-- On mobile, pickers are best suited for display in a confirmation dialog.
-- For inline display, such as on a form, consider using compact controls such as segmented dropdown buttons.
+## Installation
 
-## React components
+To make the Date and Time Pickers, work, you need to install the following packages:
```
alexfauquette, line 18:
```suggestion
To make the Date and Time Pickers work, you need to install the following packages:
```

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+```sh
+// with npm
+npm install @mui/material @emotion/react @emotion/styled
+
+// with yarn
+yarn add @mui/material @emotion/react @emotion/styled
+```
+
+<!-- #react-peer-version -->
```
alexfauquette, line 90: Does it have an interest for maintenance of the docs?
```suggestion
```

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
 ## Code setup
 
-After installation, you have to pass your chosen date library's adapter to `LocalizationProvider`.
-The supported adapters—as well as `LocalizationProvider`—are exported from both the `@mui/x-date-pickers` and `@mui/x-date-pickers-pro` packages.
+Before trying to render any component, you have to make sure that there is a `LocalizationProvider` upper in the React tree.
+This component receives your chosen date library's adapter and make it accessible to all the Date and Time Pickers component with a React context.
 
-All the pickers rendered inside this provider will have access to the adapter through a React context.
-For this reason, we recommend that you wrap your entire application with a `LocalizationProvider` so you can use the Date and Time Pickers everywhere.
+The most simple way to use `LocalizationProvider` is to wrap your entire application it.
```
alexfauquette, line 109:
```suggestion
The simplest way to use `LocalizationProvider` is to wrap your entire application in it.
```

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
 :::info
 If you need to use the Date and Time Pickers with a custom locale, have a look at the [Localized dates](/x/react-date-pickers/adapters-locale/) page.
 :::
 
+## Commonly used components
+
+Most applications only need a few of the components exposed by the Date and Time packages:
+
+{{"demo": "CommonlyUsedComponents.js"}}
```
alexfauquette, line 142: Amazing demo for introducing and playing with components 🎉

But it's the first time reader encounter a demo, and the code is complicated to read.

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,163 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+```tsx
+import { DatePicker } from '@mui/x-date-pickers/DatePicker';
+import { DatePicker } from '@mui/x-date-pickers';
+import { DatePicker } from '@mui/x-date-pickers-pro';
+```
+
+:::
+
+## Basic use cases
```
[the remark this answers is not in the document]
alexfauquette: I agree that "form props" can be discovered by reading deeper docs, or simply by typing `disabled` in the components like for all the MUI components 😉

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+### Validation
+
+If you need to add validation to your component, you can check the [Validation page](/x/react-date-pickers/validation/):
+
+{{"demo": "ValidatedComponent.js"}}
+
+## Other components
+
+### Keyboard or mouse editing
```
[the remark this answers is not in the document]
alexfauquette: "User interaction"?

docs/data/date-pickers/getting-started/getting-started.md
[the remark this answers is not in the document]
alexfauquette: Yes, you could skip them. Their main use case is if you need to create advanced components where you override the Modal/Popper. So it can be deeper in the docs

docs/data/date-pickers/getting-started/getting-started.md
alexfauquette, line 233: 😍🤩

alexfauquette (commented):
You introduce lot of useful concept to prevent user to fall into traps, but the page sounds too long.

- Reader may skip the explanation
- The demo in "Commonly used components" is a super nice introduction to what is in the package but lost in the middle


I propose to divide it in 3 pages:

- Overview
  - What are pickers with the demo from **Commonly used components**
  - Community or Pro Plan? (up to their, not technical point, a manager can read it)
  - Date library (Here we are going technical, to say "Hey, we support you're date library")
- Getting started
  - A link back the "Overview" page to select the date library and the licence (in case they skipped it)
  - The installation of `@mui/material`
  - The installation of the packages
  - The code set-up
  - Maybe the last step should be to test the setup by adding a `<DatePicker />` somwhere in their app and verify their is no warning in the console.
  - An call to action to visit the next page (or end with something like "now you're app is ready, got to next page to see how to add components")
- Introduction
  - Import component
  - ... (the rest of the page)
  
  
They will still have the same length of content to read but they are 3 different scenario, so you can read then at different moment of the day without having to find where you stopped:
- The 2 first pages are to read once
	- The first one when you are looking for which library component you want to go with (can be read on smartphone)
	- The second one when you set-up an app
- The last page is about all the big point to have in mind. It's more like a tutorial for your first interaction. You could come back in case you're not sure about one point, and reader can skip section without any problem

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -7,29 +7,29 @@ materialDesign: https://m2.material.io/components/date-pickers
...
 
-## React components
+To make the Date and Time Pickers, work, you need to install the following packages:
 
-{{"demo": "MaterialUIPickers.js"}}
+1. **The component library** (`@mui/x-date-pickers` or `@mui/x-date-pickers-pro`) to render the component.
+2. **The date library** ([Day.js](https://day.js.org/), [date-fns](https://date-fns.org/), ...) to manipulate the date.
 
-## Installation
+{{"component": "modules/components/PickersInstallationInstructions.js"}}
```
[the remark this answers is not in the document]
flaviendelangle: My approach here was to consider that the vast majority of people would understand the form without reading all those long sections.
They see "Date library", they click, they see there favorite date library, they install.

With that approach, the interactive installation is the main element and the sections and a notice guide coming after.

Other wise, the section ends up super low and is easy to miss.

docs/data/date-pickers/getting-started/getting-started.md
flaviendelangle: No idea, I copy pasted the grid doc :laughing:

docs/data/date-pickers/getting-started/getting-started.md
flaviendelangle: Yeah, this demo is interesting to help people find quickly their component, not to help them copy paste code...
We could split into 4 demos (this page has a ton of very basic demo anyway)

But my idea was that it was the "Controlled value" demo which should be the entry point for people to copy / paste, since it's closer from a real use case scenario.
I do agree that it's not very accessible right now though.

docs/data/date-pickers/getting-started/getting-started.md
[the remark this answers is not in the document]
flaviendelangle: If you remove entirely this section, would you have no place to describe our responsive behaviors outside of the component pages ? 

That can make sense
Or we could, in the future, have a page describing more in depth how we handle responsiveness (a boosted version of the `Available components` sections but with examples using `desktopMediaQuery`)

samuelsycamore:
This is definitely a big improvement, and I will follow up with a line-by-line review shortly. But on first glance, I agree with @alexfauquette's assessment that this doc should be broken up into multiple pages. I think as a general rule, we should always try to avoid introducing a scroll bar to the table-of-contents navigation whenever possible.

flaviendelangle:
I fully agree with the idea of splitting the page.
And the split proposed by @alexfauquette make sense to me, looks a lot more like a tutorial one can follow and check the boxes.

docs/data/date-pickers/getting-started/getting-started.md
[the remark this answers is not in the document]
flaviendelangle: I went for `Choose interaction style`, let's see what Sam proposes :+1:

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -76,20 +74,164 @@ The bundling strategy was taken care of by a Create React App, and no locale was
...
+
+- For keyboard and mouse editing, use the _Picker_ components:
+
+{{"demo": "BasicDatePicker.js", "hideToolbar": true, "bg": "inline"}}
+
+- For keyboard-only editing, use the _Field_ components:
+
+{{"demo": "BasicDateField.js", "hideToolbar": true, "bg": "inline"}}
+
+- For mouse-only editing, use the _Calendar / Clock_ components:
```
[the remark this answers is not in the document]
flaviendelangle: I made a proposal on the new `introduction.md` page
Still note convinced but that's better

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Date and Time Pickers - Introduction
+
+<p class="description">The Date and Time pickers expose a lot of components to fit your every needs.</p>
+
+{{"component": "modules/components/ComponentLinkHeader.js"}}
+
+## Basic usage
+
+All the components can be uncontrolled or controlled by using the `value` / `onChange` props:
```
joserodolfofreitas, line 18:
```suggestion
All the components have a `value` / `onChange` API to set and control the values:
```

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+{{"demo": "ControlledComponent.js"}}
+
+## Imports
+
+All the components exported by `@mui/x-date-pickers` are also exported by `@mui/x-date-pickers-pro` but without the nested imports.
```
joserodolfofreitas, line 24: > but without the nested imports.

What's the reason for it? Does it impact tree shaking?

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+import { DatePicker } from '@mui/x-date-pickers/DatePicker';
+import { DatePicker } from '@mui/x-date-pickers';
+import { DatePicker } from '@mui/x-date-pickers-pro';
+```
+
+## Other components
+
+### Choose interaction style
+
+The first thing you need to decide, is how the user should be able to select their value:
```
joserodolfofreitas, line 38:
```suggestion
Depending on your use case, different interaction styles are preferred.
```

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+- For input editing with a popover, use the _Picker_ components:
```
joserodolfofreitas, line 40:
```suggestion
- For input editing with a popover for mouse interaction, use the _Picker_ components:
```

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+The _Calendar / Clock_ components are rendered inside a _Popover_ on desktop and inside a _Modal_ on mobile.
+:::
+
+### Date or time editing ?
+
+The Date and Time Pickers are divided into six families of components.
+The demo below shows each one of them using their field component:
+
+{{"demo": "ComponentFamilies.js"}}
```
joserodolfofreitas, line 64: Again for this demo I would have the `<span className="plan-pro"/>` before the name and with a tooltip.

```suggestion
{{"demo": "ComponentFamilies.js"}}
```

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+Each _Picker_ is available in a desktop, mobile and responsive variant:
+
+- The desktop component (e.g. `DesktopNextDatePicker`) which works best for mouse devices and large screens.
+  It renders the views inside a popover and allows editing values directly inside the field.
+
+- The mobile component (e.g. `MobileNextDatePicker`) which works best for touch devices and small screens.
+  It renders the view inside a modal and does not allow editing values directly inside the field.
+
+- The responsive component (e.g. `NextDatePicker`) which renders the desktop component or the mobile one depending on the device it runs on.
```
joserodolfofreitas, line 76: Maybe describe the responsive first?

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -0,0 +1,144 @@
...
+
+{{"demo": "ResponsivePickers.js"}}
+
+### Find your component
+
+Use the form below to find component you need:
```
joserodolfofreitas, line 82:
```suggestion
There are many options of components available, each fitting specific use cases. Use the form below to find the component you need:
```

joserodolfofreitas (commented):
Great work setting up this introduction content!

I'm wondering if "Getting started" is too close to "Introduction" (semantically speaking).
Maybe it makes sense to call it "Installation and setup" ?

I'm also wondering if it makes sense to move the Date library with its recommendation section to the Installation page, and maybe merge the overview and introduction.

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -1,161 +1,91 @@
...
-In such case you can [force pointer precision](https://github.com/microsoft/playwright/issues/7769#issuecomment-1205106311) via browser flags or preferences.
+## Render your first component
 
-### Field components
+To make sure that everything is set up correctly, try rendering a simple `DatePicker` component:
 
-:::info
-To support RTL and some keyboard interactions, field components add some Unicode character that are invisible, but appears in the input value.
-:::
+{{"demo": "FirstComponent.js"}}
```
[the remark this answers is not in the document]
flaviendelangle: > and wrap with `LocalizationProvider` on the application level?

How would you do it ?
For now, the doc does not set a `LocalizationProvider` outside of the demo.
We could add one to enable demos without `LocalizationProvider`, if we consider that the doc is clear enough about how it should be used (it won't be possible to copy paste a demo, just like with the Joy demos which require the CSS Variables context).

docs/data/date-pickers/overview/overview.md
```diff
@@ -0,0 +1,72 @@
...
+
+{{"component": "modules/components/ComponentLinkHeader.js"}}
+
+{{"demo": "CommonlyUsedComponents.js"}}
+
+## Community or Pro Plan ?
+
+The Date and Time Pickers comes in two packages:
+
+- `@mui/x-date-pickers`, which is open-source and contains all the components to edit a date and/or a time.
```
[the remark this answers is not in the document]
flaviendelangle: We should also update the grid equivalent line then

docs/data/date-pickers/introduction/introduction.md
flaviendelangle: The reason is to avoid creating dozens of folder in `@mui/x-date-pickers-pro` that just re-export their equivalent from `@mui/x-date-pickers`

The tree shaking works fine on root imports in the whole MUI ecosystem.
Nested imports are only usefull for dev treeshaking.

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -53,8 +53,10 @@ yarn add @mui/material @mui/styled-engine-sc styled-components
 Before trying to render any component, you have to make sure that there is a `LocalizationProvider` upper in the React tree.
 This component receives your chosen date library's adapter and makes it accessible to all the Date and Time Pickers component using React context.
 
-The simplest way to use `LocalizationProvider` is to wrap your entire application in it.
-You will then be able to use the Date and Time Pickers everywhere.
+Most of the demos in the documentation use the `LocalizationProvider` wrapping only the components being demoed. That's because we aim to keep our examples atomic and functional with copy and paste.
```
alexfauquette, line 56:
```suggestion
Each demonstration in the documentation has its own `LocalizationProvider` wrapper.
This is **not** a pattern to reproduce.
The reason is to keep examples atomic and functional.
Especially when running in a CodeSandbox.
```

docs/data/date-pickers/introduction/introduction.md
```diff
@@ -79,7 +79,7 @@ Each _Picker_ is available in a desktop, mobile and responsive variant:
 
 ### Find your component
 
-Use the form below to find component you need:
+There are many options of components available, each fitting specific use cases. Use the form below to find the component you need:
```
alexfauquette, line 82: When reading it I was wondering about which option are you talking about

```suggestion
There are many components available, each fitting specific use cases. Use the form below to find the component you need:
```

samuelsycamore:
This might be beyond the scope of this PR, but the left nav bar is getting a bit cluttered now with the extra pages, and there's not a very clear visual hierarchy. Maybe we could subdivide the pages into multiple sections like Joy UI has done in their Getting Started section?

Compare Pickers vs Joy:

Screenshot 2023-01-18 at 3 28 21 PM

Screenshot 2023-01-18 at 3 28 49 PM

samuelsycamore:
I'm still working on a line-by-line review, but one thing I'm noticing is that the terminology is getting a bit confusing:

The Date and Time Pickers package contains:
- Pickers
- Fields
- Calendars and Clocks

This leads to weird situations where we have to say things like "The Date and Time Pickers can be implemented as either Pickers, Fields, or Calendars and Clocks." It feels like we need a new noun to describe the concept of the Picker components within the package (vs. the name of the package itself, which contains those components but also others). 

As an example, maybe the Pickers could be called "Combos" (ok, bad example) since they combine Fields + Calendars/Clocks. Then we could instead say "The Date and Time Pickers can be implemented as either Combos, Fields, or Calendars and Clocks." Does that make sense?

LukasTy:
> Then we could instead say "The Date and Time Pickers can be implemented as Fields, Calendars and Clocks, or Combos containing both." Does that make sense?

We would probably lean towards something like:
"The Date and Time Pickers provides the full combined picker component variants as well as standalone Fields, Calendars and Clock components".
Or am I mistaken @flaviendelangle @alexfauquette ?

flaviendelangle:
> This leads to weird situations where we have to say things like "The Date and Time Pickers can be implemented as either Pickers, Fields, or Calendars and Clocks.

Yes, and that's an issue on the whole doc.
This situation comes mainly from the name of our packages (`@mui/x-date-pickers`).
That's why I insist on saying "The Date and Time Pickers" instead of "The pickers" each time we are talking about all our components (including fields and calendar / clocks). 
But that's far from perfect.

On specific occurrences, maybe a formulation like the one proposed by @LukasTy could work better.
But for the rest, if you have a better formulation, maybe we can just ignore the name of the package and call the full list of the components "Date and Time picking components" for example.

Naming the actual pickers components differently has one big downside which is their react name.
They are named `XXXPIckers` in the code, and if in the doc we don't say "Pickers" when we talk about them, I'm not sure everybody will understand depending on the context. 

____

As a side note, we have an even worse scenario with the `StaticXXXPicker` components which are not even real pickers based on the description we are making.
But I think we can ignore this inconsistency for v6 and discuss what we want to do with them on v7 or later one.

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -0,0 +1,144 @@
...
+import { DatePicker } from '@mui/x-date-pickers-pro';
+```
+
+## Other components
+
+### Choose interaction style
+
+Depending on your use case, different interaction styles are preferred.
+
+- For input editing with a popover for mouse interaction, use the _Picker_ components:
```
LukasTy, line 40: Nitpick: This could potentially use a bit different wording in order to cover mobile dialog behaviour as well.
Also, technically both desktop and mobile shows a dialog (at least depending on the provided `role`) 🙈

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -0,0 +1,144 @@
...
+
+{{"demo": "BasicDatePicker.js", "hideToolbar": true, "bg": "inline"}}
+
+- For input-only editing, use the _Field_ components:
+
+{{"demo": "BasicDateField.js", "hideToolbar": true, "bg": "inline"}}
+
+- For inline editing, use the _Calendar / Clock_ components:
```
LukasTy, line 48: Shouldn't this demo the `Static` picker version? 🤔 
Separate sub-components could maybe be a different section on it's own?

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -0,0 +1,144 @@
...
+
+Each _Picker_ is available in a desktop, mobile and responsive variant:
+
+- The responsive component (e.g. `DatePicker`) which renders the desktop component or the mobile one depending on the device it runs on.
+
+- The desktop component (e.g. `DesktopDatePicker`) which works best for mouse devices and large screens.
+  It renders the views inside a popover and allows editing values directly inside the field.
+
+- The mobile component (e.g. `MobileDatePicker`) which works best for touch devices and small screens.
+  It renders the view inside a modal and does not allow editing values directly inside the field.
```
LukasTy, line 76: Maybe put the responsive variant description at the bottom to align with the order described above? 🤔 
```suggestion
- The desktop component (e.g. `DesktopDatePicker`) which works best for mouse devices and large screens.
  It renders the views inside a popover and allows editing values directly inside the field.

- The mobile component (e.g. `MobileDatePicker`) which works best for touch devices and small screens.
  It renders the view inside a modal and does not allow editing values directly inside the field.

- The responsive component (e.g. `DatePicker`) which renders the desktop component or the mobile one depending on the device it runs on.
```

docs/data/date-pickers/overview/CommonlyUsedComponents.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  if (isProOnly) {
+    return (
+      <Stack direction="row" spacing={0.5} component="span">
+        <Tooltip title="Included on Pro package">
+          <a href="/x/introduction/licensing/#pro-plan">
+            <span className="plan-pro" />
+          </a>
+        </Tooltip>
+        {content}
```
LukasTy, line 35: I'd suggest inverting the order of text and icon to align with other usages.
```suggestion
        {content}
        <Tooltip title="Included in Pro package">
          <a href="/x/introduction/licensing/#pro-plan">
            <span className="plan-pro" />
          </a>
        </Tooltip>
```

docs/data/date-pickers/base-concepts/ComponentFamilies.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+function ProLabel({ children }: { children: React.ReactNode }) {
+  return (
+    <Stack direction="row" spacing={0.5}>
+      <Tooltip title="Included on Pro package">
+        <a href="/x/introduction/licensing/#pro-plan">
+          <span className="plan-pro" />
+        </a>
+      </Tooltip>
+      <span>{children}</span>
```
LukasTy, line 23: Maybe `Included in` is semantically a bit more correct? 🤔 
Also, inverting order of icon and text seems a bit more correct based on other usages.
```suggestion
      <span>{children}</span>
      <Tooltip title="Included in Pro package">
        <a href="/x/introduction/licensing/#pro-plan">
          <span className="plan-pro" />
        </a>
      </Tooltip>
```

LukasTy (commented):
Great job! 🙌 
Having separate pages helps with organization and separation of concerns. 👌 

Leaving a few improvement/nitpick comments. 😉

docs/data/date-pickers/overview/CommonlyUsedComponents.tsx
flaviendelangle: @joserodolfofreitas made me reverse in this way :laughing:

LukasTy: 🙈

docs/data/date-pickers/base-concepts/base-concepts.md
flaviendelangle: I don't think so

This section says "a picker is a field plus one or several views"
And it demonstrates, the picker, the field and the view.

If I start talking about the static pickers, who are pickers but not really because I said just before that a picker had a field and the static pickers don't have one, I fear it will confuse everybody.

docs/data/date-pickers/base-concepts/base-concepts.md
flaviendelangle: Jose asked to have the responsive at the top, I'll fix the order above instead

docs/data/date-pickers/base-concepts/base-concepts.md
LukasTy: Mhh, fair point. I did not look at it from this PoV.
Sound fair in that case. 👍

docs/data/date-pickers/base-concepts/base-concepts.md
LukasTy: Sounds like a plan as well. As long as we are staying consistent 👍

docs/data/date-pickers/getting-started/getting-started.md
```diff
@@ -1,161 +1,99 @@
...
 
-:::warning
-Native input controls support by browsers [isn't perfect](https://caniuse.com/#feat=input-datetime).
-:::
+- All the adapters are exported by both `@mui/x-date-pickers` and `@mui/x-date-pickers-pro`:
 
-Native date (`type="date"`), time (`type="time"`) and date&time (`type="datetime-local"`) pickers.
+  ```tsx
+  import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
+  import { AdapterDayjs } from '@mui/x-date-pickers-pro/`AdapterDayjs`';
```
LukasTy, line 75:
```suggestion
  import { AdapterDayjs } from '@mui/x-date-pickers-pro/AdapterDayjs';
```

LukasTy (approved):
I still stand by my opinion regarding the pro icon placement, but other than that and this last syntax fix, looks great to me. 👌

PR #7153, opened 2022-12-09
[pickers] Clean `autoFocus` behavior on fields and new pickers

flaviendelangle:
A lot of small big / small fixes when our components receive `autoFocus: true`.
I tried to clarify the when the field / view should be focused.
I hope I did not introduce any regression.

### Single Input Field

- `autoFocus: true` => focus the input and select all sections
- `autoFocus: false` => don't focus the input

### Multi Input Field

- `autoFocus: true` => focus the first input and select all of its sections
- `autoFocus: false` => don't focus any input

### Desktop Single Item Pickers

If `props.open === false` on mount

- `autoFocus: true` => focus the input and select all sections
- `autoFocus: false` => don't focus the input

If `props.open === true` on mount

- `autoFocus: true` => focus the visible view
- `autoFocus: false` => focus the visible view
(the `autoFocus` has no effect)

### Desktop Range Pickers

If `props.open === false` on mount

- `autoFocus: true` => focus the first input and select all of its sections
- `autoFocus: false` => don't focus any input

If `props.open === true` on mount

- `autoFocus: true` => focus the visible view
- `autoFocus: false` => focus the visible view
(the `autoFocus` has no effect)

### Mobile Pickers

No `autoFocus` prop

### Static Pickers

- `autoFocus: true` => focus the visible view
- `autoFocus: false` => don't focus anything inside the component

### DIFF
--- a/docs/pages/x/api/date-pickers/desktop-next-date-picker.json
+++ b/docs/pages/x/api/date-pickers/desktop-next-date-picker.json
@@ -1,5 +1,6 @@
 {
   "props": {
+    "autoFocus": { "type": { "name": "bool" } },
     "className": { "type": { "name": "string" } },
     "closeOnSelect": {
       "type": { "name": "bool" },
--- a/docs/pages/x/api/date-pickers/desktop-next-date-time-picker.json
+++ b/docs/pages/x/api/date-pickers/desktop-next-date-time-picker.json
@@ -2,6 +2,7 @@
   "props": {
     "ampm": { "type": { "name": "bool" }, "default": "`utils.is12HourCycleInCurrentLocale()`" },
     "ampmInClock": { "type": { "name": "bool" } },
+    "autoFocus": { "type": { "name": "bool" } },
     "className": { "type": { "name": "string" } },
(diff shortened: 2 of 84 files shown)


alexfauquette:
> (the autoFocus has no effect)

Did not think about it when refactoring the focus management, but it's weird. It would make more sense to have the focus set on the view only when the view get open by the user


I assume nobody uses the pickers open by default. In such case, the behavior is ok

Otherwise, we should find a way to move the focus to the view only when it's opened by user interaction

flaviendelangle:
I also think the opening a picker on mount in an edge case and almost nobody does it.
What would be the best behavior for you ? If `open = true` on mount we only focus the view if `autoFocus = true` and don't focus anything otherwise ?

The picker is a popper so it make sense to me to focus it when opened, no matter how it is opened.

alexfauquette (approved):
Ok for the current behavior. I did not catch any regression in the preview. But we have a bug (not caused by this PR) #7177

flaviendelangle:
My bad, fixed by #7179

PR #7154, opened 2022-12-09
[DateRangePickers] Add shortcuts component

alexfauquette:


packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
```diff
@@ -453,10 +454,18 @@ export const usePickerValue = <
     onSelectedSectionsChange: handleFieldSelectedSectionsChange,
   };
 
+  const isValid = (testedValue: TValue) =>
+    validator({
+      adapter,
+      value: testedValue,
+      props: { ...props, value: testedValue },
+    }) === null;
```
LukasTy, line 462: Or something similar to support range pickers.
```suggestion
  const isValid = (testedValue: TValue) => {
    const validationResponse = validator({
      adapter,
      value: testedValue,
      props: { ...props, value: testedValue },
    });
    return Array.isArray(testedValue)
      ? (validationResponse as any[]).every((v) => v === null)
      : validationResponse === null;
  };
```

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { DateOrTimeView } from '../internals';
+import { useUtils } from '../internals/hooks/useUtils';
+import { VIEW_HEIGHT } from '../internals/constants/dimensions';
+// import { useLocaleText } from '@mui/x-date-pickers/src/internals/hooks/useUtils';
+
+interface PickersShortcutsItemGetValueParams<TValue> {
+  value: TValue;
+  view: any;
+  isValid: (value: TValue) => boolean;
+  adapter: any;
```
LukasTy, line 14: We'll need to type it one way or another 🤔

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,73 @@
...
+      },
+      disabled: !isValid(newValue),
+    };
+  });
+
+  return (
+    <List {...other} dense sx={{ maxHeight: VIEW_HEIGHT, overflow: 'auto' }}>
+      {items.map((item) => {
+        return (
+          <ListItem key={item.label}>
```
LukasTy, line 66: We'd probably want to make this a slot, don't you think?
Users might not want to re-implement the whole shortcuts component if they only want to override how a single shortcut looks (what it renders).

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,73 @@
...
+      label: item.label,
+      onClick: () => {
+        onChange(newValue);
+      },
+      disabled: !isValid(newValue),
+    };
+  });
+
+  return (
+    <List {...other} dense sx={{ maxHeight: VIEW_HEIGHT, overflow: 'auto' }}>
```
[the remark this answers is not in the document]
flaviendelangle: We probably don't want to define a `sx` because it forces the users to re-define those styles even if they want to apply some unrelated ones.

A `styled()` component is probably the way to go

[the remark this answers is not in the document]
LukasTy: Hm... I'm not sure about it...
Maybe go the [with this suggestion](https://github.com/mui/mui-x/pull/5651#discussion_r952343529) instead? 🤔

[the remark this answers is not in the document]
flaviendelangle: Can work indded

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
alexfauquette: I'm not sure, because the component itself is small. There is no complex logic, so I find it more secure to add an example saying:

> Here is the shortcuts slot. You get access to all the API methods used to create your shortcuts: onChange, value, isValid.

Than allowing to override a `ListItem` as a black box (ignoring it get some context from its parents `List`)

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useUtils } from '../internals/hooks/useUtils';
+import { VIEW_HEIGHT } from '../internals/constants/dimensions';
+// import { useLocaleText } from '@mui/x-date-pickers/src/internals/hooks/useUtils';
+
+interface PickersShortcutsItemGetValueParams<TValue, TDate> {
+  value: TValue;
+  view: DateOrTimeView;
+  isValid: (value: TValue) => boolean;
+  adapter: MuiPickersAdapter<TDate>;
+}
```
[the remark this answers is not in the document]
flaviendelangle: I think that if we are unsure about the usefulness of a given property, it's better to not add it and see if people struggle without it.

It keeps the API as light and easy to understand as possible.

flaviendelangle:
image

Not sure how to improve it, but the scrollbar on Linux (and I guess WIndows) makes the shortcut UI looks a lot less appealing. 

____

```tsx
<StaticNextDateRangePicker
  componentsProps={{
    shortcuts: {
      shortcuts,
    },
  }}
/>
```

Naming a prop `shortcuts` in a slot named `shortcuts` can be disorienting.
Maybe we can find another name to avoid the duplication.

flaviendelangle:
> Not sure to get you're point. I added the page under the mysterious name "shortcuts" male_detective

Oh I missed it
But still, don't you think we should group all those pages talking about slots in a folder ?
For me it was the plan when we would have new pages and now we have 3.

LukasTy:
> Yes, what about `shortcuts.items` instead of `shortcuts.shortcuts`?

Much better version! 👍 

> But still, don't you think we should group all those pages talking about slots in a folder ?
For me it was the plan when we would have new pages and now we have 3.

Indeed, we now have quite a lot separate scattered pages about various complex customizations. Putting them all under one side menu parent item could help with discoverability. 🤔

alexfauquette:
> Putting them all under one side menu parent item could help with discover ability. 

I agree, I put them under a same group, but to be consistent, it might require some rewriting, to add hyperlinks between those 3 pages. I keep this rewriting for a follow up PR

docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -18,6 +18,7 @@ By default, pickers are made of 4 sub elements present in the following order:
 - The **content** displaying the current view. Can be a calendar, or a clock.
 - The **tabs** allowing to switch between day and time views in Date Time Pickers.
 - The **action bar** allowing some interactions. Can be added with [`componentsProps.actionBar`](/x/react-date-pickers/custom-components/#action-bar) prop.
+- The **shortcuts** allowing quick selection of some values. Can be added with [`componentsProps.shortcuts`](/x/react-date-pickers/shortcuts/#adding-shortcuts)
```
LukasTy, line 21:
```suggestion
- The **shortcuts** allowing quick selection of some values. Can be added with [`slotProps.shortcuts`](/x/react-date-pickers/shortcuts/#adding-shortcuts)
```

docs/data/date-pickers/shortcuts/AdvancedRangeShortcuts.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import dayjs, { Dayjs } from 'dayjs';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
+import { Unstable_StaticNextDateRangePicker as StaticNextDateRangePicker } from '@mui/x-date-pickers-pro/StaticNextDateRangePicker';
+import { PickersShortcutsItem } from '@mui/x-date-pickers/PickersShortcuts/PickersShortcuts';
```
LukasTy, line 6: This import could probably be improved somewhat. Maybe at least an index re-export? 🤔

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
+---
+product: date-pickers
+title: Date and Time pickers - Shortcuts
```
LukasTy, line 3:
```suggestion
title: Date and Time pickers - Shortcuts
components: PickersShortcuts
```
But it also should be added as a submenu link in `docs/data/pages`.

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+# Shortcuts
+
+<p class="description">The date picker lets you add custom shortcuts.</p>
+
+## Adding shortcuts
+
+By default, pickers use the `PickersShortcuts` component to display shortcuts.
+This component accepts a `shortcurs` prop as an array of `PickersShortcutsItem`.
```
LukasTy, line 13:
```suggestion
This component accepts a `shortcuts` prop as an array of `PickersShortcutsItem`.
```

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
...
+Those items are made of two required attributes:
+
+- `label`: The string displayed on the shortcut chip. This property must be unique.
+- `getValue`: A function that returns the value associated to the shortcut.
+
+You can use `componentsProps.shortcus` to customize this props. For example to add a shortcut to Christmas Day, you can do the following:
```
LukasTy, line 19:
```suggestion
You can use `slotProps.shortcuts` to customize this prop. For example to add a shortcut to Christmas Day, you can do the following:
```

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
...
+Shortcuts on range pickers require `getValue` property to return an array with two values.
+
+{{"demo": "BasicRangeShortcuts.js", "bg": "inline"}}
+
+## Advanced shortcuts
+
+The `getValue` methods receive a `isValid` helper function.
+You can use it to test if a value is valid or not based on the [validation props](/x/react-date-pickers/validation/).
+
+In the following demonstration, it is used to get the next available week and week-end.
```
LukasTy, line 58:
```suggestion
In the following demonstration, it is used to get the next available week and weekend.
```

docs/data/pages.ts
```diff
@@ -251,8 +251,15 @@ const pages: MuiPage[] = [
         ],
       },
       { pathname: '/x/react-date-pickers/fields', title: 'Field components', newFeature: true },
-      { pathname: '/x/react-date-pickers/custom-components', title: 'Custom subcomponents' },
-      { pathname: '/x/react-date-pickers/custom-layout', title: 'Custom layout' },
+      {
+        pathname: '/x/react-date-pickers/visual-customization',
+        title: 'Visual customization',
```
LukasTy, line 256: I'm not sure about this title. 🤔 
The possible customizations do not have to be visual 🤔 
Maybe just go with?
```suggestion
        title: 'Customization',
```

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+function PickersShortcuts<TValue>(props: PickersShortcutsProps<TValue>) {
+  const { items, isLandscape, onChange, isValid, ...other } = props;
+
+  if (items == null || items.length === 0) {
+    return null;
+  }
+
+  const itemsAttributs = items.map((item) => {
```
LukasTy, line 36:
```suggestion
  const resolvedItems = items.map((item) => {
```

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <List
+      dense
+      sx={[
+        {
+          maxHeight: VIEW_HEIGHT,
+          maxWidth: 200,
+          overflow: 'hidden',
+          '&:hover': { overflow: 'auto' },
```
LukasTy, line 56: This solution doesn't work on mobile. 🙈

LukasTy (commented):
Looks really great ❤️ 
It would be really great if the layout wouldn't need to jump (increase in width) when the shortcuts are hovered. 🙈 

https://user-images.githubusercontent.com/4941090/212776796-d2535e26-7a72-457e-a2b2-bdb9dc8ed5f7.mov

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
...
+By default, the shortcuts are disabled if the returned value does not pass validation.
+Here is an example where `minDate` is set to the middle of the year.
+
+{{"demo": "DisabledDatesShortcuts.js", "bg": "inline"}}
+
+## Range shortcuts [<span class="plan-pro"></span>](/x/introduction/licensing/#pro-plan)
+
+Shortcuts on range pickers require `getValue` property to return an array with two values.
+
+{{"demo": "BasicRangeShortcuts.js", "bg": "inline"}}
```
joserodolfofreitas, line 51: The "short version" of this demo code doesn't show getValue with two values. It's not a big deal, but it may make sense to display the relevant code regarding this difference in setup.

image

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Advanced shortcuts
+
+The `getValue` methods receive a `isValid` helper function.
+You can use it to test if a value is valid or not based on the [validation props](/x/react-date-pickers/validation/).
+
+In the following demonstration, it is used to get the next available week and week-end.
+
+{{"demo": "AdvancedRangeShortcuts.js", "bg": "inline"}}
```
joserodolfofreitas, line 60: Same here; perhaps the short version of the code should focus on `getValue: ({ isValid }) => { ... }`

joserodolfofreitas (commented):
Using the shortcuts feels quite good, and bonus points for the interesting shortcuts examples in the docs. Well done, @alexfauquette!

I'm not sure, though, if the docs should be under visual customizations (only).
First, because I think it's more than "simply" visual customization. It impacts behavior and UX the most, so it may not be intuitive to go there to find the page.

Furthermore, I understand it can be used with the regular date picker, but I think we should have it under a section* in the date range page as well, which I believe is likely to hold most use cases for the shortcuts. 
*/x/react-date-pickers/date-range-picker/#pre-defined-range-shortcuts 
(at least a mini section pointing to the main page).

It may also be interesting to have an example using the custom layout to show the shortcuts on the bottom or right side.

joserodolfofreitas:
We can release the current version for access in the earlier beta and work on the scrollbar as a [separate issue](https://github.com/mui/mui-x/issues/7672).

docs/data/date-pickers/shortcuts/shortcuts.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Customization
+
+Like other [layout's subcomponent](/x/react-date-pickers/custom-layout/), the shortcuts can be customized.
+Here is an example with an horizontal shortcuts.
```
LukasTy, line 65:
```suggestion
Here is an example with horizontal shortcuts.
```

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
LukasTy: @alexfauquette 
I've experimented with a solution, which would be supported in mobile and I think that with only a tiny tweak it does work.
Could you have a look [at this example](https://codesandbox.io/s/confident-leftpad-htposs?file=/demo.tsx)? 🤔

docs/data/pages.ts
alexfauquette: Customization can also be about the behavior

LukasTy: Fair point. Most of the listed examples in the those pages are visual customizations. 👌

packages/x-date-pickers/src/PickersShortcuts/PickersShortcuts.tsx
alexfauquette: Amazing :)

I moved it into the follow-up issue, such that we can merge as it is, and add improvement after having time to think about them

PR #7156, opened 2022-12-09
Update DateTimePicker export pattern to match

kealjones-wk:
DateTimePickers export pattern doesn't match how all the other components are exported. This fixes that, see [DatePicker for ref](https://github.com/mui/mui-x/blob/next/packages/x-date-pickers/src/DatePicker/DatePicker.tsx#L62). 

This change fixes an issue with [react-docgen-typescript](https://www.npmjs.com/package/react-docgen-typescript) being able to locate the component.

Signed-off-by: Keal Jones <41018730+kealjones-wk@users.noreply.github.com>


LukasTy:
@kealjones-wk  Thank you for your contribution! 🙏 
Could you explain in more detail, what is the root cause you are trying to solve with this change?
From the looks of it, we probably have more components with exporting at the bottom of the file.

kealjones-wk:
@LukasTy So sorry, i meant for this to commit into master... ugh

PR #7176, opened 2022-12-13
[pickers] Allow to define custom view renderers on the pickers

flaviendelangle:
[Working example](https://codesandbox.io/s/date-and-time-pickers-forked-xclwey)

This PR adds a new prop `viewRenderers` on every v6 picker to override the renderer of some view.
The main use case for now is to allow people to bring back the clock for time editing to desktop if wanted.

Here is how people could bring back the clock on every desktop picker with time editing:

```tsx
const timeViewRenderers = { 
  hours: renderTimeViewClock, 
  minutes: renderTimeViewClock, 
  seconds: renderTimeViewClock
};

const theme = createTheme({
  components: {
    MuiDesktopNextTimePicker: {
      defaultProps: {
        viewRenderers: timeViewRenderers,
      },
    },
    MuiDesktopNextDateTimePicker: {
      defaultProps: {
        viewRenderers: timeViewRenderers,
      },
    },
  },
});
```


packages/x-date-pickers/src/NextDatePicker/shared.tsx
```diff
@@ -48,6 +47,14 @@ export interface BaseNextDatePickerProps<TDate>
    * @default {}
    */
   componentsProps?: BaseNextDatePickerSlotsComponentsProps<TDate>;
+  /**
+   * Define custom view renderers for each section.
+   * If `null`, the view will be editing with the field.
+   * If `undefined`, the view will be the one defined internally.
```
LukasTy, line 53: Same comment changes would apply here as well. 😉

packages/x-date-pickers/src/NextDateTimePicker/shared.tsx
```diff
@@ -90,6 +90,19 @@ export interface BaseNextDateTimePickerProps<TDate>
    * @default {}
    */
   componentsProps?: BaseNextDateTimePickerSlotsComponentsProps<TDate>;
+  /**
+   * Define custom view renderers for each section.
+   * If `null`, the view will be editing with the field.
```
LukasTy, line 95: And here as well 🙈

packages/x-date-pickers/src/NextTimePicker/shared.tsx
```diff
@@ -51,6 +50,14 @@ export interface BaseNextTimePickerProps<TDate>
    * @default {}
    */
   componentsProps?: BaseNextTimePickerSlotsComponentsProps;
+  /**
+   * Define custom view renderers for each section.
+   * If `null`, the view will be editing with the field.
```
LukasTy, line 55: Likewise. 🤔

packages/x-date-pickers/src/internals/hooks/usePicker/usePicker.ts
```diff
@@ -32,11 +31,9 @@ export const usePicker = <
     validator,
   });
 
-  const pickerViewsResponse = usePickerViews({
+  const pickerViewsResponse = usePickerViews<TValue, TView, TExternalProps, TAdditionalProps>({
     props,
     inputRef,
-    viewLookup,
-    wrapperVariant,
```
LukasTy, line 39: Was it here by mistake?
Or did you remove it by mistake? 🤔

packages/x-date-pickers/src/internals/models/props/basePickerProps.tsx
```diff
@@ -41,10 +43,16 @@ export interface BasePickerProps<TValue, TDate> extends PickerStateProps<TValue>
 }
 
 /**
- * Props common to all pickers.
+ * Props common to all pickers after the applying the default props on each picker.
```
LukasTy, line 46:
```suggestion
 * Props common to all pickers after applying the default props on each picker.
```

packages/x-date-pickers/src/internals/models/props/basePickerProps.tsx
```diff
@@ -65,6 +73,15 @@ export interface BaseNextPickerProps<TValue, TDate, TView extends DateOrTimeView
   localeText?: PickersInputComponentLocaleText<TDate>;
 }
 
+/**
+ * Props common to all pickers after the applying the default props on each picker.
```
LukasTy, line 77:
```suggestion
 * Props common to all pickers after applying the default props on each picker.
```

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
```diff
@@ -48,6 +69,11 @@ export interface UsePickerViewsBaseProps<TView extends DateOrTimeView> {
    * Array of views to show.
    */
   views: readonly TView[];
+  /**
+   * Define custom view renderers for each section.
+   * If `null`, the view will be editing with the field.
```
LukasTy, line 74: Is this comment used anywhere?
Each picker type seems to override it and provide it's own comment with one extra line.

LukasTy (commented):
This is a really nice API improvement 🚀 
We just need to mention this possibility [in docs where we mention removal of the clock](https://next.mui.com/x/migration/migration-pickers-v5/#drop-clock-in-desktop-mode). 😉 

P.S. Maybe we could adjust types so that something like this would error out? 🤔 
Screenshot 2022-12-16 at 12 27 27

packages/x-date-pickers/src/internals/hooks/usePicker/usePicker.ts
flaviendelangle: `wrapperVariant` was not used, it was probably at some point during the development of the new pickers

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
flaviendelangle: I don't think it is used anywhere in the JSDoc
It's just here to describe what the param is doing in `usePickerViews`

flaviendelangle:
@LukasTy concerning the error message, I think it's how TypeScript work.
I would like an error indeed, but not sure if we can do it without type:

#### External variable without an explicit type => no error

image

#### External variable with an explicity type => error

image


____

I did a small typing test to confirm what I was thinking
You can see that TypeScript does not complain about the `c` property when passed to `funcExpectingParent`

image

But if I explicitely say `child is of type Child`, then TypeScript complains

image

And this flexibility make sense, most of the time, it allows us to pass objects with additional props to a function (for example all the functions with a `FieldSection` param that can access the range sections without complaining.

___

Maybe in the doc we should avoid to define the renderers in a standalone variable, that way we have a correct error:

image

LukasTy:
> Maybe in the doc we should avoid to define the renderers in a standalone variable, that way we have a correct error:

Fair point. Thanks for clarifying this. It was an oversight from my point as well.
If we were to provide an example of how to do it in migration guide—I'd go with no variable just to have type safety. 👍

LukasTy (approved):
Great job! 🙌

PR #7181, opened 2022-12-13
Fix gridcoltype autocomplete

### LINKED ISSUE #7173: [data grid] Fix typescript autocomplete for GridColType on ColDef type on DataGrid
The following code has an issue where the final `| string` essentially disables autocomplete and type-checking when using the Grid out of the box. My understanding is it's an open type so we can extend the datagrid with our own column types, but it's possible to have both.

```ts
export type GridNativeColTypes =
  | 'string'
  | 'number'
  | 'date'
  | 'dateTime'
  | 'boolean'
  | 'singleSelect'
  | 'actions';

export type GridColType = GridNativeColTypes | string;
```
https://github.com/mui/mui-x/blob/3e7bfb2a3b11d3728892b245e70928c7669a0bd6/packages/grid/x-data-grid/src/models/colDef/gridColType.ts#L10

This can be improved by doing the following:

```ts 
export type GridColType = GridNativeColTypes | (string & Record<never, never>);
```

A detailed explanation of the typescript limitation and workaround can be found in the type-fest package: https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts

With this change free-text is permitted, but auto-complete suggestions will properly document the library out of the box.

Nick-Lucas:
Fixes #7173

This PR enables editor autocomplete for gridColType which should improve the developer experience of defining columns

Just an initial draft for now as there are a few ideas I'm bouncing around, small stuff but will leave some comments with my reasoning

### DIFF
--- a/packages/grid/x-data-grid/src/models/colDef/gridColType.ts
+++ b/packages/grid/x-data-grid/src/models/colDef/gridColType.ts
@@ -15,7 +15,10 @@ export type GridNativeColTypes =
  */
 type LiteralAnyString = string & Record<never, never>
 
+/**
+ * Immitate typescript's template literal syntax to document in autocomplete that the string type is open for extension.
+ */
 // eslint-disable-next-line no-template-curly-in-string
-type LiteralAnyStringPrompt = '${string}'
+type LiteralAnyStringDocPrompt = '${string}'
 
-export type GridColType = GridNativeColTypes | LiteralAnyStringPrompt | LiteralAnyString;
+export type GridColType = GridNativeColTypes | LiteralAnyStringDocPrompt | LiteralAnyString;
--- a/packages/grid/x-data-grid/src/models/colDef/gridColType.ts
+++ b/packages/grid/x-data-grid/src/models/colDef/gridColType.ts
@@ -7,4 +7,15 @@ export type GridNativeColTypes =
   | 'singleSelect'
   | 'actions';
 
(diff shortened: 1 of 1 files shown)


packages/grid/x-data-grid/src/models/colDef/gridColType.ts
```diff
@@ -7,4 +7,18 @@ export type GridNativeColTypes =
...
+ * Can be removed if this issue is ever resolved:
+ * https://github.com/Microsoft/TypeScript/issues/29729
+ */
+type LiteralAnyString = string & Record<never, never>
+
+/**
+ * Immitate typescript's template literal syntax to document in autocomplete that the string type is open for extension.
+ */
+// eslint-disable-next-line no-template-curly-in-string
+type LiteralAnyStringDocPrompt = '${string}'
```
Nick-Lucas, line 22: So this is a bit of an experiment. 

* When seeing autocomplete in typescript, it can appear to developers that you're restricted from using any value which deviates from the listed values
* Typescript has established syntax in template literals for literals with some variable/free-text element
* Typescript does not have an established approach for documenting when literals and template-literals permit "any string" as a value, it simply disabled autocomplete without workarounds like implemented here.

So my idea is to list the options with a template literal shown to help communicate that other values are permitted:

image

Nick-Lucas:
@m4theushw first crack here, it's thankfully pretty small, and shouldn't contain any breaking changes as far as I know!

PR #7184, opened 2022-12-13
[fields] Support escaped characters

### LINKED ISSUE #6368: [pickers] Support escaped characters in the format analysis method (`splitFormatIntoSections`)
### Duplicates

### Latest version

### Summary 💡

Based on https://github.com/mui/mui-x/pull/6337#discussion_r985671283

The adapter should support escaped characters.
For instance `[The] d [of] MM`.

DayJS and Moment uses the `[ ]` syntax to escape characters.
Date-fns, Luxon uses the `' '` syntax to escape characters

### Examples 🌈

_No response_

### Motivation 🔦

_No response_

### Order ID 💳 (optional)

_No response_

flaviendelangle:
Fixes #6368

## Change in the parsing logic

I did a major change in the parsing logic.
Now to go from the sections, to a parsed date, I am applying the parsing on a format without the separators.

For example, if the format is `DD/MM`, we have 2 sections `[{ formatValue: 'DD', separator: '/' }, { formatValue: 'MM' }]`.
Before, we were doing:

```ts
const dateStr = createDateStrFromSections(activeDateSections, false)
const date = utils.parse(dateStr, format);

// utils.parse('25/05', 'DD/MM') in the current example
```

Now I am doing:

```ts
const formatWithoutSeparator = sections.map((section) => section.formatValue).join(' ');
const dateWithoutSeparatorStr = sections
  .map((section) => getSectionVisibleValue(section, false))
  .join(' ');
const date = utils.parse(dateWithoutSeparatorStr, formatWithoutSeparator);

// utils.parse('25 05', 'DD MM') in the current example
```

Why ? Because date libraries are not able to parse format with escaped characters, those escaped characters are only for formatting.

I did not encounter any regression, but that's an important change and I will try to play with it a little more.
Having the separator only used for the rendering and never for the parsing has two interesting benefits outside from this PR:

1. We no longer need to store the `parsingSeparator` to keep trace of the  ` / ` separator without the spaces
2. `createDateStrFromSections` no longer need to be able to generate formats that are not for the input. So it can always apply the RTL for example.

Small side note concerning formats like `PP` (which are expanded using the `utils.expandFormat` method)
Before we were parsing using the non-expanded format since we directly used the `format` prop.
Now we are parsing using the expanded format since we are rebuilding a format from the sections.
This should not have any effect, unless our `utils.expandFormat` method is broken (but it should break on other places before this PR)


packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -387,64 +415,123 @@ export const splitFormatIntoSections = <TDate>(
...
-    if (char.match(/([A-zÀ-ú]+)/g)) {
+    if (escapedPartOfCurrentChar == null && char.match(/([A-Za-z]+)/g)) {
       currentTokenValue += char;
-    } else {
+    }
+    // If we are on the starting of the ending character of an escaped part of the format,
+    // Then we ignore this character.
+    else if (
+      escapedPartOfCurrentChar == null ||
+      (escapedPartOfCurrentChar.start !== i && escapedPartOfCurrentChar.end !== i)
```
alexfauquette, line 431: It's a bit hard to get this logic expression. Maybe would be easier with dedicated varaibles such as 

```js
const isEscapedChar = escapedPartOfCurrentChar != null
// We ignore open/close token of escaped sections
const isIgnoredChar = isEscapedChar && (escapedPartOfCurrentChar.start === i || escapedPartOfCurrentChar.end === i)

// ...

else if (!isIgnoredChar){
  // ..
}
```

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -387,64 +415,123 @@ export const splitFormatIntoSections = <TDate>(
...
+        cleanedUpString += char;
+      }
+    }
+
+    return cleanedUpString;
+  };
+
+  const expandedFormat = utils.expandFormat(format);
+  const escapedPartsInFormat = getEscapedPartsFromExpandedFormatOrDateStr(utils, expandedFormat);
+  const escapedPartsInDateStr = getEscapedPartsFromExpandedFormatOrDateStr(utils, dateStr);
```
alexfauquette, line 492: If I correctly understood the logic of `getEscapedPartsFromExpandedFormatOrDateStr` it's looking for escape token, why would the `dateStr` contain such token?

If I use `[The] 6th [of] December`, I expect the date string to be `The 6th of December` so not token to consider

alexfauquette (commented):
Just left a few implementation comments. I will have to read it a second time to get a clear view of how data are manipulated and answer your questions.

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: > If I use [The] 6th [of] December, I expect the date string to be The 6th of December so not token to consider

To be displayed in the input yes
To be parsed not
Because as explained in the PR description,  `dayjs` (and maybe others) don't handle parsing of formats with escaped characters. So before the parsing from the string, I need to remove the escaped characters from both the format and the stringified date.

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
```diff
@@ -290,6 +291,14 @@ export interface UseFieldState<TValue, TSection extends FieldSection> {
    */
   referenceValue: TValue;
   sections: TSection[];
+
+  /**
+   * The separator to insert before the first section.
+   * If it contains escaped characters, then it must not have the escaping characters.
+   * For example, on Day.js, the format `[Current year:] YYYY` has a `startSeparator` equal to `Current year:` not `[Current year:]`
+   */
+  startSeparator: string;
```
flaviendelangle, line 300: The alternative would be to put a `startSeparator` / `endSeparator` on each section (to replace the `separator` which is always at the end)

I did not wanted to do that because only the 1st section is impacted.
But with single input range fields, it becomes more complicated since the 1st section of the 2nd date has the separator as well.
So maybe, the tradeoff is not that good...

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
alexfauquette: If the user past some text, it will be "The 6th of December", not the "[The] 6th [of] December" and we have 

```js
getEscapedPartsFromExpandedFormatOrDateStr("The 6th of December") -> []

getEscapedPartsFromExpandedFormatOrDateStr("[The] 6th [of] December") -> [{start: 0, end: 4}, {start: 10, end: 13}]
```

-----

I'm not sure that copy/past needs to support the escaped format, because you do not copy an element from an input to past it in the same input. data will probably come from external input for which we have no clue about the format used.

Would it make sense to use the default format for parsing the past text, and have a distinct way of customizing this format.

For example, I make an internal app, and intarnal tools sometime use `dd-mm` or `dd/mm/yyyy`, As a developper I would like to pass a prop to `dayJs` such that the past date is managed as follow:

```js
dayjs(inputValue, ["dd-mm", "dd/mm/yyyy"], 'fr', true);
```

flaviendelangle: > If the user past some text, it will be "The 6th of December", not the "[The] 6th [of] December"

Right, I mixed up several things... 

____

> Would it make sense to use the default format for parsing the past text, and have a distinct way of customizing this format.

For the parsing, probably yes, to have something like Retool has where we test a list of formats, if possible first the format used in the field and then commonly used formats.
But that would be for later.

However, the logic we are talking about here also impacts imperative update of the field, to enable users to test there application with `fireEvents` without caring about the implementation detail of the field, just by firing an `onChange` with the new value.
And for this use case, they will probably match the format expected in the field.

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -387,64 +418,95 @@ export const splitFormatIntoSections = <TDate>(
...
-    // wrap into an isolated group such that separators can split the string in smaller ones by adding \u2069\u2068
-    return `\u2066${formattedArray.join('')}\u2069`;
-  }
-  return formattedArray.join('');
+  // \u2066: start left-to-right isolation
+  // \u2067: start right-to-left isolation
+  // \u2068: start first strong character isolation
+  // \u2069: pop isolation
+  // wrap into an isolated group such that separators can split the string in smaller ones by adding \u2069\u2068
+  return `\u2066${formattedArray.join('')}\u2069`;
```
alexfauquette, line 509: If we only generate for input, those can be added in the `startSeparator`, and the last `separator`

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -351,15 +355,39 @@ const getSectionPlaceholder = <TDate>(
   }
 };
 
+const getEscapedPartsFromExpandedFormatOrDateStr = <TDate>(
+  utils: MuiPickersAdapter<TDate>,
+  expandedFormat: string,
+) => {
+  let isInsideEscapedPart = false;
```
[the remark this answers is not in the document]
alexfauquette: Might be more buggy, but this RegExp can also do the job 🙈
`/(\[[^\]]*\])+/g`

https://regex101.com/r/uRJrCp/1

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
alexfauquette: > However, the logic we are talking about here also impacts imperative update of the field, to enable users to test there application with fireEvents without caring about the implementation detail of the field, just by firing an onChange with the new value.

The `onChange` first checks if the modified value impacts only the selected section and if yes it only parses this section. So this should not be a problem.

If the modification goes outside of the selected section, then we will have the full parsing which can be problematic.

The only solution I see is to build a regexp (or a custom function) to do the cleaning

```js
const cleanValue = (inputValue, sections) => {
	let cleanedValue = inputValue
	sections.forEach(({startSeparator, separator}) => {
		cleanedValue = cleanedValue.replace(startSeparator, ' ')
		cleanedValue = cleanedValue.replace(separator, ' ')
	})
	// remove duplicate spaces
	// ...
	return cleanedValue
}
```

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
flaviendelangle: Would it work for scenario like `[Escaped []` ? (should make `Escaped [` as escaped)
The date libraries (at least `dayjs`) supports this weird edge case and I'd like to support it as well

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: True
But for the single range input, I guess we would need to remove them from the last section of the `start` date and the first section of the `last` date, which could be more work and complexity.

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: In the `onChange` I have an early check saying that, if no section is selected, then it's probably an imperative update, in which case I re-parse the whole string just like for pasting (in its current form).

The selection will often go outside of the current form, because I don't expect people to respect all the RTL / hidden spaces characters.
But the current approach works fine, if no section is selected, then parse the string based on the format.
I thought it was more complicated than that because I was trying to parse `[The] 6th [of] January`, but parsing `The 6th of January` works great.

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
alexfauquette: The other option would be to put the `startSeparator` of the second date at the end of the first date.

But having a prefix/suffix for all section even if most of the time they are empty sounds more reliable.

It will also avoid those `getSectionsFromValue(...).sections`

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
```diff
@@ -173,13 +174,22 @@ interface FieldActiveDateManager<TValue, TDate> {
...
+  shouldSelectBoundarySelectors?: boolean;
+};
 
 export type FieldSelectedSections =
   | number
   | FieldSelectedSectionsIndexes
   | MuiDateSectionName
-  | null;
+  | null
+  | 'all';
```
flaviendelangle, line 192: I try to select the separators that are at the very beginning and end of the input only when we do Ctrl + A, or focus.

My initial check was `selectedSectionIndexes.startIndex === 0 && selectedSectionIndexes.endIndex === state.sections.length - 1`
But it does not work on single section format.

If you have a better idea, I'm interested.
It feels weird if we don't select them when doing Ctrl + A
It feels weird if we select them when clicking on the section in a single section format

flaviendelangle: https://user-images.githubusercontent.com/3309670/207849408-474f1545-7378-4d2c-980d-005bfe501d3f.mp4

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
alexfauquette: I think yes, because in bot case, you search for the first opening token, and then the next closing token and don't care about what is between

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
alexfauquette: Good point

packages/x-date-pickers/src/internals/hooks/useField/useField.interfaces.ts
alexfauquette: I don't get what's going wrong in your recording. For me the following are correct behavior

- click on before/after put the select on the associated placeholder
- ctrl+a select the entire value

alexfauquette: Introducing `all` does not look like a problem


Instead of `shouldSelectBoundarySelectors: true` you could have `includesStartSparetor` and `includesEndSparetor` for more flexibility.

The only limitation I see is selecting only the first date, because I guess the current strategy allows to select:
- `YYYY`
- `Before YYYY After - ` 
- but not `Before YYYY After`

alexfauquette: I thought about the empty section, the keyboard can be managed by modifying the `getSectionOrder`, it would also require to modify the function that select a section on click to ignore it.

Another solution could be to remove it from sections, and only add its content when generating the input value from sections, and computing sections positions.

But both sound too hacky

flaviendelangle: > Another solution could be to remove it from sections, and only add its content when generating the input value from sections, and computing sections positions.

Could be the best solution
We can experiment later on

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
flaviendelangle:
```tsx
const escapedParts: { start: number; end: number }[] = [];
const { start: startChar, end: endChar } = utils.escapedCharacters;
const regExp = new RegExp(`(\\${startChar}[^\\${endChar}]*\\${startChar})+`, 'g');

let match: RegExpExecArray | null = null;
// eslint-disable-next-line no-cond-assign
while ((match = regExp.exec(expandedFormat))) {
  escapedParts.push({ start: match.index, end: regExp.lastIndex - 1 });
}

return escapedParts;
```

The RegExp seem to work well

packages/x-date-pickers-pro/src/internal/utils/valueManagers.ts
```diff
@@ -63,55 +63,48 @@ export const rangeFieldValueManager: FieldValueManager<
...
-
-    const rawSectionsOfStartDate = getSections(start, prevDateRangeSections.startDate);
-    const rawSectionsOfEndDate = getSections(end, prevDateRangeSections.endDate);
-
-    const sectionsOfStartDate: RangeFieldSectionWithoutPosition[] = rawSectionsOfStartDate.map(
-      (section, sectionIndex) => {
-        if (sectionIndex === rawSectionsOfStartDate.length - 1) {
+      const sections = splitFormatIntoSections(utils, localeText, format, newDate);
+      return sections.map((section, sectionIndex) => {
+        if (sectionIndex === sections.length - 1) {
```
alexfauquette, line 79: If we do not have `position === 'start'` it woudl have the same behavior as the default `return`. So it's easier to read when all conditions are in the same `if`
```suggestion
        if (sectionIndex === sections.length - 1 && position === 'start') {
```

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -351,15 +351,34 @@ const getSectionPlaceholder = <TDate>(
   }
 };
 
+const getEscapedPartsFromExpandedFormat = <TDate>(
+  utils: MuiPickersAdapter<TDate>,
+  expandedFormat: string,
+) => {
+  const escapedParts: { start: number; end: number }[] = [];
+  const { start: startChar, end: endChar } = utils.escapedCharacters;
+  const regExp = new RegExp(`(\\${startChar}[^\\${endChar}]*\\${startChar})+`, 'g');
```
alexfauquette, line 360: tests are passing, but the behavior does not work. I get the following message with this simple example:

```jsx
<DateField format="MM [myLitleToken<3] YYYY" />
```

> Error: MUI: The token "myLitleToken<3" is not supported by the Date and Time Pickers.

Here is the fix
```suggestion
  const regExp = new RegExp(`(\\${startChar}[^\\${endChar}]*\\${endChar})+`, 'g');
```

The reason why tests are passing is that tests run with an adapter using `'` as escaping characters, so open/close characters are the same

alexfauquette (commented):
Sounds solid 👍

Just two small comments, one for clarity, the second to fix a typo

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: Thanks for the catch ! :see_no_evil:

packages/x-date-pickers-pro/src/internal/utils/valueManagers.ts
```diff
@@ -63,55 +63,48 @@ export const rangeFieldValueManager: FieldValueManager<
...
             ...section,
-            dateName: 'start',
-            separator: `\u2069 – \u2066`,
-            parsingSeparator: ' – ',
+            dateName: position,
+            endSeparator:
+              position === 'start'
+                ? `${section.endSeparator}\u2069 – \u2066`
+                : section.endSeparator,
           };
```
alexfauquette, line 87:
```suggestion
            endSeparator: `${section.endSeparator}\u2069 – \u2066`
          };
```

alexfauquette (approved):

PR #7189, opened 2022-12-13
[field] Fix bug introduced by RTL in single input range fields

alexfauquette:
TODO:

- [x] Add test
- [x] Verify keyboard interaction in RTL display


packages/x-date-pickers-pro/src/internal/models/fields.ts
```diff
@@ -3,7 +3,16 @@ import { SlotComponentProps } from '@mui/base/utils';
 import TextField, { TextFieldProps } from '@mui/material/TextField';
 import Stack, { StackProps } from '@mui/material/Stack';
 import Typography, { TypographyProps } from '@mui/material/Typography';
-import { BaseFieldProps } from '@mui/x-date-pickers/internals';
+import { BaseFieldProps, FieldSection } from '@mui/x-date-pickers/internals';
+
+export interface RangeFieldSection extends FieldSection {
```
flaviendelangle, line 8: You can probably drop `DateRangeFieldSection` in favor of this `RangeFieldSection`

packages/x-date-pickers-pro/src/SingleInputDateRangeField/tests/SingleInputDateRangeField.keyboard.test.tsx
```diff
@@ -0,0 +1,171 @@
+import * as React from 'react';
```
flaviendelangle, line 1: The test files are usually named `{topic}.{componentName}.test.tsx`
It is the nomenclature on the data grid and on the new pickers test files.

I would name this one `selection.SingleInputDateRangeField.test.tsx`

packages/x-date-pickers-pro/src/internal/utils/valueManagers.ts
```diff
@@ -75,34 +76,43 @@ export const rangeFieldValueManager: FieldValueManager<
...
 
-    const sectionsOfStartDate = rawSectionsOfStartDate.map((section, sectionIndex) => {
-      if (sectionIndex === rawSectionsOfStartDate.length - 1) {
+    const sectionsOfStartDate: RangeFieldSectionWithoutPosition[] = rawSectionsOfStartDate.map(
+      (section, sectionIndex) => {
+        if (sectionIndex === rawSectionsOfStartDate.length - 1) {
+          return {
+            ...section,
+            dateName: 'start',
+            separator: `\u2069${' – '}\u2066`,
```
flaviendelangle, line 85: Why do we need to use template string here ?
Is there a difference with this ?

```suggestion
            separator: '\u2069 – \u2066',
```

packages/x-date-pickers-pro/src/SingleInputDateRangeField/tests/SingleInputDateRangeField.keyboard.test.tsx
alexfauquette: For now, all the `x-pickers-pro` are like that. I will rename all of them in a follow up PR

packages/x-date-pickers-pro/src/internal/utils/valueManagers.ts
alexfauquette: Just the result of iterating on different solutions 🙈

packages/x-date-pickers-pro/src/SingleInputDateRangeField/tests/SingleInputDateRangeField.keyboard.test.tsx
flaviendelangle: @LukasTy is creating new test files using the correct nomenclature in #7135
Can be done in a follow up PR of course

flaviendelangle (approved):
Great catch !
We definitely need more tests on the single input range fields, as they are not being played with manually a lot, it's easy to break something and see it weeks later

PR #7212, opened 2022-12-15
[docs] Allow to collapse items in the DemoContainer [WIP]

flaviendelangle:


LukasTy:
As mentioned [here](https://github.com/mui/mui-x/pull/7053#discussion_r1049779032):

> I'd just change the button label to text in order to identify what will happen when it's clicked.
Like `More`, `More examples`, or `Show more`.

flaviendelangle:
I should be able to start working on this one again soon :partying_face:

flaviendelangle:
Closing it because it seems to be in contradiction with our current attempt to limit the magic done by `DemoContainer`
Can re-open if we reconsider later on.

PR #7225, opened 2022-12-16
[fields] Fix multi input fields root element props order and types

LukasTy:
Created from https://github.com/mui/mui-x/pull/7135#discussion_r1050759913

- Fixes Multi Input Range fields root element
  - props to extend `StackProps`
  - spread `{...props}` at the end to allow for our defined prop value overriding
- Update `generateProptypes` to generate more correct final `PropTypes`
  - skip `classes` & `children` from inherited types
  - patch internal code to generate readable and simple `input.type` prop type
- Adjust component props removing illogical props


patches/@mui+monorepo+5.10.16.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/node_modules/@mui/monorepo/packages/typescript-to-proptypes/src/generator.ts b/node_modules/@mui/monorepo/packages/typescript-to-proptypes/src/generator.ts
+index 2e243512..6314afe4 100644
+--- a/node_modules/@mui/monorepo/packages/typescript-to-proptypes/src/generator.ts
++++ b/node_modules/@mui/monorepo/packages/typescript-to-proptypes/src/generator.ts
+@@ -224,6 +224,17 @@ export function generate(component: t.Component, options: GenerateOptions = {}):
+         (type): type is t.LiteralType => type.type === 'LiteralNode',
+       );
+ 
++      if (context.propTypeDefinition.name === 'type') {
```
LukasTy, line 9: I'm open to suggestions on how to tackle this one, because it also might be not stable (could no longer work if we update `mui/monorepo`).
Introducing this or similar change in base repo COULD be an option, but on the other hand, they technically don't care about such change. 🙈 🤷

LukasTy: It may not be the most ideal fix as it will not allow any unknown string as `type`, which is the reason for `(string & {})` exists... 🤔 
This change avoid the following type madness due to the `(string & {})` hack:
**Before:**
```js
type: PropTypes.oneOfType([
  PropTypes.oneOf([
    'button',
    'checkbox',
    'color',
    'date',
    'datetime-local',
    'email',
    'file',
    'hidden',
    'image',
    'month',
    'number',
    'password',
    'radio',
    'range',
    'reset',
    'search',
    'submit',
    'tel',
    'text',
    'time',
    'url',
    'week',
  ]),
  PropTypes.shape({
    '__@iterator@17874': PropTypes.func.isRequired,
    anchor: PropTypes.func.isRequired,
    at: PropTypes.func.isRequired,
    big: PropTypes.func.isRequired,
    blink: PropTypes.func.isRequired,
    bold: PropTypes.func.isRequired,
    charAt: PropTypes.func.isRequired,
    charCodeAt: PropTypes.func.isRequired,
    codePointAt: PropTypes.func.isRequired,
    concat: PropTypes.func.isRequired,
    endsWith: PropTypes.func.isRequired,
    fixed: PropTypes.func.isRequired,
    fontcolor: PropTypes.func.isRequired,
    fontsize: PropTypes.func.isRequired,
    includes: PropTypes.func.isRequired,
    indexOf: PropTypes.func.isRequired,
    italics: PropTypes.func.isRequired,
    lastIndexOf: PropTypes.func.isRequired,
    length: PropTypes.number.isRequired,
    link: PropTypes.func.isRequired,
    localeCompare: PropTypes.func.isRequired,
    match: PropTypes.func.isRequired,
    matchAll: PropTypes.func.isRequired,
    normalize: PropTypes.func.isRequired,
    padEnd: PropTypes.func.isRequired,
    padStart: PropTypes.func.isRequired,
    repeat: PropTypes.func.isRequired,
    replace: PropTypes.func.isRequired,
    replaceAll: PropTypes.func.isRequired,
    search: PropTypes.func.isRequired,
    slice: PropTypes.func.isRequired,
    small: PropTypes.func.isRequired,
    split: PropTypes.func.isRequired,
    startsWith: PropTypes.func.isRequired,
    strike: PropTypes.func.isRequired,
    sub: PropTypes.func.isRequired,
    substr: PropTypes.func.isRequired,
    substring: PropTypes.func.isRequired,
    sup: PropTypes.func.isRequired,
    toLocaleLowerCase: PropTypes.func.isRequired,
    toLocaleUpperCase: PropTypes.func.isRequired,
    toLowerCase: PropTypes.func.isRequired,
    toString: PropTypes.func.isRequired,
    toUpperCase: PropTypes.func.isRequired,
    trim: PropTypes.func.isRequired,
    trimEnd: PropTypes.func.isRequired,
    trimLeft: PropTypes.func.isRequired,
    trimRight: PropTypes.func.isRequired,
    trimStart: PropTypes.func.isRequired,
    valueOf: PropTypes.func.isRequired,
  }),
]),
```

**After:**
```js
type: PropTypes.oneOf([
  'button',
  'checkbox',
  'color',
  'date',
  'datetime-local',
  'email',
  'file',
  'hidden',
  'image',
  'month',
  'number',
  'password',
  'radio',
  'range',
  'reset',
  'search',
  'submit',
  'tel',
  'text',
  'time',
  'url',
  'week',
]),
```

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.tsx
```diff
@@ -1,16 +1,17 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import Stack, { StackProps } from '@mui/material/Stack';
-import TextField, { TextFieldProps } from '@mui/material/TextField';
+import TextField from '@mui/material/TextField';
 import Typography, { TypographyProps } from '@mui/material/Typography';
 import { styled, useThemeProps } from '@mui/material/styles';
 import { useSlotProps } from '@mui/base/utils';
+import { FieldsTextFieldProps } from '@mui/x-date-pickers/internals/models';
```
flaviendelangle, line 8: We need to re-export it from `@mui/x-date-pickers/internals`

packages/x-date-pickers/src/DateField/DateField.tsx
```diff
@@ -154,11 +251,56 @@ DateField.propTypes = {
...
+   */
+  sx: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.func, PropTypes.object, PropTypes.bool])),
+    PropTypes.func,
+    PropTypes.object,
+  ]),
+  /**
+   * Type of the `input` element. It should be [a valid HTML5 input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types).
+   */
+  type: PropTypes.oneOf([
```
flaviendelangle, line 270: We can skip this one, we are forcing a value on our side and I don't think people should override it
Or if we want people to override it, we should put it before the spread

packages/x-date-pickers/src/DateField/DateField.tsx
```diff
@@ -79,10 +97,78 @@ DateField.propTypes = {
    * @default false
    */
   disablePast: PropTypes.bool,
+  /**
+   * If `true`, the label is displayed in an error state.
+   * @default false
+   */
+  error: PropTypes.bool,
```
flaviendelangle, line 104: To remove, we are always overriding it
If we want people to be able to override the `error` attribute, we should edit the behavior to put it before the spread (could be discussed)

flaviendelangle (commented):
Great job !
I just added a few suggestions of props people can't override

packages/x-date-pickers/src/DateField/DateField.tsx
LukasTy: I believe that we are changing `inputmode` and `type` can easily be changed externally. 🤔 
But I agree, that for this kind of component we should not allow it to change. 😉 👍

alexfauquette (approved):
This new propTypes strategy sounds to be a good compromise 👍

flaviendelangle (approved):

docs/scripts/generateProptypes.ts
```diff
@@ -68,14 +68,30 @@ async function generateProptypes(program: ttp.ts.Program, sourceFile: string) {
       if (['children', 'state'].includes(prop.name) && component.name.startsWith('DataGrid')) {
         return false;
       }
-      let shouldDocument = true;
+      let shouldExclude = false;
       prop.filenames.forEach((filename) => {
-        // Don't include props from external dependencies
-        if (/node_modules/.test(filename)) {
-          shouldDocument = false;
```
oliviertassinari, line 75: Maybe we should differentiate between peer dependencies and dependencies. For example, MUI X has the TextField as a peer dependency, so it doesn't control which props from the API are supported or not. So the generation could be off (tell you a prop value isn't there when using a version of MUI X that was built with mui/material@v5.6.0 while the user is on mui/material@v5.7.0.

patches/@mui+monorepo+5.10.16.patch
oliviertassinari: I would propose: https://github.com/mui/mui-x/pull/7225#discussion_r1054889927.

packages/x-date-pickers-pro/src/SingleInputDateRangeField/SingleInputDateRangeField.tsx
```diff
@@ -55,6 +55,25 @@ SingleInputDateRangeField.propTypes = {
...
+   */
+  autoFocus: PropTypes.bool,
+  className: PropTypes.string,
+  /**
+   * The color of the component.
+   * It supports both default and custom theme colors, which can be added as shown in the
+   * [palette customization guide](https://mui.com/material-ui/customization/palette/#adding-new-colors).
+   * @default 'primary'
+   */
+  color: PropTypes.oneOf(['error', 'info', 'primary', 'secondary', 'success', 'warning']),
```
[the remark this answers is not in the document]
LukasTy: > From the DX perspective, I'm not sure that any of the TextField APIs should be duplicated here. I wonder if we shouldn't lean on this approach instead: https://mui.com/material-ui/api/menu/#inheritance. The advantage is that it makes it clear where the props are coming from, and what's the structure of the component. The downside is that people have to learn the pattern once. If they already learned it from MUI Core, they should be good.

Really good point. I did not investigate the MUI core solution to this issue deep enough.
In regards to api documentation your point makes total sense and would allow better differentiation between component and inherited props.
But it still leaves some open questions on edge cases:
- what if we inherit only some of the component props? I'd imagine it might cause a bit of confusion as to which props are actually valid
- what about PropTypes then? They may once again be way out of sync, with what props are actually available on a component. Or we don't care that much about their correctness because a possibly small amount of users are not using TS?

packages/x-date-pickers/src/DateField/DateField.tsx
LukasTy: Nope, we did not look into such solution. But as Flavien mentioned, this change has eventually been reverted in favor of not supporting custom `type` value, because it does not make sense for these components.

patches/@mui+monorepo+5.10.16.patch
LukasTy: That is a good fallback to be aware of if we'll need it again. Thank you. 🙏

docs/scripts/generateProptypes.ts
LukasTy: That is a good point I did not think about. 🤔 
So, all in all, MUI (core) preference is declare inheritance and don't focus on having correct/full PropTypes?

packages/x-date-pickers-pro/src/SingleInputDateRangeField/SingleInputDateRangeField.tsx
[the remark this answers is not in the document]
oliviertassinari: > what if we inherit only some of the component props?

Then the pattern I proposed is not an option. It only works if the component forwards all the props.

> what about PropTypes then?

I think that the type should stay broad and links the source, like in this example: https://mui.com/material-ui/api/tree-item/#props `TransitionProps`.

docs/scripts/generateProptypes.ts
oliviertassinari: If the prop-type is fully declared, then when the component receives a wrong prop, the component warns twice: once in the parent and once in the actually inherited component. It might be clearer to warn once and in the lower level. It teaches developers the lower level component structure.

LukasTy: > MUI Core avoids duplicating the prop-types. If the prop-type is fully declared, then when the component receives a wrong prop, the component warns twice: once in the parent and once in the actually inherited component. It assumes it's clearer to not duplicate the prop-type, to warn once and in the lower level. It teaches developers the lower-level component structure.

It seems that this problem exists in MUI Core as well.
Did a small test with `DateField` component in a fresh JS project and my change added one extra warn/error as you mentioned, but there already are 4 others being thrown by the inherited components. 🙈 
Screenshot 2022-12-30 at 11 21 09

oliviertassinari: The number 2 is strange on these component names.

Yeah, definitely. From what I recall, we decided to duplicate some of the props because the UX of the inherited prop is not good enough on the API pages. The relevant issue: https://github.com/mui/material-ui/issues/18288.

PR #7227, opened 2022-12-16
Add missing polish translations for date pickers

KonradLinkowski:
Signed-off-by: Konrad Linkowski <linkowski.konrad97@gmail.com>

### DIFF
--- a/packages/x-date-pickers/src/locales/plPL.ts
+++ b/packages/x-date-pickers/src/locales/plPL.ts
@@ -27,10 +27,10 @@ const plPLPickers: Partial<PickersLocaleText<any>> = {
   todayButtonLabel: 'Dzisiaj',
 
   // Toolbar titles
-  // datePickerDefaultToolbarTitle: 'Select date',
-  // dateTimePickerDefaultToolbarTitle: 'Select date & time',
-  // timePickerDefaultToolbarTitle: 'Select time',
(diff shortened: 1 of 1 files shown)


LukasTy:
Thank you for your contribution! @KonradLinkowski 🙏 🙇 
Our main development branch is currently `next`. Could you open a PR with changes from that branch?
You will have a few more labels (introduced in next/v6) to optionally translate as well. 😃 
We usually suggest contributors handling the PR/changes targeting the `next` branch and we take care of back-porting those changes to master, but if you are up for it—you could two PRs with translations targeting both branches.

There is also one more label (`inputModeToggleButtonAriaLabel`) available for translation. Would you be able to provide a translation for it? 🤔

KonradLinkowski:
Hi, @LukasTy I will open a PR :)

Seems like links here https://mui.com/x/react-date-pickers/localization/#supported-locales should redirect to the `next` branch

About `inputModeToggleButtonAriaLabel`: it's hard to make it work with polish, but I will look into that :)

LukasTy:
> Seems like links here https://mui.com/x/react-date-pickers/localization/#supported-locales should redirect to the `next` branch

Fair point. I've specifically suggested making it link to master translations in case of stable version docs and [when browsing next (v6)](https://next.mui.com/x/react-date-pickers/localization/#supported-locales) it links to next locale. But yeah, I do understand the possible confusion when looking to contribute. 🙈

PR #7249, opened 2022-12-19
[pickers] Require Luxon 3.0.2 or higher

flaviendelangle:
Now that we don't allow direct adapter import from `@date-io` and that all the fields / pickers fail with older version of Luxon, I think that we can just bumb the peer dependency version.

People could use Luxon v1 / v2 for `DateCalendar` or `TimeClock` but I don't think those components are worth the overhead.


packages/x-date-pickers-pro/package.json
```diff
@@ -82,7 +82,7 @@
     "@types/luxon": "^3.1.0",
     "date-fns": "^2.29.3",
     "dayjs": "^1.11.7",
-    "luxon": "^1.28.0 || ^2.0.0 || ^3.0.2",
```
flaviendelangle, line 85: It's the `devDeps` so like on the community package I'm just putting the last version here

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -38,6 +38,12 @@ If you do not find the adapter you were using—there probably was a reason for
 +import { AdapterMomentJalaali } from '@mui/x-date-pickers-pro/AdapterMomentJalaali';
 ```
 
+### Increased Luxon minimal version
+
+The v6 `AdapterLuxon` now requires `luxon` in version `3.0.2` or higher in order to work.
```
LukasTy, line 43:
```suggestion
The v6 `AdapterLuxon` now requires `luxon` version `3.0.2` or higher in order to work.
```

LukasTy (approved):

alexfauquette (approved):

PR #7264, opened 2022-12-20
[codemod] New codemod for view component renaming

flaviendelangle:
Codemod for #6389 and #6851


packages/x-codemod/src/v6.0.0/view-components-rename/index.ts
```diff
@@ -0,0 +1,100 @@
...
+      const subPackage = matchImport(path)![3];
+      const importPath = path.node.source.value?.toString() ?? '';
+
+      return j.importDeclaration(
+        path.node.specifiers, // copy over the existing import specifiers
+        j.stringLiteral(importPath.replace(subPackage, SUB_PACKAGES[subPackage])), // Replace the source with our new source
+      );
+    });
+
+  // Rename the import sage
```
LukasTy, line 91: sage? Am I missing something or is it a typo? 🙈

packages/x-codemod/src/v6.0.0/view-components-rename/view-components-rename.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  return readFile(path.join(__dirname, fileName));
+}
+
+const TEST_FILES = ['community-nested-imports', 'community-root-imports', 'pro-root-imports'];
+
+describe('v6.0.0', () => {
+  describe('view-components-rename', () => {
+    TEST_FILES.forEach((testFile) => {
+      const actualPath = `./actual-${testFile}.spec.tsx`;
+      const expectedPath = `./expected-${testFile}.spec.tsx`;
```
LukasTy, line 17: Nitpick: I have specifically gone with using `js` files instead of `ts` for test fixtures in order to avoid red IDE and additional TS processing for files we know will usually have errors.
But if you think it adds value having them authored in `ts` as well—by no means, we can go with. 👌

LukasTy (approved):
Very thorough testing. 👍 
Looks great to me 🚀

packages/x-codemod/src/v6.0.0/view-components-rename/view-components-rename.test.ts
flaviendelangle: I am testing the renaming of the exported types, that's why I went for TS :+1:

packages/x-codemod/src/v6.0.0/view-components-rename/index.ts
flaviendelangle: :laughing: Clearly a typo

packages/x-codemod/src/v6.0.0/view-components-rename/view-components-rename.test.ts
LukasTy: Ah, fair point 🤦

PR #7272, opened 2022-12-20
[fields] Improve editing (automatic section switch, allow letter editing in digit section, allow numeric editing in letter section)

### LINKED ISSUE #7252: [pickers] Auto switch of sections in date and time Fields
### Summary 💡

When users fill in the current section of a date, the field automatically switches the focus to the next piece, so there's no need to navigate through the sections manually.

That means that if the user wants to input his birthday, he'll simply type, for instance: `05052005` for the output `05/05/2005` (considering the standard separators)

### Examples 🌈

https://react-spectrum.adobe.com/react-spectrum/DateField.html

### Motivation 🔦

Improved UX and intuitiveness.

flaviendelangle:
[Playground](https://codesandbox.io/s/date-and-time-pickers-forked-eem4o0)

Fixes #7252

- [x] Add automatic section switch 
- [x] Allow to type letters in a digit month field (typing `J` will render `01`, then typing `U` will render `06`, etc...)
- [x] Allow the to type digits on a letter month section (typing `6` to render `June`)
- [x] Clean the query after 5sec without any change 

I created two global variables to enable the new behaviors.
The goal is of course to remove them before merging and apply the behavior for everybody.

### Automatic section switch

I did not enable the jump by default, to let you play with the old and new behavior.
You can enable it using the `ponyfillGlobal.__MUI__PICKERS_ENABLE_AUTOMATIC_SECTION_SWITCH__` variable.

### Changes to the numeric editing

With the current numeric editing approach, if a user presses `0` then `1` in a day or month section, the hook does not consider it as filled because the new value is `1` and `1` can still be completed (`10-12` for the month and `10-19` for the day).

This makes the automatic section switch quite painful for these sections.

The root problem is that our numeric editing is using the section value and the key pressed to determine the new section value. So it can't differentiate `01` and `1`.

To solve this, we need to move the numeric editing to a query-based editing like the letter editing.
When editing a section, we store the typed value in a ref (this allows us to differentiate `1` and `01`).
And when the new section value can't be completed without overflowing (typing `5` on a day section) or when the size of the typed string is equal to the size of the maximum boundary (when typing `01` on a day section), we jump to the next section.

Apart from allowing a coherent automatic section switch, the main difference is that now the section value is reset when it overflows.
With the current behavior, if on a day section you type `1 - 2 - 3`, it will write `23` when React Spectrum would write `3` (if the day is the last section, otherwise it just jumps to the next section after the `2` is pressed).
With the query-based numeric editing, we would have the same behavior as React Spectrum.

I kept both behavior on this PR to let you play with it depending on `ponyfillGlobal.__MUI__PICKERS_ENABLE_QUERY_BASED_NUMERIC_EDITING__`:

- If `false` or `undefined`, the behavior is the same as on next (`1 - 2 - 3` will write `23`)
- If `true`, the query based numeric editing is enabled (`1 - 2 - 3` will write `3`)

### Review

I need to confirm that we are going for the automatic section switch and the query-based numeric editing before writing the new tests.
A lot of our tests around field editing are slightly broken (as expected).


packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -211,135 +209,7 @@ export const useField = <
...
-          if (activeSection.contentType === 'letter') {
-            return activeSection.value;
-          }
-
-          return getNewSectionValueStr(null, boundaries);
-        },
-      });
-    }
-    // TODO: Improve condition
-    else if (['/', ' ', '-'].includes(keyPressed)) {
```
flaviendelangle, line 284: I removed this logic
React Spectrum does not have it either

The only use case where it could be useful is if somebody pressed `1` then `/`.
Before, it would have moved to the next section
Now it stays on the current one

I can't just keep it as it is, otherwise, typing `12/` would move twice to the next section.

I can re-enable it only when the query is on the current section and not empty.
Not sure the complexity is worth it.

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
```diff
@@ -0,0 +1,442 @@
...
+  const applyLetterEditing: CharacterEditingApplier<TDate, TSection> = (params) => {
+    const getFirstSectionValueMatchingWithQuery: QueryApplier<TSection> = (
+      queryValue,
+      activeSection,
+    ) => {
+      switch (activeSection.dateSectionName) {
+        case 'month': {
+          const getMonthResponse = (format: string): ReturnType<QueryApplier<TSection>> => {
+            const matchingMonths = utils
+              .getMonthArray(utils.date()!)
```
LukasTy, line 155: Nitpick, but this could probably be memoized somewhere as its result should not change, unless locale changes. 🤔

LukasTy (approved):
Looks and works great! 👍 
Couldn't find a flaw, great job! 🚀

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
```diff
@@ -0,0 +1,442 @@
...
+
+        case 'meridiem': {
+          const now = utils.date()!;
+          const sectionValue = [utils.endOfDay(now), utils.startOfDay(now)]
+            .map((date) => utils.formatByString(date, activeSection.formatValue))
+            .find((meridiem) => meridiem.toLowerCase().startsWith(queryValue));
+
+          if (sectionValue == null) {
+            return { saveQuery: false };
+          }
```
alexfauquette, line 208: Could be buggy if there is a local with a meridian starting with the same letter

But not sure it worth to handle such a case

alexfauquette (commented):
Nice architecture 👍

I tried hard to break it and found only one dummy way: mixing digits and letters.

In the month section, press <kbd>m</kbd> then <kbd>0</kbd> to get `NaN`, because it will try to parse `Number('m0')`

flaviendelangle:
> In the month section, press m then 0 to get NaN, because it will try to parse Number('m0')

I added a `Number.isNaN` check before concatenating the queries and it works great

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
flaviendelangle: This line is only used when no meridiem matches, so we should have no issue even with multiple meridiems starting with the same letter

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
```diff
@@ -0,0 +1,442 @@
...
+      queryValue,
+      activeSection,
+    ) => {
+      switch (activeSection.dateSectionName) {
+        case 'month': {
+          const getMonthResponse = (format: string): ReturnType<QueryApplier<TSection>> => {
+            const matchingMonths = utils
+              .getMonthArray(utils.date()!)
+              .map((month) => utils.formatByString(month, format!))
+              .filter((month) => month.toLowerCase().startsWith(queryValue));
```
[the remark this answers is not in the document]
flaviendelangle: Could be a follow up indeed

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
flaviendelangle: We can do a follow up where we put this info in the `boundaries` (and rename it of course).

alexfauquette (approved):

alexfauquette:
> I added a Number.isNaN check before concatenating the queries and it works great

Approved too quickly, sounds better to filter `queryValue` at the beginning of appliers 

Make sure it only contains letters in `applyLetterEditing` 
Make sure it only contains digits in `applyNumericEditing`

flaviendelangle:
> Approved too quickly, sounds better to filter queryValue at the beginning of appliers

I changed the check to only run it on letter editing.
For letter editing, if we have `0m` for example, it should work fine even without any check.

flaviendelangle:
I migrated the tests to the new behavior

PR #7294, opened 2022-12-21
[docs] Fix SingleInputDateRangeField API generation

oliviertassinari:
The test location is used to generate the API.

Before https://next--material-ui-x.netlify.app/x/api/date-pickers/single-input-date-range-field/

Screenshot 2022-12-21 at 23 07 48

After https://deploy-preview-7294--material-ui-x.netlify.app/x/api/date-pickers/single-input-date-range-field/

Screenshot 2022-12-21 at 23 07 36

ℹ️ there are probably many more components that are impacted by this test name and location problem.

### DIFF
--- a/docs/pages/x/api/date-pickers/single-input-date-range-field.json
+++ b/docs/pages/x/api/date-pickers/single-input-date-range-field.json
@@ -69,7 +69,10 @@
   "slots": { "Input": { "default": "TextField", "type": { "name": "elementType" } } },
   "name": "SingleInputDateRangeField",
   "styles": { "classes": [], "globalClasses": {}, "name": "MuiSingleInputDateRangeField" },
+  "spread": true,
+  "forwardsRefTo": "HTMLDivElement",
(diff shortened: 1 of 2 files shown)


LukasTy (approved):
Great spot!
I've missed the fact of incorrect `ref` forwarding note.
Will need to have a look through other components. 👌

oliviertassinari:
The alternative solution would be to update the "describe" parsing logic in the mono repo to also be aware of this new pattern used in the date picker (group the tests inside a folder).

Personally, I prefer working with big files but starting to split beyond 2,000 lines.

flaviendelangle:
>  this new pattern used in the date picker

It's the same pattern as on the data grid tests :+1:

oliviertassinari:
I'm closing as this is only the tip of the iceberg, this PR won't really make a difference.

oliviertassinari:
For the test file location, no real preferences as long as it works and we enforce as few as possible options that work.

PR #7302, opened 2022-12-22
[DataGridPro] Use row ID as `key` of the detail panels

### LINKED ISSUE #7256: DataGridPro seems to reuse detail panel across different rows
### Duplicates

### Latest version

### Steps to reproduce 🕹

Please have a look at the example at https://codesandbox.io/s/datagrid-issue-fuf0ut.

### Current behavior 😯

If you click on different rows, you will see the same random number appearing in the details panel of different rows.

### Expected behavior 🤔

Each row should have its own details panel and thus, each row should have a unique random value.

### Context 🔦

Problem only seems to occur when only one panel is visible. If I allow multiple details panels to become visible at the same time, each one will have its own random value:
image

I find this highly unexpected and its causing issues because when I click on a row to expand the details panel, the panel will show up with the wrong details, which came from the previous row. Even if this only happens for one tick, the flicker is still annoying.

### Your environment 🌎

I was unable to run `npx @mui/envinfo` for some reason. These are the dependencies in the sandbox example:

```
"dependencies": {
    "@emotion/react": "11.10.4",
    "@emotion/styled": "11.10.4",
    "@mui/material": "5.10.7",
    "@mui/x-data-grid-pro": "5.17.6",
    "react": "18.2.0",
    "react-dom": "18.2.0"
  },
```

### Order ID 💳 (optional)

_No response_

m4theushw:
Fixes #7256 

Before: https://codesandbox.io/s/datagrid-issue-fuf0ut
After: https://codesandbox.io/s/datagrid-issue-forked-iiwzuq?file=/src/App.js

### DIFF
--- a/packages/grid/x-data-grid-pro/src/tests/detailPanel.DataGridPro.test.tsx
+++ b/packages/grid/x-data-grid-pro/src/tests/detailPanel.DataGridPro.test.tsx
@@ -25,7 +25,7 @@ import { getRow, getCell, getColumnValues } from 'test/utils/helperFn';
 const isJSDOM = /jsdom/.test(window.navigator.userAgent);
 
 describe('<DataGridPro /> - Detail panel', () => {
-  const { render } = createRenderer({ clock: 'fake' });
+  const { render, clock } = createRenderer({ clock: 'fake' });
 
   let apiRef: React.MutableRefObject<GridApi>;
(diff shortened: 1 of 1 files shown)


MBilalShafi (approved):
Should we add a test for this one?

PR #7324, opened 2022-12-26
[docs] Fix `disablePast` and `disableFuture` definition swap

alexfauquette:
From slack feedback: 

> The descriptions for `disablePast` and `disableFuture` are flipped. For example, `disablePast` says "If true disable values after the current date for date components, time for time components and both for date time components," which should be the description for `disableFuture`. Instead, `disablePast` should say: "If true disable values before the current date for date components, time for time components and both for date time components."

The first commit is the 2 lines diff I did. Others are the results of `yarn proptypes` and `yarn docs:api`


LukasTy (approved):

flaviendelangle (approved):

flaviendelangle:
I'm merging this one to avoid future Git conflict :+1:

Issue #3140, opened 2021-11-08
[data grid]Improve the filtering panel UX


chrishawn:
### Summary 💡

I feel like the UX behavior should be if the filter dialog is open i should be able to remove all the rows (~and if the last row is removed the dialog should close~).

### Examples 🌈

<img width="552" alt="Screen Shot 2021-11-08 at 4 55 13 PM" src="https://user-images.githubusercontent.com/309639/140832116-d0d21ebb-63b9-4257-8467-186596ea1f70.png">

### Motivation 🔦

it's not always obvious to people to click outside of the popup or the filters menu button to get the filtering panel closed. also if you have a rowclick event enabled and clicking off the filter can fire that event if not careful.

### Order ID 💳 (optional)

31818

flaviendelangle:
Closing the panel when deleting the last filter is quite simple.
I would be interested in the feedback of other users / of the rest of the team. Because both can make sense.

m4theushw:
On Airtable, deleting the last filter shows the following message:

<img width="500" src="https://user-images.githubusercontent.com/42154031/141121973-58dbbab9-09c1-4b78-8eea-5409edfd4a24.png" alt="image" style="max-width: 100%;">

For the DataGrid, I don't know why it was decided to do this way when the filter panel was created. We could close the panel when the last filter is deleted. cc @oliviertassinari 

> it's not always obvious to people to click of the dialog or the filters menu button to get it to toggle it closed. also if you have a rowclick event enabled and clicking off the filter can fire that event if not careful.

Only closing the panel when deleting the last filter won't solve this problem. The user could still mistakenly click in a row to close the panel. A better solution would be show a backdrop when any dialog is open. I don't mean a backdrop from [`Modal`](https://mui.com/components/modal) which hides the scrollbar, but a backdrop only on top of the virtual scroller. Also, Airtable uses this approach.

oliviertassinari:
I think that we should have @danilo-leal redo the UX of the filtering panel for us, and freeze all changes until it's done. For me, this pain was part of #1777. The panel UX if not great (to be politically correct). I personally got tricked in a different way, tried to click the arrow to close the panel 🙈

oliviertassinari:
Actually, Airtable changed its design since the last time we work on the filtering. It seems to work well 👌

![IMG_20211110_185811~2](https://user-images.githubusercontent.com/3165635/141167707-e193cc2f-774c-4667-8270-e52a21b3e23c.jpg)

[...]

PR #7326, opened 2022-12-26
[DataGridPro] Add `Remove All` option in filter panel

### LINKED ISSUE #7313: [DataGridPro] Add clear all filters button
### Duplicates

### Latest version

### Summary 💡

image

Need the `Clear All Filters` button beside `Add Filter` that will clear all the applied filters.

### Examples 🌈

_No response_

### Motivation 🔦

It's tough to clear all the filters one by one.

MBilalShafi:
Closes #7313 

- Add "Remove all" button to allow clearing multiple filters
- Allow disabling Add Filter and Remove All buttons (it could be beneficial for some corner cases like in https://github.com/mui/mui-x/pull/6947)
- Update icons of remove and remove all as per new design.

Preview: https://deploy-preview-7326--material-ui-x.netlify.app/x/react-data-grid/filtering/


gerdadesign:
> I'm not sure "Clear all" is the best naming for it. At the current state, it's behaving as a "Delete all" button.

This was mentioned in another issue as well #3140

Under the current behaviour, I would agree with changing the X icon → trash bin ([Delete Icon](https://fonts.google.com/icons?selected=Material%20Icons%3Adelete%3A)) and X CLEAR ALL → ([Delete Forever Icon](https://fonts.google.com/icons?selected=Material%20Icons%3Adelete_forever%3A)) DELETE ALL 

Delete Icons

gerdadesign:
Yeah, I would expect DELETE ALL to remove all filters and close the panel. 
I would expect CLEAR ALL to remove the values in the filter fields and keep the panel open. This would either show an empty state for the panel or show the default view of one filter row with an empty value. 

It's not entirely clear from the linked issues which behaviour is more preferred by users. I wonder if @joserodolfofreitas has any additional data points to help determine if one method is more requested than the other? 

It's also entirely possible that we just choose one convention and stick with it — either a pattern of Clearing or a pattern of Deleting, so that it's more predictable behaviour. Since each row includes Delete, I would opt for continuing the Delete pattern (which answers the original issue #7313 of not having to individually remove every filter).

joserodolfofreitas (approved):

olzhas-kalikhan:
are there plans to include this "remove all" button in v5?

PR #7327, opened 2022-12-26
[docs] Support new editing API in custom columns from the generator

### LINKED ISSUE #7300: Bug on datagrid overview page, cannot close dialog after has been opened

yaredtsy:
Closes #7300 

**Problem**

some of `data-grid-generator` renderers only use the `oldEditingApi` function to commit and stop the edit mode. they do not support the `newEditingApi`. it is causing an error if it's used with `newEditingApi`.
Demo: https://codesandbox.io/s/09htss?file=/demo.tsx try to edit the `incoTerm` column fields to reproduce the error.

**Solution**

check if `newEditingApi` is enabled and use the correct function to commit and stop the edit mode.


flaviendelangle:
Closing this one, we no longer have the old editing API so the new one is applied everywhere

PR #7334, opened 2022-12-28
[fields] Add `a11y` support

### LINKED ISSUE #6384: [fields] Refactor to use multiple elements to support a11y
The new `DateField` (and their sibling components) need custom a11y implementation to be accessible by screen readers.

W3 Aria [has an example](https://www.w3.org/WAI/ARIA/apg/example-index/spinbutton/datepicker-spinbuttons.html) of date picker component with spin buttons. Their example looks and feels (functionality) very similar to our `DateField` component behaviour, hence, it could make a very good base for a11y support in the component.

### Exploration notes
- Using [spinbutton](https://www.w3.org/WAI/ARIA/apg/example-index/spinbutton/datepicker-spinbuttons.html) approach and augmenting input element with additional aria attributes does not work. Possibly the ever changing full `value` attribute as well as custom section selecting interferes with screen readers ability to announce proper aria attribute denoted data.
- Having separate element with proper aria attributes updated based on selection does not work, because that element is not focused and it's attributes are ignored

### Possible exploration
- Having `spinbutton` divs (overlaying the underlying input element), which would intercept tabulation/click events, propagate those events to the actual field and update the spinbutton aria attributes accordingly (it seems like tedious and janky solution to achieve pretty simple result easily doable with [different field implementations](https://react-spectrum.adobe.com/react-spectrum/DateField.html))
- Provide a hidden label with full current date string (updated after any section data change) at least announcing to end user the currently selected value

## After exploration

Results uncovered in #7334 proved that the only logical path is to refactor the field component rendering in order to achieve the desired `spinbutton` behavior. Other implementations seem like hacks, without a correct tangible result.

LukasTy:
Fixes #6384

Current branch implementation contains a POC of a hacky manual a11y implementation for `date field` component.
If an implementation "along these lines" doesn't work—I don't really see a better option than a refactor to react spectrum style.

@m4theushw Could you compare [existing situation](https://next.mui.com/x/react-date-pickers/date-field/#basic-usage) with [this implementation](https://deploy-preview-7334--material-ui-x.netlify.app/x/react-date-pickers/date-field/#basic-usage) and give your opinion if going forth with this kind of implementation is worth the effort? 🤔


LukasTy:
> The only way to avoid the situations above would be to not use the selection API to communicate which date component is selected. Have you considered using something like a div under the current date component? It will be hard to match the alignment with the separator but the screen reader won't read it.

That is our plan B. Initially this approach was not chosen because of probable harder visual customizability support, but after learning the fact that with the current Fields implementation we are either having easy customization by simply using an `input` element or we have proper a11y support by choosing an ARIA supported approach, i.e. like you mentioned, what React Spectrum has chosen.

After making all these discoveries and seeing how much extra fragile measures we need to take in order to have half-baked a11y support—I tend to lean towards parking this implementation and:
1. finding the time to investigate if the necessary component changes can be shipped under a feature flag without necessary breaking changes during v6
2. deliver "plan B" refactor with v7

Would you @m4theushw agree with these options or do you still think that fine-tuning this solution is worth it during v6?

flaviendelangle:
> was there a particular reason why we went with selecting all the sections when focusing the field? Or was it just a default that was chosen

I followed Telerik behavior of selecting all sections on focus
But the HTML date input focused the 1st section, so this behavior is probably questionable.

PR #7337, opened 2022-12-29
[docs] Note the pickers breaking changes supported by the codemod

flaviendelangle:


docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -17,6 +17,21 @@ Since v6 is a major release, it contains some changes that affect the public API
...
 
+## Run codemods
+
+The `preset-safe` codemod will automatically adjust the bulk of your code to account for breaking changes in v6 for both the Date and Time Pickers and DataGrid.
+It should be only applied **once per folder.**
+
+```sh
+npx @mui/x-codemod v6.0.0/preset-safe <path>
+```
+
```
LukasTy, line 28: Maybe we should also add a similar note a in https://mui.com/material-ui/migration/migration-v4/ ?
Screenshot 2022-12-29 at 11 07 17

LukasTy (approved):

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
flaviendelangle: Indeed, I'll add it

By the way, maybe a `preset-safe-pickers` and `preset-safe-data-grid` would make sense for people wanting to differentiate the two migrations, as they are independent (and a `preset-safe` that runs both of them)

LukasTy: > By the way, maybe a preset-safe-pickers and preset-safe-data-grid would make sense for people wanting to differentiate the two migrations, as they are independent (and a preset-safe that runs both of them)

That is a very good point. 🤔 
I'd agree on introducing it. 👌 
What do you think @mui/xgrid ?

flaviendelangle: @LukasTy added 

I think we should differentiate the grid and pickers codemod in a cleaner way

Either with `v6.0.0-pickers/xxx` and `v6.0.0-grid/xxx` or `v6.0.0/pickers/xxx` and `v6.0.0/grid/xxx`, or `v6.0.0/pickers-xxx` and `v6.0.0/grid-xxx`.

Because right now, someone using only the grid have to go read all the pickers codemod to check if it impacts him.

LukasTy: > I think we should differentiate the grid and pickers codemod in a cleaner way

I would agree that going with the following makes sense. It would avoid the need to come up with unique individual codemod names for any codemod applicable (very similar) to both grid and pickers. 🤔 
- v6.0.0/pickers/xxx
- v6.0.0/data-grid/xxx
- v6.0.0/pickers/preset-safe
- v6.0.0/data-grid/preset-safe
- v6.0.0/preset-safe

flaviendelangle: That would be fine for me :+1:

cherniavskii: Looks good to me too :+1:

I don't think I understand this statement though:
> If you want to run the transformers one by one, check out the [preset-safe codemod](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md#-preset-safe) for more details.

Does `preset-safe` allow running the transformers one by one?

LukasTy: > Does `preset-safe` allow running the transformers one by one?

Indeed. It should probably link to [this heading](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md#v600) and not mention the `preset-safe`.
The copyright for that paragraph could definitely be improved 🙈

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -17,6 +17,25 @@ Since v6 is a major release, it contains some changes that affect the public API
...
+:::info
+If you want to run the transformers one by one, check out the [preset-safe codemod](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md#-preset-safe) for more details.
```
LukasTy, line 30:
```suggestion
If you want to run the transformers one by one, check out the transformers included in the [preset-safe codemod](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md#-preset-safe) for more details.
```

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
flaviendelangle: Suggestion applied

PR #7338, opened 2022-12-29
[codemod] Add new codemod for the value prop renaming on the view components

flaviendelangle:
Part of #6997


docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -169,22 +169,22 @@ Component names in the theme have changed as well:
...
 
-### Rename `date` prop to `value`
+### ✅ Rename `date` prop to `value`
 
 The `date` prop has been renamed `value` on `MonthCalendar`, `YearCalendar`, `TimeClock`, and `DateCalendar` (components renamed in previous section):
 
 ```diff
--<MonthPicker date={dayjs()} onChange={handleMonthChange} />
-+<MonthCalendar value={dayjs()} onChange={handleMonthChange} />
+-<MonthPicker date={dayjs()} />
```
LukasTy, line 177: 🙌

packages/x-codemod/src/util/renameProps.ts
```diff
@@ -1,17 +1,26 @@
-import type { Collection } from 'jscodeshift';
+import type { Collection, JSCodeshift } from 'jscodeshift';
 
 interface RenamePropsArgs {
   root: Collection<any>;
-  componentName: string;
+  componentNames: string[];
```
LukasTy, line 5: ❤️

packages/x-codemod/README.md
```diff
@@ -98,6 +98,24 @@ Renames the view components
 +<TimeClock {...props} />
 ```
 
+#### `view-components-rename-value-prop`
```
LukasTy, line 101: We should also add a list item in the list above all sections (preset-safe) paragraph. 😉

packages/x-codemod/src/v6.0.0/view-components-rename-value-prop/view-components-rename-value-prop.test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { expect } from 'chai';
+import jscodeshift from 'jscodeshift';
+import transform from '.';
+import readFile from '../../util/readFile';
+
+function read(fileName) {
+  return readFile(path.join(__dirname, fileName));
+}
+
+describe('v6.0.0', () => {
```
LukasTy, line 11: Ideally we should/could add at least one testable item into the `preset-safe` fixtures as well, to double ensure the consistency between transformations and automatically ensure that incorrect transformation order would fail a test. 🤔

LukasTy (approved):

packages/x-codemod/README.md
flaviendelangle: Added

packages/x-codemod/src/v6.0.0/view-components-rename-value-prop/view-components-rename-value-prop.test.ts
flaviendelangle: I renamed the `value` prop of the `CalendarPicker` into `date` in the `actual.spec.js` of the `preset-safe` :+1:

PR #7347, opened 2022-12-30
[docs] Add example to add back the mobile keyboard view

flaviendelangle:


packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
```diff
@@ -58,8 +59,6 @@ export const PickersLayoutContentWrapper = styled('div', {
   slot: 'ContentWrapper',
   overridesResolver: (props, styles) => styles.contentWrapper,
 })({
-  gridColumn: 2,
```
flaviendelangle, line 61: @alexfauquette was there a reason for the coordinates to be defined here but defined on the container for the action bar ? Is this because the file is defined in the same file ?

And are you just `2 2` was the right coordinates ? Shouldn't it be on multiple columns ?

alexfauquette: As we discussed in the meeting:

- (2,2) is the coordinate for when you need to place components on basic position top, top-right, right, ...
- Yes, the position can be defined somewhere else. I defined it here because it's a styled component so I had the opportunity to add CSS for it. Others are not accessible in this fils, so I used classes

packages/x-date-pickers/src/DatePicker/index.ts
```diff
@@ -1,5 +1,7 @@
 export { DatePicker } from './DatePicker';
 export type { DatePickerProps } from './DatePicker';
+
+export { DatePickerToolbar } from './DatePickerToolbar';
```
alexfauquette, line 4: Need to do the same for the `TimePickerToolbar` and `DateTimePickerToolbar`.

alexfauquette (commented):
For me, the solution with the `@ts-ignore` is good enough

packages/x-date-pickers-pro/src/internal/hooks/useStaticRangePicker/useStaticRangePicker.tsx
```diff
@@ -63,7 +63,7 @@ export const useStaticRangePicker = <
     toolbar: {
       ...componentsProps?.toolbar,
       rangePosition,
-      onDateRangePositionRange: setRangePosition,
+      onRangePositionChange: setRangePosition,
```
flaviendelangle, line 66: Having the proptypes on the `DateRangePickerToolbar` helped me find this typo :partying_face:

alexfauquette (approved):

PR #7349, opened 2022-12-30
[core] Order the slots alphabetically in the JSON files

flaviendelangle:
I have a weird CI error in https://app.circleci.com/pipelines/github/mui/mui-x/30816/workflows/b609017f-caab-4068-91e5-d88cce54eab9/jobs/176499

I ordered the slot in the JSON file to make sure we don't have changes based on how the slots are extracted from the AST.


LukasTy (approved):
LGTM. The resulting doc doesn't change one bit as the slots were already being sorted alphabetically somewhere else. 🤔

PR #7350, opened 2022-12-30
[core] Handle selection edge case

oliviertassinari:
Per https://github.com/mui/material-ui/pull/35670/files#r1059417694, this might be better

### DIFF
--- a/packages/grid/x-data-grid/src/hooks/features/clipboard/useGridClipboard.ts
+++ b/packages/grid/x-data-grid/src/hooks/features/clipboard/useGridClipboard.ts
@@ -26,15 +26,15 @@ function writeToClipboardPolyfill(data: string) {
 }
 
 function hasNativeSelection(element: HTMLInputElement) {
-  if (window.getSelection()?.toString() !== '') {
+  // When getSelection is called on an <iframe> that is not displayed Firefox will return null.
+  if (window.getSelection() && window.getSelection()!.toString() !== '') {
     return true;
   }
 
-  if (!element) {
-    return false;
-  }
-
-  if ((element.selectionEnd || 0) - (element.selectionStart || 0) > 0) {
+  // window.getSelection() returns an empty string in Firefox for selections inside a form element.
+  // See: https://bugzilla.mozilla.org/show_bug.cgi?id=85686.
+  // Instead, we can use element.selectionStart that is only defined on form elements.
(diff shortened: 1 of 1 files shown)


flaviendelangle (commented):
Is it really a `core` PR ?

oliviertassinari:
I think that it has barely any impact, so most people wouldn't care.

packages/grid/x-data-grid/src/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -26,15 +26,15 @@ function writeToClipboardPolyfill(data: string) {
 }
 
 function hasNativeSelection(element: HTMLInputElement) {
-  if (window.getSelection()?.toString() !== '') {
+  // When getSelection is called on an <iframe> that is not displayed Firefox will return null.
+  if (window.getSelection() && window.getSelection()!.toString() !== '') {
```
oliviertassinari, line 30: I can't reproduce what they say in https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection, it might be even better:

```suggestion
  if (window.getSelection()!.toString() !== '') {
```

packages/grid/x-data-grid/src/hooks/features/clipboard/useGridClipboard.ts
```diff
@@ -26,15 +26,15 @@ function writeToClipboardPolyfill(data: string) {
 }
 
 function hasNativeSelection(element: HTMLInputElement) {
-  if (window.getSelection()?.toString() !== '') {
```
oliviertassinari, line 29: @cherniavskii I think that this wasn't exactly correct. Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

>  If the object accessed or function called is undefined or null, it returns undefined instead of throwing an error.

So when the selection returns null, the chaining returns `undefined` which we then compare to `''`. it's different, so say there is a selection. But there is no selection in this case, in theory, per https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection.

> When called on an [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) that is not displayed (e.g., where display: none is set) Firefox will return null

cherniavskii (approved):
Nice catch!

PR #7356, opened 2022-12-31
[pickers] Remove redundant `variants` in theme augmentation

LukasTy:
`variants` are not supported by any component in `mui-x` repository.
Most of mui-core components do have variants and so they have additional typing and logic to support additional variants extension.
In pickers case we did not even support it anywhere.

This is technically a breaking change, but it would not impact anyone as those fields served no purpose. It was just a mistake on our side.

If you think that we should include it in the migration docs—we could do it, but I see little to no purpose in it... 🙈


flaviendelangle (approved):
Make total sense !
Not sure it's worth adding to the migration guide since it does not work.
I'm not doing sections for the props of the pickers I removed and that did nothing in v5

alexfauquette (approved):
Nice catch 👏👏👏

PR #7368, opened 2023-01-02
[codemod] Reorganize codemods and add codemod to rename column menu components

MBilalShafi:
- Some refactorings in the structure of codemods
- Codemod for new column menu released in #6619


packages/x-codemod/README.md
```diff
@@ -170,3 +171,30 @@ If you were always using the same text value in all your components, consider mo
 ```
 
 You can find more details about this breaking change in [the migration guide](https://next.mui.com/x/migration/migration-pickers-v5/#rename-the-locale-prop-on-localizationprovider).
+
+#### `column-menu-components-rename`
```
MBilalShafi, line 175: I am thinking to categorize codemods related to grid and pickers so that users using one of the products are more convenient.
At least we can consider segregating them in documentation if not in folder structure too.
Though it's not a big deal, but we can also think about dividing `preset-safe` into `preset-safe-grid` and `preset-safe-pickers` to make things more specific.

flaviendelangle: We discussed it in https://github.com/mui/mui-x/pull/7337#discussion_r1058855910

I'm in favor of also splitting the folder structure.

LukasTy: > Though it's not a big deal, but we can also think about dividing `preset-safe` into `preset-safe-grid` and `preset-safe-pickers` to make things more specific.

Yes. We agreed on it.
So feel free to either adjust it with this PR introducing 1st data-grid codemod.
Or I could adjust the structure after this codemod is merged. 😉

packages/x-codemod/src/v6.0.0/column-menu-components-rename/column-menu-components-rename.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const TEST_FILES = ['community-root-imports', 'pro-root-imports', 'premium-root-imports'];
+
+describe('v6.0.0', () => {
+  describe('column-menu-components-rename', () => {
+    TEST_FILES.forEach((testFile) => {
+      const actualPath = `./actual-${testFile}.spec.tsx`;
+      const expectedPath = `./expected-${testFile}.spec.tsx`;
+
+      describe(`Package (${testFile.replace(/-/g, ' ')})`, () => {
```
LukasTy, line 19: Nitpick, maybe you wanted to write:
```suggestion
      describe(`Package (${testFile.split(/-/g)[0]})`, () => {
```
producing `community` instead of `community root imports`? 🤔

LukasTy (commented):
Looks good from my point of view, but I'm not that well aware about the details of this change and other possible permutations.

Could you also add a minimal test case to the `preset-safe` just to safeguard against possible transformation order issues down the line. 😉

alexfauquette (approved):
Nice :)

If there is other component renaming planned, could make sense to create a helper `renameComponent` on the same idea than `renameProps`

packages/x-codemod/README.md
MBilalShafi: Done, please check if you see anything missing

LukasTy: I'd suggest updating the data grid migration doc page to be more in line with picker:
- mention running codemods
- mark breaking changes handled by codemod

LukasTy (commented):
I've played around the migration doc as well as readme file and here are my suggestions and fixes. Listing them in comment causes issues due to documentation formatting included in the diff snippets.

[diff.patch](https://github.com/mui/mui-x/files/10331875/diff.patch)

packages/x-codemod/src/v6.0.0/pickers/adapter-change-import/actual.spec.js
```diff
@@ -14,4 +14,4 @@ import JalaaliAdapter from '@date-io/jalaali';
   <LocalizationProvider dateAdapter={MomentAdapter} />
   <LocalizationProvider dateAdapter={HijriAdapter} />
   <LocalizationProvider dateAdapter={JalaaliAdapter} />
-</>
\ No newline at end of file
+</>;
```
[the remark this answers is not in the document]
MBilalShafi: Yes, seems something is bad with my prettier config 😕

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
```diff
@@ -13,6 +13,28 @@ In `package.json`, change the version of the data grid package to `next`.
...
+npx @mui/x-codemod v6.0.0/preset-safe <path>
+```
+
+:::info
+If you want to run the transformers one by one, check out the transformers included in the [preset-safe codemod for data grid](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md#preset-safe-for-data-grid) for more details.
+:::
+
+Breaking changes that are handled by this codemod are denoted by a ✅ emoji in the table of contents on the right side of the screen.
+
+If you have already applied the `v6.0.0/data-grid/preset-safe` (or `v6.0.0/preset-safe`) codemod, then you most proabably should not need to take any further action on these items. If there's a specific part of the breaking change that is not part of the codemod or needs some manual work, it will be listed in the end of each section.
```
LukasTy, line 34: The sentence becomes a bit "clumsy" with those words included. It also makes it very uncertain from our point of view. 🤔 

```suggestion
If you have already applied the `v6.0.0/data-grid/preset-safe` (or `v6.0.0/preset-safe`) codemod, then you should not need to take any further action on these items. If there's a specific part of the breaking change that is not part of the codemod or needs some manual work, it will be listed in the end of each section.
```

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
```diff
@@ -104,9 +126,15 @@ The minimum supported Node.js version has been changed from 12.0.0 to 14.0.0, si
   | `GridAggregationColumnMenuItem`                                     | `GridColumnMenuAggregationItem` |
   | `GridRowGroupingColumnMenuItems`, `GridRowGroupableColumnMenuItems` | `GridColumnMenuGroupingItem`    |
 
-- `GridFilterItemProps` has been renamed to `GridColumnMenuItemProps`.
+- The `GridFilterItemProps` has been renamed to `GridColumnMenuItemProps`.
 - Props `column` and `currentColumn` passed to `GridColumnMenu` and column menu items have been renamed to `colDef`
 
+These changes are not part of `preset-safe` codemod and you need to handle them manually:
```
LukasTy, line 132: IMHO: This calls for a separate section to increase visibility of this fact.
But I'd suggest approaching further migration docs refactor with a separate PR.
You could consider following the text style similar to MUI Core or pickers, where migration doc (sections) tell the user what will need to be done to migrate, instead of what has been done.

LukasTy (approved):
Looking great! 🙏 👍 

Left a couple possible improvements (could be future refactors).

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
MBilalShafi: Yep, agreed. The next line should be enough to make the fact clear that some sub-portions may require some additional work.
I was trying to find a different emojie for such cases, something like a yellow check mark but couldn't find a suitable one 😄

packages/x-codemod/src/v6.0.0/pickers/adapter-change-import/actual.spec.js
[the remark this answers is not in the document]
LukasTy: Those changes are probably what caused the test failures. 🤔

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
MBilalShafi: [This case](https://mui.com/material-ui/migration/v5-component-changes/#remove-withmobiledialog-component) in core seemed similar where part of the change was included in `preset-safe`, handled our case the same way by removing ✅ and adding a warning so that users are aware that some extra work may be needed.

LukasTy: Gotcha. That's also a solution, especially if it worked for MUI Core. 👍 
Although, I'd consider splitting it into a separate section or two different sub-sections (like with MUI Core: Component -> specific change).
Or maybe using a different icon, like you mentioned, something along the lines of a warning, or maybe a checkmark and a warning. 🤷 
But in any case, those are just thinking out loud nitpicks. 😉

MBilalShafi: Sure thanks for sharing the ideas. They are good and we can consider them with upcoming PRs 👍

PR #7369, opened 2023-01-02
[fields] Rename the `input` slot of the fields `textField` to avoid confusion

flaviendelangle:
Discussed in today's meeting

The main goal is to avoid the confusion between the `DateField.components.Input` slot and `TextField.inputProps` or `TextField.InputProps`.


packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.types.ts
```diff
@@ -49,10 +49,11 @@ export interface MultiInputDateRangeFieldSlotsComponent {
    */
   Root?: React.ElementType;
   /**
-   * Input rendered for the start or end date.
+   * Component rendering an HTML input and a label.
+   * Will be rendered twice: once for the start date and once for the end date.
```
alexfauquette, line 53:
```suggestion
   * Form control rendered twice: once for the start date and once for the end date.
   * Should have similar interface as `@mui/material/TextField`
```
The `form control` comes from the introduction of [`TextField`](https://mui.com/material-ui/react-text-field/#basic-textfield)
Not sure about the second line

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.types.ts
```diff
@@ -26,7 +26,12 @@ export interface UseDesktopRangePickerSlotsComponent<TDate, TView extends DateOr
   Field: React.ElementType;
   FieldRoot?: React.ElementType<StackProps>;
   FieldSeparator?: React.ElementType<TypographyProps>;
-  Input?: React.ElementType<TextFieldProps>;
+  /**
+   * Component rendering an HTML input and a label inside the smart field.
```
alexfauquette, line 30:
```suggestion
   * Component rendering an HTML input and a label inside the default field.
```

When reading that it let me think that the most important aspect is that it's the default ones. If they use a custom field and do not use `components` then it does not matter

alexfauquette (approved):
Small wording proposal, but the code looks good

docs/data/date-pickers/date-picker/CustomInput.js
```diff
@@ -54,7 +54,7 @@ export default function CustomInput() {
         label="Custom input"
         defaultValue={dayjs('2022-04-07')}
         components={{
-          Input: BrowserInput,
+          TextField: BrowserInput,
```
LukasTy, line 57: Thinking about it with a fresh head. I think I've come up with another reasonable naming option—`FieldInput`.
Pros over `TextField`:
- not directly related to the actual component used for the slot
- technically the field might not be text if the date format does not include date sections with words
- it's closed to what the actual slot replaces—a elaborate input component

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import Stack, { StackProps } from '@mui/material/Stack';
-import TextField from '@mui/material/TextField';
+import MaterialTextField from '@mui/material/TextField';
```
LukasTy, line 4: It seems that we do use `Mui` prefix if we need for these components for the cases we currently have. 🤔 
```suggestion
import MuiTextField from '@mui/material/TextField';
```

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.types.ts
LukasTy: Yeah, I'd also be cautions about using this `smart` word... It's probably the only place we have this description for the field inputs 🙈

packages/x-date-pickers-pro/src/MultiInputDateTimeRangeField/MultiInputDateTimeRangeField.types.ts
```diff
@@ -51,10 +51,11 @@ export interface MultiInputDateTimeRangeFieldSlotsComponent {
    */
   Root?: React.ElementType;
   /**
-   * Input rendered for the start or end date.
+   * Component rendering an HTML input and a label.
+   * Will be rendered twice: once for the start date-time and once for the end date-time.
```
LukasTy, line 55: I don't see us using `date-time` in written form, only for code or url parts. 🤔 
```suggestion
   * Will be rendered twice: once for the start date time and once for the end date time.
```

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.types.ts
LukasTy: Good point. We can explain more precisely what the slot actually does by default and then it's up to the user to decide what they want to do with it. 🤔 
In regards to the second line. Maybe we could format it in the following way:
```
Receives props in `import { TextFieldProps } from '@mui/material/TextField'` format.
```
Or something along these lines 🤔

LukasTy (commented):
Looking way better! 👍 
Leaving a few possible improvement comments as well as a possible slot rename suggestion. 🙈

packages/x-date-pickers-pro/src/MultiInputDateRangeField/MultiInputDateRangeField.types.ts
flaviendelangle: When I read `Form control`, my first guess is that it only contain the `FormControl` but not the actual input.
I added a small precision.

I pushed a new wording with these feedbacks.

packages/x-date-pickers/src/DateField/DateField.types.ts
```diff
@@ -53,12 +53,13 @@ export type DateFieldOwnerState<TDate> = DateFieldProps<TDate>;
 
 export interface DateFieldSlotsComponent {
   /**
-   * Input rendered.
-   * @default TextField
+   * Form control with an input to render the value.
+   * Receives the same props as '@mui/material/TextField'.
```
LukasTy, line 57:
```suggestion
   * Receives the same props as `@mui/material/TextField`.
```
Same suggestion would apply to all components having this JSDoc.
Screenshot 2023-01-03 at 14 03 55

LukasTy (approved):
Looks great if we are going with `TextField` and not even considering [this suggestion](https://github.com/mui/mui-x/pull/7369/files#r1060322151). 👍

packages/x-date-pickers/src/DateField/DateField.types.ts
flaviendelangle: Done :+1:

docs/data/date-pickers/date-picker/CustomInput.js
flaviendelangle: It was also my first idea before yesterday's meeting.
But I would prefer avoiding the term `Input` altogether :+1:

Issue #7372, opened 2023-01-02
[pickers] Remove `WrapperVariantContext` usage


LukasTy:
Created from https://github.com/mui/mui-x/pull/7105#discussion_r1048500282

`WrapperVariantContext` serves little purpose and in some cases causes issues (inconsistencies) if a component is used without this context.

Remove or replace by a prop from:
- [x] DateTimePickerTabs @LukasTy #7374
- [x] CalendarOrClockPicker `Will be removed with legacy pickers`
- [x] PickersMonth  @flaviendelangle #7382
- [x] Clock #8083
- [x] PickersYear @flaviendelangle #7382 
- [x] YearCalendar @flaviendelangle  #7382
- [x] DateRangeCalendar @flaviendelangle #7488
- [x] DateRangePickerView `Will be removed with legacy pickers`
- [ ] Remove `WrapperVariantContext` after legacy pickers are removed

flaviendelangle:
I'm tagging my self on the one I'm working on :+1:

PR #7374, opened 2023-01-02
[pickers] Remove `WrapperVariantContext` from `DateTimePickerTabs`

LukasTy:
Part of #7372

### DIFF
--- a/packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
+++ b/packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
@@ -5,10 +5,6 @@ import Tabs, { tabsClasses } from '@mui/material/Tabs';
 import { styled, useThemeProps } from '@mui/material/styles';
 import { unstable_composeClasses as composeClasses } from '@mui/utils';
 import { Time, DateRange } from '../internals/components/icons';
-import {
-  WrapperVariantContext,
-  WrapperVariant,
-} from '../internals/components/wrappers/WrapperVariantContext';
 import { DateOrTimeView } from '../internals/models';
 import { useLocaleText } from '../internals/hooks/useUtils';
 import {
@@ -16,6 +12,7 @@ import {
   getDateTimePickerTabsUtilityClass,
 } from './dateTimePickerTabsClasses';
 import { BaseTabsProps, ExportedBaseTabsProps } from '../internals/models/props/tabs';
+import { WrapperVariant } from '../internals/models/common';
 
 type TabValue = 'date' | 'time';
 
@@ -51,6 +48,11 @@ export interface ExportedDateTimePickerTabsProps extends ExportedBaseTabsProps {
    * @default Time
    */
   timeIcon?: React.ReactNode;
+  /**
+   * Variant of the component this component is inside of.
+   * @default 'desktop'
+   */
+  wrapperVariant?: WrapperVariant;
 }
 
 export interface DateTimePickerTabsProps
@@ -62,9 +64,7 @@ export interface DateTimePickerTabsProps
   classes?: Partial<DateTimePickerTabsClasses>;
 }
 
-type OwnerState = DateTimePickerTabsProps & { wrapperVariant: WrapperVariant };
-
-const useUtilityClasses = (ownerState: OwnerState) => {
+const useUtilityClasses = (ownerState: DateTimePickerTabsProps) => {
   const { classes } = ownerState;
   const slots = {
     root: ['root'],
@@ -77,7 +77,7 @@ const DateTimePickerTabsRoot = styled(Tabs, {
   name: 'MuiDateTimePickerTabs',
   slot: 'Root',
   overridesResolver: (_, styles) => styles.root,
-})<{ ownerState: OwnerState }>(({ ownerState, theme }) => ({
+})<{ ownerState: DateTimePickerTabsProps }>(({ ownerState, theme }) => ({
   boxShadow: `0 -1px 0 0 inset ${(theme.vars || theme).palette.divider}`,
   ...(ownerState.wrapperVariant === 'desktop' && {
     // TODO v6: Drop `order` with the legacy pickers
@@ -98,12 +98,12 @@ const DateTimePickerTabs = function DateTimePickerTabs(inProps: DateTimePickerTa
     timeIcon = <Time />,
     view,
     hidden = typeof window === 'undefined' || window.innerHeight < 667,
+    wrapperVariant = 'desktop',
   } = props;
 
   const localeText = useLocaleText();
-  const wrapperVariant = React.useContext(WrapperVariantContext);
   const ownerState = { ...props, wrapperVariant };
-  const classes = useUtilityClasses(ownerState);
+  const classes = useUtilityClasses(props);
 
   const handleChange = (event: React.SyntheticEvent, value: TabValue) => {
     onViewChange(tabToView(value));
--- a/packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
+++ b/packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
@@ -169,6 +169,11 @@ DateTimePickerTabs.propTypes = {
    * Currently visible picker view.
    */
   view: PropTypes.oneOf(['day', 'hours', 'minutes', 'month', 'seconds', 'year']).isRequired,
+  /**
+   * Variant of the component this component is inside of.
+   * @default 'desktop'
+   */
+  wrapperVariant: PropTypes.oneOf(['desktop', 'mobile']),
 } as any;
 
 export { DateTimePickerTabs };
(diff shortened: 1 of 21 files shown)


packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
```diff
@@ -51,6 +48,11 @@ export interface ExportedDateTimePickerTabsProps extends ExportedBaseTabsProps {
    * @default Time
    */
   timeIcon?: React.ReactNode;
+  /**
+   * Variant of the component this component is inside of.
+   * @default 'desktop'
+   */
+  wrapperVariant?: WrapperVariant;
```
LukasTy, line 55: Did experiment with making this prop private, but then casting is needed in some places. And I'm not sure if we should stop users from overriding it if they so need. 🤔

flaviendelangle:
Couldn't we handle the styling at the `PickersLayoutContentWrapper` level ?
Set `flexDirection: 'column-reverse'` and set the shadow there or change the rendering order (depending on what keyboard order we want).

Because with the current approach, `PickersLayout` handled all the positioning and the tabs say "no no I do want to be below"

LukasTy:
> Couldn't we handle the styling at the `PickersLayoutContentWrapper` level ?

We probably could, but that was not the intent of this PR.
I'll look into this option as it could probably be even more clean solution if we could use layout and avoid the need for `wrapperVariant` on the tabs. 😃 👍

packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.tsx
```diff
@@ -60,7 +60,11 @@ const DesktopNextDateTimePicker = React.forwardRef(function DesktopNextDateTimeP
         label: defaultizedProps.label,
         ampm: defaultizedProps.ampm,
       }),
-    },
+      tabs: {
+        underlineLocation: 'top',
+        ...defaultizedProps.componentsProps?.tabs,
+      },
+    } as DesktopNextDateTimePickerProps<TDate>['componentsProps'],
```
LukasTy, line 67: I've added casting as a least intrusive way of having type inference for props without re-engineering props typing to work with the hook below.

packages/x-date-pickers/src/MobileDateTimePicker/MobileDateTimePicker.tsx
```diff
@@ -101,7 +101,13 @@ export const MobileDateTimePicker = React.forwardRef(function MobileDateTimePick
         autoFocus
         DateInputProps={DateInputProps}
         components={components}
-        componentsProps={componentsProps}
+        componentsProps={{
+          ...componentsProps,
+          tabs: {
+            underlineLocation: 'bottom',
```
flaviendelangle, line 107: Why do we need to define it here ?
Couldn't it be the layout that says "the tabs is below so the underline is above" ?
Since it already has the logic

Or if we are changing the rendering order (which appears to be the case in `PickersLayout`, maybe simply use `:last-child` inside the toolbar and remove the prop altogether.

flaviendelangle: My 2nd proposal was keeping the style inside the toolbar itself, but deduce it from the DOM position, not from a prop

LukasTy: Hm. I've played around with your second suggestion. Thanks for an option I definitely did not think about.
Indeed, it does seem like the most correct solution, especially when now we are going to have correct tabulation/DOM order. 👍

packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
```diff
@@ -101,9 +93,7 @@ const DateTimePickerTabs = function DateTimePickerTabs(inProps: DateTimePickerTa
   } = props;
 
   const localeText = useLocaleText();
-  const wrapperVariant = React.useContext(WrapperVariantContext);
-  const ownerState = { ...props, wrapperVariant };
-  const classes = useUtilityClasses(ownerState);
+  const classes = useUtilityClasses(props);
```
flaviendelangle, line 96: Most of the time, we keep the `const ownerState = props` 

I don't have a strong opinion on this though

packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.tsx
flaviendelangle: Do we agree that the typing works fine without those casts ?
What are you trying to achieve with those casts ?


My main problem with keeping them without a clear gain, is that is reduces the type safety.
For example if I do a typo on a property I'm passing, without the cast it complains:

image

With the cast it does not complain:

image


We had a real scenario a few days ago with the range toolbar where I had a typo in a variable name that was swallowed by the cast.

If we need to pass a prop that the slot does not know, I would at least put the cast at the slot level, on all the slots

image

Or even apply a single `@ts-ignore` to really avoid impacting the rest of the typing.

packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
```diff
@@ -84,8 +84,17 @@ const PickersLayout = function PickersLayout<TValue, TView extends DateOrTimeVie
     >
       {toolbar}
       <PickersLayoutContentWrapper className={classes.contentWrapper}>
-        {tabs}
-        {content}
+        {wrapperVariant === 'desktop' ? (
```
flaviendelangle, line 87: I don't use the array format a lot with JSX, but with variables like this it's lighter imho:

```suggestion
        {wrapperVariant === 'desktop' ? [content, tabs] : [tabs, content]}
```

packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.tsx
LukasTy: Hm... Fair point. We can't have both benefits with as casting...
I'll try going with the new [satisfies operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html). Seems that it should do the trick of covering both requirements. 👌

flaviendelangle: If it does that's a far better solution that anything we both proposed :star_struck:

LukasTy: Need to clarify if it's safe to use this operator, but seems to be the ideal solution for such cases. 👍 
Screenshot 2023-01-06 at 17 21 49

LukasTy: These `satisfies` do not end up in production code in any way.
They seem safe to use. 👌

packages/x-date-pickers/src/DateTimePicker/DateTimePickerTabs.tsx
LukasTy: 66 vs 13 instances, but also most of the time where `ownerState` is passed—it is an extension of `props` or even a custom object built from other props.
IMHO, I see little to no sense in creating a temporary variable reference copy just to pass the same data to a function. 🤔 
In cases where we reuse the `ownerState` for other reasons (need to pass `ownerState` somewhere else) or augment the `props` object with extra info, then I do totally agree with using `ownerState`. 👌

packages/x-date-pickers/src/PickersLayout/PickersLayout.tsx
LukasTy: Sadly, doing this produces a missing `key` warning.
I'd love this much cleaner approach, but it's not easy to apply it in this case. 🤔 
Screenshot 2023-01-09 at 10 27 29

flaviendelangle (approved):
As soon as we have the `satisfies` instead of the type casting I'm good with merging :+1:

LukasTy:
> As soon as we have the `satisfies` instead of the type casting

I initially did not notice the compilation error. I'm trying to understand why it might be failing to compile on our codebase, even though we have most of the latest libraries. 🤔 
https://stackoverflow.com/questions/74677993/typescript-react-app-is-throwing-error-when-using-the-new-satisfies-operator-w

packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.tsx
LukasTy: Will leave this place as is for the time being, because of a [known bug in next](https://github.com/vercel/next.js/issues/43799).
It works without problems in IDE as well as build, but dev and test do fail.

PR #7339, opened 2022-12-29
[docs] Cleanup the doc pages of the community pickers

flaviendelangle:
A lot of small changes to unify the doc of the Date Picker, Time Picker and Date Time Picker.
I'll keep the Date Range Picker for later, as it needs a Date Range Calendar dedicated page.

The Date Picker page still have 2 demos I'd like to move to other pages (the helper text and the custom input one).
But I think this PR is big enough as it is.

PR #7380, opened 2023-01-03
[pickers] Hide am/pm controls when their is no hour view

flaviendelangle:
While working on #7339, I noticed that a time picker with just the `minutes` and `seconds` view was renderting `MM:SS AM` in the toolbar, which is confusing.


alexfauquette (approved):
Stange use case, but make sense 👍

PR #7382, opened 2023-01-03
[pickers] Stop using the `WrapperVariantContext` in `MonthCalendar` and `YearCalendar`

flaviendelangle:
Part of #7372


packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
```diff
@@ -24,6 +24,11 @@ export interface DesktopNextDatePickerProps<TDate>
   extends BaseNextDatePickerProps<TDate>,
     DesktopOnlyPickerProps<TDate>,
     BaseNextNonStaticPickerExternalProps {
+  /**
+   * Years rendered per row.
+   * @default 4
+   */
+  yearsPerRow?: 3 | 4;
```
LukasTy, line 31: Do you really think it's worth redefining these props just to have a more correct `default`?
I wouldn't mind having the following default. 🤔 
```
`4` on desktop, `3` on mobile.
```

packages/x-date-pickers/src/MonthCalendar/PickersMonth.tsx
```diff
@@ -29,17 +26,16 @@ interface PickersMonthProps extends ExportedPickersMonthProps {
...
   tabIndex: number;
+  monthsPerRow: 3 | 4;
 }
 
-interface PickersMonthOwnerState extends PickersMonthProps {
-  wrapperVariant: WrapperVariant;
-}
+type PickersMonthOwnerState = PickersMonthProps;
 
 const useUtilityClasses = (ownerState: PickersMonthOwnerState) => {
```
LukasTy, line 34: What do you think about dropping `ownerState` and just using `props` here? 🤔

packages/x-date-pickers/src/YearCalendar/pickersYearClasses.ts
```diff
@@ -24,10 +20,8 @@ export function getPickersYearUtilityClass(slot: string) {
   return generateUtilityClass('MuiPickersYear', slot);
 }
 
-export const pickersYearClasses = generateUtilityClasses('MuiPickersYear', [
+export const pickersYearClasses = generateUtilityClasses<PickersYearClassKey>('MuiPickersYear', [
   'root',
-  'modeDesktop',
-  'modeMobile',
```
LukasTy, line 30: This and same change in `pickersMonthClasses` essentially makes it a public API breaking change.
We should probably add some breaking change doc and mention in PR..? 🤔

LukasTy (commented):
Looking great ❤️ 
Leaving a few nitpick comments.

packages/x-date-pickers/src/DatePicker/DatePicker.tsx
```diff
@@ -32,6 +32,11 @@ export interface DatePickerProps<TDate>
    * @example '@media (min-width: 720px)' or theme.breakpoints.up("sm")
    */
   desktopModeMediaQuery?: string;
+  /**
+   * Years rendered per row.
+   * @default 4 on desktop, 3 on mobile
```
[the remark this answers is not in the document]
LukasTy: By the way, these props and defaults seem to be lost somewhere in the nebula 🤷 
Screenshot 2023-01-10 at 16 43 26

https://codesandbox.io/s/sweet-hellman-c6lj5k?file=/demo.tsx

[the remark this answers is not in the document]
flaviendelangle: I think I introduced a bug when merging with `next` this morning :/

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
flaviendelangle: If you use directly `YearCalendar` or `DateCalendar` this JSDoc would be wrong :/

packages/x-date-pickers/src/DatePicker/DatePicker.tsx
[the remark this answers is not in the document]
flaviendelangle: Fixed :+1:

packages/x-date-pickers/src/YearCalendar/pickersYearClasses.ts
flaviendelangle: Yes, I was waiting for us to agree on the change before discussing the migration guide.

I think this class make no real sense (why put it on those elements and not on others ?)

packages/x-date-pickers/src/MonthCalendar/PickersMonth.tsx
flaviendelangle: I'm fine with it
The core uses `const ownerState = props` fairly often (and never `ownerState={props}`)
But as long as we try to be coherent inside X.

packages/x-date-pickers/src/YearCalendar/pickersYearClasses.ts
LukasTy: > I think this class make no real sense (why put it on those elements and not on others ?)

I really don't know as well... Maybe there once was a request to allow for the items per row to be easily controlled via CSS? 🤔

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.tsx
```diff
@@ -45,6 +45,7 @@ const DesktopNextDatePicker = React.forwardRef(function DesktopNextDatePicker<TD
     viewRenderers,
     format: getDatePickerFieldFormat(utils, defaultizedProps),
     showToolbar: defaultizedProps.showToolbar ?? false,
+    yearsPerRow: 4,
```
LukasTy, line 48: `yearPerRow` still does not seem to like being overridden on new picker component. 🤷 
Screenshot 2023-01-11 at 13 32 01
https://codesandbox.io/s/sweet-hellman-c6lj5k?file=/demo.tsx

flaviendelangle: :facepalm: sorry I fixed it to quickly
Fixed

flaviendelangle:
@LukasTy I added a migration guide
I think almost nobody uses this so maybe we can even skip the variable export and hardcode the media query on the `diff`.

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.tsx
LukasTy: Does still seem to be an issue with `StaticNextDatePicker` and `StaticNextDateTimePicker`. 🙈

PR #7383, opened 2023-01-04
[field] Prevent date change on `TimeField` arrow edition

flaviendelangle:
[Behavior before](https://codesandbox.io/s/date-and-time-pickers-forked-fepeji)
[Behavior after](https://codesandbox.io/s/date-and-time-pickers-forked-onxvj9)

Press `ArrowUp`, the old one should go to the 8th of April, not the new one.

Maybe some people want the old behavior and if so we could introduce a prop to support it.
But the new behavior feels more coherent to me, especially with time range pickers were the behavior is very weird when we have `12:00 - 15:00` rendered but the field is in error because the `15:00` is actually in the day before.

### DIFF
--- a/packages/x-date-pickers-pro/src/SingleInputDateRangeField/useSingleInputDateRangeField.ts
+++ b/packages/x-date-pickers-pro/src/SingleInputDateRangeField/useSingleInputDateRangeField.ts
@@ -70,6 +70,6 @@ export const useSingleInputDateRangeField = <TDate, TChildProps extends {}>({
     valueManager: rangeValueManager,
     fieldValueManager: rangeFieldValueManager,
     validator: validateDateRange,
-    supportedDateSections: ['year', 'month', 'day'],
+    valueType: 'date',
   });
 };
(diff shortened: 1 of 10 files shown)


packages/x-date-pickers-pro/src/SingleInputDateRangeField/useSingleInputDateRangeField.ts
```diff
@@ -70,6 +70,6 @@ export const useSingleInputDateRangeField = <TDate, TChildProps extends {}>({
     valueManager: rangeValueManager,
     fieldValueManager: rangeFieldValueManager,
     validator: validateDateRange,
-    supportedDateSections: ['year', 'month', 'day'],
```
flaviendelangle, line 73: I think this new abstraction is simpler to understand and easier to reason with if we have logic to apply like the one of this PR.

packages/x-date-pickers/src/internals/hooks/useField/useField.ts
```diff
@@ -294,15 +295,24 @@ export const useField = <
...
               activeSection.dateSectionName,
               event.key as AvailableAdjustKeyCode,
-            ),
-            shouldGoToNextSection: false,
-          }),
+            );
+
+            // If the field only supports time editing, then we should never go to the previous / next day.
+            if (valueType === 'time') {
+              date = utils.mergeDateAndTime(activeDate, date);
```
flaviendelangle, line 308: The fix is here

LukasTy (approved):
Very nice fix! 💯

PR #7387, opened 2023-01-04
[docs] Fix the `DemoContainer` behavior in production build

flaviendelangle:
The `DemoContainer` don't work in production because it can't determine the name of its children and build its layout based on it.

- [x] Add a `displayName` to all pickers / fields / clocks / calendars / localization provider.
- [x] Throw an error when a child of `DemoContainer` has no display name.
- [x] Wrap the child in `DemoItem` even when no label (done inside `DemoContainer`) to take advantage of the alignment for multi input components.


LukasTy (commented):
Personally—I'm not a fan of having to add additional code and ship it to users just to have a more correct demo layout. 🤔 
Seems to somewhat cause a conflict of interest...
It will also have a strange result of having some correct component names in production react dev tools, while others will be gibberish. 🤔 

But in any case, if you did not think of any other way to achieve what we need—it's not the end of the world. We are already shipping way bigger size of PropTypes. 🙈 😆 

By the way, could we maybe fix the styling for the following as we are "already here" 😃 
Screenshot 2023-01-10 at 18 07 17
Screenshot 2023-01-10 at 18 07 51

flaviendelangle:
> But in any case, if you did not think of any other way to achieve what we need

The only other way I see is to go one level lower in abstraction and add a `componentNames` prop on `DemoContainer` for example.

LukasTy:
> The only other way I see is to go one level lower in abstraction and add a `componentNames` prop on `DemoContainer` for example.

What values would `componentNames` contain? Isn't the problem with the fact that we don't have a reliable component name in production build, just random characters..? 🤔

flaviendelangle:
> What values would componentNames contain? Isn't the problem with the fact that we don't have a reliable component name in production build, just random characters..?

Yes that's the issue
`DemoContainer` need to know what type of component it contains in order to decide how many item per row to render (and a few other things).

Right now I'm using the name of the children.
But in production it does not work.
To fix it, we can make the name of the children work in production (`displayName`)
Or explicitly pass those names to `DemoContainer`

```tsx
<DemoContainer components={['NextDatePicker', 'NextTimePicker', 'TimeField']}>
```

That's not very appealing for sure... And it can be error prone

LukasTy:
> Or explicitly pass those names to `DemoContainer`

Oh yes, that's definitely not an elegant solution... 🙈 

After all, it's always the issue when we try to go the abstraction route.

On the other hand, we could also say, that having explicit component and style configuration in the demo would have it's own benefits of showing examples of how to combine `@mui/material` components with `@mui/x-*`.
Basically, more ways to educate the user. 🤔 
Also, it seems that we might not even be able to fix [Stackblitz demos](https://github.com/stackblitz/core/issues/437) using `DemoContainer`. 😢

flaviendelangle:
My main fear of explicitly defining the rule in the demo is to loose consistency.
Since I created the `DemoContainer`, several time I changes the rules because it was not working in some scenario I did not anticipated.
And it applied automatically to all the demos which is super nice.

Passing the name of the components, at least keeps all the logic inside `DemoContainer`.

LukasTy:
> Passing the name of the components, at least keeps all the logic inside `DemoContainer`.
> I would not go even lower in abstraction with stuff like `<DemoContainer componentTypes={['picker', 'multi-input-field']} />` because the type of components evolve with time and updating everywhere will be a pain.

Maybe I'm just pretentious, but I would lean towards this unclean solution of defining component names on the `DemoContainer` in case we can't overcome the stackblitz limitation and wouldn't want to ship this component in the bundle. 🤔 But I still have hope, because StakcBlitz documentation clearly shows that specifying deps and their versions should work. 🤷

LukasTy:
> Indeed, I did not know about this limitation

So, I'm afraid I'm coming with bad news... No matter what you try and with which API, stackblitz refuses to add a dependency if it does not exist as an `npm` package. 😭 

We are basically left with three options:
1. Try pulling the `DemoContainer` file when creating a code sandbox, inject it as one of the files and replace the import path
2. Go with the solution discussed in this PR of having explicit `components` defined on `DemoContainer`
3. Ship `DemoContainer` with `@mui/x-date-pickers` 

Those options are basically in the priority of increasing "evilness" level, hence I'm gonna experiment with option 1 firstly, but it is also one of the most limiting, because the code is in `@mui/monorepo` and any change to file path or something else would result in changes needed to to this handling as well. 🤔

LukasTy:
> I don't understand the link between stackblitz and this PR. How does putting `DemoContainer` in `@mui/x-date-pickers` solves the name of component issue in production ?

I just replied here because we were discussing this fact and possibility to go with a different solution for DemoContainer. 🤔

LukasTy:
Superseded by #7574 
This solution wouldn't have been ideal as well, because of [nasty package size increase](https://github.com/mui/mui-x/pull/7387#issuecomment-1400315840).

PR #7388, opened 2023-01-04
[docs] Fix codesandboxes using `DemoContainer`

### LINKED ISSUE #7371: [docs] Pickers codeSandbox broken when using import `DemoContainer`/`DemoItem`
### Duplicates

### Latest version

### Steps to reproduce 🕹

Go to any demonstration using `DemoContainer` or `DemoItem` and open the codesandbox.

It will fail to load some dependencies.

It comes from the following line:
```js
import { DemoContainer, DemoItem } from 'docsx/src/modules/components/DemoContainer';
```

When opening the codesanbox, docs script read it and adds to `package.json` the following line: `"docsx": "latest"`

The codesandbox script should be modified to either:
remove this line and define the components at the top of the document, or put their definition in another file, and modify the import line

### Current behavior 😯

_No response_

### Expected behavior 🤔

_No response_

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #7371

After long and various experiments, we've concluded that the least evil solution of them all is to ship `DemoContainer` in the bundle.

- move `DemoContainer` to `x-date-pickers/src/internals/demo/` and re-export to be available at `@mui/x-date-pickers/internals/demo`
- replace imports


docs/data/date-pickers/date-range-picker/ResponsiveDateRangePicker.js
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import dayjs from 'dayjs';
-import { DemoContainer, DemoItem } from 'docsx/src/modules/components/DemoContainer';
+import { DemoContainer, DemoItem } from '@mui/x-date-pickers/demo/DemoContainer';
```
flaviendelangle, line 3: We probably want  to import from `'@mui/x-date-pickers/demo'`
Imports of depth 3 are not supported by our bundling strategy so it might not work depending on how codesandbox and/or Stackblitz import it (or if users want to import it).

LukasTy: CodeSandbox as well as StackBlitz does not seem to have issues with it. 🤔 
And maybe it's even better that users wouldn't be able to easily import it? 😆 🙈

flaviendelangle: If we consider it to be private API, maybe name it `internal-demo` or something like that ? 
To avoid people complaining when we break it.
If it work on both website I guess it's fine to keep the current depth.

If we consider it to be public API, then I'd go for a reduced depth.

LukasTy: So, maybe put it in `internals/demo` and re-export at `internals` level? 🤔

flaviendelangle: If you re-export at `internals` level it will be bundled on the pro package.
But put it in `internals/demo` is fine for me.

LukasTy: Moved to `internals/demo/` and re-exported to be available at `internals/demo` 😉

flaviendelangle (approved):

alexfauquette (approved):
Effectively it's much easier than the initial ideas 👍

PR #7390, opened 2023-01-04
[pickers] Support components and slots for new pickers

alexfauquette:
### About components

- internal components accept `components`, `slots`, and `slotsProps`. The `componentsProps` is removed as a duplicate of `slotsProps`.
- Props of legacy components are not modified.
- New components accept the 4 props.

### About tests

- When the goal is to test that `components` or `componentsProps` are working correctly, the test is duplicated to have it with `slots`.
- When the slots are not the main target of the test, `components` is replaced by `slots`.


flaviendelangle (commented):
Just in case, the slot nomenclature also includes a capitalization of the `slotsProps` (`Layout` instead of `layout`).
And that's going to make the migration a lot more verbose.

flaviendelangle:
The following seems to work

```tsx
type CapitalizeKeys<T extends object> = Capitalize<keyof T & string>;

type CapitalizeObjectKeys<T extends object> = {
  [key in CapitalizeKeys<T>]: Uncapitalize<key> extends keyof T ? T[Uncapitalize<key>] : never;
}

export interface ExportedPickersArrowSwitcherProps {
  /**
   * Overrideable components.
   * @default {}
   */
  components?: PickersArrowSwitcherSlotsComponent;
  /**
   * The props used for each component slot.
   * @default {}
   */
  componentsProps?: PickersArrowSwitcherSlotsComponentsProps;
  slotsProps?: CapitalizeObjectKeys<PickersArrowSwitcherSlotsComponentsProps>
  classes?: Partial<PickersArrowSwitcherClasses>;
}
```

We would need to check the TS version required and if the doc generation works correctly.

packages/x-date-pickers/src/TimeClock/TimeClock.tsx
```diff
@@ -137,7 +150,7 @@ const TimeClockRoot = styled(PickerViewRoot, {
   flexDirection: 'column',
 });
 
-const TimeCLockArrowSwitcher = styled(PickersArrowSwitcher, {
+const TimeClockArrowSwitcher = styled(PickersArrowSwitcher, {
```
alexfauquette, line 153: Typo fix

flaviendelangle: :laughing: I fixed it in #7380 as well

packages/x-date-pickers/src/MobileNextTimePicker/MobileNextTimePicker.types.ts
```diff
@@ -27,11 +28,23 @@ export interface MobileNextTimePickerProps<TDate>
   /**
    * Overrideable components.
    * @default {}
+   * @deprecated
    */
-  components?: MobileNextTimePickerSlotsComponent<TDate>;
+  components?: Partial<MobileNextTimePickerSlotsComponent<TDate>>;
```
flaviendelangle, line 33: Quel est le besoin résolue par `Partial` ici ?
J'ai retiré tous les `Partial` parce que certains slots peuvent être requis dans certains cas donc je trouve plus simple d'avoir un `?:` au niveau de chaque slot, sinon on peut vite être bloqué.

flaviendelangle: See #6463

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
```diff
@@ -27,11 +28,23 @@ export interface DesktopNextDatePickerProps<TDate>
...
    * The props used for each component slot.
    * @default {}
+   * @deprecated
    */
   componentsProps?: DesktopNextDatePickerSlotsComponentsProps<TDate>;
+  /**
+   * Overrideable components.
+   * @default {}
+   */
+  slots?: UncapitalizeObjectKeys<DesktopNextDatePickerSlotsComponent<TDate>>;
```
LukasTy, line 44: Could maybe the `SlotsAndSlotsProps` utility type be used for these cases as well? 🤔 
I see it used in some places, but in others there are lots of duplication. 🤔

packages/x-date-pickers/src/PickersLayout/usePickerLayout.tsx
```diff
@@ -117,10 +119,14 @@ const usePickerLayout = <TValue, TView extends DateOrTimeView>(
...
 
-  const Tabs = components?.Tabs;
+  const Tabs = slots?.tabs ?? components?.Tabs;
   const tabs =
     view && Tabs ? (
-      <Tabs view={view} onViewChange={onViewChange} {...componentsProps?.tabs} />
+      <Tabs
+        view={view}
+        onViewChange={onViewChange}
+        {...(slotsProps?.tabs ?? componentsProps?.tabs)}
```
LukasTy, line 128: Maybe going with the following pattern used in some other places would result in a cleaner result? 🤔 

```tsx
  ...
  components,
  componentsProps,
  slots,
  slotsProps: innerSlotsProps,
} = props;

const slotsProps = innerSlotsProps ?? componentsProps;
...
  externalSlotProps: slotsProps?.toolbar,
```

packages/x-date-pickers/src/StaticNextDateTimePicker/StaticNextDateTimePicker.tsx
```diff
@@ -41,11 +41,11 @@ const StaticNextDateTimePicker = React.forwardRef(function StaticNextDateTimePic
     viewRenderers,
     displayStaticWrapperAs,
     showToolbar: defaultizedProps.showToolbar ?? displayStaticWrapperAs === 'mobile',
```
LukasTy, line 43: @flaviendelangle @alexfauquette 
Shouldn't this prop end up in the same place we migrated the `hideTabs` prop? 🤔 

```diff
- hideTabs={false}
- showToolbar
+ slotsProps={{
+   tabs: { hidden: false },
+   toolbar: { hidden: false },
+ }}
```

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
```diff
@@ -394,9 +399,12 @@ export function PickersPopper(inProps: PickerPopperProps) {
           disableRestoreFocus
           disableEnforceFocus={role === 'tooltip'}
           isEnabled={() => true}
-          {...componentsProps?.desktopTrapFocus}
+          {...(slotsProps?.desktopTrapFocus ?? componentsProps?.desktopTrapFocus)}
```
LukasTy, line 402: I do really think that for such use cases the `const slotsProps = innerSlotsProps ?? componentsProps;` pattern would allow for cleaner end result. 🤔

packages/x-date-pickers/src/internals/hooks/useStaticPicker/useStaticPicker.tsx
```diff
@@ -30,7 +30,15 @@ export const useStaticPicker = <
   validator,
   ref,
 }: UseStaticPickerParams<TDate, TView, TExternalProps>) => {
-  const { localeText, components, componentsProps, displayStaticWrapperAs, autoFocus } = props;
+  const {
+    localeText,
+    slots,
+    slotsProps,
+    components,
+    componentsProps,
```
LukasTy, line 38: How come for `useMobilePicker` we can only keep `slots` and `slotsProps`, but for `useStaticPicker` we are leaving these props as well? 🤔 
Is it not a mistake by chance?

packages/x-date-pickers/src/internals/utils/slots-migration.ts
```diff
@@ -0,0 +1,65 @@
+// TODO v7: This file exist only to simplify typing between
```
LukasTy, line 1: Nitpick: this file name makes for interesting looking imports. 🙈 
Maybe we could go with something less related to migration? After all, it is going to live in files for the whole v6 lifecycle...
Maybe just `slots-types`? 🤔

LukasTy (commented):
Looking great! 🙌 
Looks like a huge effort. 🙈 
Leaving some comments with suggestions for improvements and some general concerns. 🤔

packages/x-date-pickers-pro/src/internal/hooks/useDesktopRangePicker/useDesktopRangePicker.tsx
```diff
@@ -125,22 +130,19 @@ export const useDesktopRangePicker = <
...
-  const componentsForField: BaseMultiInputFieldProps<DateRange<TDate>, unknown>['components'] = {
-    ...fieldProps.components,
-    TextField: components.TextField,
-    Root: components.FieldRoot,
-    Separator: components.FieldSeparator,
+  const slotsForField: BaseMultiInputFieldProps<DateRange<TDate>, unknown>['slots'] = {
+    ...fieldProps.slots,
+    textField: slots.textField,
+    root: slots.fieldRoot,
+    separator: slots.fieldSeparator,
```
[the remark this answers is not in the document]
flaviendelangle: In theory you could pass slots to the field in two ways:

```tsx
<DatePicker 
  components={{ textField: TextField }}
/>
```

or 

```tsx
<DatePicker 
  componentsProps={{ field: { components: { textField: TextField }}}}
/>
```

With the current code, the 2nd example don't work because the default `components.textField` will always override `componentsProps.field.components.textField`.
With your suggestion, the 2nd example would work.

I don't see any problem applying your suggestion :+1:

packages/x-date-pickers/src/StaticNextDateTimePicker/StaticNextDateTimePicker.tsx
flaviendelangle: I proposed it in https://github.com/mui/mui-x/pull/6445#issuecomment-1276308006

TLDR: We agreed to do it but not on #6445 to keep the PR focused on the migration to slots.

LukasTy: Created [an issue](https://github.com/mui/mui-x/issues/7472) for it. 😉

packages/x-date-pickers/src/DateCalendar/DateCalendar.tsx
```diff
@@ -306,8 +308,9 @@ export const DateCalendar = React.forwardRef(function DateCalendar<TDate>(
         disableFuture={disableFuture}
         reduceAnimations={reduceAnimations}
         labelId={gridLabelId}
+        slots={slots}
         components={components}
-        componentsProps={componentsProps}
+        slotsProps={slotsProps ?? componentsProps}
```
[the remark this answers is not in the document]
alexfauquette: Just to get it simpler inside internal components, But if one day we export them, I agree it worth the extra `??`

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
alexfauquette: It would return the following error
```
Named property 'components' of types 
'BaseNextDatePickerProps<TDate>' and 
'SlotsAndSlotsProps<DesktopNextDatePickerSlotsComponent<TDate>, DesktopNextDatePickerSlotsComponentsProps<TDate>>' 
are not identical
```

Adding `components` solve this error by overriding it. So I need to do the same with `slots`

packages/x-date-pickers/src/internals/utils/slots-migration.ts
alexfauquette: What is in it is only for migration aspect: capitalizing keys. For v7 we should delete it and make typescript pass. If latter on we need utils that are staying (I don't think we will) we should put them in a distinct file

packages/x-date-pickers/src/internals/hooks/useStaticPicker/useStaticPicker.tsx
alexfauquette: Yes, it sounds to be a mistake

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
LukasTy: Gotcha. 👍

LukasTy: I've tinkered with types and to me it still seems worth it to play around the interface extending issue in order to achieve WAY cleaner types for every component. 🤔 
I've tried doing it for several components with 0 issues...
What do you think?
```diff
diff --git a/packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.types.ts b/packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.types.ts
index 599879893..9535d727b 100644
--- a/packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.types.ts
+++ b/packages/x-date-pickers-pro/src/DateRangeCalendar/DateRangeCalendar.types.ts
@@ -15,7 +15,7 @@ import {
   PickersCalendarHeaderSlotsComponentsProps,
   DayCalendarProps,
   ExportedUseViewsOptions,
-  UncapitalizeObjectKeys,
+  SlotsAndSlotsProps,
 } from '@mui/x-date-pickers/internals';
 import { DateRange, RangePositionProps, DayRangeValidationProps } from '../internal/models';
 import { DateRangeCalendarClasses } from './dateRangeCalendarClasses';
@@ -97,7 +97,11 @@ export interface ExportedDateRangeCalendarProps<TDate>
 
 export interface DateRangeCalendarProps<TDate>
   extends ExportedDateRangeCalendarProps<TDate>,
-    Partial<RangePositionProps> {
+    Partial<RangePositionProps>,
+    SlotsAndSlotsProps<
+      DateRangeCalendarSlotsComponent<TDate>,
+      DateRangeCalendarSlotsComponentsProps<TDate>
+    > {
   /**
    * The selected value.
    * Used when the component is controlled.
@@ -121,28 +125,6 @@ export interface DateRangeCalendarProps<TDate>
    * The system prop that allows defining system overrides as well as additional CSS styles.
    */
   sx?: SxProps<Theme>;
-  /**
-   * Overrideable components.
-   * @default {}
-   * @deprecated Please use `slots` with uncapitalized properties.
-   */
-  components?: DateRangeCalendarSlotsComponent<TDate>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   * @deprecated Please use `slotsProps`
-   */
-  componentsProps?: DateRangeCalendarSlotsComponentsProps<TDate>;
-  /**
-   * Overrideable component slots.
-   * @default {}
-   */
-  slots?: UncapitalizeObjectKeys<DateRangeCalendarSlotsComponent<TDate>>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   */
-  slotsProps?: DateRangeCalendarSlotsComponentsProps<TDate>;
 }
 
 export interface DateRangeCalendarOwnerState<TDate> extends DateRangeCalendarProps<TDate> {
diff --git a/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts b/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
index 3e753295c..e501648d0 100644
--- a/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
+++ b/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
@@ -11,7 +11,7 @@ import {
 import { MakeOptional } from '../internals/models/helpers';
 import { BaseNextNonStaticPickerExternalProps } from '../internals/models/props/basePickerProps';
 import { DateView } from '../internals/models/views';
-import { UncapitalizeObjectKeys } from '../internals/utils/slots-migration';
+import { SlotsAndSlotsProps, WithoutSlots } from '../internals/utils/slots-migration';
 
 export interface DesktopNextDatePickerSlotsComponent<TDate>
   extends BaseNextDatePickerSlotsComponent<TDate>,
@@ -22,29 +22,10 @@ export interface DesktopNextDatePickerSlotsComponentsProps<TDate>
     UseDesktopPickerSlotsComponentsProps<TDate, DateView> {}
 
 export interface DesktopNextDatePickerProps<TDate>
-  extends BaseNextDatePickerProps<TDate>,
+  extends WithoutSlots<BaseNextDatePickerProps<TDate>>,
     DesktopOnlyPickerProps<TDate>,
-    BaseNextNonStaticPickerExternalProps {
-  /**
-   * Overrideable components.
-   * @default {}
-   * @deprecated Please use `slots` with uncapitalized properties.
-   */
-  components?: DesktopNextDatePickerSlotsComponent<TDate>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   * @deprecated Please use `slotsProps`
-   */
-  componentsProps?: DesktopNextDatePickerSlotsComponentsProps<TDate>;
-  /**
-   * Overrideable component slots.
-   * @default {}
-   */
-  slots?: UncapitalizeObjectKeys<DesktopNextDatePickerSlotsComponent<TDate>>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   */
-  slotsProps?: DesktopNextDatePickerSlotsComponentsProps<TDate>;
-}
+    BaseNextNonStaticPickerExternalProps,
+    SlotsAndSlotsProps<
+      DesktopNextDatePickerSlotsComponent<TDate>,
+      DesktopNextDatePickerSlotsComponentsProps<TDate>
+    > {}
diff --git a/packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.types.ts b/packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.types.ts
index e8c32ec0d..a91e553fb 100644
--- a/packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.types.ts
+++ b/packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.types.ts
@@ -11,7 +11,7 @@ import {
 import { MakeOptional } from '../internals/models/helpers';
 import { BaseNextNonStaticPickerExternalProps } from '../internals/models/props/basePickerProps';
 import { DateOrTimeView } from '../internals/models/views';
-import { UncapitalizeObjectKeys } from '../internals/utils/slots-migration';
+import { SlotsAndSlotsProps, WithoutSlots } from '../internals/utils/slots-migration';
 
 export interface DesktopNextDateTimePickerSlotsComponent<TDate>
   extends BaseNextDateTimePickerSlotsComponent<TDate>,
@@ -25,29 +25,10 @@ export interface DesktopNextDateTimePickerSlotsComponentsProps<TDate>
     UseDesktopPickerSlotsComponentsProps<TDate, DateOrTimeView> {}
 
 export interface DesktopNextDateTimePickerProps<TDate>
-  extends BaseNextDateTimePickerProps<TDate>,
+  extends WithoutSlots<BaseNextDateTimePickerProps<TDate>>,
     DesktopOnlyPickerProps<TDate>,
-    BaseNextNonStaticPickerExternalProps {
-  /**
-   * Overrideable components.
-   * @default {}
-   * @deprecated Please use `slots` with uncapitalized properties.
-   */
-  components?: DesktopNextDateTimePickerSlotsComponent<TDate>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   * @deprecated Please use `slotsProps`
-   */
-  componentsProps?: DesktopNextDateTimePickerSlotsComponentsProps<TDate>;
-  /**
-   * Overrideable component slots.
-   * @default {}
-   */
-  slots?: UncapitalizeObjectKeys<DesktopNextDateTimePickerSlotsComponent<TDate>>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   */
-  slotsProps?: DesktopNextDateTimePickerSlotsComponentsProps<TDate>;
-}
+    BaseNextNonStaticPickerExternalProps,
+    SlotsAndSlotsProps<
+      DesktopNextDateTimePickerSlotsComponent<TDate>,
+      DesktopNextDateTimePickerSlotsComponentsProps<TDate>
+    > {}
diff --git a/packages/x-date-pickers/src/DesktopNextTimePicker/DesktopNextTimePicker.types.ts b/packages/x-date-pickers/src/DesktopNextTimePicker/DesktopNextTimePicker.types.ts
index 864dfd7f0..346be7362 100644
--- a/packages/x-date-pickers/src/DesktopNextTimePicker/DesktopNextTimePicker.types.ts
+++ b/packages/x-date-pickers/src/DesktopNextTimePicker/DesktopNextTimePicker.types.ts
@@ -11,7 +11,7 @@ import {
 import { MakeOptional } from '../internals/models/helpers';
 import { BaseNextNonStaticPickerExternalProps } from '../internals/models/props/basePickerProps';
 import { TimeView } from '../internals/models/views';
-import { UncapitalizeObjectKeys } from '../internals/utils/slots-migration';
+import { SlotsAndSlotsProps, WithoutSlots } from '../internals/utils/slots-migration';
 
 export interface DesktopNextTimePickerSlotsComponent<TDate>
   extends BaseNextTimePickerSlotsComponent<TDate>,
@@ -22,29 +22,10 @@ export interface DesktopNextTimePickerSlotsComponentsProps<TDate>
     UseDesktopPickerSlotsComponentsProps<TDate, TimeView> {}
 
 export interface DesktopNextTimePickerProps<TDate>
-  extends BaseNextTimePickerProps<TDate>,
+  extends WithoutSlots<BaseNextTimePickerProps<TDate>>,
     DesktopOnlyPickerProps<TDate>,
-    BaseNextNonStaticPickerExternalProps {
-  /**
-   * Overrideable components.
-   * @default {}
-   * @deprecated Please use `slots` with uncapitalized properties.
-   */
-  components?: DesktopNextTimePickerSlotsComponent<TDate>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   * @deprecated Please use `slotsProps`
-   */
-  componentsProps?: DesktopNextTimePickerSlotsComponentsProps<TDate>;
-  /**
-   * Overrideable component slots.
-   * @default {}
-   */
-  slots?: UncapitalizeObjectKeys<DesktopNextTimePickerSlotsComponent<TDate>>;
-  /**
-   * The props used for each component slot.
-   * @default {}
-   */
-  slotsProps?: DesktopNextTimePickerSlotsComponentsProps<TDate>;
-}
+    BaseNextNonStaticPickerExternalProps,
+    SlotsAndSlotsProps<
+      DesktopNextTimePickerSlotsComponent<TDate>,
+      DesktopNextTimePickerSlotsComponentsProps<TDate>
+    > {}
diff --git a/packages/x-date-pickers/src/internals/utils/slots-migration.ts b/packages/x-date-pickers/src/internals/utils/slots-migration.ts
index 58b0c4c22..ccb8f8fc6 100644
--- a/packages/x-date-pickers/src/internals/utils/slots-migration.ts
+++ b/packages/x-date-pickers/src/internals/utils/slots-migration.ts
@@ -19,17 +19,19 @@ export type UncapitalizeObjectKeys<T extends object> = {
     : never;
 };
 
+export type WithoutSlots<T> = Omit<T, 'components' | 'componentsProps' | 'slots' | 'slotsProps'>;
+
 export interface SlotsAndSlotsProps<TSlots extends object, TSlotsProps> {
   /**
    * Overrideable components.
    * @default {}
-   * @deprecated Please use `slots` with uncapitalized properties.
+   * @deprecated Please use `slots`.
    */
   components?: TSlots;
   /**
    * The props used for each component slot.
    * @default {}
-   * @deprecated Please use `slotsProps`
+   * @deprecated Please use `slotsProps`.
    */
   componentsProps?: TSlotsProps;
   /**

```

LukasTy (commented):
Looking great!
Leaving a few more comments and a diff suggestion to simplify types in A LOT of places 🤔

packages/x-date-pickers/src/internals/components/PickersPopper.tsx
```diff
@@ -350,14 +357,14 @@ export function PickersPopper(inProps: PickerPopperProps) {
     ownerState: { wrapperVariant: 'desktop' },
   });
 
-  const PaperContent = components?.PaperContent ?? React.Fragment;
-  const Transition = components?.DesktopTransition ?? Grow;
-  const TrapFocus = components?.DesktopTrapFocus ?? MuiTrapFocus;
+  const PaperContent = slots?.paperContent ?? React.Fragment;
```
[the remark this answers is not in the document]
alexfauquette: When this will be dropped, typescript should help use to remove every thing :)
```ts
// TODO v6: Drop with the legacy pickers
export interface PickersSlotsComponent {
  /**
   * Custom component for the action bar, it is placed bellow the picker views.
   * @default PickersActionBar
   */
  ActionBar?: React.ElementType<PickersActionBarProps>;
  /**
   * Custom component wrapping the views of the desktop and static pickers (it is the direct child of the Paper component).
   * @default React.Fragment
   */
  PaperContent?: React.ElementType<{ children: React.ReactNode }>;
}
```

alexfauquette:
@LukasTy It's not exacly the same, because with your proposal, you completely override the slots'type.

Currently we do not override it, we extend it. For example, if you remove `BaseNextDatePickerSlotsComponent` from `DesktopNextDatePickerSlotsComponent` as follow, you will get an error saying 

```
Interface 'DesktopNextDatePickerProps<TDate>' incorrectly extends interface 'BaseNextDatePickerProps<TDate>'.
  Types of property 'components' are incompatible.
    Type 'DesktopNextDatePickerSlotsComponent<TDate> | undefined' is not assignable to type 'BaseNextDatePickerSlotsComponent<TDate> | undefined'.
      Type 'DesktopNextDatePickerSlotsComponent<TDate>' has no properties in common with type 'BaseNextDatePickerSlotsComponent<TDate>'.ts(2430)
```

Typescript documentation is not super explicit about that point. Seems we are concerned by the following issue
https://github.com/microsoft/TypeScript/issues/16936

```diff
--- a/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
+++ b/packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.types.ts
@@ -14,12 +14,13 @@ import { DateView } from '../internals/models/views';
 import { UncapitalizeObjectKeys } from '../internals/utils/slots-migration';
 
 export interface DesktopNextDatePickerSlotsComponent<TDate>
-  extends BaseNextDatePickerSlotsComponent<TDate>,
-    MakeOptional<UseDesktopPickerSlotsComponent<TDate, DateView>, 'Field' | 'OpenPickerIcon'> {}
+  extends MakeOptional<
+    UseDesktopPickerSlotsComponent<TDate, DateView>,
+    'Field' | 'OpenPickerIcon'
+  > {}
 
 export interface DesktopNextDatePickerSlotsComponentsProps<TDate>
-  extends BaseNextDatePickerSlotsComponentsProps<TDate>,
-    UseDesktopPickerSlotsComponentsProps<TDate, DateView> {}
+  extends UseDesktopPickerSlotsComponentsProps<TDate, DateView> {}
 
 export interface DesktopNextDatePickerProps<TDate>
   extends BaseNextDatePickerProps<TDate>,

```

alexfauquette:
The difference is error protection against bad copy past or forgotten types.

The diff in my previous comment is the case of someone forgetting to extend `BaseNextDatePickerSlotsComponent` when defining slots.
The most probable scenario would be to mix `BaseNextDatePickerSlotsComponent` with `BaseNextTimePickerSlotsComponent`, or  `BaseNextDateTimePickerSlotsComponent`, ...

Such an error would not be detected by your solution, whereas the current approach detects it. It has been a pain to setup those `components`/`componentsProps` without making mistakes, I care that if we use `Omit<>` to override them, we introduce regression without noticing it.

From what I understand the choice is between:
1. Break `extends` check with `Omit<>` to get less repetitive code
2. Keep `extends` check with a lot of duplicates between `components` and `slots` definitions

I prefer 2. because this duplication does not hurt. I never have to take care about the object defining `components` (and now `slots`) because when I want to modify them, I know it's in two interfaces of the form `XxxxSlotsComponent` and `XxxxSlotsComponentsProps`

LukasTy (approved):
> I prefer 2. because this duplication does not hurt. I never have to take care about the object defining `components` (and now `slots`) because when I want to modify them, I know it's in two interfaces of the form `XxxxSlotsComponent` and `XxxxSlotsComponentsProps`

I probably like to live dangerously and would still choose option 1, but that is also a "optimizer" in me talking.
After all, we will end up with same resulting bundle result, hence, it depends on what gives us more sanity in authoring. 👌 

Approving with one final nitpick. Great job!! 💯

packages/x-date-pickers/src/MobileNextDateTimePicker/MobileNextDateTimePicker.types.ts
```diff
@@ -27,11 +28,23 @@ export interface MobileNextDateTimePickerProps<TDate>
   /**
    * Overrideable components.
    * @default {}
+   * @deprecated Please use `slots`. Please use `slots`.
```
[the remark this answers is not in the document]
flaviendelangle: We do use slots !!!!!

PR #7392, opened 2023-01-05
[fields] Support week day formats

### LINKED ISSUE #6590: [fields] format inconsistency between fully and partially filled fields
I was writing test when I realized why it was so hard. The fields do not display the same thing when the value is filled and when it remains a section to fill.

In the following codesandbox, in the first field, type in the year the following digits: `2 0 2 1 2`. You should end up on `DD/0212`

Now if you set a full date(enter something in the day) and you try again to enter `2 0 2 1 2` you will end up on a date `01/212`

https://codesandbox.io/s/pensive-julien-eyyuqx?file=/demo.tsx

### LINKED ISSUE #7254: [pickers] Support days of the week in date and time fields
### Summary 💡

Developers may need to support formats containing the day of the week. 
The day of the week section, as with other sections, should be editable through text-typing and, ideally, up/down arrow keys, which should lead to navigation through the days in the calendar.

image

### Motivation 🔦

Proper support for some edge cases with date formats.

flaviendelangle:
Closes #7254
Fixes #6590
[Playground](https://codesandbox.io/s/date-and-time-pickers-forked-6r1sxg)

### Telerik

Supports arrow navigation (`ArrowDown` adds a day, `ArrowUp` substracts a day).
But don't support letter or digit editing, try pressing a letter while focusing the week day section in [this example](https://www.telerik.com/kendo-react-ui/components/dateinputs/dateinput/formats/), nothing changes.

### Spectrum

No support ?

### Arrow navigation on valid dates

- [x] `ArrowUp` => add a day
- [x] `ArrowDown` => subtract a day
- [x] `PageUp` => go to the end of the week
- [x] `PageDown` => go to the start of the week

### Arrow navigation on invalid dates

- [x] `ArrowUp` => add a day
- [x] `ArrowDown` => subtract a day
- [x] `PageUp` => go to the end of the week
- [x] `PageDown` => go to the start of the week

### Letter editing

- [x] Apply query editing

### Digit editing

- [x] Apply query editing

### Can't use `changeSectionValueFormat` for week days

`dayjs` (and maybe other libraries) can't parse a week day without any context (`dayjs('Monday', 'dddd')` returns an `Invalid Date`).

This method was used in 3 places

1. To check if a digit format has trailing zeroes => I can check on the start of the week. We could do the same for the rest of the section if we refactor `adjustDateSectionValue` to put its `startOf` / `endOf` methods in the object returned by `getDateSectionGetterAndSetter`
2. To use letter editing on a digit section => I am generating the whole week using the digit format and using `indexOf`
3. To use numeric editing on a letter section => I am generating the whole week using the letter format and return the `nth-1` item (press 2 returns the element of index1 ie the 2nd day of the week)

## Other editing approaches

1. Introduce a notion of *read-only* section, the day of the week would be rendered (right now it crashes) but would be handled like a dynamic separator.
4. Only support Arrow editing like Telerik.

### TODO: Add demos and tests


packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -617,15 +734,19 @@ export const getDateSectionGetterAndSetter = <TDate>(
 
   return adapterMethods[dateSectionName];
 };
-export const cleanTrailingZeroInNumericSectionValue = (value: string, maximum: number) => {
```
flaviendelangle, line 620: Date-fns has a format with a trailing zero on numeral week days (`01-07`).
But `7.toString().length === 1`, so our current method does not work (my assumption was that the maximum boundary had no trailing zero).

The new logic is simply to format the current date with the format and check the length.
The assumption is now that all values have the same length on a section with trailing zeroes (no `002` but `0204` for example).
It's a looser assumption, I hope this one will hold.

packages/x-date-pickers/src/locales/enUS.ts
```diff
@@ -70,6 +70,7 @@ const enUSPickers: PickersLocaleText<any> = {
   fieldYearPlaceholder: (params) => 'Y'.repeat(params.digitAmount),
   fieldMonthPlaceholder: (params) => (params.contentType === 'letter' ? 'MMMM' : 'MM'),
   fieldDayPlaceholder: () => 'DD',
+  fieldWeekDayPlaceholder: (params) => (params.contentType === 'letter' ? 'DDDD' : 'DD'),
```
flaviendelangle, line 73: @gerdadesign what placeholder should we put for the week day ? 
It can be either in a letter form (`Friday`, `Fri`) or in a digit for (`05`, `5`).
It's very similar to the month so I took the same approach.
But the `D` is already used and I don't have an alternative idea.

gerdadesign: @flaviendelangle  Can you elaborate with an example of this in context? 

What's the difference between Day and WeekDay, especially if it can be in letter or digit form?

gerdadesign: Ahh, ok. thank you for the clarification. 

I'm seeing a couple sources that use `EEEE` (even though that's not super intuitive). 
[Unicode](http://www.unicode.org/reports/tr35/tr35-31/tr35-dates.html#Date_Format_Patterns),  [Oracle](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) 

There's also the convention of distinguishing between capital and lowercase: using `D` or `DD` for the day of the month (**25th** of December) and using `ddd` for the shortened day of the week (Tue, Wed, Thu) and `dddd` for the long day of the week (Tuesday, Wednesday, Thursday). `d` for representing this in numeric format. 
[Example 1](https://day.js.org/docs/en/display/format), [This one](https://www.php.net/manual/en/datetimeimmutable.createfromformat.php) flips the capitalization convention. 

Since we already distinguish different fields based on capitalization (`M` for month vs `m` for minute), perhaps we can continue to this pattern with `DD` vs `dddd`? 
Though ultimately, it seems the best practice is to use today's date as an example placeholder. Would users be able to do that?

flaviendelangle: > Though ultimately, it seems the best practice is to use today's date as an example placeholder. Would users be able to do that?

Do something like `Friday dd of MMMM` to take the example above ?
Yes I don't see anything blocking.

gerdadesign: I wouldn't mix methods, but it would be nice to allow users to choose how they prefer placeholder values. 

So if I need to enter **Friday, December 16, 2022**:

Placeholder with abbreviations:  `dddd, MMMM DD, YYYY` 
OR 
Placeholder with written descriptions: `Weekday, Month Day, Year`
OR 
Placeholder with today's date as an example of format:  `Tuesday, January 10, 2023`

Generally, it seems the last option is best practice, but someone may choose a different placeholder option in different contexts.

flaviendelangle: Okay
Each one of them is fairly easy to achieve

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -357,15 +408,37 @@ export const changeSectionValueFormat = <TDate>(
...
+  if (contentType !== 'digit') {
+    return false;
+  }
+
+  if (dateSectionName === 'weekDay') {
+    return utils.formatByString(utils.startOfWeek(utils.date()!), format).length > 1;
+  }
+
+  // We can use `changeSectionValueFormat`, because  `utils.parse('1', 'YYYY')` returns `1971` instead of `1`.
+  if (dateSectionName === 'year') {
```
flaviendelangle, line 436: Fixes #6590

I can extract it to a follow up PR but the fix seems simple enough to apply it here

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -508,9 +597,9 @@ export const createDateStrForInputFromSections = (sections: FieldSection[]) => {
   return `\u2066${formattedArray.join('')}\u2069`;
 };
 
-export const getSectionsBoundaries = <TDate, TSection extends FieldSection>(
+export const getSectionsBoundaries = <TDate>(
```
flaviendelangle, line 600: I could not pass the full section anymore because of the digit editing on letter section (and the other way around).
So I'm just passing the `format` and `contentType`.

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -180,6 +205,26 @@ export const adjustInvalidDateSectionValue = <TDate, TSection extends FieldSecti
...
+        } else {
+          newDate = utils.endOfWeek(today);
+        }
+
+        return utils.formatByString(newDate, section.formatValue);
+      }
+
+      const formattedDaysInWeek = getDaysInWeekStr(utils, section.formatValue);
+      const currentDayInWeek = formattedDaysInWeek.indexOf(section.value);
+      const newDayInWeek =
```
LukasTy, line 222: There might be some check missing here, or is this an expected behavior? 🤔 

Or it could also be some check missing in `applyWeekDayChange` function.

https://user-images.githubusercontent.com/4941090/215624285-cc70a227-6f2d-4b50-8617-aee83fbbc1e5.mov

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -357,15 +408,37 @@ export const changeSectionValueFormat = <TDate>(
...
+) => {
+  if (contentType !== 'digit') {
+    return false;
+  }
+
+  if (dateSectionName === 'weekDay') {
+    return utils.formatByString(utils.startOfWeek(utils.date()!), format).length > 1;
+  }
+
+  // We can use `changeSectionValueFormat`, because  `utils.parse('1', 'YYYY')` returns `1971` instead of `1`.
```
LukasTy, line 435:
```suggestion
  // We can't use `changeSectionValueFormat`, because  `utils.parse('1', 'YYYY')` returns `1971` instead of `1`.
```

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -592,11 +695,38 @@ export const applyMeridiemChange = <TDate>(
   return date;
 };
 
+/**
+ * @template TDate
+ * @param {MuiPickersAdapter<TDate>} utils The utils to manipulate the date.*
+ * @param {TDate} date The date on which the week day must be applied.
+ * @param {string} sectionFormat The format of the week day section.
+ * @param {string} sectionValue The new value of the week day section.
+ * @returns {TDate} The date with the new meridiem.
```
LukasTy, line 704: `meridiem` probably does not belong here... 🙈

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
LukasTy: It does look and sound that this solution should be more robust 👍

packages/x-date-pickers/src/internals/hooks/useField/useFieldCharacterEditing.ts
```diff
@@ -149,71 +151,112 @@ export const useFieldCharacterEditing = <TDate, TSection extends FieldSection>({
...
   };
 
-  const applyLetterEditing: CharacterEditingApplier<TDate, TSection> = (params) => {
+  const applyLetterEditing: CharacterEditingApplier<TDate> = (params) => {
+    const findMatchingOptions = (
+      format: string,
+      options: string[],
+      queryValue: string,
+    ): ReturnType<QueryApplier<TSection>> => {
+      const matchingValues = options.filter((month) => month.toLowerCase().startsWith(queryValue));
```
LukasTy, line 160: Or am I missing something and `month` makes sense here? 🤔 
```suggestion
      const matchingValues = options.filter((option) => option.toLowerCase().startsWith(queryValue));
```

LukasTy (commented):
Great effort on this one! 🚀 
Looking and working great! 👍 
Leaving a few nitpick and improvement/question comments. 😉

packages/x-date-pickers/src/AdapterDateFns/index.ts
```diff
@@ -16,6 +16,27 @@ const formatTokenMap: MuiFormatTokenMap = {
   MMM: { sectionName: 'month', contentType: 'letter' },
   LLL: { sectionName: 'month', contentType: 'letter' },
   LLLL: { sectionName: 'month', contentType: 'letter' },
+  E: { sectionName: 'weekDay', contentType: 'letter' },
+  EE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEEE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEEEE: { sectionName: 'weekDay', contentType: 'letter' },
+  i: 'weekDay',
+  ii: 'weekDay',
```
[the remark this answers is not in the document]
flaviendelangle: We are not supporting digit formats with letters at the end correctly for now so I excluded them from the formats.
It's probably the most important improvement we can do in the future (for week days but more importantly for month day).

packages/x-date-pickers/src/AdapterMoment/index.ts
```diff
@@ -12,11 +12,16 @@ const formatTokenMap: MuiFormatTokenMap = {
...
   MMMM: { sectionName: 'month', contentType: 'letter' },
 
-  // Day of Month
+  // Day of the month
   D: 'day',
   Do: 'day',
   DD: 'day',
 
+  // Day of the week
+  e: 'weekDay',
```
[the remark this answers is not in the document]
flaviendelangle: Added (except `do` for the reason explained on the `date-fns` comment) :+1:

packages/x-date-pickers/src/AdapterDateFns/index.ts
```diff
@@ -16,6 +16,27 @@ const formatTokenMap: MuiFormatTokenMap = {
...
+  e: 'weekDay',
+  ee: 'weekDay',
+  eee: { sectionName: 'weekDay', contentType: 'letter' },
+  eeee: { sectionName: 'weekDay', contentType: 'letter' },
+  eeeee: { sectionName: 'weekDay', contentType: 'letter' },
+  eeeeee: { sectionName: 'weekDay', contentType: 'letter' },
+  iii: { sectionName: 'weekDay', contentType: 'letter' },
+  iiii: { sectionName: 'weekDay', contentType: 'letter' },
```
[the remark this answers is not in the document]
flaviendelangle: Not really, fixed
I'll clean the format token make of every adapter in a follow up, the order is not always the same :+1:

packages/x-date-pickers/src/AdapterDateFns/index.ts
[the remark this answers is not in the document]
LukasTy: Yeah, I've tested it and indeed, it was not working correctly with keyboard entry regardless of `contentType`.
It was fine only with keyboard navigation. 🤔

packages/x-date-pickers/src/AdapterDateFns/index.ts
```diff
@@ -16,6 +16,27 @@ const formatTokenMap: MuiFormatTokenMap = {
   MMM: { sectionName: 'month', contentType: 'letter' },
   LLL: { sectionName: 'month', contentType: 'letter' },
   LLLL: { sectionName: 'month', contentType: 'letter' },
+  E: { sectionName: 'weekDay', contentType: 'letter' },
+  EE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEEE: { sectionName: 'weekDay', contentType: 'letter' },
+  EEEEE: { sectionName: 'weekDay', contentType: 'letter' },
```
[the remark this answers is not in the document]
flaviendelangle: Do you mean that we can't access `Thursday` with the keyboard when using the `E` format ?
If so, I think it's something that we can expect, it remains accessible with the digits and the arrow.
Tbh, some of these formats will probably never be used for editing :grimacing:

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
```diff
@@ -180,6 +205,26 @@ export const adjustInvalidDateSectionValue = <TDate, TSection extends FieldSecti
       return utils.formatByString(newDate, section.formatValue);
     }
 
+    case 'weekDay': {
+      let newDate: TDate;
+      if (shouldSetAbsolute) {
+        if (delta > 0 || isEnd) {
+          newDate = utils.startOfWeek(today);
```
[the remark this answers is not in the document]
flaviendelangle: Indeed, it seems incoherent when pressing `Start`.

Do you agree that pressing `ArrowDown` on an empty day section should set `31` or should it set `1` ?
Spectrum sets `1`
Telerik sets `31`

In any case, I think we can fix it in another PR since the behavior is not linked only to the week days.

packages/x-date-pickers/src/AdapterDateFns/index.ts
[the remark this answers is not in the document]
LukasTy: Yeah, I meant exactly that. 👌 
Sure, I'd assume the same, that they are going to be used very rarely and shouldn't be a big problem.
But in a very rare case we could probably come up with a fix supporting keyboard entry autocompletion for longer length than the section, in case there are more than 1 found items.
Because we'd have the same issue with other "exotic" format tokens, like [`MMMMM`](https://date-fns.org/v2.29.3/docs/format#:~:text=2-,MMMMM,-J%2C%20F%2C%20...%2C%20D). 🙈

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
LukasTy: Indeed, this fix can be done in a separate PR. 👍 

To me the following would seem most logical:
- <kbd>ArrowDown</kbd> -> `1`
- <kbd>ArrowUp</kbd> -> `31`

Reason behind it: "Up" sounds like it should increase the value, sort of makes sense that it would bump the value to max in case of empty section.

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
flaviendelangle: Nice catch !
I had a bug in the digit editing, when using a letter week day section :+1:

packages/x-date-pickers/src/AdapterDateFns/index.ts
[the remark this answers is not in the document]
flaviendelangle: How do you suggest we could support it ?

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
flaviendelangle: The browser input sets `31` on arrow up.

But to take a kitchen example, when I press `-` on my electric hob, it sets the power to max whereas on my parent's one it sets to a low power.
Both have very good arguments.

Setting `1` is closer from the logic of the key pressed (arrow *down* so we sets to the value down in the list).
Setting `31` is more optimized to get to values close to the end of the month (to get to `29` I just have to do `ArrowDown` `ArrowDown` `ArrowDown` instead of `ArrowUp` `ArrowDown` `ArrowDown`, my find does not move. I personally find this behavior easier to use.

packages/x-date-pickers/src/AdapterDateFns/index.ts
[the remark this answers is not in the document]
LukasTy: I'm honestly not suggesting to go ahead and rush introducing this extra complexity until there is actual need for it.
If we'd proceed with handling every rare condition, our fields codebase and it's complexity might explode. 🙈 

But if we'd need support for it. I believe it shouldn't be too difficult allowing letter section auto-complete regardless of the section format length. If we still have > 1 possible autocomplete values based on the input—allow section autocomplete to work with more than 1 character.
For example:
If we mentioned one letter week day format:
- Entering `t` shows `T` in the section, but does not jump to the next section, because there are > 1 options
- Entering `h` still shows only `T` in the section, but does jump to the next section, because there is only one weekday option for `th` search string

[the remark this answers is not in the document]
flaviendelangle: OK
The problem here is that we are doing the auto-completion on the value formatted with the current format. We don't have the `h` of `Thursday`.
One workaround could be that if we end up with several identical values matching the query, we fallback to a full week day format.
Another would be to always use the full week day format, but that's a change that would need to be carefully checked (could we have some format in some locale that are not just shortcuts of the full format and that would be hard to use if we autocomplete based on the full format and render those weird formats ?)

Maybe we can wait for feedback on this one :+1:

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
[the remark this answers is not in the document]
LukasTy: > The browser input sets `31` on arrow up.

Well. Indeed, my assumption doesn't hold ground if the three major browsers do it the same we currently have as well. 🤔 
If major browsers did implement it this way, it's probably written in spec and we could safely go with the same implementation. 👌 
But at least we (I and browser native implementations) agree on the `Home` and `End` behaviour, where it's the same regardless of empty state or filled data.

packages/x-date-pickers/src/AdapterDateFns/index.ts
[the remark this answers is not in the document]
LukasTy: > The problem here is that we are doing the auto-completion on the value formatted with the current format. We don't have the h of Thursday.

Oh, fair point. 😞 

> One workaround could be that if we end up with several identical values matching the query, we fallback to a full week day format.

> Maybe we can wait for feedback on this one 

Indeed, based on this discussion findings it seems like the added complexity is definitely not worth it until there is enough incentive from users. 🙈

packages/x-date-pickers/src/internals/hooks/useField/useField.utils.ts
LukasTy: The bug is fixed! 🙌 
I can no longer spot or anticipate any further issues. 💯

LukasTy (approved):
Great job! 👍 
P.S. A possible follow up could be the `week of year` support. 😃

PR #7408, opened 2023-01-06
[pickers] support components and slots for new pickers

alexfauquette:
This is a draft branch where instead of using a generic to uncapitalise type, I redefine types for `slots`


alexfauquette:
Replaced by the #7390 fix

Issue #7440, opened 2023-01-08
[pickers] UX design review v6.0.0-alpha.14


oliviertassinari:
I found https://twitter.com/simonw/status/1611856336396128256 by chance, which inspired me to have a deep dive into the UX with the latest version of the pickers. The goal is to surface all the frustration points I had as an end-user. This issue can be considered a design review, with confidence levels:

## 1. Vertical height sync 90%

Open https://next.mui.com/x/react-date-pickers/date-picker/ on Android, and go to the year view, the active year isn't in the viewport:

<img width="406" alt="Screenshot 2023-01-08 at 14 17 33" src="https://user-images.githubusercontent.com/3165635/211198131-f13b5267-6f2b-4acc-960c-aeb8432054de.png">

## 2. Month edit with date field doesn't work 90%

Open https://next.mui.com/x/react-date-pickers/date-field/#customize-the-date-format try to use the [Microsoft SwiftKey keyboard](https://play.google.com/store/apps/details?id=com.touchtype.swiftkey&hl=en&gl=US&pli=1). It doesn't work in any way. The extension has 1B+ downloads (to be compared with the most used app on Android, YouTube 10B downloads). I suspect it's not specific to this keyboard extension. I have seen the same issue with Grammarly's keyboard.

<img width="389" alt="Screenshot 2023-01-08 at 14 22 13" src="https://user-images.githubusercontent.com/3165635/211198351-0eeb24cc-aac7-4bfe-b857-1d941f71eab9.png">

## 3. Popup alignment 90%

When I open https://next.mui.com/x/react-date-pickers/date-picker/#basic-usage, the popup feels misaligned:

<img width="669" alt="Screenshot 2023-01-08 at 13 57 16" src="https://user-images.githubusercontent.com/3165635/211197231-fa1b3997-5187-4a2a-a46f-78bebbba4861.png">

I would expect:

<img width="738" alt="Screenshot 2023-01-08 at 13 59 25" src="https://user-images.githubusercontent.com/3165635/211197303-7069e50b-1bb2-420a-a522-8765bd928bbf.png">

See https://m3.material.io/components/date-pickers/specs and all the others date pickers built, e.g. native ones or massively popular ones like https://jqueryui.com/datepicker/.

## 4. View reset  90%

When I switch to the year view, close the popup, and open it again. I would expect to see the month view, not:

[...]

PR #7444, opened 2023-01-09
[fields] Fix Android editing

flaviendelangle:
Part of #7440

In Android, we have 2 `onChange`, for the 1st one we need to just store the result as is in the input.
Since the RTL PR, we were storing the clean value (without the RTL characters), which was causing shift in the edited layout on the 2nd `onChange`.

### DIFF
--- a/packages/x-date-pickers/src/DateField/tests/editing.DateField.test.tsx
+++ b/packages/x-date-pickers/src/DateField/tests/editing.DateField.test.tsx
@@ -773,4 +773,78 @@ describe('<DateField /> - Editing', () => {
       expect(onChange.lastCall.firstArg).toEqualDateTime(new Date(2022, 8, 16, 3, 3, 3));
     });
   });
+
+  describe('Android editing', () => {
+    let originalUserAgent: string = '';
+
+    beforeEach(() => {
+      originalUserAgent = global.navigator.userAgent;
+      Object.defineProperty(global.navigator, 'userAgent', {
+        configurable: true,
+        writable: true,
+        value:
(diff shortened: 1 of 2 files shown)


alexfauquette (approved):
It was hard to reproduce. I don't have the bug with Gboard. 

What about adding a test to catch this error?

flaviendelangle:
Do you have an idea of how to test those Android-only behaviors ?

flaviendelangle:
I added two tests :+1:

oliviertassinari:
@flaviendelangle Thanks for looking into it 👌. 

I noticed two other things:

1. https://next.mui.com/x/react-date-pickers/date-field/#customize-the-date-format

https://user-images.githubusercontent.com/3165635/216845789-f4837f75-9fc9-4817-ae3b-eac50238ffc8.mov

2. https://next.mui.com/x/react-date-pickers/date-field/#customize-the-date-format

https://user-images.githubusercontent.com/3165635/216845986-893b299e-69bb-4966-bfb6-4f0f9222e924.mov

It can be reproduced on https://live.browserstack.com/dashboard#os=android&os_version=12.0&device=Samsung+Galaxy+S22&device_browser=chrome&zoom_to_fit=true&full_screen=true&url=https%3A%2F%2Fy2jlm1.csb.app%2F&speed=1&host_ports=google.com%2C80%2C0.

flaviendelangle:
Bug reproduced, that's not specific to Android, I'll investigate

flaviendelangle:
2nd bug reproduced.
This one is interesting. The `Backspace` logic is not handled by the `onChange` to it does not work on Android...
I created #7835 because it's not a trivial change.

PR #7456, opened 2023-01-09
[DataGrid] Stop exporting editing selector

m4theushw:
Part of #3287 

### Breaking changes

- The `gridEditRowsStateSelector` selector was removed.


packages/grid/x-data-grid/src/internals/index.ts
```diff
@@ -49,6 +49,7 @@ export {
   preferencePanelStateInitializer,
 } from '../hooks/features/preferencesPanel/useGridPreferencesPanel';
 export { useGridEditing, editingStateInitializer } from '../hooks/features/editing/useGridEditing';
+export * from '../hooks/features/editing/gridEditingSelectors';
```
flaviendelangle, line 52: I tend to avoid the "export all" pattern in the `internal` to make keep clear what we are exporting as private and prioritize public imports.

flaviendelangle (approved):

PR #7482, opened 2023-01-12
[DataGridPremium] Keep focus on first selected cell

m4theushw:
Fixes https://github.com/mui/mui-x/pull/6567#issuecomment-1322791819

Preview: https://deploy-preview-7482--material-ui-x.netlify.app/x/react-data-grid/cell-selection/

Before:

msedge_UpAU33Pp7t

After:

msedge_CNGMRJx4ga

To fix https://github.com/mui/mui-x/pull/6567#issuecomment-1322791819 initially I thought I would have to move the focus to the last clicked cell but keep the outline in the first clicked cell. To implement this we would need to store the outlined cell in the state. The need to keep the last clicked cell focused was to allow to create a range selection with arrow keys. However, I found a different way to implement this without splitting the outline/focus state. When another cell is clicked while <kbd>Shift</kbd> is pressed, I don't update the focus state, keeping the outline in the first cell. This works fine with selection via click, but to support keyboard selection is a bit trickery. When a navigation key is pressed while holding <kbd>Shift</kbd>, I disable the entire keyboard navigation logic and start tracking the focus inside the cell selection hook. I call this "virtual focus" since the user is not seeing which cell he's navigating to. By doing this, the outline is kept in the first cell while selecting via arrow keys is still possible. This method has a disadvantage which is that, once a range selection is created, pressing an arrow key **without holding <kbd>Shift</kbd>** will erase the current selection. This is because the virtual focus in on the last cell but the user is seeing the first cell as focused. Fortunately, this is how most spreadsheet apps work.

### DIFF
--- a/packages/grid/x-data-grid-premium/src/hooks/features/cellSelection/useGridCellSelection.ts
+++ b/packages/grid/x-data-grid-premium/src/hooks/features/cellSelection/useGridCellSelection.ts
@@ -198,9 +198,8 @@ export const useGridCellSelection = (
         return;
       }
 
-      const selectedCells = Object.keys(apiRef.current.unstable_getCellSelectionModel());
-      const isSelectingFirstCell = selectedCells.length === 0;
-      if (hasClickedValidCellForRangeSelection(params) && event.shiftKey && !isSelectingFirstCell) {
+      const focusedCell = gridFocusCellSelector(apiRef);
+      if (hasClickedValidCellForRangeSelection(params) && event.shiftKey && focusedCell) {
         event.preventDefault();
       }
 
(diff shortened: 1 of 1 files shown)


MBilalShafi:
I observed this behavior when playing around with this, is it expected?
focus

1. When I click on first cell and then Shift + Click on another cell, outline/focus keeps on first cell ✅ 
2. When I use arrow keys to update first cell focus and then Shift + Click on another cell, focus moves to the new cell ❓

joserodolfofreitas (approved):
Nice improvement! 👌

MBilalShafi (approved):
LGTM

Issue #7086, opened 2022-12-04
[DataGrid] Add codemods for v6 breaking changes


MBilalShafi:
Codemods for pickers can be found [here](https://github.com/mui/mui-x/issues/6997)

The v6 for `DataGrid` is bringing quite a few breaking changes.
To ease users' migration we ought to try and cover as most changes with codemods as possible.

The codemod package can be [found here](https://github.com/mui/mui-x/tree/next/packages/x-codemod). It contains all the core setup needed in order to add new codemods to it.

This is intended as an umbrella issue to group all the codemods needed for `v.6.x` migration for DataGrid.
Ideally as much migrations as possible should be provided with `preset-safe` codemod.

## The process of adding a new codemod
- One of the [existing codemods](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0) can be used as an example (i.e.: [localization-provider-rename-locale](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/localization-provider-rename-locale))
- copy or create a new folder for your codemod
- all the relevant code that is eventually shipped to the user is in `index.ts`
- update/provide tests for your newly created codemod (test fixtures are authored as js files to allow as minimal syntax as you want—without triggering ts issues)
- if your codemod is safe to add to [`preset-safe`](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/preset-safe)—update this codemod with your newly added codemod making sure the [codemods order is correct](https://github.com/mui/mui-x/blob/next/packages/x-codemod/src/v6.0.0/preset-safe/index.ts) if codemod is sensitive to that (i.e. separate props should firstly be migrated to `componentProps` and only later be passed through a codemod renaming `componentProps` to `slotProps` if we'd go with such approach)
- update [README](https://github.com/mui/mui-x/blob/next/packages/x-codemod/README.md) file with your newly added codemod
You can check [@mui/codemod](https://github.com/mui/material-ui/tree/master/packages/mui-codemod/src) for inspiration on more advanced migration cases.

## How to test your changes locally (on a different project)

P.S. `yarn link` does not create a `bin` command from a linked package
- run `build` with either `yarn workspace @mui/x-codemod build` or `yarn build` if you are in `src/packages/x-codemod`
- run `cd build && yarn pack`
- copy the built `7.tgz` file to the root of your project you want to run (test) the codemod in
[...]

PR #6985, opened 2022-11-24
[DataGrid] Refactor `GridFilterItem` props

MBilalShafi:
Closes #5604 

- [x] Rename GridFilterItem['columnField'] -> GridFilterItem['field']
- [x] Rename `GridFilterItem['operatorValue']` -> `GridFilterItem['operator']`  
- [x] Make  `GridFilterItem['operatorValue']` (renamed 'operator') required 
- [x] Update migration guide

PR #7483, opened 2023-01-12
[codemod] Rename `GridFilterItem` props

MBilalShafi:
Codemod for breaking change done in #6985
Relates to umbrella issue #7086

### Use cases:

1. ✅ Basic use case when users directly use `filterModel` prop or `initialState.filter.filterModel`
2. ✅ `filterModel` is controlled using some state variable, `initialState` is defined in some other variable
3.  `filterModel` is deduced using some business logic like some external hook etc - NOT APPLICABLE, LIMITATION ON JSCODESHIFT


packages/x-codemod/src/v6.0.0/data-grid/rename-filter-item-props/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  operatorValue: 'operator',
+};
+
+const PACKAGE_REGEXP = /@mui\/x-data-grid(-pro|-premium)?/;
+
+const GridComponents = ['DataGrid', 'DataGridPro', 'DataGridPremium'];
+
+const preRequisiteUsages: { [identifierName: string]: PreRequisiteUsage } = {
+  columnField: {
+    possiblePaths: ['initialState', 'filterModel', 'onFilterModelChange'],
```
LukasTy, line 15: I haven't investigated how do these settings directly affected the codemod, but I've added a somewhat unrelated case, but with the same topic I've raised earlier of having a prop name clash and it was renamed, even though it should not have been and is technically a bug from the users point of view. 🙈 
```tsx
<div columnField="column" filterModel={{ columnField: 'columnFilterField' }} />
```
was transformed to:
```tsx
<div field="column" filterModel={{ field: 'columnFilterField' }} />
```

LukasTy (commented):
I'm re-raising my concern in regards to a potential issue with such codemods, targeting identifiers/literals, instead of direct props on a known component. 🙈

packages/x-codemod/src/v6.0.0/data-grid/rename-filter-item-props/index.ts
MBilalShafi: That is a very solid point. And is a tradeoff between:
1. Not doing any refactors outside of our components (like something defined in a variable even in the same file won't be refactored, most common examples include objects initialized in React.useState, React.useRef, props passed from a container component and so on)
2. Allowing renaming of literals (to support use-cases outside of components) after doing some preliminary checks which reduce the possibilities of false outcomes.
  - Like in the above mentioned use-case, we are checking for these conditions before applying a literal rename to reduce the chances of false renames:
    1. There must be some components imported from `x-data-grid(pro|premium)` package.
    2. One of these components `['DataGrid', 'DataGridPro', 'DataGridPremium']` must be imported from one of the packages checked in point 1.
    3. One of these props `['initialState', 'filterModel', 'onFilterModelChange']` must be passed to one of the above components (the idea is to only rename related literals if related props are being used in the component). For example: if filterModel or initialState.filter.filterModel is used, it's more likely user is using columnField and it's safer to rename.
    
I went for 2 because IMO it will improve the impact of codemod changes, and is less likely to do unsafe changes. Even if it does, codemods is something user must not rely totally on. They should check after running codemods if it left some console errors or something which require user intervention.

This discussion also pops this question in my head, are users likely to use same names as ours in the same file? Even if a very low percentage of users are likely to do that, should it be a huge flag as more than 90% of users (IMO) are not likely to do that and they also have diffs in editors before committing to revert false changes if they happen? 🤔 

Let me know what thoughts on this.

LukasTy: > Let me know what thoughts on this.

I do mostly agree with your points and am hopeful that your changes will have a much bigger positive impact rather than a few cases of false positives. 👍 
Just wanted to hear your position and clarify if you did know about the possibly unwanted behaviour of the transformations your codemod would do. 👌 

For the most part, our prop names are not that common that you'd expect to find them on other components.
And I also agree, that cases of such an un-scoped usages, where we'd have both `DataGrid` as well as some other custom component with same names in the same file are likely rare or close to none. 🤞 🙈

LukasTy (approved):

PR #7484, opened 2023-01-12
[pickers] Go to the default view when opening a picker

flaviendelangle:
Part of #7440

We are moving back to v5 behavior on this


packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.test.tsx
```diff
@@ -92,22 +92,71 @@ describe('<DesktopNextDatePicker />', () => {
...
+      openPicker({ type: 'date', variant: 'desktop' });
+
+      fireEvent.click(screen.getByLabelText(/switch to year view/i));
+      expect(handleViewChange.callCount).to.equal(1);
+
+      // Dismiss the picker
+      userEvent.keyPress(document.activeElement!, { key: 'Escape' });
+
+      openPicker({ type: 'date', variant: 'desktop' });
+      expect(handleViewChange.callCount).to.equal(2);
```
alexfauquette, line 132: I do not have a full understanding of how the current view is managed. But this test raised a question:

Would it makes sense to switch back to the default view when we close the view. Such that when reopening, the current view will not be set and so we avoid getting two renderings:
- one for opening
- a second one for updating the `view` state

I'm wondering if removing the early return `if(!open)` in hook you modified would do the job

flaviendelangle: > Would it makes sense to switch back to the default view when we close the view.

That's an interesting question indeed
The view would probably not be reset if the user is controlling the opening state and closing it from the outside.
Not sure if this is problematic.

alexfauquette (approved):

PR #7498, opened 2023-01-12
[pickers] Replace `showToolbar` with toolbar slot `hidden` prop

### LINKED ISSUE #7472: [pickers] Replace `showToolbar` prop with `slotsProps.toolbar`
Created from: https://github.com/mui/mui-x/pull/7390#discussion_r1066653758

We have already replaced top level `hideTabs` prop with `slotsProps.tabs.hidden`.
The next logical thing would be to migrate top level `showToolbar` prop to `slotsProps.toolbar.hidden` (or something similar).

LukasTy:
Fixes #7472 

### Breaking changes

The `showToolbar` prop has been moved to the `toolbar` component slot props:

  ```diff
   <DatePicker
  -  showToolbar
  +  slotProps={{
  +    toolbar: {
  +      hidden: false,
  +    }
  +  }}
   />
  ```


docs/data/date-pickers/custom-layout/custom-layout.md
```diff
@@ -14,7 +14,7 @@ packageName: '@mui/x-date-pickers'
 
 By default, pickers are made of 4 sub elements present in the following order:
 
-- The **toolbar** displaying the selected date. Can be enforced with `showToolbar` prop.
+- The **toolbar** displaying the selected date. Can be enforced with `componentsProps: { toolbar: { hidden: false } }` prop.
 - The **content** displaying the current view. Can be a calendar, or a clock.
 - The **tabs** allowing to switch between day and time views in Date Time Pickers.
```
flaviendelangle, line 19: Should we also add the `Can be enforced` here ?

packages/x-date-pickers/src/internals/models/props/toolbar.ts
```diff
@@ -3,6 +3,11 @@ import { DateOrTimeView } from '../views';
 
 export interface BaseToolbarProps<TValue, TView extends DateOrTimeView>
   extends ExportedBaseToolbarProps {
+  /**
+   * If `true`, show the toolbar even in desktop mode.
+   * @default `true` for Desktop, `false` for Mobile (based on the chosen wrapper and `desktopModeMediaQuery` prop), `displayStaticWrapperAs === 'desktop'` for `Static` pickers
```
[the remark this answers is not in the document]
flaviendelangle: We have a lot of `Desktop` / `Mobile` and a lot of `desktop` / `mobile` in the boolean `@default` descriptions.
We should unify at some point.

You prefer the capitalized version ?

Imho we can drop the `Static` picker part and take for granted that people using the static pickers understand that it's mobile by default and can be overridden with `displayStaticWrapperAs`

```suggestion
   * @default `true` for Desktop, `false` for Mobile (based on the `desktopModeMediaQuery` prop for responsive pickers).
```

Or even

```suggestion
   * @default `true` for Desktop, `false` for Mobile.
```

packages/x-date-pickers/src/internals/models/props/toolbar.ts
```diff
@@ -3,6 +3,11 @@ import { DateOrTimeView } from '../views';
 
 export interface BaseToolbarProps<TValue, TView extends DateOrTimeView>
   extends ExportedBaseToolbarProps {
+  /**
+   * If `true`, show the toolbar even in desktop mode.
```
flaviendelangle, line 7: Not sure which one is valid, we have both in our descriptions

```suggestion
   * If `true`, shows the toolbar even in desktop mode.
```

packages/x-date-pickers/src/DesktopNextDatePicker/DesktopNextDatePicker.test.tsx
```diff
@@ -55,11 +55,11 @@ describe('<DesktopNextDatePicker />', () => {
     expect(handleChange.callCount).to.equal(1);
   });
 
-  it('prop `showToolbar` – renders toolbar in desktop mode', () => {
```
flaviendelangle, line 58: For the test files of the new pickers, can we follow the nomenclature of the other slots

```tsx
describe('Component slots: Popper', () => {
```

Can be adapter if needed, I just want to keep the consistency :+1:

packages/x-date-pickers/src/internals/models/props/toolbar.ts
```diff
@@ -25,7 +30,7 @@ export interface BaseToolbarProps<TValue, TView extends DateOrTimeView>
   toggleMobileKeyboardView?: () => void;
 }
 
-export interface ExportedBaseToolbarProps {
+export interface ExportedBaseToolbarProps extends Pick<BaseToolbarProps<any, any>, 'hidden'> {
```
flaviendelangle, line 33: Why not moving the `hidden` prop here ?
Usually we have `interface XXXProps extends ExportedXXXProps` not the other way around.

packages/x-date-pickers/src/StaticNextDateTimePicker/StaticNextDateTimePicker.tsx
```diff
@@ -41,13 +41,16 @@ const StaticNextDateTimePicker = React.forwardRef(function StaticNextDateTimePic
...
     yearsPerRow: defaultizedProps.yearsPerRow ?? (displayStaticWrapperAs === 'mobile' ? 3 : 4),
-    showToolbar: defaultizedProps.showToolbar ?? displayStaticWrapperAs === 'mobile',
     componentsProps: {
       ...defaultizedProps.componentsProps,
       tabs: {
         hidden: displayStaticWrapperAs === 'desktop',
         ...defaultizedProps.componentsProps?.tabs,
       },
+      toolbar: {
+        hidden: displayStaticWrapperAs !== 'mobile',
```
flaviendelangle, line 51:
```suggestion
        hidden: displayStaticWrapperAs === 'desktop',
```

Maybe keep the same check as for the `tabs` (same on other static pickers)

packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.tsx
```diff
@@ -50,6 +49,10 @@ const MobileNextDateRangePicker = React.forwardRef(function MobileNextDateRangeP
         sx,
         ref,
       }),
+      toolbar: {
+        hidden: false,
```
flaviendelangle, line 53: Do you think it's worth adding the `false` value ? If we just remove those lines, the toolbar receives `undefined` which works as well.

packages/x-date-pickers/src/internals/models/props/toolbar.ts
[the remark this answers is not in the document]
LukasTy: > You prefer the capitalized version ?

Nope, just lazily copy pasted it from `showToolbar` and added the static description to it. 🙈 

> Imho we can drop the Static picker part 

Agreed, we rarely specify it in other cases.

docs/data/date-pickers/custom-layout/custom-layout.md
LukasTy: Good idea. 👍

packages/x-date-pickers/src/internals/models/props/toolbar.ts
LukasTy: Do we? 🤔 I couldn't find any usages...
To me `show` sounds a bit more correct in this context.

packages/x-date-pickers/src/internals/models/props/toolbar.ts
LukasTy: Good point. Will fix this case, I thought that something seems strange props wise for this component... 🙈

packages/x-date-pickers/src/StaticNextDateTimePicker/StaticNextDateTimePicker.tsx
LukasTy: Sure, better stay coherent. 👍

packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.tsx
LukasTy: Yes, it does work with undefined, but I usually like being explicit. 🤔 
We can omit it if you think it's pointless. But on the hand—if we will ever need to change the default rule, then it will be less changes and a bit easier to discover places, where it need to happen. 🤔

packages/x-date-pickers/src/internals/models/props/toolbar.ts
flaviendelangle: > If `true`, renders as disabled.
> If `true`, calls `renderLoading` instead of rendering the day calendar.

> If `true`, disable values after the current date for date components, time for time components and both for date time components.
> If `true`, force firing the `onChange` callback


I guess it depends if we consider it do be a short notation of `the component shows` or if it's an infinite.

The core also uses both

> If `true`, expands the accordion, otherwise collapse it.

> If `true`, disables gutter padding.


> If `true`, display the list in horizontal direction.

> If `true`, hide the selected options from the list box.

Sometime it even have the explicit `it`

> If `true`, it removes the margin between two expanded accordion items and the increase of height.

They also have a few sentences with `will`

> If `true`, will allow focus on disabled items.

____

In summary, nobody agrees :laughing:

packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.tsx
flaviendelangle: True
Let's be explicit, I'll also add it to my `tabs` PR.

packages/x-date-pickers/src/internals/models/props/toolbar.ts
LukasTy: > In summary, nobody agrees 😆

Good summary. 😆 
I think that we could leave this case as is for now.
But maybe @samuelsycamore has an idea of which form would be the most correct? We could unify such cases in the future if we decide on an approach. 🤔

flaviendelangle: Clearly, let's not block this PR for it
I was thinking out loud :laughing:

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
```diff
@@ -102,6 +103,7 @@ const DateRangePickerToolbar = React.forwardRef(function DateRangePickerToolbar<
       className={clsx(className, classes.root)}
       ownerState={ownerState}
       ref={ref}
+      {...other}
```
flaviendelangle, line 106: We probably want to put the `isLandscape={false}` after the spread if we want to keep forcing the orientation (or check if it works correctly in landscape now if we want to change the behavior).

flaviendelangle (approved):
Works great :partying_face:

packages/x-date-pickers-pro/src/DateRangePicker/DateRangePickerToolbar.tsx
LukasTy: Would you say that it is not working correctly? 🤔 
Screenshot 2023-01-16 at 16 46 05

flaviendelangle: The main usage of `isLandscape` is probably the lanscape mode on mobile (without any customization from the user).
And I guess it should be working fine.

And for the manual landscape on desktop, as your screen shows, it seems to work fine.

Just wanted to notice that it was a behavior change.

LukasTy: Thanks for noting. Did a deeper comparison and noticed that it does introduce a visual regression.
Changed the spread to be at the top until the toolbar layout is fixed. 👌

PR #7503, opened 2023-01-12
[pickers] Hide the tabs by default on `DesktopNextDateTimePicker`

flaviendelangle:
I noticed this bug while reviewing #7498


LukasTy (approved):
Damn, good spot! 🥳

packages/x-date-pickers/src/DesktopNextDateTimePicker/DesktopNextDateTimePicker.tsx
```diff
@@ -61,6 +61,10 @@ const DesktopNextDateTimePicker = React.forwardRef(function DesktopNextDateTimeP
         label: defaultizedProps.label,
         ampm: defaultizedProps.ampm,
       }),
+      tabs: {
+        hidden: true,
```
LukasTy, line 65: By the way, Legacy date time picker also has this regression. 🤔

flaviendelangle: I dive a bit deeper:

On v5, we had 2 checks:
- `const showTabs = !hideTabs && typeof window !== 'undefined' && window.innerHeight > 667;` in `CalendarOrClockPicker` 
- A default `hideTabs` props applied on each picker

On v6 legacy pickers, the `CalendarOrClockPicker` check has been moved to `DateTimePickerTabs` (in #6543), and can probably be dropped (not sure what was the goal even in v5 since all the pickers are already defining a value to `hideTabs` and it works poorly with SSR if ever used).

And we indeed miss the check on the pickers, I'll add it back.

packages/x-date-pickers/src/DesktopNextTimePicker/DesktopNextTimePicker.tsx
```diff
@@ -34,8 +34,6 @@ const DesktopNextTimePicker = React.forwardRef(function DesktopNextTimePicker<TD
     ...defaultizedProps.viewRenderers,
   };
 
-  const slots = defaultizedProps.slots;
```
flaviendelangle, line 37: Some components were defining those variables.
Some components were spreading them from `defaultizedProps`
I unified it :+1:

PR #7508, opened 2023-01-13
[codemod] Remove `disableExtendRowFullWidth` prop

MBilalShafi:
Part of #7086 
Codemod for change done in https://github.com/mui/mui-x/pull/7373

- [x] Update README
- [x] Update Migration Guide


docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
```diff
@@ -65,7 +65,7 @@ The minimum supported Node.js version has been changed from 12.0.0 to 14.0.0, si
   The old behavior can be restored by using `apiRef.current.stopRowEditMode({ ignoreModifications: true })` or `apiRef.current.stopCellEditMode({ ignoreModifications: true })`.
 - The `onColumnVisibilityChange` prop was removed. Use `onColumnVisibilityModelChange` instead.
 - The `components.Header` slot was removed. Use `components.Toolbar` slot instead.
-- The `disableExtendRowFullWidth` prop was removed.
+- ✅ The `disableExtendRowFullWidth` prop was removed.
```
alexfauquette, line 68: How does user reach the same behavior without this prop?

MBilalShafi: If the purpose is to show/hide the right border for last column/header cell, `showCellVerticalBorder` and `showColumnVerticalBorder` could be used to achieve that. `disableExtendRowFullWidth` was only adding a border to the right of last cell. We can try to have a proper fix for this if requested by some users.

alexfauquette: Maybe worth adding a small sentence like 

> You can reproduce the behavior by adding a right border to rows' last cell

Otherwise, as a user reading the breaking change, I see the prop has been removed, but I have no clue about the impact it has (especially if the dev how added it is not anymore in the team). Which is not good for the confidence dev have in the migration guide

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
```diff
@@ -65,7 +65,7 @@ The minimum supported Node.js version has been changed from 12.0.0 to 14.0.0, si
   The old behavior can be restored by using `apiRef.current.stopRowEditMode({ ignoreModifications: true })` or `apiRef.current.stopCellEditMode({ ignoreModifications: true })`.
 - The `onColumnVisibilityChange` prop was removed. Use `onColumnVisibilityModelChange` instead.
```
LukasTy, line 66: I see a lot of cases (especially in data grid side) of `was removed` and only some `has been`.
I'm not sure which form is more correct.
- Prop has been removed. Please update usage.
- Prop was removed. Please update usage.

Maybe @samuelsycamore could confirm? 🤔

packages/x-codemod/README.md
```diff
@@ -209,6 +209,7 @@ npx @mui/x-codemod v6.0.0/data-grid/preset-safe <path|folder>
 The list includes these transformers
 
 - [`column-menu-components-rename`](#column-menu-components-rename)
+- [`remove-disableExtendRowFulWidth-prop`](#remove-disableExtendRowFulWidth-prop)
```
LukasTy, line 212:
```suggestion
- [`remove-disableExtendRowFullWidth-prop`](#remove-disableExtendRowFullWidth-prop)
```

packages/x-codemod/src/util/removeProps.ts
```diff
@@ -0,0 +1,21 @@
+import type { Collection, JSCodeshift } from 'jscodeshift';
+
+interface RemovePropsArgs {
+  root: Collection<any>;
+  componentNames: string[];
+  props: string[];
+  j: JSCodeshift;
+}
+
+export default function removeProps({ root, componentNames, props, j }: RemovePropsArgs) {
```
LukasTy, line 10: 🙌

packages/x-codemod/src/v6.0.0/data-grid/remove-disableExtendRowFullWidth-prop/actual.spec.js
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { DataGrid } from '@mui/x-data-grid';
```
LukasTy, line 2: If we are going with correctness with all the imports. 🙈 
```suggestion
import { DataGrid } from '@mui/x-data-grid';
import { DataGridPro } from '@mui/x-data-grid-pro';
import { DataGridPremium } from '@mui/x-data-grid-premium';
```

But also no one would be mad if this file only had the following 🙈 
```tsx
<>
  <DataGrid disableExtendRowFullWidth />
  <DataGridPro disableExtendRowFullWidth />
  <DataGridPremium disableExtendRowFullWidth />
</>
```

LukasTy (approved):

docs/data/migration/migration-data-grid-v5/migration-data-grid-v5.md
LukasTy: Thanks for clarifying. 😉 
Will continue using "_has been removed_". 👍

PR #7526, opened 2023-01-13
[DateRangeCalendar] Ignore `calendars` prop on mobile

flaviendelangle:
Fixes an inconsistency between the doc and the actual behavior

https://github.com/mui/mui-x/blob/e472a0e6848d68a115e278408e973acddbfdc4de/docs/data/date-pickers/date-range-picker/date-range-picker.md?plain=1#L79


LukasTy (approved):
🙈

PR #7529, opened 2023-01-13
[DateRangeCalendar] Ignore `showDaysOutsideCurrentMonth` when `calendars > 1`

### LINKED ISSUE #6982: [DateRangePicker] Strange selected days highlighting with `showDaysOutsideCurrentMonth`
### Steps to reproduce 🕹

Link to live example: https://codesandbox.io/s/amazing-mendeleev-j1tyyw?file=/demo.tsx

Steps:
1. set `showDaysOutsideCurrentMonth`
2. select date range spanning two months

### Current behavior 😯

The selected days outside of current month are not highlighted correctly.

Screenshot 2022-11-24 at 10 52 00

### Expected behavior 🤔

The selected days outside of current month should receive a "stable" highlighting style.
Either:
- mimic same style from the other month (in the above examples case—provide highlighting style to day 31 in November calendar)
- have only one start and end marker in calendars (both 30 and 31 days in November calendar would receive generic highlighting style)

flaviendelangle:
Fixes #6982

Only manually authored files:

- `PickersDay.tsx` 
- `DateRangeCalendar.tsx`

____

[Example before](https://codesandbox.io/s/focused-butterfly-e89fwk)
[Example after](https://codesandbox.io/s/nervous-dream-iqg3mg)


packages/x-date-pickers/src/PickersDay/PickersDay.tsx
```diff
@@ -19,11 +19,28 @@ import {
   pickersDayClasses,
 } from './pickersDayClasses';
 
+export interface ExportedPickersDayProps {
+  /**
+   * If `true`, today's date is rendering without highlighting with circle.
```
LukasTy, line 24: Maybe it's not worth going into the styling specifics in JSDoc? 🤔 
```suggestion
   * If `true`, today's date is rendered without highlight.
```

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
```diff
@@ -19,11 +19,28 @@ import {
...
+   * @default false
+   */
+  disableHighlightToday?: boolean;
+  /**
+   * If `true`, days outside the current month are rendered:
+   * - if `fixedWeekNumber` is defined, renders days to have the weeks requested.
+   * - if `fixedWeekNumber` is not defined, renders day to fill the first and last week of the current month.
+   * On range pickers, it will be ignored if more than one month is rendered.
```
LukasTy, line 32: Maybe use the code term rather than the doc one in the JSDoc?
```suggestion
   * On range pickers, it will be ignored if more than one `calendars` is rendered.
```

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
```diff
@@ -19,11 +19,28 @@ import {
...
 
+export interface ExportedPickersDayProps {
+  /**
+   * If `true`, today's date is rendering without highlighting with circle.
+   * @default false
+   */
+  disableHighlightToday?: boolean;
+  /**
+   * If `true`, days outside the current month are rendered:
+   * - if `fixedWeekNumber` is defined, renders days to have the weeks requested.
```
LukasTy, line 30: Could we look into formatting it so that the new lines are also rendered in the api doc? 🤔 
Also, how does it look like in IDE?

LukasTy (commented):
Could we also fix this inconsistency while we are at it? 🤔 
Screenshot 2023-01-16 at 11 04 37

packages/x-date-pickers/src/PickersDay/PickersDay.tsx
flaviendelangle: image

In Webstorm it looks fine

In the API page, it's super ugly like all our props with long descriptions

image

We should really priority improving this table... But not sure we can do much on this PR.

What do you have in mind ?

LukasTy: > What do you have in mind ?

I propose the following formatting:
```js
/**
 * If `true`, days outside the current month are rendered:
 *
 * - if `fixedWeekNumber` is defined, renders days to have the weeks requested.
 *
 * - if `fixedWeekNumber` is not defined, renders day to fill the first and last week of the current month.
 *
 * - if `calendars` is set to `1` on range pickers.
 * @default false
 */
showDaysOutsideCurrentMonth?: boolean;
```

This would produce the following result. 
Screenshot 2023-01-16 at 12 33 10
Screenshot 2023-01-16 at 12 33 21

flaviendelangle: I don't understand the `if `calendars` is set to `1` on range pickers.` :angel: 
I applied the rest

LukasTy: I was trying to unify the text formatting, because that last sentence about range pickers with different formatting seemed off. 🤔 
Adding it as another list item IMHO improves the coherency of the description.

flaviendelangle: Hiow about `- ignored if `calendars equals more than `1` on range pickers.` ?

With your proposal, the two first dots explains have the `if A, then B` but the 3rd is just `if A` so I'm waiting for a 2nd part.
Note that the first two dots also applies to the range pickers.

If it was a doc page it would probably be something like:

```
If `true`, days outside the current month are rendered:

- if `fixedWeekNumber` is defined, renders days to have the weeks requested.

- if `fixedWeekNumber` is not defined, renders day to fill the first and last week of the current month.

:::info
On range pickers, it will be ignored if more than one `calendars` is rendered.
:::
```

LukasTy: Fair point. Your suggestion sound more correct. 👌 👍

flaviendelangle: Done :ok_hand:

LukasTy (approved):

PR #7533, opened 2023-01-13
[codemod] Add codemod for `components` to `slots` remaining

alexfauquette:
Part of  #6997 and need #7528 to be merged first (noticed the typo when writing this PR)

I applied the codemod to the docs demonstrations

Not sure if it should be added to `preset-safe` because it's not stickily speaking a breaking change

TypeScript is failing, because the docs still contain some legacy components that do not support `slots`/`slotProps`.
We should warn about using it only after the migration of components to v6


packages/x-codemod/src/v6.0.0/pickers/rename-components-to-slots/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  root
+    .find(j.ImportDeclaration)
+    .filter(({ node }) => {
+      return node.source.value.startsWith('@mui/x-date-picker');
+    })
+
+    .forEach((path) => {
+      path.node.specifiers.forEach((node) => {
+        // Process only Joy UI components
```
alexfauquette, line 41:
```suggestion
        // Process only date-pickers components
```

flaviendelangle:
> Not sure if it should be added to preset-safe because it's not stickily speaking a breaking change

I think it should be included to the v7 preset safe since it will be a breaking change then.
But for now I would keep it as a standalone.

packages/x-codemod/src/v6.0.0/pickers/rename-components-to-slots/actual-props.spec.js
```diff
@@ -0,0 +1,22 @@
+import { NextDatePicker } from '@mui/x-date-pickers';
+import { NextDateRangePicker } from '@mui/x-date-pickers-pro';
+
+<div>
+  <NextDateRangePicker
```
flaviendelangle, line 5: Can we add another component (`TextField` for example) and check if it is renamed as well ?

flaviendelangle (approved):

packages/x-codemod/src/v6.0.0/pickers/rename-components-to-slots/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  const j = api.jscodeshift;
+
+  const printOptions = options.printOptions;
+
+  const root = j(file.source);
+
+  root
+    .find(j.ImportDeclaration)
+    .filter(({ node }) => {
+      return node.source.value.startsWith('@mui/x-date-picker');
```
LukasTy, line 36: Nitpick: Maybe this codemod could not be `pickers` specific but handle both data grid as well as pickers changes?
I believe that both packages did make the change of deprecating the old `components` naming and adding `slots`. 🤔

alexfauquette: It would not make sense with the comande line name
```sh
npx @mui/x-codemod v6.0.0/pickers/migrate-to-components-componentsProps
```

LukasTy: Of course, that's why I was thinking if it wouldn't make sense to just move this codemod to top level? 🤔

MBilalShafi: Or may be create a util and reuse in both data grid and pickers

alexfauquette: Yes, that would allow users to do the migration in two steps: the data grid then the pickers (or the opposit)

PR #7574, opened 2023-01-16
[docs] Update `DemoContainer` `components` prop using a codemod

alexfauquette:
@flaviendelangle I tried to writ what we discussed this morning. Seems to work quite good (not that much delay added).

It does not handle the case where `<DemoContainer/>` children are pickers (not `<DemoItem/>`)


flaviendelangle:
> It does not handle the case where <DemoContainer/> children are pickers (not <DemoItem/>)

Is there something preventing us from supporting it or is it just to save time in the 1st POC ?

alexfauquette:
> Is there something preventing us from supporting it or is it just to save time in the 1st POC?

Two points:
- Saving time
- I'm not sure about what could be inside a `DemoContainer`. Could it have multiple children?

flaviendelangle:
Yes it could
The API I had in mind before discussing automation was `<DemoContainer components={['DatePicker', 'SingleInputDateRangeField', 'TimeField']}`.

Which is slightly more complicated because we need to go from `DemoContainer`, check the children and if they are `DemoItem` or `LocalizationProvider` check their children.

alexfauquette:
@flaviendelangle I confirm `DemoContainer` can be done with the `<DemoContainer components={['DatePicker', 'SingleInputDateRangeField', 'TimeField']}`.

Instead of custom logic about looking for children of children, I extract all the JSX components inside DemoContainer and filter those imported from `@mui/x-date-pickers...`

Same for `Demo` item

packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
```diff
@@ -53,21 +52,52 @@ const getSupportedSectionFromChildName = (childName: string): PickersSupportedSe
...
+  children: React.ReactNode;
+  component: string;
+}
+export function DemoItem(props: DemoItemProps) {
+  const { label, children, component } = props;
 
-  const childrenCount = React.Children.count(children);
-  const childrenTypes = new Set<PickersGridChildComponentType>();
-  const childrenSupportedSections = new Set<PickersSupportedSections>();
+  if (!component) {
```
LukasTy, line 63: I'm not a fan of this extra requirement of this change.
It will pollute the demo previews a lot as well as creates two points of failure/need to edit/change in order to get "correct" layout. 🙈 

What do you think about avoiding the need of `component` prop on `DemoItem` with a tiny loss of flexibility and styling?

```diff
diff --git a/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx b/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
index 78b194dea..0fd67ed39 100644
--- a/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
+++ b/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
@@ -55,35 +55,12 @@ const getSupportedSectionFromChildName = (childName: string): PickersSupportedSe
 interface DemoItemProps {
   label?: React.ReactNode;
   children: React.ReactNode;
-  component: string;
 }
 export function DemoItem(props: DemoItemProps) {
-  const { label, children, component } = props;
-
-  if (!component) {
-    throw new Error('DemoItem has to have a `component` prop defining the component it wraps.');
-  }
-
-  const childName = component;
-  const childType = getChildTypeFromChildName(childName);
-
-  let spacing: StackProps['spacing'];
-  let sx: StackProps['sx'];
-
-  if (childType === 'multi-input-range-field') {
-    spacing = 2;
-    sx = {
-      [`& .${textFieldClasses.root}`]: {
-        flexGrow: 1,
-      },
-    };
-  } else {
-    spacing = 1;
-    sx = undefined;
-  }
+  const { label, children } = props;
 
   return (
-    <Stack direction="column" alignItems="stretch" spacing={spacing} sx={sx}>
+    <Stack direction="column" alignItems="stretch" spacing={1}>
       {label && <Typography variant="body2">{label}</Typography>}
       {children}
     </Stack>
@@ -139,6 +116,15 @@ export function DemoContainer(props: DemoGridProps) {
   } else {
     sx = { [`& > .${textFieldClasses.root}`]: { minWidth: 200 } };
   }
+  // handle extra styling previously handled via `DemoItem`
+  if (childrenTypes.has('multi-input-range-field')) {
+    sx = {
+      ...sx,
+      [`& .${textFieldClasses.root}`]: {
+        flexGrow: 1,
+      },
+    };
+  }
 
   return (
     <Stack direction={direction} spacing={spacing} sx={sx}>

```

alexfauquette:
@LukasTy I did a few modifications:

- removed all the usage of `React.children` in `DemoContainer` to only rely on the `components` prop
- replaced `components={["DatePicker]}` by `component="DatePicker"` in `DemoItem`
- Special case of ComponentExplorerNoSnap which modifies the child depending on inputs, so I hardcoded the `components`/`component` props depending on selected values. And updated the codemod such that if there is no picker component inside, it does not modify the props.
- The codemod does not respect prettier convention, so I added a prettier step in the `test_types` ci

LukasTy:
> I did a few modifications:

Great modifications! 👍 

> replaced `components={["DatePicker]}` by `component="DatePicker"` in `DemoItem`

Scratch my initial solution, it is flawed. But what do you think about the following modification potentially avoiding flaws (given that `DemoItem` is a child of `DemoContainer` and not nested deeper)?
```diff
diff --git a/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx b/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
index 78b194dea..262bb586a 100644
--- a/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
+++ b/packages/x-date-pickers/src/internals/demo/DemoContainer.tsx
@@ -55,22 +55,15 @@ const getSupportedSectionFromChildName = (childName: string): PickersSupportedSe
 interface DemoItemProps {
   label?: React.ReactNode;
   children: React.ReactNode;
-  component: string;
+  component?: string;
 }
 export function DemoItem(props: DemoItemProps) {
   const { label, children, component } = props;
 
-  if (!component) {
-    throw new Error('DemoItem has to have a `component` prop defining the component it wraps.');
-  }
-
-  const childName = component;
-  const childType = getChildTypeFromChildName(childName);
-
   let spacing: StackProps['spacing'];
   let sx: StackProps['sx'];
 
-  if (childType === 'multi-input-range-field') {
+  if (component && getChildTypeFromChildName(component) === 'multi-input-range-field') {
     spacing = 2;
     sx = {
       [`& .${textFieldClasses.root}`]: {
@@ -90,6 +83,8 @@ export function DemoItem(props: DemoItemProps) {
   );
 }
 
+DemoItem.displayName = 'DemoItem';
+
 export function DemoContainer(props: DemoGridProps) {
   const { children, components } = props;
 
@@ -101,6 +96,20 @@ export function DemoContainer(props: DemoGridProps) {
     childrenTypes.add(getChildTypeFromChildName(childName));
     childrenSupportedSections.add(getSupportedSectionFromChildName(childName));
   });
+  const cleanChildren: React.ReactNode[] = [];
+  React.Children.forEach(children, (child: any, index: number) => {
+    cleanChildren.push(
+      child?.type?.displayName !== 'DemoItem'
+        ? child
+        : {
+            ...child,
+            props: {
+              ...child.props,
+              component: components?.[index],
+            },
+          },
+    );
+  });
 
   const getSpacing = (direction: 'column' | 'row') => {
     if (direction === 'row') {
@@ -142,7 +151,7 @@ export function DemoContainer(props: DemoGridProps) {
 
   return (
     <Stack direction={direction} spacing={spacing} sx={sx}>
-      {children}
+      {cleanChildren}
     </Stack>
   );
 }

```

> Special case of ComponentExplorerNoSnap which modifies the child depending on inputs, so I hardcoded the `components`/`component` props depending on selected values. And updated the codemod such that if there is no picker component inside, it does not modify the props.

Yes, I initially skipped it's custom implementation as was going to do a change this morning. 🥱 🙈 👍 

> The codemod does not respect prettier convention, so I added a prettier step in the `test_types` ci

This is somewhat annoying modification we need to do, especially given the fact that we provide `quote: 'single',` option to `jscodeshift` 🤷

alexfauquette:
As discussed, I made `component` optional, and modify the codemod to only add it for multi input fields.

Next step would be to do not care about `DemoItem` in the codemod

docs/scripts/formattedTSDemos.js
```diff
@@ -144,6 +145,18 @@ async function main(argv) {
...
+  const paths = await getFiles(path.join(workspaceRoot, 'docs/data'));
+  const transformPath = path.resolve('scripts/addDemoItemsAttributes.js');
+
+  // Format pickers demos such that `DemoContainer` and `DemoItem` gets correct `components` prop
+  await jscodeshift(transformPath, paths, {
+    extensions: 'js,ts,jsx,tsx',
+    parser: 'tsx',
+    verbose: 2,
+    quote: 'single',
+    trailingComma: true,
```
LukasTy, line 157:
```suggestion
    trailingComma: true,
    wrapColumn: 85
```
The default of `74` is way to "opinionated" 🙈 
Maybe use a value of our `prettier`? 🤔

joserodolfofreitas (approved):
It looks like we have a solid fix to our Demo containers problem. Thanks, @alexfauquette!

-- Not on this particular PR but generally speaking about the DemoContainer:
Overall I like a lot how the demos are displayed (speaking about the visual aspect).
I think it works quite well when demoing multiple components (which was the biggest problem we were aiming to solve ).

On follow-ups, we can consider improving the DX to copy+paste the code.
With multiple components+/internals/demo, the code showcase feels a bit cluttered.
Maybe we can explore mitigating the trade-offs of displaying multiple components, by making it easier to copy the important parts of the demo.

Perhaps something like (this quick idea)
image

docs/scripts/formattedTSDemos.js
alexfauquette: I tried lot of parameters, but the prettier always modify something after

alexfauquette: To avoid problems, I moved the codemod script to a distinct command, such that everybody is free to run it or not, and avoid CI problems

I also opened follow-up issue #7691

I'm not sure this codemod will survive to the resolution of the follow-up issue, and will be happy to delete this command when solved :)

docs/scripts/populatePickersDemos.js
```diff
@@ -0,0 +1,77 @@
+const { run: jscodeshift } = require('jscodeshift/src/Runner');
+/**
+ * Transpiles TypeScript demos to formatted JavaScript.
+ * Can be used to verify that JS and TS demos are equivalent. No introduced change
+ * would indicate equivalence.
```
LukasTy, line 5: This description is probably not the most correct, is it? 🤔 
This script would only apply/change `components/component` prop on existing demos files.

LukasTy (approved):
Great job! 💯 
I think that it is a great compromise. 👌

PR #7545, opened 2023-01-13
[pikers] Drop legacy pickers

flaviendelangle:
I did not spend to much time on the doc since it's going to be impacted again by the renaming.

### Breaking changes

- The legacy pickers (pickers using a masked-based input) have been removed

  ```diff
  - DatePicker
  - DesktopDatePicker
  - MobileDatePicker
  - StaticDatePicker
  
  - TimePicker
  - DesktopTimePicker
  - MobileTimePicker
  - StaticTimePicker

  - DateTimePicker
  - DesktopDateTimePicker
  - MobileDateTimePicker
  - StaticDateTimePicker

  - DateRangePicker
  - DesktopDateRangePicker
  - MobileDateRangePicker
  - StaticDateRangePicker
  ```

The new pickers (pickers using our new field components) will be renamed in a follow up PR to replace those legacy pickers.

PR #7575, opened 2023-01-16
[pickers] Rename new pickers

flaviendelangle:
Follow up #7545
[Doc preview](https://deploy-preview-7575--material-ui-x.netlify.app/)

The diff in line of codes comes from the deletion of the `zh` and `pt` translations that were not generated anymore.


packages/x-date-pickers-pro/src/themeAugmentation/components.d.ts
```diff
@@ -13,29 +13,31 @@ export interface PickersProComponents<Theme = unknown> {
...
-  };
   MuiSingleInputDateRangeField?: {
     defaultProps?: ComponentsProps['MuiSingleInputDateRangeField'];
     styleOverrides?: ComponentsOverrides['MuiSingleInputDateRangeField'];
   };
-  MuiStaticNextDateRangePicker?: {
-    defaultProps?: ComponentsProps['MuiStaticNextDateRangePicker'];
-    styleOverrides?: ComponentsOverrides<Theme>['MuiStaticNextDateRangePicker'];
+
+  // Date Range Pickers
```
LukasTy, line 25: Do you think using sections makes more sense than alphabetical order? 😃

LukasTy (approved):
LGTM.  Great job! 🚀 💯 

We should probably go with both having new chip or none with the chip..? 🤔 
Screenshot 2023-01-17 at 10 14 18

A bit off-topic, but noticed that this section seems somewhat irrelevant given the fact that we do not support direct adapters import as per [code setup](https://next.mui.com/x/react-date-pickers/getting-started/#code-setup). 🤔 
Screenshot 2023-01-17 at 10 16 25

Another off-topic: We should probably add format to the second demo to align with the last one? 🤔 
Screenshot 2023-01-17 at 10 38 20

flaviendelangle:
> We should probably add some breaking changes notes here as well? thinking

If we merge both PRs in the same week (#7575 and this one), I think we can have a single breaking change note saying that the pickers have been replaced by their field-based equivalent and that the unstable version of the field-based have been removed.
With a link to the relevant sections of the migration guide (or inline everything but it could be very verbose)

____

> We should probably go with both having new chip or none with the chip..? thinking


`DateCalendar` exist for more than a year now (named `CalendarPicker` before)
That's why I did not put a `new feature` next to it.

`DateRangeCalendar` has been introduced for the 1st time during this alpha.
That's why I did put a `new feature` next to it.

____

> A bit off-topic, but noticed that this section seems somewhat irrelevant given the fact that we do not support direct adapters import as per [code setup](https://next.mui.com/x/react-date-pickers/getting-started/#code-setup). thinking

We should probably do a little cleanup of the field page indeed

____

packages/x-date-pickers-pro/src/themeAugmentation/components.d.ts
flaviendelangle: In the community package, I think so.
We had legacy pickers missing from this page because it's super long and it's hard to find mistake on it.
That's why I grouped it, if we have a picker with less than 4 elements, something is wrong.

For the pro package, I don't think it brings much value indeed.
I went for consistency, because I don't think it makes the file less readable.

LukasTy:
> If we merge both PRs in the same week (#7575 and this one), I think we can have a single breaking change note saying that the pickers have been replaced by their field-based equivalent and that the unstable version of the field-based have been removed.
> With a link to the relevant sections of the migration guide (or inline everything but it could be very verbose)

Fair point, yes, having smaller wall of text to convey the same idea is always preferred. 🙈 

> `DateCalendar` exist for more than a year now (named `CalendarPicker` before) That's why I did not put a `new feature` next to it.
> 
> `DateRangeCalendar` has been introduced for the 1st time during this alpha. That's why I did put a `new feature` next to it.

Mhh, I might have already mentioned this case in some other PR. Sorry for repeating myself. 🤦‍♂️ 🙈

packages/x-date-pickers-pro/src/themeAugmentation/components.d.ts
LukasTy: > We had legacy pickers missing from this page because it's super long and it's hard to find mistake on it.
> That's why I grouped it, if we have a picker with less than 4 elements, something is wrong.

Mhh, indeed. This file is not trivial to manage. 🙈 
Grouping by component types probably makes more sense. 👍

docs/data/pages.ts
```diff
@@ -277,41 +275,29 @@ const pages: MuiPage[] = [
...
-            pathname: '/x/api/date-pickers/mobile-next-date-time-picker',
-            title: 'MobileNextDateTimePicker',
-          },
-          {
-            pathname: '/x/api/date-pickers/mobile-next-time-picker',
-            title: 'MobileNextTimePicker',
+            pathname: '/x/api/date-pickers/mobile-date-time-picker',
+            title: 'MobileDateTimePicker',
           },
+          { pathname: '/x/api/date-pickers/mobile-picker', title: 'MobileTimePicker' },
```
oliviertassinari, line 300: 404 fixed in https://github.com/mui/mui-x/commit/d2bd7eef3948b1b29bf2d4c5f16e21ede4268f1f. 

In the future, we could automate the generation of these pages. The Next.js page is generated in `buildComponentsDocumentation` so I think that we could also generate these API routes, like it's done in the main repo: https://github.com/mui/material-ui/blob/8ec664cae22b027d128363b15c9ed2c9dc4402b5/docs/data/base/pages.ts#L1

PR #7577, opened 2023-01-16
[docs] List all pickers toolbar pages in api docs side menu

LukasTy:
Noticed that we are changing contents of these pages quite often, but they are not listed anywhere.
We did list `DateTimePickerTabs` component, I do not see a reason to not list these components.


flaviendelangle (approved):
Nice catch
They should probably also be listed in `api/index.md`

LukasTy:
> They should probably also be listed in `api/index.md`

Yup. I've added those in `Toolbars` group.
I've also added the `Index` Api reference submenu item as the first child—same as in Data Grid api docs. 😉

docs/pages/x/api/date-pickers/index.md
```diff
@@ -55,6 +55,13 @@
 
 - [TimeClock](/x/api/date-pickers/time-clock/)
 
+### Toolbars
+
+- [DatePickerToolbar](/x/api/date-pickers/date-picker-toolbar/)
```
flaviendelangle, line 60: Alternatively we could put it next to their picker (like for the tabs)

If we keep them here, could we keep the regular order "Date => Time => Date Time => Date Range" that we tend to use most of the time.

PR #7581, opened 2023-01-17
[test] Fix date range picker tests to work with western time zones

m4theushw:
These failing tests were bothering me for quite a long time but I always ignored them. What happens is that when the clock is mocked, the initial date is `1970-01-01T00:00:00.000Z`, however, in time zones to the west of UTC (e.g. USA or South America) the local date is actually `1969-12-31`. This causes problems when calculating the next or previous day and asserting their disabled state, because the month changes in these time zones. Here's how the calendar view looks like in my time zone (UTC -3) if I pass the date used by the fake clock to the `value`, enable the `disablePast` prop and also force to consider "now" as being the date above.

image

This situation explains why the following assertion fails in western time zones but not on the CI. Since tomorrow is `1970-01-01` in my time zone, the day `1` is disabled in the left calendar, but the test expects it to be enabled. The test is not considering that the month may have changed. In a time zone with positive offset, tomorrow would be `1970-01-02` and it would be enabled.

https://github.com/mui/mui-x/blob/fd5573cedb92640f16d8d4f5370aec3bf18b8eaa/packages/x-date-pickers-pro/src/tests/describeRangeValidation/testDayViewRangeValidation.tsx#L160

A similar wrong assertion occurs with `disableFuture` but in opposite direction.

To simulate the problem and test the solution you can change the time zone of your operating system to UTC -3 and run `yarn test:unit`.

### DIFF
--- a/docs/data/date-pickers/adapters-locale/AmPMCustomization.tsx
+++ b/docs/data/date-pickers/adapters-locale/AmPMCustomization.tsx
@@ -11,7 +11,7 @@ import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 
 const locales = ['en', 'en-gb', 'de'];
 
-type LocaleKey = typeof locales[number];
+type LocaleKey = (typeof locales)[number];
 
 export default function AmPMCustomization() {
   const [locale, setLocale] = React.useState<LocaleKey>('en');


packages/x-date-pickers-pro/src/tests/describeRangeValidation/testDayViewRangeValidation.tsx
```diff
@@ -157,7 +157,11 @@ export function testDayViewRangeValidation(ElementToTest, getOptions) {
       const yesterday = adapterToUse.addDays(now, -1);
 
       testDisabledDate(adapterToUse.format(now, 'dayOfMonth'), [false, false], isDesktop);
-      testDisabledDate(adapterToUse.format(tomorrow, 'dayOfMonth'), [false, false], isDesktop);
+      testDisabledDate(
```
LukasTy, line 160: I would suggest fixing the root cause instead.
This time zone discrepancy happens because creating a new date with `new Date` and no hours specified applies the time zone offset.
This issue can be mitigated by setting explicit hours, minutes, seconds and miliseconds when creating a date.
I've replicated the issue locally and fixed it by applying the following changes:

```diff
diff --git a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePicker.test.tsx
index 84fced809..0c621b77e 100644
--- a/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/DateRangePicker/DateRangePicker.test.tsx
@@ -8,7 +8,10 @@ import { expect } from 'chai';
 import { createPickerRenderer, stubMatchMedia, wrapPickerMount } from 'test/utils/pickers-utils';
 
 describe('<DateRangePicker />', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeConformance(
     <DateRangePicker
diff --git a/packages/x-date-pickers-pro/src/DesktopNextDateRangePicker/describes.DesktopNextDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/DesktopNextDateRangePicker/describes.DesktopNextDateRangePicker.test.tsx
index feec3d5df..016fafe10 100644
--- a/packages/x-date-pickers-pro/src/DesktopNextDateRangePicker/describes.DesktopNextDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/DesktopNextDateRangePicker/describes.DesktopNextDateRangePicker.test.tsx
@@ -10,7 +10,10 @@ import {
 } from 'test/utils/pickers-utils';
 
 describe('<DesktopNextDateRangePicker /> - Describes', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   const { clickOnInput } = buildFieldInteractions({ clock });
 
diff --git a/packages/x-date-pickers-pro/src/MobileDateRangePicker/MobileDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/MobileDateRangePicker/MobileDateRangePicker.test.tsx
index e2005e94e..d3138ae3a 100644
--- a/packages/x-date-pickers-pro/src/MobileDateRangePicker/MobileDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/MobileDateRangePicker/MobileDateRangePicker.test.tsx
@@ -25,7 +25,10 @@ const WrappedMobileDateRangePicker = withPickerControls(MobileDateRangePicker)({
 });
 
 describe('<MobileDateRangePicker />', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeConformance(
     <MobileDateRangePicker
diff --git a/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.test.tsx
index e588bdbdb..228f57aa8 100644
--- a/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/MobileNextDateRangePicker.test.tsx
@@ -13,7 +13,10 @@ import {
 import { DateRange } from '@mui/x-date-pickers-pro';
 
 describe('<MobileNextDateRangePicker />', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeConformance(<MobileNextDateRangePicker />, () => ({
     classes: {},
diff --git a/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/describes.MobileNextDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/describes.MobileNextDateRangePicker.test.tsx
index f8c87e491..64cc149b8 100644
--- a/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/describes.MobileNextDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/MobileNextDateRangePicker/describes.MobileNextDateRangePicker.test.tsx
@@ -10,7 +10,10 @@ import {
 } from 'test/utils/pickers-utils';
 
 describe('<MobileNextDateRangePicker /> - Describes', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeRangeValidation(MobileNextDateRangePicker, () => ({
     render,
diff --git a/packages/x-date-pickers-pro/src/StaticDateRangePicker/StaticDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/StaticDateRangePicker/StaticDateRangePicker.test.tsx
index f6d822673..1040a62e2 100644
--- a/packages/x-date-pickers-pro/src/StaticDateRangePicker/StaticDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/StaticDateRangePicker/StaticDateRangePicker.test.tsx
@@ -24,7 +24,10 @@ const WrappedStaticDateTimePicker = withPickerControls(StaticDateRangePicker)({
 });
 
 describe('<StaticDateRangePicker />', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeConformance(
     <StaticDateRangePicker
diff --git a/packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.test.tsx b/packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.test.tsx
index 001316372..63e36362f 100644
--- a/packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.test.tsx
+++ b/packages/x-date-pickers-pro/src/StaticNextDateRangePicker/StaticNextDateRangePicker.test.tsx
@@ -7,7 +7,10 @@ import { wrapPickerMount, createPickerRenderer, adapterToUse } from 'test/utils/
 import { describeRangeValidation } from '@mui/x-date-pickers-pro/tests/describeRangeValidation';
 
 describe('<StaticNextDateRangePicker />', () => {
-  const { render, clock } = createPickerRenderer({ clock: 'fake' });
+  const { render, clock } = createPickerRenderer({
+    clock: 'fake',
+    clockConfig: new Date(2018, 0, 1, 0, 0, 0, 0),
+  });
 
   describeConformance(<StaticNextDateRangePicker />, () => ({
     classes: {},

```

flaviendelangle (approved):

LukasTy (approved):
Thank you for taking the time to handle these oversights! 💯

PR #6964, opened 2022-11-22
[core] Unify testing of the value related props

flaviendelangle:
## Work

- [x] Add `describeValidation` on `MonthCalendar` 
- [x] New `describeValue` that checks the `value` / `defaultValue` behaviors for all components and the `onClose` / `onAccept` / `onChange` behaviors for all pickers (when using the various action bar actions for instance)

## Side notes

- We have missing `describeConformance` on pickers, with the new file structure it's very easy to find the missing one, I'll add them in a follow up PR and try to reduce the `skip` that no longer apply
- Adapt `describeValue` for range components later

PR #7587, opened 2023-01-17
[pickers] Spread props to the DOM in `DateCalendar` and `TimeClock`

flaviendelangle:


packages/x-date-pickers/src/TimeClock/TimeClock.test.tsx
```diff
@@ -15,27 +15,6 @@ import { adapterToUse, wrapPickerMount, createPickerRenderer } from 'test/utils/
 describe('<TimeClock />', () => {
   const { render } = createPickerRenderer();
 
-  describeConformance(
```
flaviendelangle, line 18: `describeConformance` was defined twice for this component

packages/x-date-pickers/src/dateViewRenderers/dateViewRenderers.tsx
```diff
@@ -16,13 +16,84 @@ export interface DateViewRendererProps<TDate, TView extends DateOrTimeView>
...
+  disabled,
+  showDaysOutsideCurrentMonth,
+  dayOfWeekFormatter,
+  sx,
+  autoFocus,
+  fixedWeekNumber,
+  displayWeekNumber,
+}: DateViewRendererProps<TDate, any>) => (
   <DateCalendar
-    {...props}
```
LukasTy, line 23: What was the problem you are solving with the removal of spread? 🤔

packages/x-date-pickers/src/TimeClock/tests/describes.TimeClock.test.tsx
```diff
@@ -17,26 +17,15 @@ import {
...
-        'componentsProp',
-        'propsSpread',
-        'reactTestRenderer',
-        // TODO: fix TimeClock to spread props to root
-        'themeDefaultProps',
-        'themeVariants',
-      ],
-    }),
-  );
+  describeConformance.only(<TimeClock />, () => ({
```
LukasTy, line 20:
```suggestion
  describeConformance(<TimeClock />, () => ({
```

packages/x-date-pickers/src/TimeClock/TimeClock.tsx
```diff
@@ -341,7 +341,7 @@ export const TimeClock = React.forwardRef(function TimeClock<TDate extends unkno
       ref={ref}
       className={clsx(classes.root, className)}
       ownerState={ownerState}
-      sx={sx}
+      {...other}
```
LukasTy, line 344: It does seem to forward the ref, but after these changes the generated documentation states that it no longer does... 🤷‍♂️

flaviendelangle: The script analyzing the test files to know if it spread the ref does not handle the `describes.XXX.test.tsx`, we need to update it, see https://github.com/mui/mui-x/pull/6964#discussion_r1036794325

LukasTy: Ah yes, forgot that. 🤦

packages/x-date-pickers/src/dateViewRenderers/dateViewRenderers.tsx
flaviendelangle: The renderer receives all the props passed to the pickers.
Including props only intended for the fields or for the picker layer (not for the view).
So if we forward everything and spread to the root, React complains that we are passing unwanted props to the root (things like `format` or `inputRef` will be passed to the DOM)

packages/x-date-pickers/src/tests/describeValidation/testDayViewValidation.tsx
```diff
@@ -181,44 +181,6 @@ export const testDayViewValidation: DescribeValidationTestSuite = (ElementToTest
       expect(screen.getByLabelText('Next month')).to.have.attribute('disabled');
     });
 
-    it('should apply minTime', function test() {
```
flaviendelangle, line 184: Those two tests were only run on the calendar and it now fire an error because `minTime` and `maxTime` is passed to the DOM

LukasTy (approved):

PR #7589, opened 2023-01-17
[pickers] Fix day calendar row and column index

### LINKED ISSUE #6886: [pickers] Improve `DayCalendar` column a11y
Created from https://github.com/mui/mui-x/pull/6144#discussion_r986975316.

When we will introduce the `week numbers` feature—the screen reader announcement of a currently active column will be in incorrect (`1st day of the week` will be announced as `column 2`).
We should address this issue with either the solution provided in the linked comment or look for a different way to fix this inconsistency.

LukasTy:
Fixes #6886 
Add explicit `aria-rowindex` on the week row and `aria-colindex` on the day to have correct numbers announced by the reader.
The indexes are skewed because of the week day label row and optional week number column.

### DIFF
--- a/packages/x-date-pickers/src/DateCalendar/DayCalendar.tsx
+++ b/packages/x-date-pickers/src/DateCalendar/DayCalendar.tsx
@@ -229,6 +229,7 @@ function WrappedDay<TDate extends unknown>({
   isDateDisabled,
   currentMonthNumber,
   isViewFocused,
+  ...other
 }: Pick<PickersDayProps<TDate>, 'onFocus' | 'onBlur' | 'onKeyDown' | 'onDaySelect'> & {
   parentProps: DayCalendarProps<TDate>;
   day: TDate;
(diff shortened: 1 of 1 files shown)


packages/x-date-pickers/src/DateCalendar/DayCalendar.tsx
```diff
@@ -272,6 +273,7 @@ function WrappedDay<TDate extends unknown>({
       isAnimating: isMonthSwitchingAnimating,
       // it is used in date range dragging logic by accessing `dataset.timestamp`
       'data-timestamp': utils.toJsDate(day).valueOf(),
+      ...other,
```
flaviendelangle, line 276: If we spread, we can remove `onDaySelect`, `onKeyDown`, `onBlur` and `onFocus` for the explicitly defined props.

flaviendelangle (approved):
Looks great :partying_face:

PR #7378, opened 2023-01-03
[docs] New page for `DateRangeCalendar`

flaviendelangle:
- [Date Range Calendar](https://deploy-preview-7378--material-ui-x.netlify.app/x/react-date-pickers/date-range-calendar/)
- [Date Range Picker](https://deploy-preview-7378--material-ui-x.netlify.app/x/react-date-pickers/date-range-picker/)

PR #7600, opened 2023-01-18
[pickers] Make sure the `className` and `sx` props are applied to the field / static root of the picker and never to the view

flaviendelangle:
- Foward the `className` and `sx` to the field on non-static pickers
- Forward `className` and `sx` to the layout root on static pickers
- Add a safeguard in the `usePickerViews` to never forward `sx` and `className` to the view. I'm adding it here instead of removing them from the view renderers, because those are props the view accepts, so when building custom views, people can expect the `className` or `sx` they receive in their view renderers to be the one to forward to the root.
- Handle the `inputRef` and `label` props in the headless pickers. initially I did it in the components because I thought single item and range pickers would have the same headless implementation (and so this hook could not decide what to do with those props). But since we have a different hook for range pickers, we can move the logic to the hook to centralize and avoid mistakes.


packages/x-date-pickers/src/MobileTimePicker/MobileTimePicker.tsx
```diff
@@ -47,10 +47,6 @@ const MobileTimePicker = React.forwardRef(function MobileTimePicker<TDate>(
         ...resolveComponentProps(defaultizedProps.slotProps?.field, ownerState),
         ...extractValidationProps(defaultizedProps),
         ref,
```
flaviendelangle, line 49: I would like to also handle the ref in `useDesktopPicker` / `useMobilePicker` / ...
But it creates a TS error when I stop passing it through the `slotProps` in the picker file.
And I struggle to fix it, so I did not touch this behavior for now.

docs/data/date-pickers/date-range-picker/ResponsiveDateRangePickers.tsx
```diff
@@ -30,6 +31,11 @@ export default function ResponsiveDateRangePickers() {
         <DemoItem label="Static variant">
           <StaticDateRangePicker
             defaultValue={[dayjs('2022-04-07'), dayjs('2022-04-10')]}
+            sx={{
```
flaviendelangle, line 34: Fixes https://github.com/mui/mui-x/pull/7378#discussion_r1065670519

It's this example who made me realize we were forwarding the `className` and `sx` to the view on the static pickers instead of the layout root.

LukasTy (commented):
Looks like a good refinement! 🙌 
Just need to take care of some final details. 😉 

These changes seem to also cause a regression with `CustomInput` cases, where a custom `textField` slot component is passed.
It no longer receives provided `className` or `sx` props it used to. 🤔

packages/x-date-pickers/src/DesktopDatePicker/DesktopDatePicker.tsx
```diff
@@ -53,10 +53,6 @@ const DesktopDatePicker = React.forwardRef(function DesktopDatePicker<TDate>(
         ...resolveComponentProps(defaultizedProps.slotProps?.field, ownerState),
         ...extractValidationProps(defaultizedProps),
         ref,
-        className,
-        sx,
```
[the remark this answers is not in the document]
flaviendelangle: I forgot to remove the spread of `defaultizedProps`
Fixed :+1:

LukasTy (approved):
LGTM.
Both the issues I mentioned are now indeed resolved. 👌

PR #7603, opened 2023-01-18
[codemod] Rename `rowsPerPageOptions` prop to `pageSizeOptions`

MBilalShafi:
Part of #7086 
Codemod for breaking change done in https://github.com/mui/mui-x/pull/7147

### DIFF
--- a/packages/x-codemod/src/v6.0.0/data-grid/preset-safe/index.ts
+++ b/packages/x-codemod/src/v6.0.0/data-grid/preset-safe/index.ts
@@ -1,12 +1,12 @@
 import transformColumnMenu from '../column-menu-components-rename';
-import removeDisableExtendRowFullWidth from '../remove-disableExtendRowFullWidth-prop';
 import renameRowsPerPageOptions from '../rename-rowsPerPageOptions-prop';
+import removeDisableExtendRowFullWidth from '../remove-disableExtendRowFullWidth-prop';
 import { JsCodeShiftAPI, JsCodeShiftFileInfo } from '../../../types';
 
 export default function transformer(file: JsCodeShiftFileInfo, api: JsCodeShiftAPI, options: any) {
   file.source = transformColumnMenu(file, api, options);
-  file.source = removeDisableExtendRowFullWidth(file, api, options);
   file.source = renameRowsPerPageOptions(file, api, options);
+  file.source = removeDisableExtendRowFullWidth(file, api, options);
 
   return file.source;
 }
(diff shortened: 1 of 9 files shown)


packages/x-codemod/src/v6.0.0/data-grid/preset-safe/index.ts
```diff
@@ -1,10 +1,12 @@
 import transformColumnMenu from '../column-menu-components-rename';
 import removeDisableExtendRowFullWidth from '../remove-disableExtendRowFullWidth-prop';
+import renameRowsPerPageOptions from '../rename-rowsPerPageOptions-prop';
 import { JsCodeShiftAPI, JsCodeShiftFileInfo } from '../../../types';
 
 export default function transformer(file: JsCodeShiftFileInfo, api: JsCodeShiftAPI, options: any) {
   file.source = transformColumnMenu(file, api, options);
   file.source = removeDisableExtendRowFullWidth(file, api, options);
+  file.source = renameRowsPerPageOptions(file, api, options);
```
LukasTy, line 9: Nitpick: This transformation is done last, but in the docs placed in the middle (second).
Doesn't make a lot of difference, but when reviewing and seeing the first files I would have assumed that the order of transformations would matter for this codemod. 🙈

packages/x-codemod/src/v6.0.0/data-grid/rename-rowsPerPageOptions-prop/index.ts
```diff
@@ -0,0 +1,17 @@
+import renameProps from '../../../util/renameProps';
+import { JsCodeShiftAPI, JsCodeShiftFileInfo } from '../../../types';
+
+const componentNames = ['DataGrid', 'DataGridPro', 'DataGridPremium'];
+const props = { rowsPerPageOptions: 'pageSizeOptions' };
```
LukasTy, line 5: Nitpick: You could have considered having a single codemod handling the rename of all the changed props.
Or wouldn't it have worked? 🤔

LukasTy (approved):
LGTM. 🙌

packages/x-codemod/src/v6.0.0/data-grid/rename-rowsPerPageOptions-prop/index.ts
MBilalShafi: That's actually a good point. 👍 
We discussed the approach and agreed on grouping codemods based on topic/PR because.

1. It will be easy that way to split the codemods and develop them iteratively as the PRs go by.
2. Possibly a user is heavily using a specific feature or set of features and they may want to run the codemods for those features only.
3. Same approach is being followed in the core, so they split codemods based on a specific component/set of related components.

Although to me it's fine other way around too as most of the stable mods will anyways go into preset-safe. So mostly it's not a big deal, we can group related codemods to save some clutter.

packages/x-codemod/src/v6.0.0/data-grid/preset-safe/index.ts
MBilalShafi: Nice catch 😄 Thanks

PR #7605, opened 2023-01-18
[docs] Clean the field page

flaviendelangle:
Part of https://github.com/mui/mui-x/pull/7575#pullrequestreview-1251325513


docs/data/date-pickers/fields/fields.md
```diff
@@ -33,23 +23,17 @@ The field components requires an adapter imported from `@mui/x-date-pickers` or
...
-New pickers using them will be released in a few months.
-
-## Form props
+The picker components (e.g. `DatePicker`) are using the field to handle their input editing.
 
-The standard form attributes (`disabled`, `readOnly`) are supported by all the field components:
+:::warning
+The single input range fields are not supported on the range pickers yet (you can't use `SingleInputDateRangeField` in `DateRangePicker`). It's coming.
 
-{{"demo": "FormProps.js", "defaultCodeOpen": false}}
```
flaviendelangle, line 52: When I wrote this page, we had no detailed page for each field.
Now each field has an equivalent of this section, so I don't think it's useful to keep it here.

docs/data/date-pickers/fields/fields.md
```diff
@@ -33,23 +23,17 @@ The field components requires an adapter imported from `@mui/x-date-pickers` or
...
 {{"demo": "DateRangeFieldExamples.js", "defaultCodeOpen": false}}
 
 ### Usage with the pickers
 
-The picker components (`DatePicker`, `DestkopDatePicker`, `MobileDatePicker`, `DateTimePicker`, ...) do not use the field components yet.
-
-New pickers using them will be released in a few months.
-
-## Form props
+The picker components (e.g. `DatePicker`) are using the field to handle their input editing.
```
flaviendelangle, line 30: If you have other things to add here, it's very short :laughing:

docs/data/date-pickers/fields/DateRangeFieldExamples.tsx
```diff
@@ -24,9 +26,15 @@ export default function DateRangeFieldExamples() {
         <DemoItem label="MultiInputTimeRangeField">
           <MultiInputTimeRangeField defaultValue={[date1, date2]} />
         </DemoItem>
+        <DemoItem label="SingleInputTimeRangeField">
```
LukasTy, line 29: Could we maybe stick to one order of components in this demo?
Let's go with either of these
- Single -> Multi
- Multi -> Single <- probably makes more sense as this order is used in individual page demos

docs/data/date-pickers/fields/fields.md
alexfauquette: You could also remove this section, and let the following warning be in the previous section

> The single input range fields are not supported...

I assume most of the reader will notice that fields are also present in other pages

alexfauquette (approved):

LukasTy (approved):
A welcome and needed cleanup 👍

docs/data/date-pickers/fields/DateRangeFieldExamples.tsx
flaviendelangle: Indeed !
Fixed

docs/data/date-pickers/fields/fields.md
flaviendelangle: I moved the warning to `Fields to edit a range` as proposed :+1:

PR #7610, opened 2023-01-18
[pickers] Support `UTC` with `dayjs`

### LINKED ISSUE #6948: [pickers] How to set a global timezone on the pickers with `dayjs`
### Order ID 💳

50497

### Duplicates

### Latest version

### The problem in depth 🔍

Hi there 👋 My team and I are currently trying to set a global timezone (UTC 00:00) for our entire React application. This is due to how the backend handles and serves dates: no times are included and is assumed to be UTC 00:00.

The expectation would be that regardless of your local timezone, a user interacting with the [DateRangePicker](https://mui.com/x/api/date-pickers/date-range-picker/) from Tokyo should see the same highlighted `today` date as a user in San Francisco, as well as any dates selected.

Current behaviour viewing from San Francisco:
image

Current behaviour viewing from Tokyo:
Screenshot 2022-11-21 at 4 30 33 PM

I've provided a [codebox demo](https://codesandbox.io/s/daterangepicker-set-global-timezone-ez6ojp?file=/src/App.js) of roughly how we're attempting to accomplish this, using the `LocalizationProvider` and passing `dayjs.utc` as the instance, which I understand is what the `DateRangePicker` will base it's dates from, but this raises the `TypeError`: `Cannot read properties of undefined (reading 'en')`. Any suggestions/feedback would be appreciated 🙏

### Your environment 🌎

[Codesandbox Demo](https://codesandbox.io/s/daterangepicker-set-global-timezone-ez6ojp?file=/src/App.js)

flaviendelangle:
Closes #6948
[Doc preview](https://deploy-preview-7610--material-ui-x.netlify.app/x/react-date-pickers/adapters-locale/#use-utc-dates)

First step in the major task of having UTC and timezone support with as many date libraries as possible
I'll start by working on `dayjs` since it's the library we a recommending.


packages/x-date-pickers/src/AdapterDayjs/index.ts
```diff
@@ -46,25 +46,36 @@ export class AdapterDayjs extends BaseAdapterDayjs implements MuiPickersAdapter<
 
   constructor(options: Opts) {
```
flaviendelangle, line 47: See https://github.com/mui/mui-x/issues/6948#issuecomment-1323220551

`dayjs.utc` is almost a full instance of `dayjs`
It just lack the `extend` and `Ls` properties.

packages/x-date-pickers/src/AdapterDayjs/AdapterDayjs.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    });
+  });
+
+  it('should return the correct week number', () => {
+    const localizedAdapter = new AdapterDayjs({ locale: 'fr' });
+
+    const dateToTest = localizedAdapter.date(new Date(2022, 10, 10));
+
+    expect(localizedAdapter.getWeekNumber!(dateToTest)).to.equal(45);
+  });
```
LukasTy, line 137: Maybe put this test inside of `Localization` describe block? After all, it is taken from the original test file 🤔

LukasTy (approved):
Have tested what I could think of.
Looks good to me. 👍 
Great job with the investigation! 💯

docs/data/date-pickers/field-customization/HeadlessBrowserDateField.js
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
```
LukasTy, line 1: Should these have gone in this PR? They don't seem to be used anyhwere 🙈

Issue #7615, opened 2023-01-18
[fields] The expanded format is not correct when using `luxon`


rodrigofariow:
### Duplicates

- [X] I have searched the existing issues

### Latest version

- [X] I have tested the latest version

### Steps to reproduce 🕹

I noticed some time formats are not consistent between `NextTimePicker` and `NextDateTimePicker`.

For instance, on `NextTimePicker` the behaviour for locale `pt-PT` is fine:
https://codesandbox.io/s/boring-albattani-bppmff?file=/demo.tsx

I'm able to write an hour within the 24h system (which is correct for pt-PT):
![image](https://user-images.githubusercontent.com/22502084/213237033-ab9a1a30-267d-4d72-86a9-24de51641f6d.png)

For `NextDateTimePicker` the behaviour for locale `pt-PT` is faulty:
https://codesandbox.io/s/vibrant-rumple-23d0uh?file=/demo.tsx

I'm not able to write an hour past 12. For instance, if I want to write 13 it just shows 1:
![image](https://user-images.githubusercontent.com/22502084/213238742-d722ba23-f5ed-427c-b6d5-83e1906ad1ec.png)

Not sure if this is a problem with `luxon` or `NextDateTimePicker` but I was hoping you could clarify.

Thanks!

### Your environment 🌎

System:
    OS: macOS 13.1
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 109.0.5414.87
    Edge: Not Found
    Firefox: Not Found
    Safari: 16.2
  npmPackages:
    @emotion/react: 11.10.4 => 11.10.4 
    @emotion/styled: 11.10.4 => 11.10.4 
    @mui/icons-material: 5.11.0 => 5.11.0 
    @mui/lab: 5.0.0-alpha.113 => 5.0.0-alpha.113 
    @mui/material: 5.11.1 => 5.11.1 
    @mui/system: 5.11.1 => 5.11.1 
    @mui/x-data-grid: 6.0.0-alpha.14 => 6.0.0-alpha.14 
    @mui/x-data-grid-generator: 6.0.0-alpha.14 => 6.0.0-alpha.14 
    @mui/x-data-grid-premium: 6.0.0-alpha.14 => 6.0.0-alpha.14 
[...]

PR #7618, opened 2023-01-19
[fields] Fix upper boundary on 12-hours sections

flaviendelangle:
Part of #7615


packages/x-date-pickers/src/TimeField/tests/editing.TimeField.test.tsx
```diff
@@ -341,6 +346,19 @@ describe('<TimeField /> - Editing', () => {
         keyStrokes: [{ value: '1', expected: '25' }],
       });
     });
+
+    it('should go to the next section when pressing `2` in a 12-hours format', () => {
```
alexfauquette, line 350: Could you add a test "I should set 3 when pressing 1 then 3" Otherwise the fix sounds good 👍

alexfauquette (approved):

packages/x-date-pickers/src/TimeField/tests/editing.TimeField.test.tsx
flaviendelangle: Done :+1:

PR #7620, opened 2023-01-19
[pickers] Update focus when opening a UI view

### LINKED ISSUE #7177: [pickers] Focus management broken with keyboard
### Duplicates

### Latest version

### Steps to reproduce 🕹

Link to live example:

Steps:
1. Go to https://next.mui.com/x/react-date-pickers/date-time-picker/#basic-usage
2. Open picker
3. Notice keyboard navigation working
4. Close picker
5. Open picker
6. press keyboard arrows up/down to see that it modifies the field, and not the focus in the calendar

https://user-images.githubusercontent.com/45398769/207285995-e5ede264-696e-4ae1-8f61-5d4bf531b4db.mp4

### Current behavior 😯

When opening the calendar with the keyboard, pickers are stuck in a mixed behavior; Field does not highlight the selected item but keyboard navigation impact them.

### Expected behavior 🤔

When opening the popper, the focus must be on the popper and so field should not react

### Context 🔦

_No response_

### Your environment 🌎

<details>

```
```
</details>

### Order ID 💳 (optional)

_No response_

alexfauquette:
Fix #7177

I did not find any effect taking care of setting the focus on open.

Why did it work the first time you open the popper and not after?

Because in `useDesktopPicker` we can see that `usePicker` is called with `autoFocusView: true` but `autoFocus` is only here for the first render (it set the default focused vie to `'day'` instead of `'null'`)

### DIFF
--- a/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
+++ b/packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
@@ -219,6 +219,7 @@ export const usePickerViews = <
     if (newView !== view) {
       setView(newView);
     }
+    setFocusedView(newView, true);
   }, [open]); // eslint-disable-line react-hooks/exhaustive-deps
 
   const layoutProps: UsePickerViewsLayoutResponse<TView> = {
--- a/packages/x-date-pickers/src/DateCalendar/tests/keyboard.DateCalendar.test.tsx
+++ b/packages/x-date-pickers/src/DateCalendar/tests/keyboard.DateCalendar.test.tsx
@@ -122,5 +122,20 @@ describe('<DateCalendar /> keyboard interactions', () => {
         });
       });
     });
+
+    describe('navigate months', () => {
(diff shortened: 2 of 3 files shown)


packages/x-date-pickers/src/internals/hooks/usePicker/usePickerViews.ts
```diff
@@ -219,6 +219,7 @@ export const usePickerViews = <
     if (newView !== view) {
       setView(newView);
     }
+    setFocusedView(newView, true);
```
LukasTy, line 222: Seems to do the trick 🙌

LukasTy (approved):
Thank you for looking into it. 👍 

By the way, was it a thing we changed in v6 or is the following another focus behavior regression?

1. Open Date picker
2. Tab to focus view switch button
3. Press <kbd>Enter</kbd> to switch view

Notice the focus stays on the switch button in v6, whereas in v5 the focus moves to the content of the switched view.

PR #7624, opened 2023-01-19
[l10n] Improve Japanese (ja-JP) locale

makoto14:
Adds missing translations for Japanese (ja-JP) locale - https://github.com/mui/mui-x/issues/3211.


alexfauquette (approved):
Thanks for your contribution 🙏

PR #7632, opened 2023-01-20
[DataGrid] Fix outdated links to localeTextConstants.ts

Vishal-Patil-Gslab:
Signed-off-by: Vishal-Patil-Gslab <100920255+Vishal-Patil-Gslab@users.noreply.github.com>


alexfauquette:
Since v4 the scripts used for the update have change, I do not know which one should be run

oliviertassinari (commented):
I think that we shouldn't much about v4, we can change the base to be on the `next` branch (we need to invest time with a 12+ month target in mind).

Looking at https://next.mui.com/x/api/data-grid/data-grid/#props, I think that we could improve things going forward:

1. We could point to the guide docs, rather than the source directly, so we can provide more context:

```jsx
  /**
   * Set the locale text of the grid.
   * You can find all the translation keys supported in [the docs](https://mui.com/x/react-data-grid/localization/#translation-keys).
   */
  localeText: Partial<GridLocaleText>;
```

2. Still, what do we do with the link to the source in the guide docs, e.g. "[the source](https://github.com/mui-org/material-ui-x/blob/HEAD/packages/grid/_modules_/grid/constants/localeTextConstants.ts)" in https://v4.mui.com/components/data-grid/localization/#translation-keys? I think that the docs markdown parser could replace all these HEAD links automatically with the version tag when rendering. Effectively, the same strategy as in https://github.com/mui/material-ui/pull/23821.

3. And then, once in the docs, developers can check older versions of the docs. We don't do it as well as https://developer.hashicorp.com/terraform/cdktf/concepts/resources or https://www.postgresql.org/docs/9.2/queries-with.html, but we could get there.

flaviendelangle:
Closing this one, the longer we wait the less relevant the fix becomes

PR #7639, opened 2023-01-20
Codemod WIP

alexfauquette:


packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * Recusive function that:
+ * Return a jscodeshift object with `value` associated to path.
+ * The path cons be such as 'tabs.hidden' whcih will return `{ tabs: { hidden: value } }`.
```
LukasTy, line 4:
```suggestion
 * The path can be such as 'tabs.hidden' which will return `{ tabs: { hidden: value } }`.
```

packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
...
+ * If object is not null, path/value will be added respecting the other rpoperties of the object.
```
LukasTy, line 5:
```suggestion
 * If object is not null, path/value will be added respecting the other properties of the object.
```

packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
...
+  if (object === null) {
+    // Simplest case, no object to take into consideration
+    const propertyToAdd = j.objectProperty(
+      j.identifier(targetKey),
+      addItemToObject(remainingPath, value, null, j),
+    );
+    return j.objectExpression([propertyToAdd]);
+  }
+
+  // Look if the oobject we got already contains the property we have to use.
```
LukasTy, line 35:
```suggestion
  // Look if the object we got already contains the property we have to use.
```

packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
...
+  const correspondingObject = (object.properties ?? []).find(
+    (property) => property.key.name === targetKey,
+  );
+
+  const propertyToAdd = j.objectProperty(
+    j.identifier(targetKey),
+    // Here we use recusion to mix the new value with the current one
```
LukasTy, line 42:
```suggestion
    // Here we use recursion to mix the new value with the current one
```

packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
+/**
```
LukasTy, line 1: Was there a specific reason you chose to author this file as `js` instead of `ts`? 🤔

packages/x-codemod/src/util/addComponentsSlots.js
```diff
@@ -0,0 +1,87 @@
...
+ * @param propName the name of the prop to edit (`'componentsProps'` for example)
+ * @param nestedPath path of the nested object keys with '.' to separate. ('tabs.hidden' for example)
+ * @param value the jscodeshift value to associate to the key
+ * @param j jscodeshift
+ */
+export const transformNestedProp = (elementPath, propName, nestedPath, value, j) => {
+  const initialAttribute = elementPath.value.openingElement.attributes?.find(
+    (attribute) => attribute.name.name === propName,
+  );
+  // The the objec in JSX element
```
LukasTy, line 64:
```suggestion
  // the object in JSX element
```

LukasTy (commented):
Great job on the codemod and the added utils! 👍

packages/x-codemod/README.md
```diff
@@ -90,6 +90,7 @@ The list includes these transformers
 - [`view-components-rename-value-prop`](#view-components-rename)
 - [`localization-provider-rename-locale`](#localization-provider-rename-locale)
 - [`text-props-to-localeText`](#text-props-to-localeText)
+- [`replace-tabs-props](#replace-tabs-props)
```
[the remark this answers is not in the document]
alexfauquette: This codemod is not in `rename-components-to-slots` because it's not a breaking change.

The idea would be to encourage doing it at the end, and include it in the v7 release.

packages/x-codemod/src/util/addComponentsSlots.js
alexfauquette: Too much TS to fix, because I only consider nested objects.

packages/x-codemod/src/v6.0.0/pickers/replace-tabs-props/actual.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+<DateTimePicker
+  componentsProps={{
+    tabs: {
+      dateRangeIcon: <LightModeIcon />,
+    },
+  }}
+>
+  ChildrenString
+</DateTimePicker>;
```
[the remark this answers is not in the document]
alexfauquette: I struggled to make it work with self-closing items, so I preferred to add a tests for both self closing and with children. But effectively, the one with children could be removed since it does not exist 🤭

packages/x-codemod/src/v6.0.0/pickers/replace-tabs-props/actual.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+<DateTimePicker hideTabs={false} dateRangeIcon={<LightModeIcon />} timeIcon={<AcUnitIcon />} />;
```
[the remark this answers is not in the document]
alexfauquette: NIce idea :)

packages/x-codemod/src/util/addComponentsSlots.js
LukasTy: Mhh, fair point. 👌

packages/x-codemod/README.md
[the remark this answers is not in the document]
LukasTy: Hm. It is indeed not a breaking change, but at the same time—a deprecated feature, that can easily be resolved.
I feel that if we make it optional and only mention it in the migration doc—we'd end up with half people using slots and other half sticking with components. 🤔 
But sure, we can try going this route with only suggesting this migration as an optional step. 👌

packages/x-codemod/src/v6.0.0/pickers/replace-tabs-props/expected.spec.js
```diff
@@ -0,0 +1,33 @@
...
+      timeIcon: <AcUnitIcon />,
+    },
+  }} />;
+
+<DateTimePicker
+  componentsProps={{
+    actionBar: {
+      actions: [],
+    },
+
```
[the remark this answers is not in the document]
alexfauquette: I would encourage users to run prettier after the codemod O:)

If you have an idea about how to solve this problem, why not. Otherwise, I prefer not to fix it since writing codemod is time-consuming (I fear half a day of Googling to find how to remove those empty lines)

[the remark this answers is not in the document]
LukasTy: Indeed, I experimented a bit with `lineTerminator` in [recast options](https://github.com/benjamn/recast/blob/master/lib/options.ts), it seems to control it, but there does not seem to be a smart option to control it's value conditionally, nor does it seem to support custom prettier config, where we could specify that empty lines should be removed. 🤔 
TL.DR.: The current solution seems sufficient 👍

PR #7645, opened 2023-01-21
[l10n] Add Czech (cs-CZ) locale

OndrejHj04:
Czech localization for Date Pickers


packages/x-date-pickers/src/locales/index.ts
```diff
@@ -6,6 +6,7 @@ export * from './deDE';
 export * from './esES';
 export * from './faIR';
 export * from './fiFI';
+export * from './csCS';
```
LukasTy, line 9: Correct me if I'm wrong, but I think that the [`BCP 47 language tag` is `cs-CZ`](https://www.techonthenet.com/js/language_tags.php).
This would mean that the file should have a `csCZ` name instead.
Naming it like this would probably also trigger a localization statistics script to pick up this locale, update the data json and include it [in the table](https://deploy-preview-7645--material-ui-x.netlify.app/x/react-date-pickers/localization/#supported-locales).

LukasTy (commented):
Thank you for your contribution! 🙏 
Left a few concerns that I've noticed.

OndrejHj04:
@LukasTy Thank you for your check now it should be alright 🚀

LukasTy:
> Thank you for your check now it should be alright 🚀

It indeed is great now. Thank you once again. 👍 
Thank you for taking the time to contribute and finalize the PR. 🙏

LukasTy (approved):

LukasTy:
@OndrejHj04 I've created a cherry-pick PR to have these translations in the next stable (v5) release.
But the stable branch has [one extra label](https://github.com/mui/mui-x/pull/7666/files#diff-398880cfac026c8566e201502bd8f89b4204d8ded35a78cd599ec31e25f37877R25) that has been removed in v6.
Would you be able to help with translation of it so that the added translations are complete? 🤔

PR #7682, opened 2023-01-23
[DataGrid] Fix browser tooltip for GridCell

### LINKED ISSUE #7671: [DataGrid/GridCell] Browser tooltip support using title attribute
### Duplicates

### Latest version

### Summary 💡

The DataGrid should support the native browser tooltip across supported browsers. Hovering the mouse over a cell should eventually trigger the native browser tooltip. 

This happens on safari as expected(as shown below). However, chrome seems not to have the same behaviour. However, this might be easily achieved by adding a title attribute as that seems to work as expected in chrome.

### Examples 🌈

This is what it looks like in safari currently and this is what I would expect it to look like in chrome as well.

Screenshot 2023-01-23 at 10 49 40

### Motivation 🔦

We are trying to accomplish this as users of out system would like to hover and read the full values in cases where the value is truncated. 

### Order ID 💳 (optional)

_No response_

thupi:
Resolves #7671

### DIFF
--- a/packages/grid/x-data-grid/src/components/cell/GridCell.tsx
+++ b/packages/grid/x-data-grid/src/components/cell/GridCell.tsx
@@ -238,7 +238,12 @@ function GridCell(props: GridCellProps) {
 
   const renderChildren = () => {
     if (children === undefined) {
-      return <div className={classes.content}>{valueToRender?.toString()}</div>;
+      const valueString = valueToRender?.toString();
+      return (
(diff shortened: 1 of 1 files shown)


thupi:
@m4theushw Seems like i did something that didn't look too good on the other PR so made this one that seems a bit cleaner :-) !

oliviertassinari:
One note, WAVE isn't very happy with this change:

Screenshot 2023-05-14 at 01 16 48

PR #7687, opened 2023-01-24
[codemod]  Update pickers component toolbar props 

alexfauquette:
Part #6997


packages/x-codemod/src/v6.0.0/pickers/replace-toolbar-props-by-slot/actual.spec.js
```diff
@@ -0,0 +1,7 @@
+<DatePicker ToolbarComponent={MyToolbar} />;
```
LukasTy, line 1: Try adding this example as well. It reveals a somewhat ugly transformation result of:
```tsx
componentsProps={{
  toolbar: {
    hidden: !false,
  },
}}
```
Maybe we could make sure such case would end up as `hidden: true`? 🤔 
```suggestion
<DatePicker ToolbarComponent={MyToolbar} showToolbar={false} />;
```

LukasTy, line 1: We could also consider adding even more props to make sure the previous ones are left intact. 🤔 
```suggestion
<DatePicker
  ToolbarComponent={MyToolbar}
  showToolbar={false}
  components={{
    ActionBar: CustomActionBar,
  }}
  componentsProps={{
    actionBar: {
      actions: [],
    },
  }} />;
```

packages/x-codemod/src/v6.0.0/pickers/replace-toolbar-props-by-slot/index.ts
```diff
@@ -0,0 +1,92 @@
...
+      });
+    });
+
+  removeProps({ root, componentNames: COMPONENTS, props: Object.keys(propsToSlots), j });
+
+  return renameComponentsSlots({
+    root,
+    componentNames: COMPONENTS,
+    translation: { dateRangeIcon: 'dateIcon' },
+    j,
```
LukasTy, line 90: Should this part be here? 🤔 🙈

LukasTy (commented):
Good job! Leaving a few improvement comments. 😉

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
```diff
@@ -444,7 +444,7 @@ For example, the `ToolbarComponent` has been replaced by a `Toolbar` component s
 
 ### Toolbar (`ToolbarComponent`)
 
-- The `ToolbarComponent` has been replaced by a `Toolbar` component slot:
+- ✅ The `ToolbarComponent` has been replaced by a `Toolbar` component slot:
```
LukasTy, line 447: Maybe we should change these bullet points into sub-sections to add visibility of partial codemod coverage via side menu?
Or maybe come up with an intermediate icon for such section heading denoting that they are partly covered by codemods? 🤔

LukasTy (approved):
Great result! ❤️

docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
alexfauquette: It's inspired from the data grid. I prefer to have the same emoji, such that you know green square means "I can skip up to the next subsection/bullet point"

LukasTy: I was referring to adding a different emoji to `### Toolbar (`ToolbarComponent`)` heading if at least some of it's subsections are covered, like in this case. Just for easier "scanning" when looking at side navigation.
Another mentioned option—change the bullet points into subheadings so that they would appear as submenu items in side nav and become more discoverable. 😃

alexfauquette: I don't get what is the interest of being "discoverable". 

It's more important to do not miss a breaking change not handled by codemod than seeing what is done by codemod

For me the current state says
> `Toolbar` is not handled. Check carefully what remains to be done.

LukasTy: Well, fair point. Maybe it's best not to excite users about having as much green checkmarks as possible, but to let them attentively follow the guide. 👌

MBilalShafi:
I am wondering if you considered adding these use-cases (a little bit advanced ones) to the codemod

### Case 1: When variables are defined separately in the same file

**Actual:**
```tsx
const toolbarPlaceholder = "__";
const toolbarFormat = "DD / MM / YYYY";
const showToolbar = true;

<DatePicker toolbarPlaceholder={toolbarPlaceholder} toolbarFormat={toolbarFormat} showToolbar={showToolbar} />;
```

**Expected:**
```tsx
const toolbarPlaceholder = "__";
const toolbarFormat = "DD / MM / YYYY";
const showToolbar = true;

<DatePicker
  componentsProps={{
    toolbar: {
      toolbarPlaceholder: toolbarPlaceholder,
      toolbarFormat: toolbarFormat,
      hidden: !showToolbar,
    },
  }} />;
```

**Outcome:** This one works as expected 🎉 May be worth adding a couple of snippets to validate it?

### Case 2: When variables are defined as object in same file and used by spread operator

Actual:
```tsx
const pickerProps = {
  toolbarPlaceholder: "__",
  toolbarFormat: "DD / MM / YYYY",
  showToolbar: true,
};

<DatePicker {...pickerProps} />;
```

Expected:
```tsx
const pickerProps = {
  toolbarPlaceholder: "__",
  toolbarFormat: "DD / MM / YYYY",
  showToolbar: true,
};

<DatePicker 
  componentsProps={{
    toolbar: { ...pickerProps }
  }} />;
```

**Outcome:** This does not work as of now (tested with `replace-toolbar-props-by-slot` codemod)

### Case 3: Variables are imported from some other file or passed as props to a wrapper 

**Actual:**
```tsx
// there could be many ways in which these variables could be stored
// 1. Passed as props object
// 2. Passed as some object in props
// 3. Stored as object in state/ref/some other data structure and imported
// 4. Variables deduced using some hook after some business logic is performed etc etc
const SomeWrapperComponent = ({ pickerProps, ...other }) => {
  return <DatePicker {...pickerProps} />;
}
```

**Expected:**
```tsx
const SomeWrapperComponent = ({ pickerProps, ...other }) => {
  return <DatePicker componentsProps={{
    toolbar: { ...pickerProps } // is it even safe to do it? I mean there could be other props that should be used as direct ones
  }} />;
}
```

**Outcome:** Doesn't work too. Not even sure if we want to support such use-cases. 🤔 

### Thoughts: 🤔

1. Should we support more advanced use-cases like case 2 and 3 mentioned above, as they may be used by some of the users?
2. If not, then should we include such codemods to preset safe or add a ✅ in front of them in migration guide? So as an end user, I'll assume that ones with ✅ will be taken care of by codemod but I may not be aware of the advanced use-cases I have to handle myself.
3. Should the same approach be followed in the use cases where we have to change the inner values instead of prop names or prop nested levels like [this](https://github.com/mui/mui-x/pull/7483/files#diff-80496158dc89754e78bdcf78686a1d43934fb4f8d3f6b359785d6c0b313217eaR22-R24) for example. Keeping in mind the values could be stored in many ways like these few:
   - Directly passed in component (simple use case)
   - In some variable or state/ref etc
   - Imported or passed as props
   
Here are the possibilities that I can think of:
- Maybe try and support more advanced use cases
- Clearly mention in README/migration guide that codemods cover the simple usecases (along with some code examples) and that the advanced ones (spread operator, importing from other files, etc, etc) should be tackled by the users themselves.
  - As a dx improvement, we can think about generating the warnings/log about the use cases needed to be handled manually

LukasTy:
Very good and on point observations @MBilalShafi 
This was already discussed via Slack, but just to re-iterate.

> 1\. Should we support more advanced use-cases like case 2 and 3 mentioned above, as they may be used by some of the users?

There might be quite a big chunk of users creating very exotic use cases, but we can't feasibly support everything in the world. IMO, writing codemods at such scale would be a full time job. 🙈 



> 2\. If not, then should we include such codemods to preset safe or add a ✅ in front of them in migration guide?

IMO, we are almost exactly clear in this regard, because we provide an example of the expected diff(s) for each (most) breaking changes. I'd think that it is somewhat self explanatory that you shouldn't expect much codemod coverage beyond the given examples. 🤔 



> 3\. Should the same approach be followed in the codemods where we have to change the inner values instead of prop names or prop nested levels like [this](https://github.com/mui/mui-x/pull/7483/files#diff-80496158dc89754e78bdcf78686a1d43934fb4f8d3f6b359785d6c0b313217eaR22-R24) for example. Keeping in mind the values could be stored in many ways

Indeed, this is a great point you did bring up. For pickers—it should be much less common having many different ways of passing in complex date objects as props, while for `DataGrid`—it might be every single use case, where the approach for the passed data is different. 🤔 
I'd say that if you have the time capacity, it might be worth exploring the feasibility of handling a bit more complex cases, but given the fact that we are working on ASTs here, I'd fear that having robust support would still not be easy to achieve.

> Clearly mention in README/migration guide that codemods cover the simple usecases (along with some code examples)

This is a very good point. It won't hurt, but might help someone a lot and will help clear things up as to what is expected outcome of the codemods! 👌 

> As a dx improvement, we can think about generating the warnings/log about the use cases needed to be handled manually

Could you clear this point up a bit? What do you mean exactly?
Do you mean that we should add a warning when a codemod does not find any code to transform for a given codemod?
Or are you suggesting that we should think of most common possible use cases that are not supported by a codemod and list them via CLI message when running? 🤔

MBilalShafi:
> Do you mean that we should add a warning when a codemod does not find any code to transform for a given codemod?

Not exactly, as this could very easily and frequently be the use-case as the users may not be using the feature or area covered by a specific codemod. Trigerring warning in such use-cases may generate false warnings which may not mean much for many users.

> are you suggesting that we should think of most common possible use cases that are not supported by a codemod and list them via CLI message when running? 

After thinking more on this, I feel this is a viable option, I was initially thinking to do it on per codemod basis.

So if we do:
1. Add a warning in migration guide to explain the use-cases not possibly covered
   - Some thing like `Codemod have some limitation. The main one being object spreading. If you use <DatePicker {...pickerProps} />  in your codebase, codemod will not be able to rename keys.
Don't forget to test you app after upgrading to ensure you have not console errors` as suggested by Alex in slack thread
3. Producing a similar minimal warning in cli on codemod execution. As a lot of people don't bother to read documentation in detail and may have missed that aspect.

It can help provide a bit safer migration experience for the developers in my opinion.

LukasTy:
> So if we do:
> 
> 1. Add a warning in migration guide to explain the use-cases not possibly covered
>    
>    * Some thing like `Codemod have some limitation. The main one being object spreading. If you use <DatePicker {...pickerProps} />  in your codebase, codemod will not be able to rename keys. Don't forget to test you app after upgrading to ensure you have not console errors` as suggested by Alex in slack thread
> 2. Producing a similar minimal warning in cli on codemod execution. As a lot of people don't bother to read documentation in detail and may have missed that aspect.
> 
> It can help provide a bit safer migration experience for the developers in my opinion.

Those are really good points! 👌 
But how would we manage the warning output when users run `preset-safe` instead of an individual codemod, which will possibly be most of the use-cases? 🤔 
Would you suggest only showing one general warning?

MBilalShafi:
> Would you suggest only showing one general warning?

Yes, that could be the simplest approach and probably the most viable too.

Something like this could do maybe?

```tsx
Not all use-cases are covered by codemods. Some advanced cases like props spreading, cross file dependencies etc should be handled manually. 

For example if a codemod renames some props but they are being passed using spread operator they may not be converted as expected.
<DatePicker {...pickerProps} />
<DataGrid {...newProps} />

Make sure to test your app before committing to ensure you don't have any console errors.
```

PR #7698, opened 2023-01-25
[codemod] Basics migration to components props

alexfauquette:
Part of #6997

### DIFF
--- a/packages/x-codemod/README.md
+++ b/packages/x-codemod/README.md
@@ -91,6 +91,7 @@ The list includes these transformers
 - [`localization-provider-rename-locale`](#localization-provider-rename-locale)
 - [`text-props-to-localeText`](#text-props-to-localeText)
 - [`replace-tabs-props](#replace-tabs-props)
+- [`migrate-to-components-componentsProps`](#migrate-to-components-componentsProps)
 
 #### `adapter-change-import`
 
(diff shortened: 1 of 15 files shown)


packages/x-codemod/README.md
```diff
@@ -220,6 +221,56 @@ Replace props used for `Tabs` in DateTime pickers by `componentsProps.tabs` prop
...
+ />
+
+ <DatePicker
+-  TrapFocusProps={{ isEnabled: () => false }}
++  componentsProps={{ desktopTrapFocus: { isEnabled: () => false }}}
+ />
+
+ <DatePicker
+-  InputProps={{ color: 'primary' }}
++  componentsProps={{ input: { InputProps: { color: 'primary' }}}}
```
LukasTy, line 257: This is technically incorrect, because of https://github.com/mui/mui-x/pull/7369 😭 
What do you think:
- should we update existing migration entries to match this change?
- create a new entry and possibly a separate codemod?
I'm leaning towards updating the existing entry, because it's a continuation of a previous breaking change. 🤔

packages/x-codemod/src/util/removeProps.ts
```diff
@@ -16,6 +16,13 @@ export default function removeProps({ root, componentNames, props, j }: RemovePr
     .find(j.JSXAttribute)
     .filter((attribute) => props.includes(attribute.node.name.name as string))
     .forEach((attribute) => {
-      j(attribute).remove();
+      // Test if we are not remove  a prop of a nested component
```
LukasTy, line 19:
```suggestion
      // Test if we are not trying to remove a prop of a nested component
```

packages/x-codemod/src/util/removeProps.ts
```diff
@@ -16,6 +16,13 @@ export default function removeProps({ root, componentNames, props, j }: RemovePr
...
+      const attributeParent = attribute.parentPath.parentPath;
+      if (
+        attributeParent.value.type === 'JSXOpeningElement' &&
+        componentNames.includes(attributeParent.value.name.name)
```
LukasTy, line 23: Shouldn't this be negated, based on the comment above? 🤔

packages/x-codemod/src/v6.0.0/pickers/migrate-to-components-componentsProps/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    trailingComma: true,
+  };
+
+  root
+    .find(j.ImportDeclaration)
+    .filter(({ node }) => {
+      return node.source.value.startsWith('@mui/x-date-picker');
+    })
+
+    .forEach((path) => {
```
LukasTy, line 33: Almost all of this transformation code could probably be re-used by other transformations doing the same or am I missing something? 🤔

alexfauquette: Yes, but I did not find a clear way to do it. The `transformNestedProp` already took a lot of time to have something flexible enough, so I just did some copy pasted

packages/x-codemod/src/util/removeProps.ts
alexfauquette:
```suggestion
      // Only remove props from components in componentNames. Not nested ones.
```

packages/x-codemod/src/util/removeProps.ts
LukasTy: Resolved based on https://github.com/mui/mui-x/pull/7698#discussion_r1086459337

packages/x-codemod/README.md
alexfauquette: Yes, we can replace it by
```
componentsProps={{ textField: { InputProps: { color: 'primary' }}}}
```

People following the alpha might already fix it.

packages/x-codemod/src/util/removeProps.ts
alexfauquette: No, it's the comment that is miss leading. I fixed it

packages/x-codemod/README.md
LukasTy: Could you please update the migration doc with the same diff as well?
We missed it in the original PR... 🙈

LukasTy (approved):
Looking great, good job! 💯

PR #7699, opened 2023-01-25
[codemod] Rename data grid selectors and events

MBilalShafi:
Part of https://github.com/mui/mui-x/issues/7086

### DIFF
--- a/packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
+++ b/packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
@@ -8,7 +8,6 @@ const props = {
   disableMultipleSelection: 'disableMultipleRowSelection',
   showCellRightBorder: 'showCellVerticalBorder',
   showColumnRightBorder: 'showColumnVerticalBorder',
-  headerHeight: 'columnHeaderHeight',
 };
 
 const componentNames = ['DataGrid', 'DataGridPro', 'DataGridPremium'];
--- a/packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
+++ b/packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
@@ -8,6 +8,7 @@ const props = {
   disableMultipleSelection: 'disableMultipleRowSelection',
   showCellRightBorder: 'showCellVerticalBorder',
   showColumnRightBorder: 'showColumnVerticalBorder',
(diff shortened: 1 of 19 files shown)


packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
```diff
@@ -8,6 +8,7 @@ const props = {
   disableMultipleSelection: 'disableMultipleRowSelection',
   showCellRightBorder: 'showCellVerticalBorder',
   showColumnRightBorder: 'showColumnVerticalBorder',
+  headerHeight: 'columnHeaderHeight',
```
LukasTy, line 11: Nitpick: We could also consider delivering such improvements via a separate PR, it would also help with increased visibility and discoverability if anyone is following the changelog. 😉

LukasTy (approved):
LGTM. But I'm not an expert in the possible usage of mentioned breaking changes and I could not know other possible usages that could not have been covered by the transformation. 🙈

packages/x-codemod/src/v6.0.0/data-grid/row-selection-props-rename/index.ts
MBilalShafi: This specific change fell under same category which was previously covered by a codemod and was a minor one, so I pushed it along with this one, but you are right, delivering under separate ones might be more helpful. 👍

LukasTy: I was thinking about both this change as well as the `GridColumnMenuItemProps` rename. 😉

MBilalShafi: Ah ok, you are right actually 😄 
Let me split this into respective ones

LukasTy: No pressure. We are still on initial beta releases, it's not that the package is widely used. 🙈

MBilalShafi: No no, it's a good idea. It'll be easier for us too to track the changes this way.
Thanks for the useful Nitpick 😉

MBilalShafi: Split respective changes into https://github.com/mui/mui-x/pull/7710 and https://github.com/mui/mui-x/pull/7711

PR #7707, opened 2023-01-26
[codemod] Rename linkOperators to logicOperators

MBilalShafi:
Codemod for change done in https://github.com/mui/mui-x/pull/7310
Part of https://github.com/mui/mui-x/issues/7086


packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/actual.spec.js
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react'
```
MBilalShafi, line 1: Could there be other ways (apart from those mentioned in this file) in which breaking changes from [this PR](https://github.com/mui/mui-x/pull/7310) could be used 🤔

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    trailingComma: true,
+  };
+
+  const matchingImports = root.find(j.ImportDeclaration).filter((path) => !!matchImport(path));
+
+  if (matchingImports.length > 0) {
+    // Rename the identifiers
+    // - import { linkOperatorInputProps } from '@mui/x-data-grid'
+    // + import { logicOperatorInputProps } from '@mui/x-data-grid'
+    root
```
MBilalShafi, line 42: Followed bulk rename approach with an additional check, i.e checking if we have some imports from `grid` packages.

The main reason is to improve the impact of codemod assuming that most percentages of the users are not likely to use the same variable names (as defined on top of this file), especially when using DataGrid in the file too.
Because if we limit it to only inside `DataGrid(Pro|Premium)` nodes it will miss most common use-cases like https://github.com/mui/mui-x/blob/3c7c7e7b86cbd83cc6de359f866690ae41d15ba6/packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/actual.spec.js#L5-L20

The failure cases should be covered by the message we are going to add in the migration guide for the user to check out for console errors after executing the codemods. [Context](https://github.com/mui/mui-x/pull/7687#issuecomment-1411747094)

What do you think @alexfauquette @LukasTy ?

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  'MuiDataGrid-filterFormLinkOperatorInput': 'MuiDataGrid-filterFormLogicOperatorInput',
+  'MuiDataGrid-withBorder': 'MuiDataGrid-withBorderColor',
+};
+
+const PACKAGE_REGEXP = /@mui\/x-data-grid(-pro|-premium)?/;
+
+const matchImport = (path: ASTPath<ImportDeclaration>) =>
+  ((path.node.source.value as any).toString() || '').match(PACKAGE_REGEXP);
+
+// Press ctrl+space for code completion
```
LukasTy, line 26: Was it left behind accidentally? 🤔

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    quote: 'single',
+    trailingComma: true,
+  };
+
+  const matchingImports = root.find(j.ImportDeclaration).filter((path) => !!matchImport(path));
+
+  if (matchingImports.length > 0) {
+    // Rename the identifiers
+    // - import { linkOperatorInputProps } from '@mui/x-data-grid'
+    // + import { logicOperatorInputProps } from '@mui/x-data-grid'
```
LukasTy, line 41: Am I missing something, but I don't see such case in tests, nor codemod readme diff... 🙈 🤷

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
LukasTy: I'd say that in this case for most transformations your approach should work nicely. 👌 
But I'd fear that `linkOperator` and `linkOperators` identifiers might be a bit too common for certain cases and could end up in certain false-positives. 🤔 

Also, if we'd go with this approach for this codemod, shouldn't we also consider doing the same with every existing codemod? At least the `DataGrid` ones? 🤔

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    // - apiRef.current.getLocaleText('filterPanelLinkOperator')
+    // + apiRef.current.getLocaleText('filterPanelLogicOperator')
+    root
+      .find(j.Literal)
+      .filter((path) => !!renamedLiterals[path.node.value as any])
+      .replaceWith((path) => j.literal(renamedLiterals[path.node.value as any]));
+
+    // Rename the classes
+    // - 'MuiDataGrid-filterFormLinkOperatorInput'
+    // + 'MuiDataGrid-filterFormLogicOperatorInput'
```
LukasTy, line 57: Have you considered/tried using `renameClassKey` utility for this transform? 🤔

LukasTy (approved):
The changes look great overall to me! 👌 

My couple of concerns are:
- the fact that we can end up with false-positive transforms if we end up in a case, where the code having the same literal has been transformed, just because a file had `@mui/x-data-grid` import, even though the change itself is unrelated to any components from this package
- the possible performance implications given a big codebase. We haven't done any performance measurements, but I do have a feeling that this kind of codemod should be slower than one narrowing down the cases, where it needs to be applied, based on jsx component node names 🤔

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
MBilalShafi: Yep, thanks for identifying 🙈

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
MBilalShafi: Correct, updated!

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/actual.spec.js
alexfauquette: Unlikely but feasible: If they have a custom filter panel, the link operator will be in the props they reuse

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
MBilalShafi: > But I'd fear that linkOperator and linkOperators identifiers might be a bit too common for certain cases and could end up in certain false-positives. 

That's actually a really good point 👍 
Reflecting again on this I came up with [this](https://github.com/mui/mui-x/blob/e0d7f4101f1b9df484687658f4db9bdbf4e40e8d/packages/x-codemod/src/util/renameIdentifiers.ts) utility method for renaming identifiers in a much safer way.

It accepts an optional `preRequisiteUsages` parameters using which we can narrow-down the identifiers we rename based on certain pre-requisites like:
1. Some imports must be done from the packages identified by `PreRequisiteUsage.packageRegex` e.g `import { something } 'x-data-grid'`
2. Some specific components from those mentioned in `PreRequisiteUsage.components` must be imported from the specified packages using 
3. Specific nested paths must be used in the components imported from specified packages using `PreRequisiteUsage.possiblePaths`. 


**Example:**
Below example will only rename `GridLinkOperator` if:
1. There's at least one import from 'x-data-grid(pro|premium)' packages.
2. Atleast one of `components` are imported from 'x-data-grid(pro|premium)' imports.
3. Atleast one of the `components` is used in JSX and must satisfy at least one of `possiblePaths`.
```tsx
const preRequisiteUsages = {
    GridLinkOperator: {
        possiblePaths: ['initialState.filter', 'filterModel', 'componentsProps.filter'],
        components: ['DataGrid', 'DataGridPro', 'DataGridPremium'],
        packageRegex: /@mui\/x-data-grid(-pro|-premium)?/,
    }
  }

renameIdentifier({ ...otherProps, preRequisiteUsages })
```

IMO, this should solve > 90% of false positives as it allows to drill down as deep as you want in the AST. And in most of the cases, even if the usages are outside the JSX nodes, we can guess if it should be updated or not based on usages in DataGrid component or whatever component we're writing codemod for.

We can also use it in other areas like https://github.com/mui/mui-x/pull/7483 to solve use case 2 efficiently.

MBilalShafi:
Thanks @LukasTy for putting in so useful points 🙏 

> the fact that we can end up with false-positive transforms if we end up in a case, where the code having the same literal has been transformed, just because a file had @mui/x-data-grid import, even though the change itself is unrelated to any components from this package

Please check https://github.com/mui/mui-x/pull/7707#discussion_r1095537690

> the possible performance implications given a big codebase. We haven't done any performance measurements, but I do have a feeling that this kind of codemod should be slower than one narrowing down the cases, where it needs to be applied, based on jsx component node names

This point is also very good but not that worrisome to me as I believe for a normal application, migration step is not that frequently happening, it may happen on a major version upgrade (ideally once a year), even if users have to wait a bit while running npx codemod, it should be OK (I am thinking about `yarn install` which also takes some time but it's kinda ok to run it once in a while).

Still, I think it's worth it to do some benchmarking to get a clear picture of how well our whole codemod suite performs. I'll try to grab some heavy application that is using `mui-x` or will create a new one and run some tests on it. (If you know about good candidate for this, please pass that on to me) 😉

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
MBilalShafi: Yes, actually it doesn't work great in out use-case as it's targeting single component and `classes` prop. For our usecase we want to handle `sx` prop and legacy styling like this too:
https://github.com/mui/mui-x/blob/e0d7f4101f1b9df484687658f4db9bdbf4e40e8d/packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/actual.spec.js#L5-L10

Will try to add this new (more generic) one to the util or replace the other with this.

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/actual.spec.js
MBilalShafi: Ah ok, that could be one of the use cases, thanks for the input.

I think as long as the other component is defined in the same file it should work with props even, [added](https://github.com/mui/mui-x/pull/7707/commits/d3617d9c80c4e1f8793f401bde088c6d92709bd3) another snippet to validate that.

But the one defined in another file is a limitations of the codemods and it must be handled manually by users.

packages/x-codemod/src/v6.0.0/data-grid/rename-linkOperators-logicOperators/index.ts
LukasTy: Yeah, by no means, feel free to update the existing util with the required logic.
I've only left the existing util in the codebase after cloning MUI Core codemod, because I thought that we'd have use for it, but so far it hasn't been used once. 🙈

PR #7709, opened 2023-01-26
[codemod] Rename `shouldDisableTime` prop

alexfauquette:
Part of #6997

### DIFF
--- a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
+++ b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
@@ -1,0 +1,5 @@
+import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+
+<div>
+  <DateTimePicker shouldDisableTime={(timeValue, view) => view === 'hours' && timeValue < 12} />
+</div>;
--- a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
+++ b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
@@ -1,5 +1,9 @@
 import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+import { TimePicker } from '@mui/x-date-pickers/TimePicker';
+
+const shouldDisable = (timeValue, view) => view === 'hours' && timeValue < 12;
 
 <div>
   <DateTimePicker shouldDisableTime={(timeValue, view) => view === 'hours' && timeValue < 12} />
+  <TimePicker shouldDisableTime={shouldDisable} />
 </div>;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -342,7 +342,7 @@ We'd like to offer a big thanks to the 9 contributors who made this release poss
   ```diff
    <DateTimePicker
   -   shouldDisableTime={(timeValue, view) => view === 'hours' && timeValue < 12}
-  +   shouldDisableTime={(time, view) => view === 'hours' && value.hour() < 12}
+  +   shouldDisableTime={(value, view) => view === 'hours' && value.hour() < 12}
    />
   ```
 
--- a/docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
+++ b/docs/data/migration/migration-pickers-v5/migration-pickers-v5.md
@@ -138,7 +138,7 @@ The picker components no longer have a keyboard view to render the input inside 
 At some point, the mobile pickers should have a prop allowing to have an editable field without opening the modal.
 :::
 
-### Rename `shouldDisableTime` prop
+### ✅ Rename `shouldDisableTime` prop
 
 The `shouldDisableTime` prop signature has been changed. Either rename the prop usage to `shouldDisableClock` or refactor usage.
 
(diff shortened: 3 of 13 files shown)


packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
```diff
@@ -0,0 +1,5 @@
+import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+
+<div>
+  <DateTimePicker shouldDisableTime={(timeValue, view) => view === 'hours' && timeValue < 12} />
```
LukasTy, line 4: We could also introduce a test case with function as a variable just for the sake wider coverage. 🙈 

```diff
diff --git a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
index 898616043..bf32e31b8 100644
--- a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
+++ b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
@@ -1,5 +1,9 @@
 import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+import { TimePicker } from '@mui/x-date-pickers/TimePicker';
+
+const shouldDisable = (timeValue, view) => view === 'hours' && timeValue < 12;
 
 <div>
   <DateTimePicker shouldDisableTime={(timeValue, view) => view === 'hours' && timeValue < 12} />
+  <TimePicker shouldDisableTime={shouldDisable} />
 </div>;
diff --git a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/expected-props.spec.js b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/expected-props.spec.js
index 784d823db..f3e13aeaf 100644
--- a/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/expected-props.spec.js
+++ b/packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/expected-props.spec.js
@@ -1,5 +1,9 @@
 import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
+import { TimePicker } from '@mui/x-date-pickers/TimePicker';
+
+const shouldDisable = (timeValue, view) => view === 'hours' && timeValue < 12;
 
 <div>
   <DateTimePicker shouldDisableClock={(timeValue, view) => view === 'hours' && timeValue < 12} />
+  <TimePicker shouldDisableClock={shouldDisable} />
 </div>;

```

LukasTy (approved):
Looks great! 💯 
Just a general question: do you know if such thing is feasible with `jscodeshift`:
Having:
```tsx
const props = {
  shouldDisableTime: (timeValue, view) => view === 'hours' && timeValue < 12,
};

<TimePicker {...props} />
```
transform it into:
```tsx
const props = {
  shouldDisableClock: (timeValue, view) => view === 'hours' && timeValue < 12,
};

<TimePicker {...props} />
```

Because our current codemods will only cover users directly using pickers from our package as well as specifying props directly on the component, but I fear that in the real world we might have quite a lot of users doing the following:
- having a wrapper component on top of ours with additional logic/defaults
- spreading the props from some other object onto the picker component

packages/x-codemod/src/v6.0.0/pickers/rename-should-disable-time/actual-props.spec.js
alexfauquette: If you feel safer with it, I add it with pleasure :)

alexfauquette:
@LukasTy I don't think it's feasible with `jscodeshift` because we use AST which represent the code as tokens, but do not set any logic between them.


For example here is a simple code and it's associated AST in JSON (I removed most of the useless parts)
```jsx
const prop = { label: "Name" }
<A {...props} />
```

```json
{
  "program": {
    "type": "Program",
    "body": [
      {
        "type": "VariableDeclaration",
        "declarations": [
          {
            "type": "VariableDeclarator",
            "id": {
              "type": "Identifier",
              "name": "props"
            },
            "init": {
              "type": "ObjectExpression",
              "properties": [
                {
                  "type": "Property",
                  "key": {
                    "type": "Identifier",
                    "name": "label"
                  },
                  "computed": false,
                  "value": {
                    "type": "Literal",
                    "value": "Nale",
                    "raw": "'Name'"
                  },
                  "kind": "init",
                  "method": false,
                  "shorthand": false
                }
              ]
            }
          }
        ],
        "kind": "const"
      },
      {
        "type": "ExpressionStatement",
        "expression": {
          "type": "JSXElement",
          "openingElement": {
            "type": "JSXOpeningElement",
            "name": {
              "type": "JSXIdentifier",
              "name": "A"
            },
            "selfClosing": true,
            "attributes": [
              {
                "type": "JSXSpreadAttribute",
                "argument": {
                  "type": "Identifier",
                  "name": "props"
                }
              }
            ]
          },
          "children": [],
          "closingElement": null
        }
      }
    ]
  }
}
```

The tree is made of two parts:
- `type: "VariableDeclaration"` which declares the definition of `props`
- `type: "ExpressionStatement"` that defines the components

The only way to get the link between them is to notice that the first one defines an object of type `"Identifier"` with the name `"props"`  which has the same name as the one used in the spread operator.

As you can imagine, this can not scale since you would have to consider all the possible ways to edit an object. Same if objects are imported (we only parse one file at a time)

LukasTy:
@alexfauquette Thank you for a detailed answer. 🙏

PR #7721, opened 2023-01-26
[pickers] Change mobile dialog close behavior

### LINKED ISSUE #7700: [pickers] Return the previous value if the user click's out the MobileDatePicker Dialog
### Duplicates

### Latest version

### Steps to reproduce 🕹

Link to live example: https://mui.com/x/react-date-pickers/date-picker/ => MobileDatePicker

1. click on the mobile date picker => select value => close => ` the value will return as it was `
2. click on the mobile date picker => select value => Ok => ` I will see the updated value `

The bug:
3. click on the mobile date picker => select value => and click outside => ` I will see the updated value `

### Current behavior 😯

Click on the mobile date picker => select value => and click outside => ` I will see the updated value `

### Expected behavior 🤔

Click on the mobile date picker => select value => and click outside => ` the value will return as it was `? 

because I didn't accept the new value 

### Context 🔦

when I click close, onChange gets called again and with the old date, so i can return the old value 
we can call the onChange `onClickOutSide` so the old value will return again with no change 

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #7700 

- Add `resetValueOnClose` property, which will reset the value to `resetFallback` (same as when clicking cancel) when closing modal.
- Default the value to `true` on mobile pickers so that clicking away or `Esc` would reset the value (seems to make sense, given that pickers do have `Cancel` and `Accept` actions in this case)

The only relevant change in in [`usePickerValue` file](https://github.com/mui/mui-x/pull/7721/files#diff-885d7c8b398de2634901fe80ba13c539c8da220b470d723c1945ed2d9a869d6b).


packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
```diff
@@ -429,7 +437,10 @@ export const usePickerValue = <
   const handleDismiss = useEventCallback(() => {
     // Set all dates in state to equal the last committed date.
     // e.g. Reset the state to the last committed value.
-    setDate({ value: dateState.committed, action: 'acceptAndClose' });
+    setDate({
+      value: resetValueOnClose ? dateState.resetFallback : dateState.committed,
+      action: 'acceptAndClose',
+    });
   });
```
alexfauquette, line 444: I don't get why the modification is here. I agree "dismiss" is mostly used when closing UI. But I doubt it's a good idea to have `onDismiss` behavior depending on `resetValueOnClose` especially if at some point we open access to hook. 

For me, it would make more sense that depending on `resetValueOnClose` we either call `onDismiss` or `onCancel`

LukasTy: Very good point. 👌

LukasTy: [Updated](https://github.com/mui/mui-x/pull/7721/commits/e6075da46b352fc125a1f8cad322a75c4dc48814)

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
```diff
@@ -206,6 +206,13 @@ export interface UsePickerValueNonStaticProps<TValue>
    * Use in controlled mode (see open).
    */
   onOpen?: () => void;
+  /**
+   * If `true`, on dismiss will reset the value to the value when the picker was opened.
+   *
+   * Might make most sense when in need of resetting the value on modal dismiss.
+   * @default `true` for mobile, `false` for desktop.
```
alexfauquette, line 213: The default value of a boolean prop should be `false`, such that you end up with either

`<Components />` or `<Component resetValueOnDismiss />` but not `<Component resetValueOnDismiss={false} />`

alexfauquette (commented):
With the new naming it makes sens, @flaviendelangle do you see coming features with which it could interfere?

packages/x-date-pickers/src/internals/hooks/usePicker/usePickerValue.ts
LukasTy: Thank you for clearing this thing up. 👍 
I've updated it, do you think it is now in a correct form? 🤔

flaviendelangle:
The behavior of the pickers on close is a topic that keeps getting discussed, I would be very cautious before introducing any new prop, to make sure that we are solving peoples pain and not just fine tuning the use case of some specific user.

I'll try to list the issues around this topic to check :+1:

LukasTy:
> I would be very cautious before introducing any new prop, to make sure that we are solving peoples pain and not just fine tuning the use case of some specific user.

I do understand and agree.
But I think that we either go with a complete overhaul of how the on change works (is propagated) or we try to cater to more use cases.
Would you agree that the current behaviour of accepting the selected value on picker dialog dismiss (either by clicking on backdrop or especially by clicking <kbd>Esc</kbd>) is somewhat weird and very opinionated. I would expect quite a lot of users wanting the `acceptValueOnDismiss={false}` behavior even on desktop. 🤔

flaviendelangle:
> Would you agree that the current behaviour of accepting the selected value on picker dialog dismiss (either by clicking on backdrop or especially by clicking Esc) is somewhat weird and very opinionated

Yes, I think people should be able to save or not when closing.

We have 2 topics here

## The default behavior

### Desktop

Saving when closing on desktop is what Spectrum is doing for example.
On Telerik however, the date is not filled at all before the last view is committed (by explicitly clicking on `Set`) so closing will reset.

Most of the discussions around this topic are linked in #4408
For example, this comment argues for the validation on close https://github.com/mui/mui-x/pull/4408#issuecomment-1103856668
And the team agreed at the time to save when closing on desktop.

This specific use case is probably the one with the smallest majority on any behavior.

### Mobile

It is the 1st feedback we get of someone wanting to reset the value (as far as I remember).
So I don't think we should the default behavior here (and you did not propose it so I think we agree on that one).

## How to let people choose another behavior

When we started to get feedback about the life-cycle, it became clear that is a very subjective topic and that if we want to make everyone happy, we need a way to customize those behaviors.
At the time, it decided to postpone the work, to let us learn better those behaviors and build something for the long term. I fear that otherwise we add half a dozen boolean props to fine tune the various life-cycle behaviors, leading to a very poor DX.

One API I had in mind was to let people override part of `UsePickerValueActions`, maybe with names describing actions more than behavior.

For example:

```tsx
<DatePicker 
  lifecycleManager={({ setDate, dateState }) => ({ 
    onDimiss: (source: 'escape' | 'click-outside') => {
      setDate({ value: dateState.resetFallback, action: 'acceptAndClose' });
    }
  }})
/>
```

But that's a low level API that can easily be done in a poor way and end up being worse that all those booleans props.

____

In summary, it is very unclear to me what the default behaviors should be and how people should be able to override them.
I think it's a topic for which we need an umbrella issue and a description on the various use cases people a struggling with.

LukasTy:
> One API I had in mind was to let people override part of UsePickerValueActions, maybe with names describing actions more than behavior.

> But that's a low level API that can easily be done in a poor way and end up being worse that all those booleans props.

Indeed, your suggested API does look very powerful and fully customizable, but at the same time, as you mentioned, it could be a bit too powerful and a bit too cumbersome to use for most. Unless we provide a more user friendly option.

I was also thinking about at least having a hybrid prop accepting both a boolean as well as a callback `shouldAcceptValueOnDismiss: boolean | (dateState, wrapperVariant) => boolean`.

Did we ever consider having hybrid props serving both a more basic case as well as a more customized one?

flaviendelangle:
> Did we ever consider having hybrid props serving both a more basic case as well as a more customized one?

It's not something I saw on our components, but it's something I saw on other set of components.

PR #7728, opened 2023-01-27
[Experiment] Testing codemod behavior

alexfauquette:


packages/x-codemod/src/v6.0.0/pickers/experiment/actual.spec.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
danieldelcore, line 1: Hey @alexfauquette, do you have any examples of how your pickers change between versions? Maybe a diff? I can see if I can quickly mock something up for you 😊

alexfauquette: Most of the modifications have already been done:
- pickers #6997
- data grid #7086

The modification I proposed in codeshift comunity comes from [`migrate-to-components-componentsProps`](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/pickers/migrate-to-components-componentsProps)
which work well with self-closing elements

But I realized it was a problem when adding the transformation to our [`preset-safe`](https://github.com/mui/mui-x/tree/next/packages/x-codemod/src/v6.0.0/pickers/preset-safe)

That's why I'm proposing to add a filtering condition on the parent
```jsx
componentNames.includes(attributeParent.value.name.name)
```

PR #7729, opened 2023-01-27
[docs] Rename `next` translation docs and remove duplicates with `-next`

LukasTy:
These there leftover docs files having `-next` in their name.
- Main doc files were just duplicates, which were removed
- `pt` and `zh` files were somehow left un-renamed
- `pt` and `zh` files do no longer seem to be updated currently, but I believe we agreed not to remove them from code for now. 🙈


alexfauquette (approved):
Not sure about what's the current policy fro translation files, but this cane can not hurt 👍

flaviendelangle (approved):

PR #7732, opened 2023-01-27
[TimePicker] Add missing `themeAugmentation` entry

### LINKED ISSUE #7730: [TimePicker] not included in `themeAugmentation`
### Duplicates

### Latest version

### Steps to reproduce 🕹

Try to add `MuiTimePicker` to theme.

https://codesandbox.io/s/eloquent-wildflower-c7m4ts?file=/demo.tsx

### Current behavior 😯

```
 Object literal may only specify known properties, but 'MuiTimePicker' does not exist in type 'Components<Omit<Theme, "components">>'. Did you mean to write 'MuiDatePicker'?ts(2322)
```

image

### Expected behavior 🤔

`MuiTimePicker` should be defined in the types as with the other pickers.

### Context 🔦

Related to #7674 

### Your environment 🌎

_No response_

### Order ID 💳 (optional)

_No response_

LukasTy:
Fixes #7730

### DIFF
--- a/packages/x-date-pickers/src/themeAugmentation/components.d.ts
+++ b/packages/x-date-pickers/src/themeAugmentation/components.d.ts
@@ -161,6 +161,11 @@ export interface PickerComponents<Theme = unknown> {
     styleOverrides?: ComponentsOverrides<Theme>['MuiStaticTimePicker'];
     variants?: ComponentsVariants['MuiStaticTimePicker'];
   };
+  MuiTimePicker?: {
+    defaultProps?: ComponentsProps['MuiTimePicker'];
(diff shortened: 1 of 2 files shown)


alexfauquette (approved):
I'm wondering if it would not be easier to automate the generation of those files

LukasTy:
> I'm wondering if it would not be easier to automate the generation of those files

Yup, I believe we could look into that. 🙈 
I was looking into coming up with a type, which would enforce themeAugmentation.spec to contain all the components from `props` file, but that is tedious to do and could still allow for human error 🤔

LukasTy:
Ended up with a different approach (https://github.com/mui/mui-x/pull/7808) preventing the ability to introduce errors in `d.ts` files.

PR #7734, opened 2023-01-27
[docs] Fix hydration crash in pickers

### LINKED ISSUE #7439: [pickers] Hydration crash
### Duplicates

### Latest version

### Steps to reproduce 🕹

Link to live example:

Steps:
1. Open http://0.0.0.0:3001/x/react-date-pickers/date-picker/

### Current behavior 😯

Screenshot 2023-01-08 at 16 54 23

### Expected behavior 🤔

No errors

### Context 🔦

I'm working on the mobile experience. It's different from #5856. This time, it's not the time that changes. It fails in localhost and with the id.

### Your environment 🌎

v6.0.0-alpha.14

### Order ID 💳 (optional)

_No response_

oliviertassinari:
Closes #7439 (https://github.com/mui/mui-x/issues/7439#issuecomment-1406630387)

https://deploy-preview-7734--material-ui-x.netlify.app/x/react-date-pickers/base-concepts/

### DIFF
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -107,8 +107,8 @@ export default function ComponentExplorerNoSnap() {
 
       return isPro
         ? `
-import { ${exportedName} } from '@mui/x-date-pickers-pro';
-import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';`
+import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';
+import { ${exportedName} } from '@mui/x-date-pickers-pro';`
         : `
 import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
 import { ${exportedName} } from '@mui/x-date-pickers';
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -106,13 +106,15 @@ export default function ComponentExplorerNoSnap() {
       const isPro = cleanName.includes('Range');
 
       return isPro
-        ? `import { ${exportedName} } from '@mui/x-date-pickers-pro';
+        ? `
+import { ${exportedName} } from '@mui/x-date-pickers-pro';
 import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';`
-        : `import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
+        : `
+import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
 import { ${exportedName} } from '@mui/x-date-pickers';
 import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
     })
-    .join('\n\n');
+    .join('\n');
 
   const content = exportedNames.map((exportedName) => {
     const Component = exportedElements[
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -202,7 +202,7 @@ import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
           </Select>
         </FormControl>
       </Stack>
-      {exportedNames.length > 0 && (
+      {exportedNames.length > 0 ? (
         <React.Fragment>
           <div>
             {docPages.map((docPage) => (
@@ -224,8 +224,7 @@ import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
             </LocalizationProvider>
           </Stack>
         </React.Fragment>
-      )}
-      {exportedNames.length === 0 && (
+      ) : (
         <Typography>This component is not available yet</Typography>
       )}
     </Stack>
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -107,14 +107,12 @@ export default function ComponentExplorerNoSnap() {
 
       return isPro
         ? `
-import { ${exportedName} } from '@mui/x-date-pickers-pro'    
-import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}'    
-`
+import { ${exportedName} } from '@mui/x-date-pickers-pro';
+import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';`
         : `
-import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}'    
-import { ${exportedName} } from '@mui/x-date-pickers'
-import { ${exportedName} } from '@mui/x-date-pickers-pro'    
-`;
+import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
+import { ${exportedName} } from '@mui/x-date-pickers';
+import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
     })
     .join('\n');
 
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -123,6 +123,7 @@ import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
 
     return (
       <DemoItem
+        key={exportedName}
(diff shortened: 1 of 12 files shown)


docs/data/date-pickers/base-concepts/ControlledComponent.tsx
```diff
@@ -11,7 +11,12 @@ export default function ControlledComponent() {
   return (
     <LocalizationProvider dateAdapter={AdapterDayjs}>
       <DemoContainer components={['DatePicker']}>
-        <DatePicker value={value} onChange={(newValue) => setValue(newValue)} />
+        <DatePicker
+          value={value}
+          onChange={(newValue) => {
+            setValue(newValue);
```
oliviertassinari, line 17: Make it clear that there is no return value (for the noobies)

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -13,7 +13,7 @@ waiAria: https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/datepicker-d
 
 {{"component": "modules/components/ComponentLinkHeader.js"}}
 
-## Basic usage
+## Controlled value
```
oliviertassinari, line 16: To be clearer? Developers don't have to control the `value`.

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -56,7 +56,7 @@ For example, the `DatePicker` is the combination of the `DateField` and the `Dat
 The _Calendar / Clock_ components are rendered inside a _Popover_ on desktop and inside a _Modal_ on mobile.
 :::
 
-### Date or time editing ?
+### Date or time editing?
```
oliviertassinari, line 59: French mistake

docs/data/date-pickers/base-concepts/base-concepts.md
```diff
@@ -81,7 +81,7 @@ Each _Picker_ is available in a responsive, desktop and mobile variant:
 
 There are many components available, each fitting specific use cases. Use the form below to find the component you need:
 
-{{"demo": "ComponentExplorerNoSnap.js", "hideToolbar": true}}
+{{"demo": "ComponentExplorerNoSnap.js", "hideToolbar": true, "bg": "inline"}}
```
oliviertassinari, line 84: I assume it was forgotten, the other demos have it.

docs/data/date-pickers/base-concepts/ComponentFamilies.tsx
```diff
@@ -14,7 +14,7 @@ import Tooltip from '@mui/material/Tooltip';
 
 function ProLabel({ children }: { children: React.ReactNode }) {
   return (
-    <Stack direction="row" spacing={0.5}>
+    <Stack direction="row" spacing={0.5} component="span">
```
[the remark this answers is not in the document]
LukasTy: Are you sure that this should close the original issue?
I believe that this page was introduced after the issue creation. And the original hydration issue is mentioned to happen in `date-picker` page. 🤔

docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
```diff
@@ -107,14 +107,12 @@ export default function ComponentExplorerNoSnap() {
 
       return isPro
         ? `
-import { ${exportedName} } from '@mui/x-date-pickers-pro'    
-import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}'    
-`
+import { ${exportedName} } from '@mui/x-date-pickers-pro';
+import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';`
```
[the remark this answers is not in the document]
LukasTy: ❤️ But maybe we could also move "`" symbol from L109 to L110 thus avoiding the first empty line when copy-pasting these examples? 🤔

docs/data/date-pickers/base-concepts/ControlledComponent.tsx
LukasTy: Or go the other—minimalism way. 🙈 😆 
```tsx
onChange={setValue}
```
In any case, this only changes one example, albeit, in the the getting started page. Do you think it is enough and leaving others as is doesn't hurt? 🤔

docs/data/date-pickers/base-concepts/base-concepts.md
LukasTy: I'm not so sure. The only other demos without a border are the `basic` demos showing small examples between paragraphs. This is possibly a huge demo, which might not even fit vertically to a screen. I feel that having the border ties the demo together and is more in line with most of the other examples.

docs/data/date-pickers/base-concepts/ComponentFamilies.tsx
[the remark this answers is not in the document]
oliviertassinari: I can't reproduce the other hydration issue.

docs/data/date-pickers/base-concepts/base-concepts.md
oliviertassinari:
```suggestion
{{"demo": "ComponentExplorerNoSnap.js", "hideToolbar": true}}
```

docs/data/date-pickers/base-concepts/ControlledComponent.tsx
oliviertassinari: > onChange={setValue}

I don't think that we can do this. I would expect more arguments, starting with the `event`.

> Do you think it is enough and leaving others as is doesn't hurt?

I was thinking of one change at a time. It's not a huge difference, but I tend to make this change anytime I came across it.

docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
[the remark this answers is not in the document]
oliviertassinari: I think that the line break is clear: 
Screenshot 2023-01-28 at 02 03 29

docs/data/date-pickers/base-concepts/ControlledComponent.tsx
LukasTy: > I don't think that we can do this. I would expect more arguments, starting with the `event`, and later the reason why the onChange event triggers.

Fair point. We are also providing the context as the second argument, possibly containing validation errors.

> It's not a huge difference, but I tend to make this change anytime I came across it.

I've gone through all v6 documentation for pickers and couldn't find a single case where we are using your suggested format.
I do understand the reason why you'd prefer it, but I believe that @flaviendelangle has chosen the shorter format to be have shorter demo previews.
IMHO, if we change the format, I'd vote for changing every usage of it to have consistency.

docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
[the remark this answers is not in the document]
LukasTy: I was referring to the fact that both the original version as well as your case has leading empty/new line in the pasted content. 🙈 
Screenshot 2023-01-28 at 22 32 22

I was thinking about something like this to avoid this empty line. 🤔 
```diff
diff --git a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
index 5e102404b..6f27ac442 100644
--- a/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
+++ b/docs/data/date-pickers/base-concepts/ComponentExplorerNoSnap.tsx
@@ -106,15 +106,13 @@ export default function ComponentExplorerNoSnap() {
       const isPro = cleanName.includes('Range');
 
       return isPro
-        ? `
-import { ${exportedName} } from '@mui/x-date-pickers-pro';
+        ? `import { ${exportedName} } from '@mui/x-date-pickers-pro';
 import { ${exportedName} } from '@mui/x-date-pickers-pro/${subPackage}';`
-        : `
-import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
+        : `import { ${exportedName} } from '@mui/x-date-pickers/${subPackage}';
 import { ${exportedName} } from '@mui/x-date-pickers';
 import { ${exportedName} } from '@mui/x-date-pickers-pro';`;
     })
-    .join('\n');
+    .join('\n\n');
 
   const content = exportedNames.map((exportedName) => {
     const Component = exportedElements[
```

docs/data/date-pickers/base-concepts/ControlledComponent.tsx
oliviertassinari: In the pickers maybe, but I I think that in the overall scope, we default to be explicit about the returned values, e.g. https://mui.com/material-ui/react-text-field/#uncontrolled-vs-controlled. I'm reverting here as it's outside of the scope of the bug fix.

